PR #6, opened 2019-11-25
Feature: Ability to enable and disable sections based on when rules

MebinAbraham:
Added the ability to enable/disable sections based on when rule evaluation.

`PathFinder` now has a new `enabled_section_ids` property which returns a list of enabled sections. I'm open to suggestions on whether we should create a new class for this, i.e. ` SectionPathFinder and SurveyPathFinder`.

I have also removed the pathfinder helper (LocalProxy) in favour of instantiating in the place it is needed. The `Router` and `HubContext` classes now require a `PathFinder` object, however, this does mean some of the arguments are duplicated, hence we need to look at refactoring this as Router can technically be initialised with just `PathFinder `as `PathFinder` has access to all the properties needed by `Router`. 

**Note:**
Since routing path is done at a section level, when evaluating the when rules to check if a section enables or not, the routing path is not taken into account. This means that if an answer in section 2 enables/disables other sections then if that answer in section 2 gets removed from the path, the sections there were initially enabled by the answer from section 2 will be unchanged. This is due to the limitation of how we calculate the path and is not one of the scenarios needed to be handled for our use case.

It is possible to add a check to schema validator to ensure the answer that drives whether a section is enabled or not must always be on the path, however, it seems unnecessary in this instance, but I am open to suggestions.

### How to review 

> Test using newly introduced schemas.

Ensure:
- You can enable and disable sections based on a given answer.
- The disabled section should not show up on the summary.
- The disabled sections should not be displayed on the hub.
- It should not be possible to access a disabled section manually.


MebinAbraham:
So there is a schema to test using radio answer and one for checkbox answers however, I can remove one of these as we should have confidence that our when rule evaluation works with all answer types. Thoughts?

app/views/contexts/hub_context.py
```diff
@@ -67,6 +67,7 @@ def __init__(
         self._schema = schema
         self._survey_complete = survey_complete
         self._placeholder_renderer = None
+        self._path_finder = path_finder
```
ajmaddaford: I don't think we should add path_finder as a dependency to this, we could just pass the enabled section ids.

scripts/build_schemas.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-mkdir -p data/en
+rm -rf data/en && mkdir -p data/en
```
ajmaddaford: We need to be careful about this as people could have things they're working on in that directory. Maybe we should have a `make clean` command?

tests/app/questionnaire/test_path_finder.py
```diff
@@ -704,3 +704,29 @@ def test_full_routing_path_with_repeating_sections(self):
...
+        path_finder = PathFinder(
+            schema=schema,
+            answer_store=answer_store,
+            metadata={},
+            progress_store=progress_store,
+        )
+
+        expected_section_ids = ['section-1', 'section-2', 'summary-section']
+
+        self.assertEqual(path_finder.enabled_section_ids, expected_section_ids)
```
ajmaddaford: It might also be worth asserting that `section-3` isn't in `enabled_section_ids`.

app/questionnaire/rules.py
```diff
@@ -241,6 +241,19 @@ def evaluate_skip_conditions(
...
+    section_enabled_conditions = section.get('enabled', [])
+    if not section_enabled_conditions:
+        return True
+
+    for condition in section_enabled_conditions:
+        if evaluate_when_rules(
+            condition['when'], schema, metadata, answer_store, list_store
+        ):
+            return True
+    return False
```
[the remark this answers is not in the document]
MebinAbraham: Moved.

app/questionnaire/path_finder.py
```diff
@@ -32,6 +34,26 @@ def __init__(
         self.progress_store = progress_store
         self.list_store = list_store
 
+    @cached_property
```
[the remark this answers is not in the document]
MebinAbraham: Not quite sure, possibly not, have changed to `@property` for now.

scripts/build_schemas.sh
MebinAbraham: They should be working within `data-source/json/` and not `data/en`?

tests/app/questionnaire/test_path_finder.py
MebinAbraham: Can do but the current assert makes asserting `section-3` isn't in `enabled_section_ids` redundant?.

MebinAbraham:
> I think we should consider moving the enabled section ids and full routing path functionality into Router. Looking at `path_finder` I think it should become a section path finder as the only method it has that works outside of a section is full routing path.

@ajmaddaford As discussed, `PathFinder` is now initialised on initialisation of the `Router` object. The `full_routing_path`/`enabled_section_ids` methods are now part of `Router`.

tests/app/questionnaire/test_path_finder.py
ajmaddaford: Oh yeah 🤦‍♂

scripts/build_schemas.sh
ajmaddaford: Most of the time yes, but I've made changes to schemas in data/en before to try things out without rebuilding.

MebinAbraham: Is that a pain point of not wanting to wait for all the census schemas to build?. I do that too. 
For now, I have introduced a new `clean` command to delete the `data` dir and a `build-schemas` command to build just the schemas not everything such as translations.

app/questionnaire/path_finder.py
```diff
@@ -22,9 +22,9 @@ def __init__(
         self,
```
ajmaddaford: I think we should also move `is_path_complete` and `get_first_incomplete_location` into Router. `get_first_incomplete_location` is called from Router anyways and `is_path_complete` is called from the block handlers that have access to a Router.

app/routes/questionnaire.py
```diff
@@ -165,23 +169,27 @@ def post_questionnaire(schema, questionnaire_store):
...
         return redirect(redirect_location.url())
+
     section = schema.get_section(section_id)
 
-    if not section:
+    if not section or section_id not in router.enabled_section_ids:
         return redirect(url_for('.get_questionnaire'))
 
-    routing_path = path_finder.routing_path(
+    routing_path = router.path_finder.routing_path(
```
ajmaddaford: We should consider hiding path_finder from callers of router. We could add a `section_routing_path` method on router that calls `path_finder.routing_path`.

tests/functional/spec/features/enabled-sections/enabled_section_checkbox.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+
+describe('Feature: Section Enabled Based On Checkbox Answers', () => {
+  beforeEach('Open survey', () => {
+    browser.openQuestionnaire('test_section_enabled_checkbox.json');
+  });
+
+  describe('Given the user selects `Section 2`', () => {
+
+    it('When the user submits and proceed to the next page, Then, section 2 should be displayed', () => {
```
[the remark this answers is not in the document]
MebinAbraham: Have decided to drop the describe as I think the first describe is enough.

app/questionnaire/path_finder.py
MebinAbraham: These now live in `Router`.

app/routes/questionnaire.py
MebinAbraham: Have made it a method in router as I think it looks cleaner. So right now, only router and calculated summary used `PathFinder`, but could look at dropping `PathFinder` from calculated summary in favour of `Router` if we think `PathFinder` must be used via router?

app/questionnaire/router.py
```diff
@@ -246,6 +237,24 @@ def full_routing_path(self):
 
         return path
 
+    def _get_first_incomplete_location(self, path):
+        for location in path:
```
MebinAbraham: This was wrapped in an `if path` before my move, but I dropped it because it should never be `None`?. Feel free to correct me if I'm mistaken.

app/questionnaire/router.py
```diff
@@ -202,23 +277,21 @@ def _get_allowable_path(self, routing_path):
...
+                    enabled_section_keys.append(section_key)
             else:
                 section_key = (section_id, None)
-                all_section_keys.append(section_key)
+                enabled_section_keys.append(section_key)
 
         incomplete_section_keys = [
             (section_id, list_item_id)
-            for section_id, list_item_id in all_section_keys
+            for section_id, list_item_id in enabled_section_keys
```
ajmaddaford: Possibly not for this pull request, but this should be a `_get_first_incomplete_section_key` method that returns as soon as it finds an incomplete section - we only care about the next incomplete section or if the survey is complete. Thoughts?

tests/functional/spec/features/enabled-sections/enabled_section_checkbox.spec.js
```diff
@@ -0,0 +1,44 @@
...
+const sectionTwo = require('../../../generated_pages/section_enabled_checkbox/section-2-block.page');
+const summary = require('../../../generated_pages/section_enabled_checkbox/summary.page');
+
+
+describe('Feature: Section Enabled Based On Checkbox Answers', () => {
+  beforeEach('Open survey', () => {
+    browser.openQuestionnaire('test_section_enabled_checkbox.json');
+  });
+
+  it('When the user selects `Section 2` and submits, Then, section 2 should be displayed', () => {
```
ajmaddaford: The comma after 'Then' needs to be removed throughout.

data-source/json/test_section_enabled_hub.json
```diff
@@ -0,0 +1,163 @@
...
+                                    {
+                                        "id": "section-2-answer",
+                                        "label": "",
+                                        "mandatory": false,
+                                        "type": "Number"
+                                    }
+                                ],
+                                "description": "This is section 2.",
+                                "id": "section-2-question",
+                                "title": "Which section is this?.",
```
ajmaddaford: Questions shouldn't end in a full stop. This is done in a couple of places.

app/questionnaire/router.py
```diff
@@ -175,11 +202,59 @@ def is_survey_complete(self):
...
+                )
+
+        return path
+
+    def _get_first_incomplete_location(self, path):
+        for location in path:
+            block = self._schema.get_block(location.block_id)
+            block_type = block.get('type')
+
+            is_location_complete = (
```
bitdivision: Not your code, but:

```
completed_block_ids = self._progress_store.get_completed_block_ids(
    section_id=location.section_id, list_item_id=location.list_item_id
)
is_location_complete = location.block_id in completed_block_ids
```

Reads easier I think :)

MebinAbraham: Should you not be 🛌  😁. But since this is used in multiple places I made a method called `is_location_complete` instead.

app/questionnaire/router.py
```diff
@@ -175,11 +202,59 @@ def is_survey_complete(self):
...
+                location.block_id
+                in self._progress_store.get_completed_block_ids(
+                    section_id=location.section_id, list_item_id=location.list_item_id
+                )
+            )
+
+            if not is_location_complete and block_type not in [
```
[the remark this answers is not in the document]
MebinAbraham: Was just copied by `path_finder.py`, but have changed.

app/questionnaire/router.py
```diff
@@ -175,11 +193,59 @@ def is_survey_complete(self):
...
+
+    def _get_first_incomplete_location(self, path):
+        for location in path:
+            block = self._schema.get_block(location.block_id)
+            block_type = block.get('type')
+
+            if not self._is_location_complete(location) and block_type not in [
+                'Summary',
+                'Confirmation',
+            ]:
```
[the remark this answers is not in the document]
MebinAbraham: Was just copied by `path_finder.py`, but have changed.

app/questionnaire/router.py
MebinAbraham: I think if we leave this then it probably won't be looked at in a long time so I've decided to refactor into `_get_first_incomplete_section_key`.

tests/app/questionnaire/test_path_finder.py
```diff
@@ -254,9 +264,9 @@ def test_routing_path_empty_routing_rules(self):
         answer_1 = Answer(answer_id='mandatory-checkbox-answer', value='Cheese')
         answer_2 = Answer(answer_id='non-mandatory-checkbox-answer', value='deep pan')
 
-        answers = AnswerStore()
-        answers.add_or_update(answer_1)
-        answers.add_or_update(answer_2)
+        answer_Store = AnswerStore()
```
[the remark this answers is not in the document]
MebinAbraham: Fixed.

tests/app/views/contexts/test_hub_context.py
```diff
@@ -1,11 +1,41 @@
...
 from app.data_model.progress_store import ProgressStore, CompletionStatus
-from app.data_model.answer_store import AnswerStore
 from app.questionnaire.questionnaire_schema import QuestionnaireSchema
+from app.questionnaire.router import Router
 from app.views.contexts.hub_context import HubContext
 
 
-def test_get_not_started_row_for_section():
+@pytest.fixture(name='schema')
+def fixture_schema():
```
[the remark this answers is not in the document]
MebinAbraham: This was to prevent disabling the file from `# pylint: disable=redefined-outer-name` and this was the suggested approach and it used in our other tests. However, I have moved the fixtures into the `conftest.py` to avoid this problem as discussed.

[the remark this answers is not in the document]
MebinAbraham: Still doesn't fix it so I used `# pylint: disable=redefined-outer-name`

jonnyshaw89:
:cry: Its like looking back at an ex-girlfriends social media and seeing what shes up to without you :rofl:

PR #7, opened 2019-11-25
Use black quote defaults

iwootten:

In adding formatting for black to eq-schema-validator, I discovered that we aren't using the defaults that black uses for quoting strings in pylint. The reason it isn't flagged is that we explicitly tell black not to reformat pylints quoted strings within pyproject.toml. The black project has said they don't intend to ever give the option to [format using singlequotes](https://github.com/psf/black/issues/118).

### How to review 

Check you agree with adopting blacks defaults.


iwootten:
Holding off on this until we have a few less open PR's which it'll cause issues on.

PR #8, opened 2019-11-27
Fix minimum/maximum date validation

berroar:
There is a bug that is stopping runner from evaluating minimum and maximum values correctly. 

This bug may have been introduced when we implemented repeating sections, as the code that evaluates the answer value for the minimum date is missing the list_item_id as an argument. The validation still works on Census Individual surveys.

We need to fix date_form.py so that the list_item_id is passed to the method that retrieves the answer value.

### How to review 
Is this a viable solution/temporary solution? Does runner need to be refactored?


berroar:
[Fix minimum and maximum date validation](https://trello.com/c/kIOcPcRv/3539-fix-minimum-and-maximum-date-validation)

app/forms/date_form.py
```diff
@@ -250,7 +251,10 @@ def get_referenced_offset_value(answer_min_or_max, answer_store, metadata):
     elif 'answer_id' in answer_min_or_max:
         schema = load_schema_from_metadata(metadata)
         answer_id = answer_min_or_max['answer_id']
-        value = get_answer_value(answer_id, answer_store, schema)
+        if location:
+            value = get_answer_value(answer_id, answer_store, schema, location.list_item_id)
+        else:
+            value = get_answer_value(answer_id, answer_store, schema, None)
```
iwootten: I don't think you need to pass the None here as it has a default in the method.

app/forms/fields.py
```diff
@@ -28,7 +28,7 @@
 
 
 def get_field(
-    answer, label, error_messages, answer_store, metadata, disable_validation=False
+    answer, label, error_messages, answer_store, metadata, location, disable_validation=False
```
iwootten: If this is defaulted as a kwarg to None, it would save adding it to each of the calling methods.

app/forms/date_form.py
```diff
@@ -226,7 +227,7 @@ def get_dates_for_single_date_period_validation(answer, answer_store, metadata):
     return minimum_referenced_date, maximum_referenced_date
 
 
-def get_referenced_offset_value(answer_min_or_max, answer_store, metadata):
+def get_referenced_offset_value(answer_min_or_max, answer_store, metadata, location):
```
iwootten: Same comment here about defaulting to None

app/forms/questionnaire_form.py
```diff
@@ -16,12 +16,14 @@
...
-    def __init__(self, schema, question_schema, answer_store, metadata, location, **kwargs):
+    def __init__(
+        self, schema, question_schema, answer_store, metadata, location, **kwargs
+    ):
         self.schema = schema
         self.question = question_schema
         self.answer_store = answer_store
         self.metadata = metadata
-        self.location = location,
+        self.location = (location,)
```
iwootten: Is there a reason why this is a tuple?

berroar: Not sure how that got there, but have removed it now.

PR #9, opened 2019-11-27
Evaluate section status on list change

JamesGardiner:
Section status needs to be re-evaluated on a list change when a section is linked to a list via when rules.

This PR introduces the mechanism for updating the section status on a post to a ListAction block type.
### How to review 
Run the full test suite.
Review code.


JamesGardiner:
[Accommodation completeness not updated when the householder list changes](https://trello.com/c/aeED9Esp/3260-accommodation-completeness-not-updated-when-the-householder-list-changes)

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -200,6 +200,13 @@ def update_section_status(
             section_status, section_id, list_item_id
         )
 
+    def sections_to_evaluate(self, filter_by: Set[str] = None) -> Union[List, None]:
+        completed_sections = self._progress_store.completed_section_keys()
+        try:
+            return list(completed_sections & filter_by)
+        except TypeError:
+            return list(completed_sections)
```
MebinAbraham: I personally prefer `if` when dealing with `NoneType` as that is which is common in this project.
```py
return (
    list(completed_sections)
    if filter_by is None
    else list(completed_sections & filter_by)
)
```

app/views/handlers/list_action.py
```diff
@@ -47,8 +49,33 @@ def handle_post(self):
...
+            self._update_section_completeness(
+                Location(section_path[0][0]), section_path[1]
+            )
+
+    def _sections_to_evaluate(self):
+        list_name = self._schema.get_block(self.rendered_block['parent_id'])['for_list']
+        associated_sections = self._schema.get_sections_associated_to_list_name(
+            list_name
+        )
+        associated_sections = {(element, None) for element in associated_sections}
```
MebinAbraham: This would mean it won't work with repeats?

So I had a block within a repeat that is only displayed when there is at least 2 members in the household.
The block wasn't displayed when there was 1 householder. Once the section for that individual was complete, I added another person, which should make the individual section be marked as in progress however it was still marked as complete.
Clicking View answers then results in a circular redirect.

app/questionnaire/questionnaire_store_updater.py
JamesGardiner: Happy to use the ternary in this instance

app/views/handlers/block.py
```diff
@@ -131,10 +131,13 @@ def _get_routing_path(self):
             list_item_id=self._current_location.list_item_id,
         )
 
-    def _update_section_completeness(self, location: Optional[Location] = None):
+    def _update_section_completeness(
+        self, location: Optional[Location] = None, routing_path: Optional = None
+    ):
```
ajmaddaford: This method shouldn't be modified to take any routing path. Everything in this class works at a block level and assumes it's the current location (unless it's a parent location). 

We should consider moving the `section_status` calculation to `questionnaire_store_updater.update_section_status`, so this method would look more like:

```python
    def _update_section_completeness(
        self, location: Optional[Location] = None
    ):
        location = location or self._current_location

        self.questionnaire_store_updater.update_section_status(
            is_complete=self.path_finder.is_path_complete(self._routing_path),
        )
```

app/views/handlers/list_action.py
```diff
@@ -47,8 +49,33 @@ def handle_post(self):
...
+        self.evaluate_sections()
         self.questionnaire_store_updater.save()
 
+    def evaluate_sections(self):
+        sections_to_evaluate = self._sections_to_evaluate()
+        section_paths = self._routing_paths(sections_to_evaluate)
+        for section_path in section_paths:
+            self._update_section_completeness(
+                Location(section_path[0][0]), section_path[1]
+            )
```
ajmaddaford: I don't see the need for the _routing_paths method. This could just be:

```python
for section in self._sections_to_evaluate():
    path = self.path_finder.routing_path(section[0])
	self._update_section_completeness(
        Location(section_path[0][0]), section_path[1]
    )
```

Also, if the suggested changes are made to the `update_section_status` method in questionnaire store updater, this could become:

```python
for section_id, list_item_id in self._sections_to_evaluate():
    path = self.path_finder.routing_path(section_id)
	self.questionnaire_store_updater.update_section_status(
        is_complete = self.path_finder.is_path_complete(path),
        section_id=section_id,
        list_item_id=list_item_id
    )
```

app/views/handlers/list_action.py
```diff
@@ -47,8 +49,33 @@ def handle_post(self):
             self.parent_location.block_id
         )
         self.questionnaire_store_updater.remove_answers(answer_ids_to_remove)
+        self.evaluate_sections()
         self.questionnaire_store_updater.save()
 
+    def evaluate_sections(self):
```
ajmaddaford: This is not just evaluating, it is updating so the method name should reflect that.

app/questionnaire/questionnaire_schema.py
```diff
@@ -38,6 +39,38 @@ def get_section_ids(self):
...
+                    yield v
+                if isinstance(v, dict):
+                    for result in self._lookup_nested_when_rules(v):
+                        yield result
+                elif isinstance(v, list):
+                    for d in v:
+                        for result in self._lookup_nested_when_rules(d):
+                            yield result
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
JamesGardiner: If a list item is a string not a dict

app/views/handlers/list_action.py
JamesGardiner: Fixed in the next push

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -200,6 +200,13 @@ def update_section_status(
             section_status, section_id, list_item_id
         )
 
+    def sections_to_evaluate(self, filter_by: Set[str] = None) -> Union[List, None]:
```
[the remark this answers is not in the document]
JamesGardiner: Moved this to the progress store but needs further work as I'm now accessing progress store via questionnaire store updater, when progress store is a private attribute

app/views/handlers/list_action.py
JamesGardiner: Initial suggestion is implemented - working on the changes to progress_store/questionnaire_store_updater

app/data_model/progress_store.py
```diff
@@ -64,14 +64,30 @@ def is_dirty(self) -> bool:
...
             for section_key, section_progress in self._progress.items()
-            if section_progress.status == CompletionStatus.COMPLETED
-        ]
+            if section_progress.status in statuses
+        }
+
+    def get_in_progress_and_completed_sections(
+        self, filter_by: Set[str] = None
+    ) -> Union[List, None]:
+        completed_sections = self.section_keys_by_status(
```
MebinAbraham: This isn't just completed sections.

app/data_model/progress_store.py
```diff
@@ -64,14 +64,30 @@ def is_dirty(self) -> bool:
...
+        return {
             section_key
             for section_key, section_progress in self._progress.items()
-            if section_progress.status == CompletionStatus.COMPLETED
-        ]
+            if section_progress.status in statuses
+        }
+
+    def get_in_progress_and_completed_sections(
+        self, filter_by: Set[str] = None
```
MebinAbraham: I can't see where this method is called with `filter_by` apart from a test?.

app/views/handlers/list_action.py
```diff
@@ -47,8 +47,22 @@ def handle_post(self):
             self.parent_location.block_id
         )
         self.questionnaire_store_updater.remove_answers(answer_ids_to_remove)
+        self.evaluate_and_update_section_status()
         self.questionnaire_store_updater.save()
 
+    def evaluate_and_update_section_status(self):
+        sections_to_evaluate = (
+            self.questionnaire_store_updater.get_in_progress_and_completed_sections()
```
MebinAbraham: This currently seems to be returning all sections that are in progress and complete, and not just the ones that have a dependency on the list we are dealing with.

app/data_model/progress_store.py
JamesGardiner: Yeah good shout - this was evaluating every completed or in progress section. I've updated in [da77aab5](https://github.com/ONSdigital/eq-questionnaire-runner/pull/9/commits/da7aab520465aca3e3d28bcf5ccfd6ccb20f4a7d)

data-source/json/test_list_change_evaluates_sections.json
```diff
@@ -0,0 +1,653 @@
+{
```
ajmaddaford: This schema is far larger than it needs to be, it should be the minimum required to test the feature.

app/data_model/progress_store.py
```diff
@@ -64,13 +64,30 @@ def is_dirty(self) -> bool:
...
-        return [
+    def section_keys_by_status(self, statuses: Iterable[str]) -> Set[str]:
+        return {
             section_key
             for section_key, section_progress in self._progress.items()
-            if section_progress.status == CompletionStatus.COMPLETED
+            if section_progress.status in statuses
+        }
+
+    def get_in_progress_and_completed_section_ids(
```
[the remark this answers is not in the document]
JamesGardiner: Yeah I initially named it with keys and don't remember what prompted me to change it. Keys makes more sense. I'll take a look at adding the `filter_by_section_ids` as well; see what fits best :+1:

tests/app/data_model/test_progress_store.py
```diff
@@ -471,3 +473,56 @@ def test_remove_progress_for_list_item_id():
...
+        ({'s1', 's2'}, [('s1', None), ('s2', None)]),
+        # Repeating sections and non repeating sections
+        ({'s1', 's4'}, [('s1', None), ('s4', '123abc')]),
+        # Only repeating sections
+        ({'s4', 's5'}, [('s4', '123abc'), ('s5', '456def')]),
+        # No filter_by paramater specified
+        (None, [('s1', None), ('s2', None), ('s4', '123abc'), ('s5', '456def')]),
+    ],
+)
+def test_in_progress_and_completed_section_ids(test_input, expected):
```
iwootten: Could these args be named more explicitly?

app/views/handlers/list_action.py
```diff
@@ -47,8 +47,27 @@ def handle_post(self):
...
+        list_name = self.parent_block['for_list']
+        section_ids_to_filter_by = self._schema.get_section_ids_dependent_on_list(
+            list_name
+        )
+
+        section_ids_to_evaluate = self.questionnaire_store_updater.get_in_progress_and_completed_section_ids(
+            section_ids_to_filter_by
+        )
+
+        for section_id, list_item_id in section_ids_to_evaluate:
```
ajmaddaford: `section_ids_to_evaluate` contains section keys, so should be named `section_keys_to_evaluate`

app/views/handlers/list_action.py
```diff
@@ -47,8 +47,25 @@ def handle_post(self):
             self.parent_location.block_id
         )
         self.questionnaire_store_updater.remove_answers(answer_ids_to_remove)
+        self.evaluate_and_update_section_status()
```
MebinAbraham: This will only update dependencies on a list action of a non-primary person list collector. Even though it is unlikely, we need to update the dependencies on an action of a primary person list action as well, as it is possible to someone to change their mind and add/remove a primary person later on.

app/questionnaire/questionnaire_schema.py
```diff
@@ -38,6 +39,25 @@ def get_section_ids(self):
...
+    def _section_ids_associated_to_list_name(self, list_name: str) -> Set:
+        section_ids = set()
+        for block in self.get_blocks():
+            for when_rule in _get_values_for_key(
+                block, 'when', ignore_keys={'question_variant', 'content_variant'}
+            ):
+                for rule in when_rule:
+                    if rule.get('list') == list_name:
+                        section_ids.add(self.get_section_id_for_block_id(block['id']))
+        return section_ids
```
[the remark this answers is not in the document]
JamesGardiner:
```python
    def _section_ids_associated_to_list_name(self, list_name: str) -> Set:
        section_ids = set()

        for section in self.get_sections():
            if section['id'] in section_ids:
                continue

            ignore_keys = {'question_variants', 'content_variants'}
            when_rules = [
                when_rule
                for when_rule in _get_values_for_key(section, 'when', ignore_keys)
            ]
            if any(
                rule.get('list') == list_name
                for when_rule in when_rules
                for rule in when_rule
            ):
                section_ids.add(section['id'])
                continue

        return section_ids
```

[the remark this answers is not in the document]
MebinAbraham: I would still use a list over set as `section_ids` are used for iteration in `section_keys` method.

app/views/handlers/primary_person_list_collector.py
```diff
@@ -26,18 +26,20 @@ def get_context(self):
         return build_question_context(self.rendered_block, self.form)
 
     def handle_post(self):
+        list_name = self.rendered_block['for_list']
+        self.evaluate_and_update_section_status_on_list_change(list_name)
```
MebinAbraham: This needs to done after the action such as once the person is added or removed.

app/views/handlers/primary_person_list_collector.py
```diff
@@ -26,18 +26,23 @@ def get_context(self):
...
```
ajmaddaford: The original line above should now be removed.

app/views/handlers/primary_person_list_collector.py
```diff
@@ -26,18 +26,23 @@ def get_context(self):
...
+            self.evaluate_and_update_section_status_on_list_change(list_name)
+
             self.questionnaire_store_updater.save()
+
         else:
-            self.questionnaire_store_updater.remove_primary_person(
-                self.rendered_block['for_list']
-            )
+            self.questionnaire_store_updater.remove_primary_person(list_name)
+            self.evaluate_and_update_section_status_on_list_change(list_name)
```
ajmaddaford: This is done before `update_answers` is called by `super().handle_post()` below.

app/views/handlers/primary_person_list_collector.py
```diff
@@ -26,18 +26,21 @@ def get_context(self):
...
         else:
-            self.questionnaire_store_updater.remove_primary_person(
-                self.rendered_block['for_list']
-            )
-            return super().handle_post()
+            self.questionnaire_store_updater.remove_primary_person(list_name)
+            super().handle_post()
+
+        self.evaluate_and_update_section_status_on_list_change(list_name)
+        self.questionnaire_store_updater.save()
```
MebinAbraham: This does mean when they are not adding we are potentially saving the state twice. Is it better to accept some duplication and avoid this?

JamesGardiner: I think at this point it's best to check that is_dirty is True.

app/views/handlers/primary_person_list_collector.py
MebinAbraham: This is still the case @JamesGardiner ?

JamesGardiner: Yes still the case - which means that a call to the method may incorrectly evaluate the current section. The current section status is evaluated correctly in the call to `super.handle_post()`. Spoke with @ajmaddaford about this last week and decided it's the least unfavourable option without a significant of refactoring to move `evaluate_and_update_section_status_on_list_change` to questionnaire_store_updater. The limitation on that being that a router would also need to be passed at the moment, which we want to avoid.

data-source/json/test_list_change_evaluates_sections_primary_person.json
```diff
@@ -0,0 +1,340 @@
+{
```
ajmaddaford: I think we should just have one test schema for this with a primary person as that schema can be used for all the test scernarios.

data-source/json/test_list_change_evaluates_sections.json
```diff
@@ -3,9 +3,9 @@
     "schema_version": "0.0.1",
     "data_version": "0.0.3",
     "survey_id": "0",
-    "title": "Test ListCollector Evaluate Sections",
+    "title": "Test ListCollector Evaluate Sections for Primary Person",
```
ajmaddaford: I think the "for Primary Person" should be dropped as it's not just for primary person.

data-source/json/test_list_change_evaluates_sections.json
```diff
@@ -3,9 +3,9 @@
...
     "theme": "default",
-    "description": "A questionnaire to test that changes to a list update section completeness",
+    "description": "A questionnaire to test that changes to a primary person list update section completeness",
```
ajmaddaford: Same here, the previous text was more appropriate.

PR #12, opened 2019-12-05
Add FieldHandler classes

iwootten:

Modified the way fields are created, in order to support using location during validation. This changes handling of fields to be created using a class so that it is possible to have access to common globals in each child field class. 

Fields are now managed through a Handler style interface, similar to how views are invoked. This means each fields logic now lives in a handler class which is decided through a map. Each handler defines a get_field method which returns the wtform field to be used. The StringHandler class is the simplest and the FieldHandler is the base that others inherit from. The most complex is the DateHandler which now includes all the logic for date validators within it. Additionally, the date field has been reworked to calculate "dependent" values prior to the field creation. There is further work that could be done in order to define dependencies in a uniform manner throughout runner in order to move the dependency logic completely outside of handlers.

### How to review 

Confirm that the issue described in https://github.com/ONSdigital/eq-questionnaire-runner/pull/8 is now resolved. Check the tests pass. Confirm you agree with the design choices made.


iwootten:
[Fix minimum and maximum date validation](https://trello.com/c/kIOcPcRv/3539-fix-minimum-and-maximum-date-validation)

app/forms/field_handlers/field_handler.py
```diff
@@ -0,0 +1,71 @@
...
+    def get_validation_message(self, message_key):
+        message = self.error_messages[message_key]
+
+        if (
+            'validation' in self.answer_schema
+            and 'messages' in self.answer_schema['validation']
+            and message_key in self.answer_schema['validation']['messages']
+        ):
+            message = self.answer_schema['validation']['messages'][message_key]
+        return message
```
MebinAbraham: Thoughts?
```py
    def get_validation_message(self, message_key):
        message = (
            self.answer_schema.get('validation', {})
            .get('messages', {})
            .get(message_key)
            or self.error_messages[message_key]
        )
        return message
```

app/forms/field_handlers/number_handler.py
```diff
@@ -0,0 +1,118 @@
...
+            else CustomIntegerField
+        )
+        return field_type(
+            label=self.label, validators=self.validators, description=self.guidance
+        )
+
+    def get_number_field_dependencies(self):
+        max_decimals = self.answer_schema.get('decimal_places', 0)
+
+        min_value = 0
```
MebinAbraham: Shouldn't this be `self.MIN_NUMBER`?

app/forms/field_handlers/number_handler.py
```diff
@@ -0,0 +1,118 @@
...
+from app.data_model.answer_store import AnswerStore
+from app.forms.fields.custom_decimal_field import CustomDecimalField
+from app.forms.fields.custom_integer_field import CustomIntegerField
+from app.forms.field_handlers.field_handler import FieldHandler
+from app.questionnaire.location import Location
+from app.validation.validators import NumberCheck, NumberRange, DecimalPlaces
+
+
+class NumberHandler(FieldHandler):
+    MANDATORY_MESSAGE = 'MANDATORY_TEXTFIELD'
```
MebinAbraham: Shouldn't this be `MANDATORY_NUMBER`?

app/forms/field_handlers/number_handler.py
```diff
@@ -0,0 +1,118 @@
...
+    MAX_NUMBER = 9999999999
+    MIN_NUMBER = -999999999
+    MAX_DECIMAL_PLACES = 6
```
MebinAbraham: This doesn't seem to be used anywhere but a test?.

app/forms/fields/custom_select_field.py
```diff
@@ -0,0 +1,35 @@
+from wtforms import SelectField
+
+
+class CustomSelectField(SelectField):
```
ajmaddaford: We should name custom fields after their functionality rather than just prefixing with `Custom`. This could be  `SelectFieldWithDetailAnswer` and the `MultipleSelectField` could be `MultipleSelectFieldWithDetailAnswer`. I can't think of sensible names for the custom Number and Decimal fields though.

app/forms/field_handlers/number_handler.py
iwootten: Should be used in get_number_field_dependencies

app/forms/field_handlers/date_handler.py
```diff
@@ -0,0 +1,160 @@
...
+        limit = None
+
+        if limit_key in self.answer_schema:
+            limit = self.get_referenced_date(limit_key)
+
+            if 'offset_by' in self.answer_schema[limit_key]:
+                offset = self.answer_schema[limit_key]['offset_by']
+                limit = self.transform_date_by_offset(limit, offset)
+
+        return limit
```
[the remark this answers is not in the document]
iwootten: I've updated the method and argument names here, but can't see passing the referenced_date makes this any simpler.

app/forms/field_handlers/field_handler.py
```diff
@@ -0,0 +1,83 @@
...
+            mandatory_message = self.get_validation_message(self.MANDATORY_MESSAGE_KEY)
+
+            return ResponseRequired(message=mandatory_message)
+
+        return validators.Optional()
+
+    def get_schema_value(self, schema_element):
+        if 'meta' in schema_element:
+            return get_metadata_value(self.metadata, schema_element['meta'])
+        if 'value' in schema_element:
```
MebinAbraham: `elif`?

app/forms/field_handlers/field_handler.py
```diff
@@ -0,0 +1,83 @@
...
+            return schema_element['value']
+        if 'answer_id' in schema_element:
```
MebinAbraham: Same here `elif`

app/forms/field_handlers/number_handler.py
```diff
@@ -0,0 +1,102 @@
...
+        return {
+            'min_exclusive': self.answer_schema.get('min_value', {}).get(
+                'exclusive', False
+            ),
+            'max_exclusive': self.answer_schema.get('max_value', {}).get(
+                'exclusive', False
+            ),
+            'min_value': min_value,
+            'max_value': max_value,
+        }
```
MebinAbraham: Thoughts?. Just to reduce some duplication.
```py
    def get_field_references(self):
        schema_min_value = self.answer_schema.get('min_value', {})
        min_value = self.get_schema_value(schema_min_value) if schema_min_value else 0

        schema_max_value = self.answer_schema.get('max_value', {})
        max_value = (
            self.get_schema_value(schema_max_value)
            if schema_max_value
            else self.MAX_NUMBER
        )

        return {
            'min_exclusive': schema_min_value.get('exclusive', False),
            'max_exclusive': schema_max_value.get('exclusive', False),
            'min_value': min_value,
            'max_value': max_value,
        }
```

app/forms/field_handlers/number_handler.py
```diff
@@ -0,0 +1,102 @@
...
+        answer_errors = self.error_messages.copy()
+
+        if (
+            'validation' in self.answer_schema
+            and 'messages' in self.answer_schema['validation']
+        ):
+            for error_key, error_message in self.answer_schema['validation'][
+                'messages'
+            ].items():
+                answer_errors[error_key] = error_message
```
MebinAbraham: Just to make it a bit easier to follow?
```py
        schema_validation_messages = self.answer_schema.get('validation', {}).get(
            'messages', {}
        )

        for error_key, error_message in schema_validation_messages.items():
            answer_errors[error_key] = error_message
```

app/forms/field_handlers/field_handler.py
iwootten: It isn't necessary on these as they all return early. Pylint will fail.

app/forms/questionnaire_form.py
```diff
@@ -305,49 +311,42 @@ def get_data(self, answer_id):
         return attr.raw_data[0] if attr.raw_data else ''
 
 
+def _option_value_in_data(answer, option, data):
+    data_to_inspect = data.to_dict(flat=False) if isinstance(data, MultiDict) else data
+
+    return option['value'] in dict(data_to_inspect).get(answer['id'], [])
+
+
 # pylint: disable=too-many-locals
```
MebinAbraham: This can now be removed.

tests/app/forms/handlers/test_date_handler.py
```diff
@@ -0,0 +1,300 @@
+# pylint: disable=unused-argument
```
ajmaddaford: Some of these tests should be moved into test files for the different date handlers. Other tests should be moved into test files for the date fields.

tests/app/forms/test_fields.py
```diff
@@ -1,22 +1,20 @@
-from wtforms import validators, StringField, FormField, SelectField
+from wtforms import StringField, FormField, SelectField
```
ajmaddaford: This file seems to be testing field handlers more than fields, so should be renamed.

tests/app/forms/validation/test_number_range_validator.py
```diff
@@ -95,6 +97,31 @@ def test_within_range_at_max(self):
...
+        }
+        returned_error_messages = answer['validation']['messages']
+
+        with self.assertRaises(Exception) as ite:
+            get_field(answer, returned_error_messages, self.store, {}, None, False)
+
+            self.assertEqual(
+                str(ite.exception),
+                'answer: set-maximum-cat value: cat for answer id: test-range is not a valid number',
+            )
```
ajmaddaford: I think all of the tests need to be re-organised to match the code changes. For example, this test is in the number range validator, but the value is resolved in the field handler.

tests/app/forms/handlers/test_duration_field.py
```diff
@@ -0,0 +1,29 @@
...
+        'mandatory': True,
+        'options': [],
+        'q_code': '11',
+        'type': 'Duration',
+        'units': ['years', 'months'],
+        'validation': {
+            'messages': {
+                'INVALID_DURATION': 'The duration entered is not valid.  Please correct your answer.',
+                'MANDATORY_DURATION': 'Please provide a duration to continue.',
+            }
```
ajmaddaford: Can `validation` be dropped if it's not used in the test?

tests/app/forms/validation/test_number_range_validator.py
```diff
@@ -2,10 +2,12 @@
 from unittest.mock import Mock, patch
 from wtforms.validators import ValidationError
 
+from app.forms.field_handlers.number_handler import NumberHandler
+from app.forms.fields.integer_field_with_separator import IntegerFieldWithSeparator
+from app.forms.field_factory import get_field_handler
```
ajmaddaford: Testing the validators shouldn't require the field or handler, and we shouldn't have any schema in these tests. If we're testing that a validator is attached to a field, then that test should live in the handler tests.

tests/app/forms/handlers/test_duration_field.py
iwootten: Thinking about this...doesn't it imply that a duration field can't be marked as mandatory?

ajmaddaford: The validators for the duration field are created within the form rather than passed to it.

iwootten: To answer my own question, no - because duration_form uses a custom .validate() that means validation is done at the form and not the field level.

tests/app/forms/handlers/test_number_handler.py
```diff
@@ -105,8 +115,200 @@ def test_percentage_field():
...
+    }
+
+    test_form_class = get_test_form_class(answer_schema)
+    form = test_form_class(DummyPostData(test_field=['9']))
+    form.validate()
+
+    assert (
+        form.errors['test_field'][0]
+        == answer_schema['validation']['messages']['NUMBER_TOO_SMALL']
+    )
```
ajmaddaford: This test looks like it's testing the min rather than the range. Possibly worth adding a test for max as well and rename to `test_min_and_max`?

tests/app/forms/handlers/test_number_handler.py
```diff
@@ -105,8 +115,200 @@ def test_percentage_field():
...
+
+
+def test_manual_min_exclusive(app):
+    answer_schema = {
+        'min_value': {'value': 10, 'exclusive': True},
+        'label': 'Min Test',
+        'mandatory': False,
+        'validation': {
+            'messages': {
+                'INVALID_NUMBER': 'Please only enter whole numbers into the field.',
```
ajmaddaford: I don't think `INVALID_NUMBER` is used in the test and should be removed. This applies to other tests with validation messages that aren't used.

tests/app/forms/handlers/test_number_handler.py
```diff
@@ -105,8 +115,200 @@ def test_percentage_field():
...
+        'decimal_places': 2,
+        'label': 'Range Test 10 to 20',
+        'mandatory': False,
+        'validation': {
+            'messages': {
+                'INVALID_NUMBER': 'Please only enter whole numbers into the field.',
+                'NUMBER_TOO_SMALL': 'The minimum value allowed is 10. Please correct your answer.',
+                'NUMBER_TOO_LARGE': 'The maximum value allowed is 20. Please correct your answer.',
+            }
+        },
```
ajmaddaford: No need for `validation` messages here?

tests/app/forms/field_handlers/test_number_handler.py
```diff
@@ -0,0 +1,333 @@
...
+        'id': 'chewies-age-answer',
+        'mandatory': False,
+        'label': 'How old is Chewy?',
+        'type': 'Number',
+        'validation': {
+            'messages': {
+                'NUMBER_TOO_LARGE': 'No one lives that long, not even Yoda',
+                'NUMBER_TOO_SMALL': 'Negative age you can not be.',
+                'INVALID_NUMBER': 'Please enter your age.',
+            }
```
ajmaddaford: Are `validation` messages needed in tests that don't use them?

app/forms/field_handlers/text_area_handler.py
```diff
@@ -0,0 +1,36 @@
...
+        validate_with = super().validators
+        if self.disable_validation is False:
+            validate_with.append(self.get_length_validator())
+        return validate_with
+
+    @cached_property
+    def max_length(self):
+        if 'max_length' in self.answer_schema and self.answer_schema['max_length'] > 0:
+            return self.answer_schema['max_length']
+        return self.MAX_LENGTH
```
MebinAbraham: While we're here?
```py
    def max_length(self):
        max_length = self.answer_schema.get('max_length', 0)
        return max_length if max_length > 0 else self.MAX_LENGTH
```

data-source/jsonnet/england-wales/individual/blocks/identity-and-health/arrive_in_country.jsonnet
```diff
@@ -14,6 +14,9 @@ local question(title) = {
       minimum: {
         answer_id: 'date-of-birth-answer',
       },
+      maximum: {
+        value: 'now',
+      },
```
MebinAbraham: Is this needed?

iwootten: Yes...?

PR #13, opened 2019-12-05
Ability to cancel from add a person screen

berroar:
This PR has been raised to add the ability for the user to cancel and return to the previous page when using an add person screen/list item add screen. See [trello for more info](https://trello.com/c/VOlzRmlO/3533-ability-to-cancel-from-add-a-person-screen-s). This functionality has not been added to the primary person list collector.

I have added the cancel_text variable which allows for a bespoke 'cancel and return' message to be used in the census schemas. The default text is ''Cancel and return to the previous question".

The functionality is the same as the 'previous' link.

### How to review 
Test that the new link is functional and is being displayed in the test_list_collector schema.

Check the new messages are being displayed for the census and css schemas.


berroar:
[Ability to cancel from add a person screen (S)](https://trello.com/c/VOlzRmlO/3533-ability-to-cancel-from-add-a-person-screen-s)

Pipfile
```diff
@@ -61,7 +61,7 @@ redis = "==3.2.1"
 flask-compress = "*"
 htmlmin = "*"
 coloredlogs = "*"
-eq-translations = {editable = true,git = "https://github.com/ONSDigital/eq-translations.git",ref = "master"}
+eq-translations = {editable = true,git = "https://github.com/ONSDigital/eq-translations.git",ref = "add-cancel-text-key"}
```
berroar: Revert before merge.

scripts/test_schemas.sh
```diff
@@ -6,7 +6,7 @@ if [ "$1" == "--local" ] || [ "$2" == "--local" ]; then
 fi
 
 if [ "$run_docker" == true ]; then
-    branch=v3
+    branch=add-cancel-text-variable
```
berroar: Revert before merge

data-source/jsonnet/northern-ireland/household/blocks/who-lives-here/visitor_list_collector.jsonnet
```diff
@@ -100,6 +100,10 @@ function(census_date) {
   add_block: {
     id: 'add-visitor',
     type: 'ListAddQuestion',
+    cancel_text: {
+      text: 'Don’t need to add anyone else?',
+      action_text: 'Return to the previous page',
+    },
```
ajmaddaford: I don't think there is a requirement for the action text to be configurable. We should just have `cancel_text` and if that's not in the schema no link should be displayed. The current fallback to `Cancel and return to the previous question` doesn't work as the previous page could be a summary.

berroar: I split it out to two separate strings because of the requirement for census that the hyperlink would only be displayed on the 'action' part of text. I see what you mean about the fall back.

berroar: This has been fixed.

templates/listaddquestion.html
```diff
@@ -1 +1,13 @@
...
+  {% if previous_location_url %}
+    {% if content.block.cancel_text %}
+        <p class="u-mt-xl" >
+           {{ content.block.cancel_text }} <a id="cancel-and-return" href="{{ previous_location_url }}">{{ _("Return to the previous page") }}</a></p>
+    {% else %}
+        <p class="u-mt-xl" >
+        <a id="cancel-and-return" href="{{ previous_location_url }}">{{ _("Cancel and return to the previous page") }}</a></p>
+    {% endif %}
+  {% endif %}
+{% endblock cancel_and_return %}
```
MebinAbraham: Petty formatting comment incoming:
```jinja2
{% extends 'question.html' %}

{% block cancel_and_return %}
  {% if previous_location_url %}
    {% if content.block.cancel_text %}
      <p class="u-mt-xl">
      {{ content.block.cancel_text }} <a id="cancel-and-return" href="{{ previous_location_url }}">{{ _("Return to the previous page") }}</a>
      </p>
    {% else %}
      <p class="u-mt-xl">
        <a id="cancel-and-return" href="{{ previous_location_url }}">{{ _("Cancel and return to the previous page") }}</a>
      </p>
    {% endif %}
  {% endif %}
{% endblock cancel_and_return %}
```

berroar: Updated this.

berroar:
I have simplified the listaddquestion.html page as we are now using the default cancel message: "Cancel and return to the previous page" as part of the census text.

MebinAbraham:
As per @ajmaddaford comment on validator, if this is to be supported for more than just the add screen, then we need to make the relevant changes here + new tests.

berroar:
> As per @ajmaddaford's comment on validator, if this is to be supported for more than just the add screen, then we need to make the relevant changes here + new tests.

I've now made the cancel link available for list edit and remove questions.

tests/functional/spec/list_collector.spec.js
```diff
@@ -96,6 +96,37 @@ describe('List Collector', function() {
...
+    it('The user is returned to the list collector when the cancel link is clicked on the remove page.', function() {
+      $(ListCollectorPage.yes()).click();
+      $(ListCollectorPage.submit()).click();
+      $(ListCollectorAddPage.firstName()).setValue('Someone');
+      $(ListCollectorAddPage.lastName()).setValue('Else');
+      $(ListCollectorAddPage.submit()).click();
+      $(ListCollectorPage.listRemoveLink(1)).click();
+      $(ListCollectorRemovePage.cancelAndReturn()).click();
+      expect(browser.getUrl()).to.contain(ListCollectorPage.pageName);
+    });
```
MebinAbraham: These check the functionality of the link. We need to add the `action_text` into a schema that is not the census and make sure that is also displayed on these pages. Probably easiest to add it to the list_collector json and edit the existing integration test.

berroar: I've now added a test for this, not sure if I need to add additional ones for the edit and remove pages too?

MebinAbraham: I would say yes because different templates handle these.

MebinAbraham:
Padding doesn't seem right.
Screenshot 2019-12-16 at 12 48 30

templates/layouts/_questionnaire.html
```diff
@@ -21,6 +21,8 @@
       </div>
   {% endblock %}
 
+  {% block cancel_and_return %}{% endblock %}
```
MebinAbraham: This is defined in `listaddquestion.html`, `listeditquestion.html` and `listremovequestion.html` but should only need to define this once and then can be used in `questionnaire.html`.

berroar: I have abstracted the code for these blocks.

berroar:
> Padding doesn't seem right, seems excessive.
> Screenshot 2019-12-16 at 12 48 30

The link is now below the submit button, but I have reduced the padding slightly.

data-source/json/test_list_collector.json
```diff
@@ -89,6 +90,7 @@
                             "edit_block": {
                                 "id": "edit-person",
                                 "type": "ListEditQuestion",
+                                "cancel_text": "Don't need to add anyone else?",
```
ajmaddaford: This should be something like "Don't need to change anything?"

data-source/json/test_list_collector.json
```diff
@@ -112,6 +114,7 @@
                             "remove_block": {
                                 "id": "remove-person",
                                 "type": "ListRemoveQuestion",
+                                "cancel_text": "Don't need to add anyone else?",
```
ajmaddaford: This should be something like "Don't need to remove this person?"

templates/partials/cancel-and-return.html
```diff
@@ -0,0 +1,10 @@
+{% block cancel_and_return %}
+  {% if previous_location_url %}
+    <p class="u-mt-l">
+      {% if content.block.cancel_text %}
+        {{ content.block.cancel_text }}
+      {% endif %}
+      <a id="cancel-and-return" href="{{ previous_location_url }}">{{ _("Return to the previous page") }}</a>
+    </p>
+  {% endif %}
+{% endblock cancel_and_return %}
```
ajmaddaford: We should rename this file `list-action.tmpl`, extend from `question.html` and amend the add/edit/remove templates to inherit from `list-action.tmpl`. That way it's clear that add/edit/remove all use the same functionality.

templates/list-action.tmpl
```diff
@@ -0,0 +1,12 @@
+{% extends 'question.html' %}
+
+{% block cancel_and_return %}
+  {% if previous_location_url %}
+    <p class="u-mt-l">
+      {% if content.block.cancel_text %}
+        {{ content.block.cancel_text }}
```
ajmaddaford: The pull request looks good. We just need to check with @charlotte-hirst if there should be a line break here so it's consistent with the return to hub link.

berroar: Checked with @charlotte-hirst, have now added a line break to make it consistent with the return to hub link.

PR #18, opened 2019-12-16
Fix answer and when rules values mismatch errors

petechd:
This fixes existing routing errors detected by new validator method which checks for answer and when values mismatches. It adds a fix to census schema to make it pass validator check after merge of new validator branch. 

### How to review 
Build schemas and check if "term-time-location" routing uses default "Yes, an address outside the UK" routing from "another-address" question.


petechd:
[Validate answer values used in routing exist](https://trello.com/c/EbHSxOaI/3481-validate-answer-values-used-in-routing-exist)

PR #25, opened 2020-01-08
Add visible write ins

petechd:
This adds new boolean "open" property to radio and checkbox answers in certain questions described on [this Trello card](https://trello.com/c/g3esxHDn/3433-visible-write-ins-m). It requires [this validator branch](https://github.com/ONSdigital/eq-schema-validator/tree/Add-visible-write-ins) to be merged before we change it back to v3. Checkboxes "open" property cannot be tested as it is only partially implemented in Design System.

### How to review 
Check that open property works as expected and write-ins are visible in "birth_gender" and "sexual_identity" questions in census schemas. Note that these two questions were requested to be radios instead of checkboxes on [this Trello card](https://trello.com/c/Wu3oraFw/3424-voluntary-questions-m). Currently, this change is not implemented in N.I. census schemas.


petechd:
[visible write ins (M)](https://trello.com/c/g3esxHDn/3433-visible-write-ins-m)

app/jinja_filters.py
```diff
@@ -237,7 +237,11 @@ def __init__(self, option, index, form, answer):
 
         if option.detail_answer_id:
             detail_answer = form["fields"][option.detail_answer_id]
-            self.other = OtherConfig(detail_answer)
+            if "open" in answer_option["detail_answer"]:
+                answer_open = answer_option["detail_answer"]["open"]
+            else:
+                answer_open = False
```
iwootten: If you move do answer_open=False first, then you can remove the else.

tests/integration/components/test_radio_open.py
```diff
@@ -0,0 +1,10 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestOpenRadio(IntegrationTestCase):
+    def setUp(self):
+        super().setUp()
+        self.launchSurvey("test_radio_open")
+
+    def test_label_appears(self):
+        self.assertInBody("Enter your favourite drink")
```
iwootten: I don't think this test is necessary, as it's duplicating the behaviour in 'When I view a write-in radio and the open option is set to true, detail answer label should be displayed' functional test.

app/jinja_filters.py
iwootten: If you make the answer_open=False first, then you can remove the else.

tests/integration/components/test_radio_open.py
petechd: If we remove this test we won't have "required test coverage of 100%" and travis will fail.

iwootten: Ok, but you shouldn't need to create an integration test to test the jinja filter changes?

data-source/json/test_radio_open.json
```diff
@@ -0,0 +1,128 @@
+{
```
ajmaddaford: This should be named something like `test_radio_detail_answer_open.json` as it's specific to the detail answer in a radio being open rather than all of the radio options.

data-source/json/test_radio_open.json
```diff
@@ -0,0 +1,128 @@
...
+							"type": "Radio",
+							"id": "radio-open-true-answer",
+							"mandatory": false,
+							"voluntary": true,
+							"options": [{
+									"label": "Coffee",
+									"value": "Coffee"
+								},
+								{
+									"description": "Select to enter answer",
```
ajmaddaford: The description "Select to enter answer" should be removed as it's misleading when the detail answer is always open.

tests/functional/spec/components/radio/radio_open_answers.spec.js
```diff
@@ -0,0 +1,28 @@
+const RadioOpenTruePage = require('../../../generated_pages/radio_open/radio-open-true.page.js');
+const RadioOpenFalsePage = require('../../../generated_pages/radio_open/radio-open-false.page.js');
+const RadioOpenNonePage = require('../../../generated_pages/radio_open/radio-open-none.page.js');
+
+describe('Given I start a Radio survey with a write-in option', function() {
+  beforeEach(function() {
+    browser.openQuestionnaire('test_radio_open.json');
+    $(RadioOpenTruePage.coffee()).click();
```
ajmaddaford: I don't think this step should be in the `beforeEach` as it could have an effect on the behaviour of the radio options.

tests/functional/spec/components/radio/radio_open_answers.spec.js
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  it('When I view a write-in radio and the open option is set to true, detail answer label should be displayed', function() {
```
ajmaddaford: The tests are mostly consistent with descriptions formatted as "When ..., Then ..." so we should do that here.

tests/functional/spec/components/radio/radio_open_answers.spec.js
```diff
@@ -0,0 +1,28 @@
...
+    expect($(RadioOpenTruePage.otherDetail()).isDisplayed()).to.equal(true);
+  });
+
+  it('When I view a write-in radio and the open option is set to false, detail answer label should not be displayed', function() {
+    $(RadioOpenTruePage.submit()).click();
+    $(RadioOpenFalsePage.iceCream()).click();
```
ajmaddaford: This test shouldn't need to select a radio option to test that the detail answer isn't shown.

tests/functional/spec/components/radio/radio_open_answers.spec.js
```diff
@@ -0,0 +1,28 @@
...
+    expect($(RadioOpenFalsePage.otherDetail()).isDisplayed()).to.equal(false);
+  });
+
+  it('When I view a write-in radio and the open option is not set, detail answer label should not be displayed', function() {
+    $(RadioOpenFalsePage.submit()).click();
+    $(RadioOpenFalsePage.iceCream()).click();
+    $(RadioOpenFalsePage.submit()).click();
+    $(RadioOpenNonePage.blue()).click();
```
ajmaddaford: Same comment as above, test shouldn't need to select the 'blue' radio option to test that the detail answer isn't shown.

tests/integration/components/test_radio_open.py
petechd: I changed it and moved into jinja filters unit tests file

data-source/json/test_radio_open.json
petechd: I changed it to test_radio_detail_answer_open

data-source/json/test_radio_open.json
petechd: I removed it from test schema. There are still two census questions affected, but I had to ask someone for exact wording on these descriptions and waiting for reply now.

tests/functional/spec/components/radio/radio_open_answers.spec.js
petechd: I removed "beforeEach", thanks.

tests/functional/spec/components/radio/radio_open_answers.spec.js
petechd: Fixed. Thanks!

tests/functional/spec/components/radio/radio_open_answers.spec.js
petechd: I got rid of that, thanks!

tests/functional/spec/components/radio/radio_open_answers.spec.js
petechd: Removed, thanks!

tests/functional/spec/components/radio/radio_visible_answers.spec.js
```diff
@@ -0,0 +1,28 @@
...
+
+describe('Given I start a Radio survey with a write-in option', function() {
+  beforeEach(function() {
+    browser.openQuestionnaire('test_radio_detail_answer_visible.json');
+  });
+
+  it('When I view a write-in radio and the visible option is set to true, Then the detail answer label should be displayed', function() {
+    $(RadioVisibleTruePage.coffee()).click();
+    expect($(RadioVisibleTruePage.otherDetail()).isDisplayed()).to.equal(true);
+  });
```
iwootten: The other detail is displayed by default, so shouldn't need a click prior to this check?

iwootten: Is this testing the default behaviour, or behaviour after a click?

If this test is actually testing that the visible write in doesn't collapse after the coffee option is selected, there should probably be an additional test with an appropriate description. The click step then may be removed from the default behaviour test here.

data-source/json/test_radio_open.json
petechd: Ok, descriptions have been removed from census questions, test schema description has been refactored. Thanks!

tests/functional/spec/components/radio/radio_visible_answers.spec.js
petechd: Fixed the first test, added new one to cover "visible write in doesn't collapse" behaviour, thanks!

data-source/json/test_radio_detail_answer_visible.json
```diff
@@ -0,0 +1,128 @@
...
+							"type": "Radio",
+							"id": "radio-visible-true-answer",
+							"mandatory": false,
+							"voluntary": true,
+							"options": [{
+									"label": "Coffee",
+									"value": "Coffee"
+								},
+								{
+									"description": "Enter answer",
```
ajmaddaford: Minor, but this description looks odd in the page and should be removed.

petechd: Removed, thanks!

app/jinja_filters.py
```diff
@@ -237,7 +237,10 @@ def __init__(self, option, index, form, answer):
 
         if option.detail_answer_id:
             detail_answer = form["fields"][option.detail_answer_id]
-            self.other = OtherConfig(detail_answer)
+            answer_visible = False
+            if "visible" in answer_option["detail_answer"]:
+                answer_visible = answer_option["detail_answer"]["visible"]
```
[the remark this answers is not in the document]
petechd: I had a chat with Mebin about this change before but I was not sure if other people will want it, now I know. Thanks.

PR #27, opened 2020-01-09
Update runner to use new schema repo

iwootten:

Updated the runner load process to load in schema dependencies from the new eq-questionnaire-schemas repo to /schemas. Build now only loads templates and compiles translation files. This should hopefully reduce times on runner due the dependency on jsonnet now being removed and translation scripts being held there too. Additionally, the feature schemas are now held in test_schemas and there is no longer a step which copies test schemas into a single directory. Instead schema folders are listed and inspected at run time. A map is built of all possible schemas which loads them from their source paths if they exist. 

Schemas from eq-questionnaire-schemas are no longer installed by default and all tests with dependencies on them have been updated. If you want to install them, you can with **make load-schemas**.

### How to review 

Runner has been updated to use the new eq-questionnaire-schemas repo, ensure that the tests pass in using it.  Check all previous build dependencies have been removed if no longer needed.


iwootten:
[Separate questionnaire build process out of runner (L)](https://trello.com/c/xp4MT3YV/3547-separate-questionnaire-build-process-out-of-runner-l)

app/utilities/schema.py
```diff
@@ -25,6 +27,41 @@
 }
 
 
+def get_schema_path_map_for_language(language_code=DEFAULT_LANGUAGE_CODE):
```
ajmaddaford: Language code is always passed, so we should drop `DEFAULT_LANGUAGE_CODE` here.

app/utilities/schema.py
```diff
@@ -25,6 +27,41 @@
...
+def get_schema_path_map():
+    language_map_codes = ["en", "cy", "ga", "eo"]
+
+    return {
+        language_code: get_schema_path_map_for_language(language_code)
+        for language_code in language_map_codes
+    }
+
+
+SCHEMA_PATH_MAP = get_schema_path_map()
```
ajmaddaford: Any reason not to use `cache.memoize` on `get_schema_path_map` as done for `load_schema_from_name`?

app/routes/schema.py
```diff
@@ -18,11 +15,5 @@ def get_schema_json_from_name(schema_name):
...
 def list_schemas():
-    schema_path = "data/en"
-    schema_names = [
-        f.replace(".json", "")
-        for f in listdir(schema_path)
-        if isfile(join(schema_path, f)) and f.endswith(".json")
-    ]
-
-    return jsonify(schema_names)
+    schemas = list(SCHEMA_PATH_MAP["en"].keys())
```
ajmaddaford: The map shouldn't be exposed outside of `app.utilities.schema`. We should add a method to `app.utilities.schema` that returns a list of all schemas for a given language code.

tests/app/utilities/test_schema.py
```diff
@@ -44,3 +45,12 @@ def test_transform_schema_name_bad_case_type_raises_error():
...
     assert get_allowed_languages(schema_name, launch_language) == expected
+
+
+def test_get_schema_path_map():
+    schema_path_map = get_schema_path_map()
+
+    assert "en" in schema_path_map.keys()
+    assert "cy" in schema_path_map.keys()
+    assert "ga" in schema_path_map.keys()
+    assert "eo" in schema_path_map.keys()
```
ajmaddaford: Some more tests here would make it clearer what structure is generated from this method.

app/utilities/schema.py
iwootten: I'd prefer not to do this as it will impact every unit test we have which loads a schema due to it's dependency on flask_caching. I believe we can make used of improved caching methods available in 3.8 once we update. This method is only called once to generate this variable, maybe we can hold off?

app/utilities/schema.py
```diff
@@ -25,6 +27,41 @@
...
+
+
+def get_schema_path_map():
+    language_map_codes = ["en", "cy", "ga", "eo"]
+
+    return {
+        language_code: get_schema_path_map_for_language(language_code)
+        for language_code in language_map_codes
+    }
+
```
JamesGardiner: This can be  made more succinct by using the Path object.

```Python
from collections import ChainMap

def schema_path_maps_for_language(language_code):
    schema_dirs = ["schemas", "test_schemas"]
    
    for schema_dir in schema_dirs:
            yield {Path(p).with_suffix('').name: p for p in list(Path(schema_dir, language_code).glob('*.json'))}

def get_schema_path_map():
    language_map_codes = ["en", "cy", "ga", "eo"]

    return {
        language_code: ChainMap(*get_schema_path_map_for_language(language_code))
        for language_code in language_map_codes
    }
```

app/utilities/schema.py
```diff
@@ -25,6 +27,41 @@
...
+
+SCHEMA_PATH_MAP = get_schema_path_map()
+
+
+def schema_exists(language_code, schema_name):
+    return (
+        language_code in SCHEMA_PATH_MAP
+        and schema_name in SCHEMA_PATH_MAP[language_code]
```
JamesGardiner: If `language_code` isn't a key in `SCHEMA_PATH_MAP` this would fail with a KeyError, so the first `in` operation is redundant

app/utilities/schema.py
```diff
@@ -115,13 +161,8 @@ def _load_schema_file(schema_name, language_code):
...
     )
 
-    try:
-        with open(schema_path, encoding="utf8") as json_data:
-            return json.load(json_data, use_decimal=True)
-
-    except FileNotFoundError as e:
-        logger.error("no schema file exists", filename=schema_path)
-        raise e
+    with open(schema_path, encoding="utf8") as json_data:
```
JamesGardiner: The context manager isn't `json_data` so this is a sightly misleading variable name. Something like `file` or `fp` would be more accurate.

app/utilities/schema.py
iwootten: The first in operation is to prevent the key error and make this method that returns a boolean?

app/utilities/schema.py
iwootten: This may be more succinct, but introduces quite a lot of complexity. I'm not sure it's worth the confusion.

app/utilities/schema.py
JamesGardiner: Ah yeah, my mistake it will short circuit :+1:

app/utilities/schema.py
JamesGardiner: I find it easier to read.

Which lines are overly complex? It may be that there is a way to reduce the complexity and keep the more succinct approach.

JamesGardiner: Just saw the latest change - I think that's a good approach and answers the above question :+1:

app/utilities/schema.py
```diff
@@ -13,7 +14,7 @@
 
 logger = get_logger()
 
-DEFAULT_SCHEMA_DIR = "data"
+SCHEMA_DIRS = ["schemas", "test_schemas"]
```
JamesGardiner: This doesn't need to be a module level variable now. It's only used once in get_schema_path_map_for_language

app/utilities/schema.py
```diff
@@ -25,6 +26,37 @@
...
+    for schema_dir in SCHEMA_DIRS:
+        schema_files.extend(glob(f"{schema_dir}/{language_code}/*.json"))
+
+    return {
+        Path(schema_file).with_suffix("").name: schema_file
+        for schema_file in schema_files
+    }
+
+
+def get_schema_path_map():
```
JamesGardiner:
```suggestion
def get_schema_path_maps():
```

iwootten: There's only a single map returned from this method.

app/utilities/schema.py
iwootten: I think there's value keeping it there in order to keep configuration of schema behaviour together. The same could be said for the other module level variables in this file.

README.md
```diff
@@ -186,11 +186,11 @@ You can also individually run the `generate_pages` and `test_functional` yarn sc
 
 To generate the pages manually you can run the `generate_pages` scripts with the schema directory. Run it from the `tests/functional` directory as follows:
 
-`./generate_pages.py ../../data/en/ ./generated_pages -r "../../base_pages"`
+`./generate_pages.py ../../tests/schemas/en/ ./generated_pages -r "../../base_pages"`
```
MebinAbraham: Is this not meant to be:
```suggestion
`./generate_pages.py ../../tests_schemas/en/ ./generated_pages -r "../../base_pages"`
```

README.md
```diff
@@ -186,11 +186,11 @@ You can also individually run the `generate_pages` and `test_functional` yarn sc
...
 
 To generate a spec file with the imports included, you can use the `generate_pages.py` script on a single schema with the `-s` argument.
 
-`./generate_pages.py ../../data/en/test_multiple_piping.json ./temp_directory -r "../../base_pages" -s spec/test_multiple_piping.spec.js`
+`./generate_pages.py ../../tests/schemas/en/test_multiple_piping.json ./temp_directory -r "../../base_pages" -s spec/test_multiple_piping.spec.js`
```
MebinAbraham: Same here about file path.

scripts/run_validator.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+branch=v3
+docker pull onsdigital/eq-schema-validator:$branch
+docker run -d -p 5001:5000 "onsdigital/eq-schema-validator:$branch"
+sleep 3
```
MebinAbraham: Shouldn't be needed anymore.

README.md
iwootten: No, it's test_schemas, but close enough.

PR #30, opened 2020-01-21
Use GitHub actions

iwootten:

Migrated our jobs to use github actions instead of travis. Bumped the version of Python to 3.7.6 to match the latest release the setup-python provides.

### How to review 

Check that all workflow jobs execute successfully. Compare against what has been removed in the travis file and confirm no steps are missing.


iwootten:
[Use GitHub actions in favour of Travis for runner](https://trello.com/c/4du21NYs/3548-use-github-actions-in-favour-of-travis-for-runner)

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,158 @@
...
+  docker-push:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+        - name: Write app version
+          run: printf $(git rev-parse HEAD) > .application-version
+        - name: Tag
+          run: |
+            export BRANCH=$(git rev-parse --abbrev-ref HEAD)
+            export TAG=`if [ "$BRANCH" == "" ]; then echo "latest"; else echo branch-$BRANCH; fi`
```
MebinAbraham:
```suggestion
            export TAG=`if [ "$BRANCH" == "" ]; then echo "latest"; else echo "branch-$BRANCH"; fi`
```

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,158 @@
...
+            echo "::set-env name=TAG::$TAG"
```
MebinAbraham: What does this do?

scripts/run_tests.sh
```diff
@@ -8,11 +8,14 @@ set -e
 echo "Starting Validator"
 ./scripts/run_validator.sh
 
-echo "Running schema tests"
-./scripts/test_schemas.sh schemas/en
+echo "Running schema validation"
+./scripts/validate_test_schemas.sh test_schemas/en
```
MebinAbraham: `test_language.json` is in `test_schemas/cy` so we can remove passing the dir to test all `test_schemas`.

scripts/run_travis.sh
```diff
@@ -8,14 +8,17 @@ set -e
 echo "Starting Validator"
```
MebinAbraham: If we are keeping this file, it should not be named `run_travis`.

.github/workflows/pull_request.yml
iwootten: Sets a tag environment variable to be used in another step. Exporting it won't pass it to the next step. https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#set-an-environment-variable-set-env

iwootten: Sets a tag environment variable to be used in another step. Exporting it won't pass it to the next step. https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#set-an-environment-variable-set-env

scripts/run_tests.sh
iwootten: We can't as validator will complain about special characters in the welsh schemas.

.github/workflows/pull_request.yml
MebinAbraham: Ah cool. 👍

scripts/run_tests.sh
MebinAbraham: Ah ya but in that case, this needs to change to `./test_schemas/en` in here:
https://github.com/ONSdigital/eq-questionnaire-runner/blob/use-github-actions/scripts/validate_test_schemas.sh#L31

MebinAbraham:
There is still a `make travis` command in the Makefile.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,158 @@
...
+      - uses: actions/checkout@v2
+      - run: |
+          echo "::set-env name=PYTHON_VERSION::$(cat .python-version)"
+      - uses: actions/setup-python@v1
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+      - name: Install apt dependencies
+        run: sudo apt-get install libsnappy-dev libgconf-2-4
+      - name: Write app version
+        run: printf $(git rev-parse HEAD) > .application-version
```
MebinAbraham: I pointed my docker-compose to point at this branch, pulled images and started the services. But when I got to the `/status` endpoint on runner, it doesn't match the current `git rev-parse HEAD`. The version I get is `27309c700371dea3a8e0d414b2b7a7f76f617397`.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,158 @@
...
+      - uses: actions/setup-python@v1
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+      - name: Install apt dependencies
+        run: sudo apt-get install libsnappy-dev libgconf-2-4
+      - name: Install Pipenv
+        run: pip install pipenv==2018.11.26
+      - name: Cache virtualenv
+        id: cache-virtualenv
+        uses: actions/cache@v1
```
JamesGardiner: Actions should be referenced by commit hash rather than tag.
```yaml
uses: actions/cache@a505c2e
```

Tags can be opaquely updated, so this ensures that the expected code is checked out. See https://julienrenaux.fr/2019/12/20/github-actions-security-risk/ for further explanation

.github/workflows/pull_request.yml
iwootten: You need to actually update your local .application-version file prior to doing docker-compose build in order to replicate the steps that travis/actions do when they create a build. Otherwise .application-version will just be whenever you last updated it locally.

I've just done that and it points to this: https://github.com/ONSdigital/eq-questionnaire-runner/commits/5c7bc4ea16d946ae5be89df6f6086603d2971bcb

.github/workflows/pull_request.yml
iwootten: These aren't random actions in github marketplace, they are ones maintained in the official github org.

JamesGardiner: True - I was aiming more to set a precedent from the start so that should any future developers add actions from marketplace the sha is used. I'm not sure where else is best to capture that information :man_shrugging:

iwootten: I think the other thing is that this applies to other dependencies of ours - e.g. pulling images from docker. 

Those tags could theoretically be updated. We also assume the latest versions of packages should be used in pip through using pipenv, so there could be risks there we may not catch too.

I think until there is something that can maintain commit hashes for us within actions, I suggest we lock to sha's for non-github marketplace actions?

.github/workflows/pull_request.yml
MebinAbraham: Ah ofc. 👍

iwootten: Hmm, looks like it's incorrectly tagging things at the moment as branch-HEAD. Will investigate.

iwootten: Looks like this is down to how the checkout works. Have updated to use the available github variables and appears to be tagging correctly once again.

.github/workflows/pull_request.yml
ajmaddaford: I agree - use version for github actions and sha's for non-github actions.

PR #21, opened 2019-12-20
Add voluntary property to radio answer

petechd:
### What is the context of this PR?
This adds new boolean "voluntary" property to radio answer in certain questions described on [this Trello card](https://trello.com/c/Wu3oraFw/3424-voluntary-questions-m). It requires [this validator branch](https://github.com/ONSdigital/eq-schema-validator/pull/178) to be merged first.

### How to review 
Check that clear button and voluntary property works as expected in census schemas. Currently, this change is not implemented in N.I. census schemas.

PR #31, opened 2020-01-22
Add clear button functionality for voluntary Radios

ajmaddaford:
Adds a clear button to Radios with a `voluntary` property set to `true`. This pull request was previously worked on by @petechd in https://github.com/ONSdigital/eq-questionnaire-runner/pull/21 and extended to work when Javascript is disabled.

### How to review 
- Use the `test_radio_voluntary` questionnaire to test the functionality
- Also test with Javascript disabled


ajmaddaford:
[Voluntary questions (M)](https://trello.com/c/Wu3oraFw/3424-voluntary-questions-m)

templates/partials/answers/radio.html
```diff
@@ -1,10 +1,21 @@
...
   "id": answer["id"],
   "name": answer["id"],
   "legend": question_title,
   "legendClasses": "u-vh",
   "radios": map_radio_config(form, answer),
   "error": error
-}) }}
+} %}
+
+{% if 'voluntary' in answer and answer["voluntary"] %}
```
MebinAbraham: This can be simplified:
```suggestion
{% if answer.voluntary %}
```

templates/partials/answers/radio.html
```diff
@@ -1,10 +1,21 @@
...
+  {% set config = config | setAttribute("clearRadios", {
+    "text": _("Clear selection"),
+    "name": "action[clear_radios]",
+    "ariaClearText": "You can clear your answer using the clear selection button after the radio inputs",
+    "ariaClearedText": "You have cleared your answer"
```
MebinAbraham: Needs to be wrapped in `_()` for translation.

templates/partials/answers/radio.html
```diff
@@ -1,10 +1,21 @@
...
   "radios": map_radio_config(form, answer),
   "error": error
-}) }}
+} %}
+
+{% if 'voluntary' in answer and answer["voluntary"] %}
+  {% set config = config | setAttribute("clearRadios", {
+    "text": _("Clear selection"),
+    "name": "action[clear_radios]",
+    "ariaClearText": "You can clear your answer using the clear selection button after the radio inputs",
```
MebinAbraham: Needs to be wrapped in `_()` for translation.

app/views/handlers/question.py
```diff
@@ -155,3 +155,13 @@ def evaluate_and_update_section_status_on_list_change(self, list_name):
                 section_id=section_id,
                 list_item_id=list_item_id,
             )
+
+    def clear_radio_answers(self):
+        answer_ids_to_remove = []
+        for answer in self.rendered_block.get("question").get("answers"):
```
MebinAbraham: The `question` and `answers` are always a valid key right?. As in do we need `.get()`?

templates/partials/answers/radio.html
ajmaddaford: I don't think any of the aria text is translated, but I'll check with @jrbarnes9

app/views/handlers/question.py
ajmaddaford: Fair point, removed the `.get`s

PR #39, opened 2020-01-29
Move newagent init to allow instrumentation

iwootten:

A number of python modules are unable to be initialised in the python newrelic agent due to the point at which it is initialised in the runner code. 

This PR makes the newrelic init run first in order to instrument every library it can. It drops the version of grpcio to the last version, due to the following issue: https://discuss.newrelic.com/t/instrumentation-error-for-python-grpc-with-spanner/91912

Greenlet has also been removed from our dependencies as it does not appear to be used and newrelic will fail to instrument it due to being detected in the environment and never launched.

### How to review 

Create a user account and license key in newrelic and set the following env vars:

EQ_NEW_RELIC_ENABLED=True
NEW_RELIC_LICENSE_KEY=< your_license_key >
NEW_RELIC_APP_NAME=questionnaire-runner-< yourname >

You should then see stats being sent to newrelic and the following message should not appear for the timeframe this change has been used. 
Screen Shot 2020-01-29 at 16 06 36

This is an alternative to #40 and we need to decide which approach we'd prefer to go with.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -40,7 +40,7 @@ flask-login = "*"
 flask-wtf = "*"
 gevent = {version = "*",platform_python_implementation = "=='CPython'"}
 google-cloud-datastore = "*"
-grpcio = "*"
+grpcio = "==1.25.0"
 gunicorn = "*"
 newrelic = "*"
 pika = "*"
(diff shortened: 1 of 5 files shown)


iwootten:
[Newrelic: Move newrelic.agent.initialize() to allow instrumentation](https://trello.com/c/SgPLksii/3583-newrelic-move-newrelicagentinitialize-to-allow-instrumentation)

PR #40, opened 2020-01-30
Use newrelic agent

iwootten:

This modifies runner to use the preferred newrelic agent on startup. Additionally this prevents library wrapping warnings within newrelic interface. A new make command 'gunicorn' has been added which can start the new relic agent based on appropriate environment vars having been set. A new gunicorn config.py has been added in order to keep settings together as it was impossible to quote vars in a way codacy was happy with. At the same time I've also stripped the server header.

It drops the version of grpcio to the last version, due to the following issue: https://discuss.newrelic.com/t/instrumentation-error-for-python-grpc-with-spanner/91912

Greenlet has also been removed from our dependencies as it does not appear to be used and newrelic will fail to instrument it due to being detected in the environment and never launched

Currently there is a monkeypatching warning shown on startup by gevent but newrelic have indicated it shouldn't be an issue. https://discuss.newrelic.com/t/seeing-monkeypatching-error-while-running-newrelic-admin-with-python-3-7-gevent-and-gunicorn/73893. It appears that gevent is showing this warning in many cases where it need not: https://github.com/gevent/gevent/pull/1239

This is an alternative to https://github.com/ONSdigital/eq-questionnaire-runner/pull/39 and we need to decide which approach we'd prefer to go with.
### How to review 

Create a user account and license key in newrelic and set the following env vars:

EQ_NEW_RELIC_ENABLED=True
NEW_RELIC_LICENSE_KEY=< your_license_key >
NEW_RELIC_APP_NAME=questionnaire-runner-< yourname >

You should then see stats being sent to newrelic and the following message should not appear for the timeframe this change has been used. 
Screen Shot 2020-01-29 at 16 06 36


iwootten:
[Newrelic: Move newrelic.agent.initialize() to allow instrumentation](https://trello.com/c/SgPLksii/3583-newrelic-move-newrelicagentinitialize-to-allow-instrumentation)

iwootten:
[AD3 – Suppress Disclosure Through HTTP Headers (Server: Gunicorn/19.9.0) (S)](https://trello.com/c/JRRCBRgt/3545-ad3-suppress-disclosure-through-http-headers-server-gunicorn-1990-s)

.development.env
```diff
@@ -16,3 +16,6 @@ EQ_REDIS_PORT=6379
 EQ_SECRETS_FILE=dev-secrets.yml
 EQ_KEYS_FILE=dev-keys.yml
 AWS_DEFAULT_REGION=eu-west-1
+GUNICORN_WORKERS=1
+GUNICORN_KEEP_ALIVE=2
+GUNICORN_CMD_ARGS=-c gunicorn_config.py
```
ajmaddaford: I think as this contains a space it may need to be wrapped in quotes to make sure nothing odd happens. In the example in the docs it's quoted - https://docs.gunicorn.org/en/stable/settings.html#settings.

run.sh
```diff
@@ -0,0 +1,12 @@
...
+if [ -n "$SECRETS_S3_BUCKET" ]; then
+    echo "Load Secrets from S3 Bucket [$SECRETS_S3_BUCKET]"
+    aws s3 sync "s3://$SECRETS_S3_BUCKET/" /secrets
+fi
+
+if [ "$EQ_NEW_RELIC_ENABLED" == "True" ]; then
+    NEW_RELIC_CONFIG_FILE=newrelic.ini newrelic-admin run-program gunicorn application:application
+else
+    gunicorn application:application
+fi
```
ajmaddaford: I realise I suggested `run.sh`, but maybe `run_gunicorn.sh` would be better 🙄

.development.env
iwootten: I've attempted adding the quotes and docker fails with them.

JamesGardiner: As an env var is parsed as a string the current implementation of gunicorn should parse it correctly https://github.com/benoitc/gunicorn/blob/7d8c92f48a6d9cee6b15fbdade6981721182d073/gunicorn/config.py#L71

iwootten: Gunicorn does with or without, but we also specify the var as part of the docker setup which doesn't like the quotes.

PR #42, opened 2020-02-04
Add lookup to text field

iwootten:

Add the typeahead component for use on a textfield and schema by adding a new parameter 'url' to textfield.

### How to review 

Open the schema. Type a few letters into the schema and note the new suggestion in the visible dropdown.


.github/workflows/pull_request.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       runs-on: ubuntu-latest
       services:
         validator:
-          image: onsdigital/eq-questionnaire-validator:latest
+          image: onsdigital/eq-questionnaire-validator:add-lookup-to-textfield
```
iwootten: Revert following validator merge.

scripts/run_validator.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
-branch=latest
+branch=add-lookup-to-text-field
```
iwootten: Revert following validator merge

iwootten:
[[C] Add lookup functionality to text field answer type (M)](https://trello.com/c/7ypG3adI/3442-c-add-lookup-functionality-to-text-field-answer-type-m)

templates/partials/answers/textfield.html
```diff
@@ -17,4 +17,21 @@
   "legendClasses": "u-vh",
   "mutuallyExclusive": mutuallyExclusive,
   "error": error
-}) }}
+} %}
+
+{% if 'url' in answer and answer["url"] %}
```
MebinAbraham: This can be simplified to `answer.url` in jinja even if url doesn't exist.

test_schemas/en/test_textfield_typeahead.json
```diff
@@ -0,0 +1,65 @@
...
+                        {
+                            "type": "Question",
+                            "id": "country-block",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "country-answer",
+                                        "label": "What is your country of birth?",
+                                        "mandatory": false,
+                                        "url": "https://gist.githubusercontent.com/rmccar/c123023fa6bd1b137d7f960c3ffa1fed/raw/368a3ea741f72c62c735c319ff7e33e3c1bfdc53/country-of-birth.json",
```
MebinAbraham: I don't think a personal gist should be committed?

iwootten: This is where the data currently is.

iwootten:
> How is this being tested?

"A functional test will follow once we have the data deployed to the cdn."

test_schemas/en/test_textfield_typeahead.json
MebinAbraham: Is it being hosted in a CDN soon?. The feature should have a test before it is merged,

iwootten: Once the work has been scheduled to do so.

The only way of getting it into a cdn seems to be a pipeline and the existing one for the design system seems to go via aws. Spoke with @ajmaddaford on how we should approach this and we decided that setting up an appropriate pipeline was a little out of the scope of this card. Hence continuing to use the gist.

MebinAbraham: So we should still write a functional test without the data being hosted in CDN then.

iwootten: I don't think there's a nice way of doing this - it would require custom endpoints within runner in order to modify csp policies.

MebinAbraham: Ah right -_-

iwootten: Possibly, but no-one realised that until I'd done the work to integrate. So we can either add the schema now and make the test part of any cdn work, or just throw this away. We chose the former.

MebinAbraham: Spoke with @ajmaddaford and agreed we can progress this by adding an integration test to ensure that the text field is rendered with the correct URL property and have confidence that the typeahead itself is tested in the design system.

templates/partials/answers/textfield.html
```diff
@@ -17,4 +17,21 @@
...
+  {% set config = config | setAttribute("typeahead", {
+    "instructions": "Use up and down keys to navigate suggestions once you\'ve typed more than two characters. Use the enter key to select a suggestion. Touch device users, explore by touch or with swipe gestures.",
+    "ariaYouHaveSelected": "You have selected",
+    "ariaMinChars": "Enter 3 or more characters for suggestions.",
+    "ariaOneResult": "There is one suggestion available.",
+    "ariaNResults": "There are {n} suggestions available.",
+    "ariaLimitedResults": "Results have been limited to 10 suggestions. Type more characters to improve your search.",
+    "moreResults": "Continue entering to improve suggestions",
+    "resultsTitle": "Suggestions",
+    "noResults": "No results found",
```
MebinAbraham: Do we need to mark these for translations?. I'm not sure about arias.

iwootten: Yes, I would imagine so.

test_schemas/en/test_textfield_typeahead.json
```diff
@@ -0,0 +1,60 @@
...
+                        {
+                            "type": "Question",
+                            "id": "country-block",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "country-answer",
+                                        "label": "What is your country of birth?",
+                                        "mandatory": false,
+                                        "url": "/json/countries",
```
ajmaddaford: I don't think it's very clear what this property is for, something like `suggestions_url` would be clearer.

app/routes/static.py
```diff
@@ -22,3 +24,10 @@ def privacy():
 @static_blueprint.route("/accessibility", methods=["GET"])
 def accessibility():
     return render_template("static/accessibility")
+
+
+@static_blueprint.route("/json/countries", methods=["GET"])
```
JamesGardiner: This could probably do with a comment to mark it as somewhere that will be deprecated once the relevant file is on the CDN

templates/partials/answers/textfield.html
```diff
@@ -17,4 +17,21 @@
...
+} %}
+
+{% if answer.suggestions_url %}
+  {% set config = config | setAttribute("typeahead", {
+    "instructions": _("Use up and down keys to navigate suggestions once you\'ve typed more than two characters. Use the enter key to select a suggestion. Touch device users, explore by touch or with swipe gestures."),
+    "ariaYouHaveSelected": _("You have selected"),
+    "ariaMinChars": _("Enter 3 or more characters for suggestions."),
+    "ariaOneResult": _("There is one suggestion available."),
+    "ariaNResults": _("There are {n} suggestions available."),
+    "ariaLimitedResults": _("Results have been limited to 10 suggestions. Type more characters to improve your search."),
```
MebinAbraham: We don't mark aria for translation in other places, can't remember if there was a specific reason for this. Worth checking.

tests/functional/spec/typeahead.js
```diff
@@ -0,0 +1,10 @@
+const TypeaheadPage = require('../generated_pages/textfield_typeahead/country-block.page.js');
+
+describe('Typeahead', function() {
+  it('Given text entered into a textfield typeahead, it will show suggestions', function() {
```
MebinAbraham: Can we update this to follow the pattern `Given When Then`?

MebinAbraham:
Can we change all reference of `typeahead` to look up` since that is the official name I believe?

iwootten:
> Can we change all reference of `typeahead` to look up` since that is the official name I believe?

I think theres debate on this given it's typeahead in the design system: https://ons-design-system.netlify.com/components/input/

PR #46, opened 2020-02-12
Age confirmation playback

petechd:
This adds new "age playback" where persons age is now represented in number of months instead of 0 years in case of a person younger than 1 year. Also, 1 year was displayed as plural form "1 years", this is now fixed. Number of days is also displayed, but I'm opened to reviewers suggestions on how it should be implemented. It calculates difference of any two dates and doesn't need to be associated  with person's age only. Because of that I changed transforms name to "calculate_difference" and placeholder to "age" in related repos: schemas and validator.

### How to review 
Use local runner to check that age confirmation question displays singular and plural forms of years, months and days depending on the dates difference.
Check if extra unit tests cover all the changes.
This PR uses age_confirmation_playback branch of eq-questionnaire-validator and eq-questionnaire-schemas.


petechd:
[Age confirmation playback](https://trello.com/c/vylaiCH1/2937-age-confirmation-playback)

app/questionnaire/placeholder_transforms.py
```diff
@@ -80,14 +81,24 @@ def format_number(self, number):
...
+            if time.years == 1:
+                return f'{time.years} {lazy_gettext("year")}'
+            return f'{time.years} {lazy_gettext("years")}'
+        if time.months:
+            if time.months == 1:
+                return f'{time.months} {lazy_gettext("month")}'
+            return f'{time.months} {lazy_gettext("months")}'
+        if time.days == 1:
+            return f'{time.days} {lazy_gettext("day")}'
+        return f'{time.days} {lazy_gettext("days")}'
```
ajmaddaford: Babel has support for this with `format_timedelta` 
 - http://babel.pocoo.org/en/latest/api/dates.html#babel.dates.format_timedelta. We should look to use that unless it's unsuitable.

petechd: I used format_timedelta and it's 3 lines now. Thanks!

app/questionnaire/placeholder_transforms.py
```diff
@@ -82,22 +81,11 @@ def format_number(self, number):
...
-                return f'{time.years} {lazy_gettext("year")}'
-            return f'{time.years} {lazy_gettext("years")}'
-        if time.months:
-            if time.months == 1:
-                return f'{time.months} {lazy_gettext("month")}'
-            return f'{time.months} {lazy_gettext("months")}'
-        if time.days == 1:
-            return f'{time.days} {lazy_gettext("day")}'
-        return f'{time.days} {lazy_gettext("days")}'
+        time = format_timedelta(timedelta, locale=flask_babel.get_locale())
```
ajmaddaford: `self.locale` should be used here to remove the need for `flask_babel`.

app/questionnaire/placeholder_transforms.py
```diff
@@ -80,13 +80,27 @@ def format_number(self, number):
         return ""
 
     @staticmethod
-    def calculate_years_difference(first_date, second_date):
-        return str(
-            relativedelta(
-                PlaceholderTransforms.parse_date(second_date),
-                PlaceholderTransforms.parse_date(first_date),
-            ).years
+    def calculate_difference(first_date, second_date):
```
MebinAbraham: @petechd Should the method name be more descriptive?. Possibly `calculate_time_difference` or something along those lines?
@ajmaddaford thoughts?

petechd: It's not the most accurate method to calculate time difference as it will never return amount of years and months or months and days in one string, only one of two, so it's best suited to calculate age?!

MebinAbraham:
Functionally works as expected. Just should decide on https://github.com/ONSdigital/eq-questionnaire-runner/pull/46#discussion_r380107509

app/questionnaire/placeholder_transforms.py
ajmaddaford: How about `calculate_date_difference`?

I think we'll come back to this when we know more about the requirements for date formatting, for example we may want to work out the difference in years and months.

PR #48, opened 2020-02-12
Updating gunicorn_config property from keep_alive to keepalive

pricem14pc:
While the gunicorn command line argument is `--keep-alive` the corresponding gunicorn_config.py  property is `keepalive` as documented here http://docs.gunicorn.org/en/stable/settings.html#keepalive

When deployed into GCP the helm value for GUNICORN_KEEP_ALIVE of 650 seconds was not being set in the config class. This caused connections to default to 2 seconds causing disconnections with the loadbalancer resulting in 502 bad gateway when under (light) load.

Test that when deployed to GCP the load balancer doesn't throw 502 "backend_connection_closed_before_data_sent_to_client" errors when under load


pricem14pc:
[gunicorn GUNICORN_KEEP_ALIVE is not being used and keepalive defaults to 2 seconds](https://trello.com/c/bsspHtkE/3662-gunicorn-gunicornkeepalive-is-not-being-used-and-keepalive-defaults-to-2-seconds)

PR #49, opened 2020-02-12
Implement strings based on a number of items

iwootten:

Adds the ability to use plural variations of a string using babel to look up an appropriate key based on a value. Based on the proposal found [here](https://github.com/ONSdigital/eq-schema-validator/blob/v3/doc/decisions/0003-plural-forms-in-schema-text.md). Added the ability to use the length of a list as a placeholder.

Both english and welsh rule maps are added but translation is not handled as part of this card.

Depends on the [following validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/4), which ignores validation of text_plural labels.

### How to review 

Check you agree with the approach and that tests pass.

### DIFF
--- a/app/questionnaire/plural_forms.py
+++ b/app/questionnaire/plural_forms.py
@@ -1,0 +1,27 @@
+from babel.plural import PluralRule
+
+from app.questionnaire.questionnaire_schema import DEFAULT_LANGUAGE_CODE
+
+
+def get_plural_form_key(count, language=DEFAULT_LANGUAGE_CODE):
+    mappings = {
+        "en": {"one": "n is 1"},
+        "cy": {
+            "zero": "n is 0",
+            "one": "n is 1",
+            "two": "n is 2",
+            "few": "n is 3",
+            "many": "n is 6",
+        },
+        "ga": {
+            "one": "n is 1",
+            "two": "n is 2",
+            "few": "n in 3..6",
+            "many": "n in 7..10",
+        },
+        "eo": {"one": "n is 1"},
+    }
+
+    plural_rule = PluralRule(mappings[language])
+
+    return plural_rule(count)
--- a/app/views/handlers/question.py
+++ b/app/views/handlers/question.py
@@ -63,7 +63,6 @@ class Question(BlockHandler):
         context["return_to_hub_url"] = self.get_return_to_hub_url()
 
         if "list_summary" in self.rendered_block:
-
             list_collector_context = ListCollectorContext(
                 language=self._language,
                 schema=self._schema,
@@ -79,6 +78,7 @@ class Question(BlockHandler):
                 ),
                 "editable": False,
             }
+
         return context
 
     def handle_post(self):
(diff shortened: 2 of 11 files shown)


iwootten:
[[B] Implement strings based on a number of items (L)](https://trello.com/c/pKafSvaX/3556-b-implement-strings-based-on-a-number-of-items-l)

app/views/handlers/question.py
```diff
@@ -134,17 +135,14 @@ def get_return_to_hub_url(self):
...
-            else:
-                question_title = question["title"]["text"]
-
-            page_title = f'{question_title} - {self._schema.json["title"]}'
+    def _get_page_title(self, question):
+        question_title = (
+            question["title"]
+            if isinstance(question["title"], str)
+            else question["title"]["text"]
+        )
```
ajmaddaford: When I run the test schema I get an error here on the `confirm-number-of-people` page:

```
...
File "/Users/andrew/dev/eq-questionnaire-runner/app/views/handlers/question.py", line 142, in _get_page_title
    else question["title"]["text"]
KeyError: 'text'
```

iwootten: This is down to not knowing what to render for a plural title until it has passed through the renderer. It is conflated with placeholder rendering. As discussed, I've added a workaround to use 'other' in titles (which should always be there).

test_schemas/en/test_plural_forms.json
```diff
@@ -0,0 +1,269 @@
+{
```
ajmaddaford: As the plural count can come from a number answer we should use that for the test schema rather than have to create list collectors etc. We could have something as simple as one question asking for a number of things and the next question confirming it.

app/questionnaire/plural_forms.py
```diff
@@ -0,0 +1,18 @@
+from babel.plural import PluralRule
+
+
+def get_plural_form_key(count, language="en"):
```
ajmaddaford: We should use `DEFAULT_LANGUAGE_CODE` rather than hardcode `en` here.

tests/integration/questionnaire/test_questionnaire_page_titles.py
```diff
@@ -100,3 +100,12 @@ def test_content_page_should_use_nested_content_text_in_page_title_if_it_exists(
...
         )
+
+    def test_plural_page_title_should_use_other_title(self):
+        self.launchSurvey("test_plural_forms")
+
+        self.post({"number-of-people-answer": 0})
+
+        self.assertEqualPageTitle(
+            "… people live here, is this correct? - Test Plural Forms"
+        )
```
ajmaddaford: We should also have integration tests for the summary content (question and answer).

app/questionnaire/plural_forms.py
```diff
@@ -0,0 +1,20 @@
...
+
+def get_plural_form_key(count, language=DEFAULT_LANGUAGE_CODE):
+    mappings = {
+        "en": {"one": "n is 1"},
+        "cy": {
+            "zero": "n is 0",
+            "one": "n is 1",
+            "two": "n is 2",
+            "few": "n is 3",
+            "many": "n is 4",
```
ajmaddaford: I think this should be `n is 6` to match the gettext plural forms we use.

scripts/run_validator.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
-branch=add-lookup-to-textfield
+branch=implement-strings-on-number-of-items
```
MebinAbraham: Can be reverted.

PR #50, opened 2020-02-13
Change interstitial page content title

petechd:
This adds level 2 heading to 'You will need to know' title element of "content" on interstitial page (interstitial block main title is not affected). The change is described on [this Trello](https://trello.com/c/1pKKk6qx/3626-interstitial-page-you-will-need-to-know-should-be-a-level-2-heading).
### How to review 
Make sure you have loaded latest templates. Check if 'You will need to know' title has a h2 class and that there are no issues with other interstitial pages.


petechd:
[Interstitial page 'You will need to know' should be a level 2 heading](https://trello.com/c/1pKKk6qx/3626-interstitial-page-you-will-need-to-know-should-be-a-level-2-heading)

templates/interstitial.html
```diff
@@ -8,6 +8,21 @@
...
+      <p>{{item.description}}</p>
+    {% endif %}
+    {%- if item.list -%}
+      <ul>
+        {%- for list_item in item.list -%}
+          <li>{{list_item}}</li>
+        {% endfor %}
+      </ul>
+    {% endif %}
+  {% endfor %}
```
MebinAbraham: This is almost identical to what is in `contents.html`, the only thing that changes is the tag for `item.title`. I think we could just set a property like `title_type` which determines whether to use `h2` or a `strong` tag.

petechd: I moved it to one file and "if" based on title type.

templates/partials/contents.html
```diff
@@ -1,6 +1,10 @@
 {% for item in contents %}
   {%- if item.title -%}
-    <strong class="u-mb-s">{{item.title}}</strong>
+    {%- if title_type == 'interstitial' -%}
```
iwootten: Can't the block type be checked here rather than having to set another variable?

PR #52, opened 2020-02-14
Update chromedriver

iwootten:

The version of chromedriver we use is a number of releases old (it is installed as a dependency of wdio-chromedriver-service), which due to a minor release of Ubuntu 18.0.4.4 is causing our functional tests to fail.

### How to review 
Ensure all functional tests pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "babel-eslint": "^10.0.1",
     "babel-register": "^6.26.0",
     "chai": "^4.2.0",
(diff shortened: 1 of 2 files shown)


iwootten:
For what it's worth, the errors we're receiving in 80 in terms of timeout shouldn't be considered something to be concerned about: https://stackoverflow.com/questions/60114639/timed-out-receiving-message-from-renderer-0-100-log-messages-using-chromedriver

PR #60, opened 2020-02-21
Custom section summaries

JamesGardiner:
Updates Runner to use a new `summary` key in questionnaire schema `section` objects.
Deprecates `add_link_text` and `empty_list_text` from `ListCollector` `summary` objects.
Deprecates `show_on_section_summary` from `ListCollector` objects.

https://github.com/ONSdigital/eq-questionnaire-validator/pull/9

### How to review 
Make sure tests pass.
Ensure functionality not impacted.

### Checklist


scripts/run_validator.sh
```diff
@@ -1,5 +1,28 @@
 #!/usr/bin/env bash
 
-branch=latest
-docker pull onsdigital/eq-questionnaire-validator:$branch
-docker run -d -p 5001:5000 "onsdigital/eq-questionnaire-validator:$branch"
+checks=4
+branch=custom-section-summaries
```
JamesGardiner: Revert before merge

JamesGardiner:
[Implement custom section summaries](https://trello.com/c/ZYz6mKNZ/3675-implement-custom-section-summaries)

templates/summary.html
```diff
@@ -16,15 +16,33 @@ <h1>{{ _("Check your answers and submit") }}</h1>
...
-
-    <h1>{{content.summary.title}}</h1>
-      {% call onsPanel({
-        "classes": "print__hidden u-mb-l"
-      }) %}
-        <p>{{ _("Check your answers before continuing to the next section") }}</p>
-      {% endcall %}
+  {% elif content.summary.summary_type == 'SectionSummary' %}
+    {% if content.summary.list_summaries %}
+        <h1>{{content.summary.title}}</h1>{% for list_summary in content.summary.list_summaries %}
```
LJBabbage: Might want to extract this out, so if we get multiple custom summaries, summary.html isn't massive

app/views/contexts/summary_context.py
```diff
@@ -67,8 +73,24 @@ def get_title_for_location(self, location):
...
+        section_id = self._schema.get_section_id_for_block_id(current_location.block_id)
+        section = self._schema.get_section(section_id)
+        section_summary = self._schema.get_summary_for_section(section_id)
         block = self._schema.get_block(current_location.block_id)
 
+        if section_summary:
+            return {
+                "summary": {
+                    "title": block["title"],
+                    "list_summaries": list(
```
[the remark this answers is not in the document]
JamesGardiner: It is - but only list summaries are implemented so far :+1:

app/views/contexts/summary_context.py
```diff
@@ -78,3 +100,67 @@ def section_summary(self, current_location):
...
+        routing_path = self._router.routing_path(
+            section["id"], current_location.list_item_id
+        )
+
+        list_collector_blocks = self._schema.get_list_collectors_for_section(
+            section, for_list=summary["for_list"]
+        )
+
+        for list_collector_block in list_collector_blocks:
+            add_link = url_for(
```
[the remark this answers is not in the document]
JamesGardiner: I'm going to keep it as `add_link`, as the key that is searched for in the resulting dictionary is add_link.

I Agree that this function is becoming too large here. I'll take a look at refactoring it out.

Makefile
```diff
@@ -24,7 +26,7 @@ test-unit:
 test-functional:
 	pipenv run ./scripts/run_tests_functional.sh
 
-validate-test-schemas:
+validate-test-schemas: run-validator
```
iwootten: The naming on these commands was changed to indicate they're running independently of one another for the github workflows. Making run a prerequisite now makes this out of sync with the  workflow and other action repos.

tests/app/questionnaire/conftest.py
```diff
@@ -569,6 +569,69 @@ def mock_relationship_collector_schema():
     }
 
 
+@pytest.fixture()
+def section_with_custom_summary():
```
iwootten: Unrelated to this but we need to look at a better method of organising fixture partials as keeping large chunks of json in conftest files for single tests is getting unmanageable.

scripts/run_validator.sh
```diff
@@ -1,5 +1,28 @@
...
+        fi
+        (( checks-- ))
+    else
+        echo "Schema Validator Running"
+        (( checks=0 ))
+    fi
+
+done
+
+exit 0
```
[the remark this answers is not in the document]
JamesGardiner: Is there a reason the Makefile isn't used in the Github Action `validate-schema`?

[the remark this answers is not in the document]
iwootten: The original reason behind not using the makefile is so it isn’t dependent on python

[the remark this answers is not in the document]
JamesGardiner: I'll revert this, and create a backlog item to track changing this across repos

tests/app/questionnaire/conftest.py
JamesGardiner: Card created on backlog :+1:

app/views/contexts/summary_context.py
```diff
@@ -66,15 +115,62 @@ def get_title_for_location(self, location):
...
+            section, for_list=summary["for_list"]
+        )[0]
+
+        add_link = self._add_link(
+            summary, current_location, section, list_collector_block
+        )
+
+        rendered_list_context = self.list_context(
+            list_collector_block, current_location.block_id
+        )
```
ajmaddaford: Doesn't this render the list items according to the `summary` element on the list collector block, rather than using the `item_title` element of the section summary?

app/views/contexts/list_collector_context.py
```diff
@@ -1,32 +1,30 @@
...
-        return rendered_summary["item_title"]
+    def __call__(self, list_collector_block, return_to=None):
+        return {
+            "list": {
+                "list_items": list(
+                    self._build_list_items_summary_context(
+                        list_collector_block, return_to=return_to
+                    )
+                ),
+                "editable": True,
```
ajmaddaford: `editable` should be passed in and used to determine if the edit/remove links should be added to the context.

app/views/contexts/summary_context.py
```diff
@@ -53,8 +74,36 @@ def summary(self, collapsible):
...
+        summary_context = {
+            "summary": {
+                "title": self._title_for_location(current_location),
+                "summary_type": "SectionSummary",
+                "answers_are_editable": True,
+                "collapsible": block.get("collapsible", False),
+            }
+        }
+
+        if not section_summary:
```
ajmaddaford: `section_summary` should be named something like `custom_section_summary` as either way a section summary is being generated and returned. 

We should also consider whether some of this logic e.g. deciding if a default or custom section summary is required, is better placed in the `SectionSummary` handler.

templates/summary.html
```diff
@@ -16,15 +16,36 @@ <h1>{{ _("Check your answers and submit") }}</h1>
...
-  {% elif content.summary.summary_type in ('SectionSummary') %}
-    {% set group = content.summary.groups %}
-
-    <h1>{{content.summary.title}}</h1>
-      {% call onsPanel({
-        "classes": "print__hidden u-mb-l"
-      }) %}
-        <p>{{ _("Check your answers before continuing to the next section") }}</p>
-      {% endcall %}
+  {% elif content.summary.summary_type == 'SectionSummary' %}
```
ajmaddaford: There is a `sectionsummary.html` that we should make use of rather than the nested elif logic in this template.

app/views/contexts/list_collector_context.py
JamesGardiner: Should the default be `True`?

templates/sectionsummary.html
```diff
@@ -1 +1,60 @@
...
+              <a data-qa="add-item-link" href="{{summary.add_link}}">{{summary.add_link_text}}</a>
+            </div>
+          </div>
+        {% endif %}
+      {% endfor %}
+  {% endif %}
+
+  <div class="u-mb-m">
+    {% include 'partials/summary/summary.html' %}
+  </div>
```
ajmaddaford: Should this `div` block be in the `if content.summary.groups` conditional?

templates/sectionsummary.html
```diff
@@ -1 +1,60 @@
-{% extends 'summary.html' %}
+{% extends 'layouts/_questionnaire.html' %}
+{% from "components/panel/_macro.njk" import onsPanel %}
+{% from "components/button/_macro.njk" import onsButton %}
+
+{% import 'macros/helpers.html' as helpers %}
+
+{% set summary_page_title = _("Summary") + " - " + survey_title %}
```
ajmaddaford: Should this should use the section summary title?

tests/functional/generate_pages.py
```diff
@@ -388,9 +388,9 @@ def process_calculated_summary(answers, page_spec):
         page_spec.write(CALCULATED_SUMMARY_LABEL_GETTER.substitute(answer_context))
 
 
-def process_list_collector_summary(schema_data, page_spec):
+def process_summary(schema_data, page_spec):
     for section in schema_data["sections"]:
-        list_collector_blocks = QuestionnaireSchema.get_visible_list_blocks_for_section(
+        list_collector_blocks = QuestionnaireSchema.get_list_collectors_for_section(
```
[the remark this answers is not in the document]
JamesGardiner: The section won't always have a `summary` attribute

app/routes/questionnaire.py
```diff
@@ -406,17 +406,14 @@ def get_view_submission(schema):
...
-                }
-            }
+            is_view_submission_response_enabled = is_view_submitted_response_enabled(
+                schema.json
+            )
+            context = final_summary_context(
+                answers_are_editable=False,
+                is_view_submission_response_enabled=is_view_submission_response_enabled,
+            )
+            del context["summary"]["summary_type"]
```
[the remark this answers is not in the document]
JamesGardiner: It's was needed for the summary context template, but broke the view submission (which uses the summary partial template). Now that I have changed the templates, this may not be required. I'll retest without the delete :+1:

app/questionnaire/questionnaire_schema.py
```diff
@@ -198,14 +201,22 @@ def get_questions(self, question_id):
...
     @staticmethod
-    def get_visible_list_blocks_for_section(section):
-        list_collector_blocks = []
-        for group in section["groups"]:
-            for block in group["blocks"]:
-                shown = block.get("show_on_section_summary", True)
-                if block["type"] == "ListCollector" and shown:
-                    list_collector_blocks.append(block)
-        return list_collector_blocks
+    def get_list_collectors_for_section(section, for_list=None):
```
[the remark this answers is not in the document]
JamesGardiner: See above - section won't always have a summary key

app/routes/questionnaire.py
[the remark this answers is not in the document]
JamesGardiner: Tested without the delete... It's no longer needed

tests/functional/generate_pages.py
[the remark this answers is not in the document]
ajmaddaford: As discussed, the only way to summarise a list is to have a `summary` attribute.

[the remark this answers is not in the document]
JamesGardiner: Added summary attribute :+1: also `edit_block` and `remove_block` kwargs now added to the method signature, as these were previously accessed via the list collector block.

app/views/contexts/section_summary_context.py
```diff
@@ -0,0 +1,160 @@
...
+        self.list_context = ListContext(
+            self._language,
+            self._schema,
+            self._answer_store,
+            self._list_store,
+            self._progress_store,
+            self._metadata,
+        )
+
+    def __call__(self, current_location):
```
LJBabbage: This whole bit just confuses me. So if our current location doesn't have a block id, like a section summary, its set as not able to collapse. However if it's a section summary why would we then build a summary for it.

JamesGardiner: Discussed offline :+1:

templates/sectionsummary.html
JamesGardiner: As in `content.summary.title`?

JamesGardiner: I'm not sure this is even needed here actually. I'll remove it and test

app/routes/questionnaire.py
```diff
@@ -47,7 +47,7 @@
 from app.views.contexts.metadata_context import (
     build_metadata_context_for_survey_completed,
 )
-from app.views.contexts.summary_context import SummaryContext
+from app.views.contexts import FinalSummaryContext
```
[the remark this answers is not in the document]
JamesGardiner: Good point - I guess `QuestionnaireSummary` as to the user the context will be singular?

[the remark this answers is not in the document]
JamesGardiner: Although, it is referred to as FinalSummaryPage  in the functional test spec - any views on this @ajmaddaford / @LJBabbage ?

app/views/contexts/section_summary_context.py
```diff
@@ -0,0 +1,160 @@
...
+            ]
+        }
+
+    def _title_for_location(self, location):
+        title = self._schema.get_section(location.section_id).get("title")
+        if location.block_id:
+            try:
+                title = self._schema.get_block(location.block_id).get("title")
+            except KeyError:
+                pass
```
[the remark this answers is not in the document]
JamesGardiner: Agree we should have a set of guidelines adopted it - it seems to be something that is easy to put on the backburner but would be helpful. In this case I guess you are referring to the try/except?

Would:
```python
title = None
if location.block_id:
    title = self._schema.get_block(location.block_id).get("title")
if not title:
    title = self._schema.get_section(location.section_id).get("title")
```
be preferable?

app/views/contexts/list_context.py
```diff
@@ -0,0 +1,84 @@
...
+
+class ListContext(Context):
+    def __call__(
+        self,
+        summary_definition,
+        for_list,
+        return_to=None,
+        editable=True,
+        edit_block=None,
+        remove_block=None,
```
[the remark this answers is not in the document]
JamesGardiner: Do you mean object attributes or do you mean override the init method to set these in the class?

templates/sectionsummary.html
JamesGardiner: It isn't required

app/views/contexts/section_summary_context.py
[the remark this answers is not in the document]
iwootten: I think so, it's more in line with the switches that fall immediately beneath it.

app/views/contexts/list_context.py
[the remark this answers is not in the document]
iwootten: I mean override the init and set them

app/routes/questionnaire.py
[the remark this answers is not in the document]
ajmaddaford: I prefer `QuestionnaireSummary`, it should be renamed anywhere it is referred to as final summary.

app/views/contexts/list_context.py
```diff
@@ -0,0 +1,84 @@
...
+        list_items = list(
+            self._build_list_items_context(
+                summary_definition,
+                for_list,
+                return_to=return_to,
+                edit_block_id=edit_block_id,
+                remove_block_id=remove_block_id,
+            )
+        )
+        return {"list": {"list_items": list_items, "editable": editable}}
```
ajmaddaford: It looks like `editable` is passed to call but not propagated on to `_build_list_items_context` so is just being used to set the `editable` property in the context. We should consider removing it from `__call__` and setting it in the context if either `edit_block_id` or `remove_block_id` are passed.

app/views/contexts/section_summary_context.py
```diff
@@ -0,0 +1,160 @@
...
+            self._list_store,
+            self._progress_store,
+            self._metadata,
+        )
+
+    def __call__(self, current_location):
+        try:
+            block = self._schema.get_block(current_location.block_id)
+            collapsible = block.get("collapsible", False)
+        except AttributeError:
```
ajmaddaford: What can raise an `AttributeError` here?

app/views/contexts/section_summary_context.py
ajmaddaford: It would be good to briefly describe here the outcome of what was discussed.

app/views/handlers/question.py
```diff
@@ -72,12 +72,17 @@ def get_context(self):
...
-                ),
-                "editable": False,
-            }
+            context.update(
+                list_collector_context(
+                    self.rendered_block["list_summary"]["summary"],
+                    for_list=self.rendered_block["list_summary"]["for_list"],
+                    editable=False,
+                    edit_block_id=self.rendered_block.get("edit_block", {}).get("id"),
+                    remove_block_id=self.rendered_block.get("remove_block", {}).get(
```
ajmaddaford: If editable is False, there is no need to pass the edit and remove block id.

templates/sectionsummary.html
```diff
@@ -1 +1,58 @@
...
+            </div>
+          </div>
+        {% endif %}
+      {% endfor %}
+  {% endif %}
+
+  {% if content.summary.is_view_submission_response_enabled %}
+    <div class="u-mt-m u-mb-s u-mb-m@s" data-qa="view-submission-text">
+      <div class="u-fs-r--b">{{ _("You will have the opportunity to view and print a copy of your answers after submitting this survey") }}</div>
+    </div>
```
ajmaddaford: `is_view_submission_response_enabled` won't be relevant for a section summary.

app/views/contexts/section_summary_context.py
JamesGardiner: A block that resolves to None

app/views/contexts/section_summary_context.py
```diff
@@ -0,0 +1,159 @@
...
+                yield self._list_summary_element(
+                    summary_element, current_location, section
+                )
+
+    def _list_summary_element(
+        self, summary: Mapping, current_location, section: Mapping
+    ) -> Iterator[Mapping]:
+        list_collector_block = self._schema.get_list_collectors_for_section(
+            section, for_list=summary["for_list"]
+        )[0]
```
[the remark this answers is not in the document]
JamesGardiner: Looks like there is an error in the schema - `visitor` in one `for_list` and `visitors` in another...

app/views/contexts/section_summary_context.py
ajmaddaford: I think this highlights another problem we will get to when we move section summaries out of blocks - where will `collapsible` be configured? I think it would make sense for it to be a property of the `summary` element on a `section`. If we do that we would need to move the list of custom summary elements into another property e.g `items`, something like:

```json
{
  "summary": {
    "collapsible": true,
    "items" : []
  }
}
```

As this is something that we are changing in this pull request, we should consider doing this now.

app/views/contexts/list_context.py
```diff
@@ -0,0 +1,90 @@
...
+            )
+
+            is_primary = list_item_id == primary_person
+
+            list_item_context = {
+                "item_title": self._get_item_title(list_item_id, is_primary),
+                "primary_person": is_primary,
+            }
+
+            if self._editable:
```
ajmaddaford: I don't think we need a `_editable` property. If the edit/remove block ids are passed, they can be added to the context.

app/views/contexts/list_context.py
```diff
@@ -0,0 +1,90 @@
...
+
+        self._context = {
+            "list": {
+                "list_items": list_items,
+                "editable": any([edit_block_id, remove_block_id]),
+            }
+        }
+
+    def get_context(self):
+        return self._context
```
ajmaddaford: This is inconsistent with `QuestionnaireSummaryContext` and `SectionSummaryContext` that both use the `__call__` approach. Shouldn't we decide a common approach and use it consistently across all contexts?

app/views/contexts/section_summary_context.py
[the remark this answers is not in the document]
ajmaddaford: As discussed, we should look at elevating the functionality we use for the add link to determine if the list collector is on the routing path.

app/views/contexts/list_context.py
JamesGardiner: What's the preference? I prefer call, but wanted to see what other people preferred?

app/views/contexts/section_summary_context.py
```diff
@@ -0,0 +1,150 @@
...
+            self._answer_store,
+            self._list_store,
+            self._progress_store,
+            self._metadata,
+            list_collector_block["summary"],
+            for_list=list_collector_block["for_list"],
+            return_to=current_location.block_id,
+            edit_block_id=list_collector_block["edit_block"]["id"],
+            remove_block_id=list_collector_block["remove_block"]["id"],
+        )
```
iwootten: The sooner we resolve dependencies ahead of their use, the better.

app/views/contexts/list_context.py
iwootten: Making non-ABC properties part of the call args would mean that we'd need to chain them down through any of the classes methods. I think I commented somewhere to this effect. Not sure of the best solution, maybe the the init initialises the class properties and the call then uses them to build/return the context?

ajmaddaford: I think moving the non-ABC properties into the init has muddied the water. The other contexts (`QuestionnaireSummaryContext` and `SectionSummaryContext`) don't override the init and pass the arguments that could vary between calls via the `__call__` method. I think we should go for consistency, even if it means passing the `__call__` args down through a few methods.

JamesGardiner: Ok - I'll revert the currently overridden init.

tests/functional/spec/features/section_summary/section_summary.spec.js
```diff
@@ -1,89 +1,88 @@
...
+      $(QuestionnaireSummaryPage.insuranceAddressAnswerEdit()).click();
       expect(browser.getUrl()).to.contain(InsuranceAddressPage.pageName);
-      $(InsuranceAddressPage.answer()).setValue('Test Address');
+      $(InsuranceAddressPage.answer()).setValue("Test Address");
       $(InsuranceAddressPage.submit()).click();
-      expect($(FinalSummaryPage.insuranceAddressAnswer()).getText()).to.contain('Test Address');
+      expect($(QuestionnaireSummaryPage.insuranceAddressAnswer()).getText()).to.contain("Test Address");
     });
   });
 });
```
iwootten: Any reason why all the quotes have been changed? This has been causing codacy to fail for a while now.

JamesGardiner: Probably my IDE settings - I'll revert these soon; I hadn't had time to look at the codacy failure yet.

app/views/contexts/section_summary_context.py
JamesGardiner: Yep agreed :+1:

JamesGardiner: One question - if a "summary" key is present in a section object, and that key does not contain an array of summary elements, should that then build a default summary view?

ajmaddaford: Yes, I think so.

tests/functional/generate_pages.py
```diff
@@ -392,11 +392,13 @@ def process_calculated_summary(answers, page_spec):
...
+def process_summary(schema_data, page_spec, collapsible):
     for section in schema_data["sections"]:
-        list_collector_blocks = QuestionnaireSchema.get_visible_list_blocks_for_section(
-            section
-        )
+        list_collector_blocks = [
+            summary_element
+            for summary_element in section.get("summary", [])
+            if summary_element["type"] == "List"
+        ]
```
ajmaddaford: These aren't list collector blocks, so something like `list_summaries` would be a better name.

app/views/contexts/section_summary_context.py
```diff
@@ -0,0 +1,154 @@
...
+                section, summary["for_list"]
+            )
+
+            if driving_question_block:
+                add_link = url_for(
+                    "questionnaire.block",
+                    block_id=driving_question_block["id"],
+                    return_to=current_location.block_id,
+                )
+        return add_link
```
ajmaddaford: I think this is existing behaviour, but this logic is broken. If there is no driving question the url for the list collector add block (that isn't on the routing path) is returned. We could return `None` and check for it in the template.

tests/app/questionnaire/conftest.py
```diff
@@ -569,6 +569,69 @@ def mock_relationship_collector_schema():
...
+    return {
+        "sections": [
+            {
+                "id": "section",
+                "summary": [
```
ajmaddaford: This needs to be updated with the `items` property.

app/views/contexts/section_summary_context.py
JamesGardiner: Yeah this emulates the current behaviour. 

I'll take a look at accessingvia the template.

JamesGardiner:
@LJBabbage - that looks like the wrong schemas are being used. All ListCollectorSummary blocks have been removed from the schemas. Pull the schema  repo, check out the custom-section-summaries branch, do `make build`, then symbolically link them to the root of the runner repo under a schemas directory :+1:

JamesGardiner:
@LJBabbage - I don't understand the need for the `type` key at this level of the document. It's implicitly a section summary by position in that it is under a `section` -> `summary` path. It wouldn't make sense to have any other type of summary at that point in the schema.

ajmaddaford:
@LJBabbage @JamesGardiner This card was one of the steps needed before removing `SectionSummary` as a block type. The `summary` property will be the only way to control section summary behaviour soon.

PR #61, opened 2020-02-23
Show section on hub and remove mid section summary code

LJBabbage:
To allow a section to be hidden on the hub and remove the code which travelled backwards through the routing path to find a section summary. The requirement for mid section summaries is not longer required.

### How to review 
As this PR makes up 1 part of 3 PRs the best way to test is by using the schema generated from move-relationships-into-own-section (eq-questionnaire-schema)  and the updated eq-questionnaire-validator branch (add_show_on_hub_to_section)


app/questionnaire/router.py
```diff
@@ -210,19 +206,23 @@ def is_path_complete(self, routing_path):
...
-    ) -> Location:
+    def get_section_return_location_when_section_complete(self, routing_path):
 
-        section_summary_location = self._get_location_of_section_summary(routing_path)
-        if section_summary_location:
-            return section_summary_location
+        last_block_id = routing_path[-1]
+        block = self._schema.get_block(last_block_id)
 
+        if block["type"] == "SectionSummary":
```
ajmaddaford: This also needs to check for `ListCollectorSummary` as currently the census schemas who lives here doesn't go back to the summary from the hub when it's complete.

app/questionnaire/router.py
```diff
@@ -210,19 +206,23 @@ def is_path_complete(self, routing_path):
             return True
         return False
 
-    def get_section_return_location_when_section_complete(
-        self, routing_path
-    ) -> Location:
+    def get_section_return_location_when_section_complete(self, routing_path):
```
JamesGardiner: Why  did the type hint get removed?

app/questionnaire/router.py
```diff
@@ -210,19 +206,23 @@ def is_path_complete(self, routing_path):
...
 
-        section_summary_location = self._get_location_of_section_summary(routing_path)
-        if section_summary_location:
-            return section_summary_location
+        last_block_id = routing_path[-1]
+        block = self._schema.get_block(last_block_id)
 
+        if block["type"] == "SectionSummary":
+            return Location(
+                block_id=routing_path[-1],
```
JamesGardiner:
```suggestion
                block_id=last_block_id,
```

tests/app/questionnaire/test_router.py
```diff
@@ -419,36 +419,7 @@ def test_get_first_incomplete_location_in_section(self):
...
-
-        self.assertEqual(
-            location_when_section_complete,
-            Location(section_id="household-section", block_id="household-summary"),
-        )
-
-    def get_section_return_location_when_section_complete_section_summary_last_block(
-        self
-    ):
+    def get_section_return_location_when_section_complete_section_summary(self):
```
JamesGardiner:
```suggestion
    def test_get_section_return_location_when_section_complete_section_summary(self):
```

app/questionnaire/router.py
```diff
@@ -214,15 +210,21 @@ def get_section_return_location_when_section_complete(
         self, routing_path
     ) -> Location:
 
-        section_summary_location = self._get_location_of_section_summary(routing_path)
-        if section_summary_location:
-            return section_summary_location
+        last_block_id = routing_path[-1]
+        block = self._schema.get_block(last_block_id)
```
ajmaddaford: Minor, but this would be slightly clearer when used below if it was named `last_block`.

tests/functional/spec/features/hub_and_spoke/hub_and_spoke.spec.js
```diff
@@ -20,6 +22,10 @@ describe('Feature: Hub and Spoke', function () {
     expect($(HubPage.summaryRowState(2)).getText()).to.contain('Not started');
   });
 
+  it('When a user views the Hub, any section with show_on_hub set to false should not appear', function() {
+    browser.openQuestionnaire(hub_and_spoke_schema);
+    expect($(HubPage.submit()).getText()).not.to.contain('Relationships');
```
MebinAbraham: I don't think this is valid. The submit button text would never contain `Relationships`. You should assert that a row for the section that has `show_on_hub` set to `false` is not being displayed on the hub.

PR #64, opened 2020-02-26
Update error behaviour and summaries

boxadesign:
Updated runner to use design system 16.1.0.

**Two breaking changes:**
**Summaries** - updated the component implementation. This is in readiness for the multiple section headings.

**Errors** - Added an id of `answer.id` to the error config in answer.html and updated the id's of each answer type to be unique to this. This is to allow a different behaviour when clicking error links and the way it focusses on inputs.

### How to review 
- Check that all existing summaries display correctly and don't have any missing content.
- Check validation errors for the following behaviour:
  - When an error link is clicked it jumps to the error panel (not input inside)
  - When error is anything other than checkbox or radio it should focus the field
  - When error is checkbox or radio it jumps to the panel and nothing is focussed
  - When JS is off it jumps to the panel and nothing is focussed


templates/partials/answers/checkbox.html
```diff
@@ -1,12 +1,13 @@
...
 {{ onsCheckboxes({
-  "id": answer["id"],
   "legend": question_title,
+  "id": answer.id,
   "legendClasses": "u-vh",
   "checkboxesLabel": '' if answer.label == '' else answer.label or _("Select all that apply"),
   "checkboxes": map_checkbox_config(form, answer),
   "mutuallyExclusive": mutuallyExclusive,
   "error": error
+  }
```
rmccar: Got an extra bracket here that is causing an error
```suggestion
```

boxadesign: OK thanks!

app/jinja_filters.py
```diff
@@ -408,10 +408,10 @@ def __init__(  # noqa: C901, R0912  pylint: disable=too-complex,too-many-branche
             and "label" in answer
             and answer["label"]
         ):
-            self.title = answer["label"]
+            self.itemTitle = answer["label"]
             self.titleAttributes = {"data-qa": answer["id"] + "-label"}
         else:
-            self.title = question["title"]
+            self.itemTitle = question["title"]
```
ajmaddaford: If `title` has been renamed `itemTitle`, shouldn't `titleAttributes` be renamed `itemTitleAttributes`?

templates/question.html
```diff
@@ -22,8 +22,9 @@
       <p>{{ ngettext('This <strong>must be corrected</strong> to continue', 'These <strong>must be corrected</strong> to continue', form.mapped_errors | length) }}.</p>
       <ul class="list list--bare">
         {% for answer_id, error in form.mapped_errors %}
+          {% set hrefValue = answer_id + '-error' if answer_id != 'date-range-question' else answer_id %}
```
ajmaddaford: Using a fixed answer id of `date-range-question` won't work as the answer id can be anything a survey author wants it to be. I'm not sure it matters though as it looks like the error panel for date ranges have an id of `<answer_id>-error` anyway?

boxadesign: OK. In this scenario the `-error` suffix isn't required....

Second date fieldset contains a date ahead of the first fieldset.

No error panel is rendered around the inputs so the `-error` id isn't generated for the panel. 

The href for the error link uses the question id and not the answer id (already doing that and only for this scenario) so that it jumps to the question block and the js finds the input to focus.

If there's some other value we could test for instead of `date-range-question` then let me know and I will update.

The main thing is that we don't want the `-error` on the href for the date range scenario.

ajmaddaford: Ah, I hadn't noticed that. I've pushed a change to move the error id generation into the code, where we already know about the distinction between question and answer errors.

boxadesign:
> @boxadesign In most cases when you click on the error link, you cannot see the question in the viewport, is this a known issues?

This is the intended behaviour. The browser should jump to the top of the error panel.

templates/partials/summary/collapsible-summary.html
```diff
@@ -7,7 +7,7 @@
   {%- if group["blocks"] | length -%}
     {%- if group.title -%}
       {%- set item = {
-        "title": group.title,
+        "groupTitle": group.title,
```
MebinAbraham: This doesn't work for me. When I change `groupTitle` to `title` then it is displayed on the summary.

templates/partials/summary/collapsible-summary.html
```diff
@@ -24,7 +24,15 @@
       {%- set summary %}
         {{ onsSummary({
           "classes": "summary--no-bottom-border",
-          "rows": map_summary_item_config(group, content.summary.summary_type, content.summary.answers_are_editable, _("No answer provided"), _("Change"), _("Change your answer for:"), content.summary.calculated_question)
+          "summaries": [
+            {
+              "groups": [
+                {
+                  "rows": map_summary_item_config(group, content.summary.summary_type, content.summary.answers_are_editable, _("No answer provided"), _("Change"), _("Change your answer for:"), content.summary.calculated_question)
+                }
```
MebinAbraham: Can we make this a bit nice on the 👀?
```jinja
                {
                  "rows": map_summary_item_config(
                            group,
                            content.summary.summary_type,
                            content.summary.answers_are_editable,
                            _("No answer provided"),
                            _("Change"),
                            _("Change your answer for:"),
                            content.summary.calculated_question
                        )
                }
```

templates/partials/summary/collapsible-summary.html
MebinAbraham: I think we should also update one of our tests to ensure that the group title is displayed as expected, as this doesn't seem to be tested currently.

templates/partials/summary/summary.html
```diff
@@ -3,11 +3,23 @@
...
-      "classes": "u-mt-m" if loop.index > 1 else ""
-    }) }}
+    {% do summaryGroups.append (
+      {
+        "groups": [
+          {
+            "groupTitle": '<div id="' + group.id +'">' + group.title + '</div>' if group.title else None,
+            "rows": map_summary_item_config(group, content.summary.summary_type, content.summary.answers_are_editable, _("No answer provided"), _("Change"), _("Change your answer for:"), content.summary.calculated_question),
+            "classes": "u-mt-m" if loop.index > 1 else ""
+          }
```
MebinAbraham: Can we format this please.
```jinja
          {
            "groupTitle": '<div id="' + group.id +'">' + group.title + '</div>' if group.title else None,
            "rows": map_summary_item_config(
                        group,
                        content.summary.summary_type,
                        content.summary.answers_are_editable,
                        _("No answer provided"),
                        _("Change"),
                        _("Change your answer for:"),
                        content.summary.calculated_question
                    ),
            "classes": "u-mt-m" if loop.index > 1 else ""
          }
```

boxadesign:
@MebinAbraham All suggestions have been implemented now

MebinAbraham:
> Apart from those minor comment, it seems to work as expected 👍

I think we should update our test to cover the issue with group title:
https://github.com/ONSdigital/eq-questionnaire-runner/pull/64#discussion_r389599101

PR #66, opened 2020-02-27
Use appropriate width class on number type fields

ajmaddaford:
### Context
The design system allows customisation of an input field to set a display width. This is not currently implemented in runner and should be.

### How to review 
Use the test schemas for numbers, currencies, units and percentages to test the changes.


app/jinja_filters.py
```diff
@@ -193,6 +193,32 @@ def should_wrap_with_fieldset_processor():
...
 
+@blueprint.app_template_filter()
+def get_width_class_for_number(answer):
+    allowable_widths = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 20]
+
+    min_value = answer.get("min_value", {}).get("value", 0)
+    max_value = answer.get("max_value", {}).get("value")
+    if not max_value:
+        return None
+
```
MebinAbraham: Petty comment, could/should be initialised when needed.
```suggestion
    max_value = answer.get("max_value", {}).get("value")
    if not max_value:
        return None

    allowable_widths = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 20]
    min_value = answer.get("min_value", {}).get("value", 0)
```

app/jinja_filters.py
```diff
@@ -193,6 +193,32 @@ def should_wrap_with_fieldset_processor():
...
+    min_value_width = len(str(min_value))
+    max_value_width = len(str(max_value))
+
+    width = min_value_width if min_value_width > max_value_width else max_value_width
+
+    width += answer.get("decimal_places", 0)
+
+    for allowable_width in allowable_widths:
+        if width <= allowable_width:
+            return "input--w-" + str(allowable_width)
```
MebinAbraham:
```suggestion
            return f"input--w-{allowable_width}"
```

app/jinja_filters.py
```diff
@@ -193,6 +193,32 @@ def should_wrap_with_fieldset_processor():
...
 
 
+@blueprint.app_template_filter()
+def get_width_class_for_number(answer):
+    allowable_widths = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 20]
+
+    min_value = answer.get("min_value", {}).get("value", 0)
+    max_value = answer.get("max_value", {}).get("value")
+    if not max_value:
+        return None
```
MebinAbraham: Should this not default to the length of our system max default: `MAX_NUMBER = 9999999999`

ajmaddaford: I did think about this, but was reluctant as it seemed like a business survey limitation. Having said that, it is a limit elsewhere so should be used here. I've moved `MAX_NUMBER` into `app.settings` so it can be accessed from both places.

app/jinja_filters.py
ajmaddaford: This isn't relevant now that there is no early return.

PR #70, opened 2020-03-05
Add driving checkbox question

LJBabbage:
This is a spike for a driving checkbox question and **should not be merged** 

### How to review 
Check the new schema matches the overflow diagram and all the scenarios work. The text is not exactly the same (although it's 99% right) and doesn't need to be reviewed, it's more the journeys/scenarios that need testing.
 

**Known issues/observations**

1. To allow the 1st variant to go to the "Who else lives at .. " (add person) the empty list check in question had to be removed as naturally it would have come through the primary list collector and have an entry, which has led to a couple of tests breaking.

2. At present I have had to use 2 list collectors to cover all the variants, which also affects any summary, if there are any better ideas please comment

3. ID's need to be looked at for multiple list collectors and will fail validator, but this felt out of scope for this PR and I ran out of time

4. "Yes, I need to add {second, third} person" has not been implemented.


JamesGardiner:
FYI - Github allows draft PRs that are ideal for these kind of spikes

PR #71, opened 2020-03-09
New transforms to add numbers & format ordinals

petechd:
This adds new transform required by "format_ordinal" to work with cardinal numbers in future. More details are on [this Trello card](https://trello.com/c/PttjF0kg/3537-b-ordinals-eg-third-in-schema-text-ml). New definitions were added in this [validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/16).

### How to review 
There are unit and integration tests to check modified schema. You can check how the transform works by going through 'test_placeholder_transform.json', where new question and interstitial page were added.


petechd:
[[B] Ordinals (e.g. third) in schema text (M)](https://trello.com/c/PttjF0kg/3537-b-ordinals-eg-third-in-schema-text-m)

app/questionnaire/placeholder_transforms.py
```diff
@@ -125,6 +125,31 @@ def parse_date(date):
...
+        elif number_to_format.endswith("2"):
+            number_to_format += "nd"
+        elif number_to_format.endswith("3"):
+            number_to_format += "rd"
+        else:
+            number_to_format += "th"
+        if determiner == "a_or_an":
+            if number_to_format.startswith("8") or number_to_format == "18th":
+                return "an " + str(number_to_format)
+            return "a " + str(number_to_format)
```
MebinAbraham: `number_to_format` is already a string. It would be better to assign it to a var instead of mutating it. We should use f-Strings here.

This can be starts with '8' or starts with '18'

petechd: Everything done except startswith() for "8" but not for "18", not everything that starts with "18" will get an "an"😂

app/questionnaire/placeholder_transforms.py
```diff
@@ -125,6 +125,36 @@ def parse_date(date):
...
 
+    @staticmethod
+    def add(lhs, rhs):
+        if isinstance(lhs, str):
+            return int(lhs) + rhs
+        return lhs + rhs
+
+    @staticmethod
+    def format_ordinal(number_to_format, determiner=None):
+        number_to_format_str = str(number_to_format)
```
LJBabbage: A cleaner and quicker (timeit) version you could use is 

suffix = {1: "st", 2: "nd", 3: "rd"}.get(
number_to_format % 10 * (number_to_format % 100 not in ["11", "12", "13"]), "th")

app/questionnaire/placeholder_transforms.py
```diff
@@ -125,6 +125,36 @@ def parse_date(date):
...
+            suffix = "th"
+        elif number_to_format_str.endswith("1"):
+            suffix = "st"
+        elif number_to_format_str.endswith("2"):
+            suffix = "nd"
+        elif number_to_format_str.endswith("3"):
+            suffix = "rd"
+        else:
+            suffix = "th"
+        if determiner == "a_or_an":
```
LJBabbage: I probably wouldn't do both in the same method, one method to do the prefix one for the suffix, making it more reusable, we might use a_or_an somewhere else

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -118,6 +118,23 @@ def test_concatenate_list(self):
             == "Milk, Eggs, Flour, Water"
         )
 
+    def test_add(self):
```
LJBabbage: As above, not sure why this is here

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -118,6 +118,23 @@ def test_concatenate_list(self):
...
+        assert self.transforms.add(1, 2) == 3
+        assert self.transforms.add("1", 2) == 3
+
+    def test_format_ordinal(self):
+        assert self.transforms.format_ordinal(1, "a_or_an") == "a 1st"
```
LJBabbage: As above, maybe split into 2 prefix, suffix

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -118,6 +118,23 @@ def test_concatenate_list(self):
...
+        assert self.transforms.format_ordinal(2, "a_or_an") == "a 2nd"
```
LJBabbage: Are we interested in just these ordinals, should we be testing more, and looping, not sure on this one.

tests/integration/questionnaire/test_questionnaire_placeholders.py
```diff
@@ -30,3 +30,31 @@ def test_title_placeholders_rendered_in_summary(self):
         self.assertInUrl("/summary/")
         self.assertInBody("What is Kevin Bacon’s date of birth?")
         self.assertInBody("68 Abingdon Road, Goathill")
+
+    def test_placeholders_rendered_in_pages(self):
```
LJBabbage: A lot going on in this test, its been added to test the ordinals, but I wouldn't have known it out of context

app/questionnaire/placeholder_transforms.py
```diff
@@ -125,6 +125,36 @@ def parse_date(date):
                 date, PlaceholderTransforms.input_date_format_month_year_only
             ).replace(tzinfo=tzutc())
 
+    @staticmethod
+    def add(lhs, rhs):
```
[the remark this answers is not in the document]
petechd: It's in accordance with the structure proposed on [this card](https://trello.com/c/PttjF0kg/3537-b-ordinals-eg-third-in-schema-text-m). I spoke with other devs and this transform should allow to add any numbers passed into it , not only answer values.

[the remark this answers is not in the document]
LJBabbage: Ah, I completely missed that, thanks

app/questionnaire/placeholder_transforms.py
```diff
@@ -125,6 +125,36 @@ def parse_date(date):
...
+        if isinstance(lhs, str):
```
LJBabbage: Why only the cast on the first value? surely they both could be strings, or even something else

app/questionnaire/placeholder_transforms.py
```diff
@@ -125,6 +125,36 @@ def parse_date(date):
...
+        if number_to_format_str.endswith(("11", "12", "13")):
+            suffix = "th"
+        elif number_to_format_str.endswith("1"):
+            suffix = "st"
+        elif number_to_format_str.endswith("2"):
+            suffix = "nd"
+        elif number_to_format_str.endswith("3"):
+            suffix = "rd"
+        else:
+            suffix = "th"
```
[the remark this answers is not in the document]
petechd: I opted for two liner as it's more readable for me, thanks👍

app/questionnaire/placeholder_transforms.py
petechd: this is real head-scratcher as it failed unit tests on 11, 12, 13(spitting out 11st, 12nd, 13rd) and got linted in strange manner with come extra comas, I opted for Andrew's solution in the end. Thanks👍

app/questionnaire/placeholder_transforms.py
petechd: "determiner" is part of the same transform as proposed on [this card](https://trello.com/c/PttjF0kg/3537-b-ordinals-eg-third-in-schema-text-m). Would have to be a completely different transform?

app/questionnaire/placeholder_transforms.py
petechd: I changed it so it checks both now, thanks!

app/questionnaire/placeholder_transforms.py
```diff
@@ -125,6 +125,31 @@ def parse_date(date):
...
+
+        if 11 <= number_to_format % 100 <= 13:
+            suffix = "th"
+        else:
+            suffix = {1: "st", 2: "nd", 3: "rd"}.get(number_to_format % 10, "th")
+
+        if determiner == "a_or_an":
+            a_or_an = (
+                "an"
+                if number_to_format_str.startswith("8") or number_to_format in [11, 18]
```
ajmaddaford: The `number_to_format_str ` variable could be removed here and just use:
```python
if str(number_to_format).startswith("8") or number_to_format in [11, 18]
```

app/questionnaire/placeholder_transforms.py
```diff
@@ -125,6 +125,31 @@ def parse_date(date):
                 date, PlaceholderTransforms.input_date_format_month_year_only
             ).replace(tzinfo=tzutc())
 
+    @staticmethod
+    def add(lhs, rhs):
+        if any([isinstance(lhs, str), isinstance(rhs, str)]):
+            return int(lhs) + int(rhs)
```
ajmaddaford: This shouldn't be necessary. If the value is set in the schema it will be a number, and the answer store should be returning a number.

I've debugged this and I think it's because the answer store returns escaped answers wrapped in a `Markup` class. This is a bug in the placeholder rendering - the values should be escaped at the point they are interpolated into the string, rather than on the way out of the answer store. I think this bug will take some effort to investigate and fix, so I'm happy for it to be raised as an issue and fixed outside of this pull request.

app/questionnaire/placeholder_transforms.py
petechd: Done, thanks👍

tests/app/questionnaire/test_placeholder_transforms.py
petechd: From what I know, we're only interested in ordinals, hence the transform name. I can add some more assertions but wouldn't looping through a block be a bit excessive?

tests/app/questionnaire/test_placeholder_transforms.py
petechd: I moved it to "prefix" and "suffix" tests now, thanks👍

tests/app/questionnaire/test_placeholder_transforms.py
petechd: This test the first transform ("add") which then passes added numbers to "format_ordinal". It is a separate transform but in schemas we use it in tandem with ordinals transform.

app/questionnaire/placeholder_transforms.py
petechd: I created a separate card for this issue, do we keep this transform as it is?

tests/integration/questionnaire/test_questionnaire_placeholders.py
```diff
@@ -42,12 +42,6 @@ def test_placeholders_rendered_in_pages(self):
...
 
-        self.post({"additional-family-members-answer": 12})
-
-        self.assertInBody("Your family size is <em>13</em>")
-
-        self.post()
-
         self.post({"total-items-answer": 2})
 
         self.assertInBody("Do you want to add <em>a 3rd</em> item?")
```
ajmaddaford: This test should also test that the placeholder '3rd' is rendered on the summary.

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -118,6 +118,25 @@ def test_concatenate_list(self):
             == "Milk, Eggs, Flour, Water"
         )
 
+    def test_add(self):
+        assert self.transforms.add(1, 2) == 3
+        assert self.transforms.add("1", 2) == 3
+
+    def test_format_ordinal_prefix(self):
```
ajmaddaford: The test method should be named to match the behaviour, something like `test_format_ordinal_with_determiner`.

tests/integration/questionnaire/test_questionnaire_placeholders.py
petechd: I've added it, thanks👍

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -118,6 +118,25 @@ def test_concatenate_list(self):
...
+        assert self.transforms.format_ordinal(1, "a_or_an") == "a 1st"
+        assert self.transforms.format_ordinal(2, "a_or_an") == "a 2nd"
+        assert self.transforms.format_ordinal(3, "a_or_an") == "a 3rd"
+        assert self.transforms.format_ordinal(4, "a_or_an") == "a 4th"
+        assert self.transforms.format_ordinal(8, "a_or_an") == "an 8th"
+        assert self.transforms.format_ordinal(11, "a_or_an") == "an 11th"
```
ajmaddaford: There should be a test for 18th.

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -118,6 +118,25 @@ def test_concatenate_list(self):
...
+
+    def test_format_ordinal_suffix(self):
+        assert self.transforms.format_ordinal(1, None) == "1st"
+        assert self.transforms.format_ordinal(2, None) == "2nd"
+        assert self.transforms.format_ordinal(3, None) == "3rd"
+        assert self.transforms.format_ordinal(4, None) == "4th"
```
ajmaddaford: Given the behaviour, I think there should be tests for:
- 11th, 12th and 13th
- 21st, 22nd, 23rd
- 111th, 112th and 113th

tests/app/questionnaire/test_placeholder_transforms.py
petechd: renamed to "test_format_ordinal_with_determiner" and "test_format_ordinal_without_determiner". 👍

tests/app/questionnaire/test_placeholder_transforms.py
petechd: Added, thanks!👍

tests/app/questionnaire/test_placeholder_transforms.py
petechd: Its 1, 2, 3, 4, 8, 11, 12, 13, 18, 21, 22, 23, 111, 112, 113 now. Was I supposed to keep first four in there?

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -118,6 +118,35 @@ def test_concatenate_list(self):
...
+        assert self.transforms.format_ordinal(18, "a_or_an") == "an 18th"
+        assert self.transforms.format_ordinal(21, "a_or_an") == "a 21st"
+        assert self.transforms.format_ordinal(22, "a_or_an") == "a 22nd"
+        assert self.transforms.format_ordinal(23, "a_or_an") == "a 23rd"
+        assert self.transforms.format_ordinal(111, "a_or_an") == "a 111th"
+        assert self.transforms.format_ordinal(112, "a_or_an") == "a 112th"
+        assert self.transforms.format_ordinal(113, "a_or_an") == "a 113th"
+
+    def test_format_ordinal_without_determiner(self):
+        assert self.transforms.format_ordinal(1, None) == "1st"
```
LJBabbage: You don't need the Nones as you have a default on the method

petechd: Fair point, done, thanks👍

PR #74, opened 2020-03-11
Remove autocomplete when using lookups

petechd:
This removes autocomplete from input field using lookup component. Previously there was a bug in Design System that caused autocomplete to be set to "null".

### DIFF
--- a/templates/partials/answers/textfield.html
+++ b/templates/partials/answers/textfield.html
@@ -32,6 +32,9 @@
     "ariaLimitedResults": "Results have been limited to 10 suggestions. Type more characters to improve your search.",
     "typeaheadData": answer.suggestions_url
   }) %}
+  {% set config = config | setAttribute("other", {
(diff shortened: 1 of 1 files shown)


templates/partials/answers/textfield.html
```diff
@@ -32,6 +32,7 @@
     "ariaLimitedResults": "Results have been limited to 10 suggestions. Type more characters to improve your search.",
     "typeaheadData": answer.suggestions_url
   }) %}
+  {% set config = config | setAttribute("autocomplete", "off") %}
```
MebinAbraham: Can these no just be added to in the initial config object?

iwootten: No, by default it should be on for a textfield and only switched off when suggestions are used.

MebinAbraham: Ah didn't see the `if`

PR #76, opened 2020-03-16
Remove the version number from deployment waits

iwootten:

The wait task currently explicitly checks the commit used as the version number from runner's .git/HEAD. For manual-tagged deployments, this will be a tag, not a commit hash. We can safely assume the version we ask to be deployed is deployed, so we can remove this check.

### How to review 
Fly the task with a reachable/unreachable endpoint and check the response is as expected.

### DIFF
--- a/ci/wait.yaml
+++ b/ci/wait.yaml
@@ -10,4 +10,4 @@ run:
   args:
     - -exc
     - |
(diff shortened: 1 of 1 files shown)


iwootten:
[Update eq-manual-tagged pipelines to use ci tasks (L)](https://trello.com/c/OAqbBjJi/3657-update-eq-manual-tagged-pipelines-to-use-ci-tasks-l)

PR #77, opened 2020-03-16
Update schema release to v0.0.8

berroar:
Updates runner to use version 0.0.8 of the eq questionnaire schemas which includes the latest changes to the HH and HI schemas.


berroar:
[[A] Wales Schema change (S)](https://trello.com/c/qSsElgFv/3731-a-wales-schema-change-s)

PR #81, opened 2020-03-17
Add language property to schemas

MebinAbraham:
Added a language property to the schema, which is now enforced by: https://github.com/ONSdigital/eq-questionnaire-validator/pull/19

### How to review 
- Does validation pass?


scripts/run_validator.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
-branch=latest
+branch=add-language-property
```
JamesGardiner: revert before merge

PR #82, opened 2020-03-18
Apply appropriate display width for radio and checkbox detail answers

berroar:
This PR has been raised to apply the appropriate display width for number input fields on radio and checkbox detail answers.

### How to review 
Use the `test_checkbox_numeric_detail_answers` and `test_radio_numeric_detail_answers` schemas to check the width of the input fields on the detail answer options.


berroar:
[[F] Update number field to set appropriate width on Text Fields within Radio options - Number of Vehicles (S)](https://trello.com/c/YRwWPUuC/3742-f-update-number-field-to-set-appropriate-width-on-text-fields-within-radio-options-number-of-vehicles-s)

app/jinja_filters.py
```diff
@@ -261,9 +268,18 @@ def __init__(self, option, index, form, answer):
...
+                detail_answer_option = answer_option["detail_answer"]
+
             detail_answer = form["fields"][option.detail_answer_id]
             answer_visible = answer_option["detail_answer"].get("visible", False)
-            self.other = OtherConfig(detail_answer, answer_visible)
+            self.other = OtherConfig(
+                detail_answer=detail_answer,
+                answer_visible=answer_visible,
+                detail_answer_option=detail_answer_option,
+            )
```
ajmaddaford: It would be cleaner to handle it in the same way as `answer_visible`, something like:

```python
answer_type = answer_option["detail_answer"].get("type", None)
self.other = OtherConfig(
    detail_answer=detail_answer,
    answer_visible=answer_visible,
    answer_type=answer_type,
)
```

berroar: I passed `detail_answer_option` rather than just the `answer_type` as I thought the `get_width_class_for_number` method which is being called in OtherConfig needed the whole answer option. Is there another way around this?

ajmaddaford: Ah, I hadn't spotted that. In that case it would be cleaner to move the visible check into  `OtherConfig` and make the naming clearer between the answer field and the answer schema. Something like:

```python
detail_answer_field = form["fields"][option.detail_answer_id]
detail_answer_schema = answer_option["detail_answer"]
self.other = OtherConfig(
    detail_answer_field, 
    detail_answer_schema
)
```

PR #84, opened 2020-03-25
fix for monthyear piping

martyncolmer:
Currently when piping in a month year answer eq runner will error.
This is because the date pased to the format date method in not valid and missing the day.
The fix uses the existing parse date method to correct the date before formatting.


JamesGardiner:
@martyncolmer can this be closed?

pricem14pc:
@martyncolmer I'm going to close this for the time being as it feels like a feature change request rather than a fix. I'm not familiar with the history but the doc states date can be of type YYYY-MM-DD or YYYY-MM. (YYYY doesn't appear to be supported). As such, the unit test for the YYYY-MM as passing as expected. Happy to discuss this for progressing as a change.

PR #89, opened 2020-03-31
Configureable textarea field sizes

berroar:
This PR adds support for configurable `TextArea` field sizes to be set using the `rows` variable in schemas.

### How to review 
Use the `test_textarea` schema to test this change.


scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=add-rows-to-textarea
```
berroar: Revert before merge

berroar:
[[F] Make the Text Area height configurable in eQ Runner (S)](https://trello.com/c/dG2kAjIV/3777-f-make-the-text-area-height-configurable-in-eq-runner-s)

tests/app/forms/field_handlers/test_text_area_handler.py
```diff
@@ -73,3 +73,24 @@ def test_get_length_validator_with_max_length_override():
     validator = text_area_handler.get_length_validator()
 
     assert validator.max == 30
+
+
+def test_get_textarea_row_length():
```
ajmaddaford: - Seems odd to use length, `test_get_textarea_rows` would be better.
- We should have another test to test the default rows i.e. no rows attribute in `answer`.

tests/app/forms/field_handlers/test_text_area_handler.py
```diff
@@ -73,3 +73,24 @@ def test_get_length_validator_with_max_length_override():
...
+        "guidance": "",
+        "id": "answer",
+        "rows": 3,
+        "label": "Enter your comments",
+        "mandatory": False,
+        "q_code": "0",
+        "type": "TextArea",
+    }
+
+    text_area_handler = TextAreaHandler(answer, {}, AnswerStore(), {})
```
ajmaddaford: The last three args are defaulted if not passed, so this can be simplified to `TextAreaHandler(answer)`.

tests/app/forms/field_handlers/test_text_area_handler.py
```diff
@@ -73,3 +73,24 @@ def test_get_length_validator_with_max_length_override():
...
+
+def test_get_textarea_row_length():
+    answer = {
+        "guidance": "",
+        "id": "answer",
+        "rows": 3,
+        "label": "Enter your comments",
+        "mandatory": False,
+        "q_code": "0",
+        "type": "TextArea",
```
ajmaddaford: We shouldn't use answer properties that aren't required here - `guidance` and `q_code` should be removed.

app/forms/fields/max_text_area_field.py
```diff
@@ -2,6 +2,7 @@
 
 
 class MaxTextAreaField(TextAreaField):
-    def __init__(self, label="", validators=None, maxlength=10000, **kwargs):
+    def __init__(self, label="", validators=None, rows=None, maxlength=10000, **kwargs):
```
ajmaddaford: As maxlength is defaulted in `TextAreaHandler` we should remove the default here so it's consistent with `rows`.

tests/app/forms/field_handlers/test_text_area_handler.py
```diff
@@ -73,3 +73,40 @@ def test_get_length_validator_with_max_length_override():
     validator = text_area_handler.get_length_validator()
 
     assert validator.max == 30
+
+
+def test_get_textarea_rows_with_default():
```
JamesGardiner, line 78: ```def test_get_text_area_rows_with_default():```

tests/app/forms/field_handlers/test_text_area_handler.py
```diff
@@ -73,3 +73,40 @@ def test_get_length_validator_with_max_length_override():
...
+
+    class TestForm(Form):
+        test_field = text_area_handler.get_field()
+
+    form = TestForm()
+
+    assert form.test_field.rows == 8
+
+
+def test_get_textarea_rows():
```
JamesGardiner, line 96: ```def test_get_text_area_rows():```

PR #91, opened 2020-04-03
Question concatenation for summary textfields

LJBabbage:
Describe what you have changed and why, link to other PRs or Issues as appropriate. Validator has been updated here to reflect the change https://github.com/ONSdigital/eq-questionnaire-validator/pull/25, naturally validator will fail until it gets merged. To check it change the branch in runner. The eagled eyed will see I have had to cast to string during the assignment operator (+=), without it the line break gets escaped and then won't render properly on the browser. I haven't found a reason for this, but I am open to suggestions

### How to review 
To allow Textfields answers on the summary to be concatenated using NewLine and Space. 
### Checklist
Visually check the summary on runner and review the code as normal


app/views/contexts/summary/question.py
```diff
@@ -43,9 +48,31 @@ def _build_answers(self, answer_store, question_schema):
...
+            if answer_value:
+                if answer_separator and concatenated_answer:
+                    answer_value = f"{answer_separator}{answer_value}"
+                concatenated_answer += str(answer_value)
+
+        return [
+            {
+                "id": f"{self.id}-concatenated-answer",
+                "value": concatenated_answer,
+                "type": "textfield",
```
ajmaddaford, line 72: - This should return a just the answer object, if it needs to be wrapped in a list then that should be done by the caller.
- I don't think `type` is needed here and should be removed.

test_schemas/en/test_summary.json
```diff
@@ -32,6 +32,90 @@
             "groups": [
                 {
                     "blocks": [
+                        {
```
ajmaddaford, line 35: We should consider a separate test schema for custom question summaries.

tests/app/views/contexts/summary/test_question.py
```diff
@@ -70,6 +70,103 @@ def test_create_question_with_answer_label_when_empty_title(self):
...
+            "id": "last-name",
+            "label": "Last name",
+            "mandatory": True,
+            "type": "TextField",
+        }
+
+        question_schema = {
+            "id": "question_id",
+            "title": "question_title",
+            "type": "GENERAL",
```
ajmaddaford, line 101: This should be `General`

tests/app/views/contexts/summary/test_question.py
```diff
@@ -70,6 +70,103 @@ def test_create_question_with_answer_label_when_empty_title(self):
...
+            "id": "postcode",
+            "label": "Postcode",
+            "mandatory": False,
+            "type": "TextField",
+        }
+
+        question_schema = {
+            "id": "question_id",
+            "title": "question_title",
+            "type": "GENERAL",
```
ajmaddaford, line 156: This should be `General`

app/views/contexts/summary/question.py
```diff
@@ -43,9 +53,21 @@ def _build_answers(self, answer_store, question_schema):
             if exclusive_option:
                 return summary_answers[-1:]
             return summary_answers[:-1]
-
         return summary_answers
 
+    def _concatenate_textfield_answers(self, answer_store, concatenation_type):
+
+        answer_separators = {"NewLine": "<br/>", "Space": " "}
```
MebinAbraham, line 60: We should also change this to `Newline`.

app/views/contexts/summary/question.py
```diff
@@ -43,9 +48,31 @@ def _build_answers(self, answer_store, question_schema):
             if exclusive_option:
                 return summary_answers[-1:]
             return summary_answers[:-1]
-
         return summary_answers
 
+    def _concatenate_textfield_answers(self, answer_store, concatenation_type):
+
+        concatenated_answer = ""
+        answer_separators = {"NewLine": "<br/>", "Space": " "}
```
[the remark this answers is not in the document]
iwootten: @MebinAbraham This doesn't mean "<br />" is invalid in html. Generally good practice to close any tag.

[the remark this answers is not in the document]
MebinAbraham: It's not invalid, the browser will convert it to `<br>` hence why it should be like that in the first place.

PR #92, opened 2020-04-06
Support for question variants based on first item in a list

berroar:
This PR adds support for question variants that determine whether the section being completed is for the first item in a list.

When rules in a schema now support the value `first` within the the `id_selector` field, for example:

```
{ "list": "people", 
    "id_selector":  "first", 
    "condition": "equals", 
    "comparison": {
         "source": "location", 
         "id": "list_item_id"},
}
```

### How to review 
Use the `test_variants_first_item_in_list` schema to test this change.

### DIFF
--- a/app/data_model/list_store.py
+++ b/app/data_model/list_store.py
@@ -45,8 +45,7 @@ class ListModel:
         try:
             return self.items[0]
         except IndexError:
-            logger.error("unable to access first item in list, list is empty")
-            raise IndexError
+            raise IndexError("unable to access first item in list, list is empty")
 
 
 class ListStore:
--- a/app/data_model/list_store.py
+++ b/app/data_model/list_store.py
@@ -48,7 +48,6 @@ class ListModel:
             logger.error("unable to access first item in list, list is empty")
 
 
-
 class ListStore:
     """ Storage for lists.
 
--- a/app/data_model/list_store.py
+++ b/app/data_model/list_store.py
@@ -45,7 +45,9 @@ class ListModel:
         try:
             return self.items[0]
         except IndexError:
-            raise IndexError(f"unable to access first item in list, list '{self.name}' is empty")
+            raise IndexError(
+                f"unable to access first item in list, list '{self.name}' is empty"
+            )
 
 
 class ListStore:
--- a/app/data_model/list_store.py
+++ b/app/data_model/list_store.py
@@ -174,3 +174,7 @@ class ListStore:
             return cls()
 
         return cls(existing_items=serialised)
+
+    def get_first_item_from_list(self, list_name):
+        if list_name in self._lists:
+            return self._lists[list_name].items[0]
--- a/app/data_model/list_store.py
+++ b/app/data_model/list_store.py
@@ -42,7 +42,11 @@ class ListModel:
 
     @property
     def first(self):
-        return self.items[0]
+        try:
+            return self.items[0]
+        except IndexError:
+            logger.error("unable to access first item in list, list is empty")
+
 
 
 class ListStore:
--- a/app/data_model/list_store.py
+++ b/app/data_model/list_store.py
@@ -40,6 +40,10 @@ class ListModel:
     def __repr__(self):
         return f"<ListModel name={self.name} items={self.items}, primary_person={self.primary_person}>"
 
+    @property
+    def first(self):
+        return self.items[0]
+
 
 class ListStore:
     """ Storage for lists.
@@ -174,7 +178,3 @@ class ListStore:
             return cls()
 
         return cls(existing_items=serialised)
-
-    def get_first_item_from_list(self, list_name):
-        if list_name in self._lists:
-            return self._lists[list_name].items[0]
--- a/app/data_model/list_store.py
+++ b/app/data_model/list_store.py
@@ -45,7 +45,7 @@ class ListModel:
         try:
             return self.items[0]
         except IndexError:
-            raise IndexError("unable to access first item in list, list is empty")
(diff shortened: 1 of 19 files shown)


berroar:
[[K] When rule logic to determine if currently first list item or not (M)](https://trello.com/c/Wk9jyUSg/3788-k-when-rule-logic-to-determine-if-currently-first-list-item-or-not-m)

app/data_model/list_store.py
```diff
@@ -174,3 +174,7 @@ def deserialise(cls, serialised: list):
             return cls()
 
         return cls(existing_items=serialised)
+
+    def get_first_item_from_list(self, list_name):
+        if list_name in self._lists:
+            return self._lists[list_name].items[0]
```
JamesGardiner, line 180: Can an empty list be accessed here? If so may need to catch any IndexErrors

app/data_model/list_store.py
```diff
@@ -174,3 +174,7 @@ def deserialise(cls, serialised: list):
             return cls()
 
         return cls(existing_items=serialised)
+
+    def get_first_item_from_list(self, list_name):
```
JamesGardiner, line 178: This could be made an attribute of the ListModel, so that the `first` id selector always returns the element at index 0 of it's `items` attribute.

You wouldn't need to change rules.py then

app/questionnaire/rules.py
```diff
@@ -256,7 +256,10 @@ def _get_when_rule_value(
     elif "meta" in when_rule:
         value = get_metadata_value(metadata, when_rule["meta"])
     elif "id_selector" in when_rule:
-        value = getattr(list_store.get(when_rule["list"]), when_rule["id_selector"])
+        if when_rule["id_selector"] == "first":
+            value = list_store.get_first_item_from_list(when_rule["list"])
+        else:
+            value = getattr(list_store.get(when_rule["list"]), when_rule["id_selector"])
```
JamesGardiner, line 262: This could be reverted if the `first` when rule id_selector was used to access a `first` attribute on the ListModel instance

JamesGardiner: So on `ListModel`, add:

```Python
@property
def first(self):
    return self.items[0]
```

app/data_model/list_store.py
```diff
@@ -40,6 +40,10 @@ def serialise(self):
     def __repr__(self):
         return f"<ListModel name={self.name} items={self.items}, primary_person={self.primary_person}>"
 
+    @property
+    def first(self):
+        return self.items[0]
```
MebinAbraham, line 45: This could throw a `KeyError` if the list is empty, we should probably handle it so it fails gracefully, and we can know from logs what the issue was.

JamesGardiner: @MebinAbraham I agree we should handle this gracefully, but do you mean handle `IndexError`?

MebinAbraham: Whoops sorry, I mean't `IndexError` for the list.

app/data_model/list_store.py
```diff
@@ -40,6 +40,13 @@ def serialise(self):
     def __repr__(self):
         return f"<ListModel name={self.name} items={self.items}, primary_person={self.primary_person}>"
 
+    @property
+    def first(self):
+        try:
+            return self.items[0]
+        except IndexError:
+            logger.error("unable to access first item in list, list is empty")
```
MebinAbraham, line 48: The exception still need to be raised, we don't want it to `return None`

app/data_model/list_store.py
```diff
@@ -40,6 +40,14 @@ def serialise(self):
...
+            raise IndexError
```
JamesGardiner, line 49:
```suggestion
            raise
```

JamesGardiner: From the docs:
`If no expressions are present, raise re-raises the last exception that was active in the current scope. If no exception is active in the current scope, a RuntimeError exception is raised indicating that this is an error.`

https://docs.python.org/3/reference/simple_stmts.html#raise

app/data_model/list_store.py
```diff
@@ -40,6 +40,13 @@ def serialise(self):
     def __repr__(self):
         return f"<ListModel name={self.name} items={self.items}, primary_person={self.primary_person}>"
 
+    @property
+    def first(self):
+        try:
+            return self.items[0]
+        except IndexError:
```
JamesGardiner, line 47:
```suggestion
        except IndexError as exc:
```

app/data_model/list_store.py
```diff
@@ -40,6 +40,13 @@ def serialise(self):
...
+            raise IndexError("unable to access first item in list, list is empty")
```
JamesGardiner, line 48:
```suggestion
            raise IndexError("unable to access first item in list, list is empty") from exc
```

tests/app/data_model/test_list_store.py
```diff
@@ -99,3 +99,19 @@ def test_repr():
...
+def test_first():
+    test_list = ListModel("people", ["abcde", "12345"])
+    assert test_list.first == "abcde"
+
+
+def test_first_raises_index_error_when_list_is_empty():
+    new_list = ListModel("people", [])
+
+    with pytest.raises(IndexError) as error:
+        new_list.first
```
JamesGardiner, line 113:
```suggestion
        new_list.first # pylint: disable=pointless-statement
```

app/data_model/list_store.py
ajmaddaford, line 48: If we're overriding the message, we should add the list name, something like:
```python
raise IndexError(f"unable to access first item in list, list '{self.name}' is empty")
```

PR #93, opened 2020-04-09
Update schema release to v0.1.2

berroar:
Updates schemas to v1.0.2 including latest content updates and changes.

### How to review 
Ensure schemas are loaded correctly and work as expected.


iwootten:
Might be worth adding the PR titles to the original release notes on eq-questionnaire-schemas, the numbers don't give a lot away.

PR #94, opened 2020-04-11
Update python version to 3.8

iwootten:

Updates runner to python 3.8, as per all validator and schema repos. 

Our main blocker to doing this was that gevent did not have support for python 3.8 and took a large amount of time to install. Version 1.5 of gevent was recently released, [providing this support](http://www.gevent.org/whatsnew_1_5.html). 

Additionally, this PR:

- removes the pinned version of grpcio, which was set when we were exploring different approaches to including new relic within runner. There has since been multiple releases of it.
- pins the version of pylint, due to errors incorrectly raised on dynamic wtform classes

### How to review 

Check the tests pass. Check if the gevent installation takes abnormal length of time to install.


iwootten:
I think it worth bumping the versions in this given the PR is now almost a month old. As I mentioned gevent has had two releases in the interim.

pricem14pc:
Added a don't merge until the performance investigation documentation has been completed and reviewed.

PR #98, opened 2020-04-16
Open PR on schemas release

petechd:
This introduces a new set of GitHub Actions that updates `Makefile` and raises new pull request whenever new version of eq-questionnaire-schemas is released. It's using a new schema version update script and new pull request yaml file which is triggered by `repository dispatch` with specific event type in schemas repo.

### How to review 
Check the code and try to trigger pull request by releasing new schemas version.


petechd:
[Test load-schemas is up to date with latest version before allowing merge (S)](https://trello.com/c/rQfJ7j8n/3775-test-load-schemas-is-up-to-date-with-latest-version-before-allowing-merge-s)

scripts/update_schemas_version.py
```diff
@@ -0,0 +1,29 @@
...
+from fileinput import FileInput
+
+import requests
+
+from requests.exceptions import RequestException
+
+
+try:
+    response = requests.get(
+        "https://api.github.com/repos/ONSdigital/eq-questionnaire-schemas/releases"
```
iwootten, line 11: You can visit https://api.github.com/repos/ONSdigital/eq-questionnaire-schemas/releases/latest to get the latest release rather than all.

scripts/update_schemas_version.py
```diff
@@ -0,0 +1,29 @@
...
+                continue
+            version = line.split(" ")[-1].strip()
+            line.replace(version, latest_tag)
+        file.close()
+        with FileInput("Makefile", inplace=True) as file:
+            for line in file:
+                print(line.replace(version, latest_tag), end="")  # noqa: T001
+        file.close()
+    else:
+        sys.exit(1)
```
iwootten, line 27: We probably want this to end with an explanation of what has happened.

scripts/update_schemas_version.py
```diff
@@ -0,0 +1,29 @@
...
+        file = open("Makefile", "r")
+        for line in file:
+            if "onsdigital/eq-questionnaire-schemas" not in line:
+                continue
+            version = line.split(" ")[-1].strip()
+            line.replace(version, latest_tag)
+        file.close()
+        with FileInput("Makefile", inplace=True) as file:
+            for line in file:
+                print(line.replace(version, latest_tag), end="")  # noqa: T001
```
iwootten, line 24: I think a better approach here would be to modify the Makefile to read the schemas version from another file. We could then have a .schemas-version and overwrite it each time this is run.

petechd: Thanks👍

scripts/update_schemas_version.py
petechd: Thanks👍

scripts/update_schemas_version.py
petechd: Thanks👍

Makefile
```diff
@@ -1,10 +1,13 @@
+FILE=schemas_version.txt
+VARIABLE=`cat $(FILE)`
```
iwootten, line 2: FILE, VARIABLE? Come on @petechd, you're better than this. 😄

Makefile
```diff
@@ -1,10 +1,13 @@
+FILE=schemas_version.txt
```
iwootten, line 1: As we already have a .python-version file, it would make sense to follow the same form for our schemas and call this .schemas-version

Makefile
petechd: I'm not familiar with this syntax but I'll give it a try😅

Makefile
petechd: Done, thanks!

Makefile
petechd: Lets hide this shameful code🤦‍♂️

scripts/update_schemas_version.py
```diff
@@ -0,0 +1,24 @@
...
+        latest_tag = response.json()["tag_name"]
+        file = open(".schemas-version", "w")
+        file.write(latest_tag)
+        file.close()
+    else:
+        logger.error("Can't check schemas version")
+        sys.exit(1)
+except RequestException:
+    logger.error("Can't check schemas version")
+    sys.exit(1)
```
iwootten, line 24: I think the if should be modified to throw a RequestException if the status_code is anything other than 200 and the else removed.

petechd: Done, thanks!

.github/workflows/schemas_release_pr.yml
```diff
@@ -0,0 +1,67 @@
...
+      - uses: actions/checkout@v2
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v2
+        with:
+          committer: GitHub <noreply@github.com>
+          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          commit-message: Update schema release version
+          title: Automated Schema Release Updates
+          body: This is an auto-generated PR with schema release update.
+          branch: schema-release
```
ajmaddaford, line 67: Is there a way we can get the schema version into this job? It would be good if we could make the commit message and title explicit - "Update schemas to version {version}".

petechd: Added, thanks!👍

.github/workflows/schemas_release_pr.yml
```diff
@@ -0,0 +1,73 @@
...
+        run: pip install pipenv==2018.11.26
+      - name: Cache virtualenv
+        id: cache-virtualenv
+        uses: actions/cache@v1
+        with:
+          path: ~/.local/share/virtualenvs/
+          key: ${{ runner.os }}-${{ env.PYTHON_VERSION }}-virtualenvs-${{ hashFiles('Pipfile.lock') }}
+      - name: Install virtual environment
+        if: steps.cache-virtualenv.outputs.cache-hit != 'true'
+        run: pipenv install --dev
```
MebinAbraham: We shouldn't need to install all Pipfile packages as the check schemas script is only using standard libraries. Can we just not use:
```yaml
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-python@v1
        with:
          python-version: 3.7
```

petechd: don't we use `requests` in schemas check?

MebinAbraham: Ya correct, I missed that sorry. I'm still not sure about installing all of the libdev libraries etc, and all 130+ Pipfile packages for this when we use 1. I wouldn't be against just installing requests but other may have diff opinions.

scripts/update_schemas_version.py
```diff
@@ -0,0 +1,23 @@
...
+    if response.status_code != 200:
+        raise RequestException(f"Got a {response.status_code} status code.")
+    latest_tag = response.json()["tag_name"]
+    file = open(".schemas-version", "w")
+    file.write(latest_tag)
+    file.close()
+
+except RequestException as ex:
+    logger.error("Can't check schemas version - %s.", ex)
+    sys.exit(1)
```
iwootten, line 23: We should replace this with a bash version to remove the dependency on python and the python step. I just found a gist to get a release number:

```suggestion
# https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c
get_latest_release() {
  curl --silent "https://api.github.com/repos/$1/releases/latest" | # Get latest release from GitHub api
    grep '"tag_name":' |                                            # Get tag line
    sed -E 's/.*"([^"]+)".*/\1/'                                    # Pluck JSON value
}

get_latest_release "onsdigital/eq-questionnaire-schemas" > .schemas-version
```

MebinAbraham: We need to make sure to update `eq-translations` as well to keep it consistent as that uses python to fetch the version etc.

scripts/update_schemas_version.sh
```diff
@@ -0,0 +1,8 @@
+# https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c
+get_latest_release() {
+  curl --silent "https://api.github.com/repos/$1/releases/latest" | # Get latest release from GitHub api
+    grep '"tag_name":' |                                            # Get tag line
+    sed -E 's/.*"([^"]+)".*/\1/'                                    # Pluck JSON value
+}
+
+get_latest_release "onsdigital/eq-questionnaire-schemas" > .schemas-version
```
ajmaddaford, line 8: Now that this is a shell script we should add the relevant lines to identify it:
```
#!/usr/bin/env bash
```
And to exit immediately on failure:
```
set -e
```

This is used in a number of our other scripts in this repo.

.github/workflows/schemas_release_pr.yml
```diff
@@ -0,0 +1,50 @@
+name: Schemas release PR
+
+on:
+  repository_dispatch:
+      types: schemas_release
+
+jobs:
+  python-dependencies:
```
MebinAbraham, line 8: This isn't needed anymore.

.github/workflows/schemas_release_pr.yml
```diff
@@ -0,0 +1,50 @@
...
+          echo "::set-env name=SCHEMAS_VERSION::$(cat .schemas-version)"
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          committer: GitHub <noreply@github.com>
+          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          commit-message: Update schemas to version ${{ env.SCHEMAS_VERSION }}
+          title: Update schemas to version ${{ env.SCHEMAS_VERSION }}
+          body: This is an auto-generated PR with schema release update.
```
ajmaddaford, line 49: I think we can drop the "with schema release update bit" as it's covered by the pull request title.

.github/workflows/schemas_release_pr.yml
petechd: All gone now, thanks!

scripts/update_schemas_version.sh
```diff
@@ -1,4 +1,4 @@
-# https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c
+#!/bin/bash
```
iwootten, line 1: I think we should keep a link to the original author's gist.

.github/workflows/schemas_release_pr.yml
ajmaddaford, line 49: I think we can be a bit more descriptive here by linking to the schemas release, something like:
> This is an auto-generated PR. Schema release https://github.com/ONSdigital/eq-questionnaire-schemas/releases/tag/v0.1.3.

scripts/update_schemas_version.sh
petechd: Done, thanks! 👍

scripts/update_schemas_version.sh
petechd: Done, thanks! 👍

.schemas-version
```diff
@@ -0,0 +1 @@
+v0.1.2
```
ajmaddaford, line 1: This should be v0.1.3

.github/workflows/schemas_release_pr.yml
petechd: Removed, thanks! 👍

scripts/update_schemas_version.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c
```
ajmaddaford, line 2: This comment should be immediately before the `get_latest_release()` line.

.schemas-version
petechd: Updated!👍

.github/workflows/schemas_release_pr.yml
petechd: Added a link, thanks!

scripts/update_schemas_version.sh
petechd: Moved it, thanks!

petechd:
It fails because the base branch for that bot PR schema release update is test-open-pr which is not set up to be analysed by Codacy. We can't open test auto PR based on master because update script and .schema-version is not in there yet. It needs to be merged first to see it working fine.

MebinAbraham:
Changes since:
- Have used GPG to sign the commit.-
- Use the schemas version passed from the event payload to avoid fetching it via curl.

To review, you should fork this and test it or you can view the workiing actions here:
	- Schema: https://github.com/ons-eq-team/eq-questionnaire-schemas/runs/612347767?check_suite_focus=true
	- Runner: https://github.com/ons-eq-team/eq-questionnaire-runner/runs/612348762?check_suite_focus=true
	- Generated pr: https://github.com/ons-eq-team/eq-questionnaire-runner/pull/1

Ensure all the secrets are uploaded as required.

PR #99, opened 2020-04-20
Remove section summary blocks

LJBabbage:
SectionSummary as blocks have been removed, instead they are defined as a summary element (which can be blank) in a section. This story updates all the schemas affected; creates a new routing rule which allows section end routing; updates return_to to return_to_summary and corrects all the related test.

N.B The branch for validator if its not checked in is remove-section-summary-block-add-section-end-routing

### How to review 
Check all SectionSummary blocks have been removed from the test_schemas and routing continues to work as expected


app/questionnaire/questionnaire_schema.py
```diff
@@ -148,6 +148,9 @@ def is_answer_in_list_collector_block(self, answer_id):
 
         return self.is_list_block_type(block["type"])
 
+    def is_summary_in_section(self, section_id: str):
+        return "summary" in self._sections_by_id.get(section_id)
+
```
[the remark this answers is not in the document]
LJBabbage: See below

app/questionnaire/router.py
```diff
@@ -71,36 +71,25 @@ def get_next_location_url(self, location, routing_path):
         Get the first incomplete block in section/survey if trying to access the section/survey end,
         and the section/survey is incomplete or gets the next default location if the above is false.
         """
-        current_block_type = self._schema.get_block(location.block_id)["type"]
         last_block_id = routing_path[-1]
-        last_block_type = self._schema.get_block(last_block_id)["type"]
         hub_enabled = self._schema.is_hub_enabled()
+        has_section_summary = self._schema.is_summary_in_section(location.section_id)
```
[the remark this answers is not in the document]
LJBabbage: The problem is we are allowing empty dicts for summaries, and there be processed as false in the next if statement. I kind of thought it would be cleaner to do this

app/questionnaire/router.py
```diff
@@ -246,6 +221,11 @@ def full_routing_path(self):
                 )
         return full_routing_path
 
+    def can_access_section_summary(self, section_id, list_item_id):
+        return self._schema.is_summary_in_section(
```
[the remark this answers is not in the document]
LJBabbage: As above

app/views/handlers/list_action.py
```diff
@@ -56,6 +65,7 @@ def handle_post(self):
         self.questionnaire_store_updater.save()
 
     def _get_location_url(self, block_id):
+
```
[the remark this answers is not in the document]
JamesGardiner: I knew you'd like that :wink:

app/questionnaire/router.py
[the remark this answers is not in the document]
JamesGardiner: Yep I've left a comment on validator re the empty summary defintiions. TLDR - we should enforce a minimum of `"enabled": true` such as we do with the hub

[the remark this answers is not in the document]
LJBabbage: Completely valid point. I just went with the story description which said they should be empty

app/routes/questionnaire.py
```diff
@@ -179,14 +182,47 @@ def get_section(schema, questionnaire_store, section_id, list_item_id=None):
...
+            previous_location_url=router.get_section_return_location_when_section_complete(
+                routing_path
+            ).url(),
+            schema=schema,
+            page_title=section_title,
+        )
+
+    if not schema.is_hub_enabled():
+        redirect_location = router.get_first_incomplete_location_in_survey()
+        return redirect(redirect_location.url())
```
ajmaddaford, line 224: As mentioned earlier, we should consider having a handler that encapsulates a lot of the logic in this method.

app/questionnaire/path_finder.py
```diff
@@ -145,6 +145,9 @@ def _evaluate_routing_rules(
             )
 
             if should_goto:
+                if "end" in rule["goto"].get("section", ""):
```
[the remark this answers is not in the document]
ajmaddaford: `end` was chosen as there won't necessarily be a summary at the end of a section.

app/questionnaire/router.py
```diff
@@ -71,36 +71,25 @@ def get_next_location_url(self, location, routing_path):
...
-                list_item_id=routing_path.list_item_id,
-            )
+            if has_section_summary:
+                return self.get_section_summary_url(location)
+            if hub_enabled:
+                return url_for(".get_questionnaire")
 
         if self.is_survey_complete() and not hub_enabled:
+            if has_section_summary:
+                return self.get_section_summary_url(location)
```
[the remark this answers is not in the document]
LJBabbage: I ummed and ahed on this for a bit as I was also unsure, but there was one functional test which breaks without it 'question_variants.spec.js'. Basically if you complete a non-hub survey, but then go back and change an answer, the old behaviour was to return to the section summary. As the card said to match current behaviour I put it in, but naturally it can be remove if its now not right

app/views/contexts/section_summary_context.py
```diff
@@ -122,7 +118,7 @@ def _list_summary_element(
             **list_context(
                 list_collector_block["summary"],
                 for_list=list_collector_block["for_list"],
-                return_to=current_location.block_id,
+                return_to_summary=current_location.section_id,
```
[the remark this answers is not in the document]
LJBabbage: Yep, completely forgot about this

app/questionnaire/router.py
```diff
@@ -246,6 +221,11 @@ def full_routing_path(self):
                 )
         return full_routing_path
 
+    def can_access_section_summary(self, section_id, list_item_id):
```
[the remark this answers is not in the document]
LJBabbage: I wonder if is_location_valid needed with can_access_section_summary already there. When the section end point gets hit, this will determine whether the survey is complete and has a summary, if it returns false then the get_section function in questionnaire will redirect it to the correct block. 

My issue would be if we did have a is_location_valid like the other blocks which raises an exception, that if someone saved and resumed and there was on the hub, that the click to sections/<section_id>/ which would instantiate a SectionSummary handler and raise an exception most times if we followed the current pattern.

LJBabbage:
Yep, I can see why, we used to have

```
 if (
            hub_enabled
            and location.block_id == last_block_id
            and self._progress_store.is_section_complete(
            )
        ):
            return url_for(".get_questionnaire")

        if (
            last_block_type == "SectionSummary"
            and current_block_type != last_block_type
            and self._progress_store.is_section_complete(
            )
        ):
            return url_for(
                "questionnaire.block",
                block_id=last_block_id,
            )
```
It's slightly odd logic, but would jump back to the section summary if the section was complete on the 2nd if. Feasibly if the section is complete we should just be able to go to the summary if it exists, so no need for location.block_id == last_block_id in the new code
It has also highlighted this potentially funny test and schema (test_questionnaire_list_change_evaluates_sections.py) Why would it go back to the list collector on a change to primary. It was getting around it as primary was neither last block or had a section summary.

ajmaddaford:
Now might be the time to deal with this differently. We could use the return_to_summary url param to make returning to the section summary explicit. It would also mean if someone navigated back from a summary via the previous links, they weren't taken straight back to the summary when submitting an answer (which I think is the correct behaviour).

LJBabbage:
Makes sense. Always pass the return_to_summary in the section summary context,  and use it unless the section becomes incomplete from a change. That would then indeed then allow the previous link to  just go back and forward on the same path the respondent already followed, unless a change led to alternative routing. I will add it in now

app/questionnaire/router.py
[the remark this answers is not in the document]
LJBabbage: Now return_to_summary is on Final summary, the above has been absorbed into the proceeding condition, so the flag can correctly be used to hand back to the final summary if it doesn't have a section summary but only if the survey is complete.

tests/functional/spec/list_collector_driving_question_checkbox.spec.js
```diff
@@ -36,27 +38,30 @@ describe('List Collector Driving Checkbox Question', function() {
...
   describe('Given the primary person is removed', function() {
     it('Then they aren\'t shown on the summary screen', function() {
-      $(SummaryPage.previous()).click();
+      $(SectionSummaryPage.previous()).click();
       $(ListCollectorTemporaryAwayPage.previous()).click();
       $(ListCollectorPage.previous()).click();
       $(AnyoneUsuallyLiveAtPage.previous()).click();
       $(PrimaryPersonListCollectorPage.noIDonTUsuallyLiveHere()).click();
       $(PrimaryPersonListCollectorPage.submit()).click();
+      $(AnyoneUsuallyLiveAtPage.submit()).click();
```
LJBabbage, line 53: See Andrews comments regarding using multiple previous clicks on why these are now needed.

app/questionnaire/router.py
```diff
@@ -66,44 +66,28 @@ def can_access_hub(self):
...
-            return url_for(
-                "questionnaire.block",
-                block_id=last_block_id,
-                list_name=routing_path.list_name,
-                list_item_id=routing_path.list_item_id,
-            )
+            if has_section_summary:
+                return self.get_section_summary_url(location)
+            if hub_enabled:
+                return url_for(".get_questionnaire")
```
MebinAbraham, line 87: We should be consistent with these, one is a method of router, the other just uses `url_for`. The questionnare url is commonly used so it makes sense for them to all live in the same place.

app/views/handlers/section_summary.py
```diff
@@ -1,20 +1,44 @@
...
+        self._router = router
+        self._section_id = current_location.section_id
+        self._list_item_id = current_location.list_item_id
+        self._routing_path = router.routing_path(
+            section_id=self._section_id, list_item_id=self._list_item_id
+        )
+        self.page_title = schema.get_title_for_section(self.current_location.section_id)
 
-    def get_context(self):
+    def can_access_section_summary(self):
```
MebinAbraham, line 20: Since this is invoked on the `SectionSummary` object I don't it is necessary to have `section_summary` in the name of the method. Can we use something like:
```py
def is_accessible(self):
```

app/questionnaire/router.py
```diff
@@ -66,44 +66,28 @@ def can_access_hub(self):
...
 
         if self.is_survey_complete() and not hub_enabled:
-            last_section_id = self._schema.get_section_ids()[-1]
-            last_block_id = self._schema.get_last_block_id_for_section(last_section_id)
-            return Location(section_id=last_section_id, block_id=last_block_id).url()
+            return self.get_last_section_last_block_location().url()
 
         block_id_index = routing_path.index(location.block_id)
         # At end of routing path, so go to next incomplete location
```
ajmaddaford, line 93: The logic after this comment is checking to see if the current location is at the end of the routing path, which duplicates some of the logic above. We should look to rationalise this logic about section summary/hub/final summary.

app/routes/questionnaire.py
```diff
@@ -179,22 +182,41 @@ def get_section(schema, questionnaire_store, section_id, list_item_id=None):
...
         section_id=section_id, list_item_id=list_item_id
     )
 
+    routing_path = router.routing_path(section_id=section_id, list_item_id=list_item_id)
+
     if section_status == CompletionStatus.COMPLETED:
-        return redirect(
-            router.get_section_return_location_when_section_complete(routing_path).url()
-        )
+        return redirect(router.get_first_location_in_section(routing_path).url())
```
ajmaddaford, line 219: I think quite a bit of this logic can be simplified by dealing with the `GET` first (as we do in the other methods). I think the logic would be something like:

```
if GET:
    if section not complete:
        redirect to first incomplete location
    if summary in section:
        render summary
    else:
        redirect to first location
```

app/routes/questionnaire.py
```diff
@@ -179,22 +182,41 @@ def get_section(schema, questionnaire_store, section_id, list_item_id=None):
...
 
+    if section_id not in router.enabled_section_ids:
+        return redirect(url_for(".get_questionnaire"))
+
+    location = Location(
+        section_id=section_id,
+        list_name=schema.get_repeating_list_for_section(section_id),
+        list_item_id=list_item_id,
+    )
+    section_summary = SectionSummary(schema, questionnaire_store, location, router)
```
ajmaddaford, line 193: Now that we have a section summary handler, the router should be initialised by that and not accessed directly.

app/views/handlers/section_summary.py
```diff
@@ -1,20 +1,44 @@
...
+        return self._schema.is_summary_in_section(
+            self.current_location.section_id
+        ) and self._questionnaire_store.progress_store.is_section_complete(
+            self._section_id, self._list_item_id
+        )
+
+    def context(self):
         section_summary_context = SectionSummaryContext(
-            self._language,
+            flask_babel.get_locale().language,
```
ajmaddaford, line 29: `language` should be passed in, as we do for block handlers so that there isn't a flask_babel dependency here.

app/questionnaire/router.py
```diff
@@ -320,3 +312,8 @@ def get_section_summary_url(location):
             section_id=location.section_id,
             list_item_id=location.list_item_id,
         )
+
+    def get_last_section_last_block_location(self):
```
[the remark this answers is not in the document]
LJBabbage: Got carried away with too many last

app/questionnaire/path_finder.py
```diff
@@ -145,6 +145,9 @@ def _evaluate_routing_rules(
             )
 
             if should_goto:
+                if "End" in rule["goto"].get("section", ""):
```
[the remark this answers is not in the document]
LJBabbage: It is in the instance, perhaps its due to the reuse of the work section, but in this case it is looking for a string and not a dict

app/questionnaire/router.py
LJBabbage: Not sure I understand, both are using url_for in router.py, I just extracted the top one into its own static method. I will re-combine to make it clearer

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
MebinAbraham: Ah ya, my bad.

app/questionnaire/router.py
MebinAbraham: As in, we now have a method which just wrap `url_for` for section summary url. Whereas, get_questionnaire is fetched directly. I think it would be clean to have a wrapper method of any use of url_for since it is commonly used.

LJBabbage: I have folded this back in for now, but Router does this in 6 places across the class and has 3 different flavours, for section, for block and for questionnaire, so we would have to unpack args in some way if we wanted to do this. This to me sounds like a class refactor, as there are a number of things in router that need a second look.

app/views/handlers/section_summary.py
LJBabbage: No longer required

app/routes/questionnaire.py
```diff
@@ -179,33 +183,46 @@ def get_section(schema, questionnaire_store, section_id, list_item_id=None):
...
         )
 
-    return redirect(
-        router.get_first_incomplete_location_for_section(
-            routing_path=routing_path
-        ).url()
+    section_summary = _get_section_summary(
+        list_item_id, questionnaire_store, schema, section_id
     )
+    return redirect(section_summary.get_next_location_url())
```
ajmaddaford, line 225: There is a lot going on here and I think more should be handled by the handler. Ideally the handler should abstract the behaviour away from the route so we can highlight the places where redirects happen. This code won't work as methods would need to be implemented in the handler, but this could look something like:

```python
    try:
        section_handler = SectionHandler(
            schema=schema,
            questionnaire_store=questionnaire_store,
            section_id=section_id,
            list_item_id=list_item_id,
            language=flask_babel.get_locale().language
        )
    except InvalidLocationException:
        return redirect(url_for(".get_questionnaire"))

    if request.method == "GET":
        if section_handler.can_display_summary():
            return _render_page(
                block_type="SectionSummary",
                context=section_handler.context(),
                current_location=section_handler.current_location,
                previous_location_url=section_handler.get_previous_location_url(),
                schema=schema,
                page_title=section_handler.page_title,
            )

        return redirect(section_handler.get_section_resume_url())

    return redirect(section_handler.get_next_location_url())
```
- The `InvalidLocationException` would deal with the `router.enabled_section_ids` check
- `section_handler.can_display_summary` would deal with checking that a section summary exists and the section is complete
- `get_section_resume_url` would deal with where the user should resume i.e. the first location if the section is complete or the first incomplete location

app/questionnaire/router.py
```diff
@@ -146,7 +140,7 @@ def get_first_incomplete_location_in_survey(self):
             if location:
                 return location
 
-        return self.get_last_section_last_block_location()
+        return self.get_last_location_in_survey()
```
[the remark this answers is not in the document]
MebinAbraham: We should change `get_first_incomplete_location_in_survey` to `get_first_incomplete_location_in_questionnaire` for consistency.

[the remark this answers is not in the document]
LJBabbage: @MebinAbraham You were the one that suggested get_last_location_in_survey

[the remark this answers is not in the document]
MebinAbraham: Yes because it is consistent with the other method. My point is, if we want to use `questionnaire` instead, then need to make sure we change the other reference also.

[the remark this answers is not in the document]
LJBabbage: Sticking with survey, Mebin is right we need to be consistent, and if I start pulling the questionnaire/survey thread I am going to be here all night, sounds a nice easy refactoring piece if we do want to change all of them across the app

tests/functional/spec/relationships.spec.js
```diff
@@ -107,109 +114,85 @@ describe('Relationships', function() {
...
         $(RelationshipsPage.submit()).click();
         expect($(RelationshipsPage.playback()).getText()).to.contain('Tom Bowden is Samuel Clemens’ …');
-
         $(RelationshipsPage.sonOrDaughter()).click();
         $(RelationshipsPage.submit()).click();
         browser.pause(20000);
         expect($(RelationshipsPage.playback()).getText()).to.contain('Tom Bowden is Olivia Clemens’ …');
       });
-
-      it('When I add a household member after answering the relationships question but remove them again before exiting the list-collector block, Then I will not be asked for the original household members relationships to the new member', function() {
```
LJBabbage, line 199: Due to the way we now explicitly go back to the summary only if its the last block or the return_to_summary flag is set. In this instance, as the respondent is going back and forward in the section it won't go back to the summary page now and this test become redundant

test_schemas/en/test_relationships.json
```diff
@@ -678,11 +689,43 @@
                             ]
                         },
                         {
-                            "type": "Confirmation",
-                            "id": "confirmation",
+                            "id": "relationship-interstitial",
                             "content": {
-                                "title": "Thank you for your answers, do you wish to submit"
-                            }
+                                "title": "Relationship section",
```
LJBabbage, line 694: I have had to add an Interstitial and change the confirmation into a summary page into this test. The reason was the test was jumping to a block multiple back by using the url, effectively simulating using previous. However as we have changed how previous works, and after the use of previous the respondent follows their path back, some of the tests associated with this will fail, as they are expecting a jump to the end. If instead we use a summary click which now is the only way to have that behaviour this will work as expected. The Interstitial went in as there was a test where the behaviour of a block after a relationship collector was being tested.

app/routes/questionnaire.py
```diff
@@ -165,47 +165,40 @@ def post_questionnaire(schema, questionnaire_store):
     return redirect(router.get_first_incomplete_location_in_survey().url())
 
 
-@questionnaire_blueprint.route("sections/<section_id>/", methods=["GET"])
-@questionnaire_blueprint.route("sections/<section_id>/<list_item_id>/", methods=["GET"])
+# pylint: disable=too-many-return-statements
+@questionnaire_blueprint.route("sections/<section_id>/", methods=["GET", "POST"])
+@questionnaire_blueprint.route(
+    "sections/<section_id>/<list_item_id>/", methods=["GET", "POST"]
+)
```
ajmaddaford, line 172: I think it may have got lost in my last review comments, but we should consider making this `GET` only and using a link styled as a button on the section summary page. This would mean one less `POST` request on every section summary in completion of a questionnaire. As this wasn't captured in the initial card we could break this out into a separate piece of work.

app/views/handlers/section.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+        if section_status != CompletionStatus.COMPLETED:
+            return self._router.get_first_incomplete_location_for_section(
+                self._routing_path
+            ).url()
+
+        return self._router.get_first_incomplete_location_for_section(
+            routing_path=self._routing_path
+        ).url()
```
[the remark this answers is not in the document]
LJBabbage: I don't know, changed it so many times, will have to check

tests/functional/base_pages/section-summary.page.js
```diff
@@ -0,0 +1,59 @@
...
+  summaryItems() { return 'table.summary__items'; }
+
+  summaryRowAction(number = 1) { return 'tbody:nth-child(' + number + ') tr td.summary__actions a'; }
+
+  summaryRowTitle(number = 1) { return 'tbody:nth-child(' + number + ') tr td.summary__item-title'; }
+
+  summaryRowValue(number = 1) { return 'tbody:nth-child(' + number + ') tr td.summary__values'; }
+}
+
+module.exports = new SectionSummaryPage();
```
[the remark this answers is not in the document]
LJBabbage: I don't have dynamic section summary generation anymore, so the one I have created has to cope with all of them.

app/routes/questionnaire.py
LJBabbage: Got to go separate story, I am starting to get frustrated and need to get this checked in

app/views/handlers/section.py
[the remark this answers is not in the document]
LJBabbage: Yep, it was falling through that function to the default return which was first location, one of the functions in router which is a bit ropey, get_first_incomplete_location_for_section can return first location in section, yeah.

tests/functional/base_pages/section-summary.page.js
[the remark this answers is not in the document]
ajmaddaford: It has to be dynamic as otherwise it won't cope with all of them, only ones with lists called people and visitors.

tests/functional/base_pages/list-collector-summary.page.js
```diff
@@ -0,0 +1,43 @@
...
+    return '[data-qa=visitors-list-summary] tbody:nth-child(' + number + ') tr td.summary__item-title';
+  }
+
+  listLabel(instance) { return `tbody:nth-child(${instance}) td:first-child`; }
+
+  peopleListLabel(listItemInstance) { return 'div[data-qa="people-list-summary"] tbody:nth-child(' + listItemInstance + ') td:first-child'; }
+
+}
+
+module.exports = new ListCollectorSummary();
```
ajmaddaford, line 43: There is code in `generate_pages.py` that handles creating page specs from `SectionSummary` definitions. This needs to be refactored into logic that generates an appropriate page object based on the contents of the section summary property or the answers in the section.

This is important as currently:
- This will not work for list summaries containing something other than `people` or `visitors`.
- References to answers have changed from things like `PersonalSummaryPage.dateOfBirthQuestion()` to `SectionSummaryPage.summaryRowTitle(2)`, which is far less readable.

app/views/handlers/section.py
```diff
@@ -0,0 +1,78 @@
...
+        section_status = self._questionnaire_store.progress_store.get_section_status(
+            section_id=self._section_id, list_item_id=self._list_item_id
+        )
+
+        if section_status != CompletionStatus.COMPLETED:
+            return self._router.get_first_incomplete_location_for_section(
+                self._routing_path
+            ).url()
+
+        return self._router.get_first_location_in_section(self._routing_path).url()
```
MebinAbraham, line 65: We already have a method to check if a section is complete.
```suggestion
        is_section_complete = self._questionnaire_store.progress_store.is_section_complete(
            section_id=self._section_id, list_item_id=self._list_item_id
        )

        if is_section_complete:
            self._router.get_first_location_in_section(self._routing_path).url()

        return self._router.get_first_incomplete_location_for_section(
            self._routing_path
        ).url()
```

app/views/handlers/list_action.py
```diff
@@ -32,14 +34,21 @@ def is_location_valid(self):
         return True
 
     def get_previous_location_url(self):
-        block_id = self._request_args.get("previous") or self._request_args.get(
-            "return_to"
-        )
+        if self._request_args.get("return_to_summary"):
```
MebinAbraham, line 37: `self._request_args.get("return_to_summary")` is used 7 times, can we add a top level property to block handler so that it is defined in one place?
```py
self._return_to_summary = self._request_args.get("return_to_summary")
```

tests/functional/spec/list_collector_primary_person.spec.js
```diff
@@ -106,9 +107,10 @@ describe('Primary Person List Collector Survey', function() {
       browser.openQuestionnaire('test_list_collector_primary_person.json');
     });
 
-    it('When the user says they do not live there, then an empty list is displayed', function() {
+    it('When the user says they do not live there, then the ', function() {
```
MebinAbraham, line 110: then the ...?

.schemas-version
```diff
@@ -1 +1 @@
-v0.1.5
+v0.1.6
```
MebinAbraham, line 1: Doesn't need to be changed here as it doesn't block this PR. It will be bumped on release.

tests/functional/generate_pages.py
```diff
@@ -647,11 +645,61 @@ def process_schema(in_schema, out_dir, spec_file, require_path=".."):
...
+            "pageFile": page_filename,
+            "type_id": section_id,
+            "type_name": camel_case(generate_pascal_case_from_id(section_id)),
+            "relativeRequirePath": relative_require,
+        }
+
+        page_spec.write(HEADER.substitute(section_context))
+        page_spec.write(CLASS_NAME.substitute(section_context))
+        page_spec.write(CONSTRUCTOR.substitute(section_context))
+        process_summary(schema_data, page_spec, False, section_summary=True)
```
ajmaddaford, line 691: Calling `process_summary` here adds methods for all answers in the questionnaire to every section summary page file e.g. `generated_pages/section_summary/house-details-section-summary.page.js` and `generated_pages/section_summary/property-details-section-summary.page.js` are almost identical.

LJBabbage: Happy to fix, but looking at the code it looks like its always been bloated
```
if block["type"] in ("Summary", "SectionSummary"):
            collapsible = block.get("collapsible", False)
            process_summary(schema_data, page_spec, collapsible)
```
Summary and SectionSummary had no separation in what was passed, therefore it would always have been a bit on the chubby side

LJBabbage:
"Just the tests to fix + run_validator.sh needs to point to latest"? run_validator.sh is pointed to latest and the tests only break as explained because it can't run load schemas, am I missing something?

tests/functional/generate_pages.py
ajmaddaford: Ah, fair point. I will raise a card for fixing that separately.

LJBabbage: No need, already done it, the whole py needs looking at. It really could do with some rationalising and tiding up. In another places I fixed a for loop that was followed by the same loop of things, when it only need to go through once.

ajmaddaford: Yeah, it's a bit of a mess. I will amend the card i raised to look into refactoring it 😀.

tests/functional/generate_pages.py
```diff
@@ -392,78 +392,71 @@ def process_calculated_summary(answers, page_spec):
...
-                            SUMMARY_ANSWER_GETTER.substitute(answer_context)
+                            SECTION_SUMMARY_ANSWER_GETTER.substitute(answer_context)
                         )
-
                         page_spec.write(
-                            SUMMARY_ANSWER_EDIT_GETTER.substitute(answer_context)
+                            SECTION_SUMMARY_ANSWER_EDIT_GETTER.substitute(
+                                answer_context
+                            )
                         )
```
ajmaddaford, line 441: I think we can drop this `if` block as it generates the same accessors as the `SUMMARY_ANSWER_GETTER` and

MebinAbraham:
> "Just the tests to fix + run_validator.sh needs to point to latest"? run_validator.sh is pointed to latest and the tests only broke because as explained it couldn't run load schemas, am I missing something?

Master isn't pointing at latest, we changed it temporarily to unblock other PRs.
https://github.com/ONSdigital/eq-questionnaire-runner/blob/remove-section-summary-blocks/scripts/run_validator.sh#L4

tests/functional/generate_pages.py
LJBabbage: We will call that 1 on the refactoring card  as SECTION_SUMMARY_ANSWER_GETTER etc wasn't added by this story

LJBabbage:
```
Master isn't pointing at latest, we changed it temporarily to unblock other PRs.
```
Ah, I see what happened, it came through on the merge,

PR #111, opened 2020-04-29
WLH content changes - return link

petechd:
This changes the text on a return link to `Cancel and return to the previous page` for a multiple questions. It is part of a larger WLH schema changes card: [schemas PR](https://github.com/ONSdigital/eq-questionnaire-schemas/pull/49).

### How to review 
Check the overflow to make sure the change is correctly implemented. Check Google sheet attached to the card to compare individual changes.

### DIFF
--- a/templates/list-action.tmpl
+++ b/templates/list-action.tmpl
@@ -6,7 +6,7 @@
       {% if content.block.cancel_text %}
         {{ content.block.cancel_text }} <br>
       {% endif %}
-      <a id="cancel-and-return" href="{{ previous_location_url }}">{{ _("Return to the previous page") }}</a>
+      <a id="cancel-and-return" href="{{ previous_location_url }}">{{ _("Cancel and return to the previous page") }}</a>
     </p>
   {% endif %}
 {% endblock cancel_and_return %}
--- a/tests/integration/questionnaire/test_questionnaire_list_collector.py
+++ b/tests/integration/questionnaire/test_questionnaire_list_collector.py
@@ -245,4 +245,4 @@ class TestQuestionnaireListCollector(IntegrationTestCase):
 
         self.get(remove_link)
 
-        self.assertInBody("Don’t need to remove this person?")
+        self.assertNotInBody("Don’t need to remove this person?")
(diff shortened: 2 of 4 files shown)


petechd:
[[B] Add content changes to WLH (S)](https://trello.com/c/AgiPdvOV/3799-b-add-content-changes-to-wlh-s)

templates/list-action.tmpl
```diff
@@ -6,7 +6,7 @@
       {% if content.block.cancel_text %}
         {{ content.block.cancel_text }} <br>
```
ajmaddaford, line 7: Looking at the Overflow diagram it looks like `cancel_text` is no longer used. If so, this needs to be removed from the schema. If this feature is no longer required the feature should be removed (from this template, test schemas, json schema and eq-translations).

templates/list-action.tmpl
```diff
@@ -6,7 +6,7 @@
...
       {% endif %}
-      <a id="cancel-and-return" href="{{ previous_location_url }}">{{ _("Return to the previous page") }}</a>
+      <a id="cancel-and-return" href="{{ previous_location_url }}">{{ _("Cancel and return to the previous page") }}</a>
```
ajmaddaford, line 9: The link has been moved in the new design. It now needs to appear between the last answer and the submit button:
image

petechd: So  I was told it is an incorrect design, it stays where it is now.

templates/list-action.tmpl
petechd: As discussed, I will remove them from schemas and feature will be removed on separate card. Thanks!

tests/integration/questionnaire/test_questionnaire_list_collector.py
```diff
@@ -245,4 +245,4 @@ def test_cancel_text_displayed_on_remove_block_if_exists(self):
 
         self.get(remove_link)
 
-        self.assertInBody("Don’t need to remove this person?")
```
ajmaddaford, line 248: The `test_cancel_text_displayed_on_remove_block_if_exists` test should be removed now we can no longer have cancel text on a remove page.

petechd: Removed, thanks 👍

PR #114, opened 2020-04-30
Fix key errors on questionnaire summary context

JamesGardiner:
Adds a boolean flag to optionally ignore custom section summaries. This fixes an issue whereby custom section summaries were causing KeyError exceptions on Summary blocks.

### How to review 
Check that the tests pass and there are no regressions.


JamesGardiner:
[Fix bug in questionnaire summary context handler](https://trello.com/c/am7JRLZH/3841-fix-bug-in-questionnaire-summary-context-handler)

tests/integration/questionnaire/test_questionnaire_custom_summary.py
```diff
@@ -0,0 +1,120 @@
...
+        filtered = [html for html in selected if text in html.get_text()]
+
+        return filtered[0].get("href")
+
+    def get_previous_link(self):
+        selector = "#top-previous"
+        selected = self.getHtmlSoup().select(selector)
+        return selected[0].get("href")
+
+    def test_happy_path(self):
```
berroar, line 23: Given how this test is named should there be a test for an unhappy path as well? Generally a bit confusing for me on what exactly is being tested here with removing a mistake etc.

JamesGardiner: Thanks for the insight Rhys - I'll try to keep this in mind when naming tests in future. In any case this has now been removed 👍

test_schemas/en/test_list_collector.json
```diff
@@ -326,6 +326,43 @@
...
+                                        "description": "",
+                                        "label": "Decimal test to 2 decimal places",
+                                        "mandatory": false,
+                                        "type": "Number",
+                                        "decimal_places": 2
+                                    }
+                                ],
+                                "description": "",
+                                "id": "test-number-range-question",
+                                "title": "Please enter test values (none mandatory)",
```
petechd, line 363: non-mandatory?

petechd, line 363: Non-mandatory?

JamesGardiner: I don't see a reason for it to be mandatory - is there something I'm missing?

test_schemas/en/test_list_collector.json
```diff
@@ -326,6 +326,43 @@
...
+                                    {
+                                        "id": "test-currency",
+                                        "description": "",
+                                        "label": "Currency Test",
+                                        "mandatory": false,
+                                        "type": "Currency",
+                                        "currency": "GBP"
+                                    },
+                                    {
+                                        "id": "square-kilometres",
```
petechd, line 344: Can we stay consistent and start answer ids with "test-"?

test_schemas/en/test_list_collector.json
petechd: I mean the literal text😆

test_schemas/en/test_list_collector_driving_checkbox.json
```diff
@@ -670,6 +670,11 @@
                             "id": "who-lives-here-summary",
                             "title": "Who Lives Here Summary",
                             "type": "SectionSummary"
+                        },
+                        {
+                            "id": "summary",
```
LJBabbage, line 675: Don't think this will work as expected when you merge it, due to the changes made in SectionSummary. Summary needs to be in a separate section as the section summary will always be the last block if its enabled, so this might go Block -> summary -> section summary

app/views/handlers/question.py
```diff
@@ -92,6 +92,7 @@ def handle_post(self):
 
         self.questionnaire_store_updater.add_completed_location()
 
+        # pylint: disable=using-constant-test
```
LJBabbage, line 95: What was the reason for this?

test_schemas/en/test_list_collector.json
JamesGardiner: :laughing:

app/views/handlers/question.py
JamesGardiner: Pylint was failing lint without it

LJBabbage: No really?

app/views/contexts/questionnaire_summary_context.py
```diff
@@ -19,17 +19,25 @@ def __call__(self, collapsible=True, answers_are_editable=True):
...
-            self._metadata,
-        )
-
         for section_id in self._router.enabled_section_ids:
-            for group in section_summary_context(Location(section_id=section_id))[
-                "summary"
-            ]["groups"]:
-                yield group
+            section = self._schema.get_section(section_id)
+            if section.get("summary"):
```
ajmaddaford, line 24: I think this should this be checking the summary `items` property. Currently any section with a section summary (custom or not) won't be on the questionnaire summary?

app/views/contexts/questionnaire_summary_context.py
```diff
@@ -19,17 +19,25 @@ def __call__(self, collapsible=True, answers_are_editable=True):
...
-            self._metadata,
-        )
-
         for section_id in self._router.enabled_section_ids:
-            for group in section_summary_context(Location(section_id=section_id))[
-                "summary"
-            ]["groups"]:
-                yield group
+            section = self._schema.get_section(section_id)
+            if section.get("summary", {}).get("items"):
```
LJBabbage, line 24: Not sure on this one, I understand we wish to hide lists in the Final summary, but if a list shares a group with another block then that will also be removed. In test_list_collector there is a block called test-number-block which is just a general question, this won't appear on the summary when I believe it should. 

N.B. This might be a pattern we shouldn't be using (i.e lists sharing groups) rather than a bug

JamesGardiner: Yeah I spoke to @ajmaddaford about this - it's expected that any questions in a section that has a custom summary definition won't display those questions on the final summary

ajmaddaford: Just to be clear, this is because we only use this in who lives here and don't have any requirements for any other behaviour. When we know more, we can build it.

app/views/contexts/questionnaire_summary_context.py
```diff
@@ -19,17 +19,25 @@ def __call__(self, collapsible=True, answers_are_editable=True):
...
+                yield Group(
+                    group,
+                    routing_path,
+                    self._answer_store,
+                    self._list_store,
+                    self._metadata,
+                    self._schema,
+                    location,
+                    self._language,
+                ).serialize()
```
ajmaddaford, line 43: As discussed, we should continue using `SectionSummaryContext`.

tests/integration/questionnaire/test_questionnaire_list_collector.py
```diff
@@ -231,3 +231,26 @@ def test_cancel_text_displayed_on_edit_block_if_exists(self):
         self.get(change_link)
 
         self.assertInBody("Don’t need to change anything?")
+
+    def test_summary_does_not_display_answers(self):
```
ajmaddaford, line 235: The name of this test isn't clear. Something like `test_questionnaire_summary_does_not_display_custom_section_summary` would be better (even though it's a bit wordy).

test_schemas/en/test_list_collector.json
```diff
@@ -321,6 +321,43 @@
...
+                                        "mandatory": false,
+                                        "type": "Number",
+                                        "decimal_places": 2
+                                    }
+                                ],
+                                "description": "",
+                                "id": "test-number-range-question",
+                                "title": "Please enter test values (non-mandatory)",
+                                "type": "General"
+                            }
```
ajmaddaford, line 360: I don't think we should be adding a question to this schema to test this. Can we use the existing `test_list_summary_on_question.json` schema? That schema already has an additional non-list question in the section.

JamesGardiner: Ah I missed this - I did initially look for an existing schema that fits the requirement of having a mid section Question block along with a ListCollector. Let me take a look

test_schemas/en/test_list_summary_on_question.json
```diff
@@ -369,6 +369,43 @@
...
+                                        "mandatory": false,
+                                        "type": "Number",
+                                        "decimal_places": 2
+                                    }
+                                ],
+                                "description": "",
+                                "id": "test-number-range-question",
+                                "title": "Please enter test values (non-mandatory)",
+                                "type": "General"
+                            }
```
ajmaddaford, line 408: Is this additional question needed given there is already a non-list question "Are any of these people related to one another"?

tests/functional/spec/list_collector.spec.js
```diff
@@ -8,6 +8,7 @@ const ListCollectorEditPage = require("../generated_pages/list_collector/list-co
 const ListCollectorRemovePage = require("../generated_pages/list_collector/list-collector-remove.page.js");
 const NextInterstitialPage = require("../generated_pages/list_collector/next-interstitial.page.js");
 const SummaryPage = require("../generated_pages/list_collector/section-summary.page.js");
+const SectionPage = require("../generated_pages/list_collector/section-summary.page");
```
ajmaddaford, line 11: This looks odd.

ajmaddaford: We should probably remove the changes to the schema and functional tests now that the schema doesn't need to be modified in this pr.

test_schemas/en/test_list_summary_on_question.json
JamesGardiner: No :+1:

tests/integration/questionnaire/test_questionnaire_list_collector.py
```diff
@@ -206,8 +204,19 @@ def test_list_summary_on_question(self):
...
 
+        self.post({"radio-mandatory-answer": "No, all household members are unrelated"})
+
+        self.assertInUrl("/sections/section/")
+
+        self.post()
+
+        self.assertInUrl("/summary/")
+
+        self.assertNotInBody("£12.00")
```
ajmaddaford, line 218: This is not testing "list summary on question" and should be moved into it's own test.

PR #119, opened 2020-05-11
Character limit to textfield (DS v17.1.4)

boxadesign:
Added the client side character limit validation to text fields.

### How to review 
The char limit will display and count exceeded limit if a `max_length` property/value exists in the schema.

This can be tested on the textfield schema which has a limit set of 20.


rmccar:
[Duplicate error messages output for mutually exclusive answers](https://trello.com/c/aN73jYc3/3719-duplicate-error-messages-output-for-mutually-exclusive-answers)

PR #120, opened 2020-05-11
Enable piping unit answers

MattElcock:
The EQAuthor team would like to add the functionality to pipe unit answer types across questions in Runner.

### How to review 
Have a look at the code and make sure it's OK. I was unsure on how to add in a test for this, so pointers on that would be amazing; I'll ask on Slack 'though as well.

### Checklist

### DIFF
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from babel.numbers import format_currency, format_decimal
 from babel.dates import format_datetime
+from babel.units import format_unit
 from dateutil.tz import tzutc
 from dateutil.relativedelta import relativedelta
 from flask_babel import ngettext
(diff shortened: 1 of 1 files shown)


ajmaddaford:
The PR is failing as the code isn't formatted as expected. This can be fixed easily by running `pipenv run black app`. After that it will fail as there are no tests for the new `format_unit` placeholder transform, these will need to be added in `tests/app/questionnaire/test_placeholder_transforms.py`.

Also, any schemas built to use this new transform won't pass schema validation until the transform is added to the [validation schemas](https://github.com/ONSdigital/eq-questionnaire-validator/tree/master/schemas/string_interpolation/transforms) (and the [definitions](https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/schemas/string_interpolation/definitions.json) updated).

pricem14pc:
Hi @MattElcock I'm going to close this for the time being but happy to progress this as a feature via our PO.

PR #122, opened 2020-05-14
Feature: Conditional section summaries

MebinAbraham:
Added the ability to have conditional section summaries so that you can choose whether a summary is shown when a section becomes complete.

### How to review 
Ensure the scenarios on the Trello card are satisfied.

### Checklist

### DIFF
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict, defaultdict
-from typing import List, Union
+from typing import List, Union, Mapping
 
 from flask_babel import force_locale
 
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -87,7 +87,7 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
     def get_show_on_hub_for_section(self, section_id):
         return self._sections_by_id.get(section_id).get("show_on_hub", True)
 
-    def get_summary_for_section(self, section_id: str) -> dict:
+    def get_summary_for_section(self, section_id: str) -> Mapping:
         return self._sections_by_id.get(section_id).get("summary", {})
 
     def show_summary_on_completion_for_section(self, section_id: str) -> bool:
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -1,5 +1,4 @@
 from collections import OrderedDict, defaultdict
-
 from typing import List, Union
 
 from flask_babel import force_locale
@@ -88,8 +87,11 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
     def get_show_on_hub_for_section(self, section_id):
         return self._sections_by_id.get(section_id).get("show_on_hub", True)
 
-    def get_summary_for_section(self, section_id: str):
-        return self._sections_by_id.get(section_id).get("summary")
+    def get_summary_for_section(self, section_id: str) -> dict:
+        return self._sections_by_id.get(section_id).get("summary", {})
+
+    def show_summary_on_completion_for_section(self, section_id: str) -> bool:
+        return self.get_summary_for_section(section_id).get("show_on_completion", False)
 
     def get_repeating_list_for_section(self, section_id):
         return self._sections_by_id.get(section_id).get("repeat", {}).get("for_list")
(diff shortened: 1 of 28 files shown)


scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=conditional-summaries-section
```
```suggestion
tag=latest
```

tests/functional/spec/features/show_section_summary_on_completion/show_section_summary_on_completion.spec.js
```diff
@@ -0,0 +1,58 @@
+import employmentStatusBlockPage from '../../../generated_pages/show_section_summary_on_completion/employment-status.page';
+import employmentSectionSummary from '../../../generated_pages/show_section_summary_on_completion/employment-section-summary.page';
+
+import proxyQuestionPage from '../../../generated_pages/show_section_summary_on_completion/proxy.page';
+import accommodationSectionSummary from '../../../generated_pages/show_section_summary_on_completion/accommodation-section-summary.page';
+
+import hubPage from '../../../base_pages/hub.page.js';
```
JamesGardiner, line 7: Nice to see modern imports in here :+1:

app/questionnaire/questionnaire_schema.py
```diff
@@ -88,8 +87,11 @@ def get_title_for_section(self, section_id):
     def get_show_on_hub_for_section(self, section_id):
         return self._sections_by_id.get(section_id).get("show_on_hub", True)
 
-    def get_summary_for_section(self, section_id: str):
-        return self._sections_by_id.get(section_id).get("summary")
+    def get_summary_for_section(self, section_id: str) -> dict:
```
JamesGardiner, line 90:
```suggestion
    def get_summary_for_section(self, section_id: str) -> Mapping:
```

MebinAbraham: This is a literal dict right?

tests/functional/spec/features/show_section_summary_on_completion/show_section_summary_on_completion.spec.js
```diff
@@ -0,0 +1,58 @@
...
+
+describe('Feature: Show section summary on completion', () => {
+
+    before("Launch survey", () => {
+        browser.openQuestionnaire("test_show_section_summary_on_completion.json");
+    });
+
+    describe('Given I am completing a section with thee summary turned off for the forward journey', () => {
```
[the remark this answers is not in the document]
MebinAbraham: Fixed. My annoying Mac's keyboard :P
https://support.apple.com/keyboard-service-program-for-mac-notebooks

tests/functional/spec/features/show_section_summary_on_completion/show_section_summary_on_completion.spec.js
```diff
@@ -0,0 +1,58 @@
...
+        it('When I reach the end of that section, Then I go straight to the hub' , () => {
```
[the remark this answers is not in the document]
MebinAbraham: Only capitalised the `Given` `When` `Then` since we do it in other specs but have changed it.

MebinAbraham:
> Works as expected, just a few grammar changes. Visually I think it looks better to have dict elements on new lines, even though there is only one element. I believe we do that everywhere else, but I won't block on that.

Regarding the dict formatting, it has been run through prettier. If we want this behaviour, we should look to change prettier.

app/questionnaire/questionnaire_schema.py
JamesGardiner: Correct, but best practice dictates using Mapping https://docs.python.org/3/library/typing.html#typing.Dict - Mapping is also used elsewhere, so the use of `dict` is inconsistent

JamesGardiner:
Should we look to change prettier config it will also need to be changed in the schema repo to ensure consistency

MebinAbraham:
> Should we look to change prettier config it will also need to be changed in the schema repo to ensure consistency

Yes we could do, but don't think it should be part of this.

JamesGardiner:
https://github.com/ONSdigital/eq-questionnaire-runner/pull/122/commits I feel like this should be blocked :stop_sign: :rofl:

JamesGardiner:
> > Should we look to change prettier config it will also need to be changed in the schema repo to ensure consistency
> 
> Yes we could do, but don't think it should be part of this.

Oh no, I agree it's a separate thing, but should we do it then the schema repo should also be taken into account

PR #125, opened 2020-05-19
Use Redis to store eQ Session data and standardise data storage classes

MebinAbraham:
- Introduces a new abstract base class to standardise our Storage backend classes.
- Cleaned up some duplication in our Storage classes.
- Use Redis to store eq-sessions instead of Datastore.
- Use consistent naming for the expires at attribute.
- Removed the `used_at` key from JTI as this is not needed. We simply try to insert the key, and if it already exists, then we know it has been used recently.
This also means that we now do not store a value for the JTI claim in Redis, hence we save some memory.
- In Redis, the key, and the value both contained the key field, therefore for EQSession, the `eq_session_id` has been dropped from the value. The key field is added back when fetching the value.

### How to review 
Ensure:
- the approach makes sense
- saving state to both Datastore and DynamoDB still works as expected.
- eq-session are stored and retrieved from Redis and not Datastore.
- app behaves as expected.

### Checklist


app/setup.py
```diff
@@ -25,9 +25,9 @@
 from app.keys import KEY_PURPOSE_SUBMISSION
 from app.helpers import get_span_and_trace
 from app.secrets import SecretStore, validate_required_secrets
-from app.storage.datastore import DatastoreStorage
-from app.storage.dynamodb import DynamodbStorage
-from app.storage.redis import RedisStorage
+from app.storage.datastore import Datastore
```
JamesGardiner, line 28: We can import these classes into `app.storage.__init__.py` and import them from there

app/storage/datastore.py
```diff
@@ -1,77 +1,39 @@
...
-
-class DatastoreStorage:
-    def __init__(self, client):
-        self.client = client
 
+class Datastore(StorageHandler):
     @Retry()
     def put(self, model, overwrite=True):
-        if not overwrite:
+        if overwrite is False:
```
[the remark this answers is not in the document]
MebinAbraham: As discussed, this was when I was going to introduce type hinting as this should be a bool, but didn't in the end as boto3 cant be done without extra packages. Therefore, I didn't want to partially introduce it.
Have reverted.

app/storage/storage.py
```diff
@@ -0,0 +1,81 @@
...
+
+        if not self._model_type:
+            raise ValueError("One of model/model_type is required")
+
+        if self._model_type not in self.TABLE_CONFIG:
+            raise KeyError("Invalid model_type provided")
+
+        self._config = self.TABLE_CONFIG[self._model_type]
+
+    @property
```
iwootten, line 44: Can these use the cached_property decorator?

MebinAbraham: The property is only ever called more than once in Datastore `put` atm. Have added anyway.

app/storage/redis.py
```diff
@@ -1,17 +1,49 @@
+import json
+
+from app.data_model.app_models import EQSession, UsedJtiClaim
 from app.storage.errors import ItemAlreadyExistsError
+from .storage import StorageModel, StorageHandler
+
 
+class Redis(StorageHandler):
+    def put(self, model, overwrite=False):
```
JamesGardiner, line 9: is overwrite not used?

MebinAbraham:
- Going to remove logic of jti/session out of the Redis class.
- Add TTL to session, like we do with the JTI
- Remove old redundant code to support eq-session not having an expiry.

app/storage/storage.py
```diff
@@ -0,0 +1,89 @@
...
+            "schema": app_models.UsedJtiClaimSchema,
+        },
+    }
+
+    def __init__(self, model=None, model_type=None):
+        self._model = model
+        self._model_type = type(model) if model else model_type
+
+        if not self._model_type:
+            raise ValueError("One of model/model_type is required")
```
ajmaddaford, line 41: This could be simpler by making `model_type` required and move the responsibility of getting the type to the caller.

app/storage/storage.py
```diff
@@ -0,0 +1,89 @@
...
+        return self._config["key_field"]
+
+    @cached_property
+    def expiry_field(self):
+        return self._config.get("expiry_field")
+
+    @cached_property
+    def item(self):
+        if self._model:
+            return self.schema.dump(self._model)
```
ajmaddaford, line 63: I don't think this is very clear. The model is only used on the load, so this seems more of a serialise method. There is also an inconsistency as the deserialisation for the `get` is done in the implementations. Maybe we should introduce `serialise` and `deserialise` methods here?

app/storage/redis.py
```diff
@@ -1,17 +1,42 @@
...
 
         if not record_created:
             raise ItemAlreadyExistsError()
+
+    def get(self, model_type, key_value):
+        storage_model = StorageModel(model_type=model_type)
+        item = self.client.get(key_value)
+
+        if item:
+            item_dict = json.loads(item.decode("utf-8"))
```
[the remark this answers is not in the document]
JamesGardiner: no, it will throw a JSONDecodeError

app/storage/redis.py
```diff
@@ -8,14 +11,17 @@ class Redis(StorageHandler):
...
-            item = json.dumps(item)
+        item.pop(storage_model.key_field)
+
+        if len(item) == 1 and storage_model.expiry_field in item:
+            # Don't store a value if the only key that is not the key_field is the expiry_field
+            value = ""
+        else:
+            value = json.dumps(item)
 
+        expires_in = model.expires_at - datetime.now(tz=tzutc())
```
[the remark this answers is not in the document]
MebinAbraham: Done

app/storage/storage.py
MebinAbraham: I did this at first, then changed it so it doesn't have to provide both. But make sense to make it the responsibility of the caller therefore, have changed.

app/storage/storage.py
MebinAbraham: `StorageModel` now handles serialization and deserialization.

app/storage/redis.py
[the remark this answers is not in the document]
MebinAbraham: When the value is `""` the `if item` statement would evaluate false. If that is not desired, we can either handle `""` explicitly or store a truthy value, or store `{}` and handle that explicitly.

Thoughts?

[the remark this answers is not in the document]
ajmaddaford: I think that's fine.

app/storage/storage.py
ajmaddaford: I think we should drop model from the init, and pass it in the `serialize` call.

MebinAbraham: Have dropped model, but re-introduces very small duplication in the storage classes, have a look, should be fine.

app/routes/session.py
```diff
@@ -98,13 +98,16 @@ def login():
...
 
 def validate_jti(decrypted_token):
-    expires = datetime.utcfromtimestamp(decrypted_token["exp"]).replace(tzinfo=tzutc())
-    if expires < datetime.now(tz=tzutc()):
+    expires_at = datetime.utcfromtimestamp(decrypted_token["exp"]).replace(
+        tzinfo=tzutc()
+    )
+    jwt_expired = expires_at < datetime.now(tz=tzutc())
+    if jwt_expired:
         raise Unauthorized
```
JamesGardiner, line 106:
```suggestion
    if expires_at < datetime.now(tz=tzutc()):
        raise Unauthorized
```
jwt_expired is only used here and doesn't really add any context that isn't obvious from the boolean comparison

MebinAbraham:
Will merge once I get a ✅ from @pricem14pc

app/routes/session.py
MebinAbraham: As discussed, will keep as is for now for readability/consistency sakes.

PR #128, opened 2020-05-27
Use consistent spelling for serialize/deserialize

MebinAbraham:
Uses the American English spelling for serialize/deserialize to make the repo consistent. The spelling serialize was chosen over serialise due to the dominant use of `serialize` in python libs, docs and even most other programming languages.

Feel free to add your thoughts, main concern is making sure we are consistent.

### How to review 
Ensure everything still works as expected.

### Checklist


iwootten:
Isn't the use of 'serialize' dominant because it's primarily in libraries written by American authors? I would understand this choice if this project were for audiences other than those in the uk, but here we're apparently choosing to spell a method incorrectly. 

We're incredibly careful on the translation throughout our apps, why would we adopt a different approach for our codebase?

MebinAbraham:
> Isn't the use of 'serialize' dominant because it's primarily in libraries written by American authors? I would understand this choice if this project were for audiences other than those in the uk, but here we're apparently choosing to spell a method incorrectly.
> 
> We're incredibly careful on the translation throughout our apps, why would we adopt a different approach for our codebase?

It's not adopting a different approach when it is already existing in our code base. Given that most of the devs said they prefer `serialize` is why I have raised this PR like this. Personally, I think either is fine, as long as we are consistent.
We can discuss this in the upcoming tech session, and make a final decision.

iwootten:
> It's not adopting a different approach when it is already existing in our code base. Given that most of the devs said they prefer `serialize` is why I have raised this PR like this. Personally, I think either is fine, as long as we are consistent.

What I mean to say is, a different approach to being incredibly careful. The code that is spelt incorrectly should be corrected.

.python-version
```diff
@@ -1 +1 @@
-3.8.2
+3.8.3
```
iwootten, line 1: This needs removing given it's unavailable in pyenv at the moment

iwootten: Going back on this given it doesn't appear quite as simple as changing a tag: https://github.com/ONSdigital/eq-questionnaire-runner/pull/129

PR #132, opened 2020-05-29
Last viewed question guidance

LJBabbage:
Last question guidance should now appear on an all blocks if the section is in progress and the respondent is returning to it, whether it be from the hub or save and resume. If it's during a standard linear journey, then naturally it shouldn't appear. 

N.B. If you go directly to the page through the url it won't appear, as we are using an url param to make it visible 

### How to review 
After the normal code review, the best way would be to run the app and try every combination you can think of in regards to when last viewed question guidance  should and shouldn't appear


tests/integration/questionnaire/test_questionnaire_last_viewed_guidance.py
```diff
@@ -0,0 +1,286 @@
+import random
```
ajmaddaford, line 1: This file is almost identical to `test_questionnaire_last_viewed_guidance_hub.py`. Should this test file only have tests for the non-hub questionnaire and the hub test file have tests for the hub questionnaire?

tests/integration/questionnaire/test_questionnaire_last_viewed_guidance.py
```diff
@@ -0,0 +1,286 @@
...
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestLastViewedGuidance(IntegrationTestCase):
+    def test_last_viewed_question_guidance_not_shown_on_survey_launch(self):
+        # Given
+        self.launchSurvey("test_last_viewed_question_guidance")
+
+        # Then last viewed question guidance should not be shown
+        self._last_viewed_question_guidance_not_shown_assertion()
```
ajmaddaford, line 12: The method name should be consistent with the generic asserts and start with assert e.g. `_assert_last_viewed_question_guidance_not_shown`

tests/integration/questionnaire/test_questionnaire_last_viewed_guidance_hub.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def last_viewed_question_guidance_shown_assertion(self, link):
+        selector = "#section-start-link"
+        selected = self.getHtmlSoup().select(selector)
+        self.assertInBody("This is the last viewed question in this section")
+        self.assertEqual(selected[0].get("href"), link)
+
+    def last_viewed_question_guidance_not_shown_assertion(self):
+        self.assertNotInUrl("last_viewed_question_guidance=True")
+        self.assertNotInBody("This is the last viewed question in this section")
```
ajmaddaford, line 207: The method naming implies these methods do the opposite of each-other, but the first method also checks the link, and the second checks the url. It would be clearer if the link checking was in an assertion method of its own and the shown assertion included a check for the url.

Also, as these assertions are shared between the two tests we should consider how we can remove the duplication.

tests/integration/questionnaire/test_questionnaire_last_viewed_guidance_hub.py
```diff
@@ -0,0 +1,226 @@
...
+            "/questionnaire/household-interstitial/"
+        )
+
+    def test_last_viewed_question_guidance_shown_on_section_in_progress_resume_relationships(
+        self
+    ):
+
+        # Given
+        response_id = self.random_respondent_id()
+        self.launchSurvey("test_last_viewed_question_guidance", reponse_id=response_id)
```
ajmaddaford, line 179: It would be worth thinking about a common method that does the response id generation and returns the generated id.

tests/functional/spec/features/last_viewed_guidance/last_viewed_guidance.spec.js
```diff
@@ -0,0 +1,56 @@
+const JwtHelper = require("../../../jwt_helper");
+
+const AddressConfirmationPage = require('../../../generated_pages/last_viewed_question_guidance/address-confirmation.page.js');
+const HouseholdInterstitialPage = require('../../../generated_pages/last_viewed_question_guidance/household-interstitial.page.js');
+const PrimaryPersonListCollectorPage = require('../../../generated_pages/last_viewed_question_guidance/primary-person-list-collector.page.js');
+
+describe('Last viewed question guidance', function () {
+
+  const last_viewed_question_text = 'This is the last viewed question in this section'
+  const last_viewed_question_url_arg='last_viewed_question_guidance=True'
```
ajmaddaford, line 10: I don't think we should check the url arg in the functional tests. It is tested in the integration tests and doesn't add any value here.

tests/functional/spec/features/last_viewed_guidance/last_viewed_guidance.spec.js
```diff
@@ -0,0 +1,56 @@
+const JwtHelper = require("../../../jwt_helper");
+
+const AddressConfirmationPage = require('../../../generated_pages/last_viewed_question_guidance/address-confirmation.page.js');
+const HouseholdInterstitialPage = require('../../../generated_pages/last_viewed_question_guidance/household-interstitial.page.js');
+const PrimaryPersonListCollectorPage = require('../../../generated_pages/last_viewed_question_guidance/primary-person-list-collector.page.js');
+
+describe('Last viewed question guidance', function () {
+
+  const last_viewed_question_text = 'This is the last viewed question in this section'
```
ajmaddaford, line 9: We should be checking for the panel rather than the text. We've had issues before when we have tried to run functional tests against schemas whilst changing language.

app/views/handlers/question.py
```diff
@@ -87,6 +89,17 @@ def get_context(self):
...
         return context
 
+    def get_last_viewed_question_guidance(self):
+        if self.last_viewed_question_guidance:
+            first_location_in_section_url = self._router.get_first_location_in_section(
+                self._routing_path
+            ).url()
+            if self._current_location != first_location_in_section_url:
+                return {
+                    "show": True,
```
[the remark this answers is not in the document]
LJBabbage: Its for James, he likes stuff like that

app/views/handlers/list_collector.py
```diff
@@ -22,7 +22,9 @@ def get_next_location_url(self):
         return super().get_next_location_url()
 
     def get_context(self):
-        question_context = build_question_context(self.rendered_block, self.form)
+        question_context = build_question_context(
+            self.rendered_block, self.form, self.get_last_viewed_question_guidance()
+        )
```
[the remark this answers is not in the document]
LJBabbage: Interesting, all but list_collector doesn't actual need to override get_context from question. You actual get a small bonus with retun_to_hub_url, which was missing on all of them. Donna was going to raise a card to fix that, bagsie that one

tests/integration/questionnaire/test_questionnaire_last_viewed_guidance.py
LJBabbage: Something weird has happened here, as they are the same, I did write one for the hub, but somehow I have overridden it, fingers crossed my history is ok

LJBabbage: I see it whats happened, I was wondering how I got coverage. I copied the file, then decided to put it all in one, because it was sharing functionality, then didn't delete the copied file, doh.

app/views/handlers/list_collector.py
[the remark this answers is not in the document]
ajmaddaford: 😎

test_schemas/en/test_last_viewed_question_guidance.json
```diff
@@ -0,0 +1,820 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "0",
+    "title": "Test PrimaryPersonListCollector",
+    "theme": "default",
+    "description": "A questionnaire to test Primary Person ListCollector",
```
MebinAbraham, line 9: Should be made relevant to the what the schema tests.

app/views/handlers/block.py
```diff
@@ -29,6 +29,7 @@ def __init__(
         self.form = None
         self.page_title = None
         self._return_to_summary = self._request_args.get("return_to_summary")
+        self.resume = self._request_args.get("resume")
```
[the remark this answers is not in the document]
ajmaddaford: Good point. We should validate the query parameters and parse them into booleans.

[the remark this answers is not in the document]
ajmaddaford: We should also consider appropriately casing the query params i.e. `true` rather than `True`.

app/questionnaire/relationship_router.py
```diff
@@ -10,7 +10,9 @@ def __init__(self, section_id, block_id, list_item_ids):
     def can_access_location(self, location):
         return location in self.path
 
-    def get_first_location_url(self):
+    def get_first_location_url(self, resume=False):
+        if resume:
+            return self.path[0].url(resume=True)
```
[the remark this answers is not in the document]
LJBabbage: Its an kwarg which goes into url_for(), so needs both bits

app/views/handlers/relationship_collector.py
```diff
@@ -48,6 +48,8 @@ def is_location_valid(self):
         return True
 
     def get_first_location_url(self):
+        if self.resume:
+            return self.relationship_router.get_first_location_url(resume=True)
```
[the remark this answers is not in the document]
LJBabbage: As above

tests/functional/spec/features/last_viewed_guidance/last_viewed_guidance.spec.js
```diff
@@ -0,0 +1,51 @@
+const JwtHelper = require("../../../jwt_helper");
+
+const AddressConfirmationPage = require('../../../generated_pages/last_viewed_question_guidance/address-confirmation.page.js');
+const HouseholdInterstitialPage = require('../../../generated_pages/last_viewed_question_guidance/household-interstitial.page.js');
+const PrimaryPersonListCollectorPage = require('../../../generated_pages/last_viewed_question_guidance/primary-person-list-collector.page.js');
+
+describe('Last viewed question guidance', function () {
+
+  const responseId = JwtHelper.getRandomString(16);
```
ajmaddaford, line 9: Can we define the extra params for `openQuestionnaire` in one place, something like:
```javascript
  const resumableLaunchParams = {
    responseId: JwtHelper.getRandomString(16),
    userId: "test_user"
  };
```

tests/functional/spec/features/last_viewed_guidance/last_viewed_guidance.spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+  describe('Given a standard survey', function () {
```
ajmaddaford, line 11: I'm not sure what standard survey means. This would be clearer as "Given the last viewed question guidance questionnaire".

tests/functional/spec/features/last_viewed_guidance/last_viewed_guidance.spec.js
```diff
@@ -0,0 +1,51 @@
...
+      $(HouseholdInterstitialPage.submit()).click();
+      $(AddressConfirmationPage.saveSignOut()).click();
+      browser.openQuestionnaire('test_last_viewed_question_guidance.json', { userId: "test_user", responseId: responseId });
+      expect(browser.getUrl()).to.contain(AddressConfirmationPage.url());
+      expect($(AddressConfirmationPage.lastViewedQuestionGuidanceLink()).getAttribute('href')).to.contain(HouseholdInterstitialPage.url());
+      expect($(AddressConfirmationPage.lastViewedQuestionGuidance()).isExisting()).to.be.true;
+
+    });
+
+    it('When the respondent answers the next question and saves and continues, then last question guidance is not shown', function () {
```
ajmaddaford, line 38: This confused me, as I don't think `AddressConfirmationPage` is the __next__ question. This would be clearer if it was "When the respondent answers the question and saves and continues, then last question guidance is not shown on the next question".

tests/functional/spec/features/last_viewed_guidance/last_viewed_guidance_hub.spec.js
```diff
@@ -0,0 +1,112 @@
...
+        $(GcsesPage.yes()).click();
+        $(GcsesPage.submit()).click();
+        browser.url(HubPage.url());
+        $(HubPage.summaryRowLink(2)).click();
+        expect(browser.getUrl()).to.contain(ALevelsPage.url());
+        expect($(ALevelsPage.lastViewedQuestionGuidanceLink()).getAttribute('href')).to.contain(GcsesPage.url());
+        expect($(ALevelsPage.lastViewedQuestionGuidance()).isExisting()).to.be.true;
+      });
+
+    it('When the respondent selects a section which is complete , then last question guidance is not shown on the summary or any link clicked from the summary', function () {
```
ajmaddaford, line 73: "last question guidance is not shown on the summary" isn't tested.

app/views/handlers/block.py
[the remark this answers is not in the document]
LJBabbage: If we used to the inverse condition to control anything then I would agree (True do this, False do that) however we don't. We are basically using the presence of the param, at which point the value largely becomes redundant. Realistically we should just pass the param if the value means nothing, which I believe is still valid. One for tech session? should only take 5 minutes (more likely 20)

test_schemas/en/test_last_viewed_question_guidance.json
LJBabbage: You won't believe what happened here after I copied and pasted another schema to start off with ...

app/questionnaire/relationship_router.py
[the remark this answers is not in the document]
MebinAbraham: Sorry my point was it should use the var value, so if it kwargs:
```suggestion
            return self.path[0].url(resume=resume)
```

app/views/handlers/block.py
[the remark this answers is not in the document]
MebinAbraham: > We should also consider appropriately casing the query params i.e. `true` rather than `True`.

I also prefer this ☝️ 

Query params has to have a value, if the params didn't equal anything it would evaluate false when using `self._request_args.get()`.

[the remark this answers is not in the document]
LJBabbage: Surely you just then do a "resume" in self._request_args

app/questionnaire/relationship_router.py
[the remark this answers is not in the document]
LJBabbage: Don't really see the point as it can only be True, but if it makes you happy and saves more comments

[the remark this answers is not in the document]
MebinAbraham: Saves me having to explain it 👍

PR #134, opened 2020-06-01
Add optional warning panel to questions

ajmaddaford:

Adds the new optional warning panel to questions.

### How to review 

Use the `test_list_collector` questionnaire to check that the feature works. 

### Checklist

* [-] New static content marked up for translation

### DIFF
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=add-question-warning
 docker pull onsdigital/eq-questionnaire-validator:$tag
 docker run -d -p 5001:5000 "onsdigital/eq-questionnaire-validator:$tag"
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=add-question-warning
+tag=latest
 docker pull onsdigital/eq-questionnaire-validator:$tag
 docker run -d -p 5001:5000 "onsdigital/eq-questionnaire-validator:$tag"
--- a/templates/partials/question.html
+++ b/templates/partials/question.html
@@ -1,13 +1,12 @@
 {% from "components/question/_macro.njk" import onsQuestion %}
 {% from "components/fieldset/_macro.njk" import onsFieldset %}
+{% from "components/panel/_macro.njk" import onsPanel %}
 
 {% set form = content.form %}
 
 {% set question_title = question.title %}
 {% set question_description = format_paragraphs(question.description) if question.description else None %}
 {% set question_instruction = format_paragraphs(question.instruction) if question.instruction else None %}
-{% set question_guidance = "" %}
-{% set question_definition = "" %}
 {% set should_wrap_with_fieldset = should_wrap_with_fieldset(question) %}
 
 {%- set question_definition -%}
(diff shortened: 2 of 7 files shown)


tests/integration/questionnaire/test_questionnaire_list_collector.py
```diff
@@ -259,3 +259,12 @@ def test_cancel_text_displayed_on_edit_block_if_exists(self):
         self.get(change_link)
 
         self.assertInBody("Don’t need to change anything?")
+
+    def test_warning_text_displayed_on_remove_block_if_exists(self):
```
LJBabbage, line 263: I wonder if we need the "if_exists"? I don't believe you can even get to the remove block if the list item doesn't exists.

templates/partials/question.html
```diff
@@ -16,10 +15,20 @@
   {%- endif -%}
 {%- endset -%}
 
+{% set question_warning %}
+  {%- if question.warning -%}
+    {% call onsPanel({
+      "id": "question-warning-" + question.id,
```
MebinAbraham, line 21: The recommended way to concatenate string in Jinja is using a tilde. I know we use `+` in a lot of places but for new code, we should keep it as recommended.
```suggestion
      "id": "question-warning-" ~ question.id,
```

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=add-question-warning
```
```suggestion
tag=latest
```

tests/integration/questionnaire/test_questionnaire_list_collector.py
ajmaddaford: This was copied from the preceding tests. I read it as meaning "if the 'warning' property exists".

tests/integration/questionnaire/test_questionnaire_list_collector.py
```diff
@@ -259,3 +259,12 @@ def test_cancel_text_displayed_on_edit_block_if_exists(self):
...
+        self.launchSurvey("test_list_collector")
+        self.post(action="start_questionnaire")
+        self.post({"anyone-else": "Yes"})
+        self.add_person("Someone", "Else")
+        remove_link = self.get_link("1", "Remove")
+        self.get(remove_link)
+        self.assertInBody("All of the information about this person will be deleted")
```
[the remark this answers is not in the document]
ajmaddaford: Possibly, this was copied from the previous test. I'm not sure we consistently check for ids in any of our tests.

templates/partials/question.html
ajmaddaford: Ok. I have updated the other occurrence of this in this file as well.

tests/integration/questionnaire/test_questionnaire_list_collector.py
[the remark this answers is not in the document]
LJBabbage: The ID would come from the panel, meaning existence of a warning panel compared to just isolated text, and therefore more reassurance of correct formatting. As its also the question id you can check the warning and question go together. However saying that, I guess its not fundamentally needed, so shouldn't block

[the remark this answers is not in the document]
ajmaddaford: I've added a test for the panel.

PR #137, opened 2020-06-02
Update JS packages and use modern JS syntax (ES6+)

MebinAbraham:
- Updated our JS packages to the latest version where suitable.
- Removed python2.7 dependency for npm.
- Use consistent eslint config for both test dir and test dir. We had two eslint config files.
- Updated eslint rules to use modern JS, ie ES6+.
- Updated existing JS syntax to conform to ES6+. Not everything is done, more could still be done.
- Removed packages that we no longer used, these include some babel etc. that was needed when we hosted our own JS files.
- WebdriverIO has been updated from v5 to v6. In v6, wdio have introduced their own [assertion library](https://webdriver.io/blog/2020/03/26/webdriverio-v6-released.html#new-assertion-library) inspired from Jest's expect library, I have not installed that as part of this as we can keep using Chai (with a warning notice outputted to the terminal), but could move towards [Jest's expect](https://www.npmjs.com/package/expect) if we deemed appropriate.
- Updated certain parts of `generate_pages.py` so that the generated JS conformed to the new rules.
- Fixed wildcard expansion in yarn commands. Running `yarn lint` etc. would not function properly for all terminals due to unquoted wildcard operators which meant files were missed from linting.

There are further changes that could be made but wanted to keep the PR somewhat digestible. It looks like a lot of changes but primarily, its just syntax changes.
 
### How to review 
- Ensure all functional tests still run and behave as expected.
- Do you agree with the changes?. Are there things that could be made better?.

### Checklist


tests/functional/generate_pages.py
```diff
@@ -43,21 +43,21 @@
...
 """
 )
 
 SPEC_EXAMPLE_TEST = Template(
     r"""
-describe('Example Test', function() {
-  beforeEach('Load the survey', function() {
-    browser.openQuestionnaire('${schema}');
+describe("Example Test", () => {
+  beforeEach"Load the survey", () => {
```
iwootten, line 56: Is this how beforeEach is called in ES6?

.github/workflows/pull_request.yml
```diff
@@ -99,6 +99,8 @@ jobs:
         EQ_RUN_FUNCTIONAL_TESTS_HEADLESS: True
       steps:
         - uses: actions/checkout@v2
+        - name: Update to latest stable Chrome
+          run: sudo apt-get update && sudo apt-get upgrade google-chrome-stable -y
```
MebinAbraham, line 103: Given our functional test run against the version of Chrome installed on the host machines, it causes issues when developing/CI.

Functional test on master won't run, if you have updated your Chrome to version 83.
In this PR I have updated chrome driver to 83, but actions does not have the latest Chrome yet, only version 81 which is incompatible with Chromedrver 83. Therefore, we need to think of a way to ensure parity between dev and CI, whether it is using a docker image etc.

tests/functional/generate_pages.py
MebinAbraham: Fixed the bracket.

.eslintrc
```diff
@@ -1,40 +1,72 @@
 {
```
JamesGardiner, line 1: We should consider getting parity across eslintrc files in eq products

scripts/run_tests_functional.sh
```diff
@@ -21,8 +21,6 @@ function display_result {
 }
 
 # Run Functional tests
-npm config set python python2.7
```
JamesGardiner, line 24: 🤟

.eslintrc
```diff
@@ -1,40 +1,72 @@
...
+      2,
+      {
+        "SwitchCase": 1
+      }
+    ],
+    "comma-dangle": 0,
+    "new-cap": 2,
+    "no-alert": 1,
+    "no-console": 2,
+    "no-dupe-class-members": 0,  // Need to fix generate_pages.py to enable this.
```
JamesGardiner, line 42: Should track this comment as a `TODO`

.babelrc
```diff
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015", "stage-2"],
+  "presets": ["babel-preset-env"]
```
rmccar, line 2: I think this should be `"@babel/preset-env"`

tests/functional/spec/features/show_section_summary_on_completion/show_section_summary_on_completion.spec.js
```diff
@@ -1,58 +1,56 @@
...
+    before("Get to hub", () => {
+      browser.url(hubPage.url());
+    });
 
-        it('When I reach the end of that section, ' +
-            'then I will be taken to the section summary to enable me to amend an answer' , () => {
-            $(hubPage.summaryRowLink(2)).click();
-            $(proxyQuestionPage.noIMAnsweringForMyself()).click();
-            $(proxyQuestionPage.submit()).click();
+    it("When I reach the end of that section, " + "then I will be taken to the section summary to enable me to amend an answer", () => {
```
rmccar, line 36: The `" + "` should probably be removed here

tests/functional/spec/features/show_section_summary_on_completion/show_section_summary_on_completion.spec.js
```diff
@@ -1,58 +1,56 @@
...
-            browser.url(hubPage.url());
-        });
+  describe("Given I have completed a section with the summary turned on for the forward journey", () => {
+    before("Get to hub", () => {
+      browser.url(hubPage.url());
+    });
 
-        it('When I return to a completed section from the hub, then I am returned to the correct section summary' , () => {
-            $(hubPage.summaryRowLink(2)).click();
+    it("When I return to a completed section from the hub, then I am returned to the correct section summary", () => {
```
rmccar, line 50: Probably a bit nit picky but I've noticed most files have a capital `T` on `then` after a comma but theres a few files with a lower case `t`. I don't know if you want to standardise this

.babelrc
rmccar: Actually forget this, I've just realised you're not using babel 7

.eslintrc
MebinAbraham: They have different rules enabled where necessary, was there anything specific you meant?. Main difference I see is validator we use single quote, but we chose double a while ago, so that should change other repos.

tests/functional/spec/features/show_section_summary_on_completion/show_section_summary_on_completion.spec.js
MebinAbraham: Not sure, I think it should be capitalised for nicer output, but we don't seem to follow that pattern consistently. But I don't think I should change that as part of this PR without a way of enforcing it also.

rmccar: Yeah no worries, just something I noticed

.eslintrc
JamesGardiner: yeah I should have been clearer - where there are settings that are required across repos we should keep them consistent. The string quotations are a really good example of where we should be using the same settings

.github/workflows/pull_request.yml
MebinAbraham: Chrome is now installed from source, so the added time is negligible now which removes the gap between actions being out of date with the latest Chrome.

PR #142, opened 2020-06-09
Remove save on signout code

ajmaddaford:
Removes redundant save on signout code and tests. This hasn't been used since the signout button was changed to a link in https://github.com/ONSdigital/eq-survey-runner/pull/2277.

As part of this work a new `signed-out` endpoint has been added to differentiate it from `sign-out`. `sign-out` does the work of signing out and redirecting to either a provided log out url or the `signed-out` page.

The sign out button logic has been simplified to remove the check for an `account_service_log_out_url`. This was to hide the sign out button on the `thank-you` page. Instead there is now an explicit property `hide_signout_button` and cookie state modification on the GET of the `thank-you` page has been removed.

### How to review 
- Check that all code related to save on signout has been removed
- Check that tests that have been removed should have been

### Checklist


templates/layouts/configs/_save-sign-out-button.html
```diff
@@ -1,45 +1,37 @@
-{% if current_user.is_authenticated %}
-  {% if account_service_log_out_url %}
+{% if current_user.is_authenticated and not hide_signout_button %}
+
     {% if save_on_signout %}
```
MebinAbraham, line 3: When is this actually false?

ajmaddaford: It's not set on:
- the introduction template 
- any error pages where they are signed in e.g. 404s, 
- static pages

MebinAbraham: Ah ya, makes sense 👍

tests/integration/questionnaire/test_questionnaire_sign_out.py
```diff
@@ -0,0 +1,47 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestSignOut(IntegrationTestCase):
+    def test_sign_out_not_signed_in_redirects_to_signed_out_page(self):
```
LJBabbage, line 5: I reckon we can drop the prefix sign_out from some of these test as its class name. This one is a bit of a tongue twister and confused me initially with the 3 sign references

tests/integration/questionnaire/test_questionnaire_sign_out.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertInUrl("/signed-out")
+
+    def test_sign_out_with_log_out_url_redirects_to_log_out_url(self):
+        self.launchSurvey(
+            "test_textfield", account_service_log_out_url="https://localhost/logout"
+        )
+        self.get("/sign-out")
+        self.assertInUrl("/logout")
+
+    def test_sign_out_after_navigating_backwards(self):
```
LJBabbage, line 21: This seems a tests for resuming and not signing out

tests/integration/questionnaire/test_questionnaire_sign_out.py
```diff
@@ -0,0 +1,47 @@
...
+        # Go back to the first page
+        self.get("/questionnaire/name-block")
+
+        self.get("/sign-out")
+        self.launchSurvey("test_textfield")
+
+        # Check we are on the second page
+        self.assertEqual("/questionnaire/summary/", self.last_url)
+
+    def test_sign_out_on_section_summary(self):
```
LJBabbage, line 36: Same as above, there are other integration tests which already test this

tests/integration/questionnaire/test_questionnaire_sign_out.py
ajmaddaford: As discussed, I have split out the resume tests from the sign out tests.

tests/integration/questionnaire/test_questionnaire_sign_out.py
ajmaddaford: As discussed, there are no other integration tests for resuming.

PR #144, opened 2020-06-11
Primary list collector save resume bug

LJBabbage:
This PR fixes the bug when a respondent initially answers negatively to a PrimaryListCollector, then changes it to positive and signs out. The negative answer added the block into the progress store, but because we didn't remove it when the answer got changed, it was possible to change the answer to positive on the PrimaryListCollector but skip putting in the name by signing out or by changing the url to the next block. Resume(or changing the URL) would incorrectly go to the next block when it should go back to the PrimaryListCollector.

I have included a comment as it's an edge case and in the future its existence might be questioned. Feel free to come up with a better name for the test, my brain put up a white flag on that one

### How to review 
Confirm the fix rectifies the bug.


tests/integration/questionnaire/test_questionnaire_list_collector_primary_person.py
```diff
@@ -68,3 +69,21 @@ def test_cannot_remove_primary_person_from_list_collector(self):
...
+        response_id = random.choices(string.digits, k=16)
+
+        # Given I initially answer 'No' to the primary person list collector
+        self.launchSurvey("test_list_collector_primary_person", reponse_id=response_id)
+        self.post({"you-live-here": "No"})
+
+        # When I change my answer to 'Yes' and sign out
+        self.get("questionnaire/primary-person-list-collector/")
+        self.post({"you-live-here": "Yes"})
+        self.post(action="sign_out")
```
ajmaddaford, line 85: This should be `self.get("/sign-out")` now that the save on signout pr has merged.

PR #147, opened 2020-06-15
Replace chained selectors in runner tests with data attribute selectors

rmccar:
Replace chained selectors in runner tests with data attribute selectors 

### How to review 
- Tests Pass


rmccar:
[Replace chained selectors in runner tests with data attribute selectors (S)](https://trello.com/c/WTsgnNkW/3697-replace-chained-selectors-in-runner-tests-with-data-attribute-selectors-s)

tests/functional/base_pages/question.page.js
```diff
@@ -31,11 +31,11 @@ class QuestionPage extends BasePage {
   }
 
   errorHeader() {
-    return "#main-content > div.panel.panel--error.u-mb-s > div.panel__header > div";
+    return '[data-qa="error-header"]';
   }
 
   errorNumber(number = 1) {
-    return `[data-qa="error-body"] ul > li:nth-child(${number}) > a`;
+    return `[data-qa="list-item-${number}"]`;
```
iwootten, line 38: I don't understand why this is being changed, the original is based off a data-qa selector?

tests/functional/generate_pages.py
```diff
@@ -101,7 +101,7 @@
 
 ANSWER_LABEL_GETTER = Template(
     r"""  ${answerName}Label() {
-    return `[for=${answerId}]`;
+    return `#${answerId}-label`;
```
iwootten: I don't see how adding the id helps here? The original seems more concise and we didn't have an issue with answer label chains.

tests/functional/generate_pages.py
```diff
@@ -193,7 +195,7 @@
 """
 
 CALCULATED_SUMMARY_LABEL_GETTER = Template(
-    r"""  ${answerName}Label() { return `[data-qa=${answerId}-label]`; }
+    r"""  ${answerName}Label() { return `#{answerId}-label`; }
```
iwootten, line 198: This is now based off a specific id, rather than a data-qa selector (which is what the card specified to move to).

tests/functional/base_pages/question.page.js
rmccar: Ive already changed this back to a data-qa attribute

tests/functional/generate_pages.py
rmccar: Im making changes in here because I needed to update to version 18 of the Design System and there are some changes to labels in there, not just because of the chained selector stuff. As part of this I have also added ids to all the labels, so thought that was an easy and specific way to select them, now that they have ids, rather than using the `for` attribute. Having said that I'm still working on this in order to get the functional tests to pass.

tests/functional/generate_pages.py
rmccar: See previous comment

tests/functional/generate_pages.py
iwootten: Does v18 of the design system no longer allow selection using `for`?

iwootten: I kind of feel that keeping the use of `for` and not having to specify the id across all answer templates/rendered pages is the preferable approach here, especially as we're only really making changes for the purposes of our tests.

rmccar: Ok no worries ill change it back

app/translations/messages.pot
```diff
@@ -567,8 +567,8 @@ msgstr ""
 
 #: templates/partials/last_viewed_question_guidance.html:10
 msgid ""
-"You can also <a id='section-start-link' href='{url}'>go back to the start"
-" of the section</a>"
+"You can also <a id='section-start-link' data-qa='last-viewed-question-"
+"guidance-link' href='{url}'>go back to the start of the section</a>"
```
ajmaddaford, line 571: Is the `data-qa` property needed here if there is a unique id?

rmccar: Yeah I can change it to use the ID here

templates/question.html
```diff
@@ -23,7 +23,7 @@
       <ul class="list list--bare">
         {% for error_id, error in form.mapped_errors %}
           <li class="list__item">
-            {{ loop.index }}) <a class="js-inpagelink" href="#{{error_id}}">{{ error }}</a>
+            {{ loop.index }}) <a class="js-inpagelink" data-qa="list-item-{{ loop.index }}" href="#{{error_id}}">{{ error }}</a>
```
ajmaddaford, line 26: `list-item-x` is too generic, this should be something like `error-link-x`

tests/functional/base_pages/hub.page.js
```diff
@@ -18,11 +18,11 @@ class HubPage extends BasePage {
   }
 
   errorHeader() {
-    return "#main-content > div.panel.panel--error.u-mb-s > div.panel__header > div";
+    return '[data-qa="error-header"]';
   }
 
   errorNumber(number = 1) {
-    return `[data-qa="error-body"] ul > li:nth-child(${number}) > a`;
+    return `[data-qa="list-item-${number}"]`;
```
ajmaddaford, line 25: There are a number of selectors that use `nth-child`. We should be consistent in how we reference elements that can be repeated.

rmccar: So do you think I should change all the nth-child selectors to use data-qa attributes too?

rmccar:
> ove the use of CSS classes to locate elem

Yeah I think thats a good idea

tests/functional/base_pages/hub.page.js
ajmaddaford: I think so, using an example:
```
tbody:nth-child(${instance}) td:last-child a[data-qa='change-item-link']
```

would be something like:
```
a[data-qa='change-item-link-${instance}']
```

Which is simpler and less fragile, as it's not dependent on that element being in a `td`, in a `tbody`.

app/jinja_filters.py
```diff
@@ -574,7 +575,7 @@ def map_list_collector_config(
                     "text": edit_link_text,
                     "ariaLabel": edit_link_aria_label.format(item_name=item_name),
                     "url": list_item.get("edit_link"),
-                    "attributes": {"data-qa": "change-item-link"},
+                    "attributes": {"data-qa": "change-item-link-" + str(item_index)},
```
MebinAbraham, line 578: We should use f-strings where we can.
```suggestion
                    "attributes": {"data-qa": f"change-item-link-{item_index}"},
```

app/jinja_filters.py
```diff
@@ -584,12 +585,23 @@ def map_list_collector_config(
                     "text": remove_link_text,
                     "ariaLabel": remove_link_aria_label.format(item_name=item_name),
                     "url": list_item.get("remove_link"),
-                    "attributes": {"data-qa": "remove-item-link"},
+                    "attributes": {"data-qa": "remove-item-link-" + str(item_index)},
```
MebinAbraham, line 588:
```suggestion
                    "attributes": {"data-qa": f"remove-item-link-{item_index}"},
```

app/jinja_filters.py
```diff
@@ -584,12 +585,23 @@ def map_list_collector_config(
...
         rows.append(
-            {"rowTitle": item_name, "rowItems": [{"icon": icon, "actions": actions}]}
+            {
+                "rowTitle": item_name,
+                "rowItems": [
+                    {
+                        "icon": icon,
+                        "actions": actions,
+                        "rowTitleAttributes": {
+                            "data-qa": "list-item-label-" + str(item_index)
```
MebinAbraham, line 600:
```suggestion
                            "data-qa": f"list-item-label-{item_index}"
```

app/views/contexts/hub_context.py
```diff
@@ -66,14 +66,16 @@ def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
...
+        self, section_name: str, section_status: str, section_id: str, section_url: str
     ) -> Mapping[str, Union[str, List]]:
 
         section_content = self.SECTION_CONTENT_STATES[section_status]
         context: Mapping = {
             "rowTitle": section_name,
             "rowItems": [
                 {
+                    "rowTitleAttributes": {"data-qa": "hub-row-title-" + section_id},
+                    "attributes": {"data-qa": "hub-row-state-" + section_id},
```
MebinAbraham, line 78: Should convert all the string concatenation done using `+` to use f-strings.

app/views/contexts/hub_context.py
```diff
@@ -136,12 +146,17 @@ def _get_rows(self, enabled_section_ids) -> List[Mapping[str, Union[str, List]]]
...
                             self._get_row_for_repeating_section(
-                                section_id, list_item_id
+                                section_id, list_item_id, list_item_index
                             )
                         )
                 else:
-                    rows.append(self._get_row_for_section(section_title, section_id))
+                    rows.append(
+                        self._get_row_for_section(section_title, section_id, section_id)
+                    )
```
MebinAbraham, line 160: I don't think we should do any this, I find this really confusing as we are passing section ID twice, which may differ if it is repeating etc.

I think we should keep all this as is, and pass in the list_item_id to `get_row_context_for_section` which can call a method to return an appropriate title id based of whether there is a list item id and whether it is repeating. It means, we have to call `self._schema.get_repeating_list_for_section(section_id)` an extra time but given these would be cached in future, the performance should be negligible.

tests/app/questionnaire/test_questionnaire_schema.py
```diff
@@ -288,7 +288,7 @@ def test_get_list_item_id_for_answer_id_without_list_item_id(
 
 
 def test_get_list_item_id_for_answer_id_without_repeat_or_list_collector(
-    question_schema
+    question_schema,
```
iwootten, line 291: Why are all these commas being added?

rmccar: These were added when I ran black

iwootten: I think you need to remove the global install of black on your system and reinstall with `pipenv install --dev` as per runners readme. The pr shouldn't be reformatting these as they haven't otherwise been changed.

rmccar: OK ill do that

app/views/contexts/hub_context.py
rmccar: Yeah I get that passing section ID in twice looks a bit odd, it just allowed me to conveniently use it for two different things. While still allowing `get_row_context_for_section` to be used for sections and repeating sections.

app/views/contexts/hub_context.py
```diff
@@ -107,28 +110,35 @@ def get_section_url(section_id, list_item_id) -> str:
...
-        return self._get_row_for_section(title, section_id, list_item_id)
+        repeating_section_id = section_id + "-" + str(list_item_index)
 
-    def _get_row_for_section(self, section_title, section_id, list_item_id=None):
+        return self._get_row_for_section(
+            title, repeating_section_id, section_id, list_item_id
+        )
+
+    def _get_row_for_section(
+        self, section_title, section_id, section_title_id, list_item_id=None
```
ajmaddaford, line 127: What is being passed here as `section_title_id` should be reverted to be passed as `section_id`, as that is what it is. The new parameter being passed should be simply `list_index` as it is elsewhere, and the concatenation with `section_id` should happen in `get_row_context_for_section`.

app/views/contexts/hub_context.py
ajmaddaford: My comment above about passing `list_index` should resolve this.

templates/hub.html
```diff
@@ -10,8 +10,9 @@
     <h1 class="u-mt-m">{{ content.title }}</h1>
     {% call onsPanel({
       "type": "info",
-      "attributes": {
-      }
+        "attributes": {
+          "data-qa": "displayed-guidance"
```
ajmaddaford, line 14: It would be clearer if this was named `hub-guidance-panel`.

templates/hub.html
```diff
@@ -10,8 +10,9 @@
...
+        }
     }) %}
         <p data-qa="guidance"><strong>{{ content.guidance }}</strong></p>
```
ajmaddaford, line 17: This should be something like `hub-guidance-text`

templates/partials/answer-guidance.html
```diff
@@ -1,13 +1,18 @@
 {% from "components/collapsible/_macro.njk" import onsCollapsible %}
 
 {% call onsCollapsible({
+  "id": answer.id,
```
ajmaddaford, line 4: I think this will lead to id clashes as the answer will be using this id.

templates/partials/question-definition.html
```diff
@@ -2,13 +2,18 @@
 
 {% for definition in question.definitions %}
   {% call onsCollapsible({
+      "id": question.id ~ "-collapsible",
```
ajmaddaford, line 5: I think this will lead to id clashes as there can be more than one definition. Also, to be consistent with the attributes below, this should be `"question-definition-" ~ loop.index`.

tests/functional/base_pages/hub.page.js
```diff
@@ -57,16 +57,16 @@ class HubPage extends BasePage {
...
-  summaryRowLink(number = 1) {
-    return `tbody:nth-child(${number}) tr td.summary__actions a`;
+  summaryRowLink(sectionId = "section-1") {
+    return `[data-qa="hub-row-link-${sectionId}"]`;
   }
 
-  summaryRowTitle(number = 1) {
-    return `tbody:nth-child(${number}) tr td.summary__item-title`;
+  summaryRowTitle(sectionId = "section-1") {
+    return `[data-qa="hub-row-title-${sectionId}"]`;
```
ajmaddaford, line 69: It doesn't make sense to have a default of `section-1` for these.

tests/functional/generate_pages.py
```diff
@@ -229,13 +229,13 @@
...
 LIST_SECTION_SUMMARY_EDIT_LINK_GETTER = Template(
-    r"""  ${list_name}ListEditLink(listItemInstance) { return `div[data-qa="${list_name}-list-summary"] tbody:nth-child(` + listItemInstance + `) td:last-child a[data-qa="change-item-link"]`; }
+    r"""  ${list_name}ListEditLink(listItemInstance) { return `div[data-qa="${list_name}-list-summary"] a[data-qa="change-item-link-` + listItemInstance + `"]`; }
 
 """
 )
 
 LIST_SECTION_SUMMARY_REMOVE_LINK_GETTER = Template(
-    r"""  ${list_name}ListRemoveLink(listItemInstance) { return `div[data-qa="${list_name}-list-summary"] tbody:nth-child(` + listItemInstance + `) td:last-child a[data-qa="remove-item-link"]`; }
+    r"""  ${list_name}ListRemoveLink(listItemInstance) { return `div[data-qa="${list_name}-list-summary"] a[data-qa="remove-item-link-` + listItemInstance + `"]`; }
```
ajmaddaford, line 238: I think these should be consistent, so:
- `list-item-label-`
- `list-item-change-link-`
- `list-item-remove-link-`

tests/functional/helpers.js
```diff
@@ -1,8 +1,8 @@
 const checkPeopleInList = (peopleExpected, listLabel) => {
-  $(listLabel(1)).waitForDisplayed();
+  $(listLabel(0)).waitForDisplayed();
 
   for (let i = 1; i <= peopleExpected.length; i++) {
-    expect($(listLabel(i)).getText()).to.equal(peopleExpected[i - 1]);
+    expect($(listLabel(i - 1)).getText()).to.equal(peopleExpected[i - 1]);
```
ajmaddaford, line 5: ids and data-qa attributes for multiple elements should all start at an index of 1.

tests/functional/base_pages/summary.page.js
```diff
@@ -5,8 +5,8 @@ class ThankYouPage extends QuestionPage {
     super("summary");
   }
 
-  summaryRowState(number = 1) {
-    return `tbody:nth-child(${number}) tr td.summary__values`;
+  summaryRowState(sectionId = "section-1") {
```
ajmaddaford, line 8: Same comment as above, the default should be removed.

app/jinja_filters.py
```diff
@@ -584,12 +585,23 @@ def map_list_collector_config(
...
 
         rows.append(
-            {"rowTitle": item_name, "rowItems": [{"icon": icon, "actions": actions}]}
+            {
+                "rowTitle": item_name,
+                "rowItems": [
+                    {
+                        "icon": icon,
+                        "actions": actions,
+                        "rowTitleAttributes": {
```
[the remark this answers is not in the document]
rmccar: I did originally change `rowTitleAttributes` to be the same level as `rowTitle` but that broke some functionality because the data-qa being set in `rowTitleAttributes` was being passed `rowItems` and not `row`. If it is at the same level as `rowTitle` it doesn't have access to it. Maybe it should be renamed? or we would have to do a refactor so that it is able to deal with the whole row inside the bit where the data-qa gets set?

app/views/contexts/hub_context.py
```diff
@@ -139,12 +149,17 @@ def _get_rows(self, enabled_section_ids) -> List[Mapping[str, Union[str, List]]]
 
                 if repeating_list:
                     for list_item_id in self._list_store[repeating_list].items:
+
+                        list_item_index = len(rows)
```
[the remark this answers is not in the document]
rmccar: So are you saying I shouldn't be using `len` here because it uses 0 based numbering?

rmccar:
[Replace chained selectors in runner tests with data attribute selectors (S)](https://trello.com/c/WTsgnNkW/3697-replace-chained-selectors-in-runner-tests-with-data-attribute-selectors-s)

app/jinja_filters.py
```diff
@@ -565,6 +565,7 @@ def map_list_collector_config(
 
     for list_item in list_items:
         item_name = list_item.get("item_title")
+        item_index = list_items.index(list_item) + 1
```
MebinAbraham, line 568: This could be:
```py
    for index, list_item in enumerate(list_items, start=1):
        item_name = list_item.get("item_title")
```

rmccar: So would you then use `index` instead of `item_index` to set the data-qa?

app/jinja_filters.py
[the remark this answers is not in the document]
ajmaddaford: Should `rowTitle` be set at the same level as `rowTitleAttributes` here? Looking at the design system template, it seems to allow for that.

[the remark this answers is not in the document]
rmccar: Yeah just checked, fair enough, ill do that then

app/views/contexts/hub_context.py
```diff
@@ -158,10 +170,15 @@ def _get_rows(self, enabled_section_ids) -> List[Mapping[str, Union[str, List]]]
                 repeating_list = self._schema.get_repeating_list_for_section(section_id)
 
                 if repeating_list:
-                    for list_item_id in self._list_store[repeating_list]:
+                    for index, list_item_id in enumerate(
+                        self._list_store[repeating_list].items, start=1
+                    ):
+
+                        list_item_index = index
```
ajmaddaford, line 177:
```suggestion
                    for list_item_index, list_item_id in enumerate(
                        self._list_store[repeating_list].items, start=1
                    ):
```

app/views/contexts/hub_context.py
```diff
@@ -127,28 +129,38 @@ def get_section_url(section_id, list_item_id) -> str:
...
+            title, section_id, list_item_index, list_item_id
+        )
+
+    def _get_row_for_section(
+        self, section_title, section_id, list_item_index=None, list_item_id=None
+    ):
+        if list_item_index is not None:
+            full_row_id = section_id + "-" + str(list_item_index)
+        else:
+            full_row_id = section_id
```
ajmaddaford, line 149:
```suggestion
        full_row_id = f"{section_id}-{list_item_index}" if list_item_index else section_id
```

I also think `full_row_id` should be renamed to `row_id` as full doesn't add anything here.

app/views/contexts/hub_context.py
```diff
@@ -127,28 +129,38 @@ def get_section_url(section_id, list_item_id) -> str:
...
             repeating_title, list_item_id
         )
 
-        return self._get_row_for_section(title, section_id, list_item_id)
+        return self._get_row_for_section(
+            title, section_id, list_item_index, list_item_id
+        )
+
+    def _get_row_for_section(
+        self, section_title, section_id, list_item_index=None, list_item_id=None
```
ajmaddaford, line 144: The new variable should be added to the end of the argument list.
```suggestion
        self, section_title, section_id, list_item_id=None, list_item_index=None
```

app/views/contexts/hub_context.py
```diff
@@ -86,13 +86,15 @@ def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
         }
 
     def get_row_context_for_section(
-        self, section_name: str, section_status: str, section_url: str
+        self, section_name: str, section_status: str, section_id: str, section_url: str
```
ajmaddaford, line 89: The new argument should be added to the end of the argument list. We should also name it `row_id` rather than `section_id`, as `section_id` means something else in other places.

app/views/contexts/hub_context.py
```diff
@@ -101,11 +103,11 @@ def get_row_context_for_section(
                                 section_name=section_name
                             ),
                             "url": section_url,
-                            "attributes": {"data-qa": "summary-actions-section-link"},
+                            "attributes": {"data-qa": f"hub-row-link-{section_id}"},
```
ajmaddaford, line 106: It would be better to move the id here so that the prefix of the attribute values were the same i.e. `hub-row-{row_id}-title`, `hub-row-{row_id}-state` and `hub-row-{row_id}-link`.

templates/partials/answer-guidance.html
```diff
@@ -1,13 +1,18 @@
...
   "headingAttributes": {
     "data-ga": "click",
     "data-ga-category": "Answer guidance",
     "data-ga-action": "Open panel",
-    "data-ga-label": _(answer_guidance.schema_item.show_guidance)
+    "data-ga-label": _(answer_guidance.schema_item.show_guidance),
+    "data-qa": "answer-guidance-title"
+  },
+  "contentAttributes": {
+    "data-qa": "answer-guidance-content"
```
ajmaddaford, line 15: As there can be more than one answer, this will need to use the answer id as well. 
```suggestion
    "data-qa": "answer-guidance-" ~ answer.id ~ "-content"
```

templates/partials/answer-guidance.html
```diff
@@ -16,7 +21,8 @@
       "data-ga": "click",
       "data-ga-category": "Answer guidance",
       "data-ga-action": "Close panel",
-      "data-ga-label": _(answer_guidance.schema_item.show_guidance)
+      "data-ga-label": _(answer_guidance.schema_item.show_guidance),
+      "data-qa": "answer-guidance-button"
```
ajmaddaford, line 25:
```suggestion
      "data-qa": "answer-guidance-" ~ answer.id ~ "-button"
```

templates/partials/answer-guidance.html
```diff
@@ -1,13 +1,18 @@
...
+  "id": "answer-guidance-" ~ answer.id,
   "classes": "u-mt-s",
   "title": _(answer_guidance.schema_item.show_guidance),
   "headingAttributes": {
     "data-ga": "click",
     "data-ga-category": "Answer guidance",
     "data-ga-action": "Open panel",
-    "data-ga-label": _(answer_guidance.schema_item.show_guidance)
+    "data-ga-label": _(answer_guidance.schema_item.show_guidance),
+    "data-qa": "answer-guidance-title"
```
ajmaddaford, line 12:
```suggestion
    "data-qa": "answer-guidance-" ~ answer.id ~ "-title"
```

templates/partials/question-definition.html
```diff
@@ -17,7 +22,8 @@
           "data-ga": "click",
           "data-ga-category": "Question definition",
           "data-ga-action": "Close panel",
-          "data-ga-label": definition.title
+          "data-ga-label": definition.title,
+          "data-qa": "question-definition-button-" ~ loop.index
```
ajmaddaford, line 26: The prefixes for these attributes should be consistent, so this would be:
```suggestion
          "data-qa": "question-definition-"  ~ loop.index ~ "-button"
```

app/jinja_filters.py
```diff
@@ -584,12 +584,21 @@ def map_list_collector_config(
...
 
         rows.append(
-            {"rowTitle": item_name, "rowItems": [{"icon": icon, "actions": actions}]}
+            {
+                "rowItems": [
+                    {
+                        "icon": icon,
+                        "actions": actions,
+                        "rowTitle": item_name,
+                        "rowTitleAttributes": {"data-qa": f"list-item-label-{index}"},
```
ajmaddaford, line 598: The index should be just after `list-item` for these three attributes.

PR #148, opened 2020-06-16
Turn HTML Minification Off / Enable Whitespace Removal

iwootten:

The results of our performance investigation show that html minification has a significant effect on runner's response time. This PR defaults the minification in runner to false and enables whitespace removal through use of jinja flags.

### How to review 

Check tests continue to pass. Review performance investigation results here: https://github.com/ONSdigital/eq-survey-runner-benchmark/tree/use-non-minified-html


iwootten:
[Fix broken relationships when using raw HTML](https://trello.com/c/5ReW1Av1/3950-fix-broken-relationships-when-using-raw-html)

app/jinja_filters.py
```diff
@@ -288,9 +289,11 @@ def __init__(self, option, index, answer):
             # the 'pre-' prefix is added to the attributes here so that html minification
             # doesn't mess with the attribute contents (the 'pre-' is removed during minification).
             # see https://htmlmin.readthedocs.io/en/latest/quickstart.html
+            attribute_key = "pre-" if settings.EQ_ENABLE_HTML_MINIFY else ""
```
MebinAbraham, line 292: Should we not be getting the config from current app config like we do elsewhere instead of accessing in directly?

app/setup.py
```diff
@@ -111,6 +111,8 @@ def create_app(  # noqa: C901  pylint: disable=too-complex, too-many-statements
     application = Flask(__name__, template_folder="../templates")
     application.config.from_object(settings)
 
+    application.jinja_env.trim_blocks = True
+    application.jinja_env.lstrip_blocks = True
```
ajmaddaford, line 115: We should consider grouping these with the other jinja environment changes from line 191, possibly in a `setup_jinja_env` method.

k8s/helm/templates/deployment.yaml
```diff
@@ -61,6 +61,8 @@ spec:
               value: /keys/keys.yml
             - name: EQ_RABBITMQ_ENABLED
               value: "False"
+            - name: EQ_ENABLE_HTML_MINIFY
+              value: "False"
```
MebinAbraham, line 65: PR says default minification in runner to False, but this is only done on deployed envs atm. `EQ_ENABLE_HTML_MINIFY` in `settings.py` should be changed to `"False"` as the default. Which would mean this change could also be removed as it doesn't override the default.

iwootten: The reason it is disabled is to boost performance for census. If we were to disable it by default, we may as well remove it entirely.

ajmaddaford: The intent was to get it merged, see the performance impact on the daily test, then decide whether to remove it completely in a separate card.

MebinAbraham: Sounds good 👍

PR #151, opened 2020-06-17
Create submission handler

LJBabbage:
The code for submitting should not be in questionnaire.py and should be in its own handler, this PR creates the hander

### How to review 
Make sure everything works as it did before


app/views/handlers/submission.py
```diff
@@ -0,0 +1,49 @@
...
+        self.questionnaire_store = questionnaire_store
+        self.full_routing_path = full_routing_path
+
+    def submit_message(self):
+        message = json.dumps(
+            convert_answers(
+                self.schema, self.questionnaire_store, self.full_routing_path
+            ),
+            for_json=True,
+        )
```
iwootten, line 23: Given that the schema and routing_path are only used to generate the message, it feels like a more suitable interface might be for the handler to be initialised only with metadata and handler.submit_message() to take the message as an arg, rather than fetch it.

app/routes/questionnaire.py
```diff
@@ -306,37 +301,12 @@ def _generate_wtf_form(block_schema, schema, current_location):
...
-
-    return redirect(url_for("post_submission.get_thank_you"))
-
-
-def _store_submitted_time_in_session(submitted_time):
-    session_store = get_session_store()
-    session_data = session_store.session_data
-    session_data.submitted_time = submitted_time.isoformat()
-    session_store.save()
+    return redirect(submission_handler.get_next_location_url())
```
ajmaddaford, line 313: The `get_next_location_url` seems unnecessary given that it's always the same url. It's clearer to see what this is by using `return redirect(url_for("post_submission.get_thank_you"))` here.

app/views/handlers/submission.py
LJBabbage: You're right, that does make sense. The handler shouldn't be going to get the message. Thanks

app/routes/questionnaire.py
```diff
@@ -306,37 +301,12 @@ def _generate_wtf_form(block_schema, schema, current_location):
...
-        tx_id=metadata.get("tx_id"),
-    )
-
-    if not sent:
-        raise SubmissionFailedException()
-
-    submitted_time = datetime.utcnow()
-    _store_submitted_time_in_session(submitted_time)
+    submission_handler.submit_message()
     get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
```
ajmaddaford, line 311: I think this should either be done in the `submit_message` method, or another method created on the handler e.g. `delete_questionnaire_state`.

app/views/handlers/submission.py
ajmaddaford: I don't think so, as we're trying to move logic away from the questionnaire routes. Maybe it would be clearer if this was named `submit_questionnaire`?

app/routes/questionnaire.py
MebinAbraham: If the above is done, I think this `submit_answers` should be scrapped as it doesn't provide much over making an instance of `SubmissionHandler` in the routes. This would be if the message was handled by submission as per https://github.com/ONSdigital/eq-questionnaire-runner/pull/151#discussion_r442030960

LJBabbage: I kept the delete state out of the handler as I considered them to be distinct things

LJBabbage: Not sure about duplication of the handler in hub/block submissions

app/routes/questionnaire.py
```diff
@@ -302,15 +301,11 @@ def _generate_wtf_form(block_schema, schema, current_location):
...
+        schema, questionnaire_store, full_routing_path
     )
-    submission_handler = SubmissionHandler(questionnaire_store.metadata)
-    submission_handler.submit_message(message)
-
-    get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
-
-    return redirect(submission_handler.get_next_location_url())
+    submission_handler.submit_questionnaire()
+    return redirect(url_for("post_submission.get_thank_you"))
```
MebinAbraham: I still think it would be clearer to instantiate these in the route, as it not much duplication. But I if we are happy keeping it like this, I think the wrapper method should be renamed to `submit_questionnaire` or `handle_submission` since it is both submitting and redirecting.

app/views/handlers/submission.py
```diff
@@ -0,0 +1,47 @@
...
+            questionnaire_id=metadata.get("questionnaire_id"),
+            case_id=metadata.get("case_id"),
+            tx_id=metadata.get("tx_id"),
+        )
+
+        if not submitted:
+            raise SubmissionFailedException()
+
+        self._store_submitted_time_in_session()
+        get_questionnaire_store(current_user.user_id, current_user.user_ik).delete()
```
ajmaddaford, line 40: Can this just be `self._questionnaire_store.delete()`?

app/routes/questionnaire.py
LJBabbage: I actually agree, and did do it separately initially as I felt the same. I will revert as you have asked twice and there is a chance you might remind me in the future when I have long forgotten

MebinAbraham: Waiting for @ajmaddaford to comment to revert :trollface:

PR #152, opened 2020-06-17
Refactor form generation into block handler

ajmaddaford:
Moves the form generation into the block handler. As well as the block handler being the more appropriate place for the form generation, this also contributes to the aim of removing logic from the top level route file.

### How to review 
- Do you agree with the implementation?
- Should further refactoring be done?

### Checklist


app/views/handlers/block.py
```diff
@@ -114,3 +109,10 @@ def _update_section_completeness(self, location: Optional[Location] = None):
             section_id=location.section_id,
             list_item_id=location.list_item_id,
         )
+
+    def _set_started_at_metadata(self):
+        collection_metadata = self._questionnaire_store.collection_metadata
+        if not collection_metadata.get("started_at"):
+            started_at = datetime.utcnow().isoformat()
+            logger.info("Survey started", started_at=started_at)
+            collection_metadata["started_at"] = started_at
```
iwootten, line 118: Mutating the metadata in this way doesn't seem great. Can't started_at be set when the collection_metadata is first created?

ajmaddaford: This isn't metadata, this is collection metadata that was created for this purpose. It's done this way so that it is only set once they have POSTed a page. This also isn't new, it was previously called from the route.

app/forms/questionnaire_form.py
```diff
@@ -377,21 +374,43 @@ def _get_error_id(answer_id):
     return f"{answer_id}-error"
 
 
+def _clear_detail_answer_field(form_data, question):
+    """
+    Checks the submitted answers and in the case of both checkboxes and radios,
```
LJBabbage, line 379: I found the comment a bit hard to follow and a little verbose. Isn't it just "clear detail_answer if parent option not selected"? Also I am not sure we need the params, nothing else in that file states them and they were obvious to me

app/views/handlers/question.py
```diff
@@ -108,9 +139,9 @@ def get_last_viewed_question_guidance_context(self):
             return {"first_location_in_section_url": first_location_in_section_url}
 
     def handle_post(self):
-        self.questionnaire_store_updater.update_answers(self.form)
-
+        self.questionnaire_store_updater.update_answers(self.form.data)
         self.questionnaire_store_updater.add_completed_location()
+        self._set_started_at_metadata()
```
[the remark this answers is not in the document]
ajmaddaford: Yes, if `handle_post` is called it means something has been POSTed and there were no validation errors.

Good spot on the consistency between the two handlers. There was duplication in the `Question` handlers `handle_post` method, so I have removed that and added a call to the parent block handers `handle_post` method.

PR #154, opened 2020-06-20
Custom section summary title

LJBabbage:
The requirement is to allow a title with placeholders to be set for section summaries

The validator branch which goes with this is add-title-to-section-summary

### How to review 
Run the app and make sure the summary takes the right title, priority is as follows

1. Title set in repeating block
2. Title set in section summary
3. Section title


app/views/contexts/section_summary_context.py
```diff
@@ -65,16 +65,25 @@ def _build_summary(self, location):
...
-            )
+        section_id = location.section_id
+        list_item_id = location.list_item_id
+
+        if list_item_id:
+            repeating_title = self._schema.get_repeating_title_for_section(section_id)
             if repeating_title:
-                title = self._placeholder_renderer.render_placeholder(
-                    repeating_title, location.list_item_id
+                return self._placeholder_renderer.render_placeholder(
```
LJBabbage, line 74: This is a pain, and why I can't get this block a bit neater. For section summary titles I must check for a placeholder as it might not have one, however repeating is assuming one.

Although completely logical to assume a placeholder for repeating, how validator is currently written it is possible to have a valid schema without a placeholder in repeating meaning this will blow up.

MebinAbraham: I find it confusing that we have multiple methods to get a section title, we have `get_repeating_title_for_section`, `get_summary_title_for_section` and `get_title_for_section`.
I think we should have a single method that return the title based on the priority, so:
```python
    def get_title_for_section(self, section_id):
        return (
            self.get_repeating_title_for_section(section_id)
            or self.get_summary_title_for_section(section_id)
            or self._sections_by_id.get(section_id).get("title")
        )
```

Doing this would mean, we can simplify the method above to:
```python
    def _title_for_location(self, location):
        title = self._schema.get_title_for_section(location.section_id)
        if isinstance(title, dict):
            return self._placeholder_renderer.render_placeholder(
                title, location.list_item_id
            )

        return title
```
You could additionally add `placeholders in `title` but at the moment, title can either be a string or a dict, and a dict implies a placeholder.

app/views/contexts/section_summary_context.py
```diff
@@ -65,16 +65,25 @@ def _build_summary(self, location):
...
                 )
-        return title
+
+        summary_title_for_section = self._schema.get_summary_title_for_section(
+            section_id
+        )
+        if summary_title_for_section:
+            return self._placeholder_renderer.render(
+                summary_title_for_section, list_item_id
+            )
```
MebinAbraham, line 84: `render` expects a dict, this could be both a dict or a string, and it is just returning the string back because `find_pointers_containing(rendered_data, "placeholders")` is empty but shouldn't get to this method if it doesn't need to.

app/views/contexts/section_summary_context.py
MebinAbraham: Since the the hub requires the default section title, you would need to call it directly.
```python
    def _title_for_location(self, location):
        section_id = location.section_id
        title = (
            self._schema.get_repeating_title_for_section(section_id)
            or self._schema.get_summary_title_for_section(section_id)
            or self._schema.get_title_for_section(section_id)
        )
        if isinstance(title, dict):
            return self._placeholder_renderer.render_placeholder(
                title, location.list_item_id
            )

        return title

File: app/questionnaire/questionnaire_schema.py [hunk 3]
```diff
@@ -87,6 +87,9 @@ def get_show_on_hub_for_section(self, section_id):
     def get_summary_for_section(self, section_id: str) -> Mapping:
         return self._sections_by_id.get(section_id).get("summary", {})
 
+    def get_summary_title_for_section(self, section_id: str):
```
MebinAbraham, line 90: `secttin.py` has a method called `get_page_title` which returns `get_title_for_section` but I think it makes sense to return `get_summary_title_for_section` first.
```python
    def get_page_title(self):
        return self._schema.get_summary_title_for_section(
            self._section_id
        ) or self._schema.get_title_for_section(self._section_id)
```

LJBabbage: Thats a good spot

app/views/contexts/section_summary_context.py
LJBabbage: You might have used more characters than the PR on this comment, but I like the dict suggestion.

app/views/contexts/section_summary_context.py
```diff
@@ -65,15 +65,17 @@ def _build_summary(self, location):
...
-        title = self._schema.get_section(location.section_id).get("title")
-        if location.list_item_id:
-            repeating_title = self._schema.get_repeating_title_for_section(
-                location.section_id
+        section_id = location.section_id
+        title = (
+            self._schema.get_repeating_title_for_section(section_id)
+            or self._schema.get_summary_title_for_section(section_id)
+            or self._schema.get_title_for_section(section_id)
+        )
```
MebinAbraham, line 73: Could be wrapped in a `get_section_summary_title_for_section` method and in future cached.

tests/app/views/contexts/test_summary_context.py
```diff
@@ -136,6 +136,32 @@ def test_build_view_context_for_section_summary(self):
         self.assertEqual(len(context["summary"]), 5)
         self.assertTrue("title" in context["summary"])
 
+    def test_build_view_context_for_section_summary_title(self):
```
MebinAbraham, line 139: Title doesn't quite read right, maybe something like these?
- `test_custom_summary_title_is_used_when_provided`
- `test_section_title_is_used_when_no_custom_section_title`

ajmaddaford: Or:
- `test_custom_section_summary_title` 
- `test_section_summary_title_is_section_title`

test_schemas/en/test_section_summary.json
```diff
@@ -132,7 +132,10 @@
         {
             "id": "house-details-section",
             "title": "House Details Section",
-            "summary": { "show_on_completion": true },
+            "summary": {
+                "show_on_completion": true,
+                "title": "Household Summary"
```
ajmaddaford, line 137: Given the requirement is to be able to have placeholders in this title, it might be worth adding placeholders to the test schema.

LJBabbage: D'oh

PR #155, opened 2020-06-22
Update runner to design system v18

rmccar:
Update Runner to use v18 of the Design System

### How to review 
- Tests Pass


rmccar:
[[C] Change SAYT list to be visible when direct match only (S)](https://trello.com/c/3xRkIGh1/3931-c-change-sayt-list-to-be-visible-when-direct-match-only-s)

rmccar:
[Replace chained selectors in runner tests with data attribute selectors (S)](https://trello.com/c/WTsgnNkW/3697-replace-chained-selectors-in-runner-tests-with-data-attribute-selectors-s)

rmccar:
[Performance investigation for removing minification of HTML output](https://trello.com/c/6SkOMFNy/3921-performance-investigation-for-removing-minification-of-html-output)

templates/partials/answers/currency.html
```diff
@@ -8,6 +8,7 @@
   "name": answer.id,
   "value": input._value() | e,
   "label": {
+    "id": answer.id + "-label",
```
MebinAbraham, line 11: We should use the recommended way to concatenate strings which is using `~` so: `answer.id ~ "-label"`.

rmccar:
[Update Runner to Design System Version 18.1.0](https://trello.com/c/Sp4o38Kz/3951-update-runner-to-design-system-version-1810)

PR #158, opened 2020-06-23
Remove response_id from downstream payload

LJBabbage:
This PR removes response_id from the downstream payload. The card also included removing it from logging, however it doesn't seem to currently

### How to review 
Check response_id has been removed from all the right places, and there is nowhere else


tests/integration/routes/test_thank_you.py
```diff
@@ -2,25 +2,6 @@
 
 
 class TestThankYou(IntegrationTestCase):
-    example_payload = {
```
LJBabbage, line 5: On my searches I found this wasn't being used

PR #159, opened 2020-06-23
Remove Flask debug toolbar

petechd:
This removes [Flask-DebugToolbar](https://flask-debugtoolbar.readthedocs.io/en/latest/) dependency from Runner, script in _setup.py_ and all associated tests & environmental variables. Any mentions in README file has been removed as well. Pipfile.lock has been updated and regenerated.

### How to review 
Run unit and integration tests, check if debug toolbar is not present in Runner.


petechd:
[Remove Flask Debug Toolbar (s)](https://trello.com/c/UACwzgkr/3511-remove-flask-debug-toolbar-s)

petechd:
> A bit confused why we had EQ_DEV_MODE when it just seems a hurdle to get over to use flask-degubtoolbar, but I have checked the code and can't see it being used anywhere else

I did some searching with different keywords and, as you said, it looks like it's only being used as a step before actual flask toolbar script.

PR #160, opened 2020-06-23
Adds Individual Response post from individual section

JamesGardiner:
Adds the content for and routing for requesting an individual response by post from an individual section.

### How to review 
Check the functionality is as described in https://trello.com/b/K1jisSFp/eq-census

### Checklist

### DIFF
--- a/app/routes/individual_response.py
+++ b/app/routes/individual_response.py
@@ -1,0 +1,139 @@
+from flask import Blueprint, g, redirect, request, url_for
+from flask_login import current_user, login_required
+from structlog import get_logger
+
+from app.authentication.no_token_exception import NoTokenException
+from app.globals import get_metadata, get_session_store
+from app.helpers.language_helper import handle_language
+from app.helpers.schema_helpers import with_schema
+from app.helpers.session_helpers import with_questionnaire_store
+from app.helpers.template_helper import render_template
+from app.utilities.schema import load_schema_from_session_data
+from app.views.handlers.individual_response import (
+    IndividualResponseHandler,
+    IndividualResponseHowHandler,
+    IndividualResponsePostAddressConfirmHandler,
+)
+
+logger = get_logger()
+
+individual_response_blueprint = Blueprint(
+    name="individual_response", import_name=__name__, url_prefix="/individual-response/"
+)
+
+
+@individual_response_blueprint.before_request
+def before_individual_response_request():
+    metadata = get_metadata(current_user)
+    if not metadata:
+        raise NoTokenException(401)
+
+    logger.bind(
+        tx_id=metadata["tx_id"],
+        schema_name=metadata["schema_name"],
+        ce_id=metadata["collection_exercise_sid"],
+        questionnaire_id=metadata["questionnaire_id"],
+    )
+
+    logger.info(
+        "individual-response request", method=request.method, url_path=request.full_path
+    )
+
+    handle_language()
+
+    session_store = get_session_store()
+    g.schema = load_schema_from_session_data(session_store.session_data)
+
+
+@individual_response_blueprint.route("/", methods=["GET", "POST"])
+@login_required
+@with_questionnaire_store
+@with_schema
+def request_individual_response(schema, questionnaire_store):
+    language_code = get_session_store().session_data.language_code
+    list_item_id = request.args.get("list_item_id")
+    individual_response_handler = IndividualResponseHandler(
+        block_definition=None,
+        schema=schema,
+        questionnaire_store=questionnaire_store,
+        language=language_code,
+        form_data=request.form,
+        list_item_id=list_item_id,
+    )
+
+    if request.method == "GET":
+        return individual_response_handler.handle_get()
+
+    return redirect(url_for(".get_individual_response_how", list_item_id=list_item_id))
+
+
+@individual_response_blueprint.route("/<list_item_id>/how", methods=["GET", "POST"])
+@login_required
+@with_questionnaire_store
+@with_schema
+def get_individual_response_how(schema, questionnaire_store, list_item_id):
+    language_code = get_session_store().session_data.language_code
+    individual_response_handler = IndividualResponseHowHandler(
+        schema=schema,
+        questionnaire_store=questionnaire_store,
+        language=language_code,
+        form_data=request.form,
+        list_item_id=list_item_id,
+    )
+
+    if request.method == "GET" or not individual_response_handler.form.validate():
+        return individual_response_handler.handle_get()
+
+    return redirect(
+        url_for(
+            ".get_individual_response_post_address_confirm", list_item_id=list_item_id
+        )
+    )
+
+
+@individual_response_blueprint.route(
+    "/<list_item_id>/post/confirm-address", methods=["GET", "POST"]
+)
+@login_required
+@with_questionnaire_store
+@with_schema
+def get_individual_response_post_address_confirm(
+    schema, questionnaire_store, list_item_id
+):
+    language_code = get_session_store().session_data.language_code
+    individual_response_handler = IndividualResponsePostAddressConfirmHandler(
+        schema=schema,
+        questionnaire_store=questionnaire_store,
+        language=language_code,
+        form_data=request.form,
+        list_item_id=list_item_id,
+    )
+
+    if request.method == "GET" or not individual_response_handler.form.validate():
+        return individual_response_handler.handle_get()
+
+    return individual_response_handler.handle_post()
+
+
+@individual_response_blueprint.route("/post/confirmation", methods=["GET", "POST"])
+@login_required
+@with_questionnaire_store
+@with_schema
+def get_individual_response_post_address_confirmation(schema, questionnaire_store):
+    language_code = get_session_store().session_data.language_code
+    IndividualResponseHandler(
+        block_definition=None,
+        schema=schema,
+        questionnaire_store=questionnaire_store,
+        language=language_code,
+        form_data=request.form,
+        list_item_id=None,
+    )
+
+    if request.method == "GET":
+        return render_template(
+            template="individual_response/confirmation",
+            display_address=questionnaire_store.metadata.get("display_address"),
+        )
+
+    return redirect(url_for("questionnaire.get_questionnaire"))
--- a/app/views/contexts/hub_context.py
+++ b/app/views/contexts/hub_context.py
@@ -4,8 +4,8 @@ from flask import url_for
 from flask_babel import lazy_gettext
 
 from app.data_model.progress_store import CompletionStatus
-
-from app.views.contexts.context import Context
+from app.questionnaire import QuestionnaireSchema
+from app.views.contexts import Context
 
 
 class HubContext(Context):
@@ -32,6 +32,13 @@ class HubContext(Context):
                 "aria_label": lazy_gettext("Start {section_name} section"),
             },
         },
+        CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED: {
+            "text": lazy_gettext("Separate census requested"),
+            "link": {
+                "text": lazy_gettext("Change or resend"),
+                "aria_label": lazy_gettext("Change or resend"),
+            },
+        },
     }
 
     def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
@@ -43,6 +50,7 @@ class HubContext(Context):
 
         if survey_complete:
             title = custom_text.get("title") or lazy_gettext("Submit survey")
+
             guidance = custom_text.get("guidance") or lazy_gettext(
                 "Please submit this survey to complete it"
             )
@@ -52,17 +60,26 @@ class HubContext(Context):
             ) or lazy_gettext("Submit survey")
             submission_guidance = hub_schema.get("submission", {}).get("guidance")
 
+            individual_response_enabled = False
+
         else:
             title = lazy_gettext("Choose another section to complete")
+
             guidance = custom_text.get("guidance") or lazy_gettext(
                 "You must complete all sections in order to submit this survey"
             )
+
             submit_button = lazy_gettext("Continue")
             submission_guidance = None
+
+            individual_response_enabled = self._individual_response_enabled(
+                self._schema
+            )
 
         return {
             "title": title,
             "guidance": guidance,
+            "individual_response_enabled": individual_response_enabled,
             "submit_button": submit_button,
             "submission_guidance": submission_guidance,
             "rows": rows,
@@ -71,7 +88,6 @@ class HubContext(Context):
     def get_row_context_for_section(
         self, section_name: str, section_status: str, section_url: str
     ) -> Mapping[str, Union[str, List]]:
-
         section_content = self.SECTION_CONTENT_STATES[section_status]
         context: Mapping = {
             "rowTitle": section_name,
@@ -85,13 +101,17 @@ class HubContext(Context):
                                 section_name=section_name
                             ),
                             "url": section_url,
+                            "attributes": {"data-qa": "summary-actions-section-link"},
                         }
                     ],
                 }
             ],
         }
 
-        if section_status == CompletionStatus.COMPLETED:
+        if section_status in (
+            CompletionStatus.COMPLETED,
+            CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED,
+        ):
             context["rowItems"][0]["icon"] = "check-green"
 
         return context
@@ -138,7 +158,7 @@ class HubContext(Context):
                 repeating_list = self._schema.get_repeating_list_for_section(section_id)
 
                 if repeating_list:
-                    for list_item_id in self._list_store[repeating_list].items:
+                    for list_item_id in self._list_store[repeating_list]:
                         rows.append(
                             self._get_row_for_repeating_section(
                                 section_id, list_item_id
@@ -148,3 +168,19 @@ class HubContext(Context):
                     rows.append(self._get_row_for_section(section_title, section_id))
 
         return rows
+
+    def _individual_response_enabled(self, schema: QuestionnaireSchema) -> bool:
+        if not schema.json.get("individual_response"):
+            return False
+
+        for_list = schema.json["individual_response"]["for_list"]
+
+        count_household_members = len(self._list_store[for_list])
+
+        if count_household_members == 0:
+            return False
+
+        if count_household_members == 1 and self._list_store[for_list].primary_person:
+            return False
+
+        return True
--- a/app/views/handlers/content.py
+++ b/app/views/handlers/content.py
@@ -1,4 +1,7 @@
 from functools import cached_property
+from typing import Union
+
+from flask import url_for
 
 from app.questionnaire.schema_utils import transform_variants
 from app.views.handlers.block import BlockHandler
@@ -13,7 +16,25 @@ class Content(BlockHandler):
         return {
             "block": self.rendered_block,
             "metadata": dict(self._questionnaire_store.metadata),
+            "individual_response_url": self._individual_response_url(),
         }
+
+    def _individual_response_url(self) -> Union[str, None]:
+        if "individual_response" in self._schema.json:
+            for_list = self._schema.json["individual_response"]["for_list"]
+            list_item_id = self._current_location.list_item_id
+
+            primary_person_id = self._questionnaire_store.list_store[
+                for_list
+            ].primary_person
+
+            if list_item_id != primary_person_id:
+                return url_for(
+                    "individual_response.request_individual_response",
+                    list_item_id=list_item_id,
+                )
+
+        return None
 
     def _render_block(self, block_id):
         block_schema = self._schema.get_block(block_id)
--- a/app/views/handlers/individual_response.py
+++ b/app/views/handlers/individual_response.py
@@ -1,0 +1,309 @@
+from functools import cached_property
+from typing import List, Mapping
+
+from flask import redirect
+from flask.helpers import url_for
+from flask_babel import lazy_gettext
+from werkzeug.exceptions import NotFound
+
+from app.data_model.progress_store import CompletionStatus
+from app.forms.questionnaire_form import generate_form
+from app.helpers.template_helper import render_template
+from app.questionnaire.placeholder_renderer import PlaceholderRenderer
+from app.questionnaire.router import Router
+from app.views.contexts.question import build_question_context
+
+
+class IndividualResponseHandler:
+    _person_name_placeholder: List[Mapping] = [
+        {
+            "placeholder": "person_name",
+            "transforms": [
+                {
+                    "arguments": {
+                        "delimiter": " ",
+                        "list_to_concatenate": {
+                            "identifier": ["first-name", "last-name"],
+                            "source": "answers",
+                        },
+                    },
+                    "transform": "concatenate_list",
+                }
+            ],
+        }
+    ]
+
+    def __init__(
+        self,
+        block_definition,
+        schema,
+        questionnaire_store,
+        language,
+        form_data,
+        list_item_id,
+    ):
+        self._block_definition = block_definition
+        self._schema = schema
+        self._questionnaire_store = questionnaire_store
+        self._language = language
+        self._form_data = form_data
+        self._list_item_id = list_item_id
+
+        self.page_title = None
+
+        if not self._is_location_valid():
+            raise NotFound
+
+    def _is_location_valid(self):
+        if not self.router.can_access_hub():
+            return False
+
+        self._list_name = self._schema.json.get("individual_response", {}).get(
+            "for_list"
+        )
+        list_model = self._questionnaire_store.list_store[self._list_name]
+
+        if not list_model:
+            return False
+
+        if self._list_item_id:
+            if self._list_item_id not in list_model:
+                return False
+
+            if self._list_item_id == list_model.primary_person:
+                return False
+
+        return True
+
+    @cached_property
+    def rendered_block(self) -> Mapping:
+        return self._render_block()
+
+    @cached_property
+    def placeholder_renderer(self):
+        return PlaceholderRenderer(
+            language=self._language,
+            schema=self._schema,
+            answer_store=self._questionnaire_store.answer_store,
+            metadata=self._questionnaire_store.metadata,
+            location=None,
+            list_store=self._questionnaire_store.list_store,
+        )
+
+    @cached_property
+    def router(self):
+        return Router(
+            schema=self._schema,
+            answer_store=self._questionnaire_store.answer_store,
+            list_store=self._questionnaire_store.list_store,
+            progress_store=self._questionnaire_store.progress_store,
+            metadata=self._questionnaire_store.metadata,
+        )
+
+    @cached_property
+    def individual_section_id(self):
+        return self._schema.json.get("individual_response", {}).get(
+            "individual_section_id"
+        )
+
+    @cached_property
+    def form(self):
+        return generate_form(
+            schema=self._schema,
+            question_schema=self.rendered_block["question"],
+            answer_store=None,
+            metadata=self._questionnaire_store.metadata,
+            form_data=self._form_data,
+        )
+
+    def get_context(self):
+        return build_question_context(self.rendered_block, self.form)
+
+    def handle_get(self):
+        individual_section_first_block_id = self._schema.get_first_block_id_for_section(
+            self.individual_section_id
+        )
+
+        if self._list_item_id:
+            previous_location_url = url_for(
+                "questionnaire.block",
+                list_name=self._list_name,
+                list_item_id=self._list_item_id,
+                block_id=individual_section_first_block_id,
+            )
+        else:
+            previous_location_url = url_for("questionnaire.get_questionnaire")
+
+        return render_template(
+            template="individual_response/interstitial",
+            language=self._language,
+            previous_location_url=previous_location_url,
+        )
+
+    def _render_block(self):
+        return self.placeholder_renderer.render(
+            self._block_definition, self._list_item_id
+        )
+
+
+class IndividualResponseHowHandler(IndividualResponseHandler):
+    block_definition: Mapping = {
+        "type": "IndividualResponse",
+        "id": "individual-response",
+        "question": {
+            "type": "Question",
+            "id": "individual-response-how",
+            "title": {
+                "text": lazy_gettext(
+                    "How would you like <em>{person_name}</em> to receive a separate census?"
+                ),
+                "placeholders": IndividualResponseHandler._person_name_placeholder,
+            },
+            "description": lazy_gettext(
+                "<p>For someone to complete a separate census, we need to send them an individual access code.</p><p>Select how to send access code</p>"
+            ),
+            "answers": [
+                {
+                    "type": "Radio",
+                    "id": "individual-response-how-answer",
+                    "mandatory": False,
+                    "default": "Post",
+                    "options": [
+                        {
+                            "label": lazy_gettext("Post"),
+                            "value": "Post",
+                            "description": lazy_gettext(
+                                "We can only send this to an unnamed resident at the registered household address"
+                            ),
+                        }
+                    ],
+                }
+            ],
+        },
+    }
+
+    def __init__(self, schema, questionnaire_store, language, form_data, list_item_id):
+        super().__init__(
+            self.block_definition,
+            schema,
+            questionnaire_store,
+            language,
+            form_data,
+            list_item_id,
+        )
+
+    def handle_get(self):
+        previous_location_url = url_for(
+            "individual_response.request_individual_response",
+            list_item_id=self._list_item_id,
+        )
+
+        return render_template(
+            "individual_response/how",
+            language=self._language,
+            content=self.get_context(),
+            previous_location_url=previous_location_url,
+        )
+
+
+class IndividualResponsePostAddressConfirmHandler(IndividualResponseHandler):
+    block_definition: Mapping = {
+        "type": "IndividualResponse",
+        "question": {
+            "type": "Question",
+            "id": "individual-response-post-confirm",
+            "title": {
+                "text": lazy_gettext(
+                    "Do you want to send an individual access code for {person_name} by post?"
+                ),
+                "placeholders": IndividualResponseHandler._person_name_placeholder,
+            },
+            "description": lazy_gettext(
+                "A letter with an individual access code will be sent to your registered household address"
+            ),
+            "guidance": {
+                "contents": [
+                    {
+                        "description": lazy_gettext(
+                            "The letter will be addressed to <strong>Individual Resident</strong> instead of the name provided"
+                        )
+                    }
+                ]
+            },
+            "answers": [
+                {
+                    "type": "Radio",
+                    "id": "individual-response-post-confirm-answer",
+                    "mandatory": True,
+                    "options": [
+                        {
+                            "label": lazy_gettext("Yes, send the access code by post"),
+                            "value": "Yes, send the access code by post",
+                        },
+                        {
+                            "label": lazy_gettext("No, send it another way"),
+                            "value": "No, send it another way",
+                        },
+                    ],
+                }
+            ],
+        },
+    }
+
+    def __init__(self, schema, questionnaire_store, language, form_data, list_item_id):
+        super().__init__(
+            self.block_definition,
+            schema,
+            questionnaire_store,
+            language,
+            form_data,
+            list_item_id,
+        )
+
+    @cached_property
+    def answer_id(self):
+        return self.rendered_block["question"]["answers"][0]["id"]
+
+    @cached_property
+    def confirm_option(self):
+        return self.rendered_block["question"]["answers"][0]["options"][0]["value"]
+
+    @cached_property
+    def selected_option(self):
+        return self.form.get_data(self.answer_id)
+
+    def handle_get(self):
+        previous_location_url = url_for(
+            "individual_response.get_individual_response_how",
+            list_item_id=self._list_item_id,
+        )
+
+        return render_template(
+            "individual_response/question",
+            language=self._language,
+            content=self.get_context(),
+            previous_location_url=previous_location_url,
+        )
+
+    def handle_post(self):
+        if self.selected_option == self.confirm_option:
+            self._questionnaire_store.progress_store.update_section_status(
+                CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED,
+                self._schema.json["individual_response"]["individual_section_id"],
+                self._list_item_id,
+            )
+            if self._questionnaire_store.progress_store.is_dirty:
+                self._questionnaire_store.save()
+
+            return redirect(
+                url_for(
+                    "individual_response.get_individual_response_post_address_confirmation"
+                )
+            )
+
+        return redirect(
+            url_for(
+                "individual_response.get_individual_response_how",
+                list_item_id=self._list_item_id,
+            )
+        )
--- a/templates/errors/500.html
+++ b/templates/errors/500.html
@@ -1,13 +1,6 @@
 {% extends 'layouts/_base.html' %}
 
-{% set page_title = _("An error has occurred") + " - " + _("Office for National Statistics") %}
-
-{% set contact_us_url = "https://census.gov.uk/contact-us/" %}
-{% if schema_theme == "census-nisra" %}
-  {% set contact_us_url = "https://census.gov.uk/ni/contact-us/" %}
-{% elif language_code == "cy" %}
-  {% set contact_us_url = "https://cyfrifiad.gov.uk/cysylltu-a-ni/" %}
-{% endif %}
+{% set page_title = _("An error has occurred") ~ " - " ~ _("Office for National Statistics") %}
 
 {% block main %}
 <h1>{{ _("Sorry, there is a problem with this service") }}</h1>
--- a/templates/individual_response/how.html
+++ b/templates/individual_response/how.html
@@ -1,0 +1,7 @@
+{% extends 'individual_response/question.html' %}
+
+{% block after_submit_button_content %}
+    {% if is_post_confirm_page %}
+      <p class='u-mt-xl'>{{ _("To request a census in a different format or for further help, please <a href='{contact_us_url}'>contact us</a>".format(contact_us_url=contact_us_url)) }}</p>
+    {% endif %}
+{% endblock %}
--- a/templates/individual_response/interstitial.html
+++ b/templates/individual_response/interstitial.html
@@ -1,0 +1,23 @@
+{% extends 'layouts/_base.html' %}
+{% import 'macros/helpers.html' as helpers %}
+{% from "components/button/_macro.njk" import onsButton %}
+
+{% set save_on_signout = true %}
+
+
+{% block main %}
+    <button type="submit" class="u-d-no" hidden></button>
+    <div class="u-mb-s">
+      <a {{ helpers.track('click', 'Navigation', 'Previous link click', position )}} href="{{ previous_location_url }}" id="top-previous">{{ _("Previous") }}</a>
+    </div>
+    <h1 class="u-mt-m">{{_("If you can't answer questions for others in your household") }}</h1>
+    <p>{{_("You can ask the people you live with to answer their own questions by sharing the household access code with them.") }}</p>
+    <p>{{_("If this is not possible, you can request a separate census for them to complete.") }}</p>
+    {{
+        onsButton({
+            "text": _("Request separate census"),
+            "classes": 'btn u-mb-m u-mt-s btn--link',
+            "attributes": {'data-qa': 'btn-submit'}
+        })
+    }}
+{% endblock %}
--- a/templates/partials/individual-response-guidance.html
+++ b/templates/partials/individual-response-guidance.html
@@ -1,0 +1,29 @@
+{% from "components/collapsible/_macro.njk" import onsCollapsible %}
+
+{% set title = _("If you can’t answer questions for this person") %}
+
+{% call onsCollapsible({
+  "classes": "u-mt-s",
+  "title": title,
+  "headingAttributes": {
+    "data-ga": "click",
+    "data-ga-category": "definition",
+    "data-ga-action": "Open panel",
+    "data-ga-label": title
+  },
+  "button": {
+    "close": _("Hide this"),
+    "contextSuffix": "content",
+    "attributes": {
+      "data-ga": "click",
+      "data-ga-category": "definition",
+      "data-ga-action": "Close panel",
+      "data-ga-label": title
+    }
+  }
+}) %}
+  <div class="u-mt-s">
+      <p>{{ _("You can <em>share your household access code</em> with the people you live with so they can complete their own sections.") }}</p>
+      <p data-qa=individual-response-url >{{ _("If this is not possible, there are <a href={url}>other ways each person can complete their own census.</a>".format(url=content.individual_response_url))}}</p>
+  </div>
+{% endcall %}
--- a/tests/integration/individual_response/test_individual_response.py
+++ b/tests/integration/individual_response/test_individual_response.py
@@ -1,0 +1,404 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class IndividualResponseTestCase(IntegrationTestCase):
+    def setUp(self):
+        super().setUp()
+        self.launchSurvey("test_individual_response")
+
+    @property
+    def individual_section_link(self):
+        return self.getHtmlSoup().findAll(
+            "a", {"data-qa": "summary-actions-section-link"}
+        )[1]["href"]
+
+    @property
+    def individual_response_link(self):
+        return (
+            self.getHtmlSoup()
+            .find("p", {"data-qa": "individual-response-url"})
+            .find_next()["href"]
+        )
+
+    def _add_no_household_members(self):
+        self.get("questionnaire/primary-person-list-collector/")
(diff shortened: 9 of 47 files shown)


JamesGardiner:
> I think this might need to go back into dev, so stopped mid review, the card isn't the clearest, I think the logic might need reviewing and one page is missing.
> 
> I Seem to be able to request an individual response, then complete their section myself. Then by using previous, I can request another individual response. I can also manipulated the URL and do things I shouldn't be able to.
> 
> The page missing is the 'pick which person' you want to request a IR for from the hub, it just gives me a error and when I look for the text in the PR from the prototype I can't find it

IR from Hub is not part of this card. I've removed the link from the template. Logic will stay, as it is getting added in next ticket anyway.

As for protecting URL abuse (in particular requesting a primary person's IR) - I asked for a design on this, as it isn't covered in the original card at all, and have been told a 404 is the preferred answer. I'll be putting this in shortly.

The ability to complete a section after requesting an IR is known at the point at which this card ends. There is a separate card for updating the hub to change the text shown when an individual's IR has been requested. The logic for stopping a person then completing the individual's section should be introduced then.

app/routes/individual_response.py
```diff
@@ -0,0 +1,320 @@
...
+
+    if request.method == "GET":
+        individual_id = request.args.get("individual_id")
+        previous_location_url = (
+            f"/questionnaire/household/{individual_id}/individual-interstitial/"
+        )
+
+        return render_template(
+            template="individual-response",
+            session_timeout=session_timeout,
```
MebinAbraham, line 75: As mentioned by Andrew, we should remove occurrences of `session_timeout` being passed to `render_template` as it is no longer used in templates.

app/routes/individual_response.py
```diff
@@ -0,0 +1,323 @@
...
+        )
+
+    individual_id = request.args.get("individual_id")
+
+    return redirect(
+        url_for(".get_individual_response_how", individual_id=individual_id)
+    )
+
+
+@individual_response_blueprint.route("/<individual_id>/how", methods=["GET", "POST"])
```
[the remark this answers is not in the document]
MebinAbraham: It would improve readability if this was `list_item_id` so that it is consistent given they are  the same thing anyway.

app/routes/individual_response.py
```diff
@@ -0,0 +1,320 @@
...
+    language_code = get_session_store().session_data.language_code
+    session_timeout = get_session_timeout_in_seconds(schema)
+    previous_location_url = f"/individual-response?individual_id={individual_id}"
+
+    first_name = questionnaire_store.answer_store.get_answer(
+        "first-name", individual_id
+    ).value
+    last_name = questionnaire_store.answer_store.get_answer(
+        "last-name", individual_id
+    ).value
```
MebinAbraham, line 112: Given the individual section id is defined in the schema, should the answer id's also be defined there for consistency?. I feel like we lose in what ever we are trying to achieve by having a mix.

app/routes/individual_response.py
```diff
@@ -0,0 +1,320 @@
...
+        ],
+    }
+
+    form = generate_form(
+        schema,
+        block_definition,
+        None,
+        questionnaire_store.metadata,
+        form_data=request.form,
+    )
```
MebinAbraham, line 150: Would be nice for this to be named args, as I always have to look into the method when passing like None/empty data structure.

app/routes/individual_response.py
```diff
@@ -0,0 +1,323 @@
...
+        url_for(".get_individual_response_how", individual_id=individual_id)
+    )
+
+
+@individual_response_blueprint.route("/<individual_id>/how", methods=["GET", "POST"])
+@login_required
+@with_questionnaire_store
+@with_schema
+def get_individual_response_how(schema, questionnaire_store, individual_id):
+    router = Router(
```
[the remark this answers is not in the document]
JamesGardiner: that's not on the card

app/routes/individual_response.py
```diff
@@ -0,0 +1,323 @@
...
+
+    session_store = get_session_store()
+    g.schema = load_schema_from_session_data(session_store.session_data)
+
+
+@individual_response_blueprint.route("/", methods=["GET", "POST"])
+@login_required
+@with_questionnaire_store
+@with_schema
+def get_request_individual_response(schema, questionnaire_store):
```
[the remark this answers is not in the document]
JamesGardiner: Yeah I agree with the handler approach. I did have the hub accessible check here, I've removed it in error.

app/routes/individual_response.py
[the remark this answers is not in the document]
JamesGardiner: Sorry that is on the card - not sure why I thought it wasn't - I've also spoken to Donna about it, and 404 is the preferred response

templates/partials/individual-response-guidance.html
```diff
@@ -0,0 +1,51 @@
...
+{% if content.individual_response.hub_context %}
+  {% set title = _("If you can’t answer someone else’s questions") %}
+  {% set query_param = "?hub_context=true" %}
+{% else %}
+  {% set query_param = "?individual_id=" ~ content.individual_response.individual_id %}
+  {% set title = _("If you can’t answer questions for this person") %}
+{% endif %}
+
+{% set link_description =
+   _("If this is not possible, there are <a href=/individual-response%(query_param)s>other ways each person can do their own census.</a>", query_param=query_param)
```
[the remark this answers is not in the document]
JamesGardiner: Yeah so originally I had tried this, but it didn't work. Looks like the issue was me using a set in line assignment instead of the set block operator. Works now 👍

app/routes/individual_response.py
JamesGardiner: What do you think @ajmaddaford & @LJBabbage? I agree with @MebinAbraham that these should go to the schema, but I want to get further opinions before moving these ids

app/routes/individual_response.py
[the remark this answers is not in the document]
JamesGardiner: Resolving this as other comments track the handler approach, whereas the hub enabled checks are now back in on these endpoints

app/routes/individual_response.py
ajmaddaford: I'm not so sure. These would normally live in the block JSON for the page being rendered. Have you tried including them in that and using the existing code to render a block?

JamesGardiner: This approach will become easier when I create the IR handler - shall we see what you think then @MebinAbraham ?

templates/individual_response/how.html
```diff
@@ -0,0 +1,67 @@
+{% extends 'layouts/_base.html' %}
```
ajmaddaford, line 1: Did you try using the `question.html` template? (it provides the error message handling and block rendering)

templates/partials/individual-response-guidance.html
```diff
@@ -0,0 +1,34 @@
+{% from "components/collapsible/_macro.njk" import onsCollapsible %}
+
+{% set url = "/individual-response?list_item_id=" ~ content.individual_response.list_item_id %}
```
ajmaddaford, line 3: The url should be generated in code, and also use `url_for`.

app/views/contexts/hub_context.py
```diff
@@ -51,24 +59,35 @@ def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
...
+                self._schema
+            )
+
         return {
             "title": title,
             "guidance": guidance,
+            "individual_response": {
+                "enabled": individual_response_enabled,
+                "hub_context": True,
+            },
```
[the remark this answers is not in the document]
JamesGardiner: Yeah this should have been removed - I was looking at ways to differentiate the previous link in the individual response page depending on whether navigation there was from the hub or the individual section interstitial

app/routes/individual_response.py
JamesGardiner: spoken to @ajmaddaford about this - given we are now building the name using placeholder rendering defined in the block handler it's not necessary to put these into the schema

app/views/contexts/individual_response_context.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class IndividualResponseContext(Context):
+    @staticmethod
+    def get_context(content, form: QuestionnaireForm) -> Mapping:
+        legend = content["question"]["title"]
+        answer = content["question"]["answers"][0]
+
+        content = build_question_context(content, form)
+        content["contact_us_url"] = "https://census.gov.uk/contact-us"
```
MebinAbraham, line 16: We need to map these as they different based on region/lang.
NISRA: `https://census.gov.uk/ni/contact-us/`
Lang `cy`: `https://cyfrifiad.gov.uk/cysylltu-a-ni/`
Else: `https://census.gov.uk/contact-us/`

Currently, these mapped in error page HTML files.

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,355 @@
...
+        ).get_context(self.rendered_block, self.form)
+
+    def list_item_id_is_primary(self):
+        list_name = self._schema.json.get("individual_response", {}).get("for_list")
+        list_model = self._questionnaire_store.list_store[list_name]
+
+        if self._list_item_id not in list_model:
+            abort(404)
+
+        return self._list_item_id == list_model.primary_person
```
MebinAbraham, line 100: I think the method name should be `is_list_item_id_valid` where valid can also false for primary person for this context which is IR?. Also, the name suggests it always returns a boolean, but is inconsistent when id is not existing.

app/routes/individual_response.py
```diff
@@ -0,0 +1,285 @@
...
+    return redirect(url_for("questionnaire"))
+
+
+def _list_item_id_is_primary(list_item_id, list_store, list_name):
+    list_model = list_store[list_name]
+
+    if list_item_id not in list_model:
+        abort(404)
+
+    return list_item_id == list_model.primary_person
```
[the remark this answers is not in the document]
JamesGardiner: the design is for primary id to go to /questionnaire and for non-existent id to 404

app/views/contexts/hub_context.py
```diff
@@ -145,3 +164,19 @@ def _get_rows(self, enabled_section_ids) -> List[Mapping[str, Union[str, List]]]
...
+
+        for_list = schema.json["individual_response"]["for_list"]
+
+        count_household_members = len(self._list_store[for_list])
+
+        if count_household_members == 0:
+            return False
+
+        if count_household_members == 1 and self._list_store[for_list].primary_person:
+            return False
```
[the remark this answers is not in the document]
JamesGardiner: yeah I originally wrote it that way, but it obfuscates the logic. I think readability is better the way it is 👍

templates/macros/helpers.html
```diff
@@ -27,3 +27,12 @@
   } %}
 <a href="mailto:{{email}}{% if subject %}?subject={{ subject|urlencode }}{% endif %}" {{attr|xmlattr}}>{{email}}</a>
 {%- endmacro -%}
+
+{%- macro contact_us_url(schema_theme, language_code) -%}
+    href="https://census.gov.uk/contact-us/"
+  {% if schema_theme == "census-nisra" %}
+    href="https://census.gov.uk/ni/contact-us/"
+  {% elif language_code == "cy" %}
+    href="https://cyfrifiad.gov.uk/cysylltu-a-ni/"
```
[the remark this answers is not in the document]
ajmaddaford: I've just tried this and I don't think it's working - the contact us link on the 404 page always seems to be `https://census.gov.uk/contact-us/`.

app/views/contexts/hub_context.py
```diff
@@ -51,24 +59,32 @@ def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
...
             submit_button = lazy_gettext("Continue")
 
+            individual_response_enabled = self._individual_response_enabled(
+                self._schema
+            )
+
         return {
             "title": title,
             "guidance": guidance,
+            "individual_response": {"enabled": individual_response_enabled},
```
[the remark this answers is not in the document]
JamesGardiner: Only because I was following the hub pattern. More than happy to change to this if it's more readable

templates/macros/helpers.html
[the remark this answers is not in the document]
JamesGardiner: I'll add a test to the Cymraeg schema

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def _render_block(self):
+        transformed_block = transform_variants(
+            self._block_definition,
+            self._schema,
+            self._questionnaire_store.metadata,
+            self._questionnaire_store.answer_store,
+            self._questionnaire_store.list_store,
+            None,
+        )
```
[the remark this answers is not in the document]
JamesGardiner: I've not seen anything to suggest this

app/views/contexts/hub_context.py
```diff
@@ -51,24 +59,32 @@ def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
...
             submit_button = lazy_gettext("Continue")
 
+            individual_response_enabled = self._individual_response_enabled(
+                self._schema
+            )
+
         return {
             "title": title,
             "guidance": guidance,
+            "individual_response_enabled": individual_response_enabled,
```
MebinAbraham, line 80: Is this used?. I can't see any usages of this in `hub.html`.

app/views/handlers/content.py
```diff
@@ -13,8 +17,28 @@ def get_context(self):
...
+            )
+
+            primary_person_id = self._questionnaire_store.list_store[
+                for_list
+            ].primary_person
+
+            if list_item_id != primary_person_id:
+                return {"url": url}
+
+        return {}
```
MebinAbraham, line 40: Why do we return a dict when there is only 1 property?. Can we not just return url or None?.

templates/partials/individual-response-guidance.html
```diff
@@ -0,0 +1,33 @@
+{% from "components/collapsible/_macro.njk" import onsCollapsible %}
+
+{% set link_description %}
+    {{ content.individual_response.link_description}}
```
MebinAbraham, line 4: I can't see where this is being passed to the view.

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,335 @@
...
+                                    "list_to_concatenate": {
+                                        "identifier": ["first-name", "last-name"],
+                                        "source": "answers",
+                                    },
+                                },
+                                "transform": "concatenate_list",
+                            }
+                        ],
+                    }
+                ],
```
MebinAbraham, line 232: Could abstract this to a top level variable that is referenced by the 2 classes.

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,335 @@
...
+            },
+            "answers": [
+                {
+                    "type": "Radio",
+                    "id": "individual-response-post-confirm-answer",
+                    "mandatory": True,
+                    "options": [
+                        {
+                            "label": lazy_gettext("Yes, send the access by post"),
+                            "value": "Yes, send the access by post",
```
MebinAbraham, line 268: The label and value could be a variable. Given we don't validate these in-code schemas, it would help protect against someone changing the label but not the value and vice versa.

templates/partials/individual-response-guidance.html
```diff
@@ -0,0 +1,33 @@
...
+      "data-ga": "click",
+      "data-ga-category": "definition",
+      "data-ga-action": "Close panel",
+      "data-ga-label": _(title)
+    }
+  }
+}) %}
+  <div class="u-mt-s">
+      {{ _("<p>You can <em>share your household access code</em> with the people you live with so they can complete their own sections.</p>") }}
+      {{ _("<p>If this is not possible, there are <a href={url}>other ways each person can do their own census.</a></p>".format(url=content.individual_response.url)) }}
```
MebinAbraham, line 31: In the prototype, this is `can complete their` instead of `can do their`.

templates/individual_response/interstitial.html
```diff
@@ -0,0 +1,25 @@
...
+    <div class="u-mb-s">
+      <a {{ helpers.track('click', 'Navigation', 'Previous link click', position )}} href="{{ previous_location_url }}" id="top-previous">{{ _("Previous") }}</a>
+    </div>
+    <h1 class="u-mt-m">{{_("If you can't answer questions for others in your household") }}</h1>
+    <p>{{_("You can ask the people you live with to answer their own questions by sharing the household access code with them.") }}</p>
+    <p>{{_("If this is not possible, you can request a separate census for them to complete.") }}</p>
+    {{
+        onsButton({
+            "text": _("Request separate census"),
+            "classes": 'u-mb-m u-mt-l',
```
MebinAbraham, line 21: Our one has too much padding and is missing the `>` in the button. This should be:
```suggestion
            "classes": 'btn u-mb-m u-mt-s btn--link',
```

### Current:

image

### Expected:

image

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,335 @@
...
+            self._questionnaire_store.metadata,
+            self._questionnaire_store.answer_store,
+            self._questionnaire_store.list_store,
+            None,
+        )
+
+        return self.placeholder_renderer.render(transformed_block, self._list_item_id)
+
+
+class IndividualResponseHowHandler(IndividualResponseHandler):
```
MebinAbraham, line 131: The submit button text during the IR journey in the prototype is `Continue` vs our `Save and continue`.

templates/individual_response/how.html
```diff
@@ -0,0 +1,40 @@
...
+                {% include "partials/question.html" %}
+            </div>
+        </div>
+    {% endblock %}
+
+    {% block submit_button %}
+        {{ super() }}
+    {% endblock submit_button %}
+
+    {{ _("<p>To request a census in a different format or for further help, please <a href={contact_us_url} >contact us</a></p>".format(contact_us_url=contact_us_url)) }}
```
MebinAbraham, line 39: The padding for this doesn't match the prototype.

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,335 @@
...
+            "guidance": {
+                "contents": [
+                    {
+                        "description": lazy_gettext(
+                            "For security reasons we can only send access codes to an "
+                            "unnamed resident at the registered household address"
+                        )
+                    }
+                ]
+            },
```
MebinAbraham, line 259: We need to get confirmation of the design, because they have changed since, the property now uses:
image

**It is hard to review against a design that is changing. Need confirmation of the design to review this against. As mentioned above, the guidance has changed, the address is no longer piped in the prototype but is here etc. So I can't review it properly without knowing what it should be.**

app/views/contexts/hub_context.py
JamesGardiner: It will be when the next card goes in 😄

app/views/handlers/content.py
JamesGardiner: Because we like consistent return types and an empty dict still evaluates to true when compared to None?

templates/individual_response/interstitial.html
JamesGardiner: good spot

app/views/handlers/individual_response.py
JamesGardiner: For every label and value?

app/views/handlers/individual_response.py
JamesGardiner: I've asked @jrbarnes9 for confirmation on this

app/routes/individual_response.py
```diff
@@ -0,0 +1,156 @@
...
+    )
+
+    logger.info(
+        "individual-response request", method=request.method, url_path=request.full_path
+    )
+
+    handle_language()
+
+    session_store = get_session_store()
+    g.schema = load_schema_from_session_data(session_store.session_data)
```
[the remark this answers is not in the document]
JamesGardiner: Normally I'd agree, but as IR is a temporary feature I prefer the separation

app/views/handlers/individual_response.py
JamesGardiner: Confirmed that this  is the new design. I'll update in this ticket.

MebinAbraham: Does this apply to all the new screens?. I think this was the one stood as a very different.

app/views/handlers/content.py
MebinAbraham: I am referring to it just returning 1 thing which is a URL. It should return a URL string or None? (which is consistent with pretty much every thing else we do). Makes sense to return a dict if we have multiple properties?.

JamesGardiner: Yeah I see what you mean - it's a hangover from implementing part of the hub story as well. I'll switch it to url/None for this story

app/views/handlers/individual_response.py
JamesGardiner: Looks like this has changed again - I will push the updated change shortly

app/routes/individual_response.py
```diff
@@ -0,0 +1,156 @@
...
+        individual_response_handler = IndividualResponseHandler(
+            block_definition=None,
+            schema=schema,
+            questionnaire_store=questionnaire_store,
+            language=language_code,
+            form_data=request.form,
+            list_item_id=list_item_id,
+        )
+    except InvalidLocationException:
+        return redirect(url_for("questionnaire.get_questionnaire"))
```
ajmaddaford, line 67: Given the current in flight pull request to change questionnaire redirects to 404s, should this now be a 404?

app/routes/individual_response.py
```diff
@@ -0,0 +1,156 @@
...
+
+    if request.method == "GET" and individual_response_handler.individual_section_id:
```
ajmaddaford, line 69: Why is `individual_section_id` checked here and not validated in the initialisation of the handler?

app/routes/individual_response.py
```diff
@@ -0,0 +1,156 @@
...
+    try:
+        individual_response_handler = IndividualResponseHowHandler(
+            schema=schema,
+            questionnaire_store=questionnaire_store,
+            language=language_code,
+            form_data=request.form,
+            list_item_id=list_item_id,
+        )
+    except InvalidLocationException:
+        return redirect(url_for("questionnaire.get_questionnaire"))
```
ajmaddaford, line 91: Same as previous comment about this being a 404.

app/views/handlers/content.py
```diff
@@ -13,8 +16,28 @@ def get_context(self):
...
+                "individual_response.request_individual_response",
+                list_item_id=list_item_id,
+            )
+
+            primary_person_id = self._questionnaire_store.list_store[
+                for_list
+            ].primary_person
+
+            if list_item_id != primary_person_id:
+                return url
```
ajmaddaford, line 37:
```suggestion
            primary_person_id = self._questionnaire_store.list_store[
                for_list
            ].primary_person

            if list_item_id != primary_person_id:
                return url_for(
                    "individual_response.request_individual_response",
                    list_item_id=list_item_id,
                )
```

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,307 @@
...
+        self._list_item_id = list_item_id
+
+        self.page_title = None
+
+        if not self.can_access_hub:
+            raise InvalidLocationException(
+                "individual response not accessible without hub"
+            )
+
+        list_name = self._schema.json.get("individual_response", {}).get("for_list")
```
ajmaddaford, line 60: As `list_name` is used throughout, it would be cleaner to store it as a property on initialisation.

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,307 @@
...
+        list_model = self._questionnaire_store.list_store[list_name]
+
+        if self._list_item_id:
+            if self._list_item_id not in list_model:
+                abort(404)
```
ajmaddaford, line 65: Similar to the work done in the in progress 404 pull request, this should raise an `InvalidLocationException` and the 404 handled in the route.

app/views/handlers/individual_response.py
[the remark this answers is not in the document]
ajmaddaford: Then we don't need to call `transform_variants`?

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,307 @@
...
+        )
+
+    def handle_post(self):
+        if self.selected_option == self.confirm_option:
+            self._questionnaire_store.progress_store.update_section_status(
+                CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED,
+                self._schema.json["individual_response"]["individual_section_id"],
+                self._list_item_id,
+            )
+            self._questionnaire_store.save()
```
ajmaddaford, line 294: The progress store `is_dirty` flag should be checked here to see if saving is necessary.

templates/partials/individual-response-guidance.html
```diff
@@ -0,0 +1,29 @@
...
+      "data-ga": "click",
+      "data-ga-category": "definition",
+      "data-ga-action": "Close panel",
+      "data-ga-label": _(title)
+    }
+  }
+}) %}
+  <div class="u-mt-s">
+      {{ _("<p>You can <em>share your household access code</em> with the people you live with so they can complete their own sections.</p>") }}
+      {{ _("<p>If this is not possible, there are <a href={url}>other ways each person can complete their own census.</a></p>".format(url=content.individual_response_url)) }}
```
ajmaddaford, line 27: The `p` tags should be outside of the translatable content.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+
+    @property
+    def individual_section_link(self):
+        return self.getHtmlSoup().findAll("a", {"class": "summary__button"})[1]["href"]
+
+    @property
+    def individual_response_link(self):
+        return [
+            td.find("a")
+            for td in self.getHtmlSoup().findAll("div", {"class": "u-mt-s"})
```
ajmaddaford, line 19: We shouldn't use classes to locate elements in tests. These should use an `id` or a `data-qa` attribute.

tests/integration/routes/test_questionnaire_language.py
```diff
@@ -194,3 +194,10 @@ def test_error_messages(self):
         self.assertInBody("Mae 1 gwall ar y dudalen hon")
         self.assertInBody("Mae'n <strong>rhaid cywiro'r</strong> rhain cyn parhau")
         self.assertInBody("Nodwch ateb i barhau")
+
+    def test_contact_us(self):
```
ajmaddaford, line 198:
```suggestion
    def test_contact_us_link(self):
```

app/routes/individual_response.py
```diff
@@ -0,0 +1,156 @@
...
+
+    session_store = get_session_store()
+    g.schema = load_schema_from_session_data(session_store.session_data)
+
+
+@individual_response_blueprint.route("/", methods=["GET", "POST"])
+@login_required
+@with_questionnaire_store
+@with_schema
+def request_individual_response(schema, questionnaire_store):
```
MebinAbraham, line 53: This route and all IR routes need to check if the list store for the IR list is empty. At the moment, you can access these pages even when the list is empty, which results in 500 errors.

app/routes/individual_response.py
```diff
@@ -0,0 +1,156 @@
...
+
+logger = get_logger()
+
+individual_response_blueprint = Blueprint(
+    name="individual_response", import_name=__name__, url_prefix="/individual-response/"
+)
+
+
+@individual_response_blueprint.before_request
+def before_individual_response_request():
```
MebinAbraham, line 27: The IR routes are missing the language toggle.

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,307 @@
...
+                "placeholders": IndividualResponseHandler._person_name_placeholder,
+            },
+            "description": lazy_gettext(
+                "<p>For someone to complete a separate census, we need to send them an individual access code.</p><p>Select how to send access code</p>"
+            ),
+            "answers": [
+                {
+                    "type": "Radio",
+                    "id": "individual-response-how-answer",
+                    "mandatory": True,
```
MebinAbraham, line 169: In the prototype, this isn't mandatory and it defaults to `post`.

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,307 @@
...
+
+class IndividualResponsePostAddressConfirmHandler(IndividualResponseHandler):
+    block_definition: Mapping = {
+        "type": "IndividualResponse",
+        "question": {
+            "type": "Question",
+            "id": "individual-response-post-confirm",
+            "title": {
+                "text": lazy_gettext(
+                    "Do you want to send an individual access code to <em>{person_name}</em> by post?"
```
MebinAbraham, line 216: In the prototype, this is `access code for` not `access code to`. And the person name is not surrounded by `em` tags, so worth checking if it should be.

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,307 @@
...
+            },
+            "answers": [
+                {
+                    "type": "Radio",
+                    "id": "individual-response-post-confirm-answer",
+                    "mandatory": True,
+                    "options": [
+                        {
+                            "label": lazy_gettext("Yes, send the access by post"),
+                            "value": "Yes, send the access by post",
```
MebinAbraham, line 240: This should be `Yes, send the access code by post`

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,307 @@
...
+            "type": "Question",
+            "id": "individual-response-post-confirm",
+            "title": {
+                "text": lazy_gettext(
+                    "Do you want to send an individual access code to <em>{person_name}</em> by post?"
+                ),
+                "placeholders": IndividualResponseHandler._person_name_placeholder,
+            },
+            "description": lazy_gettext(
+                "<p>A letter with an indivdual access code will be sent to your registered household address</p>"
```
MebinAbraham, line 221:
```suggestion
                "<p>A letter with an individual access code will be sent to your registered household address</p>"
```

templates/individual_response/post-confirmation.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'layouts/_base.html' %}
+{% from 'components/button/_macro.njk' import onsButton %}
+{% from 'components/panel/_macro.njk' import onsPanel %}
+{% from 'components/question/_macro.njk' import onsQuestion %}
+
```
MebinAbraham, line 5: This page should have `{% set save_on_signout = true %}`

templates/individual_response/how.html
```diff
@@ -0,0 +1,23 @@
...
+{% from "components/button/_macro.njk" import onsButton %}
+{% from "components/question/_macro.njk" import onsQuestion %}
+{% from "components/radios/_macro.njk" import onsRadios %}
+{% from "components/fieldset/_macro.njk" import onsFieldset %}
+
+{% set save_on_signout = true %}
+{% set answer = content.block.question.answers[0] %}
+{% set form = content.form %}
+{% set question = content.block.question %}
+{% set question_title = content.title %}
```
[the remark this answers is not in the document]
JamesGardiner: question and question_title still need to be set from what I can see

templates/individual_response/how.html
```diff
@@ -0,0 +1,23 @@
...
+{% set contact_us_url = helpers.contact_us_url(schema_theme, language_code) %}
+
+{% block after_submit_button_content %}
+    {% if content.return_to_hub_url %}
+      <p class="u-mt-xl">{{ _("Can't complete this question?") }}<br>
+      <a href="{{ content.return_to_hub_url }}">{{ _("Choose another section and return to this later") }}</a></p>
+    {% endif %}
```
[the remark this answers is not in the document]
JamesGardiner: Yeah it was leftover from hub implementation.

templates/individual_response/how.html
```diff
@@ -0,0 +1,23 @@
...
+{% from "components/question/_macro.njk" import onsQuestion %}
+{% from "components/radios/_macro.njk" import onsRadios %}
+{% from "components/fieldset/_macro.njk" import onsFieldset %}
+
+{% set save_on_signout = true %}
+{% set answer = content.block.question.answers[0] %}
+{% set form = content.form %}
+{% set question = content.block.question %}
+{% set question_title = content.title %}
+{% set contact_us_url = helpers.contact_us_url(schema_theme, language_code) %}
```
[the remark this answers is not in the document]
JamesGardiner: done - and all the contact_us_url logic removed from templates 👍

templates/individual_response/interstitial.html
```diff
@@ -0,0 +1,25 @@
...
+    <p>{{_("You can ask the people you live with to answer their own questions by sharing the household access code with them.") }}</p>
+    <p>{{_("If this is not possible, you can request a separate census for them to complete.") }}</p>
+    {{
+        onsButton({
+            "text": _("Request separate census"),
+            "classes": 'btn u-mb-m u-mt-s btn--link',
+            "attributes": {'data-qa': 'btn-submit'}
+        })
+    }}
+{% endblock %}
```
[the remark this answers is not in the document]
JamesGardiner: The reason they are different is that there is no block definition for this interstitial page

app/routes/individual_response.py
JamesGardiner: This can be removed, as it is validated in the initialisation of the handler

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+        self.get(f"individual-response/{primary_person_id}/how")
+        self.assertInBody("Page not found")
+
+    def test_individual_response_confirm_page_404_if_primary_id_used(self):
+        # Given I add a primary person
+        self._add_primary()
+
+        # When I navigate to the how endpoint using the primary person's
+        # list item id
+        self.post()
```
ajmaddaford, line 361: I don't think this is navigating to the how endpoint?

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+
+        primary_person_id = self.last_url.split("/")[3]
+        self.get(f"individual-response/{primary_person_id}/how")
+        self.assertInBody("Page not found")
```
ajmaddaford, line 353: I think this test file needs to be give a once over for consistency of the Given When Then comments. Here the When should be after the `self.post()` and there should be a Then before the assert.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+        self._add_household_no_primary()
+        self.get(self.individual_section_link)
+        self.get(self.individual_response_link)
+        self.post()
+        self.post({"individual-response-how-answer": "Post"})
+
+        # When I click the previous link
+        self.get(self.previous_link)
+
+        # Then I should be taken back to the hub
```
ajmaddaford, line 249:
```suggestion
        # Then I should be taken back to the how page
```

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+        self.assertNotInBody("If you can’t answer someone else’s questions")
+
+        # Then when I navigate to the next individual section I should see the
+        # individual response guidance
+        self.post({"proxy-answer": "Yes, I am"})
+        self.post()
+        self.assertInBody("You will need to know personal details such as")
+        self.assertInBody("If you can’t answer questions for this person")
+
+    def test_navigate_back_to_individual_page_after_viewing_ir_page(self):
```
ajmaddaford, line 210: The test name doesn't seem to match the actions.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+
+        # When I sign out and navigate to the individual response page
+        individual_response_link = self.individual_response_link
+        self.get(individual_response_link)
+        self.post()
+        self.get(self.sign_out_link)
+        self.get(individual_response_link)
+
+        # Then I should see the signed out page
+        self.assertInBody("Your session has expired")
```
ajmaddaford, line 110: This should assert for the HTTP status code, rather than the text (we wouldn't want the text of these pages changing to break these tests). 

This applies to the other checks like this in the following tests e.g. "Page not found" -> 404.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+
+    def test_404_invalid_list_item_id(self):
+        # Given the hub is enabled
+        self._add_household_no_primary()
+        # When I use an invalid id in an individual response url
+        self.get("/")
+        self.get(f"individual-response/not-an-id/how")
```
ajmaddaford, line 93:
```suggestion
        self.get("/individual-response/not-an-id/how")
```

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+
+    def test_post_confirmation_404_when_hub_not_accessible(self):
+        # Given the hub is not enabled
+        # And I try to navigate to the individual response how page
+        self.get("/individual-response/post/confirmation")
+
+        # Then I should be redirected
+        self.assertInBody("Page not found")
+
+    def test_ir_guidance_not_displayed_when_no_household_members(self):
```
ajmaddaford, line 144:
```suggestion
    def test_hub_ir_guidance_not_displayed_when_no_household_members(self):
```

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+        # Given the hub is enabled
+        # And I navigate to the hub after adding a primary person
+        self._add_no_household_members()
+
+        # Then I should not see the individual response guidance
+        self.assertInBody("Submit survey")
+        self.assertNotInBody("If you can't answer someone else's questions")
+        self.assertNotInBody("Hide")
+
+    def test_ir_guidance_not_displayed_when_only_primary(self):
```
ajmaddaford, line 154:
```suggestion
    def test_hub_ir_guidance_not_displayed_when_only_primary(self):
```

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+        # Given the hub is enabled
+        # And I navigate to the hub after adding a primary person
+        self._add_primary()
+
+        # Then I should not see the individual response guidance
+        self.assertInBody("Choose another section to complete")
+        self.assertNotInBody("If you can't answer someone else's questions")
+        self.assertNotInBody("Hide")
+
+    def test_ir_guidance_not_displayed_when_primary(self):
```
ajmaddaford, line 164: As the guidance can appear on the hub or individual section, the naming needs to be clearer. 
```suggestion
    def test_individual_section_ir_guidance_not_displayed_when_primary(self):
```

This applies to all of the tests for testing the individual section guidance.

Alternatively, it could be worth considering splitting the tests up into hub, individual section and individual response tests.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+        self.post()
+        self.post({"individual-response-how-answer": "Post"})
+
+        # When I choode to send the individual response code another way
+        self.post(
+            {"individual-response-post-confirm-answer": "No, send it another way"}
+        )
+
+        # Then I should be redirected to the how page
+        self.assertInBody("Select how to send access code")
```
ajmaddaford, line 267: It would be better to assert the url rather than page content here.

app/routes/individual_response.py
JamesGardiner: Because the test schema only has an 'en' version

templates/partials/individual-response-guidance.html
```diff
@@ -0,0 +1,29 @@
...
+      "data-ga": "click",
+      "data-ga-category": "definition",
+      "data-ga-action": "Close panel",
+      "data-ga-label": _(title)
+    }
+  }
+}) %}
+  <div class="u-mt-s">
+      <p>{{ _("You can <em>share your household access code</em> with the people you live with so they can complete their own sections.") }}</p>
+      <p>{{ _("If this is not possible, there are <a data-qa=individual-response-url href={url}>other ways each person can complete their own census.</a>".format(url=content.individual_response_url))}}</p>
```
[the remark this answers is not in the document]
JamesGardiner: Yep makes sense

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,371 @@
...
+        # individual list item page
+        self._add_household_no_primary()
+        self.get(self.individual_section_link)
+        self.get(self.individual_response_link)
+        self.post()
+
+        # When I click the previous link
+        self.get(self.previous_link)
+
+        # Then I should be taken back to the hub
```
ajmaddaford, line 278:
```suggestion
        # Then I should be taken back to the individual response introduction page
```

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,371 @@
...
+
+        # When I choose to send the individual response code another way
+        self.post(
+            {"individual-response-post-confirm-answer": "No, send it another way"}
+        )
+
+        # Then I should be redirected to the how page
+        self.assertInUrl("/how")
+
+    def test_mandatory_error_rendered_on_confirm_address(self):
```
ajmaddaford, line 312: The tests from here are not related to navigation. Either the test class name needs to change, or another appropriate class created to group them together.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,371 @@
...
+        self.post({"individual-response-how-answer": "Post"})
+
+        # When I post "Yes, send the access code by post"
+        self.post(
+            {
+                "individual-response-post-confirm-answer": "Yes, send the access code by post"
+            }
+        )
+
+        # Then I should see errors rendered correctly
```
ajmaddaford, line 343:
```suggestion
        # Then I should see the address
```

templates/individual_response/how.html
```diff
@@ -0,0 +1,11 @@
...
+
+{% set save_on_signout = true %}
+{% set question = content.block.question %}
+{% set continue_button_text = _("Continue") %}
+
+{% block after_submit_button_content %}
+    {% if is_post_confirm_page %}
+      <p class='u-mt-xl'>{{ _("To request a census in a different format or for further help, please <a href='{contact_us_url}'>contact us</a>".format(contact_us_url=contact_us_url)) }}</p>
+    {% endif %}
+{% endblock %}
```
ajmaddaford, line 11: As discussed - we should create an individual response `question.html` and this should extend that to add the extra after submit button content, removing the need for the `is_post_confirm_page` boolean.

templates/errors/500.html
```diff
@@ -2,13 +2,6 @@
 
 {% set page_title = _("An error has occurred") + " - " + _("Office for National Statistics") %}
```
ajmaddaford, line 3:
```suggestion
{% set page_title = _("An error has occurred") ~ " - " ~ _("Office for National Statistics") %}
```

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+
+    @property
+    def previous_link(self):
+        return self.getHtmlSoup().find("a", {"data-ga-action": "Previous link click"})[
+            "href"
+        ]
+
+    @property
+    def sign_out_link(self):
+        return self.getHtmlSoup().find("a", {"name": "btn-save-sign-out"})["href"]
```
MebinAbraham, line 31: I think it's worth adding this to `IntegrationTestCase` but making it work similar to `previous`.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+        self.post(
+            {
+                "individual-response-post-confirm-answer": "Yes, send the access code by post"
+            }
+        )
+        self.post()
+
+
+class TestIndividualResponseErrorStatus(IndividualResponseTestCase):
+    def test_401_how_redirects_signed_out(self):
```
MebinAbraham, line 78: `test_ir_is_not_accessible_without_session` or `test_ir_is_not_accessible_after_signout` or `test_ir_raises_401_without_session`?. Something along those lines as the current one wasn't clear to me.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+        self.assertStatusCode(404)
+
+    def test_404_post_confirmation_when_hub_not_accessible(self):
+        # Given I try to navigate to the individual response how page
+        self.get("/individual-response/post/confirmation")
+
+        # Then I should see the 404 page
+        self.assertStatusCode(404)
+
+    def test_404_individual_response_page__if_primary_id_used(self):
```
MebinAbraham, line 144:
```suggestion
    def test_404_individual_response_page_if_primary_id_used(self):
```

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+
+    def test_404_individual_response_no_list_items(self):
+        # Given I add no household members
+        self._add_no_household_members()
+
+        # When I navigate to the how endpoint using a fake id
+        self.get("/individual-response/no-id/how")
+
+        # A 404 will be returned
+
```
MebinAbraham, line 192:
```suggestion
```

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+        self.post()
+
+        # When I click the previous link
+        self.get(self.previous_link)
+
+        # Then I should be taken back to the individual response introduction page
+        self.assertInBody("If you can't answer questions for others in your household")
+
+
+class TestIndividualResponseConfirmationPage(IndividualResponseTestCase):
```
MebinAbraham, line 310: We should add a test for the `default` route of `Post`.

tests/integration/individual_response/test_individual_response.py
JamesGardiner: I've gone with `test_ir_raises_401_without_session`

tests/integration/individual_response/test_individual_response.py
```diff
@@ -0,0 +1,375 @@
...
+            self.getHtmlSoup()
+            .find("p", {"data-qa": "individual-response-url"})
+            .find_next()["href"]
+        )
+
+    @property
+    def previous_link(self):
+        return self.getHtmlSoup().find("a", {"data-ga-action": "Previous link click"})[
+            "href"
+        ]
```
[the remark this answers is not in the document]
JamesGardiner: No - I wasn't aware of it 👍

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,304 @@
...
+
+        if not list_model:
+            raise NotFound
+
+        if self._list_item_id:
+            if self._list_item_id not in list_model:
+                raise NotFound
+
+            if self._list_item_id == list_model.primary_person:
+                raise NotFound
```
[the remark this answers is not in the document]
JamesGardiner: Agreed and done

templates/individual_response/post-confirmation.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'individual_response/question.html' %}
+
+{% block after_submit_button_content %}
+    {% if is_post_confirm_page %}
+      <p class='u-mt-xl'>{{ _("To request a census in a different format or for further help, please <a href='{contact_us_url}'>contact us</a>".format(contact_us_url=contact_us_url)) }}</p>
+    {% endif %}
```
ajmaddaford, line 6: In the prototype, this text and link are only on the "how" page. Also, the text doesn't appear on screen as it's still wrapped in a conditional that is never set.

templates/individual_response/post-confirmation.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'individual_response/question.html' %}
```
ajmaddaford, line 1: The template names confused me as there are two confirmations. I think this would be better named `confirm-address.html`.

app/views/contexts/hub_context.py
```diff
@@ -85,13 +101,17 @@ def get_row_context_for_section(
                                 section_name=section_name
                             ),
                             "url": section_url,
+                            "attributes": {"data-qa": "summary-actions-summary-button"},
```
ajmaddaford, line 104: I realise this was the value of the attribute previously, but this would be better named something like `summary-actions-section-link`.

app/views/handlers/individual_response.py
```diff
@@ -0,0 +1,304 @@
...
+        individual_section_first_block_id = self._schema.get_first_block_id_for_section(
+            self.individual_section_id
+        )
+
+        previous_location_url = url_for(
+            "questionnaire.block",
+            list_name=self._list_name,
+            list_item_id=self._list_item_id,
+            block_id=individual_section_first_block_id,
+        )
```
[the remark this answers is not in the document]
JamesGardiner: I think it should raise NotFound

templates/individual_response/post-confirmation.html
JamesGardiner: actually switched these around now due to the contact us text being in the wrong template:

there is now:
	- question.html
	- interstitial.html
	- confirmation.html
	- how.html

where how inherits from individual_response/question.html

app/views/handlers/individual_response.py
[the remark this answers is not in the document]
MebinAbraham: As discussed, this was about where the previous URL goes to, as it is valid to get to this page without a list item id. The presence of it just determines where the previous and button links to.

PR #161, opened 2020-06-24
Escape string placeholder answer values

iwootten:

We currently escape all answers as they are retrieved from the answer_store. Ideally we would escape them only after transforms have been performed, but this is not possible when a transform returns html (such as format_list). Instead, this PR escapes only string answer values only from the answer store.

### How to review 

Decide if you agree on the approach, confirm tests pass.


iwootten:
[Escaped answer values in placeholder rendering (M)](https://trello.com/c/H9XlGzBW/3741-escaped-answer-values-in-placeholder-rendering-m)

PR #164, opened 2020-06-26
Fix section summary page title bug

LJBabbage:
Section page titles weren't rendering placeholder, this PR fixes that

### How to review 
Check the page title is now correct on test_section_summary (questionnaire/sections/house-details-section/)


app/routes/questionnaire.py
```diff
@@ -155,13 +155,14 @@ def get_section(schema, questionnaire_store, section_id, list_item_id=None):
...
+            section_context = section_handler.context()
             return _render_page(
                 template="SectionSummary",
-                context=section_handler.context(),
+                context=section_context,
                 current_location=section_handler.current_location,
                 previous_location_url=section_handler.get_previous_location_url(),
                 schema=schema,
-                page_title=section_handler.get_page_title(),
+                page_title=section_context.get("summary").get("title"),
```
MebinAbraham, line 165: These keys always exist therefore `.get` is redundant.
```suggestion
                page_title=section_context["summary"]["title"],
```

app/views/handlers/block.py
```diff
@@ -116,3 +117,11 @@ def _set_started_at_metadata(self):
             started_at = datetime.utcnow().isoformat()
             logger.info("Survey started", started_at=started_at)
             collection_metadata["started_at"] = started_at
+
+    def _get_safe_page_title(self, title):
+        if title and "placeholders" in title:
+            if "text_plural" in title:
+                title = title["text_plural"]["forms"]["other"]
```
LJBabbage, line 124: As the number is replaced with ... and other is always there, this seems the most logical one to use

app/views/contexts/section_summary_context.py
```diff
@@ -148,3 +158,11 @@ def _add_link(self, summary, current_location, section, list_collector_block):
                 block_id=driving_question_block["id"],
                 return_to_summary=True,
             )
+
+    def _get_safe_page_title(self, title, title_has_placeholders=False):
+        if title_has_placeholders:
+            if "text_plural" in title:
+                title = title["text_plural"]["forms"]["other"]
```
LJBabbage, line 165: As the number is replaced with ... and other is always there, this seems the most logical one to use

test_schemas/en/test_section_summary.json
```diff
@@ -187,6 +187,60 @@
...
+                                    {
+                                        "id": "number-of-people-answer",
+                                        "label": "How many people live here?",
+                                        "mandatory": true,
+                                        "type": "Number"
+                                    }
+                                ],
+                                "description": "",
+                                "id": "number-of-people-question",
+                                "title": "",
```
berroar, line 236: We should add a title here and remove the description above before my validator PR gets merged 😀

LJBabbage: Also means I don't need to do a null check on title anymore, which I had to do above

tests/app/views/contexts/test_summary_context.py
```diff
@@ -152,6 +152,38 @@ def test_custom_section_summary_title(self):
             context["summary"]["title"], "Household Summary - Semi-detached"
         )
 
+    def test_custom_section_summary_page_title_placeholder_text_replaced(self):
```
ajmaddaford, line 155: I don't think this is a "custom" section summary?

tests/app/views/contexts/test_summary_context.py
```diff
@@ -152,6 +152,38 @@ def test_custom_section_summary_title(self):
...
+            self.progress_store,
+            self.list_store,
+            self.metadata,
+        )
+        context = summary_context(current_location)
+        self.assertEqual(
+            context["summary"]["page_title"], "Household Summary - … - Section Summary"
+        )
+
+    def test_custom_section_summary_page_title_placeholder_text_plural_replaced(self):
```
ajmaddaford, line 171: I don't think this is a "custom" section summary?

test_schemas/en/test_section_summary.json
```diff
@@ -187,6 +187,60 @@
...
+                    "title": "Number of people",
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "number-of-people",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "number-of-people-answer",
+                                        "label": "How many people live here?",
```
ajmaddaford, line 229: This should be the question title, the answer label should be something like "Number of people".

test_schemas/en/test_section_summary.json
ajmaddaford: Question title is mandatory - https://github.com/ONSdigital/eq-questionnaire-validator/blob/master/schemas/questions/types/general.json#L93

MebinAbraham: Question title is, summary title isn't which is what is used by `title_for_location` method.

app/views/handlers/content.py
```diff
@@ -36,11 +35,4 @@ def _render_block(self, block_id):
...
         if content:
-            if isinstance(content["title"], str):
-                content_title = content["title"]
-            else:
-                content_title = content["title"]["text"]
-
-            page_title = f'{content_title} - {self._schema.json["title"]}'
-
-            return safe_content(page_title)
+            return self._get_safe_page_title(content["title"])
```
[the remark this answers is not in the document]
LJBabbage: Summary uses this so I think we still need to check content

app/views/handlers/block.py
```diff
@@ -116,3 +117,11 @@ def _set_started_at_metadata(self):
...
+            else:
+                title = title["text"]
```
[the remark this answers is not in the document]
LJBabbage: So there is a generator outside the class that already does this https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/questionnaire/questionnaire_schema.py#L412 but I am not overly comfortable with using it. It recursively looks for the key in sub dicts/list until it finds it, but if there was a duplication of the key in a sub dict it would also find it. There is an ignore list which helps but I think we need to be more explicit here

We would also have to map keys to different values, which isn't a problem but would need to be done i.e text plural = ["text_plural"]["forms"]["other"]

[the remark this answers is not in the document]
LJBabbage: Not a fan of the arg name either, we should try to not use 'block', considering its meaning in our schema

test_schemas/en/test_section_summary.json
ajmaddaford: Ah, context is everything. The comment is against the question title 🤷 .

app/views/handlers/block.py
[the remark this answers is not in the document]
ajmaddaford: I meant adding a method similar to https://github.com/ONSdigital/eq-translations/blob/master/eq_translations/survey_schema.py#L83.

app/views/handlers/content.py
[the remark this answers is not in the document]
ajmaddaford: This is in the content handler, I can't see where it is used by the summary.

app/views/contexts/section_summary_context.py
```diff
@@ -11,10 +12,21 @@
...
+            else title_for_location
+        )
 
         return {
             "summary": {
-                "title": self._title_for_location(current_location),
+                "title": title,
+                "page_title": safe_content(
+                    f'{self._schema.get_single_string_value(title_for_location)} - {self._schema.json["title"]}'
+                ),
```
MebinAbraham, line 29: Can we keep this as a method call to `_get_safe_page_title` as I think it will improve readability and make it consistent with other places in runner?

app/views/contexts/section_summary_context.py
```diff
@@ -155,3 +152,8 @@ def _add_link(self, summary, current_location, section, list_collector_block):
                 block_id=driving_question_block["id"],
                 return_to_summary=True,
             )
+
+    def _get_safe_page_title(self, title_for_location):
+        return safe_content(
+            f'{self._schema.get_single_string_value(title_for_location)} - {self._schema.json["title"]}'
+        )
```
MebinAbraham, line 159:
```suggestion
    def _get_safe_page_title(self, title):
        return safe_content(
            f'{self._schema.get_single_string_value(title)} - {self._schema.json["title"]}'
        )
```

app/views/handlers/content.py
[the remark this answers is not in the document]
LJBabbage: Block type 'Summary' still comes through here, which doesn't have a content element. When I change it to your suggest it just errors on any summary
```
return self._get_safe_page_title(transformed_block ["content"]["title"])
KeyError: 'content'
```
Its not something I have changed, so can't say whether its right that the summary block is using this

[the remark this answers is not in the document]
ajmaddaford: Ah, I see it now. I don't think it's right that the `Summary` handler extends the `Content` handler, I'll raise a card for it thought rather than extending this pull request further.

app/views/handlers/content.py
```diff
@@ -36,11 +35,4 @@ def _render_block(self, block_id):
...
         if content:
-            if isinstance(content["title"], str):
-                content_title = content["title"]
-            else:
-                content_title = content["title"]["text"]
-
-            page_title = f'{content_title} - {self._schema.json["title"]}'
-
-            return safe_content(page_title)
+            return self._get_safe_page_title(transformed_block.get("content")["title"])
```
ajmaddaford, line 38:
```suggestion
            return self._get_safe_page_title(content["title"])
```

PR #165, opened 2020-06-26
Auto PR on Design System Releases

iwootten:

Automatically open PR's when there is a new version of the design system. Brings runners design system dependency in line with the way schemas are released. Setting this as a scheduled job at 4.00am (I think) for now until the design-system ci is updated to use github actions and runner can then receive dispatch events from it.

### How to review 

Confirm if the PR automatically creates a new PR with design system updated. You can see an example of the PR created here: https://github.com/ONSdigital/eq-questionnaire-runner/pull/167


iwootten:
[Automatic PR's on design system releases (M)](https://trello.com/c/LeSrwQjD/3941-automatic-prs-on-design-system-releases-m)

iwootten:
Realised this needs a branch check to ensure it doesn't open multiple prs each day until the first is closed.

MebinAbraham:
> Realised this needs a branch check to ensure it doesn't open a pr each day until the first is closed.

The branch includes the version number so it should keep updating the existing open PR instead of making a new PR. Is that what you were referring to?

.github/workflows/templates_release_pr.yml
```diff
@@ -0,0 +1,43 @@
...
+
+on:
+  schedule:
+    - cron:  '0 4 * * *'
+
+jobs:
+  create-pull-request:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Update and commit schemas version
```
MebinAbraham, line 11: Shouldn't say `schemas`.

PR #170, opened 2020-06-29
Fix empty label values on checkbox questions

berroar:
We now enforce validation of non empty string values as a result of [this  validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/41). As we no longer accept empty label values, I have updated the logic for checkbox question pages so that we display an empty label if there is only one answer option. 

### How to review 
Use `test_checkbox.json` to check the changes.

### Checklist


berroar:
[Review the use of empty string values in schemas (S)](https://trello.com/c/YKDAmHxG/3743-review-the-use-of-empty-string-values-in-schemas-s)

templates/partials/answers/checkbox.html
```diff
@@ -1,10 +1,12 @@
 {% from "components/checkboxes/_macro.njk" import onsCheckboxes %}
 
+{% set number_of_checkboxes = answer.options|length %}
```
MebinAbraham, line 3:
```suggestion
{% set number_of_checkboxes = answer.options | length %}
```

tests/functional/spec/checkbox.spec.js
```diff
@@ -7,14 +7,14 @@ describe('Checkbox with "other" option', () => {
...
+    expect($("body").getText()).to.have.string("Select all that apply");
   });
 
-  it('Given a label has been set in the schema for a checkbox answer, When the checkbox answer is displayed, Then the label "Select all that apply" should be visible', () => {
+  it("Given a label has been set in the schema for a checkbox answer, When the checkbox answer is displayed, Then the label should be visible", () => {
     $(MandatoryCheckboxPage.none()).click();
     $(MandatoryCheckboxPage.submit()).click();
-    expect($("body").getText()).to.have.string("Select all that apply");
+    expect($("body").getText()).to.have.string("Select any answers that apply");
   });
```
MebinAbraham, line 18: Do we have a test for when `number_of_checkboxes == 1`?

berroar: Not at the moment, I will add one

templates/partials/answers/checkbox.html
berroar: Any reason why? Don't tend to have gaps in the jinja docs I've seen

MebinAbraham: Personally, I think it is more readable and the majority of places in runner are separated by a space when using `|` operator. Would be nice is it was fully consistent.

templates/partials/answers/checkbox.html
```diff
@@ -1,10 +1,12 @@
 {% from "components/checkboxes/_macro.njk" import onsCheckboxes %}
 
+{% set number_of_checkboxes = answer.options | length %}
+
 {{ onsCheckboxes({
   "legend": question_title,
   "id": answer.id,
   "legendClasses": "u-vh",
-  "checkboxesLabel": '' if answer.label == '' else answer.label or _("Select all that apply"),
+  "checkboxesLabel": '' if number_of_checkboxes == 1 else answer.label or _("Select all that apply"),
```
ajmaddaford, line 9: I think if label is provided it should be used, regardless of the number of checkboxes.

berroar: I've updated the logic so the answer.label should always be used

templates/partials/answers/checkbox.html
```diff
@@ -1,10 +1,18 @@
 {% from "components/checkboxes/_macro.njk" import onsCheckboxes %}
 
+{% if answer.label %}
+  {% set checkbox_label = answer.label %}
+{% elif answer.options|length == 1%}
```
MebinAbraham, line 5:
```suggestion
{% elif answer.options|length == 1 %}
```

test_schemas/en/test_checkbox.json
```diff
@@ -150,20 +149,45 @@
...
+                                        "validation": {
+                                            "messages": {}
+                                        }
+                                    }
+                                ],
+                                "id": "single-checkbox-question",
+                                "title": "Is this an estimate?",
+                                "type": "General"
+                            },
+                            "routing_rules": []
```
MebinAbraham, line 182: I know we are not good at this in existing schema but we should added keys that we dont need.

test_schemas/en/test_checkbox.json
```diff
@@ -150,20 +149,45 @@
...
+                                        "options": [
+                                            {
+                                                "label": "Estimate",
+                                                "value": "Estimate"
+                                            }
+                                        ],
+                                        "type": "Checkbox",
+                                        "validation": {
+                                            "messages": {}
+                                        }
```
MebinAbraham, line 175: Redundant keys.

test_schemas/en/test_numbers.json
```diff
@@ -291,7 +285,7 @@
                         }
                     ],
                     "id": "test",
-                    "title": ""
+                    "title": "Title"
```
ajmaddaford, line 288: Has a card been raised to review the use of group title?

test_schemas/en/test_question_guidance.json
```diff
@@ -27,7 +27,7 @@
             "groups": [
                 {
                     "id": "group",
-                    "title": "",
+                    "title": "Group",
```
ajmaddaford, line 30: All other places where a generic group title is needed use "Title" rather than "Group". This should be changed to match so that it's easy to find and remove when we sort out the group titles issue.

test_schemas/en/test_numbers.json
berroar: I can put one in the Tech Backlog

test_schemas/en/test_routing_not_affected_by_answers_not_on_path.json
```diff
@@ -150,7 +150,7 @@
                                 "title": "This page should have been skipped!",
                                 "contents": [
                                     {
-                                        "description": ""
+                                        "description": "Description"
```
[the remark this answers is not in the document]
berroar: Removed these

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=validate-empty-stings
```
ajmaddaford, line 2:
```suggestion
tag=latest
```

PR #173, opened 2020-06-30
Make CDN URL configurable

MebinAbraham:
Added the ability to specify own CDN URL. 
After discussing with some devs, it was suggested that we should provide the full path to the CDN where any assets can live, this includes design system, look up data etc. This is important because the default CDN hosts the artefacts under `/sdc/`.

### How to review 
Ensure:
- you are able to specify your own CDN
- the default ONS CDN is used when no CDN URL is provided.

Depends on: https://github.com/ONSdigital/design-system/pull/806
Git checkout the DS system locally and build it by following the README then symlink the `templates` from DS to the runner one so that this can be tested without merging the DS PR.


app/setup.py
```diff
@@ -238,18 +228,36 @@ def after_request(response):  # pylint: disable=unused-variable
...
+    csp_policy = deepcopy(CSP_POLICY)
+    cdn_root_url = _get_cdn_root_url(application)
+    updated_policy = {}
+    for k, v in csp_policy.items():
+        updated_policy[k] = v
+
+        if k != "frame-src":
+            updated_policy[k].append(cdn_root_url)
+
+    return updated_policy
```
ajmaddaford, line 246:
```suggestion
    for policy in csp_policy:
        if policy != "frame-src":
            csp_policy[policy].append(cdn_root_url)
    return csp_policy
```

app/settings.py
```diff
@@ -38,6 +38,7 @@ def get_env_or_fail(key):
     return value
 
 
+CDN_URL = os.getenv("CDN_URL", "https://cdn.ons.gov.uk/sdc")
```
ajmaddaford, line 41: I know we discussed this for a while, but I think we need to change this. 

We have two uses of this and we mutate the string in both places - either removing `/sdc` or adding `/design-system`. If we had:

```python
CDN_URL = os.getenv("CDN_URL", "https://cdn.ons.gov.uk")
CDN_ASSETS_PATH = os.getenv("CDN_ASSETS_PATH", "/sdc/design-system")
```

Then it keeps the configuration in one place, and also doesn't enforce a `design-system` folder.

MebinAbraham: Sometimes it is easier to decide once the changes are visible. As discussed, have changed 👍

app/settings.py
```diff
@@ -38,7 +38,8 @@ def get_env_or_fail(key):
     return value
 
 
-CDN_URL = os.getenv("CDN_URL", "https://cdn.ons.gov.uk/sdc")
+CDN_URL = os.getenv("CDN_URL", "https://cdn.ons.gov.uk")
+CDN_ASSETS_PATH = os.getenv("CDN_ASSETS_PATH", "/sdc/design-system/")
```
ajmaddaford, line 42: If we make the change I suggested in the deign system pr, this should be:
```suggestion
CDN_ASSETS_PATH = os.getenv("CDN_ASSETS_PATH", "/sdc/design-system")
```

app/setup.py
```diff
@@ -228,36 +227,26 @@ def after_request(response):  # pylint: disable=unused-variable
...
-    return updated_policy
+    for policy in csp_policy:
+        if policy != "frame-src":
+            csp_policy[policy].append(cdn_url)
+    return csp_policy
 
 
 def setup_secure_headers(application):
-    updated_policies = add_cdn_url_to_csp_policies(application)
+    csp_policy = add_cdn_url_to_csp_policies(application.config["CDN_URL"])
```
ajmaddaford:
```suggestion
    csp_policy = add_cdn_url_to_csp_policy(application.config["CDN_URL"])
```

app/setup.py
```diff
@@ -228,36 +227,26 @@ def after_request(response):  # pylint: disable=unused-variable
     return application
 
 
-def _get_cdn_root_url(application) -> str:
-    parsed_url = urlparse(application.config["CDN_URL"])
-    return f"{parsed_url.scheme}://{parsed_url.hostname}"
-
-
-def add_cdn_url_to_csp_policies(application) -> Dict:
+def add_cdn_url_to_csp_policies(cdn_url) -> Dict:
```
ajmaddaford: I've made some suggestions  to fix naming inconsistencies around csp policy rather than policies
```suggestion
def add_cdn_url_to_csp_policy(cdn_url) -> Dict:
```

app/setup.py
```diff
@@ -228,36 +227,26 @@ def after_request(response):  # pylint: disable=unused-variable
...
-    for k, v in csp_policy.items():
-        updated_policy[k] = v
-
-        if k != "frame-src":
-            updated_policy[k].append(cdn_root_url)
-
-    return updated_policy
+    for policy in csp_policy:
+        if policy != "frame-src":
+            csp_policy[policy].append(cdn_url)
```
ajmaddaford, line 241: This [google doc](https://developers.google.com/web/fundamentals/security/csp#policy_applies_to_a_wide_variety_of_resources) refers to the elements of the policy as directives.
```suggestion
    for directive in csp_policy:
        if directive != "frame-src":
            csp_policy[directive].append(cdn_url)
```

MebinAbraham:
> I think some tests are needed to check that the CSP policy contains the CDN and that the CDN url is in the template helper output.

There are some tests to check the CDN exists in the csp policy header.
As for the template helper, do you mean an integration test to ensure if it within the page source?

tests/integration/test_app_create.py
```diff
@@ -108,31 +108,44 @@ def test_enforces_secure_headers(self):
...
-                "connect-src 'self' https://cdn.ons.gov.uk https://cdn.eq.census-gcp.onsdigital.uk",
-                csp_policy_parts,
-            )
+    def test_csp_policy_headers(self):
+        for cdn_url in ["https://cdn.test.domain", "https://cdn.ons.gov.uk"]:
+            with self.subTest(cdn_url=cdn_url):
+                self._setting_overrides = {"EQ_ENABLE_LIVE_RELOAD": False}
+
+                if cdn_url != "https://cdn.ons.gov.uk":
+                    self._setting_overrides["CDN_URL"] = cdn_url
```
ajmaddaford, line 117: This isn't clear. It needs to be clearer that the second test is testing `cdn_url=None`

k8s/helm/templates/deployment.yaml
```diff
@@ -89,6 +89,10 @@ spec:
               value: "{{- .Values.googleTagManagerPreview }}"
             - name: COOKIE_SETTINGS_URL
               value: "{{- .Values.cookieSettingsUrl }}"
+            - name: CDN_URL
```
pricem14pc, line 92: While the defaulted cdn.url (and assets_path) are correctly being injected, does the deploy_app.sh need extending to allow these to be overridden?

PR #180, opened 2020-07-03
Allow Question title in validation messages

iwootten:

Add the ability to use question titles in *mandatory* validation messages. Adds to all field handler types and the mutually exclusive mandatory messages (which override their child answer messages).

### How to review 

Open the schema test_question_title_in_error.json and confirm submission of empty messages on both questions displays a validation message with the question title in it.


iwootten:
[[P] Ability to add question content to error message (M)](https://trello.com/c/GkUd45Wz/3955-p-ability-to-add-question-content-to-error-message-m)

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=fix-question-level-validation
```
iwootten, line 2: Reminder to switch this back once merged.

app/forms/questionnaire_form.py
```diff
@@ -128,7 +129,11 @@ def validate_mutually_exclusive_question(self, question):
         try:
             validator(answers, is_mandatory)
         except validators.ValidationError as e:
-            self.question_errors[question["id"]] = str(e)
+            error_message = str(e)
+
+            self.question_errors[question["id"]] = FieldHandler.format_question_title(
+                error_message, question.get("title")
+            )
```
ajmaddaford, line 136: If we made `question_title` a property of questionnaire_form, we could assign it the safe value, and remove the need to do it here and in the field handler.

```suggestion
            self.question_errors[question["id"]] = error_message % dict(question_title=self.question_title)
            )
```

test_schemas/en/test_question_title_in_error.json
```diff
@@ -0,0 +1,141 @@
...
+                                                    "id": "other-answer-mandatory",
+                                                    "label": "Please specify other",
+                                                    "type": "TextField"
+                                                }
+                                            }
+                                        ],
+                                        "type": "Checkbox",
+                                        "validation": {
+                                            "messages": {
+                                                "MANDATORY_CHECKBOX": "Select an answer to `%(question_title)s’ to continue"
```
ajmaddaford, line 76:
```suggestion
                                                "MANDATORY_CHECKBOX": "Select an answer to ‘%(question_title)s’ to continue"
```

test_schemas/en/test_question_title_in_error.json
```diff
@@ -0,0 +1,141 @@
...
+                                                "label": "I prefer not to say",
+                                                "description": "Some description",
+                                                "value": "I prefer not to say"
+                                            }
+                                        ]
+                                    }
+                                ],
+                                "validation": {
+                                    "messages": {
+                                        "MANDATORY_QUESTION": "Select an answer to `%(question_title)s’ to continue"
```
ajmaddaford, line 127:
```suggestion
                                        "MANDATORY_QUESTION": "Select an answer to ‘%(question_title)s’ to continue"
```

app/forms/field_handlers/field_handler.py
```diff
@@ -43,11 +46,21 @@ def label(self):
...
+            )
+        return error_message
+
     def get_validation_message(self, message_key):
         message = self.answer_schema.get("validation", {}).get("messages", {}).get(
             message_key
         ) or self.error_messages.get(message_key)
-        return message
+
+        return self.format_question_title(message, self.question_title)
```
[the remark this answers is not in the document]
iwootten: > I don't think we need to check if there is a `question_title` placeholder in the string:

We do unfortunately. If the string contains other variables (e.g. %(min)s) and we don't provide all vars in a single call to format, it will fail with undefined vars.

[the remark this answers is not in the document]
iwootten: Additionally, it would make sense to use f-strings here and doing so highlights that using "min" and "max" should be avoided as they are built-in functions.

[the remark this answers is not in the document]
ajmaddaford: So that means `question_title` can't be used if there are other placeholders in the string?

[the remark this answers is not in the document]
iwootten: Nope. Not ideal, but unless we want to come up with a dependency resolver for them I don't see how else this can be handled. Is there anywhere we'd need this behaviour?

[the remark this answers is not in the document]
ajmaddaford: I don't think a dependency resolver would be needed. I think the question title could be passed down through the fields to the validators, to where the existing interpolation of min/max etc is done. As you've said though, this implementation is probably enough for what we need right now.

If we do keep this, then the schema validator should validate all error messages to check that if `question_text` exists in an error string, no other placeholders should.

app/forms/field_handlers/number_handler.py
```diff
@@ -77,8 +79,8 @@ def _get_number_field_validators(self):
             "messages", {}
         )
 
-        for error_key, error_message in schema_validation_messages.items():
-            answer_errors[error_key] = error_message
+        for error_key in schema_validation_messages.keys():
```
ajmaddaford, line 82: Now that `validation_messages` is a property of field handler, we can drop the `schema_validation_messages` variable and use:
```suggestion
        for error_key in self.validation_messages.keys():
```

app/forms/field_handlers/field_handler.py
```diff
@@ -43,11 +48,19 @@ def label(self):
     def guidance(self):
         return self.answer_schema.get("guidance", "")
 
+    @staticmethod
+    def format_question_title(error_message, question_title):
```
ajmaddaford, line 52: The method name would be clearer as something like `interpolate_question_title_into_error` or `insert_question_title_into_error`.

app/forms/field_handlers/field_handler.py
```diff
@@ -43,11 +48,19 @@ def label(self):
...
+        error_message = error_message or ""
```
ajmaddaford, line 53: When can `error_message` not be set?

iwootten: The existing unit tests were not setting the messages. I've updated the tests to either disable validation or use MagicMock where appropriate.

app/forms/questionnaire_form.py
```diff
@@ -123,12 +125,17 @@ def validate_mutually_exclusive_question(self, question):
...
+            messages=messages, question_title=self.question_title
+        )
 
         try:
             validator(answers, is_mandatory)
         except validators.ValidationError as e:
-            self.question_errors[question["id"]] = str(e)
+            error_message = str(e)
+
+            self.question_errors[question["id"]] = error_message
```
ajmaddaford, line 137:
```suggestion
            self.question_errors[question["id"]] = str(e)
```

app/forms/questionnaire_form.py
```diff
@@ -336,7 +343,12 @@ def get_answer_fields(question, data, error_messages, answer_store, metadata, lo
                 ).get_field()
 
         answer_fields[answer["id"]] = get_field_handler(
-            answer, error_messages, answer_store, metadata, location
+            answer,
+            error_messages,
+            answer_store,
+            metadata,
+            location,
+            question_title=question.get("title"),
```
ajmaddaford, line 351: For consistency `question_title` should also be passed to the detail answer `get_field_handler` just before this.

app/forms/validators.py
```diff
@@ -405,13 +405,21 @@ def format_playback_value(value, currency=None):
     return format_number(value)
 
 
+def format_message_with_title(error_message, question_title):
+    return error_message % dict(question_title=question_title)
```
ajmaddaford, line 409: Given that the `safe_content` call is needed in multiple places, I think it would make sense to do it here instead.

iwootten:
> I might be reading the card wrong, but I was expecting the question text to be visually hidden and wrapped in the u-vh class.

This is for the feature. The schemas are written in a way to make it obvious for the developer, the wrapped titles can be added as appropriate for the census schema.

app/forms/questionnaire_form.py
```diff
@@ -26,6 +26,7 @@ def __init__(
         self.location = location
         self.question_errors = {}
         self.options_with_detail_answer = {}
+        self.question_title = self.question.get("title") or ""
```
MebinAbraham, line 29:
```suggestion
        self.question_title = self.question.get("title", "")
```

test_schemas/en/test_question_title_in_error.json
```diff
@@ -0,0 +1,141 @@
...
+                                        "type": "Checkbox",
+                                        "validation": {
+                                            "messages": {
+                                                "MANDATORY_CHECKBOX": "Select an answer to ‘%(question_title)s’ to continue"
+                                            }
+                                        }
+                                    }
+                                ],
+                                "id": "mandatory-checkbox-question",
+                                "title": "Which pizza toppings would you like?",
```
MebinAbraham, line 82: We should have an example and test of the question title in error message when the title has a placeholder.

app/forms/validators.py
```diff
@@ -405,13 +406,21 @@ def format_playback_value(value, currency=None):
     return format_number(value)
 
 
+def format_message_with_title(error_message, question_title):
```
LJBabbage, line 409: This is very minor, but I didn't immediately pick up what this method was doing from its name, when I looked at it I realised it was just formatting the error_message, perhaps format_error_message_with_title might help developers like me.

test_schemas/en/test_question_title_in_error.json
```diff
@@ -0,0 +1,141 @@
+{
```
LJBabbage, line 1: I little bit verbose, it could easily be one textfield with an error message

tests/app/forms/field_handlers/test_field_handler.py
```diff
@@ -50,3 +50,26 @@ def test_get_mandatory_validator_mandatory_with_error():
 
     assert isinstance(validate_with, ResponseRequired)
     assert validate_with.message == "This is the mandatory message for an answer"
+
+
+def test_get_mandatory_validator_mandatory_with_question_in_error():
```
LJBabbage, line 55: Mandatory twice in name?

LJBabbage, line 55: Nice tongue twister, but I can see its the format of the other tests

test_schemas/en/test_question_title_in_error.json
iwootten: Each field uses a field handler to do the formatting, while the mutually exclusive uses a different approach in the questionnaire form. This schema tests both those approaches.

app/forms/validators.py
iwootten: It could be, but then there isn't a use case where we need it to be so flexible and would require additional unit tests. I think keeping it with the one kwarg is fine until we have a need for anything more complex.

PR #185, opened 2020-07-07
Add custom submission guidance to hub

LJBabbage:
This PR adds submission guidance to the hub. The validation branch which goes with this PR is add-submission-guidance-to-hub

### How to review 
Open up test_hub_and_spoke_custom_content and make sure the submission guidance appears in the right place and the right time

N.B Formatting of the hub might need revisiting, I have placed the new element in the right place, but names of containers and margin/borders, might need revisiting by the design team


app/views/contexts/hub_context.py
```diff
@@ -50,18 +50,21 @@ def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
...
             ) or lazy_gettext("Submit survey")
+            submission_guidance = hub_schema.get("submission", {}).get("guidance", "")
 
         else:
             title = lazy_gettext("Choose another section to complete")
             guidance = custom_text.get("guidance") or lazy_gettext(
                 "You must complete all sections in order to submit this survey"
             )
             submit_button = lazy_gettext("Continue")
+            submission_guidance = ""
```
MebinAbraham, line 61: We should use `NoneType` over empty strings for non-existent keys. This also makes it consistent with the current pattern.
`submission_guidance = hub_schema.get("submission", {}).get("guidance")`
`submission_guidance = None`

templates/hub.html
```diff
@@ -28,11 +28,16 @@ <h1 class="u-mt-m">{{ content.title }}</h1>
...
+      <div class="u-mt-l">
+        <p class="u-mb-no" data-qa="submission-guidance">{{ content.submission_guidance }}</p>
+      </div>
+    {% endif %}
     {{
         onsButton({
             "text": content.submit_button,
             "submitType": 'timer',
-            "classes": 'u-mb-m u-mt-l',
+            "classes": 'u-mb-m u-mt-' + ("s" if content.submission_guidance else "xl"),
```
ajmaddaford, line 40:
```suggestion
            "classes": 'u-mb-m u-mt-' ~ ("s" if content.submission_guidance else "xl"),
```

PR #186, opened 2020-07-07
Remove expects from middle of tests

iwootten:

Remove interim expects from middle of functional tests which serve no purpose. Currently one such expect is unnecessarily testing an age.

### How to review 

Check the tests continue to pass.

### DIFF
--- a/tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
+++ b/tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
@@ -193,15 +193,7 @@ describe("Feature: Repeating Sections with Hub and Spoke", () => {
       $(ProxyPage.noIMAnsweringForMyself()).click();
       $(ProxyPage.submit()).click();
 
-      $(DateOfBirthPage.day()).setValue("07");
-      $(DateOfBirthPage.month()).setValue("07");
-      $(DateOfBirthPage.year()).setValue("1970");
-      $(DateOfBirthPage.submit()).click();
-
-      $(ConfirmDateOfBirthPage.confirmDateOfBirthYesIAmAgeOld()).click();
-      $(ConfirmDateOfBirthPage.submit()).click();
-
-      expect($(SexPage.questionText()).getText()).to.equal("What is your sex?");
+      expect($(DateOfBirthPage.questionText()).getText()).to.equal("What is your date of birth?");
     });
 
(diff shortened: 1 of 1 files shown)


tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
```diff
@@ -214,8 +209,6 @@ describe("Feature: Repeating Sections with Hub and Spoke", () => {
       $(ProxyPage.yes()).click();
       $(ProxyPage.submit()).click();
 
-      expect($(DateOfBirthPage.questionText()).getText()).to.equal("What is Samuel Clemens’ date of birth?");
-
```
ajmaddaford, line 218: These tests are testing that the question text is proxy or not, so we should keep this expect and remove the rest of the lines of this test. I don't think we need to check two separate pages for the proxy text.

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
```diff
@@ -194,15 +193,11 @@ describe("Feature: Repeating Sections with Hub and Spoke", () => {
       $(ProxyPage.noIMAnsweringForMyself()).click();
       $(ProxyPage.submit()).click();
 
-      expect($(DateOfBirthPage.questionText()).getText()).to.equal("What is your date of birth?");
-
```
ajmaddaford, line 198: These tests are testing that the question text is proxy or not, so we should keep this expect and remove the rest of the lines of this test. I don't think we need to check two separate pages for the proxy text.

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
iwootten: As discussed, each section needs to be completed for subsequent tests to follow. We can only therefore remove the interim expects.

iwootten:
[Bug: Functional test failure due to date based test](https://trello.com/c/dCsNqbya/3993-bug-functional-test-failure-due-to-date-based-test)

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
ajmaddaford: 🤦

PR #189, opened 2020-07-08
Change Error page titles

lauraedwards1:
Any pages where error validation applies, insert 'Error: ' in front of page title 
[trello card](https://trello.com/c/jVdmjwyc)

### How to review 
"Error: " added infront of page title when failed validation

### Checklist


app/views/handlers/question.py
```diff
@@ -128,6 +128,9 @@ def get_context(self):
                 )
             )
 
+        if self.form.errors or self.form.question_errors:
+            self.page_title = "Error: " + self.page_title
```
ajmaddaford, line 132: - This needs to be translatable, so should be wrapped in a call to `gettext` (this is done [here](https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/forms/field_handlers/dropdown_handler.py#L12))
- We should use a placeholder for the page title so that a translator has more context

```suggestion
            self.page_title = gettext("Error: {page_title}").format(page_title=self.page_title)
```

app/views/handlers/question.py
```diff
@@ -128,6 +129,11 @@ def get_context(self):
                 )
             )
 
+        if self.form.errors or self.form.question_errors:
+            self.page_title = gettext("Error: {page_title}").format(
+                page_title=self.page_title
+            )
+
```
ajmaddaford, line 136: We should have a test for this. We use integration tests to test the handlers, so I think the best place for it would be in the `tests/integration/questionnaire` folder. As we will be doing more work to make the page titles consistent, it's probably worth creating a new `test_page_titles.py` test file in that folder.

tests/integration/questionnaire/test_page_titles.py
```diff
@@ -0,0 +1,13 @@
...
+class TestPageTitles(IntegrationTestCase):
+    def test_should_have_error_in_page_title_when_fail_validation(self):
+        # Given
+        self.launchSurvey("test_checkbox")
+        # When
+        self.post()
+        # Then
+        self.assertEqualPageTitle(
+            "Error: Which pizza toppings would you like? - Other input fields"
+        )
```
ajmaddaford, line 13: Sorry, not sure how I missed this, but there is already a test file that tests page titles - `test_questionnaire_page_titles.py`, this test should be in that file.

PR #192, opened 2020-07-09
Merge Error Messages vs Overwrite

iwootten:

Updated messages to be merged for field handlers, where previously they used either the user defined or default dictionary.

### How to review 

Open the the test_dob_date schema and submit with an invalid year (eg 1500). Previously this would have failed.


iwootten:
[Bug: Date validation broken when using custom error messages](https://trello.com/c/6gUQBTud/3978-bug-date-validation-broken-when-using-custom-error-messages)

PR #194, opened 2020-07-10
Add interviewer note

petechd:
This adds interviewer note support in runner to interstitial and question block. The change is described on [Trello card](https://trello.com/c/rWIoaGMJ). It requires [validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/49) to be merged first.

### How to review 
New test schema (`test_interviewer_note`) has been added. Run new functional test using `./node_modules/.bin/wdio tests/functional/wdio.conf.js --spec tests/functional/spec/interviewer_note.spec.js`.

### Checklist


petechd:
[(I) Interviewer Note highlighting on heading (S)](https://trello.com/c/rWIoaGMJ/3975-i-interviewer-note-highlighting-on-heading-s)

templates/partials/question.html
```diff
@@ -3,8 +3,12 @@
 {% from "components/panel/_macro.njk" import onsPanel %}
 
 {% set form = content.form %}
-
-{% set question_title = question.title %}
+{% set interviewer_note = _("Interviewer Note:") %}
+{% if question.interviewer_note %}
+  {% set question_title = '<mark class="instruction">' + interviewer_note + '</mark>' + question.title %}
+{% elif question.title %}
+  {% set question_title = question.title %}
```
iwootten, line 10: This should be formatted in a function, the same as the description and instruction are. I think then you may be able to reuse it on the interstitial partial.

templates/interstitial.html
```diff
@@ -6,8 +6,10 @@
 
 {% block form_content %}
   {% set interstitial_instruction = format_paragraphs(block.content.instruction) if block.content.instruction else None %}
-  {% if block.content.title %}
-    <h1>{{ block.content.title }}</h1>
+  {% if block.content.interviewer_note %}
+      <h1 class="question__title"><mark class="instruction">{{_("Interviewer Note:")}}</mark> {{ block.content.title }}</h1>
+  {% elif block.content.title %}
+      <h1>{{ block.content.title }}</h1>
```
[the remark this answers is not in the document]
petechd: The reason is "interstitial" block uses plain h1 when "question" block is using h1 with a `question__title` class. If you check [design system 19.1.0 release files](https://github.com/ONSdigital/design-system/pull/809/files) you can see that  "Interviewer Note" only works when preceded by `question__title` class.

templates/partials/question.html
petechd: Is this ok to keep all this logic in frontend? I made a different variant of this change where the logic is in `question.py` and `content.py`..

templates/interstitial.html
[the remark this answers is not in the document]
iwootten: Doesn't that change mean that the content title can have the same class?

templates/partials/question.html
iwootten: It sounds like this should actually really be a helper macro and we should be avoiding using jinja filters.

templates/partials/question.html
```diff
@@ -3,8 +3,12 @@
 {% from "components/panel/_macro.njk" import onsPanel %}
 
 {% set form = content.form %}
-
-{% set question_title = question.title %}
+{% set interviewer_note = _("Interviewer Note:") %}
+{% if question.interviewer_note %}
+  {% set question_title = '<mark class="instruction">' + interviewer_note + '</mark>' + question.title %}
```
LJBabbage, line 8: Not sure we should be updating the title like this. It feels all of this should be in the macro. The extra param should be passed through onsQuestion and then all it needs is a simple 'if' around mark

test_schemas/en/test_interviewer_note.json
```diff
@@ -0,0 +1,121 @@
...
+                                "title": {
+                                    "text": "Can you confirm that {team} is your favourite team?",
+                                    "placeholders": [
+                                        {
+                                            "placeholder": "team",
+                                            "value": { "source": "answers", "identifier": "favourite-team-answer" }
+                                        }
+                                    ]
+                                },
+                                "interviewer_note": false,
```
LJBabbage, line 90: Not sure we want to go down this route, it would suggest we have to add it to ever question in every schema, and I don't fancy doing that. We do have this pattern of just using true in other places, you could enforce it in validator, but I don't believe we do that anywhere else mind.

tests/functional/spec/interviewer_note.spec.js
```diff
@@ -0,0 +1,32 @@
+import InitialInterstitialPage from "../generated_pages/interviewer_note/initial-interstitial-block.page.js";
```
LJBabbage, line 1: I believe we are keeping these in alphabetically order

tests/functional/spec/interviewer_note.spec.js
```diff
@@ -0,0 +1,32 @@
...
+import FavouriteTeamPage from "../generated_pages/interviewer_note/favourite-team-block.page.js";
+import ConfirmPage from "../generated_pages/interviewer_note/confirm-block.page.js";
+import FinalInterstitialPage from "../generated_pages/interviewer_note/final-interstitial-block.page.js";
+
+describe("Interviewer Note", () => {
+  beforeEach(() => {
```
LJBabbage, line 7: So I wouldn't use beforeEach, before would be better. This would mean you don't repeat each step, the next 'it' will just follow on where you  left off form the last 'it' You naturally would have to move the Given I started out of each test, but it will make this test a little shorter and quicker.

templates/interstitial.html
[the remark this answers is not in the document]
petechd: I added this class to both content titles. Otherwise "Interviewer Note" wouldn't work.

templates/macros/helpers.html
```diff
@@ -27,3 +27,15 @@
   } %}
 <a href="mailto:{{email}}{% if subject %}?subject={{ subject|urlencode }}{% endif %}" {{attr|xmlattr}}>{{email}}</a>
 {%- endmacro -%}
+
+{%- macro interviewer_note(interviewer_only='False', title='', type='') -%}
+  {% if (interviewer_only and (type == 'Interstitial')) %}
```
iwootten, line 32: The brackets here are unnecessary.

templates/macros/helpers.html
```diff
@@ -27,3 +27,15 @@
...
+    <h1 class="question__title"><mark class="instruction">{{_("Interviewer Note:")}}</mark> {{ title }}</h1>
+  {% elif (title and (type == 'Interstitial')) %}
```
iwootten, line 34: Again, the brackets shouldn't be needed.

templates/partials/question.html
petechd: I've added helper macro now, thanks!👍

templates/macros/helpers.html
```diff
@@ -27,3 +27,15 @@
   } %}
 <a href="mailto:{{email}}{% if subject %}?subject={{ subject|urlencode }}{% endif %}" {{attr|xmlattr}}>{{email}}</a>
 {%- endmacro -%}
+
+{%- macro interviewer_note(interviewer_only='False', title='', type='') -%}
```
iwootten, line 31: Why is interviewer_only set as a string?

test_schemas/en/test_interviewer_note.json
```diff
@@ -0,0 +1,121 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "0",
+    "title": "Interviewer Note Pages",
+    "theme": "default",
+    "description": "A questionnaire to demo interviewer note pages.",
```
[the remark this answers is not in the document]
petechd: Changed, thanks!👍

tests/functional/spec/interviewer_note.spec.js
petechd: In order now, thanks!

tests/functional/spec/interviewer_note.spec.js
petechd: Changed, thanks!

tests/functional/spec/interviewer_note.spec.js
petechd: Done, thanks!

templates/interstitial.html
```diff
@@ -1,14 +1,14 @@
...
 {% set save_on_signout = true %}
 
 {% set continue_button_text = _("Continue") %}
 
 {% block form_content %}
   {% set interstitial_instruction = format_paragraphs(block.content.instruction) if block.content.instruction else None %}
-  {% if block.content.title %}
-    <h1>{{ block.content.title }}</h1>
-  {% endif %}
+  {{ helpers.interviewer_note(block.content.interviewer_only, block.content.title, block.type) }}
```
iwootten, line 11: This feels like it should be only called if interviewer note formatting is required. It would help simplify the macro to move the interviewer_only and interstitial checks to this partial.

templates/partials/question.html
petechd: Moved to macro, thanks!

templates/macros/helpers.html
petechd: Removed👍

templates/macros/helpers.html
petechd: Removed👍

templates/macros/helpers.html
petechd: This is gone now👍

templates/interstitial.html
petechd: I had to keep block type check in macro but otherwise it's all done(I hope)👍

templates/partials/question.html
```diff
@@ -5,7 +5,12 @@
 {% import 'macros/helpers.html' as helpers %}
 
 {% set form = content.form %}
-{% set question_title= helpers.interviewer_note(question.interviewer_only, question.title) %}
+
+{% if question.interviewer_only %}
+  {% set question_title= helpers.interviewer_note(question.title) %}
+{% else %}
+  {% set question_title = question.title %}
+{% endif %}
```
iwootten, line 13: It would be good if this was formatted as per the question_description, question_instruction calls below. Additionally, you should only need to import the interviewer_note macro, rather than the helpers partial.

petechd: Formatted, imports sorted, thanks!👍

test_schemas/en/test_interviewer_note.json
petechd: We did this with `test_radio_detail_answer_visible` schema before, I was told by other devs to check `false` in `radio_visible_answers.spec.js` so I follow the same pattern here.

templates/interstitial.html
```diff
@@ -6,8 +6,10 @@
 
 {% block form_content %}
   {% set interstitial_instruction = format_paragraphs(block.content.instruction) if block.content.instruction else None %}
-  {% if block.content.title %}
-    <h1>{{ block.content.title }}</h1>
+  {% if block.content.interviewer_note %}
```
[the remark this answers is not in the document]
petechd: I checked and you can still have a `no title` interstitial page. I didn't want to use double nested `if` statements as it wouldn't be approved by other devs.

templates/macros/helpers.html
```diff
@@ -27,3 +27,11 @@
   } %}
 <a href="mailto:{{email}}{% if subject %}?subject={{ subject|urlencode }}{% endif %}" {{attr|xmlattr}}>{{email}}</a>
 {%- endmacro -%}
+
+{%- macro interviewer_note(title='', type='') -%}
+  {% if type == 'Interstitial' %}
+    <h1 class="question__title"><mark class="instruction">{{_("Interviewer Note:")}}</mark> {{ title }}</h1>
```
iwootten, line 33: I feel like "Interviewer Note" text is something that should really be defined and not only in the templates. May be worth seeing if any of the rest of the team have an opinion.

ajmaddaford: I don't see an issue with it being in the templates, it's done like that in plenty of other places.

templates/partials/question.html
```diff
@@ -2,9 +2,13 @@
...
 {% from "components/panel/_macro.njk" import onsPanel %}
 
+{% from 'macros/helpers.html' import interviewer_note %}
+
 {% set form = content.form %}
 
-{% set question_title = question.title %}
+
+{% set title= interviewer_note(question.title) if block.interviewer_only else question.title %}
+{% set question_title= question.title %}
```
iwootten, line 11: Not sure why the assignment is made to title, then also to question_title? Can't the reference to question_title be replaced with question.title below?

petechd: This change was discussed with @ajmaddaford, question_title is used by more than just the question template and in question templates it's used by legend html element so it needs to stay what it was before.

ajmaddaford: `question_title` is used by most of the answer field templates.

PR #195, opened 2020-07-10
Add submission and launch language to downstream

LJBabbage:
The launch language and the submission code now needs adding to the downstream payload

### How to review 
Make sure it works for normal submission and flusher/dumper


LJBabbage:
> What about submission language code?. Only launch has been added?

Opps, I had it originally in submission only to put it in convert_answers to think thats a daft place to put it, as it only ever used in submission. I obviously didn't restore it

app/views/handlers/submission.py
```diff
@@ -13,14 +13,18 @@ def __init__(self, schema, questionnaire_store, full_routing_path):
...
-        message = json.dumps(
-            convert_answers(
-                self._schema, self._questionnaire_store, self._full_routing_path
-            ),
-            for_json=True,
+
+        payload = convert_answers(
+            self._schema, self._questionnaire_store, self._full_routing_path
         )
+        payload[
```
JamesGardiner, line 23: this can be done inside the `convert_answers` method

app/views/handlers/submission.py
```diff
@@ -38,9 +42,7 @@ def submit_questionnaire(self):
...
 
-    @staticmethod
-    def _store_submitted_time_in_session():
-        session_store = get_session_store()
-        session_data = session_store.session_data
+    def _store_submitted_time_in_session(self):
+        session_data = self.session_store.session_data
         session_data.submitted_time = datetime.utcnow().isoformat()
-        session_store.save()
+        self.session_store.save()
```
JamesGardiner, line 48:
```suggestion
        self._session_store.save()
```

app/views/handlers/submission.py
LJBabbage: convert_answers  doesn't feel like the right location, as it's used by both flusher and dumper as well. In those 2 instances language_submission_code should not be present in the payload and therefore having it there would need extra logic for 2 out of 3 situations.

JamesGardiner: Just saw your previous comment 🤦

LJBabbage:
> There should be a test for `submission_language_code` in the submission data.

I did consider it, but its just the language code from session store (self.session_store.session_data.language_code) which is tested in multiple other locations to make sure it is correct. We naturally can test it, but being in submit_questionnaire I could do with a pointer in how to grab hold to have a look

ajmaddaford:
> > There should be a test for `submission_language_code` in the submission data.
> 
> I did consider it, but its just the language code from session store, which is tested in multiple other locations to make sure it is correct. We naturally can test it, but being in submit_questionnaire I could do with a pointer in how to grab hold to have a look

The test would be to make sure it's in the payload. We could add a `get_payload` method to the handler and test that.

LJBabbage:
> Admittedly convert_answers looks a bit busy (or at least misnamed) but could the submit_language_code be assigned within the existing payload as with launch_language_code (and the other top level attributes)? That way it would also be included when invoking /dump/submission? I take it session_store isn't in scope in converter.py so it would need to be passed in separately.
> 

So you wouldn't need to pass in the whole session_store, just the language from submission, put a default on it and then write some logic in convert_answers. Although you are right convert_answers isn't the best of names and I reckon it might tip if I put an extra 'if' in there (just did a quick test and it will) So it would need a refactor. I am easy, either or

pricem14pc:
> > Admittedly convert_answers looks a bit busy (or at least misnamed) but could the submit_language_code be assigned within the existing payload as with launch_language_code (and the other top level attributes)? That way it would also be included when invoking /dump/submission? I take it session_store isn't in scope in converter.py so it would need to be passed in separately.
> 
> So you wouldn't need to pass in the whole session_store, just the language from submission, put a default on it and then write some logic in convert_answers. Although you are right convert_answers isn't the best of names and I reckon it might tip if I put an extra 'if' in there (just did a quick test and it will) So it would need a refactor. I am easy, either or, just felt it only gets used in submission so kept it there.

Happy to leave it where it is. Only minor issue then is that /dump/submission doesn't write it out (which you could argue it should). On the plus side /flush also doesn't write it out (which as you point out it probably shouldn't)

ajmaddaford:
If Lewis refactors the adding of the submission language code into a `get_payload` method as suggested, that method could be called from the dump submission endpoint.

tests/app/views/handlers/test_submission_handler.py
```diff
@@ -0,0 +1,50 @@
+from unittest.mock import Mock
```
LJBabbage, line 1: Could argue this isn't necessary now dumper is using the handler payload method

PR #201, opened 2020-07-16
Add thank you pages for census, HH, IR, CE

LJBabbage:
There are now three distinct versions of the census page (individual responses, communal establishments household) This PR not only adds them but separates census and the default theme thank you. 

### How to review 
Use the new 3 test_schemas and make sure each one returns the correct template and text


tests/functional/spec/language_code.spec.js
```diff
@@ -124,7 +123,6 @@ describe("Language Code", () => {
     $(SummaryPage.submit()).click();
 
     expect(browser.getUrl()).to.contain("thank-you");
-    expect($(ThankYouPage.submissionSuccessfulTitle()).getText()).to.contain("Diolch am gyflwyno eich cyfrifiad");
```
LJBabbage, line 127: These checks are unnecessary and language checks are already done in integration tests.

templates/census-thank-you.html
```diff
@@ -0,0 +1,36 @@
...
+    {% endcall %}
+
+    <div class="panel__body">
+      <p class="icon--lock">{{ _("Your personal information is protected by law and will be kept confidential.") }}</p>
+    </div>
+
+    <div class="panel panel--warn panel--simple" >
+      <span class="panel__icon" aria-hidden="true">!</span>
+      <div class="panel__body">
+        <p>{{ _("Make sure you leave this page or close your browser if using a shared device.") }}</p>
```
MebinAbraham, line 32: This needs to be a link according to the card:
```
Clicking the "leave this page" link signs out and redirects to account_service_logout_url
```

app/views/handlers/thank_you.py
```diff
@@ -0,0 +1,51 @@
...
+from app.globals import get_session_store
+
+DEFAULT_THANK_YOU_TEMPLATE = "thank-you"
+CENSUS_THANK_YOU_TEMPLATE = "census-thank-you"
+
+CENSUS_TYPE_MAPPINGS = {
+    "household": "HH",
+    "communal_establishment": "CE",
+    "individual": "IR",
+}
```
JamesGardiner, line 16: Any reason you chose to make these module level variables instead of class attributes?

templates/census-thank-you.html
```diff
@@ -0,0 +1,36 @@
...
+
+  {% call onsPanel({
+    "type": "success",
+    "classes": "u-mb-s"
+    })
+  %}
+    <h1 data-qa="title"><span class="icon--check-green"></span>{{ _("Thank you for completing your census") }}</h1>
+
+      {% if content.census_type %}
+        {% if content.census_type =="IR" %}
```
JamesGardiner, line 14: This logic could be handled in the context handler, and just have the string passed in to the template

templates/thank-you.html
```diff
@@ -4,35 +4,33 @@
...
+        {{ _("Your answers were submitted for <span>{ru_name}</span> ({trading_as_name}) on {submitted_date_time}").format(
+          ru_name = content.ru_name,
+          trading_as_name = content.trad_as,
+          submitted_date_time=content.submitted_time|format_datetime) }}
+      {% else %}
+        {{ _("Your answers were submitted for <span>{ru_name}</span> on {submitted_date_time}").format(
+          ru_name = content.ru_name,
+          submitted_date_time=content.submitted_time|format_datetime) }}
+      {% endif %}
+    </p>
```
JamesGardiner, line 27: This logic should be handled in the context handler and the resulting string passed to the template

templates/census-thank-you.html
LJBabbage: Problem is I would need to have the <strong> in the backend. I had a quick chat with Andrew and we decided that the lesser of 2 evils was the logic in the HTML

app/views/handlers/thank_you.py
```diff
@@ -0,0 +1,51 @@
...
+        return build_census_thank_you_context(
+            self._cookie_session.get("display_address"), census_type_code
+        )
+
+    def get_template(self):
+        return (
+            CENSUS_THANK_YOU_TEMPLATE
+            if self._is_census_theme
+            else DEFAULT_THANK_YOU_TEMPLATE
+        )
```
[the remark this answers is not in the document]
LJBabbage: I was 50/50, but happy to change it

templates/census-thank-you.html
```diff
@@ -0,0 +1,36 @@
...
+{% from 'components/panel/_macro.njk' import onsPanel %}
+
+{% block main %}
+
+  {% call onsPanel({
+    "type": "success",
+    "classes": "u-mb-s"
+    })
+  %}
+    <h1 data-qa="title"><span class="icon--check-green"></span>{{ _("Thank you for completing your census") }}</h1>
```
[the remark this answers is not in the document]
LJBabbage: Good spot, I didn't see that

app/views/handlers/thank_you.py
```diff
@@ -0,0 +1,51 @@
...
+
+        self._cookie_session = cookie_session
+        self._is_census_theme = cookie_session.get("theme") in [
+            "census",
+            "census-nisra",
+        ]
+
+    def get_context(self):
+        if not self._is_census_theme:
+            return build_default_thank_you_context(self.session_data)
```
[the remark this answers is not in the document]
LJBabbage: We are very mixed on this point. I followed the pattern that most of handlers currently do and quite a few are also single use as well. As we have a definitive folder structure for contexts I am incline to store all the context there, I feel it makes it easier and quicker to find

templates/census-thank-you.html
```diff
@@ -0,0 +1,42 @@
...
+    {% endcall %}
+
+    <div class="panel__body">
+      <p class="icon--lock">{{ _("Your personal information is protected by law and will be kept confidential.") }}</p>
+    </div>
+
+    <div class="panel panel--warn panel--simple" >
+      <span class="panel__icon" aria-hidden="true">!</span>
+      <div class="panel__body">
+        <p>{{ _("Make sure you <a href='{url}'>leave this page</a> or close your browser if using a shared device.").format(url = "/sign-out")  }}</p>
```
MebinAbraham, line 38: This is in bold in the design.

test_schemas/en/test_thank_you_census_communal_establishment.json
```diff
@@ -0,0 +1,66 @@
+{
```
MebinAbraham, line 1: Shame we don't have unit tests for the handlers, otherwise schemas like these could be avoided.

MebinAbraham, line 1: Actually, I was mistaken, we do have unit tests for handlers. Therefore, in my opinion, we should write unit tests for the `ThankYou` handler instead.

LJBabbage: Not really, I think we have one for the block factory. I think we need to be consistent and something that should be brought up in tech session if we want to do it. This does have coverage, and if we do want tests on all handlers you can put my face on it

MebinAbraham: We started adding them, i.e. the Submission handler you recently introduced, so I don't see why we wouldn't want unit test for newly introduced ones?. 

A lot of the existing one don't have them because they weren't very unit testable at the point it was introduced due to app dependencies. We should add a card for that, but this was around that we started adding it, so why wouldn't we want unit test for new ones?

LJBabbage: I wouldn't say we have one for the submission handler, it just checks the submission_language_code nothing else and thats only because I couldn't get hold of it initially. I actually nearly deleted it when dumper was changed to use it then giving coverage 

The only other one, isn't actually testing any handlers, it only tests the factory and only for exceptions (a bit weird). With my pragmatic hat on I don't think removing tests and schemas that match the pattern for everything else and give us coverage, to write new ones is great use of time after the work has been done. However I would support this approach going forward if that decision is made before a story is started. Will get Mark to confirm this is a sensible approach.

app/views/handlers/submission.py
```diff
@@ -34,6 +34,7 @@ def submit_questionnaire(self):
         if not submitted:
             raise SubmissionFailedException()
 
+        cookie_session["display_address"] = metadata.get("display_address")
```
ajmaddaford, line 37: The card specifies the address should be stored in the session rather than the cookie. Although not implemented yet, storing this in the cookie doesn't align with the decisions made in https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/doc/architecture/decisions/0010-cookies.md.

test_schemas/en/test_thank_you_census_communal_establishment.json
MebinAbraham: We don't usually specify in the card to "write unit test", it a given. I fold.

PR #208, opened 2020-07-22
Spike immutable questionnaire schema

MebinAbraham:
Currently, the questionnaire schema is mutable, and in most places that we mutate any part of the schema JSON, we take a `deepcopy` to prevent mutating the original object. However,  it is still possible to accidentally mutate it, and this could cause several issues for any future uses of the object.

This PR makes the questionnaire schema JSON immutable, which means a copy must be made when we need to mutate the object.

---

Libraries investigated:
- [immutables.Map by MagicStack](https://github.com/MagicStack/immutables).
- [ImmutableDict by werkzeug](https://werkzeug.palletsprojects.com/en/1.0.x/datastructures/#werkzeug.datastructures.ImmutableDict).

#### immutables.Map vs ImmutableDict

The main reason for trying `immutables.Map` was for its ability to mutate and return an immutable copy of the object you are working with. For example:
```py
map = immutables.Map(a=1, b=2)
map2 = map.set('a', 10)
print(map, map2)
# will print:
#   <immutables.Map({'a': 1, 'b': 2})>
#   <immutables.Map({'a': 10, 'b': 2})>
```
To achieve the same thing when using `ImmutableDict`, we would have to do the following:
```py
map = ImmutableDict({"a":1, "b":2})
map2 = deepcopy(map) # make a deep copy
map2 = dict(map2) # turn it into a dict so that it is mutable. Note: this only makes the top level keys mutable.
map2["a"] = 10
print(map, map2)
# will print:
#   ImmutableDict({'a': 1, 'b': 2})
#   {'a': 10, 'b': 2} (would need to convert it back to ImmutableDict if required)
```

Running the same [benchmark](https://gist.github.com/1st1/292e3f0bbe43bd65ff3256f80aa2637d) as MagicStack but for `ImmutableDict` showed that deepcopying `ImmutableDict` is an expensive operation and considerably slower than `immutables.Map().set()`.

---

Basic implementation using immutables.Map: https://github.com/ONSdigital/eq-questionnaire-runner/commit/461994025d69e2e83232a421b464d6cc722bf23a
Basic implementation using ImmutableDict: [`e6ee830` (#208)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/208/commits/e6ee830e8f772510ab7d0e778b573c8f076acbc8)

---

I also investigated passing a custom object hook when calling `json.load()` however, this only work for JSON objects and not lists, therefore I decided to keep the serialization all in 1 place. There are further changes to be made. However, I did the initial work to demo an immutable `QuestionnaireSchema`.

In this PR, the latest implementation makes the questionnaire JSON immutable on initialization using `ImmutableDict`. I went `ImmutableDict` for the following reasons:
- any performance costs should be negligible since the questionnaire schema is cached.
- one less dependency.
- don't have to check two instances. `isinstance(a, dict)` works with both `dict` and `ImmutableDict`, whereas, for `immutables.Map`, we would require `isinstance(a, (dict, immutables.Map))`.
- get the initial work done and make schemas immutable, then potentially look at `immutables.Map` in future.
- with the proposed changed below, we will most likely not be making the most out fo `immutates.Map().set()` to warrant the extra dependency.

### Suggestions
1. As a first pass, look at making runner work with `ImmutableDict`.
1. Investigate the mutations currently done in `QuestionnaireSchema`. We currently mutate schema on initialization to add `parent_id` into relevant blocks and also group questions/answers by id. We should look at changing the approach and/or look at the uses of it, as I don't think some are even used any more.
1. Look at reducing the number of places we currently need to mutate the schema object.
1. Decide on how schema objects are passed around and handled in runner. One issue that was highlighted when making the schema object immutable was that it was hard to follow and understand, at what points in our codebase we have an immutable object vs a mutable object. Even with the current changes, there is a mix of mutable and immutable objects being passed around. Therefore, we should look to make this consistent in a future card.
	- As a further piece of work, we should consider adding python typings to highlight when a schema object is mutable/immutable. This should make the codebase more readable and make writing new code simpler.

> Will discuss further in tech session
### How to review 
- Does the suggested approach make sense to take forward?.
- Are there any other libraries that we should look at?

### Checklist


LJBabbage:
> Running the same benchmark as MagicStack but for ImmutableDict showed that deepcopying ImmutableDict is an expensive operation and considerably slower than immutables.Map().set()

Probably needs a slight clarification, as at low numbers (<30 items) deepcopy is comparible and in fact more efficient around the 5 item mark. Although not relevant for our schema, we don't want to get into the habbit of  saying ImmutableDict is always better. Looks to me break point is around 50 items, when deepcopy really starts to struggle.

MebinAbraham:
> > Running the same benchmark as MagicStack but for ImmutableDict showed that deepcopying ImmutableDict is an expensive operation and considerably slower than immutables.Map().set()
> 
> Probably needs a slight clarification, as at low numbers (<30 items) deepcopy is comparible and in fact more efficient around the 5 item mark, looking at those benchmark tests. Although not relevant for our schema, we don't want to get into the habbit of saying ImmutableDict is always better. Looks to me break point is around 50 items
> 
> Looks good and on running everything worked as expected

I agree, and especially for our use cache it shouldn't be a massive deal anyway since the schema is cached on load.

ajmaddaford:
I think keeping the first pass simple and using `ImmutableDict` is sensible. It achieves the aim of protecting unintended modification of the questionnaire schema with the minimum of impact on dependent code. 

If we need to keep the augmentation of the parent ids into the schema, then we should do that before making it immutable.

MebinAbraham:
Closing the Spike with current context, will open new PR updated details.

MebinAbraham:
Closing the Spike with current context, will open new PR updated details.

PR #210, opened 2020-07-22
Bug return to summary

LJBabbage:
This PR fixed the bug regarding return_to_summary when show_summary_on_completion was false and also introduces the return_to_final_summary flag in the url. When editing an answer from a summary, whether it is section or final, it should return back to the page it came from, assuming nothing has changed which would affect completeness.

### How to review 
Make sure return_to_summary returns to the section summary and return_to_final_summary returns to the final summary in the right situations.


app/views/contexts/section_summary_context.py
```diff
@@ -10,8 +10,8 @@
 
 
 class SectionSummaryContext(Context):
-    def __call__(self, current_location):
-        summary = self._build_summary(current_location)
+    def __call__(self, current_location, return_to="summary"):
```
LJBabbage, line 13: This is a pain and I am not particularly happy with it. As questionnaire summary (i.e final summary) produces multiple section summaries to display content, I have had to pass "return_to" down through the layers to eventually the block (summary) which creates the link.  I couldn't find an alternative way.

app/views/contexts/calculated_summary_context.py
```diff
@@ -33,6 +33,7 @@ def build_groups_for_section(self, section):
                 self._schema,
                 location,
                 self._language,
+                return_to="final_summary",
```
ajmaddaford, line 36: Given that we use `return_to` here with a string value, we should consider changing the url parameters to be the same. So instead of `return_to_summary` and `return_to_final_summary`, we would have `return_to=summary` or `return_to=final_summary` .

app/views/contexts/section_summary_context.py
ajmaddaford: I don't think there is another way without further refactoring.

app/views/contexts/calculated_summary_context.py
LJBabbage: lol, that was what I originally thought, but was convinced otherwise

app/views/handlers/list_action.py
```diff
@@ -33,7 +33,7 @@ def is_location_valid(self):
         return True
 
     def get_previous_location_url(self):
-        if self._return_to_summary:
+        if self._return_to == "summary":
```
LJBabbage, line 36: I was expecting to have to deal return_to="final_summary here" however whether it is by design or a bug, you can't have list items appear on the final summary. https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/views/contexts/questionnaire_summary_context.py#L33

MebinAbraham: By design for Census.

LJBabbage: Thanks Mebin

app/views/contexts/questionnaire_summary_context.py
```diff
@@ -34,5 +34,7 @@ def _build_all_groups(self):
                 break
 
             location = Location(section_id=section_id)
-            for group in section_summary_context(location)["summary"]["groups"]:
+            for group in section_summary_context(location, "final_summary")["summary"][
+                "groups"
+            ]:
```
MebinAbraham, line 39: Could we make the `"final_summary"` a keyword argument here, just for clarity and readability?. Threw me at first, then realised it was the `return_to` value.

app/views/contexts/section_summary_context.py
MebinAbraham, line 13: Thought on `return_to` being an attribute of `SectionSummaryContext` to remove some unneeded passing around of the arg and to avoid hardcoding to `summary` in multiple places?

app/questionnaire/router.py
```diff
@@ -75,26 +75,26 @@ def get_next_location_url(self, location, routing_path, return_to_summary=False)
         is_section_complete = self._progress_store.is_section_complete(
             location.section_id, location.list_item_id
         )
+        if return_to == "summary" and is_section_complete:
+            return self._get_section_url(location)
+
+        if return_to == "final_summary" and is_section_complete:
```
MebinAbraham, line 81: It is best practice to use hyphens in URLs over underscores.

LJBabbage: I used to be a web developer for 7 years, fancy doing that, I wonder what else I have forgotten

app/views/contexts/section_summary_context.py
LJBabbage: It considered it, but thought it might look a bit odd with the an __init__ with just return_to in, as I wouldn't add it to Context. Might do it and see how it looks

LJBabbage:
> My only other comment is about the term for summary and final summary. We use `summary` to mean `section summary` and `final_summary` for the questionnaire summary. Personally, I don't think it is the clearest. Thoughts on `section-summary` and `final-summary`/`questionnaire-summary`?. We still have the notion of "SectionSummary" in runner.

I agree, once in route.py what summary means can become lost. I think final summary is generally accepted as meaning the main summary at the end, but I think more clarity around section-summary wouldn't hurt.

app/views/contexts/section_summary_context.py
LJBabbage: Had a look, but no real benefit, SectionSummaryContext isn't passing into group, just some of its attributes,  which is called on _build_summary in SectionSummaryContext, so it would save one arg pass, but add an __init__

PR #212, opened 2020-07-22
Add change or resend individual response feature

ajmaddaford:
Adds the ability change or resend a requested individual response.

### How to review 
Review the journeys as documented in the Trello card.

### Checklist


app/views/handlers/individual_response.py
```diff
@@ -182,30 +205,199 @@ class IndividualResponseHowHandler(IndividualResponseHandler):
...
         return render_template(
-            "individual_response/how",
+            "individual_response/question",
             language=self._language,
             content=self.get_context(),
             previous_location_url=previous_location_url,
+            show_contact_us_guidance=True,
+        )
+
+    def handle_post(self):
```
JamesGardiner, line 247: Is this if statement required? I think `url_for` will skip a param pair if the value is `None`

tests/integration/individual_response/test_individual_response.py
```diff
@@ -402,3 +390,135 @@ def test_default_post_routing(self):
...
+class TestIndividualResponseChange(IndividualResponseTestCase):
+    def test_hub_change_link_goes_to_change_page(self):
+        # Given I request an individual response by post
+        self._request_individual_response()
+
+        # When I navigate to the hub and click on the change individual response link
+        self.get("/questionnaire")
+        self.get(self.individual_section_link)
+
+        # Then I should see change individual response page
```
JamesGardiner, line 404:
```suggestion
        # Then I should see the change individual response page
```

app/views/handlers/individual_response.py
```diff
@@ -182,30 +205,199 @@ class IndividualResponseHowHandler(IndividualResponseHandler):
...
             form_data,
             list_item_id,
         )
 
     def handle_get(self):
-        previous_location_url = url_for(
-            "individual_response.request_individual_response",
-            list_item_id=self._list_item_id,
-        )
+        if self._request_args.get("journey") == "change":
```
JamesGardiner, line 228: Given this request arg isn't used on _every_ journey through the IR scenarios, only on the change scenario, then it should probably be called something else. `action` might make more sense in this case, such as `action=change`

ajmaddaford: The intention is this param will be used for other journeys. For future stories we will need to know whether they came from the hub or from the individual section.

app/views/handlers/individual_response.py
ajmaddaford: I didn't know that `url_for` skips `None` values, good to know. I will update all occurrences of this.

app/views/handlers/individual_response.py
```diff
@@ -182,29 +205,190 @@ class IndividualResponseHowHandler(IndividualResponseHandler):
...
+        return redirect(
+            url_for(
+                ".get_individual_response_post_address_confirm",
+                list_item_id=self._list_item_id,
+                journey=self._request_args.get("journey"),
+            )
+        )
+
+
+class IndividualResponseChangeHandler(IndividualResponseHandler):
```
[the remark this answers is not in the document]
ajmaddaford: I did think about that, but I don't see why as it works even if the status isn't `INDIVIDUAL_RESPONSE_REQUESTED`.

app/views/handlers/individual_response.py
```diff
@@ -182,29 +205,190 @@ class IndividualResponseHowHandler(IndividualResponseHandler):
...
+    def _update_section_completeness(self):
+        routing_path = self.router.routing_path(
+            self.individual_section_id, self._list_item_id
+        )
+        status = (
+            CompletionStatus.COMPLETED
+            if self.router.is_path_complete(routing_path)
+            else CompletionStatus.IN_PROGRESS
+        )
+        self._update_section_status(status)
```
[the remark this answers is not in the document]
ajmaddaford: Good spot. I will look into fixing this.

app/views/handlers/individual_response.py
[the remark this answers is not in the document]
ajmaddaford: Arguably this could be considered as a "choose how you want to complete the questions for someone" screen.

app/views/handlers/individual_response.py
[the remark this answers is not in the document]
ajmaddaford: Fixed. The status is set to `NOT_STARTED` if there are no completed block ids for that section in the progress store. This also works out better as it means we don't need to generate a routing path in these cases 🙂 .

PR #214, opened 2020-07-23
Fix method to use local time in date handler

petechd:
This changes date_handler.py to use `now()` instead of `utcnow()` when getting a referenced date. Previously when time was 
between 12am and 1am the test was failing because of timezone difference.

### How to review 
New test was added. Run unit tests using `make test-unit` to check if the date is correct when test time is less than an hour to 12am in Universal Coordinated Time and local time is BST.


petechd:
[Tests fail between 12am and 1am due to use of utcnow in date_handler.py (s)](https://trello.com/c/pBmJYDJh/3999-tests-fail-between-12am-and-1am-due-to-use-of-utcnow-in-datehandlerpy-s)

app/forms/field_handlers/date_handler.py
```diff
@@ -71,7 +71,7 @@ def get_referenced_date(self, key):
         value = self.get_schema_value(self.answer_schema[key])
 
         if value == "now":
-            value = datetime.utcnow().strftime("%Y-%m-%d")
+            value = datetime.now().strftime("%Y-%m-%d")
```
LJBabbage, line 74: This confuses me. Ignoring the tests, if someone called this function at 00:01 (assume British Summer Time) it would have returned the day before as it was UTC? Have we been dealing with that somewhere else or has that been wrong?

petechd: Yes, datetime.utcnow() would return the day before compared to `datetime.now()` if your local time is 00:01 BST.

MebinAbraham: When doing comparison etc we should always be using `utcnow()`, therefore this should be changed back.

The test is failing between 12 and 1 failing because in the test we are comparing a `utc` timezone date with a local timezone date. So this needs to be changed here:
https://github.com/ONSdigital/eq-questionnaire-runner/blob/change-timezone-in-date-handler/tests/app/forms/field_handlers/test_date_handler.py#L107-L109

petechd: Ok, all sorted now @MebinAbraham.

PR #217, opened 2020-07-27
Remove complete/incomplete, add warning and rationalise

LJBabbage:

This PR adds the warning message to the hub. As guidance has been replaced it can be removed and the schema rationalised 

The validator branch which goes with this is add-warning-to-hub-and-rationalise

### How to review 
With test_hub_and_spoke_custom_content make sure the right text appears at the right time

### Checklist


tests/functional/base_pages/hub.page.js
```diff
@@ -33,12 +33,12 @@ class HubPage extends BasePage {
...
-  displayedGuidance() {
-    return '[data-qa="hub-guidance-panel"]';
+  displayedWarning() {
+    return '[data-qa="warning"]';
   }
 
-  displayedSubmissionGuidance() {
-    return '[data-qa="submission-guidance"]';
+  displayedGuidance() {
+    return '[data-qa="guidance"]';
```
ajmaddaford, line 41: It's minor, and not introduced in this pr, but I can't see why we use "displayed". These should just be `warning` and `guidance`.

LJBabbage: I thought the same, but went with it as I thought there might be a reason. However if neither us knows, best to change it

app/views/contexts/hub_context.py
```diff
@@ -42,47 +42,35 @@ class HubContext(Context):
     }
 
     def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
-        hub_schema = self._schema.get_hub()
+        hub_submission_schema = self._schema.get_hub().get("submission", {})
```
MebinAbraham, line 45: We should move this to after `if survey_complete:` given this is only needed when survey is complete.

app/views/contexts/hub_context.py
```diff
@@ -42,47 +42,35 @@ class HubContext(Context):
...
-
-            guidance = custom_text.get("guidance") or lazy_gettext(
-                "You must complete all sections in order to submit this survey"
-            )
-
             submit_button = lazy_gettext("Continue")
-            submission_guidance = None
-
+            guidance = None
+            warning = None
```
MebinAbraham, line 62:
```suggestion
            guidance, warning = None, None
```

LJBabbage:
> Mebin's comments on the translations pull request led me to copying the `test_language` schema into runner. Running this schema has highlighted a potential issue - the final summary doesn't use the custom strings, which it possibly should.
> 
> This will need to be looked at as it may affect the census individual schema.

Probably best to split this issue out into its own card, considering this is changes to the hub text which doesn't have any impact on final summary.

ajmaddaford:
> > Mebin's comments on the translations pull request led me to copying the `test_language` schema into runner. Running this schema has highlighted a potential issue - the final summary doesn't use the custom strings, which it possibly should.
> > This will need to be looked at as it may affect the census individual schema.
> 
> Probably best to split this issue out into its own card, considering this is changes to the hub text which doesn't have any impact on final summary.

We could do, but that would mean changing the schema for hub again as I think the definition needs to move out from the `hub` property i.e. the same schema properties should be used for both places.

app/views/contexts/hub_context.py
LJBabbage: My eyes bleed

LJBabbage:
No worries, will do, I didn't see the connection, but if there is one I will get it fixed here

LJBabbage:
I have now mirrored the translation repo languages. There seems to be a be an odd gap in the translations repo which we probably need to have a look at, questionnaire description is one and there was another, but I can't recall at the moment. Also after running the script in translations escapes character were used which I am not 100% sure we want

test_schemas/cy/test_language_hub.json
```diff
@@ -0,0 +1,67 @@
+{
```
LJBabbage, line 1: I fixed and updated all of test_languages for all languages only to find validator wouldn't allow me to have a final summary and a hub, so I had to add a separate schema, doh. It seems daft not to include the changes to test_languages even thought they weren't in the end needed

tests/functional/base_pages/hub.page.js
```diff
@@ -29,16 +29,16 @@ class HubPage extends BasePage {
...
 
-  displayedGuidance() {
-    return '[data-qa="hub-guidance-panel"]';
+  Warning() {
+    return '[data-qa="warning"]';
   }
 
-  displayedSubmissionGuidance() {
-    return '[data-qa="submission-guidance"]';
+  Guidance() {
```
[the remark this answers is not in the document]
LJBabbage: Lazy deleting on my part

test_schemas/cy/test_language_hub.json
ajmaddaford: Having considered this a bit more, I don't think we should have two test language schemas. I think it's enough to have one schema that uses a hub, rather than a final summary. The custom submission text translations will be verified by the hub schema. When the final summary is updated to use the custom submission text, we can test the English version, and should have confidence that the submission elements will be translated (as they are tested as part of the hub schema).

app/utilities/schema.py
```diff
@@ -18,6 +18,7 @@
 
 LANGUAGES_MAP = {
     "test_language": [["en", "cy"], ["en", "ga"]],
+    "test_language_hub": [["en", "cy"]],
```
ajmaddaford, line 21: This can go now.

LJBabbage:
> I tested the schemas by changing the tag in the run validator script to `add-warning-to-hub-and-rationalise`. This should really be done as part of the pull request (and reverted when the validator changes merge).

LJBabbage:
> I tested the schemas by changing the tag in the run validator script to `add-warning-to-hub-and-rationalise`. This should really be done as part of the pull request (and reverted when the validator changes merge).

I think both ways have their merits, doing it this ways blocks the merging of this PR till validator is correct, which I think is the right order, it also limits the amount of changes after reviews. Changing the branch in the scripts naturally ties it together a little better, but could lead to accidental merging and extra commits.

ajmaddaford:
> I think both ways have their merits, doing it this ways blocks the merging of this PR till validator is correct and merged first, which I think is the right order, it also limits the amount of changes after reviews. Changing the branch in the scripts naturally ties it together a little better, but could lead to accidental merging and an extra unreviewed commit.

Possibly, but I think the expectation is a pull request should be green before people spend time reviewing it. It's also not great that a pull request is being approved while there is a failing test.

PR #219, opened 2020-07-28
Use primary block for generating primary edit links

iwootten:

The links generated on the summary page are incorrect for primary person, meaning that it will break if only a primary person is entered and then the change link is used on the summary to update their details. We've confirmed it is not necessary for the question to be different when editing the primary person.

### How to review 

Add only a primary person to household, proceed to summary, try to edit that person. See if you agree with the changes.


iwootten:
[WLH breaking on return to change details](https://trello.com/c/i295qHIo/4017-wlh-breaking-on-return-to-change-details)

app/views/contexts/section_summary_context.py
```diff
@@ -95,6 +95,15 @@ def _list_summary_element(self, summary, current_location, section) -> Mapping:
...
         )[0]
 
+        primary_person_blocks = self._schema.get_list_collectors_for_list(
+            section, for_list=summary["for_list"], primary=True
+        )
+        primary_person_edit_block_id = (
+            primary_person_blocks[0]["add_or_edit_block"]["id"]
+            if len(primary_person_blocks) > 0
+            else None
+        )
```
MebinAbraham, line 105: We could save fetching the primary person list collector block by checking if the list store for the `for_list` contains a primary person. I think we should also use a generator. 

To add to that, if we check the list count, then we can use the normal list collector when we know it is available, i.e. when count is above 1, therefore a better UX given as it changes the text for the user.
```py
        primary_person_edit_block_id = None
        current_list = self._list_store[summary["for_list"]]
        if len(current_list) == 1 and current_list.primary_person:
            primary_person_edit_block_id = self._schema.get_list_collectors_for_list(
                section, for_list=summary["for_list"], primary=True
            )[0]["add_or_edit_block"]["id"]
```

ajmaddaford:
The change link for primary person on a list collector goes to the list collector edit screen. This may be acceptable, but we should check that it is. If so, we should implement Mebin's suggestion so that the change link from the section summary behaves in the same way.

iwootten:
> This may be acceptable, but we should check that it is. 

Checked with Laura and she's confirmed that this is ok.

PR #221, opened 2020-08-04
Spike using ListCollector actions

iwootten:

Removes the use of add_answer and remove_answer from list collectors and migrates them to instead use actions. Adds a "RemoveAnswersForListItem" action to support new answer removal action.

Migrating schemas is fairly simple - just adding the appropriate action to the block/answer option in question. Test schemas have been migrated as part of this PR, census schemas are likely to be easier due to duplication simplified by use of jsonnet. I had started working on a script to do this, but it was simpler in the end to migrate manually.

Most of the work here is likely in updating validator to migrate to these new actions and updating the logic in the list collector validator. Update - Added a quick validator branch https://github.com/ONSdigital/eq-questionnaire-validator/pull/53

### How to review 

Check approach, see if you agree with implementation.

### DIFF
--- a/app/views/handlers/list_collector.py
+++ b/app/views/handlers/list_collector.py
@@ -7,6 +7,7 @@ from app.views.contexts import ListContext
 class ListCollector(Question):
     def __init__(self, *args):
         self._is_adding = False
+
         super().__init__(*args)
 
     def get_next_location_url(self):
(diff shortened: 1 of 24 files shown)


iwootten:
[Spike: Using actions for list_collectors (1 day)](https://trello.com/c/qU73InX1/4037-spike-using-actions-for-listcollectors-1-day)

PR #224, opened 2020-08-05
casefold equals comparison values

JamesGardiner:
Changes runner equals comparison operators to be case insensitive.

### How to review 
Check that schema equals operations are case insensitive. Can be done on the NISRA census household schema by entering lowercased county strings in the workplace-country-answer to trigger the workplace-outside-northern-ireland routing rule

### Checklist


app/questionnaire/rules.py
```diff
@@ -62,6 +62,15 @@ def evaluate_condition(condition, answer_value, match_value):
     """
     answer_and_match = answer_value is not None and match_value is not None
 
+    if condition in {"equals", "not equals", "equals any", "not equals any"}:
```
LJBabbage, line 65: Do we need to cater for all the 'contains'? I guess we have more control over them being pre-determined lists.

.gitignore
```diff
@@ -110,4 +110,7 @@ templates/layout
 tags
 
 # Ignore wdio/chromedriver logs
-tests/functional/logs/
\ No newline at end of file
+tests/functional/logs/
+
+# `typing` folder
```
LJBabbage, line 115: Not sure this comment adds much, I still don't really know what would be in typing to ignore

app/questionnaire/rules.py
```diff
@@ -93,6 +102,13 @@ def evaluate_condition(condition, answer_value, match_value):
     return match_function(answer_value, match_value)
 
 
+def casefold(value):
```
LJBabbage, line 105: Is this because of the other languages?

app/questionnaire/rules.py
```diff
@@ -62,6 +62,15 @@ def evaluate_condition(condition, answer_value, match_value):
     """
     answer_and_match = answer_value is not None and match_value is not None
 
+    if condition in {"equals", "not equals", "equals any", "not equals any"}:
+
+        answer_value = casefold(answer_value)
+
+        if isinstance(match_value, list):
+            match_value = list(map(casefold, match_value))
```
LJBabbage, line 70: match_value = list(map(casefold, match_value)) if list(map(casefold, match_value)) else casefold(match_value) 

maybe?

tests/app/questionnaire/test_rules.py
```diff
@@ -97,6 +97,37 @@ def test_evaluate_rule_equals_with_number(self):
         self.assertFalse(evaluate_rule(when, 2))
         self.assertTrue(evaluate_rule(when, 0))
 
+    def test_evaluate_rule_equals_with_string(self):
```
LJBabbage, line 100: Think I would like to see something referencing the case difference in the test name or a comment

tests/app/questionnaire/test_rules.py
```diff
@@ -97,6 +97,37 @@ def test_evaluate_rule_equals_with_number(self):
...
+        when = {"value": "answervalue", "condition": "equals"}
+
+        self.assertTrue(evaluate_rule(when, "answerValue"))
+        self.assertTrue(evaluate_rule(when, "answervalue"))
+        self.assertFalse(evaluate_rule(when, "answer-value"))
```
LJBabbage, line 105: I find these hyphen comparison a bit odd, any reason for doing it?

app/questionnaire/rules.py
JamesGardiner: Yeah was decided not to look at the contains, as we define the lists

app/questionnaire/rules.py
JamesGardiner: Yes - casefold takes into account non ASCII char codes

tests/app/questionnaire/test_rules.py
JamesGardiner: I added them here to test the negation 👍

app/questionnaire/rules.py
JamesGardiner: I think this would fail for not iterable data

tests/app/questionnaire/test_rules.py
LJBabbage: I understood why. Belt and braces programming

JamesGardiner: I think copy and paste programming is more apt 😉

app/questionnaire/rules.py
```diff
@@ -62,6 +62,15 @@ def evaluate_condition(condition, answer_value, match_value):
     """
     answer_and_match = answer_value is not None and match_value is not None
 
+    if condition in {"equals", "not equals", "equals any", "not equals any"}:
+
+        answer_value = casefold(answer_value)
```
MebinAbraham, line 67: Given we have a lot of routing for census using non str answers like dates, this func would be excepting AttributeError a lot, therefore, would we rather not do:
```py
answer_value = (
            casefold(answer_value) if isinstance(answer_value, str) else answer_value
        )
```
or make this check part of `casefold()`. Same point for `match_value`. Thoughts?

JamesGardiner: Unless `isinstance` can be shown to be qualitatively better in this scenario then I'll keep the duck typing approach and not hard code to the `str` builtin

MebinAbraham: `isinstance()` would be more performant if most cases weren't strs but given it should be negligible I'm happy to leave as is.

test_schemas/en/test_routing_case_insensitive_text_field.json
```diff
@@ -0,0 +1,112 @@
...
+                            "type": "Question",
+                            "id": "country-text-field",
+                            "question": {
+                                "id": "country-text-field-question",
+                                "title": "Name a country you have visited",
+                                "type": "General",
+                                "answers": [
+                                    {
+                                        "id": "country-text-field-answer",
+                                        "label": "country-text-field-answer",
```
MebinAbraham, line 40: Why is the label like an id?

test_schemas/en/test_routing_case_insensitive_text_field.json
```diff
@@ -0,0 +1,112 @@
...
+            "groups": [
+                {
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "country-text-field",
+                            "question": {
+                                "id": "country-text-field-question",
+                                "title": "Name a country you have visited",
+                                "type": "General",
```
MebinAbraham, line 36: In some schemas we have guidance to say what to type to test routing, ie with dates. I think we should do the same as I wouldn't know what to type to test this without looking at the source.

MebinAbraham: Like these: https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/test_schemas/en/test_routing_date_equals.json#L45-L60

PR #228, opened 2020-08-06
Add languages to format_ordinal

berroar:
Updated the `format_ordinal` function to handle Gaelic, Ulster Scots and Welsh ordinals.

### How to review 
Check that the oridnals are correct as per the spreadsheet attached to the trello card.

### Checklist


berroar:
[[B] Update placeholder transforms to support ordinals in other languages (M)](https://trello.com/c/YtM4mIWo/3806-b-update-placeholder-transforms-to-support-ordinals-in-other-languages-m)

app/questionnaire/placeholder_transforms.py
```diff
@@ -129,21 +129,51 @@ def parse_date(date):
...
+                    12: "fed",
+                    13: "eg",
+                    14: "eg",
+                    15: "fed",
+                    16: "eg",
+                    17: "eg",
+                    18: "fed",
+                    19: "eg",
+                    20: "fed",
+                }.get(number_to_format, "fed")
```
MebinAbraham, line 176: Can we separate getting the suffix into a different method to make this a bit clearer?

app/questionnaire/placeholder_transforms.py
```diff
@@ -147,6 +143,42 @@ def format_ordinal(number_to_format, determiner=None):
 
         return f"{number_to_format}{suffix}"
 
+    @staticmethod
+    def get_ordinal(language_code, number_to_format):
```
MebinAbraham, line 147: Shouldn't be static, as it means you don't need to pass language code. Also, method name should be something like `get_suffix_for_ordinal`.

berroar: I renamed it to use `indicator` as `suffix` is not the correct term

app/questionnaire/placeholder_transforms.py
```diff
@@ -129,23 +129,54 @@ def parse_date(date):
...
+                12: "fed",
+                13: "eg",
+                14: "eg",
+                15: "fed",
+                16: "eg",
+                17: "eg",
+                18: "fed",
+                19: "eg",
+                20: "fed",
+            }.get(number_to_format, "fed")
```
MebinAbraham, line 179: Actually, these don't match the google sheets?

berroar: Ah, think I'd just accidentally missed 11 somehow

berroar: The rest (22+) are not on the spreadsheet, but they follow welsh ordinal rules

app/questionnaire/placeholder_transforms.py
```diff
@@ -129,23 +129,53 @@ def parse_date(date):
...
+                11: "eg",
+                12: "fed",
+                13: "eg",
+                14: "eg",
+                15: "fed",
+                16: "eg",
+                17: "eg",
+                18: "fed",
+                19: "eg",
+            }.get(number_to_format, "fed")
```
MebinAbraham, line 178: Could be shortned to:
```suggestion
            return {
                1: "af",
                2: "il",
                3: "ydd",
                4: "ydd",
                5: "ed",
                6: "ed",
                11: "eg",
                13: "eg",
                14: "eg",
                16: "eg",
                17: "eg",
                19: "eg",
            }.get(number_to_format, "fed")
```

PR #232, opened 2020-08-07
Add uwsgi run option

ajmaddaford:
Adds the option to run under uWSGI. There are three options within this - plain uWSGI, threaded or asynchronous (via gevent).

This branch has been used to run the tests in https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/55.

### How to review 
- Run the app locally with the Makefile commands `run-uwsgi`, `run-uwsgi-threads` or `run-uwsgi-async`
- Check that `run-gunicorn` works as it did previously
- Run under Docker


run_app.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -euo pipefail
```
MebinAbraham, line 3: Will need to set `EQ_NEW_RELIC_ENABLED=False` in `.development.env`

iwootten:
I can run by setting the EQ_NEW_RELIC_ENABLED=False as mentioned. Runner under docker (also on master) also seems to fail `Exception: Setting 'EQ_SUBMITTED_RESPONSES_TABLE_NAME' Missing`.

ajmaddaford:
@iwootten That error looks like docker isn't being passed the required environment variables from the `.development.env` file. I ran using the `docker-compose.yml` file and it worked.

uwsgi.ini
```diff
@@ -0,0 +1,12 @@
+[uwsgi]
+http = :5000
+module = application:application
+buffer-size = 32768
```
MebinAbraham, line 4: Was 32k a guesstimate or do we know our max request size?

ajmaddaford: Our payloads vary between 6-10k, so it was a guesstimate based on that with headroom.

PR #235, opened 2020-08-13
Spike Address Lookup Question Type

MebinAbraham:
The initial approach to integrating the address lookup.

#### Options considered:
**1. New answer type:**
https://github.com/ONSdigital/eq-questionnaire-runner/pull/241

**2. New question type (spike)**
This option was ideal to make address lookup work which met the requirements while making minimal changes.
As implemented, the new question type allows us to have the 4 separate address answer fields which means we can store the answers separately while reusing existing functionality to concatenate these together for the user on the summary using question summary.

**Caveats:**
Due to the way the address component is built, there are a few rules we need to adhere to:
- The answer IDs must be consistent and must have the same parent id. To do this, I used the format:
	- `{question_id}-{field_id}` for example `address-line1`, `address-line2`, `address-town` etc.
- Only address line 1 can be mandatory, the front-end component doesn't support address other mandatory fields, so questionnaire validator will need to ensure that address line 1 is and is the only mandatory answer.
- Have not added support for answer guidance to each answer field. Not required for the current Address lookup.

**Known issues:**
- If the API is down, returning to the address lookup question from the summary displays the lookup when it shouldn't. This will be fixed soon in the DS.
- Error panel shown on the manual fields will persist and be displayed on the lookup page which results in a double error panel. This will also be fixed soon in the DS.

- Ensure the address lookup works as expected.
- Ensure answers are in the payload as expected.
- Do you agree with the approach?
- Was anything missed?
- Compare the two spike approaches.

### Checklist


MebinAbraham:
Going to spike an answer type approach with a new custom answer store object.

ajmaddaford:
There are some downsides to the new question type:
- We are adding logic to the jinja filters that we've said we need to move away from
- There are a number of schema properties that need to be defined in a specific way for this to work, and align to the fixed values in the jinja filter. This would also mean extra validator logic to ensure the schema for addresses is valid, which again we've said we'd like to avoid.

I think we should spike the answer type approach, then we can compare the two and choose the best option. 

In order for us to be able to store each address field independently, we could store the answer as an object in the answer store, something like:

```
{
  "line1": "dsvdsv",
  "line2": "dvdsv",
  "town": "dsvdsv",
  ...
}
```

MebinAbraham:
https://github.com/ONSdigital/eq-questionnaire-runner/pull/241 chosen as the preferred way to implementing addresses in runner.

MebinAbraham:
https://github.com/ONSdigital/eq-questionnaire-runner/pull/241 chosen as the preferred way to implementing addresses in runner.

PR #238, opened 2020-08-14
Request IR by Post from Hub

iwootten:

Added a person selector to the individual response. The selector is only shown when requested via the hub and 2 or more non-primary householders exist. The selector is skipped on paths from the hub where conditions aren't met.

### How to review 

Check to see if you agree with the approach.

### Checklist


app/views/handlers/individual_response.py
```diff
@@ -491,13 +512,109 @@ def handle_post(self):
...
+            request_args.get("list_item_id"),
+        )
+
+    @cached_property
+    def selected_option(self):
+        answer_id = self.rendered_block["question"]["answers"][0]["id"]
+        return self.form.get_data(answer_id)
+
+    @cached_property
+    def selected_list_item(self):
```
LJBabbage, line 566: Very minor, but I would absorb selected_option in here, I understood what select_list_item meant without the extra method and with it not being used anywhere else I feel it could sit as one.

app/views/handlers/individual_response.py
```diff
@@ -491,13 +512,109 @@ def handle_post(self):
...
+                name_answers.append(
+                    (
+                        list_item,
+                        questionnaire_store.answer_store.get_answers_by_answer_id(
+                            ["first-name", "last-name"], list_item_id=list_item
+                        ),
+                    )
+                )
+
+        self.people_names = {
```
LJBabbage, line 546: I think we lose a bit of context with the var name, when I got to the handle_get and looking at len(self.people_names) > 1 I couldn't tell if this include primary or not. On checking back I can see its has been  removed, the name should perhaps reflect that

tests/integration/individual_response/test_individual_response.py
```diff
@@ -310,6 +320,97 @@ def test_introduction_previous_goes_to_hub(self):
...
+        self._add_household_no_primary()
+
+        self.get(self.individual_response_link)
+        self.post()
+
+        # Then I should skip the member selector
+        self.assertInUrl("/how")
+
+    def test_goes_to_who_selector(self):
+        # Given I add a household member
```
LJBabbage, line 345: 2 household members / or just members

tests/integration/individual_response/test_individual_response.py
```diff
@@ -310,6 +320,97 @@ def test_introduction_previous_goes_to_hub(self):
...
+        self._add_household_multiple_members_no_primary()
+
+        self.get(self.individual_response_link)
+        self.post()
+
+        # Then I should be taken to the member selector
+        self.assertInUrl("/who")
+
+    def test_previous_returns_to_hub(self):
+        # Given I add a household member
```
LJBabbage, line 356: Same as above

app/views/handlers/individual_response.py
```diff
@@ -491,13 +512,109 @@ def handle_post(self):
...
                 "individual_response.get_individual_response_how",
                 list_item_id=self._list_item_id,
+                return_to=self._return_to,
+            )
+        )
+
+
+class IndividualResponseWhoHandler(IndividualResponseHandler):
+    def __init__(self, schema, questionnaire_store, language, request_args, form_data):
+        self._list_name = schema.json.get("individual_response", {}).get("for_list")
```
LJBabbage, line 531: As we are calling schema.json.get("individual_response", {}) a few times now, would it be best to have its own method in QuestionnaireSchema?

app/routes/individual_response.py
```diff
@@ -63,10 +65,10 @@ def request_individual_response(schema, questionnaire_store):
         list_item_id=list_item_id,
     )
 
-    if request.method == "GET":
-        return individual_response_handler.handle_get()
+    if request.method == "POST":
```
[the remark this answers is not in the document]
iwootten: The default fallback option should really always be the call to the get. The only other place I can see this is 3 cases in questionnaire.py, where the logic on the get condition is confusing because it involves a call to the post form. They should probably be changed too.

app/views/handlers/individual_response.py
```diff
@@ -491,13 +512,109 @@ def handle_post(self):
...
+        list_model = questionnaire_store.list_store[self._list_name]
```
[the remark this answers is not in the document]
iwootten: It's a model in the list store, not the store itself.

app/views/handlers/individual_response.py
iwootten: This is the approach adopted in the previous class, I think it's clearer to separate them.

app/views/contexts/hub_context.py
```diff
@@ -202,14 +179,16 @@ def _individual_response_enabled(self) -> bool:
         if not self._schema.json.get("individual_response"):
             return False
 
-        count_household_members = len(self._list_store[self._for_list])
+        for_list = (
+            self._schema.json["individual_response"]["for_list"]
+            if self._schema.json.get("individual_response")
```
LJBabbage, line 184: At this stage you know individual_response is true, due to the if above it

app/routes/individual_response.py
```diff
@@ -152,10 +155,31 @@ def get_individual_response_post_address_confirmation(schema, questionnaire_stor
...
+        return redirect(url_for("questionnaire.get_questionnaire"))
+
+    return render_template(
+        template="individual_response/confirmation",
+        display_address=questionnaire_store.metadata.get("display_address"),
+    )
+
+
+@individual_response_blueprint.route("/who", methods=["GET", "POST"])
```
MebinAbraham, line 167: The previous link for this journey should back to the previous page. It should only go to the hub when on you are back on the `IR request launcher` page.

Another two scenarios:
- **1 non-primary person only:**
	- The previous link on the how page is also adding list item id to the url which is taking it back to the individual section from `IR request launcher` page instead of the hub. Regardless of the number of people, the previous should go back to the hub if we know they came from the hub?.

- **at least 2 non-primary person**
	- The previous link from any IR page is going straight to the hub, instead of back sequentially to the `IR request launcher` page then the hub.

Hopefully that is clear, if not, happy to discuss further.

iwootten: I think the card has overstated this somewhat which has made it confusing "Previous link should always go back to hub no matter how far along the IR journey you are (ensure alternate journey goes to individual section introduction page)."

JamesGardiner: I agree that comment has overstated it @iwootten - I think it was added to differentiate between the hub originated journey and the individual section originated journey, but that would be default behaviour as you pointed out 😄

app/views/handlers/individual_response.py
```diff
@@ -225,16 +235,38 @@ def __init__(
...
-        if self._request_args.get("journey") == "change":
+        if self._return_to_hub:
             previous_location_url = url_for(
-                "individual_response.get_individual_response_change",
-                list_item_id=self._list_item_id,
+                "individual_response.get_individual_response_who",
+                return_to=self._return_to,
             )
-        else:
+        elif self._request_args.get("journey") == "change":
```
MebinAbraham, line 243: I think we should use `journey` instead of `return_to` when the immediate page isn't the return to location. return to makes sense if the next page or previous page would be the return to page, but `journey` was recently introduced to track the journey between multiple screens.

app/views/handlers/individual_response.py
```diff
@@ -225,16 +235,38 @@ def __init__(
...
+        else:
+            self._list_name = self._schema.get_individual_response_list()
+            list_model = self._questionnaire_store.list_store[self._list_name]
+            non_primary_members = 0
+
+            for list_item in list_model.items:
+                if list_item != list_model.primary_person:
+                    non_primary_members += 1
+
+            if non_primary_members > 1:
```
MebinAbraham, line 257: This isn't needed if we persist the request args, because that would tell you if you came from the hub.

app/views/handlers/individual_response.py
```diff
@@ -137,22 +140,29 @@ def handle_get(self):
...
             language=self._language,
             previous_location_url=previous_location_url,
         )
 
+    def handle_post(self):
+        if self._list_item_id:
+            return redirect(
+                url_for(".get_individual_response_how", list_item_id=self._list_item_id)
+            )
+        return redirect(url_for(".get_individual_response_who", return_to="hub"))
```
MebinAbraham, line 164: At the moment, when there is only 1 non-primary person, this is redirecting to the who route which then redirects to how route, but we should be avoiding that extra redirect as we can work out at this point whether to go to the who route or the how route based on the count of non-primary list items.

templates/hub.html
```diff
@@ -34,6 +35,10 @@ <h1 class="u-mt-m">{{ content.title }}</h1>
         <p class="u-mb-no" data-qa="guidance">{{ content.guidance }}</p>
       </div>
     {% endif %}
+    {% if content.individual_response_url %}
+    {% set title = _("If you can’t answer someone else’s questions") %}
+    {% include 'partials/individual-response-guidance.html' %}
```
MebinAbraham, line 40:
```suggestion
      {% set title = _("If you can’t answer someone else’s questions") %}
      {% include 'partials/individual-response-guidance.html' %}
```

app/views/handlers/individual_response.py
iwootten: This is able to be edited though?

MebinAbraham: Do you mean the query param can be edited?. If so, yes but that shouldn't matter as long as it is one of the ones we are expecting, as we need to handle that scenario anyway.

iwootten: As discussed offline, it's likely this logic doesn't belong in this handler if we skip over it when there is only a single member.

MebinAbraham (commented):
The previous links are not working correctly. If you go to the request IR page, the previous goes to a 404 page. If you step forward to the how screen, then clicking previous twice goes back to the individual section even though for both journeys I started from the hub.

app/views/handlers/individual_response.py
```diff
@@ -137,22 +139,44 @@ def handle_get(self):
...
+
+        self._list_name = self._schema.get_individual_response_list()
+        list_model = self._questionnaire_store.list_store[self._list_name]
+        non_primary_members = []
+
+        for list_item in list_model.items:
+            if list_item != list_model.primary_person:
+                non_primary_members.append(list_item)
+
+        if len(non_primary_members) == 1:
```
MebinAbraham, line 172: I think we should have a property on list model to return the non primary list item ids. Which can then be reused here, as well as in:
- https://github.com/ONSdigital/eq-questionnaire-runner/pull/238/files#diff-1b7aa7bc01cf9df6f320a4fd4a937b8aR544-R545
- https://github.com/ONSdigital/eq-questionnaire-runner/pull/238/files#diff-c8d1a13e851293923c3a04e3f7d0d1c0R186-R188

app/data_model/list_store.py
```diff
@@ -39,6 +40,14 @@ def __getitem__(self, list_item_index):
     def __len__(self):
         return len(self.items)
 
+    @cached_property
+    def non_primary_people(self):
+        non_primary = []
+        for item in self.items:
+            if item != self.primary_person:
+                non_primary.append(item)
+        return non_primary
```
MebinAbraham, line 49: Could be simplified using a list comprehension.
```suggestion
        return [item for item in self.items if item != self.primary_person]
```

app/views/contexts/hub_context.py
```diff
@@ -175,18 +174,26 @@ def _get_rows(self, enabled_section_ids) -> List[Mapping[str, Union[str, List]]]
...
 
         count_household_members = len(self._list_store[for_list])
 
-        if count_household_members == 0:
-            return False
-
-        if count_household_members == 1 and self._list_store[for_list].primary_person:
+        if count_household_members == 0 or (
+            count_household_members == 1 and self._list_store[for_list].primary_person
+        ):
```
MebinAbraham, line 188: This can now be simplied to:
```py
if not self._list_store[for_list].non_primary_people:
	return False
```

app/views/handlers/individual_response.py
```diff
@@ -137,22 +139,46 @@ def handle_get(self):
...
+                url_for(".get_individual_response_how", list_item_id=self._list_item_id)
+            )
+
+        self._list_name = self._schema.get_individual_response_list()
+        list_model = self._questionnaire_store.list_store[self._list_name]
+        non_primary_members = []
+
+        for list_item in list_model.items:
+            if list_item != list_model.primary_person:
+                non_primary_members.append(list_item)
```
MebinAbraham, line 170: This should be replaced by `list_model.non_primary_people`.

app/views/handlers/individual_response.py
```diff
@@ -137,22 +139,41 @@ def handle_get(self):
...
-        if self._list_item_id:
+        if self._is_hub_journey:
+            previous_location_url = url_for("questionnaire.get_questionnaire")
+        else:
             previous_location_url = url_for(
                 "questionnaire.block",
                 list_name=self._list_name,
                 list_item_id=self._list_item_id,
                 block_id=individual_section_first_block_id,
             )
```
MebinAbraham, line 150: This logic means if you manually list the url, which is possible, you would get a 404 because there is no query param. Therefore, we should by default go to the help if there is no query param.

MebinAbraham (commented):
Apart from my last comment, everything seems to be as expected. 👍

MebinAbraham (approved):

PR #241, opened 2020-08-17
Spike Address Lookup Answer Type

MebinAbraham:
The initial approach to integrating the address lookup.

#### Options considered:
**1. New answer type (spike)**
New address answer type with new answer object structure to support the required fields.
```json
{
  "answer_id": "address",
  "value": {
    "line1": "7 Evelyn Street",
    "town": "Barry",
    "postcode": "CF63 4EN"
  }
}
```

**2. New question type**
https://github.com/ONSdigital/eq-questionnaire-runner/pull/235

**Caveats:**
Due to the way the address component is built, there are a few rules we need to adhere to:
- With this approach, there is no support to allow any other overriding labels or setting max length etc.
- Mandatory true only applies to address line 1.
- Have not added support for answer guidance. Not required for the current Address lookup.

**Known issues:**
- If the API is down, returning to the address lookup question from the summary displays the lookup when it shouldn't. This will be fixed soon in the DS.
- Error panel shown on the manual fields will persist and be displayed on the lookup page which results in a double error panel. This will also be fixed soon in the DS.

- Ensure the address lookup works as expected.
- Ensure answers are in the payload as expected.
- Do you agree with the approach?
- Was anything missed?
- Compare the two spike approaches.

### Checklist


ajmaddaford:
My preference is to use an address answer type over a question type.

- The logic is in runner, and doesn't need extra schema validation to ensure they lookup is configured as it should be
- The schema is simple
- The WTForms work is all done prior to rendering i.e. not using logic in jinja filters
- It aligns well with the fixed design of the design system address component

We will need to think about how we reference answers within the composite address answer e.g. for placeholders.

pricem14pc (approved):
Functionally tested this POC implementation against the WL AIMS. Front end behaving as expected and the payload as per spike proposal.

MebinAbraham:
This approach will be progressed seperately.

PR #245, opened 2020-08-18
Add IR link to remove person page

berroar:
Adds the link to request an IR form the remove person page.

Also removes validation checking if the hub is enabled for IR journeys.

### How to review 
Use the individual response schema and remove a household member to test the change.

### Checklist


berroar:
I have duplicated the `_individual_response_url` method from the `content.py` and added it to the `listRemoveQuestion` hander, so this probably needs to be refactored so that it lives somewhere else - just not sure where is best to put it

berroar:
[[H] add IR link to remove person page (M)](https://trello.com/c/gZJ66zIC/4065-h-add-ir-link-to-remove-person-page-m)

app/views/handlers/list_remove_question.py
```diff
@@ -1,3 +1,6 @@
+from typing import Union
+from flask import url_for
```
JamesGardiner, line 2:
```suggestion

from flask import url_for
```

app/views/handlers/__init__.py
```diff
@@ -8,4 +13,17 @@
...
+
+def individual_response_url(schema: Mapping, list_item_id: str, questionnaire_store: QuestionnaireStore) -> Union[str, None]:
+    if "individual_response" in schema.json:
+        for_list = schema.json["individual_response"]["for_list"]
+
+        if list_item_id != questionnaire_store.list_store._lists[for_list].primary_person:
+            return url_for(
+                "individual_response.request_individual_response",
+                list_item_id=list_item_id,
+            )
```
MebinAbraham, line 28: This would need to pass a query param to handle where the previous link goes to. Currently clicking previous goes to the introduction page but it should return to the remove screen.

app/views/handlers/list_remove_question.py
```diff
@@ -1,8 +1,12 @@
+<<<<<<< HEAD
+=======
 from typing import Union
 
 from flask import url_for
 
+>>>>>>> 46f0e4d75e5b09e34846da4d19e07189466f4757
```
JamesGardiner:
```suggestion
```

app/views/handlers/list_remove_question.py
```diff
@@ -1,8 +1,12 @@
+<<<<<<< HEAD
+=======
```
JamesGardiner:
```suggestion
```

JamesGardiner (changes requested):
As this now interacts with an implemented feature, it should be added to the list_collector.spec.js feature tests 👍

app/views/handlers/__init__.py
```diff
@@ -8,4 +14,23 @@
...
+
+
+def individual_response_url(
+    schema: QuestionnaireSchema,
+    list_item_id: str,
+    questionnaire_store: QuestionnaireStore,
+    journey: str = None,
+) -> Union[str, None]:
+    if "individual_response" in schema.json:
+        for_list = schema.json["individual_response"]["for_list"]
```
iwootten, line 28: Given schema is only used for 'for_list', it would make sense to only pass its value. The recent merge to master includes a method to fetch this from schema now.

app/views/handlers/individual_response.py
```diff
@@ -68,8 +68,9 @@ def __init__(
             raise NotFound
 
     def _is_location_valid(self):
-        if not self.router.can_access_hub():
-            return False
+        self._list_name = self._schema.json.get("individual_response", {}).get(
+            "for_list"
+        )
```
MebinAbraham, line 73: There is a method to get the IR list in questionnaire schema, we should use that.

app/views/handlers/individual_response.py
```diff
@@ -139,7 +140,15 @@ def handle_get(self):
...
         )
 
-        if self._list_item_id:
+        if self._request_args.get("journey") == "remove-person":
+            previous_location_url = url_for(
+                "questionnaire.block",
+                list_name=self._list_name,
+                list_item_id=self._list_item_id,
+                block_id="remove-person",
+            )
```
MebinAbraham, line 149: The journey needs to persist, at the moment, the previous works correct for only the first screen. If you progress forward, then essentially clicking previous multiple times won't take you back to where you came from hence when you proceed forward in the journey, it also needs to persist the journey param.

templates/listremovequestion.html
```diff
@@ -1,4 +1,6 @@
 {% extends 'list-action.tmpl' %}
 
+{% set show_individual_response_guidance = True %}
```
MebinAbraham, line 3: This should only be set if IR is enabled.

JamesGardiner (approved):

templates/partials/question.html
```diff
@@ -20,6 +20,13 @@
   {%- endif -%}
 {%- endset -%}
 
+{% set individual_response_guidance %}
+   {%- if show_individual_response_guidance == True -%}
+      {% set title = _("If you can’t answer questions for this person") %}
+      {% include 'partials/individual-response-guidance.html' %}
+   {%- endif -%}
+{% endset %}
```
MebinAbraham, line 28: Should this link go to the "Change" page if they have already requested an IR because that is what we do from the hub. Worth clarifying. Otherwise, works as expected. 👍

app/views/handlers/individual_response.py
```diff
@@ -139,7 +138,15 @@ def handle_get(self):
             self.individual_section_id
         )
 
-        if self._list_item_id:
+        if self._request_args.get("journey") == "remove-person":
+            previous_location_url = url_for(
+                "questionnaire.block",
+                list_name=self._list_name,
+                list_item_id=self._list_item_id,
+                block_id="remove-person",
```
MebinAbraham, line 146: I think this should be fetched dynamically, otherwise it will break if the id were to change.

app/views/handlers/list_remove_question.py
```diff
@@ -29,3 +30,20 @@ def handle_post(self):
...
+        journey = "remove-person"
+        context = super().get_context()
+        context["individual_response_enabled"] = self.individual_response_enabled()
+        context["individual_response_url"] = individual_response_url(
+            self._schema.get_individual_response_list(),
+            self._current_location.list_item_id,
+            self._questionnaire_store,
+            journey,
+        )
+        return context
```
MebinAbraham, line 49:
```suggestion
        context = super().get_context()
        context["individual_response_enabled"] = self.individual_response_enabled()
        context["individual_response_url"] = individual_response_url(
            self._schema.get_individual_response_list(),
            self._current_location.list_item_id,
            self._questionnaire_store,
            journey="remove-person",
        )
        return context
```

templates/partials/question.html
berroar: Confirmed the current journey with @donnap573

app/views/handlers/individual_response.py
```diff
@@ -139,7 +138,17 @@ def handle_get(self):
...
         )
 
-        if self._list_item_id:
+        if self._request_args.get("journey") == "remove-person":
+            previous_location_url = url_for(
+                "questionnaire.block",
+                list_name=self._list_name,
+                list_item_id=self._list_item_id,
+                block_id=self._schema.get_remove_block_id_for_list(
+                    self._schema.get_individual_response_list()
```
MebinAbraham, line 147: You already have the list name `self._list_name`, so don't need to fetch it again.

MebinAbraham (approved):

PR #248, opened 2020-08-20
Request IR by text message

iwootten:

Adds the journey to request an individual response by text message shown in the prototype. Encrypts the mobile number from view in the url. Validation is not handled and will follow on a separate card.

NB: I couldn't see why the person_id would be present in the url so I've not included it. This also follows how the post url structure. I'm away until the end of the month so this will need to be picked up by someone else in my absence.

### How to review 

See if you agree with the approach.


tests/integration/individual_response/test_individual_response.py
```diff
@@ -433,6 +433,114 @@ def test_previous_from_confirm_returns_via_hub_route(self):
         self.assertInUrl(f"/individual-response/{list_item_id}/how?journey=hub")
 
 
+class TestIndividualResponseTextHandler(IndividualResponseTestCase):
+    def test_display_mobile_on_confirmation_page(self):
```
JamesGardiner, line 437:
```suggestion
    def test_display_mobile_number_on_confirmation_page(self):
```

templates/individual_response/mobile_confirmation.html
```diff
@@ -0,0 +1,24 @@
+{% extends 'layouts/_base.html' %}
```
JamesGardiner, line 1: I don't think mobile_confirmation is the best name for this template. `sms_confirmation` or `text_message_confirmation` would be more explicit

tests/integration/individual_response/test_individual_response.py
```diff
@@ -433,6 +433,114 @@ def test_previous_from_confirm_returns_via_hub_route(self):
...
+        # Given I navigate to the confirmation page
+        self._add_household_no_primary()
+        self.get(self.individual_section_link)
+        self.get(self.individual_response_link)
+        self.post()
+        self.post({"individual-response-how-answer": "Text message"})
+        self.post({"individual-response-enter-number-answer": "0763921456"})
```
JamesGardiner, line 444: Is this a synthetic number and has it been screened to check it doesn't belong to anyone?

JamesGardiner (commented):
Couple of minor comments

MebinAbraham (commented):
Happy with the changes, feature is as expected.
As discussed, just waiting on decision about lowering the max age for `URLSafeTimedSerializer` should be/whether there should be one and if not it should be `URLSafeSerializer`.

templates/individual_response/mobile_confirmation.html
ajmaddaford: I've gone with `confirmation-text-message.html` and renamed the other template to `confirmation-post.html`.

tests/integration/individual_response/test_individual_response.py
ajmaddaford: I've just done a quick search and it doesn't look like it is. Ofcom lists appropriate ranges for dummy numbers - https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbers-for-drama - so I'll create one in that range and update all of the tests.

app/views/handlers/individual_response.py
```diff
@@ -622,3 +641,190 @@ def handle_post(self):
...
+                ),
+                "placeholders": IndividualResponseHandler._person_name_placeholder_possessive,
+            },
+            "answers": [
+                {
+                    "type": "TextField",
+                    "id": "individual-response-enter-number-answer",
+                    "mandatory": True,
+                    "label": "UK mobile number",
+                    "description": "This will not be stored and only used once to send the access code",
```
MebinAbraham, line 664: Needs to be marked for translations.

app/views/handlers/individual_response.py
```diff
@@ -622,3 +641,190 @@ def handle_post(self):
...
+        )
+
+    @staticmethod
+    def block_definition(mobile_number) -> Mapping:
+        return {
+            "type": "IndividualResponse",
+            "question": {
+                "type": "Question",
+                "id": "individual-response-text-confirm",
+                "title": "Is this mobile number correct?",
```
MebinAbraham, line 763: Needs to be marked for translation.

MebinAbraham (approved):

tests/integration/individual_response/test_individual_response.py
JamesGardiner: That's a good find - not seen it before but definitely good to know

JamesGardiner (approved):

PR #249, opened 2020-08-21
Fix redirecting to summary when unavailable

petechd:
This fixes issues caused by redirecting to summary when summary isn't available. It was caused in CCS schemas by `is_last_block_in_section` check being True while section is not complete if the routing path contained backwards routing. This bug is described on [this Trello card](https://trello.com/c/SyiXbNvr). One of the tests still needs refactoring as it doesn't recreate required scenario.

### How to review 
Check if unit tests are passing.

### Checklist


petechd:
[Bug: Resume panel showing on CCS](https://trello.com/c/SyiXbNvr/4028-bug-resume-panel-showing-on-ccs)

app/questionnaire/router.py
```diff
@@ -63,6 +63,11 @@ def can_access_hub(self):
             if section_id in self.enabled_section_ids
         )
 
+    def can_display_summary_for_section(self, section_id, list_item_id=None):
```
LJBabbage, line 66: I would usually go by our tree make up, so it would be section before summary, and maybe remove the for, can_display_section_summary

app/views/handlers/list_action.py
```diff
@@ -47,7 +47,13 @@ def get_section_summary_url(self):
 
     def get_next_location_url(self):
         if self._return_to == "section-summary":
-            return self.get_section_summary_url()
+            if self.router.can_display_summary_for_section(
```
LJBabbage, line 50: Not sure this is all necessary, I can possibly understand the completeness check in can_display_summary_for_section, but checking for a summary seems unnecessary as  you can't have a return_to = section_summary unless it came from the section summary. The only other way would be playing with URLs, which all bets are then off

tests/app/questionnaire/test_router.py
```diff
@@ -216,15 +218,40 @@ def test_return_to_final_summary_next_location_url(self):
 
         self.assertEqual(next_location, expected_location)
 
+    def test_last_block_on_path_routes_to_first_incomplete_location_in_section_when_section_not_complete(
```
LJBabbage, line 221: test_return_to_first_incomplete_location_when_last_block_in_section_in_progress?

tests/app/questionnaire/test_router.py
```diff
@@ -216,15 +218,40 @@ def test_return_to_final_summary_next_location_url(self):
...
+                    "block_ids": ["block-1"],
+                }
+            ]
+        )
+        router = Router(
+            schema, self.answer_store, self.list_store, progress_store, self.metadata
+        )
+        current_location = Location(section_id="section-1", block_id="block-1")
+        routing_path = RoutingPath(
+            ["block-1", "block-2", "block-1"], section_id="section-1"
```
LJBabbage, line 241: Is this a valid path?

LJBabbage (changes requested):

tests/app/questionnaire/test_list_action.py
```diff
@@ -0,0 +1,64 @@
+import pytest
```
LJBabbage, line 1: Not sure we were testing handlers?

tests/app/questionnaire/test_router.py
petechd: Ok, this one is as you suggested now👍

app/questionnaire/router.py
petechd: Done, thanks👍

app/questionnaire/router.py
```diff
@@ -71,20 +76,23 @@ def get_next_location_url(self, location, routing_path, return_to=None):
...
+        if return_to and is_section_complete:
+            if return_to == "section-summary":
                 return self._get_section_url(location)
 
-            if is_section_complete and return_to == "final-summary":
+            if return_to == "final-summary":
                 return self.get_last_location_in_survey().url()
 
+        # If the routing path contains backwards routing, it is possible for is_last_block_in_section to be True while section is not complete.
+        # For example, when the path is ["block-a", "block-b", block-a"] which is possible when "block-b" routes back to "block-a".
```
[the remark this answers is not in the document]
petechd: Done, thanks👍

tests/app/questionnaire/test_router.py
petechd: Yes, this recreates our ccs schemas behaviour where you are forced to go back to primary person list collector if there is no primary person the first time round.

app/questionnaire/router.py
```diff
@@ -71,20 +76,23 @@ def get_next_location_url(self, location, routing_path, return_to=None):
         Get the next location in the section. If the section is complete, determine where to go next,
         whether it be a summary, the hub or the next incomplete location.
         """
-
-        if return_to:
-            is_section_complete = self._progress_store.is_section_complete(
-                location.section_id, location.list_item_id
-            )
-            if is_section_complete and return_to == "section-summary":
+        is_section_complete = self._progress_store.is_section_complete(
```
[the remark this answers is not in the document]
petechd: I refactored it so focus is on is_section_complete now, thanks!

app/views/handlers/list_action.py
petechd: This is needed for a situation where all elements from a list are removed, if I remove all people in test_list_collector than summary with no householders is not displayed and I go back to list collector initial question..

tests/app/questionnaire/test_list_action.py
petechd: We don't, at least not list action. It's because this resume panel bug cannot be recreated in test schema, as far as I am aware. I mocked a list action situation from ccs schema.

tests/app/questionnaire/test_list_action.py
```diff
@@ -0,0 +1,64 @@
...
+        self.router = Router(
+            self.schema,
+            self.answer_store,
+            self.list_store,
+            self.progress_store,
+            self.metadata,
+        )
+        self.router.can_display_section_summary = Mock(return_value=False)
+
+    def test_list_action(self):
```
berroar, line 51: Might be clearer if this test name is more specific about what's being tested

tests/app/questionnaire/test_list_action.py
```diff
@@ -0,0 +1,64 @@
...
+from app.questionnaire.location import Location
+from app.questionnaire.router import Router
+from app.data_model.answer_store import AnswerStore
+from app.data_model.list_store import ListStore
+from app.data_model.progress_store import ProgressStore, CompletionStatus
+from tests.app.app_context_test_case import AppContextTestCase
+
+
+class TestListAction(AppContextTestCase):
+    def setUp(self):
```
berroar, line 16: I think this setUp method needs to be tidied, some of the bits in the set up seem like they are only relevant to the test below

tests/app/questionnaire/test_list_action.py
petechd: Changed it, thanks👍

tests/app/questionnaire/test_list_action.py
petechd: I had to keep some of them where they are because of linting, but unnecessary attributes were removed.👍

app/views/handlers/list_action.py
```diff
@@ -47,7 +47,13 @@ def get_section_summary_url(self):
 
     def get_next_location_url(self):
         if self._return_to == "section-summary":
-            return self.get_section_summary_url()
+            if self.router.can_display_section_summary(
+                self.parent_location.section_id, self.parent_location.list_item_id
+            ):
+                return self.get_section_summary_url()
+
+            return self.router.get_section_resume_url(self._routing_path, resume=False)
```
ajmaddaford, line 55: What scenario is this for?

JamesGardiner (commented):
This could do with a feature test (well, a regression test, but implemented as a feature test) to ensure this doesn't occur again

tests/app/questionnaire/test_list_action.py
ajmaddaford: I think this test highlights why we don't currently unit test the handlers - so many mocks are needed that it's not clear what is being tested, and there is a high chance we end up testing something that can't happen with schemas. I think we should replace this with a test schema and an integration test.

app/views/handlers/list_action.py
LJBabbage: So this is interesting, firstly I couldn't see where this would happen in ccs, but I can in a couple of our test schemas, test_relationships being one. So if you want a test_schema to do an integration test you can use it with no changes. Add a person and go to the section summary and then add another and it will hit that separation condition you are looking for, because although section-summary is in the url, it naturally will stop the section summary appearing as the section is no longer complete.

Saying that it is edge and I don't think we do this outside of tests. Added to that I think the section endpoint should be handling this, it already goes to the right page its the resume panel thats unwanted, correct? This therefore feels to me we should be able to pass resume true/false to the section endpoint.

ajmaddaford: I understand that, but current runner logic is to only return to the summary if it's possible immediately, else continue the normal journey. In the test_relationships example this would be to return to the parent list collector. There is a separate card in the backlog to look into only showing questions that need completing after navigating from a section summary.

tests/app/questionnaire/test_list_action.py
petechd: This test has been replaced with integration test, thanks!

app/questionnaire/router.py
JamesGardiner, line 79: This doesn't need to be assigned as it's only used once
```suggestion
```

app/questionnaire/router.py
```diff
@@ -71,20 +76,23 @@ def get_next_location_url(self, location, routing_path, return_to=None):
...
+            location.section_id, location.list_item_id
+        )
+        is_last_block_in_section = routing_path[-1] == location.block_id
+        if is_section_complete:
```
JamesGardiner, line 83:
```suggestion
	if self._progress_store.is_section_complete(
		location.section_id, location.list_item_id
	):
```

JamesGardiner (changes requested):

app/questionnaire/router.py
petechd: Done, thanks!👍

app/questionnaire/router.py
petechd: Done, thanks!👍

app/views/handlers/list_action.py
ajmaddaford: I've spoken to @MebinAbraham about this and it's been agreed with Donna that we can drop the logic about going to the next incomplete block if we can't go straight back to the summary. In that scenario it should go to the parent list collector (which it will if you remove the `get_section_resume_url` line).

petechd: Ok, done, thanks👍

LJBabbage (approved):

app/questionnaire/router.py
```diff
@@ -139,13 +146,13 @@ def get_first_incomplete_location_in_survey_url(self):
...
 
-    def get_section_resume_url(self, routing_path):
+    def get_section_resume_url(self, routing_path, resume=True):
         section_key = (routing_path.section_id, routing_path.list_item_id)
 
         if section_key in self._progress_store:
             location = self._get_first_incomplete_location_in_section(routing_path)
             if location:
-                return location.url(resume=True)
+                return location.url(resume=resume) if resume else location.url()
```
ajmaddaford, line 155: I don't think these changes are needed now that the extra call to this method has been removed.

petechd: Removed, thanks👍

app/views/handlers/list_add_question.py
```diff
@@ -12,5 +12,10 @@ def handle_post(self):
             self.parent_block["for_list"]
         )
         self.questionnaire_store_updater.update_answers(self.form.data, list_item_id)
+        if self.questionnaire_store_updater.is_dirty():
+            self._routing_path = self.router.routing_path(
+                section_id=self._current_location.section_id,
+                list_item_id=self._current_location.list_item_id,
+            )
```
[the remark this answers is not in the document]
petechd: Removed, thanks!

ajmaddaford (approved):

JamesGardiner (approved):

PR #253, opened 2020-08-25
Output schema theme in dataLayer

petechd:
This adds an identifying key:value in the `dataLayer[ ]` array to allow Google Analytics to segment events by. There will be scenarios where it is not included at all (e.g. 404 where a user hasn't had an authenticated session). This has been agreed with the analytics engineers.

### How to review 
Change if this new `dataLayer[]` is generated correctly in head element of each page.

### Checklist


petechd:
[Output NISRA or Non-NISRA in dataLayer array (S)](https://trello.com/c/w6ZtGKNl/4087-output-nisra-or-non-nisra-in-datalayer-array-s)

petechd:
[Output NISRA or Non-NISRA in dataLayer array (S)](https://trello.com/c/w6ZtGKNl/4087-output-nisra-or-non-nisra-in-datalayer-array-s)

app/helpers/template_helper.py
```diff
@@ -120,3 +121,13 @@ def safe_content(content):
...
+
+
+def get_data_layer(schema_theme):
+    if schema_theme == "census-nisra":
+        return "dataLayer = [{nisra : true}];"
+
+    if schema_theme == "census":
+        return "dataLayer = [{nisra : false}];"
+
+    return "dataLayer = [];"
```
MebinAbraham, line 133: I think we should return a list instance and not a string, so:
```suggestion
    if schema_theme == "census-nisra":
        return [{"nisra": True}]

    if schema_theme == "census":
        return [{"nisra": False}]

    return []
```
Then in `_base.html` you can do:
```jinja
	<script nonce="{{ csp_nonce() }}">
      dataLayer = {{ data_layer | tojson }}
    </script>
```

petechd: But do we want nisra key as a string? Proposed output is `[{nisra : [true | false] }]`..

MebinAbraham: It is still a valid object. `{"nisra": false}` and `{nisra: false}` are identical in JS.

petechd: Done! thanks!

berroar (approved):

app/helpers/template_helper.py
```diff
@@ -78,6 +78,7 @@ def render_template(template, **kwargs):
         language_code=get_locale().language,
         survey_title=cookie_session.get("survey_title"),
         cdn_url=cdn_url,
+        data_layer=get_data_layer(cookie_session.get("theme")),
```
JamesGardiner, line 81: It's probably worth defining a theme variable here, given the number of times the cookie_session is accessed

MebinAbraham (approved):
The semicolon is now missing from the JS statement but shouldn't matter as it is still valid.

petechd:
> The semicolon is now missing from the JS statement but shouldn't matter as it is still valid.

As far as I know, semicolon in JS is only optional - good practices, indicates end of a statement, but that's all.

app/helpers/template_helper.py
petechd: Done, thanks!

MebinAbraham (approved):

JamesGardiner (approved):

PR #264, opened 2020-09-01
Introduce isort

JamesGardiner:
Enforce import order using `isort` with `black` compatible configuration

### How to review 

### Checklist


Makefile
```diff
@@ -20,6 +20,7 @@ lint:
 
 format:
 	yarn format
+	pipenv run isort **/*.py application.py
```
ajmaddaford, line 23: The lint command should also be updated so that the sort order can be enforced. The isort docs mention the `--diff` flag which could be used for this.

ajmaddaford (changes requested):

Makefile
JamesGardiner: The `run_lint_python.sh` file contains the `isort --check` command and uses the exit code to determine success or failure. This is the same pattern we follow for the other lint commands for our python code.

berroar (approved):

scripts/run_lint_python.sh
```diff
@@ -21,13 +21,16 @@ function display_result {
...
 
-pylint --reports=n --output-format=colorized --rcfile=.pylintrc -j 0 ./app ./tests ./scripts
+pylint --reports=n --output-format=colorized --rcfile=.pylintrc -j 0 **/*.py *.py
 # pylint bit encodes the exit code to allow you to figure out which category has failed.
 # https://docs.pylint.org/en/1.6.0/run.html#exit-codes
 # We want to fail on all errors so don't check for specific bits in the output; but if we did in future, see:
 # http://stackoverflow.com/questions/6626351/how-to-extract-bits-from-return-code-number-in-bash
 display_result $? 1 "Pylint linting check"
 
+isort --check **/*.py *.py
```
MebinAbraham, line 31: Not sure why this isn't working. So I made change to import order to cause a failure, but when I run `make lint` it says there are no issues but if I run `pipenv run isort --check **/*.py *.py` manually then it gets picked up?.

Looks like isort isnt picking up the file at all, so `make format` isn't working either. The file I edited was `questionnaire.py`.

Makefile
MebinAbraham, line 23: This should use the same pattern as lint `**/*.py *.py`

MebinAbraham (changes requested):

scripts/run_lint_python.sh
JamesGardiner: when you say questionnaire.py, do you you mean `app/roots/questionnaire.py`?

JamesGardiner: actually yeah I can see the issue - it's running it based off of the directory that run_test_lint is sat in

JamesGardiner: Ok - I should have RTFM. There is a `--recursive` flag that works correctly

MebinAbraham (approved):

ajmaddaford (approved):

PR #265, opened 2020-09-01
Updating 401 error messages

berroar:
This PR updates the content for 401 error message pages. 

A new error type has been added `NoQuestionnaireStateException`

### How to review 
Check that the content for the 401 error pages match the spreadsheet attached to the Trello Card.

### Checklist


berroar:
[[P] 401 error messages (scenario 1 and 2) (M)](https://trello.com/c/hMomUBJJ/4089-p-401-error-messages-scenario-1-and-2-m)

app/routes/errors.py
```diff
@@ -41,9 +44,15 @@ def _render_error_page(status_code, template=None):
...
-@errors_blueprint.app_errorhandler(CSRFError)
 @errors_blueprint.app_errorhandler(NoTokenException)
+@errors_blueprint.app_errorhandler(CSRFError)
+@errors_blueprint.app_errorhandler(401)
+def no_cookie(error=None):
+    log_error(error, 401)
+    return _render_error_page(401, "no-cookie")
+
+
+@errors_blueprint.app_errorhandler(NoQuestionnaireStateException)
```
ajmaddaford, line 55: This should be one 401 handler with a check for the session cookie, as a cookie can exist and a 401 get raised.

app/routes/individual_response.py
```diff
@@ -27,11 +29,12 @@
 )
 
 
+@login_required
```
ajmaddaford, line 32: The `login_required` should be removed from all uses of this blueprint, as having it here means it doesn't need to be on each use of the blueprint.

app/routes/individual_response.py
```diff
@@ -27,11 +29,12 @@
...
 @individual_response_blueprint.before_request
 def before_individual_response_request():
     metadata = get_metadata(current_user)
     if not metadata:
-        raise NoTokenException(401)
+        raise NoQuestionnaireStateException(401)
```
ajmaddaford, line 37: We don't make use of the code returned here, so we can drop the `401`.

templates/errors/no-cookie.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'layouts/_base.html' %}
+{% from "components/panel/_macro.njk" import onsPanel %}
+
+{% import 'macros/helpers.html' as helpers %}
+
+{% set survey_title = _("Session expired") %}
+
+{% block main %}
+  <h1>{{ _("Sorry there is a problem") }}</h1>
+  <p>{{ _('To access this page you need to <a href="%(account_service_link)s">enter your 16-character access code</a> again to continue your census.', account_service_link=account_service_url) }}</p>
```
ajmaddaford, line 10: I think the error message from the card for this ends after the link i.e. the "again to continue your census" should be removed.

templates/errors/session-expired.html
```diff
@@ -6,22 +6,7 @@
...
-      <p>{{ _('We have saved your progress and you will need to <a href="%(account_service_link)s">enter your unique code</a> to access the survey again.', account_service_link=account_service_url) }}</p>
-    {% endif %}
-    {% if using_edge %}
-      <p>If you have been timed out unexpectedly on more than one occasion, please try using a different browser. If you’re still having problems completing the census, <a href="https://census.gov.uk/contact-us/">contact us</a>.</p>
-    {% endif %}
-  {% endcall %}
-
+  <h1>{{ _("Your session has expired due to inactivity") }}</h1>
+  <p>{{ _("To help protect your information we have timed you out") }}.</p>
+  <p>{{ _('You will need to enter <a href="%(account_service_link)s">enter your 16-character access code</a> again to continue your census.', account_service_link=account_service_url) }}</p>
```
ajmaddaford, line 11: Variables in strings should use `{}` and `.format` as we do in other templates.

templates/errors/no-cookie.html
ajmaddaford: I think `account_service_url` will only be set if there is a cookie. We could introduce a fallback url if there is no cookie, but that could only go to one place which may not work for nisra.

ajmaddaford (changes requested):

templates/errors/no-cookie.html
berroar: I checked with @donnap573 and that extra bit of the error message is correct as we're supposed to be matching the Google spreadsheet attached to the Trello card

app/routes/errors.py
```diff
@@ -41,11 +44,14 @@ def _render_error_page(status_code, template=None):
...
 
 
-@errors_blueprint.app_errorhandler(401)
-@errors_blueprint.app_errorhandler(CSRFError)
 @errors_blueprint.app_errorhandler(NoTokenException)
-def unauthorized(error=None):
+@errors_blueprint.app_errorhandler(CSRFError)
+@errors_blueprint.app_errorhandler(NoQuestionnaireStateException)
+@errors_blueprint.app_errorhandler(401)
+def no_cookie(error=None):
```
MebinAbraham, line 51: I think `unauthorized` makes sense given this handles no cookies and expired cookie.

tests/integration/session/test_timeout.py
```diff
@@ -23,6 +23,7 @@ def test_when_session_times_out_server_side_401_is_returned(self):
         time.sleep(5)
         self.get(self.last_url)
         self.assertStatusUnauthorised()
+        self.assertInBody("Your session has expired due to inactivity")
```
MebinAbraham, line 26: Need to also add a test for the other scenario.

templates/errors/no-cookie.html
ajmaddaford: I had checked the Google sheet, and have just re-checked. The error text for the "Sorry, there is a problem" 401 page doesn't have the text about continuing.

app/routes/errors.py
```diff
@@ -41,11 +44,14 @@ def _render_error_page(status_code, template=None):
...
     log_error(error, 401)
+    if not cookie_session:
+        return _render_error_page(401, "no-cookie")
     return _render_error_page(401, "session-expired")
```
MebinAbraham, line 55: Not sure if this is working as expected. If you now go to a questionnaire endpoint, the error you get first time is not the same one if you refresh the page because you don't have a cookie on the first request, but then we sent one back, so on a refresh you do have a cookie.

templates/errors/no-cookie.html
berroar: Sorry I must have checked the other page 🙈

ajmaddaford: Yeah, I did that first!

app/routes/errors.py
ajmaddaford: I think it should exist for any endpoints that allow a POST. I've had a look at the Flask docs and it looks like we can use [request.url_rule.methods](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Request.url_rule) for this. 

As all templates go via the `render_template` method we could add an `include_csrf_token` template param using something like:

```python
include_csrf_token=True if "POST" in request.url_rule.methods else False
```

Then make the inclusion of the csrf token in `_base.html` conditional on this.

app/helpers/template_helpers.py
```diff
@@ -62,7 +62,7 @@ def render_template(template, **kwargs):
     google_tag_mananger_context = get_google_tag_mananger_context()
     cdn_url = f'{current_app.config["CDN_URL"]}{current_app.config["CDN_ASSETS_PATH"]}'
     contact_us_url = get_contact_us_url(theme, get_locale().language)
-    include_csrf_token = True if "POST" in request.url_rule.methods else False
+    include_csrf_token = bool("POST" in request.url_rule.methods)
```
MebinAbraham, line 65: `bool` isnt needed. `POST in request.url_rule.methods` will return a bool.

templates/errors/no-cookie.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'layouts/_base.html' %}
+{% from "components/panel/_macro.njk" import onsPanel %}
```
iwootten, line 2: I think this may be redundant if the formatting has now been removed?

app/routes/questionnaire.py
```diff
@@ -89,6 +93,19 @@ def get_questionnaire(schema, questionnaire_store):
...
+                schema, questionnaire_store, router.full_routing_path()
+            )
+            submission_handler.submit_questionnaire()
+            return redirect(url_for("post_submission.get_thank_you"))
+
+        return redirect(router.get_first_incomplete_location_in_survey_url())
+
     if not router.can_access_hub():
         redirect_location_url = router.get_first_incomplete_location_in_survey_url()
         return redirect(redirect_location_url)
```
ajmaddaford, line 111: The `is_hub_enabled` check in the POST handling should be removed, and `router.can_access_hub` moved from after the POST handling to be the first check.

```suggestion
    if not router.can_access_hub():
        redirect_location_url = router.get_first_incomplete_location_in_survey_url()
        return redirect(redirect_location_url)
        
    if request.method == "POST":
        if router.is_survey_complete():
            submission_handler = SubmissionHandler(
                schema, questionnaire_store, router.full_routing_path()
            )
            submission_handler.submit_questionnaire()
            return redirect(url_for("post_submission.get_thank_you"))

        return redirect(router.get_first_incomplete_location_in_survey_url())
```

app/routes/questionnaire.py
```diff
@@ -76,7 +80,7 @@ def before_post_submission_request():
     )
 
 
-@questionnaire_blueprint.route("/", methods=["GET"])
+@questionnaire_blueprint.route("/", methods=["GET", "POST"])
 @login_required
```
ajmaddaford, line 84: There are still a number of `login_required` annotations that are no longer necessary now that the two blueprints in this file have it.

app/routes/questionnaire.py
ajmaddaford: It looks like the suggested change for this was made and then reverted?

templates/errors/no-cookie.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'layouts/_base.html' %}
+
+{% set survey_title = _("Session expired") ~ " - " ~ _("Census 2021") %}
+
+{% block main %}
+  <h1>{{ _("Sorry there is a problem") }}</h1>
+  <p>{{ _("To access this page you need to <a href='{url}'>enter your 16-character access code</a>.").format(url = account_service_url) }}</p>
```
ajmaddaford, line 7: `account_service_url` isn't set.

ajmaddaford (changes requested):
I still get the no cookie 401 page, then the timed out 401 page on a refresh.

templates/errors/no-cookie.html
berroar: Is there a fallback url I am able to use for this or does the content need to be changed?

ajmaddaford: There isn't yet, but there probably should be. The difficulty will be we can only have one url, which will cause issues if they have tried to access the NI census. We shouldn't direct them to the wrong access code page.

berroar: I've added the fallback URL now

app/helpers/template_helpers.py
```diff
@@ -57,12 +58,19 @@ def render_template(template, **kwargs):
         get_locale().language, theme or "census"
     )
     page_header_context.update({"title": cookie_session.get("survey_title")})
-    google_tag_mananger_context = get_google_tag_mananger_context()
+    google_tag_manager_context = get_google_tag_manager_context()
     cdn_url = f'{current_app.config["CDN_URL"]}{current_app.config["CDN_ASSETS_PATH"]}'
     contact_us_url = get_contact_us_url(theme, get_locale().language)
+    include_csrf_token = True if cookie_session.get(USER_IK) is not None else False
```
MebinAbraham, line 64: Can be simplified to `cookie_session.get(USER_IK) is not None`

app/helpers/template_helpers.py
```diff
@@ -57,12 +58,19 @@ def render_template(template, **kwargs):
...
+    account_service_url = (
+        cookie_session.get("account_service_url")
+        if cookie_session
+        else "https://census.gov.uk/en/start"
```
MebinAbraham, line 68: This could use the get base url method?.

MebinAbraham: It should prob also default to the base URL even if you have a cookie session but no account service URL given it is not a required claim.

app/helpers/template_helpers.py
```diff
@@ -57,12 +59,21 @@ def render_template(template, **kwargs):
...
     cdn_url = f'{current_app.config["CDN_URL"]}{current_app.config["CDN_ASSETS_PATH"]}'
     contact_us_url = get_contact_us_url(theme, get_locale().language)
+    include_csrf_token = (
+        "POST" in request.url_rule.methods if request.url_rule else False
+    )
+    account_service_url = (
+        f"{CENSUS_BASE_URL}en/start"
+        if not cookie_session or cookie_session.get("account_service_url")
+        else cookie_session.get("account_service_url")
+    )
```
ajmaddaford, line 70: Even if there is no cookie sent in a request, we can still check for values in `cookie_session` as we do for `theme` above. So this can be simplified to:
```suggestion
    account_service_url = cookie_session.get("account_service_url", f"{CENSUS_BASE_URL}en/start")
```

app/routes/questionnaire.py
```diff
@@ -180,7 +166,6 @@ def get_section(schema, questionnaire_store, section_id, list_item_id=None):
 @questionnaire_blueprint.route(
     "<list_name>/<list_item_id>/<block_id>/", methods=["GET", "POST"]
 )
-@login_required
```
ajmaddaford, line 183: There are still a few methods with unnecessary `login_required` decorators now that we do it on the blueprint.

tests/integration/questionnaire/test_questionnaire_endpoints.py
```diff
@@ -36,16 +36,6 @@ def test_post_invalid_questionnaire_location_raises_404(self):
...
-    def test_post_on_questionnaire_route_without_hub_raises_404(self):
-        # Given
-        self.launchSurvey("test_textfield")
-
-        # When
-        self.post(url=f"/questionnaire/")
-
-        # Then
-        self.assertStatusNotFound()
-
```
ajmaddaford, line 48: This test should be updated to what it now does, rather than removed.

tests/integration/session/test_timeout.py
```diff
@@ -23,9 +23,25 @@ def test_when_session_times_out_server_side_401_is_returned(self):
...
+        self.assertInBody("To help protect your information we have timed you out")
+
+    def test_submission_complete_timeout(self):
+        self.launchSurvey("test_timeout")
+        self.post()
+        self.post()
+        time.sleep(4)
+        self.get(self.last_url)
+        self.assertStatusUnauthorised()
+        self.assertInBody("Your census has been submitted")
```
ajmaddaford, line 47: As the string `This page is no longer available` also appears on this page, it would be better to check that than census specific text.

ajmaddaford (changes requested):

tests/integration/questionnaire/test_questionnaire_endpoints.py
```diff
@@ -36,6 +36,16 @@ def test_post_invalid_questionnaire_location_raises_404(self):
         # Then I am shown a 404 page
         self.assertStatusNotFound()
 
+    def test_post_on_questionnaire_route_without_hub_raises_404(self):
```
ajmaddaford:
```suggestion
    def test_post_on_questionnaire_route_without_hub_redirects_to_first_incomplete location(self):
```

app/routes/errors.py
```diff
@@ -41,11 +46,16 @@ def _render_error_page(status_code, template=None):
     )
 
 
-@errors_blueprint.app_errorhandler(401)
-@errors_blueprint.app_errorhandler(CSRFError)
 @errors_blueprint.app_errorhandler(NoTokenException)
+@errors_blueprint.app_errorhandler(CSRFError)
+@errors_blueprint.app_errorhandler(NoQuestionnaireStateException)
+@errors_blueprint.app_errorhandler(401)
```
ajmaddaford, line 52: Minor, but I can't see why the ordering changed here.
```suggestion
@errors_blueprint.app_errorhandler(401)
@errors_blueprint.app_errorhandler(CSRFError)
@errors_blueprint.app_errorhandler(NoTokenException)
@errors_blueprint.app_errorhandler(NoQuestionnaireStateException)

```

tests/integration/questionnaire/test_questionnaire_endpoints.py
```diff
@@ -36,15 +36,17 @@ def test_post_invalid_questionnaire_location_raises_404(self):
...
+    ):
         # Given
         self.launchSurvey("test_textfield")
 
         # When
         self.post(url=f"/questionnaire/")
 
         # Then
-        self.assertStatusNotFound()
+        self.assertInUrl("name-block")
```
MebinAbraham, line 49: Was this changed intentionally to not 404 on a POST to the questionnaire root?

templates/errors/submission-complete.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'layouts/_base.html' %}
```
MebinAbraham, line 1: This page has an exit button which doesn't do anything.

app/routes/questionnaire.py
MebinAbraham: @ajmaddaford was your suggestion with the intention to also go to first location even if hub is not enabled?. Which is what it will do, even though previously it would 404.

tests/integration/questionnaire/test_questionnaire_endpoints.py
berroar: Yeah that bit of code was removed in one of the changes above

app/routes/questionnaire.py
ajmaddaford: Yes. It seemed odd (and unnecessary) for the POST to behave differently from the GET.

templates/errors/submission-complete.html
berroar: I can't see a button when testing it? 
image

MebinAbraham: It will show up when your session hasn't expired. So try what you did but on a diff schema that doesn't timeout that quick.

berroar: I've added a fix for this

templates/errors/no-cookie.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'layouts/_base.html' %}
+
+{% set survey_title = _("Session expired") ~ " - " ~ _("Census 2021") %}
```
MebinAbraham, line 3: The page title should match the h1, so this should be `Sorry there is a problem | Census 2021`

templates/errors/session-expired.html
```diff
@@ -1,27 +1,9 @@
...
-    <p>{{ _("To help protect your information we have signed you out") }}.</p>
-    {% if account_service_url %}
-      <p>{{ _('We have saved your progress and you will need to <a href="%(account_service_link)s">enter your unique code</a> to access the survey again.', account_service_link=account_service_url) }}</p>
-    {% endif %}
-    {% if using_edge %}
-      <p>If you have been timed out unexpectedly on more than one occasion, please try using a different browser. If you’re still having problems completing the census, <a href="https://census.gov.uk/contact-us/">contact us</a>.</p>
-    {% endif %}
-  {% endcall %}
-
+  <h1>{{ _("Your session has expired due to inactivity") }}</h1>
```
MebinAbraham, line 6: The page title should match h1. Would be good to get confirmation but that is the general principle.

templates/errors/submission-complete.html
```diff
@@ -0,0 +1,8 @@
...
+
+{% set page_title = _("Submission Complete") ~ " - " ~ _("Census 2021") %}
+
+{% block main %}
+<h1>{{ _("This page is no longer available") }}</h1>
```
MebinAbraham, line 6: Again here, also the page title makes sense, it would be good to check where we can keep it as is, or the title should match h1.

templates/errors/no-cookie.html
berroar: So as discussed checked with Laura and this will be updated to `Page is not available`

tests/integration/session/test_timeout.py
```diff
@@ -23,9 +23,25 @@ def test_when_session_times_out_server_side_401_is_returned(self):
...
+        self.assertInBody("To help protect your information we have timed you out")
+
+    def test_submission_complete_timeout(self):
+        self.launchSurvey("test_timeout")
+        self.post()
+        self.post()
+        time.sleep(4)
+        self.get(self.last_url)
+        self.assertStatusUnauthorised()
+        self.assertInBody("This page is no longer available")
```
MebinAbraham, line 47: I would also assert the page title here.

tests/integration/session/test_timeout.py
```diff
@@ -23,9 +23,25 @@ def test_when_session_times_out_server_side_401_is_returned(self):
         time.sleep(5)
         self.get(self.last_url)
         self.assertStatusUnauthorised()
+        self.assertInBody("Your session has expired due to inactivity")
+
+    def test_alternate_401_page_is_displayed_when_no_cookie(self):
+        self.get("/session")
+        self.assertStatusUnauthorised()
+        self.assertInBody("Sorry there is a problem")
```
MebinAbraham, line 31: I would also assert the page title here.

tests/integration/session/test_timeout.py
```diff
@@ -23,9 +23,25 @@ def test_when_session_times_out_server_side_401_is_returned(self):
...
 
     def test_schema_defined_timeout_cant_be_higher_than_server(self):
         self.launchSurvey("test_timeout")
         time.sleep(4)
         self.get(self.last_url)
         self.assertStatusUnauthorised()
+        self.assertInBody("To help protect your information we have timed you out")
```
MebinAbraham, line 38: I would also assert the page title here.

MebinAbraham (approved):

ajmaddaford (approved):

PR #267, opened 2020-09-02
Add new post submission exception

iwootten:

Add a new post submission exception/error which displays after a submission and token timeout.

### How to review 

Open test_timeout and let it expire, refresh and note the content of the page.
Open test_timeout again, step through and submit, let it expire, refresh and note the (different) content of the page

### Checklist

### DIFF
--- a/app/routes/questionnaire.py
+++ b/app/routes/questionnaire.py
@@ -12,7 +12,7 @@ from flask_login import current_user, login_required
 from structlog import get_logger
 from werkzeug.exceptions import NotFound
 
-from app.authentication.no_token_exception import NoTokenException
+from app.authentication.exceptions import NoTokenException, PostSubmissionException
 from app.globals import get_metadata, get_session_store, get_session_timeout_in_seconds
 from app.helpers.language_helper import handle_language
 from app.helpers.schema_helpers import with_schema
@@ -43,6 +43,10 @@ post_submission_blueprint = Blueprint(
 @questionnaire_blueprint.before_request
 def before_questionnaire_request():
     metadata = get_metadata(current_user)
+
+    if cookie_session.get("submitted", False):
+        raise PostSubmissionException(401)
+
     if not metadata:
         raise NoTokenException(401)
 
@@ -66,7 +70,10 @@ def before_questionnaire_request():
 @post_submission_blueprint.before_request
 def before_post_submission_request():
     session_store = get_session_store()
+
     if not session_store or not session_store.session_data:
+        if cookie_session.get("submitted", False):
+            raise PostSubmissionException(401)
         raise NoTokenException(401)
 
     session_data = session_store.session_data
--- a/templates/errors/submission-complete.html
+++ b/templates/errors/submission-complete.html
@@ -1,0 +1,8 @@
+{% extends 'layouts/_base.html' %}
+
+{% set page_title = _("No token") ~ " - " ~ _("Census 2021") %}
+
+{% block main %}
+<h1>{{ _("This page is no longer available") }}</h1>
+<p>{{ _("Your census has been submitted.") }}</p>
+{% endblock %}
--- a/templates/errors/submission-complete.html
+++ b/templates/errors/submission-complete.html
@@ -1,6 +1,6 @@
 {% extends 'layouts/_base.html' %}
 
-{% set page_title = _("No token") ~ " - " ~ _("Census 2021") %}
+{% set page_title = _("Submission Complete") ~ " - " ~ _("Census 2021") %}
 
 {% block main %}
 <h1>{{ _("This page is no longer available") }}</h1>
--- a/templates/errors/submission-complete.html
+++ b/templates/errors/submission-complete.html
@@ -4,5 +4,5 @@
 
 {% block main %}
 <h1>{{ _("This page is no longer available") }}</h1>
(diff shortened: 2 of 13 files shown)


iwootten:
[[P] Add post submission cookie state and exception handler (scenario 3) (M)](https://trello.com/c/vkyTBBtC/4096-p-add-post-submission-cookie-state-and-exception-handler-scenario-3-m)

berroar (approved):
Tested this change, LGTM

tests/functional/spec/exit.spec.js
```diff
@@ -20,6 +20,6 @@ describe("Post submission exit", () => {
     $(CensusThankYouPage.exit()).click();
     browser.back();
     expect(browser.getUrl()).to.contain("submitted/thank-you");
-    expect($("body").getHTML()).to.contain("Your session has expired");
+    expect($("body").getHTML()).to.contain("Your census has been submitted");
```
LJBabbage, line 23: Probably needs the extra string ("This page is no longer available") as the thank you page could feasibly include "Your census has been submitted"

tests/integration/session/test_timeout.py
```diff
@@ -29,3 +29,13 @@ def test_schema_defined_timeout_cant_be_higher_than_server(self):
...
+        self.assertInBody("To help protect your information we have signed you out")
+
+    def test_submission_complete_timeout(self):
+        self.launchSurvey("test_timeout")
+        self.post()
+        self.post()
+        time.sleep(4)
+        self.get(self.last_url)
+        self.assertStatusUnauthorised()
+        self.assertInBody("Your census has been submitted")
```
LJBabbage, line 41: As above

templates/errors/submission-complete.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'layouts/_base.html' %}
```
LJBabbage, line 1: I think the story is a bit short here on design, not you. I was imagining it to look like the other session timeout page (i.e the green box), but with different text

LJBabbage (changes requested):

tests/integration/session/test_timeout.py
iwootten: Point taken, but in this case it shouldn't be needed as it's already checked it's a 401

LJBabbage (commented):
Should this also not be themed based?

tests/integration/session/test_timeout.py
LJBabbage: True, just the functional test then

templates/errors/submission-complete.html
iwootten: I think this is to follow the other card (as it was broken out of it) which states the green panel is to be removed. https://trello.com/c/hMomUBJJ/4089-p-401-error-messages-scenario-1-and-2-m

iwootten:
> Should this also not be themed based?

I think it is isn't it? 
Screenshot from 2020-09-03 14-29-51
Screenshot from 2020-09-03 14-28-54

LJBabbage:
> I think it is isn't it? Or do you mean it should _only_ be displayed for census theme?

The later, as naturally you wouldn't want "your census has been submitted" for a business survey

iwootten:
> > I think it is isn't it? Or do you mean it should _only_ be displayed for census theme?
> 
> The later, as naturally you wouldn't want "your census has been submitted" for a business survey

Checked with @donnap573 and @berroar here: https://surveydatacollection.slack.com/archives/CPFEPBT27/p1599141000026100 and it's been decided it's the same for all surveys for now.

LJBabbage (approved):

templates/errors/submission-complete.html
MebinAbraham, line 1: This page has an exit button that doesn't link anywhere. The button only appear when the cookie session is still valid therefore we should link to the normal exit page which atm is the census home page.

app/routes/questionnaire.py
```diff
@@ -37,6 +37,10 @@
 @questionnaire_blueprint.before_request
 def before_questionnaire_request():
     metadata = get_metadata(current_user)
+
+    if cookie_session.get("submitted", False):
+        raise PostSubmissionException(401)
```
MebinAbraham, line 42: Do we need a new exceptioon?. I think we should only raise a new one if we want to potentially log and alert based on that. This could be done similar to Rhys' change where the check is done in the error handler. This could raise `NoQuestionnaireStateException` which Rhys introduced. Then in the error handler we would do:
```py
if cookie_session.get("submitted", False):
    return _render_error_page(401, "submission-complete")
```
https://github.com/ONSdigital/eq-questionnaire-runner/pull/265/files#diff-0dd9f13aea9445b78356c3213cb91ab0R48-R56

So these 2 PRs are slightly inconsistent between each other right now.
Thoughts?

MebinAbraham (changes requested):

app/routes/questionnaire.py
iwootten: There's no need to have multiple exceptions if one covers all the requirements, especially as it is just empty. Given that's the case, I'd just say I make those changes after Rhys's merges.

iwootten:
Closing this and making it a part of https://github.com/ONSdigital/eq-questionnaire-runner/pull/265

PR #269, opened 2020-09-03
Extend schemas to resolve suggestions url language

petechd:
This adds new transform to help resolving suggestions urls language for NISRA schemas. It needs merging before [this schemas PR](https://github.com/ONSdigital/eq-questionnaire-schemas/pull/152) and after [validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/60).

### How to review 
Two unit tests has been added to support this and reach 100% test coverage.

### Checklist

### DIFF
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -185,6 +185,4 @@ class PlaceholderTransforms:
         return ""
 
     def format_suggestions_url(self, url, list_json):
-        if self.language == "en":
-            return url + list_json
-        return ""
+        return "{}{}".format(url, list_json) if self.language is "en" else ""
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -183,3 +183,8 @@ class PlaceholderTransforms:
             return item
 
         return ""
+
+    def format_suggestions_url(self, url, list_json):
+        if self.language == "en":
+            return url + list_json
+        return ""
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -185,4 +185,4 @@ class PlaceholderTransforms:
         return ""
 
     def format_suggestions_url(self, url, list_json):
-        return "{}{}".format(url, list_json) if self.language is "en" else ""
+        return "{}{}".format(url, list_json) if self.language == "en" else ""
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -185,5 +185,5 @@ class PlaceholderTransforms:
 
         return ""
 
-    def format_suggestions_url(self, url, list_json):
+    def format_nisra_suggestions_url(self, url, list_json):
         return "{}{}".format(url, list_json) if self.language == "en" else ""
--- a/tests/app/questionnaire/test_placeholder_transforms.py
+++ b/tests/app/questionnaire/test_placeholder_transforms.py
@@ -206,7 +206,7 @@ class TestPlaceholderParser(unittest.TestCase):
 
     def test_nisra_english_suggestions_url(self):
         assert (
-            self.transforms.format_suggestions_url(
+            self.transforms.format_nira_suggestions_url(
                 "https://cdn.eq.census-gcp.onsdigital.uk/data/v3.0.0/ni/en/",
                 "countries-of-birth.json",
             )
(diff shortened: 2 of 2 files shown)


petechd:
[[C] Extend schema build process to resolve suggestions URL language (M)](https://trello.com/c/WSPjAC5y/4079-c-extend-schema-build-process-to-resolve-suggestions-url-language-m)

app/questionnaire/placeholder_transforms.py
```diff
@@ -183,3 +183,8 @@ def first_non_empty_item(self, items):
             return item
 
         return ""
+
+    def format_suggestions_url(self, url, list_json):
+        if self.language == "en":
+            return url + list_json
```
iwootten, line 189: We should use .format() to concatenate strings.

app/questionnaire/placeholder_transforms.py
```diff
@@ -183,3 +183,8 @@ def first_non_empty_item(self, items):
             return item
 
         return ""
+
+    def format_suggestions_url(self, url, list_json):
+        if self.language == "en":
```
iwootten, line 188: Don't we also want to concatenate when it's "cy"?

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -203,3 +203,23 @@ def test_first_non_empty_item_no_valid(self):
         list_to_filter = [None, None]
 
         assert self.transforms.first_non_empty_item(list_to_filter) == ""
+
```
iwootten, line 206: There should be unit tests for the english and welsh language versions of england/wales too.

app/questionnaire/placeholder_transforms.py
```diff
@@ -183,3 +183,8 @@ def first_non_empty_item(self, items):
...
+            return url + list_json
+        return ""
```
iwootten, line 190: You could achieve the same behaviour here supplying the contatenated url as the arg:

    def format_suggestions_url(self, url):
        return url if self.language in ["en", "cy"] else return ""

iwootten (changes requested):

app/questionnaire/placeholder_transforms.py
petechd: No, because "cy" is not used in this transform, it's a different placeholder altogether. If I started transforming "cy" and "en" I would need to know which "en" I'm transforming from a region code.

tests/app/questionnaire/test_placeholder_transforms.py
petechd: Wales/England schemas are not transformed here👍

app/questionnaire/placeholder_transforms.py
iwootten: Then I think this transform should be renamed `format_nisra_suggestions_url` or something similar as at the moment ithe naming implies it can be used against any url.

tests/app/questionnaire/test_placeholder_transforms.py
iwootten: As mentioned, we should name the method more appropriately as this is currently unclear.

app/questionnaire/placeholder_transforms.py
petechd: Done, thanks👍

app/questionnaire/placeholder_transforms.py
petechd: Done, thanks👍

app/questionnaire/placeholder_transforms.py
petechd: I had it as `format_nisra_suggestions_url` at first but changed it to more universal name after and forgot to change it back when I realised that region code for Eng "en" will be needed🤦‍♂️.

tests/app/questionnaire/test_placeholder_transforms.py
petechd: Done, thanks!

MebinAbraham:
I think there was some misunderstand, as this PR isn't needed. The idea was to resolve the suggestion URL at build time as we know what they should be for a given region/language. In the schemas repo, after the build and translation phase, we need to resolve the suggestion URLs.

iwootten:
> I think there was some misunderstanding, as this PR isn't needed. The idea was to resolve the suggestion URL at build time as we know what they should be for a given region/language. In the schemas repo, after the build and translation phase, we need to resolve the suggestion URL, therefore, there is no placeholder for runner to map, the values should already be resolved.

I was wondering about this :smile:

PR #276, opened 2020-09-08
Add Publisher Backend - Pub/Sub 

MebinAbraham:
Adds support for GCP's Pub/Sub as a publisher backend for fulfilment requests (IR).

### How to review 
Ensure you can request an IR using Pub/Sub and that the fulfilment request payload is as expected.

For dev, the fulfilment request is outputted to logs. To use GCP Pub/Sub, you need to:

---

1. **__INITIAL SET UP__**

---

**TESTING LOCALLY**: (Using launcher and runner locally with GCP Pub/Sub)
	
 - Build a GCP environment using Concourse/Terraform. (this will create a Pub/Sub topic called `eq-fulfilment-topic`)
 - Authenticate with gcloud SDK using:
	- `gcloud auth login`
	- `gcloud auth application-default login`
	- `gcloud config set project <your-project-id>`
 - Set the following env vars:
	- `EQ_PUBLISHER_BACKEND: pubsub`
	- `EQ_PUB_SUB_TOPIC_ID: eq-fulfilment-topic`
 - Install the latest Pipfile.lock packages.

**TESTING IN GCP**: (Using deployed launcher and runner with GCP Pub/Sub)
  - Build a GCP environment using Concourse/Terraform while ensuring you are using the correct branch and image tag. For example, when deploying with Concourse, be sure to update the Git source **branch** and deployment **image tag** to:
	
	Note: Since we don't push to GCR for runner, you will need to build and push your own image to GCR to deploy to Concourse. Alternatively you can use the image I have already pushed.
	- **runner**: add-pub-sub-publisher
	
	- **launcher**: add-census-claims-to-ir-schema

---

2. __**TEST MESSAGE PUBLICATIONS**__

---
  - Go to Pub/Sub in GCP Console and click the `eq-fulfilment-topic` topic.
  - Click `VIEW MESSAGES`, this will prompt you to create a `Subscription` which is needed to view messages.
  - Request an IR using `test_individual_response` schema. If you are testing locally, you will need to use https://github.com/ONSdigital/eq-questionnaire-launcher/pull/21 (the branch has already been updated in the docker-compose YAML)
  - Click `VIEW MESSAGES` in the GCP UI, select the subscription you just created and click `PULL`. You should now see the IR response fulfilment request payload. Test multiple scenarios and ensure the payload matches the specification.

Known issue:
- The GCP Pub/Sub library relies on `grpc` and threading for which `gevent` only has experimental support for. Further investigation will be done to try to use `gthread` over `gevent`.

### Checklist


docker-compose-dev-linux.yml
```diff
@@ -16,7 +16,7 @@ services:
             - "6379:6379"
 
     eq-questionnaire-launcher:
-        image: onsdigital/eq-questionnaire-launcher:latest
+        image: onsdigital/eq-questionnaire-launcher:add-census-claims-to-ir-schema
```
MebinAbraham, line 19: Reminder to revert before merge
```suggestion
        image: onsdigital/eq-questionnaire-launcher:latest
```

docker-compose-dev-mac.yml
```diff
@@ -12,7 +12,7 @@ services:
             - "6379:6379"
 
     eq-questionnaire-launcher:
-        image: "onsdigital/eq-questionnaire-launcher:latest"
+        image: "onsdigital/eq-questionnaire-launcher:add-census-claims-to-ir-schema"
```
MebinAbraham, line 15: Reminder to revert before merge

```suggestion
        image: "onsdigital/eq-questionnaire-launcher:latest
```

docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
       - eq-env
 
   eq-questionnaire-runner:
-    image: onsdigital/eq-questionnaire-runner:latest
+    image: onsdigital/eq-questionnaire-runner:add-census-claims-to-ir-schema
```
MebinAbraham, line 16: Reminder to revert before merge
```suggestion
    image: onsdigital/eq-questionnaire-runner:latest
```

LJBabbage (commented):
Getting an Exception in GCP: Unknown EQ_PUBLISHER_BACKEND. This using our most recent pipeline with the branch add-pub-sub-publisher

MebinAbraham:
> Getting an Exception in GCP: Unknown EQ_PUBLISHER_BACKEND. This using our most recent pipeline with the branch add-pub-sub-publisher

You need to update both the branch and the image tag.

LJBabbage (commented):
As discussed, I am seeing latency issues when deployed to GCP

.development.env
```diff
@@ -16,7 +17,7 @@ EQ_SECRETS_FILE=dev-secrets.yml
 EQ_KEYS_FILE=dev-keys.yml
 AWS_DEFAULT_REGION=eu-west-1
 GUNICORN_KEEP_ALIVE=2
-GUNICORN_CMD_ARGS=-c gunicorn_config.py
+GUNICORN_CMD_ARGS="-c gunicorn_config.py"
```
MebinAbraham, line 20: Revert quotes
```suggestion
GUNICORN_CMD_ARGS=-c gunicorn_config.py
```

app/publisher/publisher.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class PubSub(PublisherClient):
+    def __init__(self, topic_id):
+        super().__init__()
+
+        _, project_id = google.auth.default()
+        self.topic_path = super().topic_path(project_id, topic_id)
+
+    def _publish(self, message):
```
LJBabbage, line 18: Just for readability and my preference is to put private methods after public. It also makes the class flow better when reading, if I am following the code naturally publish_and_resolve comes before _publish

app/publisher/publisher.py
```diff
@@ -0,0 +1,37 @@
...
+        logger.info("publishing message")
+        publish_future: Future = super().publish(self.topic_path, message)
+
+        return publish_future
+
+    def publish_and_resolve_message(self, message: bytes):
+        publish_future = self._publish(message)
+        try:
+            # Resolve the future
```
LJBabbage, line 27: This comment doesn't add much for me

app/publisher/publisher.py
```diff
@@ -0,0 +1,37 @@
...
+            publish_future.result()
+        except Exception as ex:  # pylint:disable=broad-except
+            raise PublicationFailed(ex)
+
+
+class LogPublisher:
+    @staticmethod
+    def publish_and_resolve_message(message):
+        logger.info("publishing message")
+        logger.info("message", message=message)
```
LJBabbage, line 37: If this is needed then it surely can be combined into one output. Any issue with potentially putting the message in the logs?

app/publisher/publisher.py
```diff
@@ -0,0 +1,37 @@
...
+        try:
+            # Resolve the future
+            publish_future.result()
+        except Exception as ex:  # pylint:disable=broad-except
+            raise PublicationFailed(ex)
+
+
+class LogPublisher:
+    @staticmethod
+    def publish_and_resolve_message(message):
```
LJBabbage, line 35: publish_and_resolve_message? Isn't this just logging?

app/setup.py
```diff
@@ -324,6 +327,22 @@ def setup_submitter(application):
         raise Exception("Unknown EQ_SUBMISSION_BACKEND")
 
 
+def setup_publisher(application):
+    if application.config["EQ_PUBLISHER_BACKEND"] == "pubsub":
+        topic_id = application.config.get("EQ_PUB_SUB_TOPIC_ID")
```
LJBabbage, line 332: "get" on one but not the other?

app/setup.py
```diff
@@ -324,6 +327,22 @@ def setup_submitter(application):
...
+
+        if not topic_id:
+            raise Exception("Setting EQ_PUB_SUB_TOPIC_ID Missing")
```
LJBabbage, line 335: Isn't it just "EQ_PUB_SUB_TOPIC_ID Missing" as we have not set anything yet?

app/setup.py
```diff
@@ -324,6 +327,22 @@ def setup_submitter(application):
...
+
+        application.eq["publisher"] = PubSub(topic_id)
+
+    elif application.config["EQ_PUBLISHER_BACKEND"] == "log":
```
LJBabbage, line 339: Unsure why we have a log publisher, what is it used for?

app/views/handlers/individual_response.py
```diff
@@ -132,6 +141,58 @@ def form(self):
     def get_context(self):
         return build_question_context(self.rendered_block, self.form)
 
+    def _get_fulfilment_code(self, mobile_number=None):
```
LJBabbage, line 144: I wonder if this shouldn't be fulfilment_type rather than mobile_number, it makes me think fulfilment codes are usually mobile numbers, perhaps they are. Also its an assumption that if its not mobile it must be postal. Probably more explicit is better here when calling _get_fulfilment_code

app/views/handlers/individual_response.py
```diff
@@ -132,6 +141,58 @@ def form(self):
...
+            GB_WLS_REGION_CODE: "P_UAC_UACIP2B",
+            GB_NIR_REGION_CODE: "P_UAC_UACIP4",
+        }
+
+        region_code = self._metadata["region_code"]
+        return (
+            fulfilment_codes_map_for_sms[region_code]
+            if mobile_number
+            else fulfilment_codes_map_for_postal[region_code]
+        )
```
LJBabbage, line 161: If it was more explicit I would be tempted to put it in one dict, something like 

```
fulfilment_codes = {
            "sms": {
                GB_ENG_REGION_CODE: "UACITA1",
                GB_WLS_REGION_CODE: "UACITA2B",
                GB_NIR_REGION_CODE: "UACITA4",
            },
            "postal": {
                GB_ENG_REGION_CODE: "P_UAC_UACIP1",
                GB_WLS_REGION_CODE: "P_UAC_UACIP2B",
                GB_NIR_REGION_CODE: "P_UAC_UACIP4",
            }

        }
        region_code = self._metadata["region_code"]
        return fulfilment_codes.get(type, {}).get(region_code)
```

app/views/handlers/individual_response.py
```diff
@@ -132,6 +141,58 @@ def form(self):
...
+
+    def _get_fulfilment_request_payload(self, mobile_number=None):
```
LJBabbage, line 163: Same as above

tests/app/views/handlers/test_individual_response.py
```diff
@@ -0,0 +1,175 @@
+# pylint: disable=protected-access
```
LJBabbage, line 1: Finally got your handler test

tests/integration/test_app_create.py
```diff
@@ -227,6 +229,50 @@ def test_defaults_to_adding_the_log_submitter_to_the_application(self):
...
+            application = create_app(self._setting_overrides)
+
+        # Then
+        assert isinstance(application.eq["publisher"], PubSub)
+
+    def test_pub_sub_topic_id_not_set_raises_exception(self):
+        # Given
+        self._setting_overrides["EQ_PUBLISHER_BACKEND"] = "pubsub"
+
+        # WHEN
```
LJBabbage, line 262: This must be an important WHEN

LJBabbage (changes requested):
Deployed into GCP and bar some latency issue, works as expected. Some minor suggestions, nothing major

app/publisher/publisher.py
```diff
@@ -0,0 +1,37 @@
...
+        _, project_id = google.auth.default()
+        self.topic_path = super().topic_path(project_id, topic_id)
+
+    def _publish(self, message):
+        logger.info("publishing message")
+        publish_future: Future = super().publish(self.topic_path, message)
+
+        return publish_future
+
+    def publish_and_resolve_message(self, message: bytes):
```
ajmaddaford, line 24: This method should be just called `publish` as the resolve is irrelevant to the calling method.

app/publisher/publisher.py
```diff
@@ -0,0 +1,37 @@
+import google
```
ajmaddaford, line 1: We should define an abstract publisher base class to make it clear how a publisher should be implemented.

app/settings.py
```diff
@@ -49,6 +49,9 @@ def get_env_or_fail(key):
 EQ_SECRETS_FILE = os.getenv("EQ_SECRETS_FILE", "secrets.yml")
 EQ_KEYS_FILE = os.getenv("EQ_KEYS_FILE", "keys.yml")
 
+EQ_PUBLISHER_BACKEND = os.getenv("EQ_PUBLISHER_BACKEND")
+EQ_PUB_SUB_TOPIC_ID = os.getenv("EQ_PUB_SUB_TOPIC_ID")
```
ajmaddaford, line 53: The app could publish to more than one topic. This variable name should be specific to its use. The Terraform variable is `eq-fulfilment-topic`, so this could be `EQ_FULFILMENT_TOPIC`.

In the Terraform we refer to this as "name" rather than "id". We should be consistent.

app/publisher/publisher.py
```diff
@@ -0,0 +1,37 @@
...
+        logger.info("publishing message")
+        publish_future: Future = super().publish(self.topic_path, message)
+
+        return publish_future
+
+    def publish_and_resolve_message(self, message: bytes):
+        publish_future = self._publish(message)
+        try:
+            # Resolve the future
+            publish_future.result()
```
ajmaddaford, line 28: According to the docs this method returns a message id. We should capture that use it in a log message to say it was published.

app/publisher/publisher.py
```diff
@@ -0,0 +1,37 @@
...
+        except Exception as ex:  # pylint:disable=broad-except
```
ajmaddaford, line 29: Why the broad exception here?

app/setup.py
ajmaddaford, line 332: The topic id should be used on publishing rather than on initialisation. The publisher could be used to publish to multiple topics.

app/views/handlers/individual_response.py
```diff
@@ -14,6 +18,10 @@
 from app.questionnaire.router import Router
 from app.views.contexts.question import build_question_context
 
+GB_ENG_REGION_CODE = "GB-ENG"
+GB_WLS_REGION_CODE = "GB-WLS"
+GB_NIR_REGION_CODE = "GB=NIR"
```
ajmaddaford, line 23:
```suggestion
GB_NIR_REGION_CODE = "GB-NIR"
```

app/publisher/publisher.py
MebinAbraham: My personal preference is to define all underscore/Dunder methods first. Obviously in Python in most cases the order you define methods doesn't matter but I did it like this as it is a common convention to define methods before they are invoked.

app/publisher/publisher.py
MebinAbraham: I added it as it may not be obvious to those who have not worked with Futures that `result()` is resolving the future.

app/publisher/publisher.py
MebinAbraham: Combined these into 1 log message. Also, this is only for dev so logging the message has no bad consequences.

app/publisher/publisher.py
MebinAbraham: Yup, but it needs to define the same interface as all other publishers (Pub/Sub) as the app has no knowledge of which publisher it is using. This is now enforced using Pythons ABC.

app/setup.py
MebinAbraham: Have changed.

app/setup.py
MebinAbraham: It is saying "The application setting eq_pub_sub_topic_id is missing" and not that "we are setting something". I followed the existing pattern used in runner but happy to change this and existing ones as I can see where the confusion came from.

app/setup.py
MebinAbraham: It is used for development same as Log Submitter. We don't use the actual GCP Pub/Sub as we decided not to use an emulator.

app/views/handlers/individual_response.py
MebinAbraham: My mistake, this should have not passed the number. Have changed to be as above.

app/views/handlers/individual_response.py
MebinAbraham: This one uses the numbers and I don't think it is worth adding a "type" filter. But yes I could essentially split these into get payload for sms/postal. If we think this method is doing too much then I am happy to split this.

tests/integration/test_app_create.py
MebinAbraham: The one I copied must have also been important 😄 . Have updated.

app/publisher/publisher.py
MebinAbraham: Updated but doing this and the ABC change meant I had to remove the `PublisherClient` inheritance due to miss matching method signatures.

app/publisher/publisher.py
MebinAbraham: Done but had to remove the `PublisherClient` inheritance due to miss matching method signatures.

app/settings.py
MebinAbraham: Have updated the variable name. However, as for ID/Name, the correct term here would be "id" as topic name is referred to as the full topic path in GCP. But happy to use name if we want to be consistent with existing Terraform implementation but also happy to updated terraform implementation to id. Let me know you preference.

app/publisher/publisher.py
MebinAbraham: Wasn't sure if we wanted to log it but have now added it to the log message.

app/publisher/publisher.py
MebinAbraham: Because when resolving the future, any number of exceptions can be raised. The design pattern is:
```
# Attempt to get the exception if there is one.
# If there is not one, then we know everything worked, and we can
# return an appropriate value.
```
The appropriate value in our case being the message id. If we want to know about specific errors like "TimeoutError" then that can be handled explicitly.

app/setup.py
MebinAbraham: Have added as part of the ABC changes. Also, I have kept the logic of raise an error if the fulfilment topic env var is not set. However, I could also handle it not being set on publish if we didn't want to do it on start up.

app/views/handlers/individual_response.py
MebinAbraham: Addressed in https://github.com/ONSdigital/eq-questionnaire-runner/pull/276#discussion_r487585426

app/views/handlers/individual_response.py
```diff
@@ -130,6 +139,58 @@ def form(self):
...
+                    "caseId": self._metadata["case_id"],
+                    "contact": mobile_number_mapping,
+                }
+            },
+        }
+
+        return json.dumps(message).encode("utf-8")
+
+    def _publish_fulfilment_request(self):
+        message = self._get_fulfilment_request_payload()
```
ajmaddaford, line 190: This never passes a mobile number.

app/views/handlers/individual_response.py
```diff
@@ -826,6 +888,7 @@ def handle_get(self):
     def handle_post(self):
         if self.selected_option == self.confirm_option:
             self._update_section_status(CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED)
+            self._publish_fulfilment_request()
```
ajmaddaford, line 891: The mobile number needs to be passed here.

app/views/handlers/individual_response.py
```diff
@@ -130,6 +139,58 @@ def form(self):
...
+                "fulfilmentRequest": {
+                    **individual_case_id_mapping,
+                    "fulfilmentCode": self._get_fulfilment_code(fulfilment_type),
+                    "caseId": self._metadata["case_id"],
+                    "contact": mobile_number_mapping,
+                }
+            },
+        }
+
+        return json.dumps(message).encode("utf-8")
```
ajmaddaford, line 187: As discussed, the building of the payload should be in it's own class, rather than part of the handler.

ajmaddaford (changes requested):

app/publisher/publisher.py
MebinAbraham: Added a log message on error.

app/settings.py
```diff
@@ -49,6 +49,9 @@ def get_env_or_fail(key):
 EQ_SECRETS_FILE = os.getenv("EQ_SECRETS_FILE", "secrets.yml")
 EQ_KEYS_FILE = os.getenv("EQ_KEYS_FILE", "keys.yml")
 
+EQ_PUBLISHER_BACKEND = os.getenv("EQ_PUBLISHER_BACKEND")
+EQ_FULFILMENT_TOPIC_ID = os.getenv("EQ_FULFILMENT_TOPIC_ID")
```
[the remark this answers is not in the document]
MebinAbraham: Done

app/publisher/publisher.py
```diff
@@ -0,0 +1,46 @@
...
+logger = get_logger(__name__)
+
+
+class Publisher(ABC):
+    @abstractmethod
+    def publish(self, topic_id, message):
+        pass  # pragma: no cover
+
+
+class PubSub(Publisher):
```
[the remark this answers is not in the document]
MebinAbraham: Done

app/views/handlers/individual_response.py
MebinAbraham: 🤦 Good spot. Fixed.

app/views/handlers/individual_response.py
MebinAbraham: Added a new class to fetch the payload.

app/helpers/uuid_helper.py
```diff
@@ -0,0 +1,9 @@
+from uuid import UUID
```
JamesGardiner, line 1: from being in the helpers module we know this is a helper, so no need to add the `_helper` suffix to the module name

app/publisher/publication_failed.py
```diff
@@ -0,0 +1,2 @@
+class PublicationFailed(Exception):
```
JamesGardiner, line 1: This should be in an `exceptions` module: `app/publisher/exceptions.py`

app/helpers/uuid_helper.py
MebinAbraham: I agree 100% but I kept it to be consistent our existing ones.

app/publisher/publisher.py
```diff
@@ -0,0 +1,51 @@
...
+
+class PubSubPublisher(Publisher):
+    def __init__(self):
+        self._client = PublisherClient()
+        _, self._project_id = google.auth.default()
+
+    def _publish(self, topic_id, message):
+        logger.info("publishing message", topic_id=topic_id)
+        topic_path = self._client.topic_path(self._project_id, topic_id)
+        publish_future: Future = self._client.publish(topic_path, message)
```
JamesGardiner, line 27: This var should be something like `response`, as it doesn't matter it's a future as the future will get resolved

app/publisher/publisher.py
```diff
@@ -0,0 +1,51 @@
...
+        return publish_future
+
+    def publish(self, topic_id, message: bytes):
+        publish_future = self._publish(topic_id, message)
```
JamesGardiner, line 31: Same again, `publish_future` should be something like `response`

app/publisher/publisher.py
MebinAbraham: It's not really a response, I would say it becomes a response when you resolve it.

app/publisher/publisher.py
```diff
@@ -0,0 +1,51 @@
...
+                topic_id=topic_id,
+                message_id=message_id,
+            )
+        except Exception as ex:  # pylint:disable=broad-except
+            logger.error(
+                "message publication failed",
+                topic_id=topic_id,
+                exc_info=ex,
+            )
+            raise PublicationFailed(ex)
```
JamesGardiner, line 46: `raise PublicationFailed from ex` will explicitly set the `__cause__`

app/publisher/publisher.py
JamesGardiner: Google disagrees https://googleapis.dev/python/pubsub/latest/publisher/api/client.html 😆

app/setup.py
```diff
@@ -324,6 +327,17 @@ def setup_submitter(application):
...
 
+def setup_publisher(application):
+    if application.config["EQ_PUBLISHER_BACKEND"] == "pubsub":
+        application.eq["publisher"] = PubSubPublisher()
+
+    elif application.config["EQ_PUBLISHER_BACKEND"] == "log":
+        application.eq["publisher"] = LogPublisher()
+
+    else:
+        raise Exception("Unknown EQ_PUBLISHER_BACKEND")
```
JamesGardiner, line 338: Why the base Exception?

app/publisher/publisher.py
MebinAbraham: As discussed, changed to keep it inline with official library,

app/views/handlers/individual_response.py
```diff
@@ -842,3 +867,59 @@ def handle_post(self):
...
+        return fulfilment_codes[self._fulfilment_type][region_code]
+
+    @cached_property
+    def payload(self) -> bytes:
+        message = {
+            "event": {
+                "type": "FULFILMENT_REQUESTED",
+                "source": "QUESTIONNAIRE_RUNNER",
+                "channel": "EQ",
+                "dateTime": datetime.now(tz=tzutc()).isoformat(),
```
JamesGardiner, line 912: Should this definitely be UTC? In the payload description there is nothing to denote it (i.e. it doesn't end in `Z` which would indicate UTC timezone

app/setup.py
MebinAbraham: Just because that is what we currently do, but ya it could be more specific but I would then suggest updating all existing ones.

app/views/handlers/individual_response.py
MebinAbraham: As discussed, I confirmed this with Mark. Python by default won't show 'Z' for Zulu time, it would show the offset which is fine from RM's perspective.

app/publisher/publication_failed.py
MebinAbraham: Done.

app/publisher/publisher.py
MebinAbraham: As discussed, this was not needed as the exceptions is chained correct by default in our case.

LJBabbage (approved):

app/setup.py
JamesGardiner: When you say all existing ones, do you mean across Runner?

app/views/handlers/individual_response.py
JamesGardiner: Sorry, should have been more specific, the string in the payload didn't contain any denotion of timezone or offset. If this is known by RM then that's fine, but ideally we should be explicit

app/publisher/publisher.py
```diff
@@ -0,0 +1,51 @@
...
+        try:
+            # Resolve the future
+            message_id = response.result()
+            logger.info(  # pragma: no cover
+                "message published successfully",
+                topic_id=topic_id,
+                message_id=message_id,
+            )
+        except Exception as ex:  # pylint:disable=broad-except
+            logger.error(
```
JamesGardiner, line 41:
```suggestion
            logger.exception(
```

logging within exception blocks should use the `exception` method, so that the exc_info is explicitly set

app/publisher/publisher.py
```diff
@@ -0,0 +1,51 @@
...
+                "message publication failed",
+                topic_id=topic_id,
+                exc_info=ex,
```
JamesGardiner, line 44:
```suggestion
```

app/setup.py
MebinAbraham: Well certainly within this file, as our pattern currently is to raise generic exception when env vars are messing/not as expected..

JamesGardiner (approved):

app/views/handlers/individual_response.py
```diff
@@ -826,19 +848,78 @@ def handle_get(self):
...
+        self._fulfilment_type = "sms" if self._mobile_number else "postal"
+
+    def _get_individual_case_id_mapping(self) -> Mapping:
+        return (
+            {}
+            if self._metadata.get("case_type") == "SPG"
+            else {"individualCaseId": str(uuid4())}
+        )
+
+    def _get_mobile_number_mapping(self) -> Mapping:
```
ajmaddaford, line 885: This method is called to populate the `contact` property so would make more sense as:
```suggestion
    def _get_contact_mapping(self) -> Mapping:
```

app/views/handlers/individual_response.py
```diff
@@ -826,19 +848,78 @@ def handle_get(self):
...
+                GB_ENG_REGION_CODE: "P_UAC_UACIP1",
+                GB_WLS_REGION_CODE: "P_UAC_UACIP2B",
+                GB_NIR_REGION_CODE: "P_UAC_UACIP4",
+            },
+        }
+
+        region_code = self._metadata["region_code"]
+        return fulfilment_codes[self._fulfilment_type][region_code]
+
+    @cached_property
```
ajmaddaford, line 905: I'm not sure this should be cached. Given there is a timestamp in it I would expect it to be generated each time I called it.

app/views/handlers/individual_response.py
```diff
@@ -130,6 +139,18 @@ def form(self):
     def get_context(self):
         return build_question_context(self.rendered_block, self.form)
 
+    def _publish_fulfilment_request(self, serialized_mobile_number=None):
+        topic_id = current_app.config["EQ_FULFILMENT_TOPIC_ID"]
+        mobile_number = (
+            URLParamSerializer().loads(serialized_mobile_number)
+            if serialized_mobile_number
+            else None
+        )
```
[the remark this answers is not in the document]
MebinAbraham: Good point, done.

app/views/handlers/individual_response.py
MebinAbraham: Updated all ocurrences.

app/views/handlers/individual_response.py
MebinAbraham: We only call this once anyway, so removed.

ajmaddaford (approved):

PR #277, opened 2020-09-08
Maintain name duplication state

iwootten:

Implements capture of household name duplicates. Adds a new key "same_name_items" to the ListStore which holds list_item_ids for all those list items that have a duplicate. Most of the work is done in `questionnaire_store_updater.update_same_name_items` which updates the duplicates following an update to a list.

Depends on https://github.com/ONSdigital/eq-questionnaire-validator/pull/63 to add same_name_answer_ids key to the schema.

### How to review 

Use the test_same_name_items.json to observe the behaviour, using /dump/debug to see how `same_name_items` changes as the list store items are updated.

### Checklist


iwootten:
[[T] Maintain duplicate first/last name list state (L)](https://trello.com/c/gbS9YV0C/4111-t-maintain-duplicate-first-last-name-list-state-l)

LJBabbage (changes requested):
Works as expected, a couple of suggestions. One other point, which we didn't discuss, and it is an edge case, but how do we want to represent 2 different names duplication, eg. John A Smith, John B Smith, Martha A Smith and Martha B Smith. A the moment this produces one list with.all 4 items in

app/data_model/list_store.py
```diff
@@ -20,10 +20,12 @@ def __init__(
         name: str,
         items: Optional[List] = None,
         primary_person: Optional[str] = None,
+        same_name_items: Optional[List] = None,
```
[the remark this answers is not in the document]
iwootten: That's true, though it's referred to as same_name_items everywhere else and there are currently no other use cases for duplicates. I think therefore keeping it named the same here makes more sense.

iwootten:
> Works as expected, a couple of suggestions. One other point, which we didn't discuss, and it is an edge case, but how do we want to represent 2 different names duplicated, eg. John A Smith, John B Smith, Martha A Smith and Martha B Smith. At the moment this produces one list with all 4 items in

We did discuss this in the tech session this week - if the name appears in the list, we show the middle name. There isn't a need to identify which other item it is duplicated with, simply that it appears in the same_name_items list.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -158,19 +158,51 @@ def remove_list_item_and_answers(self, list_name: str, list_item_id: str):
...
+            answers = self._questionnaire_store.answer_store.get_answers_by_answer_id(
+                answer_ids=same_name_answer_ids, list_item_id=current_list_item_id
+            )
+            current_names = [str(answer.value).lower() for answer in answers if answer]
+            current_list_item_name = " ".join(current_names)
+
+            if current_list_item_name in people_names:
+                people_names[current_list_item_name] += [current_list_item_id]
+            else:
+                people_names[current_list_item_name] = [current_list_item_id]
```
MebinAbraham, line 193: The If else could be avoided by using a `defaultdict`

tests/app/questionnaire/test_questionnaire_store_updater.py
```diff
@@ -297,3 +301,39 @@ def test_update_relationship_question_completeness_no_relationship_collectors(se
...
+                {"answer_id": "first-name", "value": "Joe", "list_item_id": "abcdef"},
+                {"answer_id": "last-name", "value": "Bloggs", "list_item_id": "abcdef"},
+                {"answer_id": "first-name", "value": "Joe", "list_item_id": "ghijkl"},
+                {"answer_id": "last-name", "value": "Bloggs", "list_item_id": "ghijkl"},
+                {
+                    "answer_id": "first-name",
+                    "value": "Martha",
+                    "list_item_id": "xyzabc",
+                },
+                {"answer_id": "last-name", "value": "Bloggs", "list_item_id": "xyzabc"},
```
MebinAbraham, line 317: Would be good to add some middle names.

MebinAbraham (changes requested):

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -158,19 +158,51 @@ def remove_list_item_and_answers(self, list_name: str, list_item_id: str):
...
+        if not same_name_answer_ids:
+            return
+
+        list_model = self._questionnaire_store.list_store[list_name]
+
+        for current_list_item_id in list_model:
+            answers = self._questionnaire_store.answer_store.get_answers_by_answer_id(
+                answer_ids=same_name_answer_ids, list_item_id=current_list_item_id
+            )
+            current_names = [str(answer.value).lower() for answer in answers if answer]
```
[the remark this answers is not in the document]
iwootten: Answer values can be of any one of a number of types right?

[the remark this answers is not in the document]
iwootten: Unaware this existed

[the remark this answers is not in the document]
MebinAbraham: Answer could be diff type but this imp is so specific to name where they wont be non strings?

app/questionnaire/questionnaire_store_updater.py
iwootten: This and the comment above were on outdated commits, because I failed to commit earlier in the day.

LJBabbage (approved):

app/views/handlers/list_add_question.py
```diff
@@ -13,4 +13,8 @@ def handle_post(self):
         )
         self.questionnaire_store_updater.update_answers(self.form.data, list_item_id)
 
+        self.questionnaire_store_updater.update_same_name_items(
+            self.parent_block["for_list"], self._get_same_name_answer_ids()
+        )
+
```
ajmaddaford, line 19: Could the updating of the same name items be done in the `handle_post` of `list_action.py` rather than in each of the add/edit/remove actions?

app/views/handlers/primary_person_list_collector.py
```diff
@@ -31,11 +32,20 @@ def handle_post(self):
...
+
             self.evaluate_and_update_section_status_on_list_change(list_name)
             self.questionnaire_store_updater.save()
         else:
             self.questionnaire_store_updater.remove_primary_person(list_name)
 
+            if same_name_answer_ids:
+                self.questionnaire_store_updater.update_same_name_items(
+                    list_name, same_name_answer_ids
+                )
```
ajmaddaford, line 48: Any reason to check for `same_name_answer_ids` here, but not above?

app/views/handlers/primary_person_list_collector.py
```diff
@@ -31,11 +32,20 @@ def handle_post(self):
             self._primary_person_id = (
                 self.questionnaire_store_updater.add_primary_person(list_name)
             )
+
+            self.questionnaire_store_updater.update_same_name_items(
+                list_name, same_name_answer_ids
+            )
```
ajmaddaford, line 38: I don't think this is needed here as the name is not entered until the next page, so we wouldn't know if it was a duplicate or not.

app/views/handlers/primary_person_list_collector.py
iwootten: This was a hangover from get_same_name_answer_ids originally defaulting to False.

tests/integration/questionnaire/test_questionnaire_list_collector_primary_person.py
```diff
@@ -6,8 +6,13 @@
 
 
 class TestQuestionnaireListCollector(IntegrationTestCase):
-    def get_link(self, rowIndex, text):
-        selector = f"tbody:nth-child({rowIndex}) td:last-child a"
+    def get_previous_link(self):
+        selector = "#top-previous"
+        selected = self.getHtmlSoup().select(selector)
+        return selected[0].get("href")
```
ajmaddaford, line 12: I can't see this being used (also, there is a `previous` method on `IntegrationTestCase`).

tests/integration/questionnaire/test_questionnaire_same_name_items.py
```diff
@@ -0,0 +1,113 @@
...
+class TestQuestionnaireSameNameItems(IntegrationTestCase):
+    def get_link(self, row_index, text):
+        selector = f"tbody:nth-child({row_index}) td:last-child a"
+        selected = self.getHtmlSoup().select(selector)
+
+        filtered = [html for html in selected if text in html.get_text()]
+
+        return filtered[0].get("href")
+
+    def test_dump_same_name_items(self):
```
ajmaddaford, line 15: `_dump` should be dropped from all method names.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -158,19 +158,43 @@ def remove_list_item_and_answers(self, list_name: str, list_item_id: str):
...
+
+        for current_list_item_id in list_model:
+            answers = self._questionnaire_store.answer_store.get_answers_by_answer_id(
+                answer_ids=same_name_answer_ids, list_item_id=current_list_item_id
+            )
+            current_names = [answer.value.casefold() for answer in answers if answer]
+            current_list_item_name = " ".join(current_names)
+
+            if current_list_item_name in people_names:
+                matching_list_item_id = people_names[current_list_item_name]
```
MebinAbraham, line 191: Could use a walrus operator:
```py
            if matching_list_item_id := people_names.get(current_list_item_name):
                same_name_items |= {current_list_item_id, matching_list_item_id}
```

app/views/handlers/list_action.py
```diff
@@ -55,6 +55,10 @@ def get_next_location_url(self):
         return self.parent_location.url()
 
     def handle_post(self):
+        self.questionnaire_store_updater.update_same_name_items(
+            self.parent_block["for_list"],
+            self.parent_block.get("same_name_answer_ids", []),
```
MebinAbraham, line 60: The default of `[]` isn't required given `update_same_name_items` exits early if `same_name_answer_ids` is falsey and doesn't try iterate by default. The type hint should be updated 
```py
same_name_answer_ids: Optional[List[str]]
```
Applies to all occurrences.

app/questionnaire/questionnaire_store_updater.py
iwootten: Hipster

app/views/handlers/primary_person_question.py
```diff
@@ -34,7 +39,11 @@ def get_next_location_url(self):
         )
 
     def handle_post(self):
+        same_name_answer_ids = self.parent_block.get("same_name_answer_ids", [])
```
MebinAbraham, line 42:
```suggestion
        same_name_answer_ids = self.parent_block.get("same_name_answer_ids")
```

MebinAbraham (approved):

PR #281, opened 2020-09-11
Remove census wording from CCS thank you page

LJBabbage:
CCS on submission should not mention the census, it said it earlier but got away with it.

### How to review 
Using test_thank_you_ccs.json or the ccs schema itself, confirm on the thank you page the text has changed, and says survey rather the census

### Checklist


berroar (commented):
Should the naming for the test schema and test be more generic? For example using something like `test_non_census_survey` rather than `test_ccs`?

LJBabbage:
> Should the naming for the test schema and test be more generic? For example using something like `test_non_census_survey` rather than `test_ccs`?

So the code I am trying to exercise is

```
        if "ccs" not in schema_name:
            for census_type in self.CENSUS_TYPE_MAPPINGS:
                if census_type in schema_name:
                    census_type_code = self.CENSUS_TYPE_MAPPINGS[census_type]
                    break
```
Because I have referenced ccs directly, I would try to keep my test name to the same

berroar (approved):

app/views/handlers/thank_you.py
```diff
@@ -48,11 +48,14 @@ def get_context(self):
...
-        for census_type in self.CENSUS_TYPE_MAPPINGS:
-            if census_type in self.session_data.schema_name:
-                census_type_code = self.CENSUS_TYPE_MAPPINGS[census_type]
-                break
+
+        if "ccs" not in schema_name:
+            for census_type in self.CENSUS_TYPE_MAPPINGS:
+                if census_type in schema_name:
+                    census_type_code = self.CENSUS_TYPE_MAPPINGS[census_type]
+                    break
```
ajmaddaford, line 58: The logic here could be kept as it was if `CENSUS_TYPE_MAPPINGS` was updated to be:

```python
    CENSUS_TYPE_MAPPINGS = {
        "census_household": "HH",
        "census_communal_establishment": "CE",
        "census_individual": "IR",
    }
```
Also, although the work was done in a previous pr, the census type code should be done in `build_census_thank_you_context`.

test_schemas/en/test_thank_you_no_census_type_map.json
```diff
@@ -0,0 +1,66 @@
+{
```
LJBabbage, line 1: After the update I am half tempted to get rid of this schema, however we do need a test, and if I use a different schema in the test it potential conceals what is happening

ajmaddaford (approved):

PR #286, opened 2020-09-16
Override _resolve_custom_page_title_vars for add blocks

JamesGardiner:
List add blocks cannot use the list item id, as it does not exist. Instead, this PR overrides the _resolv_custom_page_title_vars method to return the length of the list defined by the current Location object, plus one.

### How to review 
Test the custom page titles for add blocks render correctly.

### DIFF
--- a/app/views/handlers/list_add_question.py
+++ b/app/views/handlers/list_add_question.py
@@ -1,3 +1,5 @@
+from typing import MutableMapping
+
 from app.views.handlers.list_action import ListAction
 
 
@@ -13,3 +15,14 @@ class ListAddQuestion(ListAction):
         )
         self.questionnaire_store_updater.update_answers(self.form.data, list_item_id)
         return super().handle_post()
+
+    def _resolve_custom_page_title_vars(self) -> MutableMapping:
+        # For list add blocks, no list item id is yet available. Any custom `page_title`
+        # `list_item_position` string formatter is instead resolved to the length of
+        # the list identified by the `list_name` attribute of the current Location
+        # object.
+        list_length = len(
+            self._questionnaire_store.list_store[self._current_location.list_name]
+        )
+
+        return {"list_item_position": list_length + 1}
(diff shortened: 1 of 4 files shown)


berroar (approved):
Tested the change, the page titles are being rendered correctly for the add person pages

tests/functional/spec/custom-page-titles.spec.js
```diff
@@ -20,12 +20,20 @@ describe("Feature: Custom Page Titles", () => {
...
+      expect(expectedPageTitle).to.equal("Person 1");
+
+      $(ListCollectorAddPage.firstName()).setValue("Marcus");
+      $(ListCollectorAddPage.lastName()).setValue("Twin");
+      $(ListCollectorAddPage.submit()).click();
+      $(ListCollectorPage.yes()).click();
+      $(ListCollectorPage.submit()).click();
+      expectedPageTitle = browser.getTitle();
+      expect(expectedPageTitle).to.equal("Person 2");
     });
```
MebinAbraham, line 37: I think we should update the edit and remove test to also show this feature working?

app/views/handlers/list_add_question.py
```diff
@@ -13,3 +15,14 @@ def handle_post(self):
         )
         self.questionnaire_store_updater.update_answers(self.form.data, list_item_id)
         return super().handle_post()
+
+    def _resolve_custom_page_title_vars(self) -> MutableMapping:
+        # For list add blocks, no list item id is yet available. Any custom `page_title`
+        # `list_item_position` string formatter is instead resolved to the length of
+        # the list identified by the `list_name` attribute of the current Location
+        # object.
```
ajmaddaford, line 23: Some of this comment is redundant (it's clear from the code how we get the current list name), and some is incorrect (it should be the list length plus one). Maybe something like:
```suggestion
        # For list add blocks, no list item id is yet available. Instead, we resolve 
        # `list_item_position` to the position in the list it would be if added.
```

tests/functional/spec/custom-page-titles.spec.js
JamesGardiner: Done in d13c06cf

MebinAbraham (approved):

PR #288, opened 2020-09-17
Update design system to version 24.0.3

ons-eq-team:
Design system release: https://github.com/ONSdigital/design-system/releases/tag/24.0.3

Removes the use of autosuggest data after a fix to the design system no longer requiring it. This was causing tests to intermittently fail.


MebinAbraham (approved):
Address works as expected,

JamesGardiner (approved):

PR #290, opened 2020-09-17
Add section repeating page title

LJBabbage:
This PR adds page_title into repeating sections

### How to review 
Using test_custom_page_titles.json, confirm that the page title is correct when in a repeat section which has a page_title value. 

N.B the suffix Census 2021 is currently missing if the block has a page_title, another card will be raised to fix this

### Checklist


app/views/contexts/section_summary_context.py
```diff
@@ -40,11 +40,28 @@ def __call__(
...
+                if section_repeating_page_title
+                else custom_page_title
+            )
             return self._resolve_custom_page_title(custom_page_title, current_location)
 
+        title_for_location = (
+            f"{title_for_location}: {section_repeating_page_title}"
+            if section_repeating_page_title
+            else title_for_location
+        )
```
ajmaddaford, line 64: It doesn't matter if there is a custom page title or not, the `section_repeating_page_title` should have a `{list_item_position}` placeholder that needs resolving.

ajmaddaford (changes requested):
There is quite a lot of duplication. I think we need to consider how we can have one common method to append the section repeating title if it exists.

app/views/contexts/section_summary_context.py
LJBabbage: Would that be the case if there was not a repeat in a section?

LJBabbage:
This PR https://github.com/ONSdigital/eq-questionnaire-runner/pull/293 will have an affect on this PR. Because of refactoring it is not possible to have a custom page title do anything different to a normal page title when dealing with the survey title suffix, this means I can't fail but to break tests. To avoid writing code which will be removed looking at the dependent PR, I have removed survey title as the prefix and temporarily fixed the tests. Because of that this card should not be merged

app/views/handlers/block.py
```diff
@@ -78,6 +79,41 @@ def router(self):
...
+            }
+
+        content_page_title = transformed_block_page_title or self._get_content_title(
+            transformed_block
+        )
+        self._set_page_title(content_page_title)
+        rendered_question = self.placeholder_renderer.render(
+            transformed_block, self._current_location.list_item_id
+        )
+        return rendered_question
```
ajmaddaford, line 115: I don't understand why the question and content specific code has been moved here. Could that code stay in the respective handlers, and call the common `_set_page_title` method?

LJBabbage: Can do, just trying to keep the transform in the block

test_schemas/en/test_custom_page_titles.json
```diff
@@ -811,6 +811,23 @@
                                 "type": "General"
                             },
                             "type": "Question"
+                        },
+                        {
+                            "type": "Question",
+                            "id": "date-of-birth",
+                            "question": {
+                                "id": "date-pipe-question",
+                                "title": "What is your date of birth",
```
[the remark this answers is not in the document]
LJBabbage: Surely that's a bad title, and not someone we should try to fix in the code.

app/views/handlers/block.py
```diff
@@ -125,17 +125,34 @@ def _set_started_at_metadata(self):
 
     def _get_safe_page_title(self, page_title):
         page_title = self._schema.get_single_string_value(page_title)
-        title = self._schema.json["title"]
 
-        return safe_content(f"{page_title} - {title}")
+        return safe_content(f"{page_title}")
```
ajmaddaford:
```suggestion
        return page_title
```

app/views/handlers/block.py
```diff
@@ -125,17 +125,34 @@ def _set_started_at_metadata(self):
...
+        list_item_position = self._questionnaire_store.list_store.list_item_position(
+            self.current_location.list_name, self.current_location.list_item_id
+        )
+        return {"list_item_position": list_item_position}
 
-        return {}
+    def _get_content_title(self, transformed_block):
+        content = transformed_block.get("content")
+        if content:
+            return self._get_safe_page_title(content["title"])
```
ajmaddaford, line 139: This logic should be in the content block handler.

app/views/handlers/block.py
```diff
@@ -125,16 +125,33 @@ def _set_started_at_metadata(self):
...
+        )
+        if section_repeating_page_title:
+            self.page_title = f"{self.page_title}: {section_repeating_page_title}"
+
+        if self.page_title and (
+            self._current_location.list_item_id
+            or self.block["type"] == "ListAddQuestion"
+        ):
+            page_title_vars = self._resolve_custom_page_title_vars()
+            self.page_title = self.page_title.format(**page_title_vars)
```
ajmaddaford, line 157: - `self.page_title` should be set once at the end of the method
- The method should always be called with a title, so we don't need to check for `page_title` in the second conditional

```suggestion
    def _set_page_title(self, page_title):
        section_repeating_page_title = (
            self._schema.get_repeating_page_title_for_section(
                self._current_location.section_id
            )
        )
        if section_repeating_page_title:
            page_title = f"{page_title}: {section_repeating_page_title}"

        if (
            self._current_location.list_item_id
            or self.block["type"] == "ListAddQuestion"
        ):
            page_title_vars = self._resolve_custom_page_title_vars()
            page_title = page_title.format(**page_title_vars)

        self.page_title = page_title
```

app/views/handlers/content.py
```diff
@@ -8,42 +8,31 @@
...
-
-        return self.placeholder_renderer.render(
+        content_page_title = transformed_block.get(
+            "page_title"
+        ) or self._get_content_title(transformed_block)
+        self._set_page_title(content_page_title)
+        rendered_question = self.placeholder_renderer.render(
             transformed_block, self._current_location.list_item_id
         )
+        return rendered_question
```
ajmaddaford, line 27: The `rendered_question` variable name is misleading and unnecessary.

app/views/handlers/question.py
```diff
@@ -54,6 +50,29 @@ def questionnaire_store_updater(self):
...
+            self._questionnaire_store.metadata,
+            self._questionnaire_store.answer_store,
+            self._questionnaire_store.list_store,
+            self._current_location,
+        )
+        question_page_title = transformed_block.get(
+            "page_title"
+        ) or self._get_safe_page_title(transformed_block["question"]["title"])
+
+        self._set_page_title(question_page_title)
```
ajmaddaford, line 67: As this is in the question handler, `question_page_title` can just be `page_title`.
```suggestion
        page_title = transformed_block.get(
            "page_title"
        ) or self._get_safe_page_title(transformed_block["question"]["title"])

        self._set_page_title(page_title)
```

test_schemas/en/test_custom_page_titles.json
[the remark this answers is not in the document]
ajmaddaford: If a question title (rather than page title) is used it will end in `?`.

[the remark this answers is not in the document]
LJBabbage: True, but I can do nothing about this

test_schemas/en/test_custom_page_titles.json
```diff
@@ -814,11 +814,10 @@
                         },
                         {
                             "type": "Question",
-                            "id": "date-of-birth-block",
-                            "page_title": "Custom question page title",
+                            "id": "date-of-birth",
                             "question": {
                                 "id": "date-pipe-question",
-                                "title": "What is your date of birth?",
+                                "title": "Date of birth",
```
ajmaddaford, line 820: I understand this was changed to hide the problem of the question mark in page titles, but this should be reverted - the question title should always be a question. We could add a `page_title` property here as the question text in the page title should be covered by other tests.

ajmaddaford (approved):

test_schemas/en/test_custom_page_titles.json
LJBabbage: No problem will revert, but I deliberately left it without a page_title, thus giving me a repeating section with a page_title, but a block without

JamesGardiner:
@LJBabbage can the `dont merge` label be removed now that https://github.com/ONSdigital/eq-questionnaire-runner/pull/293 has been merged?

app/questionnaire/questionnaire_schema.py
```diff
@@ -224,6 +224,9 @@ def get_repeating_list_for_section(self, section_id):
     def get_repeating_title_for_section(self, section_id):
         return self._sections_by_id.get(section_id).get("repeat", {}).get("title")
 
+    def get_repeating_page_title_for_section(self, section_id):
+        return self._sections_by_id.get(section_id).get("repeat", {}).get("page_title")
```
JamesGardiner, line 228: We should have a `get_repeat_for_section` method given the `self._sections_by_id.get(section_id).get("repeat", {})` is used in three places

JamesGardiner (changes requested):

iwootten:
> @LJBabbage can the `dont merge` label be removed now that #293 has been merged?

FWIW, I didn't even bother reviewing this because that label was there.

JamesGardiner (approved):

PR #295, opened 2020-09-18
Add transforms to support users with matching names

iwootten:

Adds a number of transforms to placeholders for displaying middle names when first and last name match.

Changes to the `test_list_collector_same_name_items` now demonstrate both transforms at work. Full names for all householders will be displayed on the summary page if any householder has duplicate names to another, the edit page will show middle names only if that person shares their first/last name with someone else in the household.

NB. Originally I had used a new source of "list_item_id" to return the list_Item_id, but this seemed very specific and after discussion felt it better we add a "location" source. The renderer already takes both a list_item_id and location and we actually return that of list_item_id in the new transforms - to use location requires significant change to the existing context design (as they don't always have the property) which didn't seem required for this feature. If we ever have a need for other location properties in transforms, then we can perhaps look at refactoring contexts to support them.

### How to review 

### DIFF
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=resolve-name-display
+tag=latest
 docker pull onsdigital/eq-questionnaire-validator:$tag
 docker run -d -p 5001:5000 "onsdigital/eq-questionnaire-validator:$tag"
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=resolve-name-display
 docker pull onsdigital/eq-questionnaire-validator:$tag
 docker run -d -p 5001:5000 "onsdigital/eq-questionnaire-validator:$tag"
--- a/test_schemas/en/test_list_collector_same_name_items.json
+++ b/test_schemas/en/test_list_collector_same_name_items.json
@@ -315,14 +315,36 @@
                                             "placeholder": "person_name",
                                             "transforms": [
                                                 {
+                                                    "transform": "name_is_duplicate",
                                                     "arguments": {
-                                                        "delimiter": " ",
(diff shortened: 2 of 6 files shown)


iwootten:
[[T] Resolve name to be displayed when duplicate names exist (M)](https://trello.com/c/cIBH9hWa/4112-t-resolve-name-to-be-displayed-when-duplicate-names-exist-m)

ajmaddaford (changes requested):

test_schemas/en/test_list_collector_same_name_items.json
```diff
@@ -315,14 +315,36 @@
...
-                                                            "identifier": ["first-name", "last-name"],
-                                                            "source": "answers"
+                                                        "same_name_item_ids": {
+                                                            "source": "list",
+                                                            "id_selector": "same_name_items",
+                                                            "identifier": "people"
+                                                        },
+                                                        "list_item_id": {
+                                                            "source": "location",
+                                                            "identifier": "list_item_id"
```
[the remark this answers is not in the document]
iwootten: I think list will be reserved, so we'd need to use something else `list_to_check`?

test_schemas/en/test_list_collector_same_name_items.json
```diff
@@ -315,14 +315,36 @@
...
                                                         }
-                                                    },
-                                                    "transform": "concatenate_list"
+                                                    }
+                                                },
+                                                {
+                                                    "transform": "format_name",
+                                                    "arguments": {
+                                                        "include_middle_name": { "source": "previous_transform" },
```
ajmaddaford, line 334: `include_middle_name` should be `include_middle_names` to align with the `middle_names` property.

ajmaddaford (approved):

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=resolve-name-display
```

MebinAbraham (approved):

PR #299, opened 2020-09-22
Customise final summary text

berroar:
Final summary page needs the following to be customisable using the submission definition in the schemas: `Title`, `Warning`, `Declaration`, `Submit button`

### How to review 
You can use the `test_summary_with_submission_text.json` to test this change

### Checklist


berroar:
[Make Final summary customisable (S)](https://trello.com/c/CffmGdMl/4146-make-final-summary-customisable-s)

app/views/handlers/summary.py
```diff
@@ -15,4 +17,20 @@ def get_context(self):
         block = self._schema.get_block(self._current_location.block_id)
         collapsible = block.get("collapsible", False)
         context = questionnaire_summary_context(collapsible)
+
+        submission_schema = self._schema.get_submission()
+
+        if submission_schema:
```
JamesGardiner, line 23:
```suggestion
        if submission_schema := self._schema.get_submission():
```

app/views/handlers/summary.py
```diff
@@ -15,4 +17,20 @@ def get_context(self):
...
+            context["title"] = submission_schema.get("title")
+            context["submit_button"] = submission_schema.get("button")
+            context["guidance"] = submission_schema.get("guidance")
+            context["warning"] = submission_schema.get("warning")
```
JamesGardiner, line 27: Are you simply merging the whole of the `submission_schema` Dict? If so, you could just merge the two...

```Python
context = {**context, **submission_schema}
```

JamesGardiner (changes requested):

LJBabbage (changes requested):
A few minor comments,  we should also extend test_summary_context.py to cover these new scenarios.

app/views/handlers/summary.py
berroar: Not sure if this will work now after I've made the changes from Lewis' comments

tests/integration/questionnaire/test_questionnaire_is_skipping_question.py
```diff
@@ -25,7 +25,6 @@ def test_final_summary_not_available_after_invalidating_section(self):
         self.assertInBody("Were you forced to complete section 2?")
         self.post()
 
-        self.assertInBody("Check your answers before submitting")
```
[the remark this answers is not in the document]
berroar: That bit of text was no longer in the template, so I have now replaced it with the bit of guidance that is

berroar:
> A few minor comments, we should also extend test_summary_context.py to cover these new scenarios.

@LJBabbage  Found it a difficult to add a test for this, maybe because the code I added is in the handler, not in the contexts? Not sure we usually add tests for the handlers

app/views/handlers/summary.py
JamesGardiner: I think you are correct

LJBabbage:
> @LJBabbage Found it a difficult to add a test for this, maybe because the code I added is in the handler, not in the contexts? Not sure we usually add tests for the handlers

Sorry Rhys I didn't notice until I have just looked again, you have put the extra code in the handler, when it should be in the context (QuestionnaireSummaryContext) This then makes it easier to test as well

app/views/handlers/summary.py
```diff
@@ -15,4 +17,24 @@ def get_context(self):
         block = self._schema.get_block(self._current_location.block_id)
         collapsible = block.get("collapsible", False)
         context = questionnaire_summary_context(collapsible)
+
+        if submission_schema := self._schema.get_submission():
```
LJBabbage, line 21: This should be in QuestionnaireSummaryContext

app/views/handlers/summary.py
```diff
@@ -15,4 +17,24 @@ def get_context(self):
...
+                "Please submit this survey to complete it"
+            )
+            context["warning"] = submission_schema.get("warning") or None
+        else:
+            context["title"] = lazy_gettext("Check your answers and submit")
+            context["submit_button"] = lazy_gettext("Submit answers")
+            context["guidance"] = lazy_gettext(
+                "Please submit this survey to complete it"
+            )
+            context["warning"] = None
```
LJBabbage, line 38: This should be in QuestionnaireSummaryContext

app/views/handlers/summary.py
LJBabbage, line 21: You don't need the walrus, it can just be
```
submission_schema := self._schema.get_submission():
```
and then dump the else

berroar: Thanks, I've moved the code now and added a test 👍

JamesGardiner (approved):

tests/app/views/contexts/test_summary_context.py
```diff
@@ -82,14 +82,33 @@ def test_build_summary_rendering_context(self):
             self.language,
             self.schema,
             self.answer_store,
-            self.progress_store,
             self.list_store,
+            self.progress_store,
```
LJBabbage, line 86: Not sure why the position change

tests/functional/spec/features/summary/summary_screen.spec.js
```diff
@@ -3,6 +3,8 @@ import RadioPage from "../../../generated_pages/summary/radio.page.js";
 import SummaryPage from "../../../generated_pages/summary/summary.page.js";
 import TestNumberPage from "../../../generated_pages/summary/test-number-block.page.js";
 
+import DessertBlockPageSubmission from "../../../generated_pages/summary_with_submission_text/dessert-block.page.js";
```
LJBabbage, line 6: We generally keep these together and in alphabetical order

templates/summary.html
```diff
@@ -8,13 +8,20 @@
...
+          }) %}
+          <p data-qa="warning">{{ content.warning }}</p>
+        {% endcall %}
+    {% endif %}
 
     {% call onsPanel({
       "classes": "print__hidden u-mb-l"
     }) %}
-      <strong>{{ _("Please submit this survey to complete it") }}</strong>
-      <p>{{ _("Check your answers before submitting") }}</p>
```
LJBabbage, line 17: I couldn't see on the card this being removed, was it requested by Donna?

LJBabbage (changes requested):
Only point that needs clarifying  "Check your answers before submitting"  to be removed?

tests/app/views/contexts/test_summary_context.py
berroar: They were the wrong way around, now matches the `QuestionnaireSummaryContext` order in the class

templates/summary.html
berroar: @LJBabbage confirmed with Donna that this text is no longer required

LJBabbage (approved):

PR #301, opened 2020-09-23
Add Address selector to placeholder

LJBabbage:
Placeholders now have selector elements to find fields in Addresses  

N.B I do have reservations over the proposed solution names as explained on the validator PR

### How to review 
Using test_address make sure on entering a full address on the first page that only the first line appears on the confirmation page


scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=add-selector-to-placeholder-address-type
```

test_schemas/en/test_address.json
```diff
@@ -60,6 +60,44 @@
                                     }
                                 ]
                             }
+                        },
+                        {
+                            "type": "Question",
+                            "id": "address-confirmation",
+                            "question": {
+                                "type": "General",
```
MebinAbraham, line 68: Could be a confirmation question

MebinAbraham (commented):
Feature works as expected, will approve once validator has been addressed.

app/questionnaire/placeholder_parser.py
```diff
@@ -80,7 +80,12 @@ def _resolve_answer_value(self, value_source):
                 self._lookup_answer(each_identifier, list_item_id)
                 for each_identifier in value_source["identifier"]
             ]
-        return self._lookup_answer(value_source["identifier"], list_item_id)
+        answer = self._lookup_answer(value_source["identifier"], list_item_id)
+        return (
+            answer.get(value_source["selector"])
+            if "selector" in value_source
+            else answer
+        )
```
ajmaddaford, line 88: It's not been introduced by this change, but this highlights that the values in `dict` answers aren't being escaped. We can either address it in this pull request, or raise a card for it?

ajmaddaford: You can test this by adding html into the address line 1 answer and seeing it rendered on the confirmation page.

test_schemas/en/test_address.json
LJBabbage: I reckon it will be ok, considering its a test schema

app/questionnaire/placeholder_parser.py
LJBabbage: If we would usually store HTML and escaped on playback, then it realistically should be this card. The only reason I therefore would suggest a different story would be because of the testing deadline for DAC tomorrow. It wouldn't matter to them

MebinAbraham (approved):

ajmaddaford (approved):

PR #302, opened 2020-09-23
Add address answer lookup options

ajmaddaford:
Adds address lookups to the address answer type. The behaviour of the lookup is configured using the `lookup_options` object from the schema.

### How to review 
You will need you IP whitelisted for the address lookup to work. Ask in the dev channel on Slack if you're not sure how.

- Use the `test_address_lookups.json` schema to test the lookups.
- Confirm that the uprn field is stored in the answer store (use the `/dump/debug` endpoint)
- Verify that the schema options populate the `data-type-` attributes in the html 

### Checklist


templates/partials/answers/address.html
```diff
@@ -23,9 +23,43 @@
...
+    "typeMore": _("Enter more of the address to get results"),
+    "autocomplete": "new-password",
+    "errorTitle": _("There is a problem with your answer"),
+    "errorMessageEnter": _("Enter an address"),
+    "errorMessageSelect": _("Select an address"),
+    "errorMessageAPI": _("Sorry, there was a problem loading addresses"),
+    "errorMessageAPILinkText": _("Enter address manually"),
+    "options": {
+        "region_code": answer.lookup_options.region_code | lower,
+        "one_year_ago": "true" if answer.lookup_options.one_year_ago is defined and answer.lookup_options.one_year_ago else "false",
```
MebinAbraham, line 59: How come these are these string values and not boolean?

tests/integration/components/test_address.py
```diff
@@ -54,3 +54,55 @@ def test_optional_address_displays_no_answer_provided_on_summary_when_not_provid
...
+            {
+                "address-mandatory-line1": "first address",
+                "address-mandatory-uprn": "123456789",
+            }
+        )
+
+        self.post({})
+
+        # Then
+        self.assertNotInBody("123456789")
```
MebinAbraham, line 108: I think it is also worth adding a test to assert that the UPRN is in the source when you go back if you post with the UPRN.

MebinAbraham (commented):
Looks good but I think we should add a test to assert that the UPRN does in fact get stored in the answer store.

templates/partials/answers/address.html
ajmaddaford: The macro expects strings (the values are assigned to html attributes).

MebinAbraham: Ah ofcourse. 👍

tests/integration/components/test_address.py
ajmaddaford: Yeah, fair point. I have added it as an assertion in the `test_address_fields_are_populated_in_when_revisiting_the_page` test.

templates/partials/answers/address.html
MebinAbraham: As discussed, this may need looking at in future as this doesn't match the doc. Hopefully in a future update, the macro itself can do the conversion.

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=extend-address-answer-type-for-lookups
```

MebinAbraham (approved):

MebinAbraham:
So we discussed that lookup itself didn't warrant a functional test and that the DS should be relied on for that. However, thoughts on adding a test when the lookup is enabled but the endpoint is not available?. Or are we happy to put that test as part of the DS as well?.
There is still a small risk in not testing it in runner, hopefully we can in future if the address index had a mock endpoint.

ajmaddaford:
@MebinAbraham I think there are a number of different errors that the address api could return, and it's the responsibility of the design system component to handle that, and therefore test for it. 

I do think there is a wider issue about being sure that the way runner uses the design system macros is correct and works.

app/jinja_filters.py
```diff
@@ -23,6 +23,7 @@ def format_number(value):
 
 
 def get_formatted_address(address_fields):
+    address_fields.pop("uprn", None)
```
JamesGardiner, line 26: Is `uprn` guaranteed to be present in the address_fields dict?

JamesGardiner (approved):
LGTM - one minor comment about whether UPRN is always expected in the address field. If so, it doesn't need need the default None value.

app/jinja_filters.py
MebinAbraham: It isn't guaranteed if you are not using the lookup.

JamesGardiner: It's a very very minor comment in any case - just if I see the above I assume that it will legitimately be missing as a key in some scenarios

ajmaddaford: No - when the lookup isn't used the `uprn` field is not stored in the address answer in the answer store.

PR #303, opened 2020-09-23
Add mobile number validation

iwootten:

Updates runner to introduce a new answer type `MobileNumber` which validates mobile numbers based on the requirements of the card (copied in below for reference):

***Validation:***
- Needs to ignore spaces
- Remove leading: 044, +44, 0 
- Remove all: Brackets   - all types, Hyphen, Full Stop, Comma, Whitespace, TAB space, Forward slash

- Check for numbers only - 'Valid UK number' error if not
- Check if 10 characters- 'Valid UK number' error if not

Replay the number they have entered in the way it has been entered

***Error messages:***
Validation message for field left empty:
- Enter a UK mobile number

Validation for all invalid answers, with an example of a valid format (as we do on email):
- Enter a UK mobile number in a valid format, for example, 07700 912345 or +44 7700 912345

The sanitised number will be: 7 xxx xxx xxx
The sanitised number code needs to be common code

### How to review 


iwootten:
[[H] Runner to Validate mobile telephone number answers (M)](https://trello.com/c/vtg2cAV1/4110-h-runner-to-validate-mobile-telephone-number-answers-m)

app/forms/validators.py
```diff
@@ -424,3 +421,26 @@ def __call__(self, answer_values, is_mandatory):
                 self.messages["MANDATORY_QUESTION"], self.question_title
             )
             raise validators.ValidationError(message)
+
+
+def sanitise_mobile_number(data):
+    for char in [" ", ".", ",", "\t", "-", "{", "}", "[", "]", "(", ")", "/"]:
+        data = data.replace(char, "")
+
+    return data.lstrip("+044").lstrip("044").lstrip("0")
```
MebinAbraham, line 430: `lstriip` Removes all leading character specified by the argument.
```py
data.lstrip("+044")
```
However, I am not sure `lstrip` is what we want. It does make it more flexible but it strips any number of the specified characters, for example, it would accept numbers like `+440000044x xxx xxx xxx` as it would interpret this as `xxxxxxxxxx`.

app/forms/validators.py
```diff
@@ -424,3 +421,26 @@ def __call__(self, answer_values, is_mandatory):
...
+    def __call__(self, form, field):
+        data = sanitise_mobile_number(field.data)
+
+        if len(data) != 10:
+            raise validators.StopValidation(self.messages["INVALID_MOBILE_NUMBER"])
+
+        try:
+            int(data)
+        except ValueError:
+            raise validators.ValidationError(self.messages["INVALID_MOBILE_NUMBER"])
```
MebinAbraham, line 446: Could combine these given try both display the same message?.
```py
        if len(data) != 10 or not data.isdigit():
            raise validators.StopValidation(self.messages["INVALID_MOBILE_NUMBER"])
```

MebinAbraham (changes requested):

tests/app/forms/validation/test_mobile_number_validation.py
```diff
@@ -0,0 +1,95 @@
...
+            validator(mock_form, mock_field)
+        except StopValidation:
+            self.fail("Valid phone number raised StopValidation")
+
+    def test_country_code_valid(self):
+        validator = MobileNumberCheck()
+
+        mock_form = Mock()
+        mock_field = Mock()
+        mock_field.data = "+0447812 395623"
```
ajmaddaford, line 90: Any valid phone numbers should come from the range 07700 900000 to 900999 (from https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbers-for-drama) to ensure we don't accidentally use a real phone number.

test_schemas/en/test_mobile_number.json
```diff
@@ -0,0 +1,58 @@
...
+                        {
+                            "type": "Question",
+                            "id": "mobile-number-block",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "mobile-number-answer",
+                                        "label": "UK mobile Number",
+                                        "mandatory": true,
+                                        "q_code": "0",
```
ajmaddaford, line 39: `q_code` is not a required property and should be removed.

test_schemas/en/test_mobile_number.json
```diff
@@ -0,0 +1,58 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "001",
+    "title": "Other input fields",
```
ajmaddaford, line 7: This should be updated.

templates/partials/answers/mobilenumber.html
```diff
@@ -0,0 +1,23 @@
...
+    "id": answer.id ~ "-label",
+    "text": answer.label,
+    "description": answer.description
+  },
+  "attributes": {
+    "data-qa": "input-text"
+  },
+  "legend": question_title,
+  "legendClasses": "u-vh",
+  "mutuallyExclusive": mutuallyExclusive,
```
ajmaddaford, line 19: We should consider having a test for a mobile number input in a mutually exclusive question.

app/forms/error_messages.py
```diff
@@ -54,6 +55,9 @@
         "Enter a 'period to' date later than the 'period from' date"
     ),
     "INVALID_DURATION": lazy_gettext("Enter a valid duration"),
+    "INVALID_MOBILE_NUMBER": lazy_gettext(
+        "Enter a UK mobile number in a valid format, for example, 07700 912345 or +44 7700 912345"
```
ajmaddaford, line 59: Where did this number come from? A google search says it might belong to someone 😬 

We should pick one in the range 07700 900000 to 900999 (from https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbers-for-drama) to ensure we don't accidentally use a real phone number.

iwootten: The card

tests/app/forms/validation/test_mobile_number_validation.py
iwootten: Fair enough, I'll update this - it was written as part of the validation message we put forward during planning.

templates/partials/answers/mobilenumber.html
iwootten: We don't have a use case for this though. Can we not remove this arg from the template until such a time as there is?

tests/app/forms/validation/test_mobile_number_validation.py
iwootten: Ah no, my mistake I must of copied this area code from elsewhere during testing. I'll update.

templates/partials/answers/mobilenumber.html
ajmaddaford: Yeah, or that.

app/forms/validators.py
```diff
@@ -424,3 +421,19 @@ def __call__(self, answer_values, is_mandatory):
                 self.messages["MANDATORY_QUESTION"], self.question_title
             )
             raise validators.ValidationError(message)
+
+
+def sanitise_mobile_number(data):
+    data = re.sub(r"[\s.,\t\-{}\[\]()/]", "", data)
+    return re.sub(r"^(\+044|044|0)", "", data)
```
ajmaddaford, line 428: The `+44` example in the error message doesn't pass validation as we're not stripping `+44`here.

tests/app/forms/field_handlers/test_mobile_number_handler.py
```diff
@@ -0,0 +1,25 @@
+from wtforms import Form, StringField
+
+from app.forms.field_handlers.mobile_number_handler import MobileNumberHandler
+
+
+def test_phone_number_handler():
+    string_field_definition = {
```
ajmaddaford, line 7: I think it would be clearer if this was named something like `answer_schema`.

app/forms/validators.py
MebinAbraham: I think `+044` was meant to be `+44` however, I don't think it's far-fetched to assume people do enter `+044` tbh.

tests/app/forms/validation/test_mobile_number_validation.py
```diff
@@ -0,0 +1,71 @@
...
+        validator = MobileNumberCheck()
+
+        mock_form = Mock()
+        mock_field = Mock()
+        mock_field.data = "+0447700 900333"
+
+        try:
+            validator(mock_form, mock_field)
+        except ValidationError:
+            self.fail("Valid phone number raised ValidationError")
```
MebinAbraham, line 71: I think this should be a subtest with the different valid formats we expect. Such as `+44` 0xxz etc.

iwootten: Why? This is already well tested in the sanitise test already.

app/forms/field_handlers/mobile_number_handler.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class MobileNumberHandler(FieldHandler):
+    MANDATORY_MESSAGE_KEY = "MANDATORY_MOBILE_NUMBER"
+
+    @cached_property
+    def validators(self):
+        validate_with = super().validators
+
+        if not self.disable_validation:
```
MebinAbraham, line 16: Do we actually use this for anything that doesn't have `options` or textarea?

app/forms/validators.py
iwootten: This was incorrect according to the card, so I've updated.

tests/app/forms/validation/test_mobile_number_validation.py
MebinAbraham: I don't understand the need of this test then?

app/forms/field_handlers/mobile_number_handler.py
iwootten: It's available on all the FieldHandlers. I've followed what's available in StringHandler as its basically the same as this bar the validation.

tests/app/forms/validation/test_mobile_number_validation.py
iwootten: Fair enough, I've removed it

app/views/handlers/individual_response.py
```diff
@@ -844,7 +845,7 @@ def handle_get(self):
     def handle_post(self):
         if self.selected_option == self.confirm_option:
             self._update_section_status(CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED)
-            self._publish_fulfilment_request(self.mobile_number)
+            self._publish_fulfilment_request(sanitise_mobile_number(self.mobile_number))
```
MebinAbraham, line 848: Had a chat with @ajmaddaford and come to a conclusion that this should probably be done when the fulfilment request payload is created. Otherwise, we have no guard against the IR payload containing an unsanatized number given we can't integration test the IR journey while checking the payload.

templates/partials/answers/mobilenumber.html
```diff
@@ -0,0 +1,22 @@
...
+    "text": answer.label,
+    "description": answer.description
+  },
+  "attributes": {
+    "data-qa": "input-text"
+  },
+  "legend": question_title,
+  "legendClasses": "u-vh",
+  "error": error
+} %}
```
MebinAbraham, line 20: This should conform the telephone input defined in: https://ons-design-system.netlify.app/components/input/
Such as: Setting the type to tel to bring up the correct keypad on touch devices.

tests/app/forms/validation/test_mobile_number_validation.py
```diff
@@ -0,0 +1,71 @@
...
+        validator = MobileNumberCheck()
+
+        mock_form = Mock()
+        mock_field = Mock()
+        mock_field.data = "൦൧൨൩൪൫൬൭൮൯"
+
+        with self.assertRaises(ValidationError) as ite:
+            validator(mock_form, mock_field)
+
+        self.assertEqual(error_messages["INVALID_MOBILE_NUMBER"], str(ite.exception))
```
MebinAbraham, line 71: Can't be certain this is 100% testing the non utf character since the exception is generic. I.e the length of that is only 10 and doesn't start with 0 etc. Could add a single non utf8 char in with utf8 numbers and test it is invalid i.g: `07700൫900333`

MebinAbraham (approved):

tests/app/forms/validation/test_mobile_number_validation.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_non_numeric_string_invalid(self):
+        validator = MobileNumberCheck()
+
+        mock_form = Mock()
+        mock_field = Mock()
+        mock_field.data = "07700e900333"
```
MebinAbraham, line 54: Should remove 1 number, as this could be failing because it has length greater than 10 and not because it has a non numeric.

tests/app/forms/validation/test_mobile_number_validation.py
iwootten: Not sure I follow - because it's a string 10 in length is exactly why the test proves the error is thrown correctly?

MebinAbraham: My example had 11 chars because the validator would remove the firs `0` hence making it 10 in length, hence the `len(data) != 10` would pass and the non 0-9 check would fail. But I changed my mind, I think what you have is fine.

ajmaddaford (approved):

PR #305, opened 2020-09-24
Limit the number of confirmation emails that can be sent

berroar:
Sets a limit to the number of confirmation emails that can be sent. 

The limit is set using the `CONFIRMATION_EMAIL_REQUEST_LIMIT` environment variable.

If the limit has been reached, then the option to send another email is no longer displayed on the `thank-you` and `confirmation-email/send` pages. 

**Note**:  The content on the Trello card states that an error message needs to be raised when the limit is reached, but after further discussion it was decided this is not needed, and to stop the option of sending the emails from being displayed instead.

### How to review 
Set the new environment variable `CONFIRMATION_EMAIL_REQUEST_LIMIT` in your config, and use the `test_confirmation_email` schema to test this change.  

### Checklist


berroar:
[[M] Limit on number of emails sent (M)](https://trello.com/c/zLq7H1Ru/4069-m-limit-on-number-of-emails-sent-m)

LJBabbage (changes requested):
This is bit more complicated than it could be, and we are instantiating classes we don't need. It would be far simpler to stop ConfirmationEmail being created in thank you if the count was too high. Doing this will mean no more logic is needed, as the class will act like it was a census schema without a confirmation email flag in the schema

In the other 2 end points we should try to again try to avoid instantiating ConfirmationEmail and then EmailForm if we don't need thrm. We should be checking in the session first and then responding accordingly

ajmaddaford (changes requested):

app/routes/questionnaire.py
```diff
@@ -279,19 +284,22 @@ def get_thank_you(schema):
...
 @post_submission_blueprint.route("confirmation-email/send", methods=["GET", "POST"])
 def send_confirmation_email():
-    if not get_session_store().session_data.confirmation_email_sent:
+    if not get_session_store().session_data.confirmation_email_count:
         raise NotFound
 
     confirmation_email = ConfirmationEmail()
 
+    if email_limit_exceeded():
+        return redirect(url_for("post_submission.get_thank_you"))
```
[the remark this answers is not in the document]
MebinAbraham: Or `email_limit_exceeded` should be property of `ConfirmationEmail` which ThankYou page has access to?. Thoughts?

app/routes/questionnaire.py
```diff
@@ -279,19 +284,22 @@ def get_thank_you(schema):
...
+        content=thank_you.get_context(hide_confirmation),
         survey_id=schema.json["survey_id"],
         page_title=thank_you.get_page_title(),
     )
 
 
 @post_submission_blueprint.route("confirmation-email/send", methods=["GET", "POST"])
 def send_confirmation_email():
-    if not get_session_store().session_data.confirmation_email_sent:
+    if not get_session_store().session_data.confirmation_email_count:
```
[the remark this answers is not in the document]
berroar: If I remove this then a couple of the tests brake e.g. asserting that the POST to the thank you page without completing the questionnaire raises a 404 - so do we want to do this/should the tests be removed?

k8s/helm/templates/deployment.yaml
```diff
@@ -121,3 +121,5 @@ spec:
               value: "{{- .Values.newRelic.licenseKey }}"
             - name: NEW_RELIC_APP_NAME
               value: "{{- .Values.newRelic.appName }}"
+            - name: CONFIRMATION_EMAIL_REQUEST_LIMIT
+              value: "10"
```
MebinAbraham, line 125: You should make this configurable using `Values` as there is potential we want to increase this in prod etc.

app/views/handlers/confirmation_email.py
```diff
@@ -27,5 +28,14 @@ def get_page_title(self):
...
         session_store.save()
+
+
+def email_limit_exceeded():
+    return (
+        True
+        if get_session_store().session_data.confirmation_email_count
+        >= current_app.config["CONFIRMATION_EMAIL_REQUEST_LIMIT"]
+        else False
+    )
```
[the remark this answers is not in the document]
berroar: Added extra logic for this

app/routes/questionnaire.py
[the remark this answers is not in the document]
ajmaddaford: Ah, it's probably something wider than this pull request then. We should really be checking in the `post_submission_blueprint.before_request` whether the questionnaire is complete or not.

app/views/handlers/confirmation_email.py
```diff
@@ -27,5 +28,17 @@ def get_page_title(self):
...
+def email_limit_exceeded():
+    session_store = get_session_store()
+
+    return (
+        True
+        if session_store
+        and session_store.session_data.confirmation_email_count
+        >= current_app.config["CONFIRMATION_EMAIL_REQUEST_LIMIT"]
+        else False
+    )
```
iwootten, line 44:
```suggestion
    return (
        session_store
        and session_store.session_data.confirmation_email_count
        >= current_app.config["CONFIRMATION_EMAIL_REQUEST_LIMIT"]
    )
```

app/routes/questionnaire.py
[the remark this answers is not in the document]
ajmaddaford: I think that sounds good as it would be consistent with the other handlers.

templates/census-thank-you.html
```diff
@@ -48,10 +48,12 @@ <h1 data-qa="title">
     <p>{{ _("Your personal information is protected by law and will be kept confidential") }}</p>
   {% endcall %}
 
-  {% if form %}
-    <h2>{{ _("Get confirmation email") }}</h2>
-    <p>{{ _("If you would like to be sent confirmation that you have completed your census enter your email address") }}</p>
-    {% include 'partials/email-form.html' %}
+  {% if content.hide_confirmation == False %}
```
ajmaddaford, line 51: `hide_confirmation` should be `show_confirmation_email_form`.

templates/confirmation-email-sent.html
```diff
@@ -14,9 +14,11 @@
   <h1 data-qa="confirmation-text">{{ _("A confirmation email has been sent to {email}").format(email = content.email)  }}</h1>
  {% endcall %}
 
-  <p><strong>{{ _("Didn't receive an email?")  }}</strong></p>
-  <p>{{ _("It can take a few minutes for the email to arrive. If it doesn't arrive, check your junk mail, or you can
-    <a href='{send_confirmation_email_url}' id='send-another-email'>send another confirmation email</a>.</p>").format(send_confirmation_email_url = content.send_confirmation_email_url)  }}
+  {% if content.show_send_another_email_guidance == False %}
```
ajmaddaford, line 17: I think the logic is a bit off here, as this should be:
```suggestion
  {% if content.show_send_another_email_guidance %}
```

k8s/helm/templates/deployment.yaml
```diff
@@ -121,3 +121,5 @@ spec:
               value: "{{- .Values.newRelic.licenseKey }}"
             - name: NEW_RELIC_APP_NAME
               value: "{{- .Values.newRelic.appName }}"
+            - name: CONFIRMATION_EMAIL_REQUEST_LIMIT
+              value: "{{- .Values.confirmationEmailRequestLimit }}"
```
ajmaddaford, line 125: There should be a default in `values.yaml`, so that this only needs to be provided for environments that need to change it.

ajmaddaford (changes requested):

app/routes/questionnaire.py
```diff
@@ -324,6 +330,9 @@ def get_confirmation_email_sent():
                 "post_submission.send_confirmation_email"
             ),
             "hide_signout_button": False,
+            "show_send_another_email_guidance": True
+            if not email_limit_exceeded()
+            else False,
```
ajmaddaford, line 335:
```suggestion
            "show_send_another_email_guidance": not email_limit_exceeded(),
```

templates/census-thank-you.html
```diff
@@ -48,10 +48,12 @@ <h1 data-qa="title">
...
-  {% if form %}
-    <h2>{{ _("Get confirmation email") }}</h2>
-    <p>{{ _("If you would like to be sent confirmation that you have completed your census enter your email address") }}</p>
-    {% include 'partials/email-form.html' %}
+  {% if content.show_confirmation_email_form %}
+    {% if form %}
+      <h2>{{ _("Get confirmation email") }}</h2>
+      <p>{{ _("If you would like to be sent confirmation that you have completed your census enter your email address") }}</p>
+      {% include 'partials/email-form.html' %}
+    {% endif %}
```
ajmaddaford, line 56: The `show_confirmation_email_form` conditional isn't needed now that form is only present when the limit hasn't been reached.
```suggestion
{% if form %}
  <h2>{{ _("Get confirmation email") }}</h2>
  <p>{{ _("If you would like to be sent confirmation that you have completed your census enter your email address") }}</p>
  {% include 'partials/email-form.html' %}
```

app/routes/questionnaire.py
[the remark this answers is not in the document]
berroar: I've added this to init, not sure if there's more I can refactor now within `thank_you.py` or not

app/views/handlers/confirmation_email.py
```diff
@@ -6,10 +7,17 @@
 from app.views.contexts.email_form_context import build_confirmation_email_form_context
 
 
+class EmailLimitExceeded(Exception):
```
MebinAbraham, line 10: I think this should be consistent with the class name so `ConfirmationEmailLimitExceeded`. Also, should it be `ConfirmationEmailLimitExceeded` or `ConfirmationEmailLimitReached`.

app/views/handlers/thank_you.py
```diff
@@ -35,6 +38,7 @@ def __init__(self, schema):
         self.confirmation_email = (
             ConfirmationEmail(self.PAGE_TITLE)
             if self._schema.get_submission().get("confirmation_email")
+            and not email_limit_exceeded()
             else None
         )
```
MebinAbraham, line 43: Not sure what peoples thoughts will be but now that `ConfirmationEmail` checks if the limit has been reached, the above code would check the if the limit has been reached twice. To avoid that, this could be something like:
```py
        self.confirmation_email = self._get_confirmation_email()

    def _get_confirmation_email(self):
        if not self._schema.get_submission().get("confirmation_email"):
            return None
        
        try:
            return ConfirmationEmail(self.PAGE_TITLE)
        except EmailLimitExceeded:
            return None
```

app/settings.py
```diff
@@ -103,4 +103,9 @@ def get_env_or_fail(key):
 MAX_NUMBER = 9999999999
 
 COOKIE_SETTINGS_URL = os.getenv("COOKIE_SETTINGS_URL")
+
+CONFIRMATION_EMAIL_REQUEST_LIMIT = int(
+    os.getenv("CONFIRMATION_EMAIL_REQUEST_LIMIT", "10")
```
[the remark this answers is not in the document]
berroar: I've taken `request` out now

app/views/handlers/confirmation_email.py
berroar: Updated to use reached

app/views/handlers/thank_you.py
berroar: Yeah I think this is probably better than checking twice, updated now

k8s/helm/values.yaml
```diff
@@ -46,3 +46,5 @@ cdn:
 
 address_lookup_api:
   url: "https://whitelodge-ai-api.census-gcp.onsdigital.uk"
+
+confirmationEmailRequestLimit: 10
```
ajmaddaford, line 50: This should be changed to `confirmationEmailLimit ` to be consistent with the environment variable change

app/views/contexts/thank_you_context.py
```diff
@@ -47,6 +47,7 @@ def build_census_thank_you_context(
         "census_type": census_type_code,
         "hide_signout_button": False,
         "sign_out_url": url_for("session.get_sign_out"),
+        "show_confirmation_email_form": True if confirmation_email_form else False,
```
ajmaddaford, line 50: I don't think this is used anymore?

ajmaddaford (changes requested):

app/views/handlers/confirmation_email.py
```diff
@@ -27,5 +34,15 @@ def get_page_title(self):
...
+
+
+def email_limit_reached():
+    session_store = get_session_store()
+
+    return (
+        session_store
+        and session_store.session_data.confirmation_email_count
+        >= current_app.config["CONFIRMATION_EMAIL_LIMIT"]
+    )
```
ajmaddaford, line 48: I think we could make this a little neater if we made this a static method within the `ConfirmationEmail` class. That way we wouldn't need to import `email_limit_reached` in `questionnaire.py` - we could call `ConfirmationEmail.is_limit_reached()` instead.

```suggestion

    @staticmethod
    def is_limit_reached():
        session_store = get_session_store()

        return (
            session_store
            and session_store.session_data.confirmation_email_count
            >= current_app.config["CONFIRMATION_EMAIL_LIMIT"]
        )
```

templates/confirmation-email-sent.html
```diff
@@ -14,9 +14,11 @@
   <h1 data-qa="confirmation-text">{{ _("A confirmation email has been sent to {email}").format(email = content.email)  }}</h1>
  {% endcall %}
 
-  <p><strong>{{ _("Didn't receive an email?")  }}</strong></p>
-  <p>{{ _("It can take a few minutes for the email to arrive. If it doesn't arrive, check your junk mail, or you can
-    <a href='{send_confirmation_email_url}' id='send-another-email'>send another confirmation email</a>.</p>").format(send_confirmation_email_url = content.send_confirmation_email_url)  }}
+  {% if content.show_send_another_email_guidance %}
+    <p><strong>{{ _("Didn't receive an email?")  }}</strong></p>
+    <p>{{ _("It can take a few minutes for the email to arrive. If it doesn't arrive, check your junk mail, or you can
+      <a href='{send_confirmation_email_url}' id='send-another-email'>send another confirmation email</a>.</p>").format(send_confirmation_email_url = content.send_confirmation_email_url)  }}
```
ajmaddaford, line 20: This was already incorrect, but it should be fixed here. The new line and closing `p` tag shouldn't be in this string:
```suggestion
    <p>{{ _("It can take a few minutes for the email to arrive. If it doesn't arrive, check your junk mail, or you can <a href='{send_confirmation_email_url}' id='send-another-email'>send another confirmation email</a>.").format(send_confirmation_email_url = content.send_confirmation_email_url)}}</p>
```

ajmaddaford (approved):

tests/integration/integration_test_case.py
```diff
@@ -55,7 +55,10 @@ def _set_up_app(self):
 
         configure_logging()
 
-        setting_overrides = {"EQ_ENABLE_HTML_MINIFY": False}
+        setting_overrides = {
+            "EQ_ENABLE_HTML_MINIFY": False,
+            "CONFIRMATION_EMAIL_LIMIT": 2,
```
LJBabbage, line 60: Probably better putting this in the test, as this doesn't make it global to all tests. Being here also hides the fact its being overridden to some extent. We do this in a number of places, for reference look at test_timeout.py

```
    def setUp(self):
        settings.CONFIRMATION_EMAIL_LIMIT = 2
        super().setUp()
```

LJBabbage (changes requested):
One minor point, but the rest looks good.

tests/integration/integration_test_case.py
berroar: Moved this

LJBabbage (approved):

PR #310, opened 2020-09-28
Prevent IR request link from appearing on all interstitial pages

petechd:
This change prevents individual response link from appearing on all interstitial pages. [Trello card](https://trello.com/c/vp9wsodA).

### How to review 
Manually run `test_individual_response` schema to check if this change works, use _test_individual_response.py_ integration test for automated testing.

### Checklist


petechd:
[Bug IR request link appearing on all interstitial pages](https://trello.com/c/vp9wsodA/4161-bug-ir-request-link-appearing-on-all-interstitial-pages)

app/views/handlers/__init__.py
```diff
@@ -38,3 +39,12 @@ def individual_response_url(
                 journey=journey,
             )
     return None
+
+
+def show_individual_response_link(location, routing_path):
+    if (
+        location.block_id == routing_path[0]
+        and location.list_item_id != "primary-person"
```
MebinAbraham, line 47: `lisT_item_id` Is never `primary-person`.

app/views/handlers/__init__.py
```diff
@@ -38,3 +39,12 @@ def individual_response_url(
...
+        and location.list_name == "household"
```
MebinAbraham, line 48: In the schema, you can get the section id of the individual section:
```
"individual_response": {
        "for_list": "household",
        "individual_section_id": "individual-section"
    },
```
You then just need to check if the current location is the first block in that section.

app/views/handlers/__init__.py
petechd: Sorry, I was checking something and left i there..

templates/interstitial.html
```diff
@@ -19,7 +19,7 @@ <h1 class="question__title">{{ block.content.title }}</h1>
   {% endif %}
   {% set contents = block.content.contents %}
   {% include 'partials/contents.html' %}
-  {% if content.individual_response_url %}
+  {% if content.individual_response_url and content.show_individual_response_link %}
```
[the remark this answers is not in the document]
petechd: So, it's only one flag now, _show_individual_response_link_. Thanks👍

app/views/handlers/__init__.py
petechd: Removed all this and added `schema.json.get("individual_response", {}).get("individual_section_id")` to search for section id and `schema.get_first_block_id_for_section(section)` to compare with current block. Just need to add some tests for coverage now, thanks!

app/views/handlers/__init__.py
```diff
@@ -38,3 +39,9 @@ def individual_response_url(
                 journey=journey,
             )
     return None
+
+
+def show_individual_response_link(location, schema):
+    section = schema.json.get("individual_response", {}).get("individual_section_id")
+    if location.block_id == schema.get_first_block_id_for_section(section):
+        return True
```
iwootten, line 47: This currently doesn't return false.
```suggestion
    return location.block_id == schema.get_first_block_id_for_section(section)
```

app/views/handlers/content.py
```diff
@@ -34,6 +34,14 @@ def get_context(self):
                 self._current_location.list_item_id,
                 self._questionnaire_store,
             ),
+            "show_individual_response_link": show_individual_response_link(
+                self._current_location, self._schema
+            )
```
iwootten, line 39: Currently there's a lot of duplication here.

Given both the hub/interstitial templates already determine visibility based on existence of the url, I would expect the existing individual_response_url function to be updated to factor in the requirements of the interstitial pages, rather than a new variable be created and only used on interstitial templates.

iwootten (changes requested):

app/views/handlers/__init__.py
petechd: Added your suggestion, thanks!

app/views/handlers/content.py
```diff
@@ -33,7 +33,8 @@ def get_context(self):
                 self._schema.get_individual_response_list(),
                 self._current_location.list_item_id,
                 self._questionnaire_store,
-            ),
+            )
+            and show_individual_response_link(self._current_location, self._schema),
```
MebinAbraham, line 37: You can either check `show_individual_response_link` in templates or set `individual_response_url` to None when `show_individual_response_link` is False. The first way is more explicit.
Any preference @iwootten?

MebinAbraham (commented):
Would need to add a test to ensure this bug doesn't resurface,

app/views/handlers/content.py
iwootten: I think the best thing would be for the value of show_individual_response_link (as `show_individual_response`) is passed to the interstitial/hub templates and that is used to determine visibility. We should update my pr changes following it's merge so it factors into the value of show_individual_response_link

iwootten: For what it's worth, I think my PR and this should have been done as a single card. They both require thinking about when we decide to show the link.

MebinAbraham: So in your recent change, for the hub we passed the URL to the templates only when we know we can show it. Are you suggesting that should be changed, and we should pass, both the URL and show_ir flag?

iwootten: I don't mind either way - I think passing the flag through makes sense and I don't actually think we should be setting the url attribute at all if we don't have one. I backtracked on this approach on my own pr though as I think there's some subtleties to how the url is generated that may cause issues.

petechd: Making `individual_response_url` visible on `show_individual_response` flag set to True is breaking all sorts of tests, seems like `individual_response_url` is needed in many other places that use handlers __init__ methods, it's not only used by interstitial pages. Or should we change that `show_individual_response` method so it evaluates something else then first block of individual section?

iwootten: I think that's likely the reason why I gave up on that approach - I say just go with the appraoch @MebinAbraham has suggested then

petechd: I tried your change @MebinAbraham  and I get `individual_response_url` evaluated to None when I try to access IR link from remove person screen, so what should be `show_individual_response` method check, other than first block of individual section? This is the only test it currently breaks on.
Screenshot 2020-09-29 at 13 14 38

app/views/handlers/content.py
petechd: Duplication removed, extra variable gone. Just need to add tests now👍.

app/views/handlers/__init__.py
```diff
@@ -38,3 +39,8 @@ def individual_response_url(
                 journey=journey,
             )
     return None
+
+
+def show_individual_response(location, schema):
+    section = schema.json.get("individual_response", {}).get("individual_section_id")
```
iwootten, line 45: This method would be better placed in questionnaire_schema class.

app/questionnaire/questionnaire_schema.py
```diff
@@ -497,3 +497,10 @@ def _get_error_messages(self):
             messages.update(self.json["messages"])
 
         return messages
+
+    @staticmethod
+    def show_individual_response(location, schema):
+        section = schema.json.get("individual_response", {}).get(
```
iwootten, line 503: I think you've misunderstood my original comment here - the first line here that returns the section_id should be a separate method I think and shouldn't be static. I think this method name should probably change too as it doesn't describe the check. Something like `is_individual_response_interstitial` makes more sense.

MebinAbraham: For this method, `is_first_block_in_indiividual_response` is probably more appropriate since it being an interstitial is irrelevant.

app/views/handlers/__init__.py
```diff
@@ -38,3 +39,8 @@ def individual_response_url(
                 journey=journey,
             )
     return None
+
+
+def is_first_block_in_individual_response(location, schema):
```
MebinAbraham, line 44: I think we should move this to the content handler since it only applies to that.

app/views/handlers/__init__.py
iwootten: This call to individual_section_id is also in individual_response and should be refactored to use the common method mentioned above.

MebinAbraham: As Ian suggest, fetching the section id can be moved to questionnaire schema. And then we should update individual_response handler to use that new method. currently it is:
```py
@cached_property
    def individual_section_id(self):
        return self._schema.json.get("individual_response", {}).get(
            "individual_section_id"
        )
```

app/views/handlers/__init__.py
petechd: Done!

app/views/handlers/content.py
```diff
@@ -47,3 +44,9 @@ def _get_content_title(self, transformed_block):
         content = transformed_block.get("content")
         if content:
             return self._get_safe_page_title(content["title"])
+
+    def _is_first_block_in_individual_response(self, location, schema):
```
MebinAbraham, line 48: Might be better suited as `is_block_first_block_in_individual_response`. Also you only need to pass the block_id not the location object.

petechd: Ok, done. Renamed and passed only block_id now👌.

app/questionnaire/questionnaire_schema.py
petechd: @iwootten that first line is now a method in Questionnaire Schema class. Thanks!

app/views/handlers/content.py
petechd: Done!

app/views/handlers/content.py
```diff
@@ -45,8 +45,6 @@ def _get_content_title(self, transformed_block):
...
             return self._get_safe_page_title(content["title"])
 
-    def _is_first_block_in_individual_response(self, location, schema):
-        section = schema.json.get("individual_response", {}).get(
-            "individual_section_id"
-        )
-        return location.block_id == schema.get_first_block_id_for_section(section)
+    def _is_block_first_block_in_individual_response(self, block_id, schema):
+        section_id = schema.get_individual_response_individual_section_id()
+        return block_id == schema.get_first_block_id_for_section(section_id)
```
MebinAbraham, line 50: I realised we should be checking first loction on the path not the schema. So it should be something like:
```py
    def _is_individual_response_first_location(self):
        individual_section_id = self._schema.get_individual_response_individual_section_id()
        return (
            self.current_location.section_id == individual_section_id
            and self._current_location.block_id == self._routing_path[0]
        )
```
Feel free to get suggestion on method naming.

app/views/handlers/__init__.py
petechd: This is now using that new method, thanks!

app/views/handlers/content.py
petechd: In my defence, I did use ` location.block_id == routing_path[0]` at the beginning of this PR. Refactored, thanks 👍.

iwootten (approved):

app/questionnaire/questionnaire_schema.py
```diff
@@ -497,3 +497,8 @@ def _get_error_messages(self):
             messages.update(self.json["messages"])
 
         return messages
+
+    def get_individual_response_individual_section_id(self):
+        return self._questionnaire_json.get("individual_response", {}).get(
+            "individual_section_id"
+        )
```
MebinAbraham, line 504: Would be nice to group this with the other IR methods in this class for clarity.

test_schemas/en/test_individual_response.json
```diff
@@ -333,6 +333,19 @@
                             "id": "individual-interstitial",
                             "type": "Interstitial"
                         },
+                        {
+                            "content": {
+                                "contents": [
+                                    {
+                                        "description": "Another interstitial page to check if individual response guidance is not present."
```
MebinAbraham, line 340:
```suggestion
                                        "description": "Another interstitial page to ensure that the individual response guidance is only displayed for the first block in the individual section."
```

MebinAbraham (approved):

app/questionnaire/questionnaire_schema.py
petechd: Moved it to ir group now.

MebinAbraham (approved):

PR #311, opened 2020-09-28
Introduce new 'show_on_hub' property to IR

iwootten:

NISRA have decided not to have the IR link on the hub (will only be available on individual introduction and remove person pages). We need to add a new property to the schema to support this.

### How to review 

Use the existing `test_individual_response` schema and confirm that the links to the individual section only show upon the introduction for the individual section.


iwootten:
[NI - remove IR link displayed from the hub (M)](https://trello.com/c/X3dILIul/4162-ni-remove-ir-link-displayed-from-the-hub-m)

berroar (commented):
All seems to work OK, do we need to add a test for this though?

iwootten:
> All seems to work OK, do we need to add a test for this though?

I'd actually thought I'd modified the original schema here, but looks like I missed it.

berroar (approved):

test_schemas/en/test_individual_response_hub_disabled.json
```diff
@@ -0,0 +1,418 @@
+{
```
MebinAbraham, line 1: Can the name be `test_individual_response_on_hub_disabled` as i thought it was IR but hub is disabled, not that IR on the hub is disabled.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -79,6 +79,22 @@ def _request_individual_response(self):
...
+
+    def test_show_on_hub_false(self):
+        self._add_household_no_primary()
+
+        with self.assertRaises(AttributeError) as error:
+            individual_response_link = self.individual_response_link
+
+        self.assertEqual(
+            str(error.exception), "'NoneType' object has no attribute 'find_next'"
+        )
```
MebinAbraham, line 95: Personally, I don't find this very clear on why this is happening. I think `self. individual_response_link` should return `None` when there isn't a link and this assert to be updated to `self.assertIsNone(self.individual_response_link)`. Thoughts?

test_schemas/en/test_individual_response_hub_disabled.json
iwootten: Also updated in the validator repo

tests/integration/individual_response/test_individual_response.py
```diff
@@ -79,6 +80,19 @@ def _request_individual_response(self):
         self.post()
 
 
+class TestIndividualResponseHubDisabled(IndividualResponseTestCase):
```
MebinAbraham, line 83: `TestIndividualResponseHubDisabled ` > `TestIndividualResponseOnHubDisabled`

tests/integration/individual_response/test_individual_response.py
```diff
@@ -79,6 +80,19 @@ def _request_individual_response(self):
...
+    def setUp(self):
+        super().setUp()
+        self.launchSurvey(
+            "test_individual_response_on_hub_disabled", region_code="GB-ENG"
+        )
+
+    def test_show_on_hub_false(self):
+        self._add_household_no_primary()
+
+        self.assertIsNone(self.individual_response_link)
```
MebinAbraham, line 93: I think we should also assert that you are on the hub, as this would pass for any other page also.

MebinAbraham (approved):

PR #314, opened 2020-09-28
Make web server settings configurable

MebinAbraham:
Makes the web server settings configurable so that we can't easily swap out web server/settings as well as enable and disable the use of GRPC for Datastore.
Because it seemed appropriate, I have renamed `GUNICORN_KEEP_ALIVE` to `HTTP_KEEP_ALIVE` to be more explicit.

### How to review 
Ensure:
- app works using the different web server commands in Makefile
- the env vars are actually being applied as expected.
- helm deployment of the app works as expected. Ensure you can changes settings as expected. You can use https://github.com/ONSdigital/eq-pipelines/pull/87 to test.

### Checklist


.development.env
```diff
@@ -16,7 +16,7 @@ EQ_NEW_RELIC_ENABLED=False
 EQ_SECRETS_FILE=dev-secrets.yml
 EQ_KEYS_FILE=dev-keys.yml
 AWS_DEFAULT_REGION=eu-west-1
-GUNICORN_KEEP_ALIVE=2
+HTTP_KEEP_ALIVE=2
```
JamesGardiner, line 19: Is this a TCP or HTTP keep alive?

Dockerfile
```diff
@@ -7,12 +7,13 @@ RUN apt update && apt install -y curl unzip libsnappy-dev build-essential jq
...
 
-ENV WEB_SERVER gunicorn
+ENV WEB_SERVER_TYPE gunicorn
+ENV WEB_SERVER_WORKER_CLASS gevent
 ENV WEB_SERVER_WORKERS 3
-ENV GUNICORN_KEEP_ALIVE 2
-ENV GUNICORN_CMD_ARGS -c gunicorn_config.py
 ENV WEB_SERVER_THREADS 10
 ENV WEB_SERVER_ASYNC_CORES 10
+ENV HTTP_KEEP_ALIVE 2
```
JamesGardiner, line 15:
```suggestion
ENV WEB_SERVER_KEEP_ALIVE 2
```

I don't think we should make an assumption about what protocol is being used for the keepalive

.development.env
JamesGardiner, line 19:
```suggestion
WEB_SERVER_KEEP_ALIVE=2
```

I don't think we should make an assumption about what protocol is being used for the keepalive

JamesGardiner (changes requested):
Also needs the gRPC flag to be configurable 👍

Makefile
```diff
@@ -55,17 +55,20 @@ link-development-env:
 run: build link-development-env
 	pipenv run flask run
 
-run-gunicorn: link-development-env
-	WEB_SERVER=gunicorn pipenv run ./run_app.sh
+run-gunicorn-gevent: link-development-env
+	WEB_SERVER_TYPE=gunicorn WEB_SERVER_WORKER_CLASS=gevent pipenv run ./run_app.sh
+
+run-gunicorn-gthread: link-development-env
```
JamesGardiner, line 61: What do you think about making an assumption that gRPC will always be on when using gthread?

MebinAbraham:
> Also needs the gRPC flag to be configurable 👍

Done

Dockerfile
MebinAbraham: I did HTTP to make it explicit. For uwsgi, it has http_keepalive and so_keepalive for sockets.

.development.env
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/314#discussion_r496597536

Makefile
MebinAbraham: Could do, what about uwsgi?. Also, for dev, it shouldn't make much of a difference.

JamesGardiner: uWSGI as well

app/setup.py
```diff
@@ -279,7 +279,7 @@ def setup_dynamodb(application):
 
 
 def setup_datastore(application):
-    client = datastore.Client(_use_grpc=False)
+    client = datastore.Client(_use_grpc=application.config["DATASTORE_USE_GRPC"])
```
ajmaddaford, line 282: There is an environment variable (GOOGLE_CLOUD_DISABLE_GRPC) used by the library that we could consider using - https://github.com/googleapis/python-datastore/blob/master/google/cloud/datastore/client.py#L60

Although, that is misleading if the PubSub library ignores it 🤔

Makefile
```diff
@@ -55,17 +55,20 @@ link-development-env:
...
-	WEB_SERVER=uwsgi pipenv run ./run_app.sh
+	WEB_SERVER_TYPE=uwsgi pipenv run ./run_app.sh
 
 run-uwsgi-threads: link-development-env
-	WEB_SERVER=uwsgi-threads pipenv run ./run_app.sh
+	WEB_SERVER_TYPE=uwsgi-threads pipenv run ./run_app.sh
 
 run-uwsgi-async: link-development-env
-	WEB_SERVER=uwsgi-async pipenv run ./run_app.sh
+	WEB_SERVER_TYPE=uwsgi-async pipenv run ./run_app.sh
```
ajmaddaford, line 71: The run commands for gunicorn and uwsgi are inconsistent. We should consider `WEB_SERVER_TYPE` of `gunicorn-threads` and `gunicorn-async`.

k8s/helm/values.yaml
```diff
@@ -13,6 +13,16 @@ googleTagManagerId: ""
 googleTagManagerAuth: ""
 googleTagManagerPreview: ""
 
+webServer:
+  type: "gunicorn"
+  workerClass: "gevent"
+  workers: "7"
+  threads: "10"
+  asyncCores: "10"
```
ajmaddaford, line 21: If this only applies to uwsgi, this should be named something like `uwsgiAsyncCores`.

app/setup.py
MebinAbraham: Myself and James did discuss this env var previous;y but chose to use our own one to have finer control on which lib has grpc enabled etc.

Makefile
MebinAbraham: Fair point however, we would still need `worker_class = os.getenv("WEB_SERVER_WORKER_CLASS")` given we use a py config file and not cmd line args.

ajmaddaford: I was thinking the env var would be set in the run_app script.

k8s/helm/values.yaml
ajmaddaford: Sorry, I should have been clearer. This should be changed everywhere e.g. the env var should be `WEB_SERVER_UWSGI_ASYNC_CORES` (or `UWSGI_ASYNC_CORES`).

Makefile
MebinAbraham: Update webserver type to be consistent and to use env var in run_app script.

k8s/helm/values.yaml
MebinAbraham: I did wonder 😄 
I went with `UWSGI_ASYNC_CORES` because we have `GUNICORN_CMD_ARGS`.

run_app.sh
```diff
@@ -2,16 +2,16 @@
...
+    run_command="WEB_SERVER_WORKER_CLASS=gthread gunicorn application:application"
+elif [ "$WEB_SERVER_TYPE" = "uwsgi" ]; then
     run_command="uwsgi uwsgi.ini --workers ${WEB_SERVER_WORKERS}"
-elif [ "$web_server" = "uwsgi-threads" ]; then
+elif [ "$WEB_SERVER_TYPE" = "uwsgi-threads" ]; then
     run_command="uwsgi uwsgi.ini --workers ${WEB_SERVER_WORKERS} --enable-threads --threads ${WEB_SERVER_THREADS}"
-elif [ "$web_server" = "uwsgi-async" ]; then
-    run_command="uwsgi uwsgi.ini --module patched:application --workers ${WEB_SERVER_WORKERS} --single-interpreter --gevent ${WEB_SERVER_ASYNC_CORES}"
+elif [ "$WEB_SERVER_TYPE" = "uwsgi-async" ]; then
+    run_command="uwsgi uwsgi.ini --module patched:application --workers ${WEB_SERVER_WORKERS} --single-interpreter --gevent ${UWSGI_ASYNC_CORES}"
```
ajmaddaford, line 14: I think we need a different environment variable name. uWSGI is parsing this and reading it as a value for an unknown directive `async-cores`:
```
[strict-mode] unknown config directive: async-cores
```

MebinAbraham: Have renamed `UWSGI_ASYNC_CORES` to `WEB_SERVER_UWSGI_ASYNC_CORES`

ajmaddaford (approved):

JamesGardiner (approved):

pricem14pc:
Do we need to add these to the list of environment variable params in deploy_app.yaml to allow them to be overridden?

```
  CDN_URL:
  WEB_SERVER_TYPE:
  WEB_SERVER_WORKERS:
  WEB_SERVER_THREADS:
  WEB_SERVER_UWSGI_ASYNC_CORES:
  DATASTORE_USE_GRPC:
```

MebinAbraham:
> Do we need to add these to the list of environment variable params in deploy_app.yaml to allow them to be overridden?
> 
> ```
>   CDN_URL:
>   WEB_SERVER_TYPE:
>   WEB_SERVER_WORKERS:
>   WEB_SERVER_THREADS:
>   WEB_SERVER_UWSGI_ASYNC_CORES:
>   DATASTORE_USE_GRPC:
> ```

The ones introduced in this PR has already been added I believe.

MebinAbraham:
@ajmaddaford @JamesGardiner Made minor changes to the README, can you make sure you are happy with the changes. Thanks

JamesGardiner (approved):

ajmaddaford (approved):

PR #318, opened 2020-09-29
Add response limit for IR

LJBabbage:
This PR introduces a response limit to individual response requests. Once the limit is hit a 429 is generate with a message highlighting that fact

### How to review 
First in your settings change (or set the env variable) EQ_INDIVIDUAL_RESPONSE_LIMIT to something sensible as its currently set at 75, once done, request an IR and make sure it errors on the limit that you have set

### Checklist


app/routes/errors.py
```diff
@@ -71,6 +72,12 @@ def method_not_allowed(error=None):
     return _render_error_page(405, template="404")
 
 
+@errors_blueprint.app_errorhandler(IndividualResponseLimitExceeded)
+def too_many_ir_requests(error=None):
+    log_error(error, 429)
+    return _render_error_page(429, template="429-ir")
```
MebinAbraham, line 78: We should avoid abbreviations.

app/views/handlers/individual_response.py
```diff
@@ -147,6 +151,26 @@ def _publish_fulfilment_request(self, mobile_number=None):
...
+            )
+
+    def _update_individual_response_count(self):
+        response_metadata = self._questionnaire_store.response_metadata
+
+        if response_metadata.get("individual_response_count"):
+            response_metadata["individual_response_count"] += 1
+        else:
+            response_metadata["individual_response_count"] = 1
+        self._questionnaire_store.save()
```
MebinAbraham, line 172: Personally, I think all of this is more appropriate as part of the `FulfilmentRequest` class.

app/views/handlers/individual_response.py
```diff
@@ -574,9 +598,11 @@ def handle_get(self):
         )
 
     def handle_post(self):
+        self._check_individual_response_count()
         if self.selected_option == self.confirm_option:
             self._update_section_status(CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED)
             self._publish_fulfilment_request()
+            self._update_individual_response_count()
```
MebinAbraham, line 605: Checking the limit should only be relevant when they are requesting another IR, as in
when `self.selected_option == self.confirm_option`.
You could then swap the order of the calls to be:
```py
self._publish_fulfilment_request()
```
where `_publish_fulfilment_request` handles all the related checks to do with limits and raise the expection as needed.

app/views/handlers/individual_response.py
LJBabbage: I had it in there to begin with, but moved it out. I didn't like the extra logic in publish_fulfilment_request. Publish to me feels like something that happens after checks have been done, I contemplated a name change but decided against it. I also felt if you had hit the limit you shouldn't be updating the status (although you could argue its wrong way round anyway) and doing a check on selected option is redundant

LJBabbage: Not sure about that one, it feels like it should be resolved sooner and not effectively under _publish_fulfilment_request which to me is a method once everything has been checked. As a kicker as soon as I put it in FulfilmentRequest I won't be able to use you test handler as I will lack app_context

MebinAbraham: I'm not against avoiding it to prevent the need for app context in tests tbh. But maybe `_publish_fulfilment_request` could handle the check to see if it is possible to publish before carrying out the action?

ajmaddaford: I agree it should be kept outside of the `FulfilmentRequest` class, @MebinAbraham's suggestion of checking in `_publish_fulfilment_request` makes sense, as that would cover both Post and Mobile options.

ajmaddaford, line 172: Calling save from this method will lead to a double save, as it's also called from the `_update_section_status` method. Saving should be done independently of both (and only if the questionnaire store has been updated).

LJBabbage: Good spot, but do we need to check the store is dirty? If it gets past the count check, the count will always go up by 
 one, so it will always be dirty

LJBabbage: Ah, something else uses it, I will make a method for just saving questionnaire store with a dirty check

app/views/handlers/individual_response.py
```diff
@@ -139,12 +143,41 @@ def get_context(self):
...
+
+        if response_metadata.get("individual_response_count"):
+            response_metadata["individual_response_count"] += 1
+        else:
+            response_metadata["individual_response_count"] = 1
+
+    def _update_questionnaire_store_on_publish(self):
+        self._update_section_status(CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED)
+        self._update_individual_response_count()
+        self._save_questionnaire_store()
```
ajmaddaford, line 175: Calling this makes the saving conditional on the section status being updated, but if we are updating the individual response count, we want to be sure we are saving. It might be better to call the save explicitly here; and remove the  `_save_questionnaire_store` method as there would only be one other method that calls it.

k8s/deploy_app.sh
```diff
@@ -24,7 +24,8 @@ helm upgrade --install \
     --set-string autoscaler.targetCPUUtilizationPercentage="${TARGET_CPU_UTILIZATION_PERCENTAGE}" \
     --set-string newRelic.enabled="${EQ_NEW_RELIC_ENABLED}" \
     --set-string newRelic.licenseKey="${NEW_RELIC_LICENSE_KEY}" \
-    --set-string newRelic.appName="${NEW_RELIC_APP_NAME}"
+    --set-string newRelic.appName="${NEW_RELIC_APP_NAME}" \
+    --set-string individualResponseLimit="${EQ_INDIVIDUAL_RESPONSE_LIMIT}"
```
ajmaddaford, line 28: This should be removed as it will break any deployments that don't set `EQ_INDIVIDUAL_RESPONSE_LIMIT`, as seen with `CONFIRMATION_EMAIL_LIMIT` recently in https://github.com/ONSdigital/eq-questionnaire-runner/pull/330.
```suggestion
    --set-string newRelic.appName="${NEW_RELIC_APP_NAME}"
```

We need to decide how we do this so that the default will be used if the environment variable isn't set.

ajmaddaford (approved):

app/views/handlers/individual_response.py
```diff
@@ -24,6 +24,10 @@
 GB_NIR_REGION_CODE = "GB-NIR"
 
 
+class IndividualResponseLimitExceeded(Exception):
+    pass
```
MebinAbraham, line 28: I think this should be updated to `IndividualResponseLimitReached` be consistent with https://github.com/ONSdigital/eq-questionnaire-runner/blob/97dd209557e66c829728eae7c43b3e736ef8ad10/app/views/handlers/confirmation_email.py#L10-L11

MebinAbraham (changes requested):
Looks good, just a minor comment.

app/views/handlers/individual_response.py
```diff
@@ -153,12 +157,37 @@ def get_context(self):
...
 
+    def _check_individual_response_count(self):
+        if (
+            self._questionnaire_store.response_metadata.get(
+                "individual_response_count", 0
+            )
+            >= current_app.config["EQ_INDIVIDUAL_RESPONSE_LIMIT"]
+        ):
+            raise IndividualResponseLimitReached(
+                "Individual response limit has been exceeded"
```
MebinAbraham, line 175:
```suggestion
                "Individual response limit has been reached"
```

MebinAbraham (approved):

PR #322, opened 2020-09-30
Add IR page titles

iwootten:

Individual response currently has missing page titles, they should be updated as per the spreadsheet in the card.

### How to review 
Use `test_individual_response.json` Ensure the page titles now match those in the spreadsheet.

### Checklist


iwootten:
[Add page titles when requesting an individual response (M)](https://trello.com/c/DwhIFGyw/4182-add-page-titles-when-requesting-an-individual-response-m)

ajmaddaford (changes requested):
We already do person number substitution in for Question and Interstitial pages. I understand that that won't work for this as it relies on the questionnaire schema, but these changes should be consistent. 

- The placeholders should use `{}` rather than `%()s`
- The placeholder in the string should be `{list_item_position}`
- The interpolation should be handled by a common method, like `_set_page_title` does in the block handler

app/views/handlers/individual_response.py
```diff
@@ -73,24 +73,34 @@ def __init__(
...
-    def _is_location_valid(self):
-        self._list_name = self._schema.get_individual_response_list()
+    @cached_property
+    def _list_model(self):
+        return self._questionnaire_store.list_store[self._list_name]
 
-        list_model = self._questionnaire_store.list_store[self._list_name]
+    @cached_property
+    def _list_item_position(self):
+        return self._list_model.index(self._list_item_id)
```
ajmaddaford, line 86: The person number is off by one. Using the list store `list_item_position` method would fix this.

app/views/handlers/individual_response.py
```diff
@@ -73,24 +73,34 @@ def __init__(
...
+
+    def get_full_page_title(self, page_title):
+        if self._list_item_id:
+            page_title += ": Person {list_item_position}".format(
```
ajmaddaford, line 90: The separator (`: `) should be separate from the rest of the string, and `Person {list_item_position}` needs to be wrapped in a `lazy_gettext` call.

JamesGardiner: Done in 8e5d7c91

app/views/handlers/individual_response.py
JamesGardiner: Done in 8e5d7c91

app/views/handlers/individual_response.py
```diff
@@ -73,24 +73,38 @@ def __init__(
...
+    @cached_property
+    def _list_model(self):
+        return self._questionnaire_store.list_store[self._list_name]
 
-        list_model = self._questionnaire_store.list_store[self._list_name]
+    @cached_property
+    def _list_item_position(self):
+        return self._questionnaire_store.list_store.list_item_position(
+            self._list_item_id
+        )
```
ajmaddaford, line 88: `list_item_position` takes two arguments - the list name and the list item id.

JamesGardiner: yeah just realised this failed - fixed in latest push

ajmaddaford (approved):

app/routes/individual_response.py
```diff
@@ -250,4 +254,7 @@ def individual_response_text_message_confirmation(schema, questionnaire_store):
     return render_template(
         template="individual_response/confirmation-text-message",
         mobile_number=mobile_number,
+        page_title=individual_response_handler.get_full_page_title(
```
LJBabbage, line 257: Not sure get_full_page_title adds much for me, in the other handlers we do similar, but we don't call it full_page_title

LJBabbage (changes requested):
Should we not have a test for this as well?

app/routes/individual_response.py
JamesGardiner: are you suggesting just calling it `page_title`?

LJBabbage: yep

JamesGardiner: Done in 864439b9

LJBabbage (approved):
Just needs formatting

PR #324, opened 2020-09-30
Add IR failure template and handle failure

JamesGardiner:
Handles individual response publish failures by returning the individual response error template using a 500 error code.


JamesGardiner:
[[H] Individual response request unhappy path (M)](https://trello.com/c/1FgOkpti/4131-h-individual-response-request-unhappy-path-m)

app/views/handlers/individual_response.py
```diff
@@ -843,7 +856,19 @@ def handle_get(self):
...
+            except PublicationFailed:
+                retry_url = url_for(
+                    "individual_response.individual_response_text_message_confirm",
+                    list_item_id=self._list_item_id,
+                    **self._request_args,
+                )
+                return (
+                    render_template("errors/individual_response", retry_url=retry_url),
+                    500,
+                )
```
ajmaddaford, line 870: I think we need to decide a consistent approach for rendering error pages on exceptions. In #318 this is handled in `errors.py`.

app/views/handlers/individual_response.py
```diff
@@ -574,7 +575,19 @@ def handle_get(self):
...
+            except PublicationFailed:
+                retry_url = url_for(
+                    "individual_response.individual_response_post_address_confirm",
+                    list_item_id=self._list_item_id,
+                    **self._request_args,
+                )
+                return (
+                    render_template("errors/individual_response", retry_url=retry_url),
+                    500,
+                )
```
MebinAbraham, line 589: If the try except was moved into `_publish_fulfilment_request` you could avoid the duplication in the `handle_post` for mobile and post. You can work out the retry URL from the fact that you have a mobile number etc. The method name might then be more appropriate as `handle_fulfiilment_request` or similar.

app/views/handlers/individual_response.py
JamesGardiner: I actually prefer the approach in #318 - it's at the least consistent with what we currently do

app/views/handlers/individual_response.py
JamesGardiner: not sure about `handle_fulfiilment_request` - I think I prefer `handle_fulfilment_request` 😛

JamesGardiner: Actually, this gets uglier when you consider the handle_post then has to check the return value of the fulfilment request method to determine if it has failed or not - in this instance I think it is better to be explicit

MebinAbraham: It doesn't if you are going to redirect to the new page on the exception?

JamesGardiner: This disappears in any case if I follow the approach in #318, as I can handle the logic in the registered error handler.

I'm  making this change now btw, so let me know what you prefer when I make the push

JamesGardiner: Updated in f18f3317 - I've created a new Exception class,  `FulfilmentRequestFailedException`, to handle this apart from generic PubSub PublicationFailed errors.

templates/errors/fulfilment_request.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'layouts/_base.html' %}
+
+{% set page_title = _("Sorry, there was a problem sending the access code") %}
+
+{% block main %}
+<h1>{{ _("Sorry, there was a problem sending the access code") }}</h1>
+<p>{{ _("You can try to <a href='{retry_url}'>request a new access code again</a>.".format(retry_url = retry_url)) }}</p>
+<p>{{ _("If this problem keeps happening, please contact us for help.") }}</p>
+<p>{{ _("<a href='{url}'>Contact us</a> if you need to speak to someone about your survey.").format(url = contact_us_url)  }}</p>
```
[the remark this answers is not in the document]
JamesGardiner: I've requested feedback from Donna and Laura C on this

tests/integration/individual_response/test_individual_response.py
```diff
@@ -211,6 +231,35 @@ def test_404_individual_response_no_list_items(self):
...
+
+        # Given I add a household member
+        self._add_household_no_primary()
+        self._request_individual_response_by_post()
+        self.assertEqualPageTitle(
+            "Sorry, there was a problem sending the access code - Census 2021"
+        )
+        self.assertInSelectorCSS(
+            self.last_url, "a", href=re.compile("individual-response")
+        )
```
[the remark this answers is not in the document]
JamesGardiner: Shouldn't the retry url be the same as the last url? That's what I am testing with the above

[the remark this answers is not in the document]
ajmaddaford: Oh I see now, but that isn't clear. I think it's confusing to search for something using part of the url we're testing for, and also makes the test fragile as it could return another unrelated link. In other templates we use a `data-qa` attribute to select elements directly. As the link is within a string that will be translated, we should consider setting this on the `p` tag (to keep the translatable string as simple as possible), and select the `a` element within that.

app/routes/errors.py
```diff
@@ -37,13 +37,13 @@ def log_error(error, status_code):
...
 
-def _render_error_page(status_code, template=None):
+def _render_error_page(status_code, template=None, **kwargs):
     handle_language()
     template = template or status_code
     using_edge = request.user_agent.browser == "edge"
 
     return (
-        render_template(template=f"errors/{template}", using_edge=using_edge),
+        render_template(template=f"errors/{template}", using_edge=using_edge, **kwargs),
```
ajmaddaford, line 46: I realise this wasn't added in this pull request, but `using_edge` is no longer used and should be removed (https://github.com/ONSdigital/eq-questionnaire-runner/commit/d7079f94d03c5aa0c2ffb5823faa03fe1e8412ef#diff-b5a180a4c7c2be4b2d9e6472ed965d80).

JamesGardiner: I meant to bring that up with you yesterday - I'll remove it now

templates/errors/fulfilment_request.html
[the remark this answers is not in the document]
JamesGardiner: This has been confirmed as correct

tests/integration/individual_response/test_individual_response.py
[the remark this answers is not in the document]
JamesGardiner: I'll add an attribute to the `p` element of `data-qa=contact-us`

app/routes/errors.py
JamesGardiner: done in dd3818d

tests/integration/individual_response/test_individual_response.py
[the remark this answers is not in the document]
JamesGardiner: done in dd3818d

ajmaddaford (approved):

app/routes/errors.py
```diff
@@ -92,3 +93,27 @@ def internal_server_error(error=None):
 def http_exception(error):
     log_error(error, error.code)
     return _render_error_page(error.code)
+
+
+@errors_blueprint.app_errorhandler(FulfilmentRequestFailedException)
```
MebinAbraham, line 98: This could handle `PublicationFailed` directly and avoid an extra custom exception.

app/views/handlers/individual_response.py
```diff
@@ -155,9 +160,12 @@ def get_context(self):
...
         fulfilment_request = FulfilmentRequest(self._metadata, mobile_number)
-        return current_app.eq["publisher"].publish(
-            topic_id, message=fulfilment_request.payload
-        )
+        try:
+            return current_app.eq["publisher"].publish(
+                topic_id, message=fulfilment_request.payload
+            )
+        except PublicationFailed:
+            raise FulfilmentRequestFailedException
```
MebinAbraham, line 168: We already raise our custom `PublicationFailed`, so why do we raise another custom Exception?. We could just make the handler support `PublicationFailed` then this try expect isn't needed. Thoughts?

MebinAbraham (changes requested):

app/views/handlers/individual_response.py
JamesGardiner: Because I didn't want to use the generic PublicationFailed to execute Individual Response code. It seems to me that the pattern for publish should be Publish -> Error -> Raise PublicationFailed, allow the calling method to handle that in it's own way. Any future pub sub methods aren't then tied to the Individual Response fulfilment requests

MebinAbraham: Ya good point. 👍

MebinAbraham (approved):

PR #325, opened 2020-09-30
Ensure middle name exists when formatting name

iwootten:

Resolved an issue where the format_name transform incorrectly displayed 'None' for middle names when peoples first and last names matched.

### How to review 
Using `test_list_collector_same_name_items`, enter 2 people with the same name but don't enter a middle name for either. Go to the list summary and see the result.


iwootten:
[Same name differentiation returns None if middle name is missing](https://trello.com/c/UDAR7poX/4185-same-name-differentiation-returns-none-if-middle-name-is-missing)

MebinAbraham (approved):
Works as expected,

ajmaddaford (approved):

PR #329, opened 2020-10-01
Update Individual Response button on Interstitial page

berroar:
The `Request separate census` button on the Individual Response interstitial page should be a '[Call to action](https://ons-design-system.netlify.app/components/button/#call-to-action)' button with chevron, using the `button` type anchor link with URL.

### How to review 
Use the `test_individual_response` schema to check the change

### Checklist


berroar:
[Change call to action on /individual-response/ launcher from form sumbit button to anchor link](https://trello.com/c/DwnsWQdo/4188-change-call-to-action-on-individual-response-launcher-from-form-sumbit-button-to-anchor-link)

templates/individual_response/interstitial.html
```diff
@@ -17,6 +17,7 @@ <h1 class="u-mt-m">{{_("If you can't answer questions for others in your househo
         onsButton({
             "text": _("Request separate census"),
             "classes": 'btn u-mb-m u-mt-s btn--link',
+            "url": "",
```
ajmaddaford, line 20: The chevron is missing from the button the as the `url` property is empty. For this to work we need to remove the `POST` from the endpoint and on a GET request determine the URL for the next page. That URL should be used here, and will be whatever we currently redirect to on the POST request.

app/views/handlers/individual_response.py
```diff
@@ -150,6 +150,24 @@ def handle_get(self):
             self.individual_section_id
         )
 
+        self._list_name = self._schema.get_individual_response_list()
```
ajmaddaford, line 153: `self._list_name` is already set in the init.
```suggestion
```

app/views/handlers/individual_response.py
```diff
@@ -150,6 +150,24 @@ def handle_get(self):
...
+            )
+        elif len(list_model.non_primary_people) == 1:
+            next_location_url = url_for(
+                ".individual_response_how",
+                list_item_id=list_model.non_primary_people[0],
+                journey="hub",
+            )
+        else:
+            next_location_url = url_for(".individual_response_who", journey="hub")
+
```
ajmaddaford, line 170: There is a lot going on in this method now. I think we should create `get_previous_location_url` and `get_next_location_url` methods, and call them from here.

tests/integration/individual_response/test_individual_response.py
```diff
@@ -21,6 +21,10 @@ def individual_response_link(self):
         if response_paragraph:
             return response_paragraph.find_next()["href"]
 
+    def get_next_location_url_from_interstitial_page(self):
+        submit_button = self.getHtmlSoup().find("a", {"data-qa": "btn-submit"})
+        return submit_button.attrs["href"]
```
ajmaddaford, line 26: This method should be consistent with the `individual_section_link` and `individual_response_link` methods. It should be defined as a `@property` and named something like `individual_response_start_link`.

ajmaddaford (changes requested):

app/views/handlers/individual_response.py
```diff
@@ -160,59 +160,56 @@ def _publish_fulfilment_request(self, mobile_number=None):
...
-        if len(list_model.non_primary_people) == 1:
-            return redirect(
-                url_for(
-                    ".individual_response_how",
-                    list_item_id=list_model.non_primary_people[0],
-                    journey="hub",
-                )
-            )
-        return redirect(url_for(".individual_response_who", journey="hub"))
+            return url_for("questionnaire.get_questionnaire")
```
ajmaddaford, line 212: Same as the comment above about the `elif` and `else`.

app/views/handlers/individual_response.py
```diff
@@ -160,59 +160,56 @@ def _publish_fulfilment_request(self, mobile_number=None):
...
+                list_item_id=list_model.non_primary_people[0],
+                journey="hub",
+            )
+        else:
+            return url_for(".individual_response_who", journey="hub")
+
+    def _get_previous_location_url(self):
         individual_section_first_block_id = self._schema.get_first_block_id_for_section(
             self.individual_section_id
         )
```
ajmaddaford, line 194: As this is only used in the second conditional, these lines should be moved there.

ajmaddaford (changes requested):

ajmaddaford (approved):

LJBabbage (approved):
The moving around of stuff confusion me a touch, but it looks ok and works as expected

petechd (approved):

PR #334, opened 2020-10-06
Add same name functionality to individual response

ajmaddaford:
- Implements the same name functionality for the individual response pages:
  - For the "who" page, middle names are shown for all householders where two or more people who share the same first and last names
  - All other individual response pages only show middles names for those who share the same first and last names with someone else in the household

- Fixes a bug was fixed that stopped duplicate names (when two or more people share **all** names) appearing on the "who" page 🐛 .

- The way `block_definition` works in the individual response handler has been refactored. As the definition needed to pass the list name to the placeholder transforms, it has been converted into a method. This has the added advantage that it has removed the unnecessary init duplication across most of the individual response handlers 🙂 .

### How to review 
- Does it work with different combinations of first / middle / last names?
- Do the tests cover all scenarios?

### Checklist

* [-] New static content marked up for translation
* [-] Newly defined schema content included in eq-translations repo


MebinAbraham (approved):
👍

berroar (approved):
Tested the changes, all seem to work as expected

PR #335, opened 2020-10-06
Escape html in rendered composite answers

petechd:
This adds dictionary escaping method that checks composite answer (address) for a question and summary screen. Change is described on [this Trello card](https://trello.com/c/QeNTMaSQ).

### How to review 
To recreate the problem use test_address.json and enter HTML in the first line. Check if on schema's confirmation question and summary screen address is escaped.

### Checklist


petechd:
[HTML in composite answers (S)](https://trello.com/c/QeNTMaSQ/4173-html-in-composite-answers-s)

app/questionnaire/placeholder_parser.py
```diff
@@ -52,9 +52,15 @@ def _lookup_answer(self, answer_id: str, list_item_id: str = None):
...
                     for list_item in answer.value
                 ]
-            return (
-                escape(answer.value) if isinstance(answer.value, str) else answer.value
-            )
+
+            if isinstance(answer.value, str):
+                return escape(answer.value)
+            elif isinstance(answer.value, dict):
+                answer_dict = {}
```
berroar, line 59: Do we need to create a new dict here? Could we update the existing dict instead?

app/views/contexts/summary/question.py
```diff
@@ -22,6 +22,14 @@ def _get_answer(self, answer_store, answer_id):
...
             if isinstance(answer.value, str):
                 return escape(answer.value)
+            elif isinstance(answer.value, dict):
+                answer_dict = {}
+                for answer_field in answer.value:
+                    if isinstance(answer.value[answer_field], str):
+                        answer_dict[answer_field] = escape(answer.value[answer_field])
+                    else:
+                        answer_dict[answer_field] = answer.value[answer_field]
+                return answer_dict
```
ajmaddaford: The dictionary `items` method can be used to make this a little clearer:
```suggestion
                escaped_answer = {}
                for key, value in answer.value.items():
                    escaped_answer[key] = escape(value) if isinstance(value, str) else value
                return escaped_answer
```

app/questionnaire/placeholder_parser.py
petechd: Removed extra dicts in both methods, thanks!

app/questionnaire/placeholder_parser.py
```diff
@@ -52,9 +52,15 @@ def _lookup_answer(self, answer_id: str, list_item_id: str = None):
                     else answer.value
                     for list_item in answer.value
                 ]
-            return (
-                escape(answer.value) if isinstance(answer.value, str) else answer.value
-            )
+
+            if isinstance(answer.value, str):
+                return escape(answer.value)
+            elif isinstance(answer.value, dict):
```
ajmaddaford, line 58: This can be an `if` rather than an `elif` as the preceding `if` has a return.

ajmaddaford (changes requested):
The two approaches differ - one escapes only strings in dicts (which is correct), the other escapes everything. To reduce the duplication and ensure consistent escaped answers, I think we should move the functionality to a `get_escaped_answer` method on the `AnswerStore`. This could call the existing `get_answer` method and escape the values before returning. 

Tests also need to be added.

app/questionnaire/placeholder_parser.py
petechd: Changed to `if`, thanks!

app/views/contexts/summary/question.py
petechd: Done, thanks!

app/data_models/answer_store.py
```diff
@@ -134,3 +136,30 @@ def remove_all_answers_for_list_item_id(self, list_item_id: str):
 
     def serialize(self):
         return list(self.answer_map.values())
+
+    def get_escaped_answer(self, answer_id, list_item_id):
```
LJBabbage, line 140: Maybe a bit neater, I am not a huge fan of over using the same var. I also managed to get a walrus in before James suggested it :)

```
          if answer := self.get_answer(answer_id, list_item_id):
            if isinstance(answer.value, list):
                escaped_list = []
                for list_item in answer.value:
                    escaped_list.append(escape(list_item)) if isinstance(
                        list_item, str
                    ) else escaped_list.append(list_item)
                return escaped_list

            if isinstance(answer.value, dict):
                escaped_dict = {}
                for key, value in answer.value.items():
                    escaped_dict[key] = (
                        escape(value) if isinstance(value, str) else value
                    )
                return escaped_dict

            return (
                escape(answer.value) if isinstance(answer.value, str) else answer.value
            )

        return None

```

ajmaddaford, line 140:
```suggestion
    def get_escaped_answer_value(self, answer_id, list_item_id):
```

app/data_models/answer_store.py
```diff
@@ -134,3 +136,30 @@ def remove_all_answers_for_list_item_id(self, list_item_id: str):
...
+        answer = self.get_answer(answer_id, list_item_id)
+        if answer:
```
ajmaddaford:
```suggestion
        if answer := self.get_answer(answer_id, list_item_id):
```

LJBabbage (changes requested):

app/data_models/answer_store.py
petechd: Wow, this is compacted!

app/data_models/answer_store.py
```diff
@@ -134,3 +136,27 @@ def remove_all_answers_for_list_item_id(self, list_item_id: str):
...
+
+    def get_escaped_answer(self, answer_id, list_item_id):
+        if answer := self.get_answer(answer_id, list_item_id):
+            if isinstance(answer.value, list):
+                escaped_list = []
+                for list_item in answer.value:
+                    escaped_list.append(escape(list_item)) if isinstance(
+                        list_item, str
+                    ) else escaped_list.append(list_item)
+                return escaped_list
```
ajmaddaford, line 148: The original version used list comprehension to avoid the need for a variable:
```suggestion
                return [
                    escape(list_item)
                    if list_item and isinstance(list_item, str)
                    else answer.value
                    for list_item in answer.value
                ]
```

ajmaddaford (commented):
Also, tests for the new method should be added to `test_answer_store.py`.

petechd:
> Also, tests for the new method should be added to `test_answer_store.py`.

Added some tests and addressed Andrew's comments.

app/data_models/answer_store.py
petechd: Done, thanks!

app/data_models/answer_store.py
petechd: Renamed, thanks!

app/data_models/answer_store.py
```diff
@@ -137,15 +137,15 @@ def remove_all_answers_for_list_item_id(self, list_item_id: str):
     def serialize(self):
         return list(self.answer_map.values())
 
-    def get_escaped_answer(self, answer_id, list_item_id):
+    def get_escaped_answer_value(self, answer_id, list_item_id):
```
ajmaddaford, line 140: I think `list_item_id` should be optional (default to `None`), to be consistent with the `get_answer` method.

app/questionnaire/placeholder_parser.py
```diff
@@ -41,7 +41,7 @@ def __call__(self, placeholder_list: Sequence[Mapping]) -> Mapping:
         return self._placeholder_map
 
     def _lookup_answer(self, answer_id: str, list_item_id: str = None):
-        return self._answer_store.get_escaped_answer(answer_id, list_item_id)
+        return self._answer_store.get_escaped_answer_value(answer_id, list_item_id)
```
ajmaddaford, line 44: If the `list_item_id` defaults to `None`, the `_lookup_answer` method is redundant.

tests/app/data_model/test_answer_store.py
```diff
@@ -28,6 +28,20 @@ def basic_answer_store():
...
+    answer_store.add_or_update(
+        Answer(answer_id="answer5", value=["<p>abc123</p>", 1], list_item_id="xyz987")
+    )
+    answer_store.add_or_update(
+        Answer(
+            answer_id="answer6",
+            value={"item1": "<p>abc123</p>", "item2": 1},
+            list_item_id="xyz987",
+        )
+    )
```
ajmaddaford, line 43: If `list_item_id` is not a required param for `get_escaped_answer_value`, it can be dropped from the tests.

petechd: Done, thanks!

app/questionnaire/placeholder_parser.py
petechd: Removed the whole `_lookup_answer` method, thanks!

app/data_models/answer_store.py
petechd: Used ` list_item_id=None` here, I hope that's what you wanted, thanks!

app/data_models/answer_store.py
JamesGardiner: Now if only we weren't type checking in if statements 😢

ajmaddaford (approved):

LJBabbage (approved):

PR #338, opened 2020-10-07
Add uwsgi http-keepalive and http-timeout flags

JamesGardiner:

### How to review 

### Checklist


uwsgi.ini
```diff
@@ -1,12 +1,17 @@
...
+http-keepalive=1
+http-timeout=90
+http-connect-timeout=5
 module = application:application
 buffer-size = 32768
 
 strict = true
 master = true
 need-app = true
+single-interpreter = true
```
MebinAbraham, line 13: Can now remove `--single-interpreter` from `run_app.sh`?

uwsgi.ini
```diff
@@ -1,12 +1,17 @@
 [uwsgi]
+; Set up uwsgi HTTP router on the specified address with HTTP
```
ajmaddaford, line 2: I don't think this comment adds anything.

uwsgi.ini
```diff
@@ -1,12 +1,17 @@
...
 http = :5000
+http-keepalive=1
+http-timeout=90
```
ajmaddaford, line 5: Why 90? The GCP load balancer timeout is 30 seconds, so this should only be a bit longer than that (something like 35 seconds).

uwsgi.ini
```diff
@@ -1,12 +1,17 @@
 [uwsgi]
+; Set up uwsgi HTTP router on the specified address with HTTP
 http = :5000
+http-keepalive=1
```
ajmaddaford, line 4: We should add a comment that this enables `http-keepalive`, rather than sets a duration of 1.

JamesGardiner: I don't think that's correct, it looks like it *is* setting the duration https://github.com/unbit/uwsgi/blob/uwsgi-2.0/plugins/http/http.c#L658

ajmaddaford: 🤔  That line you've linked to implies it only does something when it is greater than 1..

The assumption that it was a boolean came from reading discussions like this - https://github.com/unbit/uwsgi/issues/2018.

If it is the timeout in seconds, it should use the `HTTP_KEEP_ALIVE` environment variable to be consistent with the Gunicorn implementation.

run_app.sh
```diff
@@ -7,11 +7,11 @@ if [ "$WEB_SERVER_TYPE" = "gunicorn-async" ]; then
...
     run_command="gunicorn application:application --worker-class gthread"
 elif [ "$WEB_SERVER_TYPE" = "uwsgi" ]; then
-    run_command="uwsgi uwsgi.ini --workers ${WEB_SERVER_WORKERS}"
+    run_command="uwsgi uwsgi.ini --workers ${WEB_SERVER_WORKERS} --http-keepalive=${HTTP_KEEP_ALIVE}"
 elif [ "$WEB_SERVER_TYPE" = "uwsgi-threads" ]; then
-    run_command="uwsgi uwsgi.ini --workers ${WEB_SERVER_WORKERS} --enable-threads --threads ${WEB_SERVER_THREADS}"
+    run_command="uwsgi uwsgi.ini --workers ${WEB_SERVER_WORKERS} --enable-threads --threads ${WEB_SERVER_THREADS} --http-keepalive=${HTTP_KEEP_ALIVE}"
 elif [ "$WEB_SERVER_TYPE" = "uwsgi-async" ]; then
-    run_command="uwsgi uwsgi.ini --module patched:application --workers ${WEB_SERVER_WORKERS} --single-interpreter --gevent ${WEB_SERVER_UWSGI_ASYNC_CORES}"
+    run_command="uwsgi uwsgi.ini --module patched:application --workers ${WEB_SERVER_WORKERS} --gevent ${WEB_SERVER_UWSGI_ASYNC_CORES}"
```
ajmaddaford, line 14: Does this not need the `http-keepalive` set?

ajmaddaford (commented):
The pull request title needs updating, and also the commit message when this merges.

ajmaddaford (approved):

PR #354, opened 2020-10-19
Remove 'to continue' from default error messages

berroar:
Remove `to continue` from all default error messages.

Also updates the address lookup template validation message from 'Select an address' to 'Select or manually enter an address'

### How to review 
Check nothing is broken by this change/tests still pass

### Checklist

### DIFF
--- a/app/forms/error_messages.py
+++ b/app/forms/error_messages.py
@@ -8,20 +8,20 @@ from flask_babel import lazy_gettext
 # validators - using `.format` for non-WTForm field validation would introduce
 # inconsistency.
 error_messages = {
-    "MANDATORY_QUESTION": lazy_gettext("Enter an answer to continue"),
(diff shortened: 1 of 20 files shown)


berroar:
[Validation message changes (S)](https://trello.com/c/QNzkjl9z/4235-validation-message-change)

LJBabbage (changes requested):
What are we doing with the other .po files (eo, cy and ga)? At the moment they don't line up and naturally we can't update the key as the translation will be wrong

berroar:
> What are we doing with the other .po files (eo, cy and ga)? At the moment they don't line up

Hadn't really thought about those as I normally just run`make translation-templates` which just updates this .po file, but I think there are cards in the feature backlog to generate/upload all the translations files

LJBabbage (approved):

PR #355, opened 2020-10-19
Upgrade deprecated k8s apis

JamesGardiner:
Upgrades the deprecated Deployment and Ingress k8s apis

### How to review 
Using a pipeline, target this branch of the runner git repository for deployment to the relevant GKE cluster. Once deployed, run load against the environment to trigger an autoscaling event. Also, use `kubectl get hpa` and assert that the autoscaling numerator is not `<unknown>`.

It would also be beneficial to attempt to upgrade an existing Deployment and Ingress api version, as this more closely replicates the upgrade path for our dev and formal environments. Helm will not upgrade the existing Ingress, so delete it, then attempt to upgrade.

### Checklist

**Marked as dont merge due to this breaking CI**


k8s/helm/templates/ingress.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.ingress.enabled -}}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
```
MebinAbraham, line 2: Any reason this is still using the beta and not `networking.k8s.io/v1`?.

MebinAbraham (commented):
Both new and existing env worked as expected. For existing env, the current ingress has to be deleted but I think it might be better to delete the ingress and the deployment and do a fresh deploy to ensure there are no consequences on residual config

k8s/helm/templates/ingress.yaml
JamesGardiner: As per the k8s docs:
```
The v1.22 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:

Ingress in the extensions/v1beta1 API version will no longer be served
Migrate to use the networking.k8s.io/v1beta1 API version, available since v1.14. Existing persisted data can be retrieved/updated via the new version.
```

We don't actually _need_ to upgrade now (it isn't deprecated till `v1.22`, but we might as well do it while we aren't running the prod service?

MebinAbraham: Given we are updating it here, I think we should only use the beta if we can't use `v1` which I don't think is the case.

JamesGardiner: As far as I can see, the docs say to use v1beta1 for ingress - is there other documentation that states otherwise?

JamesGardiner: v1 is not available until our cluster is on 1.19

MebinAbraham (approved):

pricem14pc (approved):
Manually deleted Runner Ingres via the console
New Runner Ingres successfully created on deployment
Ran a performance test against new deployment
Application autoscaled up and down as expected at 50% CPU

PR #356, opened 2020-10-20
Fix change/remove links on section summary

petechd:
This fixes the issue of change and remove links returning wrong edit-screen url address (`people-living-here-edit-person` instead of `any-more-people-living-here-edit-person`).

### How to review 
To recreate/check this issue don't add primary person and anyone on the first list collector. After moving to section summary try to click on edit link of a person you've added on the second list collector. Before this issue was fixed runner used to redirect to edit screen using incorrect url address/routing (`people-living-here-edit-person` instead of `any-more-people-living-here-edit-person`). New test and test schema were added.

### Checklist


petechd:
[P1: Change & remove links not working HH schemas](https://trello.com/c/p0MUaKS2/4229-p1-change-remove-links-not-working-hh-schemas)

app/views/contexts/section_summary_context.py
```diff
@@ -122,10 +122,25 @@ def _custom_summary_elements(self, section_summary, current_location, section):
...
         )
 
+        list_collector_blocks_on_path = [
+            list_collector_block
+            for list_collector_block in self._schema.get_list_collectors_for_list(
+                section, for_list=summary["for_list"]
+            )
+            if list_collector_block["id"] in routing_path.block_ids
+        ]
+        if len(list_collector_blocks_on_path) == 0:
```
JamesGardiner, line 137:
```suggestion
        if not list_collector_blocks_on_path:
```

tests/integration/questionnaire/test_questionnaire_list_collector_primary_person.py
```diff
@@ -131,3 +131,36 @@ def test_changing_answer_from_no_to_yes_on_primary_person_list_collector_resumes
...
+
+        self.post({"another-anyone-else": "No"})
+
+        self.assertInUrl("/questionnaire/sections/section/")
+
+        first_person_change_link = self.get_link(1, "Change")
+
+        self.get(first_person_change_link)
+
+        self.assertInBody("Change details for Marie Day")
```
JamesGardiner, line 166: This also needs the remove link tested

app/views/contexts/section_summary_context.py
```diff
@@ -122,10 +122,25 @@ def _custom_summary_elements(self, section_summary, current_location, section):
 
     def _list_summary_element(self, summary, current_location, section) -> Mapping:
         current_list = self._list_store[summary["for_list"]]
-        list_collector_block = self._schema.get_list_collector_for_list(
-            section, for_list=summary["for_list"]
+
+        routing_path = self._router.routing_path(
```
JamesGardiner, line 126: Now that we get the routing path here, we can change the method signature for the `_add_link` method so that it takes a routing path - it will save us accessing it twice

JamesGardiner (changes requested):

app/views/contexts/section_summary_context.py
petechd: Done, thanks!

app/views/contexts/section_summary_context.py
petechd: Moved routing_path and current_location attributes out so add_link is a static method now. Thanks!

tests/integration/questionnaire/test_questionnaire_list_collector_primary_person.py
petechd: Added remove test so it should be all covered now👍

app/views/contexts/section_summary_context.py
```diff
@@ -122,10 +122,25 @@ def _custom_summary_elements(self, section_summary, current_location, section):
...
         )
 
+        list_collector_blocks_on_path = [
+            list_collector_block
+            for list_collector_block in self._schema.get_list_collectors_for_list(
+                section, for_list=summary["for_list"]
+            )
+            if list_collector_block["id"] in routing_path.block_ids
+        ]
+        if not list_collector_blocks_on_path:
```
JamesGardiner, line 137: Just for readability I'd probably add a blank line here

JamesGardiner (approved):

app/views/contexts/section_summary_context.py
```diff
@@ -122,10 +122,26 @@ def _custom_summary_elements(self, section_summary, current_location, section):
...
+            )
+            if list_collector_block["id"] in routing_path.block_ids
+        ]
+
+        if not list_collector_blocks_on_path:
+            list_collector_block = self._schema.get_list_collector_for_list(
+                section, for_list=summary["for_list"]
+            )
+        else:
+            list_collector_block = list_collector_blocks_on_path[0]
```
MebinAbraham, line 143: This evaluates both sides, so I don't see a reason to check the negative side first. Also, this is potentially fetching the list twice when nothing is on the path. But ultimately, if none of the list collectors are on the path then calling 
```py
			self._schema.get_list_collector_for_list(
                section, for_list=summary["for_list"]
            )
```
won't help as that is already not on the path. If none of the collectors are on the path, then you should fetch the driving question.

test_schemas/en/test_list_collector_primary_and_two_list_collectors.json
```diff
@@ -0,0 +1,518 @@
+{
```
MebinAbraham, line 1: Primary person is not relevant to this scenario. The scenario we need to create is, when you have a list collector and the list collector is not on the path. To do this, you do not need a primary person. I think the only thing you need is a driving question and a list collector.

app/views/contexts/section_summary_context.py
ajmaddaford: I can see this was already being done (incorrectly) for the add link, but a routing path should not be being generated within a context; and doing so here will unnecessarily generate a routing path multiple times. Instead, the routing path should be passed to the context.

test_schemas/en/test_list_collector_primary_and_two_list_collectors.json
JamesGardiner: Isn't the scenario different to this though? It should be:

1) One list
2) Two list collectors, each with their own driving questions
3) Nothing added to list on 1st list collector
4) Item added to list on 2nd list collector
5) Change/remove links work for that list item

MebinAbraham: Umm not sure, is it not simply ensuring we generate a link based on the list collector that is available or driving question if list collector is not available. Having 2 list collectorws or 3 won't matter, it is the fact that, the available one is chosen?. Also, we don't have requirements for 2 driving questions and that prob opens up a bunch of other scenarios we haven't thought about.

app/views/contexts/section_summary_context.py
petechd: I moved it out from context so it's being passed to it now. Thanks!

test_schemas/en/test_list_collector_primary_and_two_list_collectors.json
JamesGardiner: You are correct re the need for only one ListColectorDrivingQuestion, but is there not still a need for two ListCollectors, as the issue that we have identified is that the add/remove links are being generated for the first list collector, which isn't on the path

app/views/contexts/section_summary_context.py
petechd: Refactored it so it's not negative first, waiting for James and his review before I make any more changes.

test_schemas/en/test_list_collector_primary_and_two_list_collectors.json
MebinAbraham: As discussed @JamesGardiner, the change and remove links are only relevant when there is at least 1 item on the list, which is only possible by having more than 1 list collector, therefore the driving question is not required. 

A list collector with a primary person collector is already handled I believe, so to test this scenario for change/remove links, I believe what you need is just 2 list collectors as James mentioned. (Primary person collector is not needed).


@petechd

app/views/contexts/section_summary_context.py
```diff
@@ -113,17 +116,34 @@ def _title_for_location(self, location):
...
+            if list_collector_block["id"] in routing_path.block_ids
+        ]
+
+        list_collector_block = (
+            list_collector_blocks_on_path[0]
+            if list_collector_blocks_on_path
+            else self._schema.get_list_collector_for_list(
+                section, for_list=summary["for_list"]
+            )
         )
```
MebinAbraham, line 147: This can actually be:
```py
        list_collector_block = (
            list_collector_blocks_on_path[0]
            if list_collector_blocks_on_path
            else None
        )
```
This will allow you to remove the need of routing path from `self._add_link`.

app/views/contexts/section_summary_context.py
```diff
@@ -171,10 +189,8 @@ def _list_summary_element(self, summary, current_location, section) -> Mapping:
...
 
-    def _add_link(self, summary, current_location, section, list_collector_block):
-        routing_path = self._router.routing_path(
-            section["id"], current_location.list_item_id
-        )
+    @staticmethod
+    def _add_link(summary, section, list_collector_block, routing_path):
 
         if list_collector_block["id"] in routing_path:
             return url_for(
```
MebinAbraham, line 196: Once my comment about removing routing path from this method is done, you can update this to:
```py
        if list_collector_block:
            return url_for(
```

MebinAbraham (commented):
The edit and remove links are handled by `_build_list_items_context` in `list_context.py` which still needs updating to use the collector that is on the path.

app/views/contexts/section_summary_context.py
petechd: Done, thanks!

app/views/contexts/section_summary_context.py
petechd: Done👍

test_schemas/en/test_list_collector_primary_and_two_list_collectors.json
petechd: Ok, two list collectors now, thanks!

app/views/contexts/section_summary_context.py
```diff
@@ -13,9 +13,12 @@
 
 class SectionSummaryContext(Context):
     def __call__(
-        self, current_location: Location, return_to: str = "section-summary"
+        self,
+        current_location: Location,
+        routing_path,
```
JamesGardiner, line 18: This could
```suggestion
        routing_path: RoutingPath,
```

JamesGardiner: ☝️ obvs will need to import RoutingPath for this as well

app/views/contexts/section_summary_context.py
```diff
@@ -137,46 +149,53 @@ def _list_summary_element(self, summary, current_location, section) -> Mapping:
...
-            self._progress_store,
-            self._metadata,
-        )
+        add_link = self._add_link(summary, section, list_collector_block)
 
         rendered_summary = self._placeholder_renderer.render(
             summary, current_location.list_item_id
         )
 
+        if list_collector_blocks_on_path:
```
JamesGardiner, line 159: This is the 2nd time `list_collector_blocks_on_path` is checked for it's truthiness value. An alternative would be to move the instantions:

```
edit_block_id = None
remove_block_id = None
```

to the start of the method, then override them inside the if block on line 138

JamesGardiner (changes requested):

app/views/contexts/section_summary_context.py
```diff
@@ -137,46 +156,46 @@ def _list_summary_element(self, summary, current_location, section) -> Mapping:
...
+        add_link = self._add_link(summary, section, list_collector_block)
 
         rendered_summary = self._placeholder_renderer.render(
             summary, current_location.list_item_id
         )
 
+        list_summary_context = {}
+
+        if list_collector_block:
+            list_context = ListContext(
```
MebinAbraham, line 168: The context should be conditional on the list collector block as you may still have a primary person list collector.

app/views/contexts/section_summary_context.py
petechd: I moved it up and there is only one if statement based on `list_collector_blocks_on_path` now, if only I could compact it more😆.

app/views/contexts/section_summary_context.py
JamesGardiner: What is the scenario for this, and does it need a test given the tests are passing?

app/views/contexts/section_summary_context.py
```diff
@@ -113,18 +113,37 @@ def _title_for_location(self, location):
...
-    def _list_summary_element(self, summary, current_location, section) -> Mapping:
+    def _list_summary_element(
+        self, summary, current_location, section, routing_path
+    ) -> Mapping:
         current_list = self._list_store[summary["for_list"]]
-        list_collector_block = self._schema.get_list_collector_for_list(
-            section, for_list=summary["for_list"]
-        )
+
+        list_collector_blocks_on_path = [
```
JamesGardiner, line 130: A few things here:

1) We should get a variable that contains all of the list collector blocks - `list_collector_blocks`
2) Using `list_collector_blocks`, create `list_collector_blocks_on_path`, which is already being done here
3) The test on line 167 can then be negated:
```
if not list_collector_block:
    list_collector_block = list_collector_blocks[0]
```

that way, there will always be a list context available, even when there are no list collector blocks on the path but primary person is added.

To test for the scenario, we probably need a new schema. It should have a PrimaryPersonListCollector, and a single ListCollector, with a ListCollectorDrivingQuestion. The test should determine primary person is displayed correctly on the section summary, when there is a primary person added, but no list collectors on the path (i.e. the negative ListCollectorDrivingQuestion is selected)

JamesGardiner (changes requested):

app/views/contexts/section_summary_context.py
JamesGardiner: @MebinAbraham does the above fix the issue you raised?

app/views/contexts/section_summary_context.py
MebinAbraham: As discussed, this is for when you have a primary person with no list collectors on the path. We should add a test for it but could be seperate.

app/views/contexts/section_summary_context.py
MebinAbraham: Looks good.

petechd: Test schema added with that specific scenario tested. Changes James recommended added. There is this side effect now where I needed to add edit block to list context as normal primary person edit block id is not used to resolve change link(in this new test scenario)🤔.

app/views/contexts/section_summary_context.py
petechd: Added a test and new test schema. Thanks!

MebinAbraham: This should be conditional.

app/views/contexts/section_summary_context.py
```diff
@@ -113,18 +113,42 @@ def _title_for_location(self, location):
...
             if summary_element["type"] == "List":
                 yield self._list_summary_element(
-                    summary_element, current_location, section
+                    summary_element, current_location, section, routing_path
                 )
 
-    def _list_summary_element(self, summary, current_location, section) -> Mapping:
+    def _list_summary_element(
+        self, summary, current_location, section, routing_path
+    ) -> Mapping:
```
MebinAbraham, line 127: You can remove the list comprehension and I think we should make `ListContext` a property. The add_link logic could stay as is if you fetch the first block at the end.

I think we need to split this method up as it is doing a lot of things. And I think we can do that by generating the routing path on `__init__`. At the moment, it is passed to every method, I don't see why we can generate the routing path on the initialisation of the object, then you don't have to pass the routing path around to all the different methods.


```py
    def _list_summary_element(
        self, summary, current_location, section, routing_path
    ) -> Mapping:
        list_collector_block = None
        edit_block_id, remove_block_id, primary_person_edit_block_id = None, None, None
        current_list = self._list_store[summary["for_list"]]

        list_collector_blocks = list(
            self._schema.get_list_collectors_for_list(
                section, for_list=summary["for_list"]
            )
        )

        list_collector_blocks_on_path = [
            list_collector_block
            for list_collector_block in list_collector_blocks
            if list_collector_block["id"] in routing_path.block_ids
        ]

        if list_collector_blocks_on_path:
            list_collector_block = list_collector_blocks_on_path[0]
            edit_block_id = list_collector_block["edit_block"]["id"]
            remove_block_id = list_collector_block["remove_block"]["id"]

        add_link = self._add_link(summary, section, list_collector_block)

        if len(current_list) == 1 and current_list.primary_person:
            primary_person_block = self._schema.get_list_collector_for_list(
                section, for_list=summary["for_list"], primary=True
            )

            primary_person_edit_block_id = primary_person_block["add_or_edit_block"][
                "id"
            ]
            edit_block_id = primary_person_block["add_or_edit_block"]["id"]

        rendered_summary = self._placeholder_renderer.render(
            summary, current_location.list_item_id
        )

        list_collector_block = list_collector_block or list_collector_blocks[0]

        list_context = ListContext(
            self._language,
            self._schema,
            self._answer_store,
            self._list_store,
            self._progress_store,
            self._metadata,
        )
        list_summary_context = list_context(
            return_to="section-summary",
            edit_block_id=edit_block_id,
            remove_block_id=remove_block_id,
            primary_person_edit_block_id=primary_person_edit_block_id,
        )

        return {
            "title": rendered_summary["title"],
            "type": rendered_summary["type"],
            "add_link": add_link,
            "add_link_text": rendered_summary["add_link_text"],
            "empty_list_text": rendered_summary["empty_list_text"],
            "list_name": rendered_summary["for_list"],
            **list_summary_context,
        }
```

app/views/contexts/section_summary_context.py
```diff
@@ -136,47 +160,52 @@ def _list_summary_element(self, summary, current_location, section) -> Mapping:
...
+            summary, section, list_collector_block, list_collector_blocks_on_path
         )
 
         rendered_summary = self._placeholder_renderer.render(
             summary, current_location.list_item_id
         )
 
+        list_summary_context = {}
+
+        if list_collector_block:
```
JamesGardiner, line 175:
```suggestion
```

this needs to be not conditional 👍

app/views/contexts/section_summary_context.py
```diff
@@ -13,9 +14,12 @@
...
 class SectionSummaryContext(Context):
     def __call__(
-        self, current_location: Location, return_to: str = "section-summary"
+        self,
+        current_location: Location,
+        routing_path: RoutingPath,
+        return_to: str = "section-summary",
     ) -> Mapping:
-        summary = self._build_summary(current_location, return_to)
+        summary = self._build_summary(current_location, return_to, routing_path)
```
JamesGardiner, line 22:
```suggestion
		self._routing_path = routing_path
        summary = self._build_summary(current_location, return_to)
```

app/views/contexts/section_summary_context.py
```diff
@@ -61,7 +65,7 @@ def get_page_title(
 
         return page_title
 
-    def _build_summary(self, location, return_to):
+    def _build_summary(self, location, return_to, routing_path):
```
JamesGardiner, line 68:
```suggestion
    def _build_summary(self, location, return_to):
```

app/views/contexts/section_summary_context.py
```diff
@@ -75,17 +79,13 @@ def _build_summary(self, location, return_to):
             summary_elements = {
                 "custom_summary": list(
                     self._custom_summary_elements(
-                        section["summary"]["items"], location, section
+                        section["summary"]["items"], location, section, routing_path
```
JamesGardiner, line 82:
```suggestion
                        section["summary"]["items"], location, section, self._routing_path
```

app/views/contexts/section_summary_context.py
```diff
@@ -113,18 +113,42 @@ def _title_for_location(self, location):
         )
         return title
 
-    def _custom_summary_elements(self, section_summary, current_location, section):
+    def _custom_summary_elements(
+        self, section_summary, current_location, section, routing_path
```
JamesGardiner, line 117:
```suggestion
        self, section_summary, current_location, section
```

app/views/contexts/section_summary_context.py
```diff
@@ -113,18 +113,42 @@ def _title_for_location(self, location):
...
+    ):
         for summary_element in section_summary:
             if summary_element["type"] == "List":
                 yield self._list_summary_element(
-                    summary_element, current_location, section
+                    summary_element, current_location, section, routing_path
```
JamesGardiner, line 122:
```suggestion
                    summary_element, current_location, section, self._routing_path
```

app/views/contexts/section_summary_context.py
```diff
@@ -113,18 +113,42 @@ def _title_for_location(self, location):
...
                 )
 
-    def _list_summary_element(self, summary, current_location, section) -> Mapping:
+    def _list_summary_element(
+        self, summary, current_location, section, routing_path
```
JamesGardiner, line 126:
```suggestion
        self, summary, current_location, section
```

app/views/contexts/section_summary_context.py
```diff
@@ -113,18 +113,42 @@ def _title_for_location(self, location):
...
+            list_collector_block
+            for list_collector_block in self._schema.get_list_collectors_for_list(
+                section, for_list=summary["for_list"]
+            )
+        ]
+
+        list_collector_blocks_on_path = [
+            list_collector_block
+            for list_collector_block in list_collector_blocks
+            if list_collector_block["id"] in routing_path.block_ids
```
JamesGardiner, line 140:
```suggestion
            if list_collector_block["id"] in self._routing_path.block_ids
```

tests/app/views/contexts/test_summary_context.py
```diff
@@ -541,7 +551,10 @@ def test_context_for_driving_question_summary_empty_list():
         {},
     )
 
-    context = summary_context(current_location)
+    context = summary_context(
+        current_location,
+        routing_path=RoutingPath(["anyone-usually-live-at"], section_id="section"),
```
JamesGardiner, line 556:
```suggestion
```

app/views/contexts/section_summary_context.py
petechd: This conditional has been removed with changes from previous comment.

tests/app/views/contexts/test_summary_context.py
petechd: Removing routing path will break the test as it's required positional argument.

app/views/contexts/section_summary_context.py
```diff
@@ -37,121 +64,142 @@ def __call__(
...
         }
 
-    def get_page_title(
-        self, current_location: Location, title_for_location: str
-    ) -> str:
+    @cached_property
+    def section(self):
+        return self._schema.get_section(self.current_location.section_id)
+
+    @cached_property
```
JamesGardiner, line 71:
```suggestion
    @property
```

app/views/contexts/section_summary_context.py
```diff
@@ -1,31 +1,58 @@
...
 from typing import Mapping
 
 from flask import url_for
 
 from app.helpers.template_helpers import safe_content
 from app.questionnaire import QuestionnaireSchema
 from app.questionnaire.location import Location
+from app.questionnaire.routing_path import RoutingPath
 
+from ...data_models import AnswerStore, ListStore, ProgressStore
```
JamesGardiner, line 11:
```suggestion
from app.data_models import AnswerStore, ListStore, ProgressStore
```

JamesGardiner: If you merge the above you will need to resort the imports

app/views/contexts/section_summary_context.py
petechd: Done, thanks!

app/views/contexts/section_summary_context.py
petechd: Done!

LJBabbage (approved):
Everything that Mebin has asked for has been done. I have checked it works and done a quick review and I am happy

test_schemas/en/test_list_collector_two_list_collectors.json
```diff
@@ -0,0 +1,443 @@
...
+                            "routing_rules": [
+                                {
+                                    "goto": {
+                                        "block": "list-collector"
+                                    }
+                                }
+                            ],
+                            "type": "Question"
+                        },
+                        {
```
JamesGardiner, line 86: This list collector needs a driving question, and it needs to be answered in the negative to make sure the list collector is not on the path

JamesGardiner (changes requested):

JamesGardiner (approved):

PR #358, opened 2020-10-21
Updated DS version, added mandatory property to address template

boxadesign:
The address autosuggest requires a `mandatory` property and boolean value for client side validation. 

Note - there is an issue with the manual address view validation (server side) where-by the error summary isn't rendered. This will be fixed in another PR. 

### How to review 
Check the address_lookups schema and check validation works as expected when on the lookup view.


MebinAbraham (commented):
So now if you don't enter anything the error panel is as expected however, if you type something in and submit without selecting an address, it now says `Enter an address` instead of `Select or manually enter an address`.

boxadesign:
> So now if you don't enter anything the error panel is as expected however, if you type something in and submit without selecting an address, it now says `Enter an address` instead of `Select or manually enter an address`.

image

MebinAbraham:
> > So now if you don't enter anything the error panel is as expected however, if you type something in and submit without selecting an address, it now says `Enter an address` instead of `Select or manually enter an address`.
> 
> image

Sorry I should have mentioned that it only happens if you first submit with no answer so that the current screen says "Enter an address", then you type something and submit without clicking an address.

boxadesign:
> > > So now if you don't enter anything the error panel is as expected however, if you type something in and submit without selecting an address, it now says `Enter an address` instead of `Select or manually enter an address`.
> > 
> > 
> > image
> 
> Sorry I should have mentioned that it only happens if you first submit with no answer so that the current screen says "Enter an address", then you type something and submit without clicking an address. So most likely a separate bug, unrelated to this.

This will be dealt with in the next DS release.

MebinAbraham (approved):
Error panel is displaying as expected.

ajmaddaford (approved):

PR #365, opened 2020-10-26
Add definition to interstitial

LJBabbage:
This PR adds the ability to add a definition(s) into a interstitial page

N.B As this has been added to contents in both the template and validator, it should work for anything which uses those elements. However as agreed in stand up, there are currently only tests for Interstitial (content and content variant). Equally I have only updated the functional test generate pages to reflect that

### How to review 
Using test_interstitial_definition.json make sure definitions work

### Checklist


templates/partials/definition.html
```diff
@@ -0,0 +1,33 @@
...
+      "data-ga-category": "Definition",
+      "data-ga-action": "Close panel",
+      "data-ga-label": definition_id,
+      "data-qa": "definition-"  ~ definition_id ~ "-button"
+    }
+  }
+})%}
+  {% set contents = item.definition.contents %}
+  {% include 'partials/contents.html' %}
+{% endcall %}
```
MebinAbraham, line 33: I think this partial should be called for `question-definitions.html` to reduce the duplication as they are pretty much identical. Only thing you have to do is pass the id in as a parameter.

LJBabbage: Its not just the id, data-ga-category, data-qa and the loop index would also need adding

MebinAbraham: The loop index is the same as the id. i.e `"question-definition-" ~ loop.index` is also the ID which could be passed in. 
If the ga category really needs to be "question definition" we could pass that in, it is 1 extra attribute only to make a total of 2.

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=add-definition-to-common-def
```

tests/functional/generate_pages.py
```diff
@@ -651,6 +697,10 @@ def process_block(
             append_spec_page_import(block_context, spec_file)
 
 
+def _get_definitions_in_block_contents(block_contents):
+    return [element for element in block_contents if element.get("definition")]
```
MebinAbraham, line 701: Given we don't need the return value from the lement, it would be clearer as `definition in element`
```suggestion
    return [element for element in block_contents if "definition" in element]
```

tests/functional/generate_pages.py
```diff
@@ -409,6 +427,19 @@ def process_final_summary(schema_data, page_spec, collapsible, section_summary=F
         page_spec.write(COLLAPSIBLE_SUMMARY_GETTER)
 
 
+def process_definition(definition_title, page_spec):
+    camel_case_definition_title = "".join(
+        x for x in definition_title.title() if not x.isspace()
+    )
```
MebinAbraham, line 433: I think we need to consider using the index not the title. With a proper question, the ids and element names will be extremely long.

templates/partials/contents.html
```diff
@@ -1,14 +1,23 @@
 {% for item in contents %}
+  {%- if item.definition -%}
+    {% set definition = item.definition %}
+    {% set definition_id = 'definition-' ~ definition.title.strip() | lower | replace(' ', '-') %}
```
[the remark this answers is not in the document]
LJBabbage: Not a fan of that at all, elements should be distinguishable, we should avoid indexes if at all possible. Indexes in tests are horrible and easily misread

tests/functional/generate_pages.py
LJBabbage: Meh

templates/partials/contents.html
```diff
@@ -1,14 +1,23 @@
...
+    {% set category = 'Definition' %}
```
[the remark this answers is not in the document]
LJBabbage: Because this is in contents, it doesn't have to be interstitial

templates/partials/contents.html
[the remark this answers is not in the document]
MebinAbraham: The id depends on the content of the title in this case which is not what we do with normal definitions where you can have any number of definitions. Same applies here, as you can have any number of definitions. Plus using the title will result in unnecessarily long IDs and func names.
If we don't want to use indexes because they are arrays, then each should have an "id" attribute.

tests/functional/generate_pages.py
LJBabbage: I also have other content elements in there, the index will not be concurrent

templates/partials/contents.html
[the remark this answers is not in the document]
LJBabbage: Normal question definitions are wrong, you shouldn't have id's which can't be determined. However I will changed it. Can you recommend how I can have a coherent index when I have other elements mixed in with contents like description

templates/partials/contents.html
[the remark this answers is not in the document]
MebinAbraham: Ya I guess.

templates/partials/contents.html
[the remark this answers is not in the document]
MebinAbraham: Looks like you got it. It not the nicest thing but I think it is better than generating an id by manipulating user defined text. Ideally we would have an "id" attribute as part of the object.

tests/functional/generate_pages.py
```diff
@@ -629,6 +654,21 @@ def process_block(
...
+                for variant in block["content_variants"]:
+                    block_contents = variant["content"]["contents"]
+                    definitions.extend(
+                        _get_definitions_in_block_contents(block_contents)
+                    )
+            else:
+                block_contents = block["content"].get("contents", [])
+                definitions = _get_definitions_in_block_contents(block_contents)
+
+            for i in range(len(definitions)):
```
JamesGardiner, line 669:
```suggestion
            for index, _ in enumerate(definitions):
            	process_definition(index, page_spec)
```

JamesGardiner (changes requested):

JamesGardiner:
Just the minor change request from me, apart from that looks good

templates/partials/contents.html
```diff
@@ -1,14 +1,26 @@
+{% set definition_count = namespace(value=0) %}
+
 {% for item in contents %}
+  {%- if item.definition -%}
+    {% set definition = item.definition %}
+    {% set definition_id = 'definition-' ~ definition_count.value %}
```
ajmaddaford, line 6: Could `loop.index` be used here like it is in `question-definition.html`?

tests/functional/generate_pages.py
```diff
@@ -409,6 +427,13 @@ def process_final_summary(schema_data, page_spec, collapsible, section_summary=F
         page_spec.write(COLLAPSIBLE_SUMMARY_GETTER)
 
 
+def process_definition(index, page_spec):
+    definition_context = {"definitionIndex": index}
+    page_spec.write(DEFINITION_TITLE_GETTER.substitute(definition_context))
+    page_spec.write(DEFINITION_CONTENT_GETTER.substitute(definition_context))
+    page_spec.write(DEFINITION_BUTTON_GETTER.substitute(definition_context))
```
MebinAbraham, line 434: The existing definition uses an index to get the data attributes like this:
```js
definitionTitle(definitionIndex) { return `[data-qa='question-definition-0-title']`; }
```
However, as you can see the existing one is wrong because it should actually be:
```js
definitionTitle(definitionIndex) { return `[data-qa='question-definition-${definitionIndex}-title']`; }
```
I think we should update it to do the same, so we don't have 2 ways of processing a definition. You could use the existing `Template` partial for both question definition and definition in content pages. The only thing you have to do is pass in the "id" where id is either "definition" or "question-definition". 

I have made this git gist to demo what I mean. It fixes the existing definition while make both definitions types reuse the same template.
https://gist.github.com/MebinAbraham/6449e983bb01d09f05cd8757b5acdaa7

templates/partials/contents.html
LJBabbage: Problem is I have other elements in there as well not just definition, so loop.index would look odd and would be hard to test.

ajmaddaford: Ah, I see. The value should start at 1 to be consistent with `loop.index` (which starts at 1).

test_schemas/en/test_interstitial_definition.json
```diff
@@ -0,0 +1,173 @@
+{
```
MebinAbraham, line 1: Schema name shouldn't mention interstitial since it is not specific to interstitial.

test_schemas/en/test_interstitial_definition.json
```diff
@@ -0,0 +1,173 @@
...
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "0",
+    "title": "Interstitial definition",
+    "theme": "default",
+    "description": "A questionnaire to demo a definition in an interstitial page.",
```
MebinAbraham, line 9: Contents should be updated to match the change.

tests/functional/spec/interstitial_definition.spec.js
```diff
@@ -0,0 +1,43 @@
+import InterstitialDefinitionPage from "../generated_pages/interstitial_definition/interstitial-definition.page";
+
+describe("Component: Interstitial Definition", () => {
+  describe("Given I launch the interstitial definition questionnaire", () => {
```
MebinAbraham, line 4: Same in this file, think the reference should be to contents definition as it is not specific to interstitial.

MebinAbraham (commented):
Looks good, just few minor comments about reference.

JamesGardiner (approved):

MebinAbraham (approved):

PR #368, opened 2020-10-27
Add unrelated relationships routing

ajmaddaford:
Adds routing for the "Related to anyone else?" question into the relationship router. 

### How to review 
- Add some people and choose "Unrelated" twice to see the "Related to anyone else?" question. Note that you will need at least five people as the question will only appear when there are at least two people left to form relationships with.
- Use other schemas to test relationships without the new question continue to work

### Checklist


JamesGardiner (changes requested):

app/questionnaire/relationship_router.py
```diff
@@ -2,54 +2,126 @@
...
+    def __init__(
+        self,
+        answer_store,
+        relationship_store,
+        section_id,
+        list_name,
+        list_item_ids,
+        relationships_block_id,
+        unrelated_block_id=None,
+        unrelated_answer_id=None,
```
JamesGardiner, line 17: Type hints here would be useful

MebinAbraham (changes requested):
Haven't finished the review but thought I'd post this so this can be progressed meanwhile.

app/questionnaire/relationship_router.py
```diff
@@ -2,54 +2,126 @@
...
+        current_index = path_for_individual.index(current_location)
+        return [
+            location.to_list_item_id
+            for location in path_for_individual[current_index + 1 :]
+        ]
 
-    @staticmethod
-    def _generate_relationships_routing_path(
-        section_id, block_id, list_item_ids, list_name
+    def _generate_relationships_routing_path(self):
```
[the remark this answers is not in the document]
ajmaddaford: Fair point, changed. I also updated the individual path method to be consistent.

app/questionnaire/relationship_router.py
```diff
@@ -2,54 +2,126 @@
...
+        path = []
+        for from_item in self.list_item_ids:
+            from_index = self.list_item_ids.index(from_item)
+            path += self._get_individual_relationships_routing_path(
+                from_list_item_id=self.list_item_ids[from_index],
+                to_list_item_ids=self.list_item_ids[from_index + 1 :],
+            )
```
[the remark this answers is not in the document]
ajmaddaford: Updated. I have also renamed `from_item` to `from_item_id` for clarity and replaced `self.list_item_ids[from_index]` with `from_list_item_id`.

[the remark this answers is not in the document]
ajmaddaford: Updated to use enumerate. I have also renamed `from_item` to `from_item_id` for clarity and replaced `self.list_item_ids[from_index]` with `from_list_item_id`.

You have also suggested to use `extend` rather than `+=`, is there a reason for that? I preferred `+=` as it avoided nesting brackets and was easier to read.

app/questionnaire/relationship_router.py
ajmaddaford: Done.

app/questionnaire/relationship_router.py
```diff
@@ -2,54 +2,126 @@
...
+        path_for_individual = [
+            location
+            for location in self.path
+            if location.list_item_id == current_location.list_item_id
+        ]
+        current_index = path_for_individual.index(current_location)
+        return [
+            location.to_list_item_id
+            for location in path_for_individual[current_index + 1 :]
+        ]
```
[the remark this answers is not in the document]
ajmaddaford: I thought about this scenario when writing the test, but then completely forgot about it when implementing it 🤦 .

I will refactor this to just use the list rather than the path.

app/questionnaire/relationship_router.py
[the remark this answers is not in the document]
MebinAbraham: No specify reason for `extend`, it is just what I normally use but more than happy for it to stay as `+=` 👍

app/questionnaire/relationship_router.py
JamesGardiner: Awesome 👍

tests/functional/spec/relationships-unrelated.spec.js
```diff
@@ -0,0 +1,92 @@
...
+      });
+
+      it("When I answer 'Unrelated' twice, Then I will be asked if anyone else is related with a list of the remaining people", () => {
+        $(RelationshipsPage.unrelated()).click();
+        $(RelationshipsPage.submit()).click();
+        $(RelationshipsPage.unrelated()).click();
+        $(RelationshipsPage.submit()).click();
+        expect($(RelatedToAnyoneElsePage.questionText()).getText()).to.contain("Are any of these people related to you?");
+        expect($(RelatedToAnyoneElsePage.listLabel(1)).getText()).to.equal("Daniel Davis");
+        expect($(RelatedToAnyoneElsePage.listLabel(2)).getText()).to.equal("Eve Elliot");
```
MebinAbraham, line 33: Can we assert `listLabel(3)` is none to highlight it has only rendered 2?

tests/functional/spec/relationships-unrelated.spec.js
```diff
@@ -0,0 +1,92 @@
...
+        $(RelatedToAnyoneElsePage.no()).click();
+        $(RelatedToAnyoneElsePage.submit()).click();
+        expect($(RelationshipsPage.questionText()).getText()).to.contain("Thinking of Betty Burns, Carla Clark is their");
+      });
+
+      it("When I click previous, Then I will go back to the 'related to anyone else' question for the first person", () => {
+        $(RelationshipsPage.previous()).click();
+        expect($(RelatedToAnyoneElsePage.questionText()).getText()).to.contain("Are any of these people related to you?");
+        expect($(RelatedToAnyoneElsePage.no()).isSelected()).to.be.true;
+      });
```
MebinAbraham, line 64: I know there is now an integration test for it, but we should assert that only the relevant people are listed after clicking Previous from No.

tests/integration/questionnaire/__init__.py
```diff
@@ -8,3 +8,21 @@ class QuestionnaireTestCase(IntegrationTestCase):
     def add_person(self, first_name: str, last_name: str):
         self.post({"anyone-else": "Yes"})
         self.post({"first-name": first_name, "last-name": last_name})
+        number_of_people = self.number_of_people()
+        return self.get_list_item_ids()[number_of_people - 1]
+
+    def number_of_people(self):
+        return len(self.getHtmlSoup().find_all(attrs={"data-list-item-id": True}))
```
MebinAbraham, line 15:
```suggestion
        return len(self.get_list_item_ids())
```

tests/integration/questionnaire/test_questionnaire_relationships.py
```diff
@@ -2,23 +2,43 @@
...
 
+    def test_get_first_relationship_with_resume(self):
+        self.launchSurvey("test_relationships")
+        first_list_item_id = self.add_person("Marie", "Doe")
+        second_list_item_id = self.add_person("John", "Doe")
+        self.post({"anyone-else": "No"})
+        self.get("/questionnaire/relationships?resume=True")
+        self.assertInUrl(
+            f"/questionnaire/relationships/people/{first_list_item_id}/to/{second_list_item_id}/?resume=True"
+        )
```
MebinAbraham, line 29: The issue with getting the first relation on resume is that you get a panel saying `This is the last viewed question in this section` which isn't necessarily correct.

MebinAbraham (commented):
Looks good, the story seems to be as expected, just a few minor comments.

MebinAbraham (commented):
Not a big deal but at the moment the answer to `related-to-anyone-else-answer` is persisted even when the strategy no longer applies, i.e. we remove people so that there aren't 5 people hence the answer to that question is redundant. But the question is, should it be removed?

app/views/handlers/relationships/unrelated_question.py
```diff
@@ -1,61 +1,39 @@
...
-
-        return True
-
-    def get_previous_location_url(self):
-        pass
-
-    def get_next_location_url(self):
-        pass  # pragma: no cover
+        previous_item_index = list_model.index(previous_location.to_list_item_id)
+        return list_model[previous_item_index + 1 :]
```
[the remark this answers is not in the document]
ajmaddaford: The name was a compromise between making it clear and not making it too long. Originally I had `get_remaining_relationship_to_list_item_ids_for_individual`.

I chose `remaining` as I thought of them as the remaining relationships (from the current location) as we are looping over the relationships, rather than any notion of ones that may have been answered.  We could name it something more generic like `get_list_summary_list_item_ids`, but that doesn't give any indication of how we're choosing the list item ids.

tests/functional/spec/relationships-unrelated.spec.js
ajmaddaford: We could, but my thoughts were the list is always printed in order, so if we're checking that the first person in the list is "Daniel Davis", then we're also checking that the first three names are not present.

tests/functional/spec/relationships-unrelated.spec.js
ajmaddaford: Done.

app/views/handlers/relationships/unrelated_question.py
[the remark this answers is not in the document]
MebinAbraham: I don't think there is any clear or nice way we will achieve this so I am happy to keep it as is but maybe we could add a comment about what the method really does?

tests/functional/spec/relationships-unrelated.spec.js
MebinAbraham: Fair enough.

tests/integration/questionnaire/test_questionnaire_relationships.py
ajmaddaford: This is replicating existing behaviour. 

I refactored the relationship router to return locations rather than urls, and so the test was refactored into an integration test. The previous [test](https://github.com/ONSdigital/eq-questionnaire-runner/pull/368/files#diff-67a62cde5555ec46cd9db95e79b9829b8218fca230e20e5f347b31a7779df21dL45) was on the relationship router.

app/views/handlers/relationships/unrelated_question.py
```diff
@@ -1,61 +1,39 @@
...
 
     def handle_post(self):
-        pass  # pragma: no cover
+        self.questionnaire_store_updater.update_answers(
+            self.form.data, list_item_id=self.current_location.list_item_id
+        )
+        self.questionnaire_store_updater.save()
```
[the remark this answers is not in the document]
ajmaddaford: Good spot. I'll take a look at the submitter and see if it's more appropriate in this pull request or should be added to the other story.

app/views/handlers/relationships/unrelated_question.py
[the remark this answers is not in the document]
ajmaddaford: 👍  I've added a comment.

ajmaddaford:
> Not a big deal but at the moment the answer to `related-to-anyone-else-answer` is persisted even when the strategy no longer applies, i.e. we remove people so that there aren't 5 people hence the answer to that question is redundant. But the question is, should it be removed?

I think not. I think that whether the strategy applies or not is decided by the routing path, and we don't remove non-relationship answers that aren't on the routing path due to routing changes. 

This has made me think of another scenario that doesn't quite work as expected though:

- Add six people
- Answer "Unrelated" twice
- Answer "No" to the unrelated question
- Go back to the second "Unrelated" relationship and select something else
- Choose "Unrelated" for the next relationship
- The unrelated question is pre-populated with "No". Ideally this should not be answered as the list is now different.

tests/integration/questionnaire/test_questionnaire_relationships.py
MebinAbraham: Fair enough. I have added a card to track this issue.

MebinAbraham:
@ajmaddaford 

Regarding:
> I think not. I think that whether the strategy applies or not is decided by the routing path, and we don't remove non-relationship answers that aren't on the routing path due to routing changes.

So I was checking the dump debug but since it doesn't exist in the submission payload yet as we haven't done the logic around only adding it to the submission what is on the path. So once you address that, I guess my concern is resolved.

Regarding:
> This has made me think of another scenario that doesn't quite work as expected though:
>
> Add six people
Answer "Unrelated" twice
> Answer "No" to the unrelated question
> Go back to the second "Unrelated" relationship and select something else
> Choose "Unrelated" for the next relationship
> The unrelated question is pre-populated with "No". Ideally this should not be answered as the list is > now different.

Ya that isn't great. We can't really work out when it changed without storing additional state?

ajmaddaford:
> So I was checking the dump debug since it doesn't exist in the submission payload yet as we haven't done the logic around only adding it to the submission what is on the path. So once you address that, I guess my concern is resolved.

👍 

> Ya that isn't great. We can't really work out when it changed without storing additional state?

Yeah, I think we would need to store either the previous list item id, or the list of list item ids we used for the list when the unrelated question was answered. I think it's out of scope for this pull request and I'll ask Andy if he wants us to look into it.

MebinAbraham (commented):
Looks good, happy with the feature. Just wondering if we want to consider only displaying the list items that actually have remaining relationships.
So if the list summary has 2 people to display but the answer to 1 already existed, you essentially have 1 item id left, hence the requirement of at least 2 renaming items isn’t fulfilled, and we don’t show the unrelated strategy?

MebinAbraham (approved):
👍

ajmaddaford:
@MebinAbraham 
> Looks good, happy with the feature. Just wondering if we want to consider only displaying the list items that actually have remaining relationships.
> So if the list summary has 2 people to display but the answer to 1 already existed, you essentially have 1 item id left, hence the requirement of at least 2 renaming items isn’t fulfilled, and we don’t show the unrelated strategy?

I think this is something for a future iteration. The current relationships routing path implementation is based on the routing path implementation, where we don't look ahead as we build the path. Although it would be possible, we would need to think carefully about the repercussions of introducing it.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -51,25 +73,68 @@ def convert_answers_to_payload_0_0_3(
...
-        block_type
-    ):
-        answers_ids_in_add_block = schema.get_answer_ids_for_list_items(
-            list_collector_block["id"]
+    unrelated_block_id, unrelated_answer_id = None, None
+    if unrelated_block := relationships_block.get("unrelated_block"):
+        unrelated_block_id = unrelated_block["id"]
+        unrelated_answer_id = schema.get_first_answer_id_for_block(
+            unrelated_block["id"]
         )
```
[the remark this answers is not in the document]
ajmaddaford: Done.

JamesGardiner (approved):

PR #371, opened 2020-10-28
Change eq-session to use persistent storage (Datastore)

ajmaddaford:
Switches session storage back to Datastore. We have experienced issues (high CPU and connection count) with Redis at high levels of load and don't want to get into managing multi-node Redis infrastructure if we don't have to.

A further card will make the session storage backend configurable.

This has been tested as part of the [performance benchmark investigation](https://github.com/ONSdigital/eq-survey-runner-benchmark/blob/master/doc/performance-investigations/0009-eq-session-datastore-over-redis/summary.md).

### How to review 
Run the app and check that it works

### Checklist


JamesGardiner (approved):

app/routes/questionnaire.py
```diff
@@ -69,14 +69,10 @@ def before_questionnaire_request():
...
 
-@login_required
 @post_submission_blueprint.before_request
+@login_required
 def before_post_submission_request():
-    session_store = get_session_store()
-    if not session_store or not session_store.session_data:
-        raise NoQuestionnaireStateException(401)
-
-    session_data = session_store.session_data
```
MebinAbraham, line 79: 👍

MebinAbraham (approved):

app/routes/questionnaire.py
ajmaddaford: 🙂

JamesGardiner (approved):

PR #372, opened 2020-10-28
Hide IR guidance on hub when survey is complete

berroar:

Individual Response guidance is now hidden on the hub when the survey is complete.

### How to review 

Use the `test_individual_response` schema to check that the IR guidance is not displayed on the hub when the survey is complete

### Checklist


berroar:
[Changes to hub at submission point (s)](https://trello.com/c/nWy4e24I/4277-changes-to-hub-at-submission-point-s)

templates/hub.html
```diff
@@ -35,7 +35,7 @@ <h1 class="u-mt-m">{{ content.title }}</h1>
         <p class="u-mb-no" data-qa="guidance">{{ content.guidance }}</p>
       </div>
     {% endif %}
-    {% if content.individual_response_url %}
+    {% if content.individual_response_url and not content.survey_complete %}
```
MebinAbraham, line 38: What if `individual_response_url` wasn't set when survey was complete?

LJBabbage: As survey complete is only used for IR at the moment, and it adds logic in the template I think not setting  individual_response_url makes more sense. However not something I will block on

LJBabbage (approved):

templates/hub.html
berroar: I've updated it now as probably better to use less logic

MebinAbraham (approved):

PR #374, opened 2020-10-28
Add feedback call to action

JamesGardiner:
Adds the feedback call to action to the census thank you and email confirmation pages, along with a feedback submission limit of 10.

### How to review 
Using the test_feedback schema, ensure the call to action is displayed correctly, and that the link takes you to the appropriate feedback page.

Test that it is *not* shown on the test_thank_you_census_household, test_confirmation_email and the census household schemas (these don't have the feedback turned on).

Test that the done button on the feedback sent page takes you to the thank you page.

Test that the feedback call to action does not render when feedback has been sent.

Test that submitting the feedback form on all attempts above 11 results in rendering the 429 template with the correct content.

### Checklist


tests/integration/routes/test_feedback.py
```diff
@@ -62,7 +61,28 @@ def test_valid_feedback(self):
...
+            self.get(self.SEND_FEEDBACK_URL)
+            self.post(
+                {
+                    "feedback-type": "Page design and structure",
+                    "feedback-text": "Feedback",
+                }
+            )
+
+        # When I view the feedback sent page
+        self.assertInUrl(self.SEND_FEEDBACK_URL)
```
ajmaddaford, line 78: The comment implies an action, rather than an assertion.

I also think this test is missing an assertion for the error on GET when the limit is reached. I think we should:
- loop 10 times to reach the limit
- assert that a GET on the `SEND_FEEDBACK_URL` returns the error
- assert that a POST to the `SEND_FEEDBACK_URL` returns the error

tests/integration/routes/test_feedback.py
```diff
@@ -62,7 +61,28 @@ def test_valid_feedback(self):
...
+
+        # Then an appropriate error is shown
+        self.assertInBody(
+            "You have reached the maximum number of times for submitting feedback"
+        )
+
+    def test_submission(self):
```
ajmaddaford, line 85:
```suggestion
    def test_multiple_submissions(self):
```

tests/integration/routes/test_feedback.py
```diff
@@ -74,8 +94,8 @@ def test_second_submission(self):
         # When I go to the feedback send page again
         self.get(self.SEND_FEEDBACK_URL)
 
-        # Then I get redirected to the sent page
-        self.assertInUrl(self.SENT_FEEDBACK_URL)
+        # Then I go to the feedback send page
+        self.assertInUrl(self.SEND_FEEDBACK_URL)
```
ajmaddaford, line 98: I think we should extend this test to test a second submission of feedback.

tests/integration/routes/test_feedback.py
```diff
@@ -131,6 +151,18 @@ def test_feedback_page_previous(self):
...
+        self.launchSurvey("test_feedback")
+        self.post({"answer_id": "Yes"})
+
+        # When I view the thank you page
+        self.post()
+
+        # Then I should see the feedback call to action
+        self.assertInBody("What do you think about this service?")
+        self.assertInSelectorCSS("/submitted/feedback/send", class_="feedback__link")
+
```
ajmaddaford, line 165: I can't see a test for the navigation from the "Done" button on feedback sent.
Also, the previous link text should be tested as it's "Back" rather than "Previous" for this screen.

ajmaddaford (changes requested):
I don't think there is a test schema that has confirmation email and feedback. I think there should be as we should have tests for:
- accessing feedback from the confirmation email sent page
- the back link from feedback in this scenario
- the done link from feedback sent in this scenario

tests/functional/base_pages/question.page.js
```diff
@@ -93,6 +93,14 @@ class QuestionPage extends BasePage {
...
     return '[data-qa="guidance"]';
   }
+
+  feedback() {
+    return ".feedback";
+  }
+
+  feedbackLink() {
+    return ".feedback a";
+  }
```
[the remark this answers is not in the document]
JamesGardiner: Yeah I was considering where to put this, but the only shared root was here and I didn't want to duplicate. I'll add it to both if you think it's cleaner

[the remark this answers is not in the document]
JamesGardiner: done

tests/integration/routes/test_feedback.py
```diff
@@ -62,7 +61,28 @@ def test_valid_feedback(self):
         self.assertInUrl(self.SENT_FEEDBACK_URL)
         self.assertInBody("Thank you for your feedback")
 
-    def test_second_submission(self):
+    def test_eleventh_submission_errors(self):
```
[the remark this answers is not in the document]
JamesGardiner: Done, also updated it to use `EQ_FEEDBACK_LIMIT + 1` as the upper bound on the range

templates/census-thank-you.html
```diff
@@ -49,12 +50,25 @@ <h1 data-qa="title">
     <p>{{ _("Your personal information is protected by law and will be kept confidential.") }}</p>
   {% endcall %}
 
+  <hr class="u-mt-no">
```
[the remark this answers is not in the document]
JamesGardiner: @jrbarnes9 ?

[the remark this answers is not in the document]
jrbarnes9: yeah only render if if the confirmation email form is present.  It's not required if only the feedback CTA follows.

tests/integration/routes/test_feedback.py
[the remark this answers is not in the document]
ajmaddaford: 👌

app/routes/errors.py
```diff
@@ -78,7 +80,41 @@ def method_not_allowed(error=None):
...
+        "You have reached the maximum number of times for submitting feedback"
+    )
+    contact_us_message = lazy_gettext(
+        "If you need to give more feedback, please <a href='{url}'>contact us</a>."
+    )
+
+    return _render_error_page(
+        429,
+        template="429",
+        page_title=lazy_gettext(
```
LJBabbage, line 112: Bit of duplication, could be page_title=title

tests/integration/routes/test_feedback.py
```diff
@@ -62,20 +62,91 @@ def test_valid_feedback(self):
...
         self.assertInBody("Thank you for your feedback")
 
-    def test_second_submission(self):
-        # Given I launch and complete the test_feedback questionnaire, and provide feedback
+    def test_feedback_error_message_on_get_when_limit_reached(self):
+        # Given I launch and complete the test_feedback questionnaire, and provide
+        # feedback the maximum number of times
+        self._launch_and_complete_questionnaire()
+
+        for _ in range(0, EQ_FEEDBACK_LIMIT):
```
berroar, line 70: Only minor but with the other limit tests (IR, email) we set a lower value for the limit in `SetUp` so wouldn't need these loops

app/routes/questionnaire.py
```diff
@@ -291,14 +290,24 @@ def get_thank_you(schema):
...
-                    "post_submission.get_confirmation_email_sent",
+                    ".get_confirmation_email_sent",
                     email=confirmation_email.get_url_safe_serialized_email(),
                 )
             )
 
+    show_feedback_call_to_action = True
+
+    try:
+        Feedback(schema, form_data=request.form)
```
LJBabbage, line 301: This will check the submitted_time twice as its already done in ThankYou

LJBabbage (changes requested):
The feedback limit needs to be added to deployment yaml

app/routes/questionnaire.py
JamesGardiner: I don't think there is a more elegant way to do this without making multiple calls to `get_session_store`

JamesGardiner: Done by moving the check out of the Feedback and ThankYou classes and into the routes before_request method

app/routes/questionnaire.py
```diff
@@ -288,37 +292,48 @@ def get_thank_you(schema):
...
+        Feedback(schema, session_store, form_data=request.form)
+    except (FeedbackNotEnabled, FeedbackLimitReached):
+        show_feedback_call_to_action = False
+
     return render_template(
         template=thank_you.template,
-        content=thank_you.get_context(),
+        content={
+            **thank_you.get_context(),
+            **{"show_feedback_call_to_action": show_feedback_call_to_action},
```
ajmaddaford, line 314:
```suggestion
            "show_feedback_call_to_action": show_feedback_call_to_action,
```

app/views/handlers/thank_you.py
```diff
@@ -21,13 +19,11 @@ class ThankYou:
...
-    def __init__(self, schema):
-        self.session_store: SessionStore = get_session_store()
-        self.session_data: SessionData = self.session_store.session_data
+    def __init__(self, schema: QuestionnaireSchema, session_store: SessionStore):
+        self.session_store: SessionStore = session_store
         self._schema: QuestionnaireSchema = schema
 
-        if not self.session_data.submitted_time:
-            raise NotFound
+        self.session_data: SessionData = self.session_store.session_data
```
ajmaddaford, line 26: I don't think a property for `session_data` is necessary. The feedback handler just uses `_session_store.session_data.<property>` when accessing data from the session.

templates/partials/feedback-call-to-action.html
```diff
@@ -0,0 +1,13 @@
+{% from 'components/feedback/_macro.njk' import onsFeedback %}
+
+{% if content.show_feedback_call_to_action %}
```
ajmaddaford, line 3: The conditional should be in the parent template, as currently `onsFeedback` is imported even if the call to action isn't going to be displayed.

templates/partials/feedback-call-to-action.html
```diff
@@ -0,0 +1,13 @@
...
+{{
+    onsFeedback({
+        "id": "feedback u-mt-xl",
```
ajmaddaford, line 6: The `id` still has a css class present.

tests/functional/spec/census_thank_you.spec.js
```diff
@@ -0,0 +1,19 @@
+import HouseholdConfirmationPage from "../generated_pages/thank_you_census_household/household-confirmation.page";
+import Summary from "../generated_pages/thank_you_census_household/summary.page";
+
+import ThankYouPage from "../base_pages/thank-you.page";
+
+describe("Thank You Census Household", () => {
+  describe("Given I launch the Thank You Census Household schema", () => {
```
ajmaddaford, line 7:
```suggestion
  describe("Given I launch a census schema without feedback enabled", () => {
```

tests/integration/routes/test_feedback.py
```diff
@@ -62,20 +62,101 @@ def test_valid_feedback(self):
...
         self.assertInUrl(self.SENT_FEEDBACK_URL)
 
-        # When I go to the feedback send page again
+        # When I request the send feedback page
         self.get(self.SEND_FEEDBACK_URL)
 
-        # Then I get redirected to the sent page
-        self.assertInUrl(self.SENT_FEEDBACK_URL)
+        # Then I go to the feedback send page
+        self.assertInUrl(self.SEND_FEEDBACK_URL)
```
ajmaddaford, line 140: I think these lines can go as they are now covered by the `test_multiple_submissions` test?

tests/integration/routes/test_feedback.py
```diff
@@ -62,20 +62,101 @@ def test_valid_feedback(self):
...
+        self.get(self.SEND_FEEDBACK_URL)
+        self.post(
+            {"feedback-type": "Page design and structure", "feedback-text": "Feedback"}
+        )
+
+        # When I request the send feedback page
+        self.get(self.SEND_FEEDBACK_URL)
+
+        # Then I go to the feedback send page
+        self.assertInUrl(self.SEND_FEEDBACK_URL)
```
ajmaddaford, line 159: I'm not sure why we need to go back to the send page here. We should just assert that we're on the feedback sent page after the second `post`.

tests/integration/routes/test_feedback.py
```diff
@@ -62,20 +62,101 @@ def test_valid_feedback(self):
...
+
+    def test_send_feedback_back_breadcrumb(self):
+        # Given I launch and complete the test_feedback questionnaire
+        self._launch_and_complete_questionnaire()
+
+        # When I view the send feedback page
+        self.get(self.SEND_FEEDBACK_URL)
+
+        # Then I should see the Back breadcrumb
+        self.assertInSelectorCSS("Back", class_="breadcrumb__link")
```
ajmaddaford, line 124: The link has an id that could be used here?

ajmaddaford (changes requested):

templates/partials/feedback-call-to-action.html
JamesGardiner: I keep deleting this, honestly...

tests/integration/routes/test_feedback.py
```diff
@@ -131,7 +203,91 @@ def test_feedback_page_previous(self):
...
+
+        # Then the done button should navigate to the thank you page
+        selector = "[data-qa='btn-done']"
+        self.assertInSelector("/submitted/thank-you", selector)
+
     def _launch_and_complete_questionnaire(self):
         self.launchSurvey("test_feedback")
         self.post({"answer_id": "Yes"})
-        self.post()
+        self.post({"feedback-type": "Page design and structure"})
```
ajmaddaford, line 293: This shouldn't have changed as this post is the survey submit button.

tests/integration/routes/test_feedback.py
```diff
@@ -131,7 +203,91 @@ def test_feedback_page_previous(self):
...
+        # When I request the feedback page
+        self.get("/submitted/feedback/send")
+
+        # Then I am able to submit feedback
+        self.post(
+            {"feedback-type": "Page design and structure", "feedback-text": "Feedback"}
+        )
+        self.assertInUrl("/submitted/feedback/sent")
+
+    def test_feedback_cta_visible_on_email_confirmation(self):
```
ajmaddaford, line 230:
```suggestion
    def test_feedback_call_to_action_visible_on_email_confirmation(self):
```

ajmaddaford (approved):

LJBabbage (approved):

PR #375, opened 2020-10-29
Remove empty list placeholder text from WLH summary

petechd:
This removes empty household or visitors list placeholder text from WLH section summary in all census and ccs schemas. This change is described on [following Trello card](https://trello.com/c/IScBQvN6). Placeholder was removed from `section_summary_context.py`, summary template and some tests had to be modified accordingly. It needs to merge first before [schemas PR](https://github.com/ONSdigital/eq-questionnaire-schemas/pull/191).

### How to review 

### Checklist


petechd:
[bug - Remove placeholder text on WLH summary](https://trello.com/c/IScBQvN6/4279-bug-remove-placeholder-text-on-wlh-summary)

berroar:
Think this is now displaying `None` rather than empty text 
image

petechd:
> Think this is now displaying `None` rather than empty text
Made it an empty string instead.

berroar (approved):

app/views/contexts/section_summary_context.py
```diff
@@ -159,7 +159,9 @@ def _list_summary_element(self, summary, current_location, section) -> Mapping:
             "type": rendered_summary["type"],
             "add_link": add_link,
             "add_link_text": rendered_summary["add_link_text"],
-            "empty_list_text": rendered_summary["empty_list_text"],
+            "empty_list_text": rendered_summary["empty_list_text"]
+            if "empty_list_text" in summary.keys()
+            else "",
```
ajmaddaford, line 164: This could be simpler with a `get`:
```suggestion
            "empty_list_text": rendered_summary.get("empty_list_text")
```

It also shouldn't be setting an empty string, as that will mean the template will output an empty `div`. The template should be amended to add a check for `empty_list_text`. So in `sectionsummary.html`:

```
            {% else %}
              <div class="u-mb-s">
                {{ summary.empty_list_text }}
              </div>
            {% endif %}
```

will become:
```
            {% elif summary.empty_list_text %}
              <div class="u-mb-s">
                {{ summary.empty_list_text }}
              </div>
            {% endif %}
```

ajmaddaford (changes requested):

app/views/contexts/section_summary_context.py
petechd: Added and done! Thanks👍

templates/sectionsummary.html
```diff
@@ -31,7 +31,7 @@ <h2>{{summary.title}}</h2>
           {% if summary.list.list_items %}
             {% set list = summary.list %}
             {% include 'partials/summary/list-summary.html' %}
-            {% else %}
+            {% elif summary.empty_list_text %}
```
ajmaddaford, line 34: It's minor, but the indentation is off here, and as we're making changes, we should fix it. The `elif` should align with the `if` (and it looks like everything below it down to the `endfor` needs to be shifted left).

petechd: Thanks, fixed now!

ajmaddaford (approved):

PR #383, opened 2020-11-03
Add no label to checkbox answers templates

petechd:
This adds new "no_label" Boolean field to checkbox answer template. It is required by this [card](https://trello.com/c/91PIVgpv) and associated with [this change](https://github.com/ONSdigital/eq-questionnaire-validator/pull/79) in validator and [schemas PR](https://github.com/ONSdigital/eq-questionnaire-schemas/pull/193).

### How to review 
Run NI census schemas and check if `religion`, `childhood_religion` and `sexual_orientation` questions don't have checkbox answer labels.

### Checklist


petechd:
[P1: Checkbox default text to be removed - NI](https://trello.com/c/91PIVgpv/4266-p1-checkbox-default-text-to-be-removed-ni)

petechd:
[P1: Checkbox default text to be removed - NI](https://trello.com/c/91PIVgpv/4266-p1-checkbox-default-text-to-be-removed-ni)

templates/partials/answers/checkbox.html
```diff
@@ -2,13 +2,21 @@
 
 {% if answer.label %}
   {% set checkbox_label = answer.label %}
-{% elif answer.options|length == 1 %}
-  {% set checkbox_label = '' %}
-{% else %}
+{% elif answer.no_label != True and answer.options|length != 1 %}
```
LJBabbage, line 5: The double negative makes this tricky to read, and there is now opportunity to fix the issue with answer.options|length == 1 also creating an empty element in the HTML. Something like 

```
{% if answer.no_label or answer.options|length == 1%}
   {{ onsCheckboxes({
    "legend": question_title,
    "id": answer.id,
    "legendClasses": "u-vh",
    "checkboxes": map_checkbox_config(form, answer),
    "error": error
  }) }}

{% else %}

  {{ onsCheckboxes({
  "legend": question_title,
  "id": answer.id,
  "legendClasses": "u-vh",
  "checkboxesLabel": answer.label or _("Select all that apply"),
  "checkboxes": map_checkbox_config(form, answer),
  "error": error
  }) }}

{% endif %}
```

LJBabbage (changes requested):

templates/partials/answers/checkbox.html
petechd: Makes it much shorter, done!

LJBabbage (approved):

LJBabbage (changes requested):
Sorry approved, then realised we probably should have a test

petechd:
> Sorry approved, then realised we probably should have a test

Added new functional test with `show_label` property set to false.

test_schemas/en/test_checkbox.json
```diff
@@ -32,6 +32,67 @@
...
+                                                "q_code": "6",
+                                                "description": "Choose any other topping",
+                                                "value": "Other",
+                                                "detail_answer": {
+                                                    "mandatory": true,
+                                                    "id": "null-label-other-answer-mandatory",
+                                                    "label": "Please specify other",
+                                                    "type": "TextField"
+                                                }
+                                            }
```
ajmaddaford, line 86: Test schemas should be as small as possible to focus on what is being tested:
- There is no need for this many options when testing for no label on the answer, 2 or 3 is enough
- `q_code` is optional, so should be removed (I can see this is the same for existing questions in this test, but they can be removed too).

test_schemas/en/test_checkbox.json
```diff
@@ -32,6 +32,67 @@
             "groups": [
                 {
                     "blocks": [
+                        {
```
ajmaddaford, line 35: I don't think adding new blocks in test schemas before existing ones is a good thing. It means other un-related tests need to be updated to add the new block. I think we should add a new schema, and have the three different variants and their tests - default label (select all that apply), no label and custom label.

test_schemas/en/test_checkbox.json
```diff
@@ -32,6 +32,67 @@
...
+                            "type": "Question",
+                            "id": "null-label-checkbox",
```
ajmaddaford, line 37: `null` is how we've implemented the schema for this, but as a feature I think `no label` is clearer and more appropriate.

petechd:
New `null` value for label now added!

templates/partials/answers/checkbox.html
```diff
@@ -1,20 +1,22 @@
...
-  {% set checkbox_label = '' %}
+{% if answer.label == None or answer.options|length == 1 %}
+  {{ onsCheckboxes({
+    "legend": question_title,
+    "id": answer.id,
+    "legendClasses": "u-vh",
+    "checkboxes": map_checkbox_config(form, answer),
+    "mutuallyExclusive": mutuallyExclusive,
+    "error": error
+  }) }}
```
ajmaddaford, line 11: There is now duplication of the `onsCheckboxes` call, so they could un-knowingly diverge. I think this should follow the same pattern as `radio.html` where the macro config that is the same is set, and then an attribute (`checkboxesLabel` in this case) would be set if applicable before calling the macro.

ajmaddaford (changes requested):

templates/partials/answers/checkbox.html
petechd: Added new functional tests with new test schema with all those features you requested plus checkbox template is now using syntax as in radio template. Thanks!

test_schemas/en/test_checkbox_label.json
```diff
@@ -0,0 +1,132 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "0",
+    "title": "Other input fields",
+    "theme": "default",
+    "description": "A questionnaire to demo checkbox label variants.",
```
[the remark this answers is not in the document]
petechd: Added `title` and `description`, thanks!

ajmaddaford (approved):

petechd:
> Sorry approved, then realised we probably should have a test
Functional tests added now.

LJBabbage (approved):

PR #386, opened 2020-11-03
Apply latest footer component

LJBabbage:
This PR apply the latest footer from the design system and adds the correct URSL

N.B At present the page for terms and conditions page has not been published, so you will get a 404

### How to review 
Check the footer has changed and the links work as expected

### Checklist


templates/layouts/configs/_footer.html
```diff
@@ -1,62 +1,56 @@
...
+                        "url": footer_urls.terms_and_conditions,
+                        "target": "_blank"
+                    }
+                ]
+            }
+        ],
+        "copyrightDeclaration": {
+            "copyright": _("Crown copyright and database rights 2020 OS 100019153."),
+            "text": _("Use of address data is subject to the"),
+            "link": _("terms and conditions"),
```
MebinAbraham, line 51: No related directly to your PR but translations won't really work well for this, I thought Andrew raised this previously.

MebinAbraham (commented):
Looks good, just a minor comment.

templates/layouts/configs/_footer.html
```diff
@@ -1,62 +1,56 @@
...
+            "url": footer_urls.terms_and_conditions,
+            "target": "_blank"
```
MebinAbraham, line 53: This link doesn't open in a new page even with target of "_blank"

MebinAbraham (commented):
Does the old runner pages like privacy need to be kept?. If they are not being used, then I think they should be removed.

templates/layouts/configs/_footer.html
LJBabbage: You're right, but not much I can do, bar raise a bug in the design system

app/helpers/template_helpers.py
```diff
@@ -115,15 +117,42 @@ def get_census_base_url(schema_theme: str, language_code: str) -> str:
...
+    if schema_theme == "census-nisra":
+        footer_urls[
+            "help"
+        ] = f"{base_url}help/help-with-the-questions/online-questions-help/"
+    else:
+        footer_urls[
+            "help"
+        ] = f"{base_url}help/how-to-answer-questions/online-questions-help/"
+
+    return footer_urls
```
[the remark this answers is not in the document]
LJBabbage: I bit chunky monkey like this, but will change it

LJBabbage:
> Does the old runner pages like privacy need to be kept?. If they are not being used, then I think they should be removed.

No idea, just went with the design

templates/layouts/configs/_footer.html
LJBabbage: Confirmed as a bug in DS and will be fixed there

MebinAbraham (approved):

berroar (approved):
Tested footer links in english and welsh, all works as expected

PR #389, opened 2020-11-04
Impute unrelated relationships

ajmaddaford:
Updates relationships in the relationship store when an answer is provided for the "Related to anyone else?" question:
- If `Yes` is selected, any `Unrelated` relationships that remain for that person are removed. Note that other relationships (i.e. not `Unrelated` are left intact)
- If `No` is selected, the remaining relationships for that person are added to the relationship store with a relationship of `Unrelated`

### How to review 
- Try different routes to test the behaviour
- Check that the appropriate relationships are stored (use `/dump/debug`)
- Check that the submission payload is as expected (use `/dump/submission`)

### Checklist


JamesGardiner (approved):

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=add-actions-for-unrelated-blocks
```

app/views/handlers/relationships/relationship_question.py
```diff
@@ -34,7 +34,14 @@ def unrelated_answer_id(self):
             return self._schema.get_first_answer_id_for_block(self.unrelated_block_id)
         return None
 
-    @property
+    @cached_property
+    def unrelated_no_answer_value(self):
```
MebinAbraham, line 38: I can't see where this has been used?

app/submitter/convert_payload_0_0_3.py
```diff
@@ -118,21 +118,12 @@ def add_relationship_answers(
...
-        if location.to_list_item_id:
-            if relationship_answer := relationship_store.get_relationship(
-                location.list_item_id, location.to_list_item_id
-            ):
-                relationship_answers.append(relationship_answer)
-        else:
+        if location.block_id == unrelated_block_id:
             if unrelated_answer := answer_store.get_answer(
                 unrelated_answer_id, list_item_id=location.list_item_id
             ):
```
MebinAbraham, line 128: These could be combined. Not sure if how walrus operator plays nice when it is not the first expression.

app/views/handlers/relationships/relationship_collector.py
```diff
@@ -14,12 +15,16 @@ def is_location_valid(self):
...
+            relationship_answer,
+        )
+        self.relationship_store.add_or_update(relationship)
         self.questionnaire_store_updater.update_relationships_answer(
             relationship_store=self.relationship_store,
-            form_data=self.form.data,
             relationships_answer_id=self.relationships_answer_id,
-            list_item_id=self._current_location.list_item_id,
-            to_list_item_id=self._current_location.to_list_item_id,
         )
```
MebinAbraham, line 28: I think we should check if the relationship store is dirty first and instead of waiting until the answer store checks the value against the existing value.

MebinAbraham (commented):
Looks good.

app/views/handlers/relationships/relationship_question.py
ajmaddaford: On line 67.

app/submitter/convert_payload_0_0_3.py
ajmaddaford: I think I avoided this because the walrus expression needed to be wrapped in brackets, and thought it would be less readable with the nested brackets for the `get_answer` call.

I've tried this and it actually looks more readable than I thought it would, once it's formatted with black it ends up looking like this:

```python
        if location.block_id == unrelated_block_id and (
            unrelated_answer := answer_store.get_answer(
                unrelated_answer_id, list_item_id=location.list_item_id
            )
        ):
            answers_payload.add_or_update(unrelated_answer)
```

What do you think?

app/views/handlers/relationships/relationship_collector.py
ajmaddaford: Good spot. This was originally done in questionnaire store updater and I didn't notice that when I moved it.

app/submitter/convert_payload_0_0_3.py
MebinAbraham: I think that is fine and I prefer it over the two ifs.

app/views/handlers/relationships/relationship_question.py
MebinAbraham: IDE was being weird. 🤔

app/submitter/convert_payload_0_0_3.py
ajmaddaford: Yes, I ran lint and all lint checks passed 🙂

ajmaddaford: I've just pushed this change.

MebinAbraham (approved):

PR #390, opened 2020-11-04
Fix cookie banner content

rmccar:
On staging there is no content in the cookie banner. At some point the cookie banner has had the params change and this PR is to update the implementation in runner

image

### How to review 
Check that the cookie banner displays correctly

### Checklist


rmccar:
[Cookies banner content missing](https://trello.com/c/gnsaXLQo/4294-cookies-banner-content-missing)

templates/layouts/_base.html
```diff
@@ -56,8 +56,10 @@
   {% if cookie_settings_url %}
     {{
       onsCookiesBanner({
-          "html": _('We use cookies to improve your experience of census.gov.uk'),
-          "secondaryButtonUrl": cookie_settings_url
+        "secondaryButtonUrl": cookie_settings_url,
+        "statementTitle": _('Tell us whether you accept cookies'),
+        "statementText": _('We use <a href="{cookie_settings_url}">cookies to collect information</a> about how you use census.gov.uk. We use this information to make the website work as well as possible and improve our services.'),
+        "confirmationText": _('You’ve accepted all cookies. You can <a href="{cookie_settings_url}">change your cookie preferences</a> at any time.'),
```
MebinAbraham, line 62: The URLs needs to be interpolated. And I used single to avoid escaping quotes, as browser will use double quotes for `href` automatically.
```suggestion
        "statementText": _("We use <a href='{cookie_settings_url}'>cookies to collect information</a> about how you use census.gov.uk. We use this information to make the website work as well as possible and improve our services.").format(cookie_settings_url=cookie_settings_url),
        "confirmationText": _("You’ve accepted all cookies. You can <a href='{cookie_settings_url}'>change your cookie preferences</a> at any time.").format(cookie_settings_url=cookie_settings_url),
```

MebinAbraham (commented):
This makes me wonder why we don't have a functional test as we wouldn't know if it broke again without it.

LJBabbage (commented):
The content looks good. It might be my set up, but if I click on either why we use cookies or cookies preference, it just adds True to the url and ends up going to 404

.development.env
```diff
@@ -24,3 +24,4 @@ WEB_SERVER_UWSGI_ASYNC_CORES=10
 CDN_URL=https://cdn.eq.census-gcp.onsdigital.uk
 CDN_ASSETS_PATH=/design-system
 ADDRESS_LOOKUP_API_URL=https://whitelodge-ai-api.census-gcp.onsdigital.uk
+COOKIE_SETTINGS_URL=True
```
MebinAbraham, line 27: This could be defaulted to `#` so the page doesn't link anywhere external. Will fix the issue Lewis highlighted.

rmccar: Yeah good idea. Ill sort that

tests/integration/components/test_render_cookies_banner.py
```diff
@@ -0,0 +1,10 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestRenderCookiesBanner(IntegrationTestCase):
+    def setUp(self):
+        super().setUp()
+        self.launchSurvey("test_dropdown_mandatory")
+
+    def test_cookies_banner_renders(self):
+        self.assertInBody("Tell us whether you accept cookies")
```
MebinAbraham, line 10: I think we should rename this to `TestCookieBanner` and the test to `test_cookies_banner_contents` as there is no rendering as integration tests don't use JS.

tests/integration/components/test_render_cookies_banner.py
```diff
@@ -0,0 +1,10 @@
+from tests.integration.integration_test_case import IntegrationTestCase
```
MebinAbraham, line 1: File name needs updating. And just making sure that this test only covers the content and not the JS behaviour as if that broke, we still would be able to tell.

MebinAbraham (approved):

templates/layouts/_base.html
```diff
@@ -56,8 +56,10 @@
   {% if cookie_settings_url %}
     {{
       onsCookiesBanner({
-          "html": _('We use cookies to improve your experience of census.gov.uk'),
-          "secondaryButtonUrl": cookie_settings_url
+        "secondaryButtonUrl": cookie_settings_url,
+        "statementTitle": _('Tell us whether you accept cookies'),
+        "statementText": _("We use <a href='{cookie_settings_url}'>cookies to collect information</a> about how you use census.gov.uk. We use this information to make the website work as well as possible and improve our services.").format(cookie_settings_url=cookie_settings_url),
+        "confirmationText": _("You’ve accepted all cookies. You can <a href='{cookie_settings_url}'>change your cookie preferences</a> at any time.").format(cookie_settings_url=cookie_settings_url),
```
ajmaddaford, line 62: I think `primaryButtonText`, `secondaryButtonText` and `confirmationButtonText` need to be passed to the macro so that translations will work.

ajmaddaford (changes requested):

LJBabbage (approved):

ajmaddaford (approved):

PR #398, opened 2020-11-05
Move post submission warning panel to footer

petechd:
This moves post submission warning panel from pre footer template to the main footer. Changed described on this [Trello](https://trello.com/c/qQQNtgsR).

### How to review 
Check post submission thank you page.

### Checklist


petechd:
[Move post submission "leave this page or close your browser" warning into the footer (XS)](https://trello.com/c/qQQNtgsR/4239-move-post-submission-leave-this-page-or-close-your-browser-warning-into-the-footer-xs)

templates/layouts/configs/_footer.html
```diff
@@ -52,5 +51,10 @@
             "url": footer_urls.terms_and_conditions,
             "target": "_blank"
         }
-    })
-}}
+} %}
+
+{% if content and content.submitted_time %}
+  {% set config = config | setAttribute("footerWarning", _("Make sure you <a href='/sign-out'>leave this page</a> or close your browser if using a shared device")) %}
```
MebinAbraham, line 57: The `href` should use a variable and the URL should use `url_for` when using our own endpoints.

petechd: Made it a variable using `content.sign_out_url`. Thanks!

LJBabbage (changes requested):
This will only work on the thank you page, it also needs to work on the feedback and confirmation email pages.

app/views/contexts/thank_you_context.py
```diff
@@ -31,6 +31,7 @@ def build_census_thank_you_context(
 ) -> Mapping:
 
     context = {
+        "submitted_time": session_data.submitted_time,
```
MebinAbraham, line 34: This needs to be added to the context of all post submission pages.

LJBabbage: As my comment above, do we want to do that? should it not live in template_helper

LJBabbage:
As discussed, all post_submission_blueprint need a flag added, as the endpoints are already protected, it can just be show_warning_message: True. I believe there is some rationalisation that can be done on all end points (hide_signout_button, show_warning_message, sign_out_url etc), but this is possibly out of scope for this story

petechd:
> As discussed, all post_submission_blueprint need a flag added, as the endpoints are already protected it can just be show_warning_message: True with no extra logic. I believe there is some rationalisation that can be done on all end points (hide_signout_button, show_warning_message, sign_out_url etc), but this is possibly out of scope for this story


I've added new `show_warning` flag (should it be `show_warning_message`?). Tests to feedback and confirmation email added too. Thanks!👍

LJBabbage (approved):
Looks good, I probably wouldn't have bothered testing the warning message on each endpoint as its exercising the same thing, but I guess it guarantees it's there, so doesn't need to be remove. Your call on show_warning_message /show_warning, either or works for me

MebinAbraham (commented):
Since this is not census specific, I think the flag needs to be added to the default thank you context also.

petechd:
> Since this is not census specific, I think the flag needs to be added to the default thank you context also.

Added to default thank you page and renamed flag to what you proposed, thanks!

app/views/contexts/email_form_context.py
```diff
@@ -3,6 +3,7 @@
 
 def build_confirmation_email_form_context(email_confirmation_form):
     context = {
+        "show_sign_out_warning": True,
         "hide_signout_button": False,
```
berroar, line 7: Minor but seems like we should use either `sign_out` or `signout` for both these vars

petechd: since we already use "sign_out_url" I will change it to `sign_out`

MebinAbraham (approved):

PR #402, opened 2020-11-09
Add NI and language variations to the footer

berroar:
Updates the footer component so that the ONS logo is displayed in Welsh when the Census is being completed in Welsh, and the NISRA logo when an NI census is being completed.

### How to review 
Use an NI census schema and check that the NISRA logo is being displayed in the footer.
Use a Welsh census schema and check that ONS logo in the footer displays the Welsh/English logo when you toggle between the languages.

### Checklist


berroar:
[P1: Add NI variation and language variation to the logos in the footer](https://trello.com/c/0si0NrHC/4300-p1-add-ni-variation-and-language-variation-to-the-logos-in-the-footer)

app/helpers/template_helpers.py
```diff
@@ -131,6 +131,7 @@ def get_footer_urls(language_code: str, base_url: str, schema_theme: str):
         accessibility_statement_path = "hygyrchedd/"
         privacy_and_data_protection_path = "preifatrwydd-a-diogelu-data/"
         terms_and_conditions_path = "telerau-ac-amodau/"
+        ons_logo_language = "cy"
```
MebinAbraham, line 134: I think this should just be passed as `language` which can be used for anything that needs language.

app/helpers/template_helpers.py
```diff
@@ -141,13 +142,16 @@ def get_footer_urls(language_code: str, base_url: str, schema_theme: str):
...
+        ons_logo_language = "en"
 
     return {
         "help": f"{base_url}{help_path}",
         "cookies": f"{base_url}{cookies_path}",
         "accessibility_statement": f"{base_url}{accessibility_statement_path}",
         "privacy_and_data_protection": f"{base_url}{privacy_and_data_protection_path}",
         "terms_and_conditions": f"{base_url}{terms_and_conditions_path}",
+        "schema_theme": schema_theme,
+        "ons_logo_language": ons_logo_language,
```
MebinAbraham, line 154: I think this should just be passed as `language` which can be used for anything that needs language.

templates/layouts/configs/_footer.html
```diff
@@ -44,6 +54,7 @@
                 ]
             }
         ],
+        "poweredBy": nisraLogo,
```
MebinAbraham, line 57: If this only applies to NISRA then this should be done conditionally:
```jinja2
{% if footer_urls.schema_theme == "census-nisra" %}
    {% set config = config | setAttribute("poweredBy", {
        "logo": 'nisra-logo-black-en',
        "alt": 'NISRA - Northern Ireland Statistics and Research Agency'
        }) 
    %}
{% endif %}
```

app/helpers/template_helpers.py
berroar: but if it's eo or ga I don't think it'll display properly

ajmaddaford (commented):
The header and footer implementations aren't consistent. For the header we set the `header` property on `pageConfig`, but for the footer we call the `onsFooter` macro rather than using the `pageConfig.footer` property. 

I think we should create the footer context in code and pass it to `pageConfig.footer` in `_base.html`.

templates/layouts/configs/_footer.html
MebinAbraham: Will change once Andrews comment is done.

app/helpers/template_helpers.py
MebinAbraham: Will change once Andrew's comment is done.

app/helpers/template_helpers.py
```diff
@@ -42,6 +42,90 @@ def get_page_header_context(language, theme):
     return context.get(theme)
 
 
+def get_footer_context(language_code, theme, base_url, contact_us_url, sign_out_url):
+    footer_urls = get_footer_urls(language_code, theme)
+
+    footer_warning = (
+        lazy_gettext(
+            f"Make sure you <a href='{sign_out_url}'>leave this page</a> or close your browser if using a shared device"
+        )
```
ajmaddaford, line 51: This would fail to lookup the translation as the signout url would be interpolated before the translation is looked up.
```suggestion
        lazy_gettext(
            "Make sure you <a href='{sign_out_url}'>leave this page</a> or close your browser if using a shared device"
        ).format(sign_out_url=sign_out_url)
```

app/helpers/template_helpers.py
```diff
@@ -42,6 +42,90 @@ def get_page_header_context(language, theme):
...
+        "copyrightDeclaration": {
+            "copyright": lazy_gettext(
+                "Crown copyright and database rights 2020 OS 100019153."
+            ),
+            "text": lazy_gettext("Use of address data is subject to the"),
+            "link": lazy_gettext("terms and conditions"),
+            "url": f"{base_url}/{footer_urls['terms_and_conditions_path']}",
+            "target": "_blank",
+        },
+        "footerWarning": footer_warning,
```
ajmaddaford, line 68: The `footerWarning` property shouldn't be set if there is no warning. It would be better to update the context if applicable before returning from this method.

app/helpers/template_helpers.py
```diff
@@ -42,6 +42,90 @@ def get_page_header_context(language, theme):
...
+    }
+    context = {
+        "census": {
+            **default_context,
+            "schema_theme": theme,
+            "lang": language_code,
+        },
+        "census-nisra": {
+            **default_context,
+            "schema_theme": theme,
```
ajmaddaford, line 118: Is `schema_theme` used?

app/helpers/template_helpers.py
```diff
@@ -124,31 +211,29 @@ def get_contact_us_url(language_code: str, base_url: str):
...
-        "privacy_and_data_protection": f"{base_url}{privacy_and_data_protection_path}",
-        "terms_and_conditions": f"{base_url}{terms_and_conditions_path}",
-    }
+        return {
+            "help_path": help_path,
+            "cookies_path": "cookies/",
+            "accessibility_statement_path": "accessibility/",
+            "privacy_and_data_protection_path": "privacy-and-data-protection/",
+            "terms_and_conditions_path": "terms-and-conditions/",
+        }
```
ajmaddaford, line 236: I think the single return was done to ensure this method always returns the a dict with the same properties.

app/translations/messages.pot
```diff
@@ -1,1297 +0,0 @@
-# Translations template for PROJECT.
```
ajmaddaford, line 1: I don't think this should be removed?

app/helpers/template_helpers.py
```diff
@@ -63,21 +141,24 @@ def render_template(template, **kwargs):
...
-    footer_urls = get_footer_urls(language_code, base_url, theme)
     include_csrf_token = request.url_rule and "POST" in request.url_rule.methods
     account_service_url = cookie_session.get(
         "account_service_url", f"{CENSUS_BASE_URL}en/start"
     )
+    sign_out_url = url_for("session.get_sign_out")
+    content_urls = get_static_content_urls(language_code, base_url, theme)
+    footer_context = get_footer_context(
+        language_code, theme, content_urls, sign_out_url
+    )
```
MebinAbraham, line 152: The theme might not exist at this point so it needs a default of "census". At the moment, without a cookie and non census themes do not render a footer.

MebinAbraham (approved):

app/helpers/template_helpers.py
```diff
@@ -42,6 +42,90 @@ def get_page_header_context(language, theme):
     return context.get(theme)
 
 
+def get_footer_context(
+    language_code, static_content_urls, sign_out_url, theme="census"
```
ajmaddaford, line 46: We are defaulting theme to census in more than one place. Looking at the code, I think we could change line 142 to:
```python
theme = cookie_session.get("theme", DEFAULT_THEME)
```

And define `DEFAULT_THEME` at the top of the file.

Then we can remove the defaulting from the page header and footer context calls.

app/helpers/template_helpers.py
```diff
@@ -42,6 +42,90 @@ def get_page_header_context(language, theme):
     return context.get(theme)
 
 
+def get_footer_context(
```
ajmaddaford, line 45: This method should have the same cache decorator (`@lru_cache(maxsize=None)`) as the `get_page_header_context` call. This ensures that the context is only ever generated once for any given combination of arguments.

app/helpers/template_helpers.py
```diff
@@ -117,37 +207,34 @@ def get_census_base_url(schema_theme: str, language_code: str) -> str:
...
 
-def get_contact_us_url(language_code: str, base_url: str):
-    if language_code == "cy":
-        return f"{base_url}cysylltu-a-ni/"
-
-    return f"{base_url}contact-us/"
-
-
-def get_footer_urls(language_code: str, base_url: str, schema_theme: str):
+def get_static_content_urls(language_code: str, base_url: str, schema_theme: str):
```
ajmaddaford, line 210: This method should have the same cache decorator (`@lru_cache(maxsize=None)`) as the `get_page_header_context` call. This ensures that the urls are only generated once for any given combination of arguments.

app/helpers/template_helpers.py
berroar: This causes an `unhashable type` error as the static urls are a dict.

app/helpers/template_helpers.py
```diff
@@ -42,6 +43,88 @@ def get_page_header_context(language, theme):
...
+    if request.blueprint == "post_submission":
+        default_context["footerWarning"] = lazy_gettext(
+            "Make sure you <a href='{sign_out_url}'>leave this page</a> or close your browser if using a shared device"
+        ).format(sign_out_url=sign_out_url)
+
+    context = {
+        "default": {
+            **default_context,
+            "lang": language_code,
+        },
```
ajmaddaford, line 110: The page header context also handles `social` and `northernireland` themes, so this should too for consistency.
We should be able to move `lang` into the default context, and the `census-nisra` dict merge below will overwrite it.

ajmaddaford (approved):

PR #405, opened 2020-11-10
Add missing auto-suggest properties to address lookup

MebinAbraham:
Add missing auto-suggest properties to address lookup used by screen readers. Checked with Matt and confirmed that these are only needed for address lookups and not textfield suggestion lookups.

### How to review 
Try the address lookup and ensure results are grouped when search term is too broad i.e. `CF63`
To test `"There are {n} for {x}"` you will have to use the VoiceOver feature and hover over a grouped result, which should read like `There are 8 addresses for Court Road, Barry, CF63 1AB` etc.

Before:
Screenshot 2020-11-10 at 17 22 03

After:
Screenshot 2020-11-10 at 17 21 46

### Checklist

### DIFF
--- a/templates/partials/answers/address.html
+++ b/templates/partials/answers/address.html
@@ -44,6 +44,8 @@
     "ariaOneResult": _("There is one suggestion available."),
     "ariaNResults": _("There are {n} suggestions available."),
     "ariaLimitedResults": _("Results have been limited to 10 suggestions. Type more characters to improve your search"),
+    "ariaGroupedResults": _("There are {n} for {x}"),
+    "groupCount": _("{n} addresses"),
     "moreResults": _("Enter more of the address to improve results"),
     "resultsTitle": _("Select an address"),
     "noResults": _("No results found. Try entering a different part of the address"),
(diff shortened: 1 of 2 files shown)


templates/partials/answers/address.html
```diff
@@ -44,6 +44,8 @@
     "ariaOneResult": _("There is one suggestion available."),
     "ariaNResults": _("There are {n} suggestions available."),
     "ariaLimitedResults": _("Results have been limited to 10 suggestions. Type more characters to improve your search"),
+    "ariaGroupedResults": _("There are {n} for {x}"),
```
ajmaddaford, line 47: I think this needs to be `There are {n} addresses for {x}`, as I think this would resolve to something like `There are 5 for CF23`.

https://github.com/ONSdigital/design-system/blob/61db1aaadca98deca77e7ee40d1dd15654024278/src/components/input/autosuggest/autosuggest.ui.js#L433

ajmaddaford (changes requested):

templates/partials/answers/address.html
ajmaddaford: So @MebinAbraham  has investigated and this is correct - it uses the `groupCount` property interpolated into the `ariaGroupedResults` message. We shouldn't really be doing that, but that would need to be fixed in the design system.

ajmaddaford (approved):

berroar (approved):

PR #407, opened 2020-11-11
Update design system to version 27.0.0

ons-eq-team:
This is an auto-generated PR.
Design system release: https://github.com/ONSdigital/design-system/releases/tag/27.0.0


MebinAbraham:
Approving this but footer is visually broken due to a breaking change in the DS however, there is also a bug in the DS which means that it can't be fixed without another release. Therefore, this can be progressed to unblock the other card and footer issue can be tracked separately.

MebinAbraham (approved):

berroar (approved):

PR #414, opened 2020-11-13
Integrate dropdown lists for feedback

petechd:
This adds dropdown list feature to feedback form in runner as described on this [Trello card](https://trello.com/c/GP8rtsdS). Customised based on different form types of schemas [`C`, `I` and `H`]. Feedback form context needed some refactoring after addition of new detail answer id.

### How to review 
New test schemas with different form types added together with new integration tests.

### Checklist


petechd:
[Integrate Drop-down lists for feedback (S)](https://trello.com/c/GP8rtsdS/4287-integrate-drop-down-lists-for-feedback-s)

app/views/handlers/feedback.py
```diff
@@ -114,6 +61,125 @@ def handle_post(self):
         self._session_store.session_data.feedback_count += 1
         self._session_store.save()
 
+    def get_question_schema(self):
+        if self._schema.form_type == "C":
```
LJBabbage, line 65: There is quite a bit of duplication here, I would be more incline to do this 

```
detail_answers_option_map = {
            "C": ["General", "This establishment", "People who live here"],
            "I": ["General", "Accommodation", "Personal details"],
            "H": ["General", "People who live here", "Visitors"],
            }

        options = ({"label": lazy_gettext(value), "value": value} for value in detail_answers_option_map.get(self._schema.form_type or "H"))

```
Not sure about the default to "H" which both this and yours does, might want to check what we want to do if there is no form_type

LJBabbage (changes requested):

app/views/handlers/feedback.py
petechd: Refactored, I'll ask about the default option, thanks!👍

app/views/handlers/feedback.py
```diff
@@ -114,6 +61,106 @@ def handle_post(self):
         self._session_store.session_data.feedback_count += 1
         self._session_store.save()
 
+    def get_question_schema(self):
```
ajmaddaford, line 64: This could be defined as a cached property, which would also be consistent with the `form` property:
```python
    @cached_property
    def question_schema(self):
```

app/views/contexts/feedback_form_context.py
```diff
@@ -14,7 +14,13 @@ def build_feedback_context(question_schema, form):
...
     }
 
-    answer_ids = (answer["id"] for answer in question_schema["answers"])
+    answer_ids = []
+    for answer in question_schema["answers"]:
+        answer_ids.append(answer["id"])
+        if answer["type"] in ("Checkbox", "Radio"):
+            for option in answer["options"]:
+                if "detail_answer" in option:
+                    answer_ids.append(option["detail_answer"]["id"])
```
ajmaddaford, line 23: I can see why this has been done, but the duplication in this file and the question context is not ideal. We could replace most of this file with a call to `build_question_context`, and append what we need before returning the context:

```python
def build_feedback_context(question_schema, form):
    block = {
        "question": question_schema
    }
    context = build_question_context(block, form)
    context["hide_sign_out_button"] = False
    context["question"] = question_schema
    context["sign_out_url"] = url_for("session.get_sign_out")

    return context
```

app/views/handlers/feedback.py
petechd: Done, thanks!

app/views/handlers/feedback.py
```diff
@@ -114,6 +61,106 @@ def handle_post(self):
...
+                    "type": "Radio",
+                    "id": "feedback-type",
+                    "mandatory": True,
+                    "options": [
+                        {
+                            "label": lazy_gettext("The census questions"),
+                            "value": lazy_gettext("The census questions"),
+                            "detail_answer": {
+                                "type": "Dropdown",
+                                "id": "feedback-type-dropdown",
```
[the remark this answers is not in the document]
petechd: Changed to what you've suggested! Thanks!

app/views/contexts/feedback_form_context.py
petechd: Calling `build_question_context` now, thanks!

app/views/handlers/feedback.py
```diff
@@ -114,6 +61,106 @@ def handle_post(self):
...
+                "Household and accommodation",
+                "Personal details",
+                "Health",
+                "Qualifications",
+                "Employment",
+            ],
+        }
+
+        options = (
+            {"label": lazy_gettext(value), "value": value}
```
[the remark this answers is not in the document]
petechd: Ok, added lazy_gettext translation of both label and value now, I tried to translate `detail_answers_option_map` before it's being passed but that doesn't seem to work..

tests/integration/routes/test_feedback.py
```diff
@@ -309,6 +309,32 @@ def test_feedback_submitted_done_button_after_email_confirmation(self):
         selector = "[data-qa='btn-done']"
         self.assertInSelector("/submitted/thank-you", selector)
 
+    def test_feedback_dropdown_form_type(self):
```
LJBabbage, line 312: I am unsure what this is testing and it looks like there are 2 tests in one here. For me this is exercising the details answer using questionnaire form, but do we not test that already? We naturally need one feedback form with a topic, but I don't think we need multiple

LJBabbage (changes requested):

tests/integration/routes/test_feedback.py
petechd: Added two new testes as you recommended, test schemas removed, irrelevant tests removed! Thanks!

ajmaddaford (approved):

LJBabbage (approved):

PR #415, opened 2020-11-13
Fix unrelated block routing when question variants exist

ajmaddaford:
Fixes a bug with routing from the "No" answer for the "Related to anyone else" question when variants exist with different "No" answer values. Before this fix the "No" answer wasn't skipping the remaining relationships for that person.

### How to review 
- Add six people
- Answer unrelated, unrelated, related to anyone else no, unrelated, unrelated, related to anyone else no
- You should be taken to the next persons relationships questions i.e. skip the rest of the relationship questions for that person

This should also be tested in the census schemas.

### Checklist


berroar (approved):
Tested using test and Census schemas and it now follows the expected behaviour

test_schemas/en/test_relationships_unrelated.json
```diff
@@ -512,41 +512,150 @@
...
+                                                                                            "id": "list_item_id"
+                                                                                        }
+                                                                                    },
+                                                                                    "delimiter": " "
+                                                                                }
+                                                                            }
+                                                                        ]
+                                                                    }
+                                                                ],
+                                                                "text": "No, none of these people are related to {person_name}"
```
LJBabbage, line 635: It would good to have these obviously different as I didn't pick up the placeholder in the second variant initially and then thought they were the same, meaning I didn't think the schema was exercising the bug

tests/functional/spec/relationships-unrelated.spec.js
```diff
@@ -52,7 +52,7 @@ describe("Unrelated Relationships", () => {
       });
 
       it("When I select 'No' to the 'related to anyone else' question, Then I will be taken to the first relationship for the second person", () => {
```
LJBabbage, line 54: The 'No' needs updating

LJBabbage (approved):
Works for me, a couple of minor points that don't have to fixed.

test_schemas/en/test_relationships_unrelated.json
ajmaddaford: The text was lifted from the census schemas, do you have a suggestion for something different that makes sense?

tests/functional/spec/relationships-unrelated.spec.js
ajmaddaford: I decided not to change this as I think the sentence describes the action of selecting the "No" answer, rather than specifying the exact text.

test_schemas/en/test_relationships_unrelated.json
LJBabbage: Like I said it's so minor it can stay like it is, and why I approved. I just prefer test_schemas to be obvious what it's trying to demonstrate

" No, {person_name} is not related to anyone else in the household"

ajmaddaford: I'm going to stick with the original text as it is consistent with the other "No" answer in referring to the list of people that appears on this page.

PR #416, opened 2020-11-13
Add Feedback payload creation and upload to GCS

LJBabbage:
This PR creates the payload and upload to GCS

### How to review 
Using the [pipeline PR](https://github.com/ONSdigital/eq-pipelines/pull/104) build the application and make sure when using a questionnaire with feedback enabled sends feedback it creates an entry in the GCS bucket. This also needs 
testing locally, as it defaults to the logs if not deployed on GCP

### Checklist


app/views/handlers/feedback.py
```diff
@@ -111,6 +115,18 @@ def get_page_title(self):
...
 
     def handle_post(self):
+        feedback_data = {
+            "feedback_text": self.form.data["feedback-type"],
+            "feedback_topic": self.form.data["feedback-text"],
+        }
+
+        feedback_upload = current_app.eq["feedback"].upload(
+            self._schema, feedback_data, self._session_store.session_data
+        )
```
MebinAbraham, line 125: As specified on the card, it was agreed that add all non form data attributes should be included as metadata and payload. Therefore, this could be simplified by passing 2 things into the method, `data` and `metadata` where metadata is a subset of data without "`feedback_text` and `feedback_topic`.

app/submitter/submitter.py
```diff
@@ -141,3 +143,46 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
+    def __init__(self, bucket_name):
+        client = storage.Client()
+        self.bucket = client.get_bucket(bucket_name)
+        self.bucket.retention_period = 86400
+
+    def upload(self, schema, feedback_data, session_data):
+        object_key = str(uuid4())
+        blob = self.bucket.blob(object_key)
+        blob.metadata = {
+            "feedback_count": session_data.feedback_count + 1,
```
MebinAbraham, line 158: We shouldn't be incrementing in 2 places but this would be addressed by my comment about passing in data and metadata that needs to be uploaded instead of constructing the object here.

app/submitter/submitter.py
```diff
@@ -141,3 +143,46 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
+            "form_type": schema.form_type,
+            "language_code": session_data.language_code,
+            "object_key": object_key,
```
MebinAbraham, line 161: Not sure we need this in the payload as anytime you interact with a blob, you have access to it's name which is the key. With it being a random UUID, it doesn't provide  any additional filtering etc.

app/views/handlers/feedback.py
```diff
@@ -19,6 +19,10 @@ class FeedbackLimitReached(Exception):
     pass
 
 
+class FeedbackUploadFailedException(Exception):
```
MebinAbraham, line 22: To be consistent with the other two exceptions.
```suggestion
class FeedbackUploadFailed(Exception):
```

tests/app/submitter/test_submitter.py
```diff
@@ -262,3 +264,55 @@ def test_send_message_adds_metadata():
...
+        )
+        # Then
+        bucket = client.return_value.get_bucket.return_value
+        blob = bucket.blob.return_value
+
+        assert blob.metadata["feedback_count"], 1
+        assert blob.metadata["form_type"], "H"
+        assert blob.metadata["language_code"], "cy"
+        assert blob.metadata["region_code"], "GB-ENG"
+        assert blob.metadata["tx_id"], "12345"
```
MebinAbraham, line 308:
```suggestion
        assert blob.metadata["feedback_count"] == 1
        assert blob.metadata["form_type"] == "H"
        assert blob.metadata["language_code"] == "en"
        assert blob.metadata["region_code"] == "GB-ENG"
        assert blob.metadata["tx_id"] == "12345"
```

tests/integration/routes/test_feedback.py
```diff
@@ -309,6 +311,23 @@ def test_feedback_submitted_done_button_after_email_confirmation(self):
...
+
+        # When I post my feedback
+        self.post(
+            {"feedback-type": "Page design and structure", "feedback-text": "Feedback"}
+        )
+
+        # Then I should see an error page
+        self.assertStatusCode(500)
+        self.assertEqualPageTitle("Sorry, there is a problem - Census 2021")
+        self.assertInBody("submit your feedback again")
```
MebinAbraham, line 329: We should also assert that the retry URL is as expected,

app/submitter/submitter.py
```diff
@@ -141,3 +143,46 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
+            "region_code": schema.region_code,
+            "submitted_at": session_data.submitted_time,
+            "tx_id": session_data.tx_id,
+        }
+        blob.upload_from_string(str(feedback_data).encode("utf8"))
```
MebinAbraham, line 166: I think it would be good to set the correct content type for the blob. The `upload_from_string` method also accepts a `content_type` which we should set to `application/json` as it defaults to `text/plain`.

app/submitter/submitter.py
```diff
@@ -141,3 +143,46 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
+    def upload(self, schema, feedback_data, session_data):
+        object_key = str(uuid4())
+        blob = self.bucket.blob(object_key)
+        blob.metadata = {
+            "feedback_count": session_data.feedback_count + 1,
+            "form_type": schema.form_type,
+            "language_code": session_data.language_code,
+            "object_key": object_key,
+            "region_code": schema.region_code,
+            "submitted_at": session_data.submitted_time,
```
MebinAbraham, line 163: The card specified the feedback submitted time not the questionnaire submitted time. However, I am not sure we even want/need submitted time as all blobs have a built-in [timeCreated](https://cloud.google.com/storage/docs/json_api/v1/objects) value. I am not sure if big query can use that or whether it needs explicit timestamp data.

MebinAbraham (changes requested):

app/submitter/submitter.py
LJBabbage: This isn't an increment, session_data.feedback_count stays the same. I can't increment session_data.feedback_count until the upload is successfully, I can't see how its possible to know it that

app/submitter/submitter.py
LJBabbage: I did think that as well, but then thought it did say it on the card

app/submitter/submitter.py
MebinAbraham: Based on my other comments this should change anyway as you would be able to create the object upfront and pass to upload but in this instance you could increment the session data which was passed in here and session data is only then saved if the upload is successful.

LJBabbage: You could separate the increment and the save, but that has future bug written all over it.

MebinAbraham: It would be done in the same code block based on my changed.

app/submitter/submitter.py
```diff
@@ -141,3 +141,42 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
+            "form_type": meta_data["form_type"],
+            "language_code": meta_data["language_code"],
+            "region_code": meta_data["region_code"],
+            "tx_id": meta_data["tx_id"],
+        }
+        blob.upload_from_string(
+            str(data).encode("utf8"), content_type="application/json"
+        )
+
+        return True
```
MebinAbraham, line 164: This should be something like:
```py
    def upload(self, data, metadata):
		logger.info("uploading feedback")

        blob = self.bucket.blob(str(uuid4()))
        blob.metadata = metadata
        blob.upload_from_string(
            str(data).encode("utf8"), content_type="application/json"
        )

        return True

```
The topic around "object_key" is still open for discussion as I don't think it needs to be in the payload. Happy to discuss at tech session.

app/views/handlers/feedback.py
```diff
@@ -58,7 +63,27 @@ def get_page_title(self):
...
+        metadata = {
+            "feedback_count": session_data.feedback_count,
+            "form_type": self._schema.form_type,
+            "language_code": session_data.language_code,
+            "object_key": str(uuid4()),
+            "region_code": self._schema.region_code,
+            "tx_id": session_data.tx_id,
+        }
+
+        feedback_upload = current_app.eq["feedback"].upload(data, metadata)
```
MebinAbraham, line 82: `data` should be a superset of `metadata`, so like:
```py
        metadata = {
            "feedback_count": session_data.feedback_count,
            "form_type": self._schema.form_type,
            "language_code": session_data.language_code,
            "object_key": str(uuid4()),
            "region_code": self._schema.region_code,
            "tx_id": session_data.tx_id,
        }
        data = {
            "feedback_text": self.form.data["feedback-type"],
            "feedback_topic": self.form.data["feedback-text"],
            **metadata
        }
```

app/views/handlers/feedback.py
```diff
@@ -58,7 +63,27 @@ def get_page_title(self):
         return self.PAGE_TITLE
 
     def handle_post(self):
+        session_data = self._session_store.session_data
         self._session_store.session_data.feedback_count += 1
+
+        data = {
+            "feedback_text": self.form.data["feedback-type"],
+            "feedback_topic": self.form.data["feedback-text"],
```
MebinAbraham, line 71: The new `'feedback-type-question-category'` should be conditionally added to `data` if if has a valid value.

LJBabbage: Its on a different card

app/views/handlers/feedback.py
LJBabbage: Not sure why we would do this

MebinAbraham: Because that is one of our requirements mentioned on the card.
- `Add all non form data attributes as metadata and payload. Form data to be payload only. `

app/views/handlers/feedback.py
MebinAbraham: Okay didn't realise one existed but I also think it is suitable to add here given this it is 2 lines of code and doing a separate card for that and reviews would just be longer. 

However, regardless if the above is done separately, this reminds me that it can't be tested anyway without adding unit test for handler. Therefore, it probably should be a different class that builds the payload like we do with IR which can then be unit tested as at the moment, we have no guarantee that the submitted feedback will arrive in GCS as expected without manual testing.

LJBabbage: This is a integration test that checks the blobs payload and metadata

app/views/handlers/feedback.py
LJBabbage: It's not how I read this, however I will change it for a easy life

app/views/handlers/feedback.py
MebinAbraham: > There is an integration test that checks the blobs payload and metadata and this is how submitter does it

Sorry I don't see it, where exactly?. If you are referring to the test of GCSFeedback class then that is not the same thing, that is testing that the submitter can submit to GCS, but we need assurances of the feedback payload. For submitter, we introduced a submission handler which has a `get_payload` method which we partially test so this could be similar to that but the payload for feedback has to be tested independently to the submitter.

LJBabbage: That's not quite true, I added the that test because I couldn't get to the submission_language_code, the rest of submission payload is not tested

MebinAbraham: > That's not quite right, I added that test because I couldn't get to the submission_language_code, the rest of submission payload is not tested. 

Because that is what you added so tested the new bit you introduced. But because we don't have a test to check the submission payload shouldn't be a reason to not have a test for this...

> The integration test creates a GCSFeedback class gives it metadata and the data, then asserts the blob is correct which will be uploaded, this is more than submitter does, I don't follow what more we need

`GCSFeedback` is a submitter, it tests that the submitter works, it has nothing to do with feedback really. It tests that given some data and metadata (which could be anything in any structure you set up in test), the blob is as expected which is not the same as testing that for a feedback submission the object contains the expected properties and values like we do with IR and `ConfirmationEmail`.

app/setup.py
```diff
@@ -29,7 +29,13 @@
 from app.publisher import LogPublisher, PubSubPublisher
 from app.secrets import SecretStore, validate_required_secrets
 from app.storage import Datastore, Dynamodb, Redis
-from app.submitter import GCSSubmitter, LogSubmitter, RabbitMQSubmitter
+from app.submitter import (
+    GCSFeedback,
+    GCSSubmitter,
+    LogFeedback,
+    LogSubmitter,
+    RabbitMQSubmitter,
```
ajmaddaford, line 37: I think as the new feedback submitters are types of submitter they should be named something like `FeedbackGCSSubmitter` and `FeedbackLogSubmitter`.

app/setup.py
```diff
@@ -341,6 +349,21 @@ def setup_publisher(application):
...
 
 
+def setup_feedback(application):
+    if application.config["EQ_FEEDBACK_BACKEND"] == "gcs":
+        bucket_id = application.config.get("EQ_GCS_FEEDBACK_BUCKET_ID")
+
+        if not bucket_id:
+            raise Exception("Setting EQ_GCS_FEEDBACK_BUCKET_ID Missing")
+
+        application.eq["feedback"] = GCSFeedback(bucket_name=bucket_id)
```
ajmaddaford, line 359: I think it would be clearer if this was:
```suggestion
        application.eq["feedback_submitter"] = GCSFeedback(bucket_name=bucket_id)
```

app/submitter/submitter.py
```diff
@@ -141,3 +141,36 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
                 self._disconnect(connection)
+
+
+class GCSFeedback:
+    def __init__(self, bucket_name):
+        client = storage.Client()
+        self.bucket = client.get_bucket(bucket_name)
+
+    def upload(self, data, metadata):
+        blob = self.bucket.blob(metadata["object_key"])
```
ajmaddaford: Does `object_key` need to be in the metadata?

app/submitter/submitter.py
```diff
@@ -141,3 +141,36 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
+        blob.upload_from_string(
+            str(data).encode("utf8"), content_type="application/json"
+        )
+
+        return True
+
+
+class LogFeedback:
+    @staticmethod
+    def upload(data, meta_data):
```
ajmaddaford: For consistency with elsewhere in runner, `meta_data` should be `metadata`.

app/submitter/submitter.py
```diff
@@ -141,3 +141,36 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
+        logger.info("uploading feedback")
+        logger.info(
+            "feedback payload",
+            data=data,
+            feedback_count=meta_data["feedback_count"],
+            form_type=meta_data["form_type"],
+            language_code=meta_data["language_code"],
+            object_key=meta_data["object_key"],
+            region_code=meta_data["region_code"],
+            tx_id=meta_data["tx_id"],
```
ajmaddaford: This can be done with:
```suggestion
            **meta_data,
```

But as we're logging the data under a `data` key, I think this should be logged as `meta_data=meta_data`.

app/views/handlers/feedback.py
```diff
@@ -58,7 +63,28 @@ def get_page_title(self):
...
+            "feedback_count": session_data.feedback_count,
+            "form_type": self._schema.form_type,
+            "language_code": session_data.language_code,
+            "object_key": str(uuid4()),
+            "region_code": self._schema.region_code,
+            "tx_id": session_data.tx_id,
+        }
+        data = {
+            "feedback_text": self.form.data["feedback-type"],
+            "feedback_topic": self.form.data["feedback-text"],
```
ajmaddaford:
```suggestion
            "feedback_topic": self.form.data["feedback-type"],
            "feedback_text": self.form.data["feedback-text"],
```

Also, if we're using `topic` rather than `type` here, then it should be renamed to `topic` in the form to be consistent.

app/views/handlers/feedback.py
```diff
@@ -58,7 +62,23 @@ def get_page_title(self):
...
+        session_data.feedback_count += 1
+        feedback_upload = FeedbackUpload(
+            session_data.feedback_count,
+            self._schema.form_type,
+            session_data.language_code,
+            self._schema.region_code,
+            session_data.tx_id,
+            self.form.data,
+        )
+        feedback_upload = current_app.eq["feedback_submitter"].upload(
```
ajmaddaford, line 75: The same variable name (`feedback_upload`) should not be used for something else here.

app/views/handlers/feedback.py
```diff
@@ -174,3 +194,40 @@ def is_limit_reached(session_data: SessionData) -> bool:
     @staticmethod
     def is_enabled(schema: QuestionnaireSchema) -> bool:
         return schema.get_submission().get("feedback")
+
+
+class FeedbackUpload:
```
ajmaddaford, line 199: I'm not sure of the naming here. For me, it's not an upload until we attempt to upload it somewhere. It might we worth considering creating a `FeedbackMetadata` and `FeedbackMessage` class?

app/views/handlers/feedback.py
```diff
@@ -58,7 +62,23 @@ def get_page_title(self):
...
-        self._session_store.session_data.feedback_count += 1
+        session_data = self._session_store.session_data
+        session_data.feedback_count += 1
+        feedback_upload = FeedbackUpload(
+            session_data.feedback_count,
+            self._schema.form_type,
+            session_data.language_code,
+            self._schema.region_code,
+            session_data.tx_id,
+            self.form.data,
```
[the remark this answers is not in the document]
LJBabbage: Passing self.form.data is a pain as feedback-type-question-category, might not be there, meaning I will have to do a check before and after the call

app/views/handlers/feedback.py
```diff
@@ -58,7 +62,30 @@ def get_page_title(self):
...
+            self.form.data.get("feedback_type"),
+            self.form.data.get("feedback_type_question_category"),
+        )
+
+        feedback_upload = current_app.eq["feedback_submitter"].upload(
+            feedback_metadata.metadata, feedback_message.payload
+        )
+
+        if not feedback_upload:
+            raise FeedbackUploadFailed()
```
JamesGardiner, line 87:
```suggestion
        if not current_app.eq["feedback_submitter"].upload(
            feedback_metadata.metadata, feedback_message.payload
        ):
            raise FeedbackUploadFailed()
```

app/views/handlers/feedback.py
```diff
@@ -174,3 +201,47 @@ def is_limit_reached(session_data: SessionData) -> bool:
...
+class FeedbackMetadata:
+    def __init__(self, feedback_count, form_type, language_code, region_code, tx_id):
+        self.feedback_count = feedback_count
+        self.form_type = form_type
+        self.language_code = language_code
+        self.region_code = region_code
+        self.tx_id = tx_id
+
+    @property
+    def metadata(self) -> Mapping:
```
JamesGardiner, line 215: Is this  a pattern used elsewhere? I'd expect a `FeedbackMetadata` class or similar metadata class to return a representation of itself in this format using it's `__call__` method?

app/views/handlers/feedback.py
```diff
@@ -174,3 +201,47 @@ def is_limit_reached(session_data: SessionData) -> bool:
...
+        feedback_text,
+        feedback_type,
+        feedback_type_question_category=None,
+    ):
+        self.feedback_text = feedback_text
+        self.feedback_type = feedback_type
+        self.feedback_type_question_category = feedback_type_question_category
+
+    @property
+    def payload(self) -> Mapping:
```
JamesGardiner, line 237: As a FeedbackPayload class is inherently a representation of a payload data structure, the use of the `payload` method here should ideally be a call to `__call__`

tests/app/submitter/test_submitter.py
```diff
@@ -4,7 +4,13 @@
 from mock import Mock, call, patch
 from pika.exceptions import AMQPError
 
-from app.submitter.submitter import GCSSubmitter, RabbitMQSubmitter
+from app.data_models.session_data import SessionData
```
JamesGardiner, line 7:
```suggestion
from app.data_models import SessionData
```

tests/app/submitter/test_submitter.py
```diff
@@ -4,7 +4,13 @@
...
+from app.questionnaire.questionnaire_schema import QuestionnaireSchema
```
JamesGardiner, line 8:
```suggestion
from app.questionnaire import QuestionnaireSchema
```

tests/app/submitter/test_submitter.py
```diff
@@ -4,7 +4,13 @@
...
+from app.submitter.submitter import (
```
JamesGardiner, line 9:
```suggestion
from app.submitter import (
```

app/routes/errors.py
```diff
@@ -19,7 +19,7 @@
 from app.views.handlers.confirmation_email import (
     ConfirmationEmailFulfilmentRequestPublicationFailed,
 )
-from app.views.handlers.feedback import FeedbackLimitReached
+from app.views.handlers.feedback import FeedbackLimitReached, FeedbackUploadFailed
 from app.views.handlers.individual_response import (
```
JamesGardiner, line 23:
```suggestion
from app.views.handlers import (
```

The `IndividualResponseFulfilmentRequestPublicationFailed`and `IndividualResponseLimitReached` need to be added to the handlers init file for the above to work

app/submitter/__init__.py
```diff
@@ -1,3 +1,15 @@
-from .submitter import GCSSubmitter, LogSubmitter, RabbitMQSubmitter
+from .submitter import (
```
JamesGardiner, line 1: 👍

JamesGardiner (changes requested):

app/routes/errors.py
LJBabbage: eh?

app/setup.py
```diff
@@ -344,6 +352,23 @@ def setup_publisher(application):
         raise Exception("Unknown EQ_PUBLISHER_BACKEND")
 
 
+def setup_feedback(application):
+    if application.config["EQ_FEEDBACK_BACKEND"] == "gcs":
+        bucket_id = application.config.get("EQ_GCS_FEEDBACK_BUCKET_ID")
+
+        if not bucket_id:
```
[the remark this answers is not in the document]
LJBabbage: This won't work, without brackets and the comma

app/views/handlers/feedback.py
```diff
@@ -174,3 +199,45 @@ def is_limit_reached(session_data: SessionData) -> bool:
...
+        self.tx_id = tx_id
+
+    def __call__(self) -> Mapping:
+        return {
+            "feedback_count": self.feedback_count,
+            "form_type": self.form_type,
+            "language_code": self.language_code,
+            "region_code": self.region_code,
+            "tx_id": self.tx_id,
+        }
```
MebinAbraham, line 219: This can be simplified:
```suggestion
        return vars(self)
```

app/views/handlers/feedback.py
```diff
@@ -58,7 +62,28 @@ def get_page_title(self):
...
+            session_data.language_code,
+            self._schema.region_code,
+            session_data.tx_id,
+        )
+
+        feedback_message = FeedbackPayload(
+            self.form.data.get("feedback_text"),
+            self.form.data.get("feedback_type"),
+            self.form.data.get("feedback_type_question_category"),
+        )
```
MebinAbraham, line 80: The ids are wrong. They should be:
```suggestion
        feedback_message = FeedbackPayload(
            self.form.data.get("feedback-text"),
            self.form.data.get("feedback-type"),
            self.form.data.get("feedback-type-question-category"),
        )
```

app/views/handlers/feedback.py
```diff
@@ -174,3 +199,45 @@ def is_limit_reached(session_data: SessionData) -> bool:
...
+        payload = {
+            "feedback_text": self.feedback_text,
+            "feedback_type": self.feedback_type,
+        }
+        if self.feedback_type_question_category:
+            payload[
+                "feedback_type_question_category"
+            ] = self.feedback_type_question_category
+
+        return payload
```
MebinAbraham, line 243: This can also be simplified to:
```suggestion
        payload = vars(self)
        if not self.feedback_type_question_category:
            del payload["feedback_type_question_category"]

        return payload
```

app/views/handlers/feedback.py
```diff
@@ -174,3 +199,45 @@ def is_limit_reached(session_data: SessionData) -> bool:
...
         return schema.get_submission().get("feedback")
+
+
+class FeedbackMetadata:
+    def __init__(self, feedback_count, form_type, language_code, region_code, tx_id):
+        self.feedback_count = feedback_count
+        self.form_type = form_type
+        self.language_code = language_code
+        self.region_code = region_code
+        self.tx_id = tx_id
```
MebinAbraham, line 210: Have you looked at whether we need timestamp based on https://github.com/ONSdigital/eq-questionnaire-runner/pull/416#discussion_r524527678
As if Big Query just load JSON data from GCS then I don't think it will have access to the "blobs" attributes.

LJBabbage: Yep, I have now had it confirmed to add feedback submission date and done so

JamesGardiner (approved):

app/views/handlers/feedback.py
```diff
@@ -174,3 +200,41 @@ def is_limit_reached(session_data: SessionData) -> bool:
...
+
+    def __call__(self) -> Mapping:
+        return {
+            "feedback_count": self.feedback_count,
+            "form_type": self.form_type,
+            "language_code": self.language_code,
+            "region_code": self.region_code,
+            "tx_id": self.tx_id,
+            "feedback_submission_date": datetime.utcnow().strftime("%Y-%m-%d"),
```
MebinAbraham, line 220: It think this should be a iso format timestamp and not just a date, it's common practice to include date and time. Will also align with other payloads like IR and email.

app/views/handlers/feedback.py
```diff
@@ -174,3 +200,41 @@ def is_limit_reached(session_data: SessionData) -> bool:
...
+        }
```
MebinAbraham, line 221: This can still be simplied either of 2 ways.
1. set the datetime on initialisation (which could mean the time could be off by few millisecond but not really and issue):
```py
    def __init__(self, feedback_count, form_type, language_code, region_code, tx_id):
        self.feedback_count = feedback_count
        self.form_type = form_type
        self.language_code = language_code
        self.region_code = region_code
        self.tx_id = tx_id
        self.date_time = datetime.now(tz=tzutc()).isoformat()

    def __call__(self) -> Mapping:
        return vars(self)
```

2. Still use `vars` and update the dict.
```py
    def __call__(self) -> Mapping:
        metadata = vars(self)
        metadata["date_time"] = datetime.now(tz=tzutc()).isoformat()
        return metadata
```

app/submitter/submitter.py
```diff
@@ -141,3 +143,31 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
+    def __init__(self, bucket_name):
+        client = storage.Client()
+        self.bucket = client.get_bucket(bucket_name)
+
+    def upload(self, metadata, payload):
+        blob = self.bucket.blob(str(uuid4()))
+        blob.metadata = metadata
+        blob.upload_from_string(
+            str({**payload, **metadata}).encode("utf8"), content_type="application/json"
+        )
```
MebinAbraham, line 158: I think either payload should be updated in place since metadata is a subset of payload or assign a new variable for clarity instead of unpacking 2 dicts into a new dict.
```suggestion
        payload.update(metadata)
        blob = self.bucket.blob(str(uuid4()))
        blob.metadata = metadata
        blob.upload_from_string(
            str(payload).encode("utf8"), content_type="application/json"
        )
```

MebinAbraham (commented):
Looks good, last few minor comments.

app/submitter/submitter.py
MebinAbraham, line 158: Since this is explicitly needed as JSON this should be:
```suggestion
        blob.upload_from_string(
            data=json.dumps(payload), content_type="application/json"
        )
```

app/views/handlers/feedback.py
LJBabbage: It's supposed to be just a date and matches the requirements as stated on big data card

app/views/handlers/feedback.py
LJBabbage: Doesn't match the key in the doc, isn't the format required, and is clearer as already it

app/submitter/submitter.py
LJBabbage: I don't think this is needed

LJBabbage: Content type is set, so is this needed?

app/views/handlers/feedback.py
MebinAbraham: I dont see a reason not to be datetime but fine lets not hold thing up any longer.

app/views/handlers/feedback.py
MebinAbraham: It reduces the references from 3 to 1.

app/submitter/submitter.py
MebinAbraham: Yes, content type is used by the browser only, atm, this is still a string so json serialistion is needed.

MebinAbraham (approved):

app/submitter/submitter.py
```diff
@@ -141,3 +144,32 @@ def send_message(self, message, tx_id, questionnaire_id, case_id=None):
...
+    def __init__(self, bucket_name):
+        client = storage.Client()
+        self.bucket = client.get_bucket(bucket_name)
+
+    def upload(self, metadata, payload):
+        payload.update(metadata)
+        blob = self.bucket.blob(str(uuid4()))
+        blob.metadata = metadata
+        blob.upload_from_string(
+            json.dumps(payload).encode("utf8"), content_type="application/json"
```
MebinAbraham, line 159: Just remembered that since you are serialising the payload directly you need to update the payload with metadata before hand.
```py
payload.update(metadata)
```

tests/app/submitter/test_submitter.py
```diff
@@ -262,3 +262,49 @@ def test_send_message_adds_metadata():
...
+        assert blob.metadata["region_code"] == "GB-ENG"
+
+        blob_contents = blob.upload_from_string.call_args[0][0]
+
+        assert (
+            blob_contents
+            == b'{"feedback-type": "Feedback type", "feedback-text": "Feedback text", '
+            b'"feedback_count": 1, "feedback_submission_date": "2021-03-23", '
+            b'"form_type": "H", "language_code": "cy", "region_code": "GB-ENG", "tx_id": "12345"}'
+        )
```
MebinAbraham, line 309: I think you should update this to also check it contains metadata.

ajmaddaford (approved):

PR #418, opened 2020-11-16
Cache questionnaire schemas on app start up

MebinAbraham:
- To potentially reduce some higher than normal response times due to cold starts, this PR preloads the questionnaire schema caches on app start up instead of caching the schema on the first request to a given schema. Only schemas in the `schemas` dir is cached on start up.
- Added some extra tests for loading schemas that was failing coverage on this branch.
- Removed wrapper method used in tests which was identical to `load_schema_from_name`.

### How to review 
Ensure the all schemas including their language variants are cached on app start up.

To test you can either:
- load the schema from launcher and ensure the `loading schema` log line is not present when switching between different schemas.
- debug the app and check the cache size and hit ratio using [CacheInfo](https://docs.python.org/3/library/functools.html#functools.lru_cache)

### Checklist


app/setup.py
```diff
@@ -30,6 +30,7 @@
 from app.secrets import SecretStore, validate_required_secrets
 from app.storage import Datastore, Dynamodb, Redis
 from app.submitter import GCSSubmitter, LogSubmitter, RabbitMQSubmitter
+from app.utilities.schema import load_questionnaire_schemas_into_cache
```
JamesGardiner, line 33:
```suggestion
from app.utilities.schema import cache_questionnaire_schemas
```

app/setup.py
```diff
@@ -142,6 +143,8 @@ def before_request():  # pylint: disable=unused-variable
         ),
     )
 
+    load_questionnaire_schemas_into_cache()
```
JamesGardiner, line 146:
```suggestion
    cache_questionnaire_schemas()
```

app/utilities/schema.py
```diff
@@ -14,7 +15,8 @@
 
 logger = get_logger()
 
-DEFAULT_SCHEMA_DIRS = ["schemas", "test_schemas"]
+SCHEMA_DIR = "schemas"
+DEFAULT_SCHEMA_DIRS = (SCHEMA_DIR, "test_schemas")
```
JamesGardiner, line 19: What is the purpose of defining `SCHEMA_DIR` over this?
```suggestion
DEFAULT_SCHEMA_DIRS = ("schemas", "test_schemas")
```

app/utilities/schema.py
```diff
@@ -27,10 +29,12 @@
...
 
-def get_schema_path_map_for_language(language_code):
+def get_schema_path_map_for_language(
+    language_code: str, dirs: Optional[Tuple[str, ...]] = None
+) -> Mapping:
     schema_files = []
 
-    for schema_dir in DEFAULT_SCHEMA_DIRS:
+    for schema_dir in dirs or DEFAULT_SCHEMA_DIRS:
         schema_files.extend(glob(f"{schema_dir}/{language_code}/*.json"))
```
JamesGardiner, line 38: This section could be simplified if the Path object was used consistently throughout the method instead of being used only in the return statement

JamesGardiner: Also, `schema_files` could be created using a list comprehension here

app/utilities/schema.py
```diff
@@ -139,7 +148,7 @@ def _load_schema_file(schema_name, language_code):
             language_code=language_code,
         )
 
-    if not schema_exists(language_code, schema_name):
+    if not _schema_exists(language_code, schema_name):
```
JamesGardiner, line 151: This is a duplicate of the call to _schema_exists on line 141; the result of this function could be set to a var and re-used

app/utilities/schema.py
```diff
@@ -80,9 +86,12 @@ def load_schema_from_session_data(session_data):
     return load_schema_from_metadata(vars(session_data))
 
 
+def load_schema_from_name(schema_name, language_code=DEFAULT_LANGUAGE_CODE):
+    return _load_schema_from_name(schema_name, language_code)
```
JamesGardiner, line 90: Is there any value in this? Could we not just drop the `_` from `_load_schema_from_name` and call it directly?

app/utilities/schema.py
```diff
@@ -180,4 +189,10 @@ def load_schema_from_url(survey_url, language_code):
...
 
 def get_schema_file_path(schema_name, language_code):
-    return SCHEMA_PATH_MAP.get(language_code, {}).get(schema_name)
+    return get_schema_path_map().get(language_code, {}).get(schema_name)
+
+
+def load_questionnaire_schemas_into_cache():
+    for language_code, schemas in get_schema_path_map(dirs=(SCHEMA_DIR,)).items():
+        for schema in schemas:
+            load_schema_from_name(schema, language_code)
```
JamesGardiner, line 198:
```suggestion
	[[load_schema_from_name(schema, language_code) for schema in schemas]
	    for language_code, schema in get_schema_path_map(dirs=(SCHEMA_DIR,)).items()]
```

JamesGardiner (changes requested):

app/utilities/schema.py
MebinAbraham: So when create app caches the schemas, the `dir` can be passed in as we don't care to cache test schemas.

app/utilities/schema.py
MebinAbraham: Didn't give it much thought as it was existing, will have a look.

app/utilities/schema.py
MebinAbraham: Didn't catch it since it was existing but good shout, will fix.

app/utilities/schema.py
MebinAbraham: It was so that `load_schema_from_name` can be used with args or keyword args and still fetch from cache, as the cache depends on the method signature. If we are happy with, the rule that you must not use keyword args or must match whatever the format for loading the cache is then ya I could remove.

app/utilities/schema.py
MebinAbraham: I didn't make it a comprehension as it wasn't returning and in this instance thought it was more readable.

app/utilities/schema.py
MebinAbraham: Actually looking at the code, it is done like that because it checks if the schema exists for the requested non default language code, then if it doesn't it sets language code to "en" and then checks that the "en" schema exists.

app/utilities/schema.py
MebinAbraham: Have used a `dict` comprehension for this method.

JamesGardiner (approved):

app/utilities/schema.py
```diff
@@ -27,34 +29,35 @@
...
-    language_map_codes = ["en", "cy", "ga", "eo"]
+@lru_cache(maxsize=None)
+def get_schema_path_map(dirs: Optional[Tuple[str, ...]] = None) -> Mapping:
+    language_map_codes = {"en", "cy", "ga", "eo"}
 
     return {
-        language_code: get_schema_path_map_for_language(language_code)
+        language_code: get_schema_path_map_for_language(
+            language_code, dirs or DEFAULT_SCHEMA_DIRS
+        )
```
ajmaddaford, line 50: I think the schemas endpoint should use this method rather than `get_schema_path_map_for_language` so that it gets the benefit of using the cached list rather than walking the filesystem again. If we do that, I think it's worth folding the `get_schema_path_map_for_language` method into this one, something like:

```python
    language_codes = {"en", "cy", "ga", "eo"}

    schema_path_map = {}
    for language_code in language_codes:
        schema_path_map[language_code] = {
            Path(schema_file).with_suffix("").name: schema_file
            for schema_dir in dirs or DEFAULT_SCHEMA_DIRS
            for schema_file in glob(f"{schema_dir}/{language_code}/*.json")
        }

    return schema_path_map
```

app/utilities/schema.py
```diff
@@ -180,4 +186,10 @@ def load_schema_from_url(survey_url, language_code):
 
 
 def get_schema_file_path(schema_name, language_code):
-    return SCHEMA_PATH_MAP.get(language_code, {}).get(schema_name)
+    return get_schema_path_map().get(language_code, {}).get(schema_name)
+
+
+def cache_questionnaire_schemas():
+    for language_code, schemas in get_schema_path_map(dirs=(SCHEMA_DIR,)).items():
```
ajmaddaford, line 193: I think this would be simpler if instead of passing the directories we passed a boolean, something like `include_test_schemas`. It could be defaulted to `False` (production default), and passed `True` by the `schemas` endpoint call to get the test schemas included.

app/utilities/schema.py
MebinAbraham: Done, this is now 1 method. I decided to make the whole thing a dict comprehension, let me know what you think in terms of readability, personally I think it is just as readable.

app/utilities/schema.py
MebinAbraham: Done 👍

app/utilities/schema.py
```diff
@@ -27,34 +29,35 @@
...
+        )
         for language_code in language_map_codes
     }
 
 
-SCHEMA_PATH_MAP = get_schema_path_map()
-
-
-def schema_exists(language_code, schema_name):
+@lru_cache(maxsize=None)
```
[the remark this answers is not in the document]
MebinAbraham: Technically it can be called more than one for the same arg if a request is made to a schema for a non default language which doesn't exist and it falls back to the cached English version. However, given that shouldn't really happen, I have removed it.

app/utilities/schema.py
ajmaddaford: The dict comprehension is fine, I think the language codes should be defined as a constant though.

app/utilities/schema.py
```diff
@@ -30,11 +30,12 @@
 
 
 @lru_cache(maxsize=None)
-def get_schema_path_map(dirs: Optional[Tuple[str, ...]] = None) -> Mapping:
+def get_schema_path_map(include_test_schemas: Optional[bool] = False) -> Mapping:
+    dirs = DEFAULT_SCHEMA_DIRS if include_test_schemas else SCHEMA_DIR
```
ajmaddaford, line 34: What do you think about having a`TEST_SCHEMA_DIR` variable rather than `DEFAULT_SCHEMA_DIRS` and creating the dirs list here?

app/utilities/schema.py
```diff
@@ -139,15 +134,17 @@ def _load_schema_file(schema_name, language_code):
...
+    if not _schema_exists(language_code, schema_name):
         logger.error(
             "no schema file exists",
             schema_name=schema_name,
             language_code=language_code,
         )
         raise FileNotFoundError
 
-    schema_path = get_schema_file_path(schema_name, language_code)
+    schema_path = get_schema_path_map(include_test_schemas=True)[language_code][
```
ajmaddaford, line 145: To be consistent with the variable name used earlier:
```suggestion
    schema_path_map = get_schema_path_map(include_test_schemas=True)[language_code][
```

app/utilities/schema.py
ajmaddaford: Nice, looks much better 🙂

app/utilities/schema.py
MebinAbraham: Dir is now created in-place as it is not used elsewhere, however I still used module level variables for the dirs.

app/utilities/schema.py
MebinAbraham: Done. Updated another 2 also.

MebinAbraham: Have corrected the incorrect variable names in some instances.

app/utilities/schema.py
```diff
@@ -27,34 +29,32 @@
...
 
     return {
-        language_code: get_schema_path_map_for_language(language_code)
-        for language_code in language_map_codes
+        language_code: {
+            Path(schema_file).with_suffix("").name: schema_file
+            for schema_dir in dirs
+            for schema_file in glob(f"{schema_dir}/{language_code}/*.json")
+        }
+        for language_code in {"en", "cy", "ga", "eo"}
```
ajmaddaford, line 49: I think these language codes should be defined in a constant.

ajmaddaford (approved):

app/utilities/schema.py
MebinAbraham: Module level?

PR #422, opened 2020-11-18
Update googletagmanager querystring

pricem14pc:
A breaking googletagmanager api change means the `gtm_preview` parameter needs to be removed from the request in the snippet. The presence of this param causes the request to fail with a 404.

Check that a GTM integrated environment no longer results in a 404 for the gtm.js resource when an EQ page loads.

While investigating (and fixing) this initial issue it was found that a subsequent content-security-policy is breached by the current GTM script. The connect-src does not allow the necessary GTM domains for the scripts to function. This is not being resolved by this PR as we first need to better understand the implications of relaxing the CSP for this domain (as the relaxation should not be required for GTM to function).

### DIFF
--- a/templates/layouts/_base.html
+++ b/templates/layouts/_base.html
@@ -85,7 +85,7 @@
       (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
       new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
       j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-      'https://www.googletagmanager.com/gtm.js?id='+i+dl+ '&gtm_auth={{ google_tag_manager_auth }}&gtm_preview={{ google_tag_manager_preview }}&gtm_cookies_win=x';f.parentNode.insertBefore(j,f);
+      'https://www.googletagmanager.com/gtm.js?id='+i+dl+ '&gtm_auth={{ google_tag_manager_auth }}&gtm_cookies_win=x';f.parentNode.insertBefore(j,f);
       })(window,document,'script','dataLayer','{{ google_tag_manager_id }}');
     </script>
     <!-- End Google Tag Manager -->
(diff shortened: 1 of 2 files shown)


JamesGardiner (approved):

templates/layouts/_base.html
```diff
@@ -85,7 +85,7 @@
       (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
       new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
       j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-      'https://www.googletagmanager.com/gtm.js?id='+i+dl+ '&gtm_auth={{ google_tag_manager_auth }}&gtm_preview={{ google_tag_manager_preview }}&gtm_cookies_win=x';f.parentNode.insertBefore(j,f);
+      'https://www.googletagmanager.com/gtm.js?id='+i+dl+ '&gtm_auth={{ google_tag_manager_auth }}&gtm_cookies_win=x';f.parentNode.insertBefore(j,f);
```
MebinAbraham, line 88: So if `google_tag_manager_preview` is no longer required then it should be removed from runner and the pipelines.

MebinAbraham (changes requested):

templates/layouts/_base.html
pricem14pc: Not yet. It needs to be tested across the different environments. I have separate branches for that. Once it's been confirmed its working in multiple integrated environments that's when it will be removed.

MebinAbraham: I thought this was tested in staging?.

pricem14pc: It has been. However, there are different GTM containers for different environments. Testing only in staging (now we've removed the env value) does not confirm that the newly formed request sends the events to the correct GTM container. That can only be confirmed by testing in more environments (essentially WL/BL) hence the need to roll this change out to other integrated environments.

MebinAbraham (approved):

templates/layouts/_base.html
MebinAbraham: Can we make sure we have a card to remove this from runner and pipelines if other envs works as expected then.

PR #424, opened 2020-11-19
Removing GOOGLE_TAG_MANAGER_PREVIEW

pricem14pc:
Following removal of the `gtm_preview` request parameter in the GTM snippet it has been confirmed we no longer need to support this value in Runner and references to it can be removed.

Ensure the environment variable no longer exists and any previous references to it through the HTML templates, the template helpers, and helm deployments have been removed.

The application should deploy, and GTM integration should function as expected, without this value.


ci/deploy_app.yaml
```diff
@@ -26,6 +25,12 @@ params:
   MIN_REPLICAS:
   MAX_REPLICAS:
   TARGET_CPU_UTILIZATION_PERCENTAGE:
+  CDN_URL:
+  WEB_SERVER_TYPE:
+  WEB_SERVER_WORKERS:
+  WEB_SERVER_THREADS:
+  WEB_SERVER_UWSGI_ASYNC_CORES:
+  DATASTORE_USE_GRPC:
```
pricem14pc, line 33: Added these as part of this PR as they were previously missing. While it doesn't impact automated pipelines it does mean they can now be overridden by a local developer fly exec to a GCP env which is a useful approach.

templates/layouts/_base.html
```diff
@@ -94,10 +94,10 @@
 
 
 {% block bodyStart %}
-  {% if google_tag_manager_id and google_tag_manager_auth and google_tag_manager_preview  %}
+  {% if google_tag_manager_id and google_tag_manager_auth  %}
     <!-- Google Tag Manager (noscript) -->
     <noscript>
-      <iframe src="https://www.googletagmanager.com/ns.html?id={{ google_tag_manager_id }}&gtm_auth={{ google_tag_manager_auth }}&gtm_preview={{ google_tag_manager_preview }}&gtm_cookies_win=x"
+      <iframe src="https://www.googletagmanager.com/ns.html?id={{ google_tag_manager_id }}&gtm_auth={{ google_tag_manager_auth }}&&gtm_cookies_win=x"
```
pricem14pc, line 100: This was not (but could have been) removed in the previous PR as it was not part the testing.

ajmaddaford, line 100:
```suggestion
      <iframe src="https://www.googletagmanager.com/ns.html?id={{ google_tag_manager_id }}&gtm_auth={{ google_tag_manager_auth }}&gtm_cookies_win=x"
```

MebinAbraham (approved):
Apart from Andrew's comment, seems to be as expected 👍

ajmaddaford (approved):

PR #428, opened 2020-11-24
Index session expires_at field

ajmaddaford:
Adds a property to storage models to allow passing which fields should be indexed. This has been used to enable an index on the session `expires_at` field so we can select expired sessions for deletion.

### How to review 
I tested this by deploying it to my GCP environment. It may also be possible to test it locally by using the `gcloud` cli to interact with the mock Datastore.

In the GCP UI you should see the `expires_at` column heading text colour is black rather than grey:
image

If you have an environment that has session data before the `expires_at field` was indexed, you'll need to run a GQL query to see this:
```
SELECT * FROM `eq-session` ORDER BY expires_at
```

### Checklist


MebinAbraham (approved):

app/storage/datastore.py
```diff
@@ -18,7 +18,12 @@ def put(self, model, overwrite=True):
         key_value = getattr(model, storage_model.key_field)
 
         key = self.client.key(storage_model.table_name, key_value)
-        entity = Entity(key=key, exclude_from_indexes=tuple(serialized_item.keys()))
+        exclude_from_indexes = [
+            field
+            for field in serialized_item.keys()
+            if field not in storage_model.index_fields
+        ]
+        entity = Entity(key=key, exclude_from_indexes=tuple(exclude_from_indexes))
```
JamesGardiner, line 26:
```suggestion
        exclude_from_indexes = tuple(
            field
            for field in serialized_item.keys()
            if field not in storage_model.index_fields
        )
        entity = Entity(key=key, exclude_from_indexes=exclude_from_indexes)
```

app/storage/storage.py
```diff
@@ -44,6 +45,10 @@ def key_field(self):
     def expiry_field(self):
         return self._config.get("expiry_field")
 
+    @cached_property
+    def index_fields(self):
+        return self._config.get("index_fields") or []
```
JamesGardiner, line 50:
```suggestion
        return self._config.get("index_fields", [])
```

JamesGardiner (changes requested):

pricem14pc (approved):
Tested in GCP and new eq-session data is indexed and queryable as expected.
Existing eq-session data remain unindexed (as expected).
GQLs referencing indexed fields act only on that indexed data (as expected).
gsutil dataflow tasks successfully target that indexed data (e.g for purging)
Further performance testing of impact of the index will be carried out.

JamesGardiner (approved):

PR #434, opened 2020-11-24
Update Python packages to latest

ajmaddaford:
Updates all un-pinned Python packages to the latest version. 

#### Datastore gRPC

The default connection for Datastore is now gRPC. It looks like the latest Datastore library is broken when HTTP is used instead of gRPC. As we have committed to using gRPC this isn't an issue for us. I will investigate further to determine if it is an issue with the library or the way we're using it.

#### flask-compress

flask-compress has been updated from 1.5.0 to 1.8.0. The default compression algorithm in 1.5.0 was gzip, the latest version defaults to brotli. Running a benchmark test against runner in GCP showed that response times were 20-30% higher when using brotli. Switching the compression algorithm back to gzip returned the response times back to their previous range.

#### responses

responses has been fixed to 0.12.0 as the latest version of responses is incompatible with the moto library.  A fix to moto has been merged but not released yet - spulec/moto#3466.

### How to review 
Test that the app works, both locally and in GCP.

### Checklist


JamesGardiner (approved):

MebinAbraham (approved):
Haven't noticed any issues so ✅

PR #440, opened 2020-11-26
Unpin marshmallow library

ajmaddaford:
Unpins the marshmallow library.

`functools.partial` is not supported by mypy, so the lines that call the partial functions need have the 'type:ignore' comment - https://github.com/python/mypy/issues/1484

### How to review 
Check that the app works as expected.

### Checklist


MebinAbraham (approved):

app/data_models/app_models.py
```diff
@@ -58,7 +58,7 @@ class QuestionnaireStateSchema(Schema, DateTimeSchemaMixin):
     version = fields.Integer()
 
     @post_load
-    def make_model(self, data):
+    def make_model(self, data, many, **kwargs):
```
MebinAbraham, line 61: Based on the examples in the doc, it doesn't look like `many` is needed when `pass_many=False` which is the default as it is followed by `**kwargs`.

ajmaddaford: It shouldn't be, but I got an error:
```
strip_whitespace() got an unexpected keyword argument 'many'
```

app/data_models/app_models.py
```diff
@@ -47,7 +47,7 @@ class DateTimeSchemaMixin:
     updated_at = fields.DateTime()
 
     @pre_dump
-    def set_date(self, data, many, **kwargs):
+    def set_date(self, data, **kwargs):
```
MebinAbraham, line 50: Perfect example of our lint not working as expected 😞. Should be an `Unused argument`.

MebinAbraham (approved):

JamesGardiner (approved):

PR #445, opened 2020-12-01
Add question error panel

LJBabbage:
Add an error panel in the template for question errors. 

### How to review 
Check that a question is wrapped in a red border panel on error. This will affect both mutually exclusive and date range 

### Checklist


berroar (approved):
Error panel is now being displayed correctly

templates/partials/question.html
```diff
@@ -107,5 +110,23 @@
   {{ individual_response_guidance }}
   {{ question_definition }}
   {{ question_guidance }}
-  {{ question_answers }}
+
+  {% if question_error %}
+    {% set config = {
+      "text": question_error,
+      "id": question.id ~ '-error',
```
MebinAbraham, line 117: The question error anchor link needs to be updated to anchor to the error panel. It currently still links to the question itself. Checklist item 3 on the card for more info.

We already have a method for it which is just used fo answers at the moment.
https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/forms/questionnaire_form.py#L386-L387

MebinAbraham (changes requested):
Anchor link needs fixing but apart from that, the error panel seems to be as expected.

templates/partials/question.html
LJBabbage: I saw it linked to the question and thought that was fine, however I didn't appreciate that the answer links to the error panel and not the answer, so will bring it inline.

MebinAbraham (approved):

PR #456, opened 2020-12-07
Remove last viewed guidance for relationships

LJBabbage:
Last viewed guidance is appearing on relationships incorrectly. Because relationships always starts on the first relationship if resumed, the last viewed guidance panel will be displayed incorrectly. 

N.B I have updated the tests as this was identified as a bug, however it's equally viable to remove them all together as we are removing functionality. Will go with the consensus on this PR

### How to review 
Using a census or any relationship schema. Enter a couple of people into the list collector, complete the relationships formed, then go back to the list collector and add a new person, this should take you through relationships again, and the guidance box should not appear

### Checklist


tests/integration/questionnaire/test_questionnaire_relationships.py
```diff
@@ -18,23 +18,23 @@ def test_valid_relationship(self):
...
-    def test_get_first_relationship_with_resume(self):
+    def test_resume_should_not_show_last_viewed_guidance(self):
         self.launchSurvey("test_relationships")
         first_list_item_id = self.add_person("Marie", "Doe")
         second_list_item_id = self.add_person("John", "Doe")
         self.post({"anyone-else": "No"})
         self.get("/questionnaire/relationships?resume=True")
         self.assertInUrl(
-            f"/questionnaire/relationships/people/{first_list_item_id}/to/{second_list_item_id}/?resume=True"
+            f"/questionnaire/relationships/people/{first_list_item_id}/to/{second_list_item_id}"
```
ajmaddaford, line 28: This assertion is true with or without resuming, it should assert instead that `resume=True` is not in the url.

LJBabbage: I yeah, lol

MebinAbraham (approved):

JamesGardiner (approved):

ajmaddaford (approved):

PR #461, opened 2020-12-10
Fix 500 error on POST to /questionnaire/relationships

ajmaddaford:

Sending a POST request to `/questionnaire/relationships` was failing because it was falling into the non-relationship block handling code. Although we wouldn't expect this behaviour, it has been observed during testing, and should be handled properly. To fix this `POST` has been enabled on the `/questionnaire/relationships` endpoint, which consequently does the same as the `GET` and redirects to the first relationship.

### How to review 
I used the `test_relationships` schema, added some people and their relationships, then edited the HTML of a page in the browser Developer Tools to set `action="/questionnaire/relationships"` on the `form` element. 

### Checklist

### DIFF
--- a/app/routes/questionnaire.py
+++ b/app/routes/questionnaire.py
@@ -220,7 +220,7 @@ def block(schema, questionnaire_store, block_id, list_name=None, list_item_id=No
 
 @questionnaire_blueprint.route(
     "relationships/",
-    methods=["GET"],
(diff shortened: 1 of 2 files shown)


JamesGardiner (approved):

pricem14pc (approved):
As per test steps

PR #463, opened 2020-12-14
Use 'add link' feature in DS summary component

petechd:
The 'Add someone to this household' and 'Add a visitor' links in /sections/people-who-live-here-and-overnight-visitors/ are now removed and added as part of the summary component using the `summaryLink` array object as documented in [Design system](https://ons-design-system.netlify.app/components/summary/).

### How to review 
Go to first summary in HH schema and check if you get same behaviour as before the changes in DS were made.

### Checklist


petechd:
[DS v29.0.2 - Update runner to use new 'add link' feature in DS summary component (S)](https://trello.com/c/RX6A6F8Q/4354-ds-v2902-update-runner-to-use-new-add-link-feature-in-ds-summary-component-s)

templates/sectionsummary.html
```diff
@@ -37,10 +39,8 @@ <h2>{{summary.title}}</h2>
               {{ summary.empty_list_text }}
             </div>
           {% endif %}
-          {% if summary.add_link %}
-            <div>
-              <a data-qa="add-item-link" href="{{summary.add_link}}">{{summary.add_link_text}}</a>
-            </div>
+          {% if not summary.list.list_items %}
```
ajmaddaford, line 42: Could this be an `else` off he initial `if` statement on line 34?

templates/sectionsummary.html
```diff
@@ -37,10 +39,8 @@ <h2>{{summary.title}}</h2>
               {{ summary.empty_list_text }}
             </div>
           {% endif %}
-          {% if summary.add_link %}
```
ajmaddaford, line 40: The add link was optional in the previous version, does the new version need to handle this?

templates/sectionsummary.html
```diff
@@ -26,8 +26,10 @@ <h1>{{content.summary.title}}</h1>
...
     {% for summary in content.summary.custom_summary %}
       {% if summary.type == 'List' %}
-
-          <div {{'class="u-mb-l"' if not loop.last }} data-qa="{{summary.list_name}}-list-summary">
+          {% set add_link = summary.add_link %}
+          {% set add_link_text = summary.add_link_text %}
+          {% set list_name = summary.list_name %}
+          <div {{'class="u-mb-l"' if not loop.last }} data-qa="{{list_name}}-list-summary">
           <h2>{{summary.title}}</h2>
           {% if summary.list and summary.list.list_items %}
```
ajmaddaford, line 34: There is already a check for an empty list in `list-summary.html` that checks the length of `list_items`. We should consider moving the empty list logic out of this template and into that one.

MebinAbraham, line 34: Now that the summary component support everything we need for list collectors, I don't think any of the logic around list items are needed. The component support, list items, add links, empty placeholder text as well as the group header. So it can be updated to something like:
```jinja2
        {% if summary.type == 'List' %}
          {% set add_link = summary.add_link %}
          {% set add_link_text = summary.add_link_text %}
          {% set empty_list_text = summary.empty_list_text %}
          {% set summary_title = summary.title %}

          <div {{'class="u-mb-l"' if not loop.last }} data-qa="{{summary.list_name}}-list-summary">
          {% if summary.list %}
            {% set list = summary.list %}
          {% endif %}
          </div>
        {% endif %}
```

Then `list-summary.html` just needs updating to something like:
```jinja2

  {% if list.editable %}
    {% set headers = ["Name of person", "Action"] %}
    {% set rows = map_list_collector_config(
      list.list_items,
      "person",
      _("Change"),
      _("Change details for {item_name}"),
      _("Remove"),
      _("Remove {item_name}")
    ) %}
  {% else %}
    {% set headers = ["Name of person"] %}
    {% set rows = map_list_collector_config(
      list.list_items,
      "person"
    ) %}
  {% endif %}

  {% set groups_config = {
    "groupTitle": summary_title,
    "headers": headers,
    "rows": rows,
    "summaryLink": {"url": add_link, "text": add_link_text, "dataQa": "add-item-link"},
  } %}

  {% if empty_list_text %}
    {% do groups_config.update({"placeholderText": empty_list_text}) %}
  {% endif %}

  {{ onsSummary({
    "withinQuestion": true,
    "summaries": [
      {
        "groups": [
          groups_config
        ]
      }
    ],
    "classes": "u-mt-s u-mb-m"
  }) }}
```

This means there is one template for list summary and it is using the actual DS component as intended.

With the proposed changes, the padding of the continue button is slightly large and may need reducing.

petechd: @MebinAbraham I used your code from above, extra space on the summary was caused by `u-mb-m` class used in our onsSummary I believe. So I removed that class and spacing is the same as in staging now.

petechd: This has been overhauled using Mebin's code below, so this doesn't use length of `list_items` anymore.

templates/sectionsummary.html
petechd: This has been removed and no longer applies..

templates/sectionsummary.html
petechd: This is removed now👍

templates/partials/summary/list-summary.html
```diff
@@ -1,38 +1,42 @@
...
-            "headers": headers,
-            "rows": rows,
-          }
-        ]
-      }
-    ],
-    "classes": "u-mt-s u-mb-m"
-  }) }}
+{% set groups_config = {
+  "groupTitle": summary_title,
```
MebinAbraham, line 22: I think in context of this file, we prob should update this to just `title` or `group_title`.

templates/partials/summary/list-summary.html
```diff
@@ -1,38 +1,42 @@
...
+  "headers": headers,
+  "rows": rows,
+  "summaryLink": {"url": add_link, "text": add_link_text, "dataQa": "add-item-link"},
+} %}
 
+{% if empty_list_text %}
+  {% do groups_config.update({"placeholderText": empty_list_text}) %}
 {% endif %}
```
MebinAbraham, line 30: The empty list text and the title is not relevant to list collector pages, so either we update both of them to be an `if` statement to update the config if it exists or we just pass it to the DS with null values which is handled as expected.

So we could remove the if and just simplify to this where all config is always passed even when it may not exist and let the component handle it.
```jinja2
{% set groups_config = {
  "groupTitle": summary_title,
  "headers": headers,
  "rows": rows,
  "summaryLink": {"url": add_link, "text": add_link_text, "dataQa": "add-item-link"},
  "placeholderText": empty_list_text
} %}
```

Thoughts? @petechd  @ajmaddaford

MebinAbraham (commented):
Looks good, just minor things to consider.

templates/partials/summary/list-summary.html
petechd: I will simplify it since Jon told me that we will never use `placeholderText` element in census.

petechd: This is simpler, since placeholderText won't be created when there is no value I would go with this approach.

templates/partials/summary/list-summary.html
petechd: Changed it now, thanks!

templates/partials/summary/list-summary.html
```diff
@@ -1,38 +1,39 @@
...
-    ],
-    "classes": "u-mt-s u-mb-m"
-  }) }}
+{% set groups_config = {
+  "groupTitle": title,
+  "headers": headers,
+  "rows": rows,
+  "summaryLink": {"url": add_link, "text": add_link_text, "dataQa": "add-item-link"},
+  "placeholderText": empty_list_text,
+} %}
```
MebinAbraham, line 27: This could just now be inline since there is no additional logic.

templates/partials/summary/list-summary.html
```diff
@@ -1,38 +1,39 @@
...
-            "headers": headers,
-            "rows": rows,
-          }
-        ]
-      }
-    ],
-    "classes": "u-mt-s u-mb-m"
-  }) }}
+{% set groups_config = {
+  "groupTitle": title,
```
MebinAbraham, line 22: Looks like this can't be called title because "title" already exists, and you end up with 2 titles. So need to be `list_title` or something unique.

MebinAbraham (approved):
Approving as I will be away, and it looks almost there. The comment about title needs fixing. A second reviewer can confirm it has been done.

templates/partials/summary/list-summary.html
petechd: Inline now, thanks!

templates/partials/summary/list-summary.html
petechd: Changed, thanks!

templates/partials/summary/list-summary.html
```diff
@@ -1,38 +1,37 @@
...
+{{ onsSummary({
+  "withinQuestion": true,
+  "summaries": [
+    {
+      "groups": [
+        {
+          "groupTitle": list_title,
+          "headers": headers,
+          "rows": rows,
+          "summaryLink": {"url": add_link, "text": add_link_text, "dataQa": "add-item-link"},
```
JamesGardiner, line 30: There is an issue with the Design System implementation of this summaryLink component - instead of `dataQa` it should take a key `attributes` that maps to html element attributes - for example:

```
"attributes": {
  "data-qa": "add-item-link"
}
```

We should wait for that change to merge and update this line accordingly before merging this change.

templates/partials/summary/list-summary.html
```diff
@@ -1,38 +1,43 @@
...
+          "groupTitle": list_title,
+          "headers": headers,
+          "rows": rows,
+          "summaryLink": {
+            "url": add_link,
+            "text": add_link_text,
+            "attributes": {
+              "data-qa": "add-item-link"
+            }
+          },
```
ajmaddaford, line 36: As `summaryLink` is always added, when there is no `add_link` or `add_link_text` the following html is incorrectly added to the html
```html
<div class="summary__add-link u-bt"><a data-qa="add-item-link" href></a></div>
```

petechd: I've added `add_link` conditional now, sets `summaryLink` as an attribute of group config if the link exists. Thanks for help!

petechd:
I've added div with a margin class to `question.html` where `list-summary.html` is called so the issue visible in `test_list_summary_on_question` test schema is resolved. Thanks @ajmaddaford!

templates/partials/summary/list-summary.html
petechd: Done, thanks!

ajmaddaford (approved):

PR #467, opened 2020-12-15
Refactor URLSafeSerializer

JamesGardiner:
Removes the URLParamSerializer in favour of a function that returns an instance of URLSafeSerializer.

Uses the `current_app.secret_key` as the signing key, and the current session's ID as the salt.

### How to review 
Use a schema that requires serialization of url parameters (test_feedback_email_confirmation is a good one to use).

Enter your param that will be serialized on the relevant page (e.g. email or mobile number). Test it is serialised and deserialised on the relevant pages.

Copy the serialised param and use it in a new session on the same page. A 500 error should be served as a result of a BadSignature exception.


tests/app/helpers/test_url_safe_serializer.py
```diff
@@ -0,0 +1,68 @@
...
+
+        with self.app_request_context("/status"):
+            with patch(
+                "app.authentication.authenticator.get_session_store",
+                return_value=self.session_store,
+            ):
+                self.session_store.create(
+                    "eq-session-id", "user_id", self.session_data, self.expires_at
+                )
+                cookie_session[EQ_SESSION_ID] = "eq-session-id"
```
ajmaddaford, line 45: I think it's a bit confusing to use the value of `EQ_SESSION_ID` as the session id.

tests/app/helpers/test_url_safe_serializer.py
```diff
@@ -0,0 +1,68 @@
...
+        test_string = "test-string"
+
+        with self.app_request_context("/status"):
+            with patch(
+                "app.authentication.authenticator.get_session_store",
+                return_value=self.session_store,
+            ):
+                self.session_store.create(
+                    "eq-session-id", "user_id", self.session_data, self.expires_at
+                )
```
ajmaddaford, line 44: I don't think these lines are needed. Neither the code being tested or the `/status` endpoint use the session.

tests/app/helpers/test_url_safe_serializer.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def test_safe_serializer_raises_bad_signature_when_salts_differ(self):
+        with self.app_request_context("/status"):
+            with patch(
+                "app.authentication.authenticator.get_session_store",
+                return_value=self.session_store,
+            ):
+                self.session_store.create(
+                    "eq-session-id", "user_id", self.session_data, self.expires_at
+                )
```
ajmaddaford, line 60: Same as above, I think these lines can go.

ajmaddaford (commented):
I think there should also be an integration test to test the application behaviour when the serialisation fails.

app/data_models/fulfilment_request.py
```diff
@@ -1,9 +1,9 @@
-import json
 from abc import ABC, abstractmethod
 from datetime import datetime
 from typing import Mapping
 from uuid import uuid4
 
+import simplejson as json
```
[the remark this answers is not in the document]
JamesGardiner: No, they were to create consistency in Runner. Happy to put into a separate PR?

app/storage/storage_encryption.py
```diff
@@ -12,12 +12,12 @@
...
-        if user_ik is None:
-            raise ValueError("User ik must be set")
-        if pepper is None:
-            raise ValueError("Pepper must be set")
+        if not user_id:
+            raise ValueError("user_id must be set")
+        if not user_ik:
+            raise ValueError("user_ik must be set")
+        if not pepper:
+            raise ValueError("pepper must be set")
```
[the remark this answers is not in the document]
JamesGardiner: Yep, my mistake. I'll revert.

tests/app/helpers/test_url_safe_serializer.py
JamesGardiner: Fixed - now using uuid4 strings

tests/app/helpers/test_url_safe_serializer.py
JamesGardiner: Yes, works fine without patching the session. I've removed in both tests

LJBabbage (approved):
Looks good to me

JamesGardiner:
@ajmaddaford integration test for serialisation failure added.

app/data_models/fulfilment_request.py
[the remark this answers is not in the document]
ajmaddaford: Yes 👍

tests/app/helpers/test_url_safe_serializer.py
```diff
@@ -0,0 +1,55 @@
...
+            language_code=None,
+            launch_language_code=None,
+            survey_url=None,
+            ru_name="ru_name",
+            ru_ref="ru_ref",
+            case_id="case_id",
+            questionnaire_id="questionnaire_id",
+        )
+        self.session_store = SessionStore("user_ik", "pepper", "eq_session_id")
+        self.expires_at = datetime.now(tzutc()) + timedelta(seconds=5)
```
ajmaddaford, line 33: I think this can all go now that we're not using a session.

tests/integration/integration_test_case.py
```diff
@@ -193,6 +193,15 @@ def sign_out(self):
...
         return self.get(selected["href"])
 
+    def exit(self):
+        """
+        GETs the sign-out url from the exit button. Does not follow the external
+        redirect.
+        """
+        url = self.getHtmlSoup().find("a", {"name": "btn-exit"})["href"]
+        environ, response = self._client.get(url, as_tuple=True, follow_redirects=False)
+        self._cache_response(environ, response)
```
ajmaddaford, line 203: We should make `follow_redirects` a param of `get` (defaulted to `True`). This could then be:
```suggestion
        self.client.get(url, follow_redirects=False)
```

JamesGardiner: Yep, that's a cleaner implementation. I'll update.

ajmaddaford (approved):

PR #480, opened 2021-01-04
Fix html escaping when using striptags issue

ajmaddaford:
An issue was identified recently that allows HTML injection via input values used in the HTML attributes `data-ga-action` and `aria-label`. This was due to the use of the `striptags` Jinja filter, which in addition to stripping HTML tags, also un-escapes strings.

Runner was using `striptags` to remove HTML from error messages passed to Google Analytics. This was because the error message could contain the question text (used for the visually hidden part of the error message in mandatory radio and checkbox answers). This however, highlights another issue - the question text can contain user provided information. To fix this, and mitigate the risk for future error message content, the `data-ga-action` now sends the question type, answer type, or "Confirmation Email" depending on where it is set, rather than the error message.

The design system templates used `striptags` to remove HTML from question text used in aria labels on summaries. This was removed from the design system in https://github.com/ONSdigital/design-system/pull/1257 - the expectation is that runner should not provide HTML in aria label strings. The aria label issue affected runner in two places - list collector summaries and final summaries. As the item names used in list collectors never use HTML, the design system change fixed this. For the final summary, as the question text is used in the `Change` and `Remove` link aria labels, HTML tags needed to be stripped from the question text. This has been done in the jinja filters that build the final summary.

### How to review 
- Check that `striptags` is only used in the jinja filter for the final summary
- Verify that the HTML is removed and content escaped properly
- Ensure that you've run `make load-design-system-templates` or `make build` to load the latest design system templates.

#### Test google analytics fix

For this you will need to generate an error that contains the question text with a user provided value placeholder. The `test_summary` schema has been updated for this:
- Enter the value `BSI">` for the dessert question and submit
- Submit the `dessert-confirmation` quesiton without answering
- The page (with user error) should render properly, you should not see any HTML containing `data-ga-action`  in the rendered page

#### Test aria label fix

Continue the journey started above to the summary. The page should be rendered properly with no corrupt HTML. The aria labels should contain the question text with appropriately escaped HTML for the dessert answer.

The list collector summary can be tested by using any list collector schema and using the `BSI">` value for the first name.

### Checklist


app/jinja_filters.py
```diff
@@ -14,6 +14,16 @@
 blueprint = flask.Blueprint("filters", __name__)
 
 
+def mark_safe(context, value):
+    if context.autoescape:
+        value = Markup(value)
+    return value
```
JamesGardiner, line 20:
```suggestion
    return Markup(value) if context.autoescape else value
```

JamesGardiner (approved):

MebinAbraham (approved):
Seems to be as expected. I did notice that, for dev when you use the dump/ endpoint the content is not escaped so look messed up when using a browser. But since it is only for dev and technically because it is a JSON response, it should be fine.

PR #487, opened 2021-01-07
Allow multiple lookups selection

petechd:
This adds support to multiple lookups selection in [autosuggest component](https://ons-design-system.netlify.app/components/autosuggest/). This feature is described in [Trello card](https://trello.com/c/RqJlAYzx). It requires this [validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/86) to merge first.

### How to review 
New functional test was added and textfield suggestions schema was expanded. Run new functional test, run through `test_textfield_suggestions.json` manually.

### Checklist

### DIFF
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=lookups-allow-multiple-selection
+tag=latest
 docker pull onsdigital/eq-questionnaire-validator:$tag
 docker run -d -p 5001:5000 "onsdigital/eq-questionnaire-validator:$tag"
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=lookups-allow-multiple-selection
 docker pull onsdigital/eq-questionnaire-validator:$tag
 docker run -d -p 5001:5000 "onsdigital/eq-questionnaire-validator:$tag"
--- a/templates/partials/answers/textfield.html
+++ b/templates/partials/answers/textfield.html
@@ -43,7 +43,7 @@
     "ariaNResults": "There are {n} suggestions available.",
     "ariaLimitedResults": "Results have been limited to 10 suggestions. Type more characters to improve your search.",
     "autosuggestData": answer.suggestions.url,
-    "allowMultiple": answer.suggestions.allow_multiple | lower if answer.suggestions.allow_multiple is defined,
+    "allowMultiple": "true" if answer.suggestions.allow_multiple else "false"
   }) %}
   {% set config = config | setAttribute("autocomplete", "off") %}
 {% endif %}
--- a/templates/partials/answers/textfield.html
+++ b/templates/partials/answers/textfield.html
@@ -30,7 +30,7 @@
   }) %}
 {% endif %}
 
-{% if answer.suggestions_url %}
+{% if answer.suggestions.url %}
   {% set config = config | setAttribute("autosuggest", {
     "instructions": _("Use up and down keys to navigate suggestions once you\'ve typed more than two characters. Use the enter key to select a suggestion. Touch device users, explore by touch or with swipe gestures."),
     "moreResults": _("Continue entering to improve suggestions"),
@@ -42,8 +42,8 @@
     "ariaOneResult": "There is one suggestion available.",
     "ariaNResults": "There are {n} suggestions available.",
     "ariaLimitedResults": "Results have been limited to 10 suggestions. Type more characters to improve your search.",
-    "autosuggestData": answer.suggestions_url,
-    "allowMultiple": answer.allow_multiple | lower if answer.allow_multiple is defined,
+    "autosuggestData": answer.suggestions.url,
+    "allowMultiple": answer.suggestions.allow_multiple | lower if answer.suggestions.allow_multiple is defined,
   }) %}
   {% set config = config | setAttribute("autocomplete", "off") %}
 {% endif %}
--- a/templates/partials/answers/textfield.html
+++ b/templates/partials/answers/textfield.html
@@ -30,7 +30,7 @@
   }) %}
 {% endif %}
 
-{% if answer.suggestions.url %}
+{% if answer.suggestions %}
   {% set config = config | setAttribute("autosuggest", {
     "instructions": _("Use up and down keys to navigate suggestions once you\'ve typed more than two characters. Use the enter key to select a suggestion. Touch device users, explore by touch or with swipe gestures."),
     "moreResults": _("Continue entering to improve suggestions"),
--- a/templates/partials/answers/textfield.html
+++ b/templates/partials/answers/textfield.html
@@ -42,7 +42,8 @@
     "ariaOneResult": "There is one suggestion available.",
     "ariaNResults": "There are {n} suggestions available.",
     "ariaLimitedResults": "Results have been limited to 10 suggestions. Type more characters to improve your search.",
-    "autosuggestData": answer.suggestions_url
+    "autosuggestData": answer.suggestions_url,
+    "allowMultiple": answer.allow_multiple | lower if answer.allow_multiple is defined,
   }) %}
   {% set config = config | setAttribute("autocomplete", "off") %}
 {% endif %}
--- a/test_schemas/en/test_textfield_suggestions.json
+++ b/test_schemas/en/test_textfield_suggestions.json
@@ -47,6 +47,26 @@
                             "routing_rules": []
                         },
                         {
+                            "type": "Question",
+                            "id": "country-block-multiple-selections",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "country-multiple-selections-answer",
+                                        "label": "What is your country of birth?",
+                                        "mandatory": false,
+                                        "suggestions_url": "https://cdn.eq.census-gcp.onsdigital.uk/data/v5.0.0/gb/en/countries-of-birth.json",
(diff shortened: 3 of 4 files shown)


petechd:
[FEATURE: Socio Demographic look-ups - Allow multiple selections (S)](https://trello.com/c/RqJlAYzx/4407-feature-socio-demographic-look-ups-allow-multiple-selections-s)

test_schemas/en/test_textfield_suggestions.json
```diff
@@ -46,6 +46,26 @@
...
+                                        "suggestions_url": "https://cdn.eq.census-gcp.onsdigital.uk/data/v5.0.0/gb/en/countries-of-birth.json",
+                                        "allow_multiple": true,
+                                        "type": "TextField"
+                                    }
+                                ],
+                                "id": "country-of-birth-multiple-selections-question",
+                                "title": "Title",
+                                "type": "General"
+                            },
+                            "routing_rules": []
```
MebinAbraham, line 67: Should take out the empty `"routing_rules": []`.

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=lookups-allow-multiple-selection
```

tests/functional/spec/textfield_suggestions.spec.js
```diff
@@ -8,3 +9,22 @@ describe("Suggestions", () => {
     expect($$(".js-autosuggest-listbox li").length).to.not.equal(0);
   });
 });
+
+describe("Suggestions", () => {
+  it("Given I open a textfield with a suggestions url, when I have entered text and picked suggestion from a list, then after typing more text it will show new suggestions", () => {
```
MebinAbraham, line 14: This should mention it allows multiple suggestion.

MebinAbraham (dismissed):
Feature works as expected. There are some issues with the DS component, will raise will Matt.

test_schemas/en/test_textfield_suggestions.json
petechd: Removed `"routing_rules": [].` now, thanks!

tests/functional/spec/textfield_suggestions.spec.js
petechd: Added to description, thanks!

test_schemas/en/test_textfield_suggestions.json
```diff
@@ -36,15 +36,38 @@
                                         "id": "country-answer",
                                         "label": "What is your country of birth?",
                                         "mandatory": false,
-                                        "suggestions_url": "https://cdn.eq.census-gcp.onsdigital.uk/data/v5.0.0/gb/en/countries-of-birth.json",
+                                        "suggestions": {
+                                            "url": "https://cdn.eq.census-gcp.onsdigital.uk/eq-lookup-suggestions-data/v5.3.0/gb/en/countries-of-birth.json",
+                                            "allow_multiple": true
```
ajmaddaford, line 41: Should this be `false`?

petechd: Yes, sorry, only second question is tested for multiple lookups.

test_schemas/en/test_textfield_suggestions.json
```diff
@@ -36,15 +36,38 @@
...
+                            }
+                        },
+                        {
+                            "type": "Question",
+                            "id": "country-block-multiple-selections",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "country-multiple-selections-answer",
+                                        "label": "What is your country of birth?",
```
ajmaddaford, line 58: It was already wrong in this schema, but we should change it now. "What is your country of birth?" should be the question text and this should be something like "Country". This should also be fixed for the first question.

petechd: Fixed now, thanks!

test_schemas/en/test_textfield_suggestions.json
```diff
@@ -34,17 +34,40 @@
...
+                                        "mandatory": false,
+                                        "suggestions": {
+                                            "url": "https://cdn.eq.census-gcp.onsdigital.uk/eq-lookup-suggestions-data/v5.3.0/gb/en/countries-of-birth.json",
+                                            "allow_multiple": true
+                                        },
+                                        "type": "TextField"
+                                    }
+                                ],
+                                "id": "country-of-birth-multiple-selections-question",
+                                "title": "What is your country of birth?",
```
ajmaddaford, line 68: I've just noticed that the question doesn't match what we're asking for - multiple countries. Maybe the question could be "What countries have you visited?" (the answer label should then be "Countries").

ajmaddaford, line 68: I've just noticed that the question doesn't match what we're asking for - multiple countries. Also, the multiple selections doesn't work properly with countries of birth as some of the values already contain commas. We should use either the national identities list or passport countries.

Maybe the question could be "What passports do you hold?" (the answer label should then be "Passports") and use the passport countries list.

petechd: It uses passport countries list now, new ids, label and title were added, thanks!

tests/functional/spec/textfield_suggestions.spec.js
```diff
@@ -8,3 +9,22 @@ describe("Suggestions", () => {
...
+    browser.keys("Ita");
+    $("#multiple-country-answer-listbox li").waitForDisplayed();
+    expect($$(".js-autosuggest-listbox li").length).to.equal(2);
+    $("#multiple-country-answer-listbox__option--0").click();
+    $(MultipleSuggestionsPage.multipleCountry()).click();
+    // Browser needs to pause before typing starts, otherwise this sequence is executed almost instantly
+    browser.pause(500);
+    browser.keys(" United");
+    $("#multiple-country-answer-listbox li").waitForDisplayed();
+    expect($$(".js-autosuggest-listbox li").length).to.equal(3);
```
ajmaddaford, line 28: Fixing on the number of suggestions displayed makes the test fragile, and it shouldn't be needed as all we're concerned with is whether the suggestions are displayed or not.

petechd: It checks if list of suggestions is visible now and if one of suggestions can be picked. For some reason functional tests cannot get list of passport countries during build process and fails.

templates/partials/answers/textfield.html
```diff
@@ -42,7 +42,8 @@
     "ariaOneResult": "There is one suggestion available.",
     "ariaNResults": "There are {n} suggestions available.",
     "ariaLimitedResults": "Results have been limited to 10 suggestions. Type more characters to improve your search.",
-    "autosuggestData": answer.suggestions_url
+    "autosuggestData": answer.suggestions.url,
+    "allowMultiple": answer.suggestions.allow_multiple | lower if answer.suggestions.allow_multiple is defined,
```
MebinAbraham, line 46: The comma is invalid as this is the last statement, Jinja2 is too nice and doesn't fail. Given the value will always be a boolean, "is defined" is not needed, and we could remove the "lower". Also, given we send "Undefined" if allow multiple isn't defined, I think an else clause should be added for completion.
```suggestion
    "allowMultiple": "true" if answer.suggestions.allow_multiple else "false"
```

MebinAbraham (approved):

tests/functional/spec/textfield_suggestions.spec.js
```diff
@@ -8,3 +9,25 @@ describe("Suggestions", () => {
...
+describe("Suggestions", () => {
+  it("Given I open a textfield with a suggestions url that allows multiple suggestions, when I have entered text and picked suggestion from a list, then after typing more text it will show new suggestions", () => {
+    browser.openQuestionnaire("test_textfield_suggestions.json");
+    $(SuggestionsPage.country()).setValue("United States of America");
+    $(SuggestionsPage.submit()).click();
+    $(MultipleSuggestionsPage.multipleCountry()).click();
+    browser.pause(500);
+    browser.keys("Ita");
+    $("#multiple-country-answer-listbox li").waitForExist();
+    $("#multiple-country-answer-listbox__option--0").click();
```
ajmaddaford, line 22: I don't think we should need to pause. The only other place we use pause is for the timeout test, where it is necessary. I understand that without this it was failing the functional test in Github actions, but maybe if tried using `waitForExist` on the option?

```suggestion
    browser.keys("Ita");
    $("#multiple-country-answer-listbox__option--0").waitForExist();
    $("#multiple-country-answer-listbox__option--0").click();
```

petechd: I tried and functional test failed..

ajmaddaford: Sorry, it does look like a workaround is needed as the javascript doesn't seem to work if the box is selected immediately. Ideally we would hook into a DOM change when the Javascript initialises. I was hoping to use `autosuggest-input--initialised ` (https://github.com/ONSdigital/design-system/blob/6ebe5e75c41daf8324cf2be1ebce388567f7e625/src/components/input/autosuggest/autosuggest.ui.js#L118) but that is set before the event listeners are initialised.

For now I think we should continue with the pause. Can we clarify the comment so that it's clear why the pause exists, so instead of "otherwise this sequence is executed almost instantly" we could say "to allow for the autosuggest Javascript to initialise".

ajmaddaford (approved):

tests/functional/spec/textfield_suggestions.spec.js
petechd: Added corrected comment to both `browser.pause()` instances. Thanks!

PR #489, opened 2021-01-08
Purge sessions from Datastore

LJBabbage:
This PR adds the yaml needed to purge eq-sessions from datastore.

### How to review 
Make sure you have a GCP deployment with sessions in (if you haven't you will need to create one and launch a survey, you will also need to remove --set=00:00:00 from the script, as it is designed to purge before 12am on the day its run). Once run check to see a job is created in Dataflow and then the current sessions are removed

### Note:
The Dataflow API will probably not be enabled in your GCP environment, to enable you will need to run your infrastructure again with it pointed at the branch **[enable-dataflow-service](https://github.com/ONSdigital/eq-terraform-gcp/pull/26)** or turn it on manually


ci/purge_sessions_from_datastore.yaml
```diff
@@ -0,0 +1,27 @@
...
+      $SERVICE_ACCOUNT_JSON
+      EOL
+
+      gcloud auth activate-service-account --key-file "${GOOGLE_APPLICATION_CREDENTIALS}"
+      gcloud config set project "${PROJECT_ID}"
+      gcloud dataflow jobs run purgesession \
+      --region europe-west2 \
+      --gcs-location gs://dataflow-templates/2020-11-30-00_RC00/Datastore_to_Datastore_Delete \
+      --parameters \
+      datastoreReadGqlQuery="SELECT * FROM \`eq-session\` WHERE expires_at < `date --set=00:00:00 +%s`",\
```
MebinAbraham, line 25: If we are getting midnight then we should use `--date` as `--set` sets the system clock. We should ensure we work with UTC also.
```sh
date --utc --date "00:00:00" +%s
```

ci/purge_sessions_from_datastore.yaml
```diff
@@ -0,0 +1,27 @@
...
+      export GOOGLE_APPLICATION_CREDENTIALS=/root/gcloud-service-key.json
+      cat >$GOOGLE_APPLICATION_CREDENTIALS <<EOL
+      $SERVICE_ACCOUNT_JSON
+      EOL
+
+      gcloud auth activate-service-account --key-file "${GOOGLE_APPLICATION_CREDENTIALS}"
+      gcloud config set project "${PROJECT_ID}"
+      gcloud dataflow jobs run purgesession \
+      --region europe-west2 \
+      --gcs-location gs://dataflow-templates/2020-11-30-00_RC00/Datastore_to_Datastore_Delete \
```
MebinAbraham, line 23: The card specifies that the dataflow template version should be parameterised.

LJBabbage: Can't see the point in that, but I will change it

JamesGardiner:
I may be doing something wrong here, but running the following command did not delete any eq-session data for me:

```sh
HTTPS_PROXY=localhost:8118 PROJECT_ID=james-test-292314 DATAFLOW_TEMPLATE_VERSION=2020-11-30-00_RC00 EXPIRATION_TIME_OFFSET=3600 fly -t census-prod execute --config ci/purge_expired_sessions_from_datastore.yaml
```

JamesGardiner:
> I may be doing something wrong here, but running the command did not delete any eq-session data for me

My fault - had old sessions that aren't indexed. Retested and worked fine 👍🏻

JamesGardiner (approved):

ci/purge_expired_sessions_from_datastore.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - -exc
+    - |
+      export GOOGLE_APPLICATION_CREDENTIALS=/root/gcloud-service-key.json
+      cat >$GOOGLE_APPLICATION_CREDENTIALS <<EOL
+      $SERVICE_ACCOUNT_JSON
+      EOL
+
+      gcloud auth activate-service-account --key-file "${GOOGLE_APPLICATION_CREDENTIALS}"
+      gcloud config set project "${PROJECT_ID}"
+      gcloud dataflow jobs run purgesession \
```
MebinAbraham, line 23: I think this job should be blocking until the dataflow job has been completed then it should report the status back. At the moment, this would just run and without checking GCP console ourselves, we wouldn't know if it passed or failed. We should do what we do for backups and restore and report the state. The task should pass/fails based on the job state.

LJBabbage: Which is the blocking part in backups and restore in the script?

LJBabbage: The state of the job will go to 'JOB_STATE_PENDING' and at best takes 2 minutes to change to failed or succeeded, when census is running this could take 30+ minutes. I am not sure polling the job every few seconds to see the job state is a great idea

MebinAbraham: The datastore `import` and `export` commands are synchronous by default (blocking), they do have an `--async` flag, but we chose not to use that, so we can report the progress. The dataflow by default is async so we won't know the state without polling. I am not sure if dataflow has a flag that will block it (worth checking). But even if we have to poll it, I don't see any issues with that, it would be the recommended way to automate it if it has no blocking option. We could easily just poll it every 60 seconds. 

I don't think it's very useful to have a CI job that triggers a job at 4am which will pass and report to Slack as passed but may have failed the actual job in the background. Without checking the state, someone will have to check the state manually every time to see if the job was actually successful.

LJBabbage: Dataflow doesn't have an export command, so I am going to have to call the describe every minute

ci/purge_expired_sessions_from_datastore.yaml
```diff
@@ -0,0 +1,44 @@
...
+      gcloud config set project "${PROJECT_ID}"
+
+      DATAFLOW_JOB=$(gcloud dataflow jobs run purgesession \
+      --region europe-west2 \
+      --gcs-location gs://dataflow-templates/"${DATAFLOW_TEMPLATE_VERSION}"/Datastore_to_Datastore_Delete \
+      --parameters \
+      datastoreReadGqlQuery="SELECT * FROM \`eq-session\` WHERE expires_at < $((`date --utc +%s` - EXPIRATION_TIME_OFFSET)) ",\
+      datastoreReadProjectId="${PROJECT_ID}",\
+      datastoreDeleteProjectId="${PROJECT_ID}" \
+      --format="value(id.scope())")
```
MebinAbraham, line 31: Can the formatting be updated to make it more readable?.

ci/purge_expired_sessions_from_datastore.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - |
+      export GOOGLE_APPLICATION_CREDENTIALS=/root/gcloud-service-key.json
+      cat >$GOOGLE_APPLICATION_CREDENTIALS <<EOL
+      $SERVICE_ACCOUNT_JSON
+      EOL
+
+      gcloud auth activate-service-account --key-file "${GOOGLE_APPLICATION_CREDENTIALS}"
+      gcloud config set project "${PROJECT_ID}"
+
+      DATAFLOW_JOB=$(gcloud dataflow jobs run purgesession \
```
MebinAbraham, line 24: The job name should be snake case or pascal case.

ci/purge_expired_sessions_from_datastore.yaml
```diff
@@ -0,0 +1,44 @@
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: google/cloud-sdk
+params:
+  SERVICE_ACCOUNT_JSON: ((gcp.service_account_json))
+  PROJECT_ID:
+  DATAFLOW_TEMPLATE_VERSION:
+  EXPIRATION_TIME_OFFSET:
```
MebinAbraham, line 10: I think these should be defaulted to make fly executing easier as the values doesn't matter too much. Alternatively, the README should link to where you can find the dataflow template versions.

ci/README.md
```diff
@@ -59,3 +59,18 @@ fly -t <target-concourse> execute \
...
+```sh
+PROJECT_ID=<project_id> \
+DATAFLOW_TEMPLATE_VERSION=<dataflow_template_version> \
+EXPIRATION_TIME_OFFSET=<expiration_time_offset> \
+fly -t <target-concourse> execute \
+  --config ci/purge_expired_sessions_from_datastore.yaml
+```
+
+- `DATAFLOW_TEMPLATE_VERSION` is the template version on GCP you want to use
+- `EXPIRATION_TIME_OFFSET` is the offset in seconds expired sessions will be purged from (i.e 3600 would delete all sessions more than an hour old)
```
MebinAbraham, line 76: Any reasons seconds were used or maybe not hours?. Either way, I think the variable should indicate if it is hours or seconds similar to how we do it with runner variables. It makes it less vague as when I look at the eq-pipelies repo variables file, I can't tell this is in seconds. Maybe `EXPIRATION_TIME_OFFSET_IN_SECONDS` for clarity.

MebinAbraham (commented):
Looks good, task executing as expected. Few minor comments.

ci/README.md
LJBabbage: Thought about hours, but came to the conclusion that there might be a case where we might want to remove sessions closer to the actually time.

MebinAbraham: Ya fair enough 👍

ci/purge_expired_sessions_from_datastore.yaml
LJBabbage: Bash is never pretty, and there is not much I can do here

ci/purge_expired_sessions_from_datastore.yaml
MebinAbraham: D'oh, I think I was actually getting at kebab-case but oh well, better than one previous so 👍 .

MebinAbraham (approved):
👍

ci/README.md
```diff
@@ -59,3 +59,19 @@ fly -t <target-concourse> execute \
...
+PROJECT_ID=<project_id> \
+DATAFLOW_TEMPLATE_VERSION=<dataflow_template_version> \
+EXPIRATION_TIME_OFFSET_IN_SECONDS=<expiration_time_offset_in_seconds> \
+fly -t <target-concourse> execute \
+  --config ci/purge_expired_sessions_from_datastore.yaml
+```
+There are defaults for both DATAFLOW_TEMPLATE_VERSION and EXPIRATION_TIME_OFFSET_IN_SECONDS if not set
+
+- `DATAFLOW_TEMPLATE_VERSION` is the template version on GCP you want to use
+- `EXPIRATION_TIME_OFFSET` is the offset in seconds expired sessions will be purged from (i.e 3600 would delete all sessions more than an hour old)
```
MebinAbraham, line 77:
```suggestion
- `EXPIRATION_TIME_OFFSET_IN_SECONDS ` is the offset in seconds expired sessions will be purged from (i.e 3600 would delete all sessions more than an hour old)
```

ci/purge_expired_sessions_from_datastore.yaml
```diff
@@ -0,0 +1,44 @@
...
+      EOL
+
+      gcloud auth activate-service-account --key-file "${GOOGLE_APPLICATION_CREDENTIALS}"
+      gcloud config set project "${PROJECT_ID}"
+
+      DATAFLOW_JOB=$(gcloud dataflow jobs run purge_session \
+      --region europe-west2 \
+      --gcs-location gs://dataflow-templates/"${DATAFLOW_TEMPLATE_VERSION}"/Datastore_to_Datastore_Delete \
+      --parameters \
+      datastoreReadGqlQuery="SELECT * FROM \`eq-session\` WHERE expires_at < $((`date --utc +%s` - EXPIRATION_TIME_OFFSET_IN_SECONDS)) ",\
```
MebinAbraham, line 28:
```suggestion
      datastoreReadGqlQuery="SELECT * FROM \`eq-session\` WHERE expires_at < $((`date --utc +%s` - EXPIRATION_TIME_OFFSET_IN_SECONDS))",\
```

MebinAbraham (changes requested):
minor

MebinAbraham (approved):

PR #490, opened 2021-01-11
Add cloud task publisher to runner

berroar:

Runner should now publish confirmation email requests to a Cloud Task queue rather than a PubSub endpoint.

### How to review 

Use the `test_confirmation_email` schema to check the changes

### Checklist


berroar:
[EQ Runner Cloud Task Publisher (S)](https://trello.com/c/sH5GGx7f/4443-eq-runner-cloud-task-publisher-s)

tests/integration/routes/test_confirmation_email.py
```diff
@@ -1,14 +1,18 @@
...
 from tests.integration.integration_test_case import IntegrationTestCase
 
 
 class TestEmailConfirmation(IntegrationTestCase):
-    def setUp(self):
-        settings.CONFIRMATION_EMAIL_LIMIT = 2
-        super().setUp()
+    def setUp(self) -> None:
+        with patch(
+            "app.publisher.publisher.google.auth._default._get_explicit_environ_credentials",
```
JamesGardiner, line 11: I think this needs to be added to the non-integration test case as well

app/cloud_tasks/exceptions/__init__.py
```diff
@@ -0,0 +1,2 @@
+class CloudTaskCreationFailed(Exception):
```
MebinAbraham, line 1: Can we standardise where we place custom exceptions?. For cloud tasks, it's within it's own exceptions module, for Pub/Sub it is within `exceptions.py` and for submitter it is a file for a specific exception. It would be nicer if we can make all these consistent. Personally, I think using `exceptions.py` in the working module is fine. This is what libs like redispy, google auth etc does.

app/views/handlers/confirmation_email.py
```diff
@@ -63,16 +64,13 @@ def get_page_title(self):
         return self.page_title
 
     def _publish_fulfilment_request(self):
-        topic_id = current_app.config["EQ_SUBMISSION_CONFIRMATION_TOPIC_ID"]
```
MebinAbraham, line 66: Reference to `EQ_SUBMISSION_CONFIRMATION_TOPIC_ID` can be removed from deployment.yaml

app/setup.py
```diff
@@ -342,6 +345,15 @@ def setup_submitter(application):
         raise Exception("Unknown EQ_SUBMISSION_BACKEND")
 
 
+def setup_task_client(application):
+    if application.config["EQ_CLOUD_TASK_CLIENT"] == "cloud-tasks":
```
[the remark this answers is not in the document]
berroar: Renamed this

app/cloud_tasks/exceptions/__init__.py
berroar: Moved these to an `exceptions.py`

app/views/handlers/confirmation_email.py
berroar: Have removed this now

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -0,0 +1,55 @@
...
+            self._project_id, "europe-west2", queue_name
+        )
+
+        url = f"https://europe-west2-{self._project_id}.cloudfunctions.net/eq-submission-confirmation-consumer"
+        self._task = {
+            "http_request": {
+                "http_method": HttpMethod.POST,
+                "url": url,
+                "oidc_token": {
+                    "service_account_email": "cloud-task-http-function-invok@james-test-292314.iam.gserviceaccount.com"
```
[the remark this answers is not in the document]
berroar: Updated this with the service account that @LJBabbage provided

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, queue_name: str):
+        self._client = CloudTasksClient()
+
+        _, self._project_id = auth.default()
+
+        self._parent = self._client.queue_path(
+            self._project_id, "europe-west2", queue_name
+        )
+
+        url = f"https://europe-west2-{self._project_id}.cloudfunctions.net/eq-submission-confirmation-consumer"
```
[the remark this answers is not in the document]
berroar: Made this configurable now

app/settings.py
```diff
@@ -59,8 +59,11 @@ def utcoffset_or_fail(date_value, key):
 
 EQ_PUBLISHER_BACKEND = os.getenv("EQ_PUBLISHER_BACKEND")
 EQ_FULFILMENT_TOPIC_ID = os.getenv("EQ_FULFILMENT_TOPIC_ID", "eq-fulfilment-topic")
-EQ_SUBMISSION_CONFIRMATION_TOPIC_ID = os.getenv(
-    "EQ_SUBMISSION_CONFIRMATION_TOPIC_ID", "eq-submission-confirmation-topic"
+EQ_SUBMISSION_CONFIRMATION_QUEUE = os.getenv(
+    "EQ_SUBMISSION_CONFIRMATION_QUEUE", "eq-submission-confirmation"
+)
+EQ_SUBMISSION_CONFIRMATION_BACKEND = os.getenv(
+    "EQ_SUBMISSION_CONFIRMATION_BACKEND", "cloud-tasks"
```
MebinAbraham, line 66: The default should be removed to enforce that it has to be set explicitly. The validation should check the value instead of defaulting to log publisher to be consistent with other backends.

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -0,0 +1,57 @@
...
+
+from app.settings import SUBMISSION_CONFIRMATION_CONSUMER
+
+from .exceptions import CloudTaskCreationFailed
+
+logger = get_logger(__name__)
+
+
+class CloudTaskPublisher:
+    def __init__(self, queue_name: str):
```
MebinAbraham, line 14: We should keep this similar to Pub/Sub where the queue name is provided when you call `create` as you could have multiple queues.

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -0,0 +1,57 @@
...
+                    "Content-type": "application/json",
+                },
+            },
+        }
+
+    def _create(self, payload: bytes) -> Task:
+        logger.info("creating cloud task")
+        task = self._task.copy()
+        task["http_request"]["body"] = payload
+        return self._client.create_task(request={"parent": self._parent, "task": task})
```
MebinAbraham, line 41: `create_task` does support a mapping in its logic but the type hint is expecting a `CreateTaskRequest`, so the IDE is not a fan of the mismatch. It can be fixed, if we updated this to:
```py
request = CreateTaskRequest(mapping={"parent": self._parent, "task": task})
return self._client.create_task(request=request)
```
This is what the lib does if we don't provide a `CreateTaskRequest` ourselves.

app/setup.py
```diff
@@ -342,6 +345,15 @@ def setup_submitter(application):
         raise Exception("Unknown EQ_SUBMISSION_BACKEND")
 
 
+def setup_task_client(application):
+    if application.config["EQ_SUBMISSION_CONFIRMATION_BACKEND"] == "cloud-tasks":
+        application.eq["task-client"] = CloudTaskPublisher(
+            application.config["EQ_SUBMISSION_CONFIRMATION_QUEUE"]
+        )
+    else:
+        application.eq["task-client"] = LogCloudTaskPublisher()
```
ajmaddaford: I think we need to consider the naming. `EQ_SUBMISSION_CONFIRMATION_BACKEND` implies it's specific to submission confirmation, but the assignment is to a generic `task-client`.

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -0,0 +1,57 @@
...
+            self._project_id, "europe-west2", queue_name
+        )
+
+        url = f"https://europe-west2-{self._project_id}.cloudfunctions.net/{SUBMISSION_CONFIRMATION_CONSUMER}"
+        self._task = {
+            "http_request": {
+                "http_method": HttpMethod.POST,
+                "url": url,
+                "oidc_token": {
+                    "service_account_email": f"cloud-function-invoker@{self._project_id}.iam.gserviceaccount.com"
```
MebinAbraham, line 29: FYI, the email has been updated in Lewis' PR. It is now `cloud-functions@`

app/setup.py
MebinAbraham: Makes sense. Only feedback backend then stands out against the rest. But since you can choose your "queue" at the time of the action, it makes sense to be generic.

app/setup.py
```diff
@@ -342,6 +345,19 @@ def setup_submitter(application):
...
 
 
+def setup_task_client(application):
+    if application.config["EQ_SUBMISSION_CONFIRMATION_BACKEND"] == "cloud-tasks":
+        queue = application.config["EQ_SUBMISSION_CONFIRMATION_QUEUE"]
+
+        if not queue:
+            raise Exception("Setting EQ_SUBMISSION_CONFIRMATION_QUEUE Missing")
+
+        application.eq["task-client"] = CloudTaskPublisher(queue)
```
MebinAbraham, line 355: The queue should not be passed here. The Task publisher can be generic like Pub/Sub then the queue is passed at the time of task creation. Basically the same principle as Pub/Sub.

app/views/handlers/confirmation_email.py
```diff
@@ -63,16 +65,16 @@ def get_page_title(self):
...
-            return current_app.eq["publisher"].publish(
-                topic_id, message=fulfilment_request.message
+            return current_app.eq["task-client"].create_task(
+                payload=fulfilment_request.payload,
+                queue_name=EQ_SUBMISSION_CONFIRMATION_QUEUE,
             )
-        except PublicationFailed:
-            raise ConfirmationEmailFulfilmentRequestPublicationFailed
+        except PublicationFailed as exc:
+            raise ConfirmationEmailTaskPublicationFailed from exc
```
ajmaddaford, line 77: `create_task` doesn't raise a `PublicationFailed` exception, I think this should be catching `CloudTaskCreationFailed` exceptions.

There is also some inconsistencies in naming between `ConfirmationEmailTaskPublicationFailed` and `CloudTaskCreationFailed` (publication vs creation).

app/views/handlers/confirmation_email.py
```diff
@@ -92,21 +94,19 @@ def is_enabled(schema: QuestionnaireSchema):
...
     session_data: SessionData
     schema: QuestionnaireSchema
 
-    def _payload(self) -> Mapping:
-        return {
-            "fulfilmentRequest": {
+    @property
+    def payload(self):
+        return json.dumps(
+            {
```
ajmaddaford, line 105: We made a decision a while ago to use the `fulfilmentRequest` wrapper to be consistent with how we queue external requests from runner. I can't see why we are removing that. In doing this, we have also dropped the extra metadata (transaction id).

app/views/handlers/confirmation_email.py
```diff
@@ -92,21 +94,19 @@ def is_enabled(schema: QuestionnaireSchema):
...
                 "email_address": self.email_address,
+                "personalisation": {"address": self.session_data.display_address},
```
ajmaddaford, line 107: Given that `personalisation` is specific to the choice of email provider (gov notify), I think we should revert to the flat structure we had previously and do the work to construct the gov notify payload in the cloud task.

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -0,0 +1,57 @@
...
+
+        logger.info("creating cloud task")
+        task = self._task.copy()
+        task["http_request"]["body"] = payload
+        return self._client.create_task(request={"parent": self._parent, "task": task})
+
+    def create_task(self, payload: bytes, queue_name: str) -> None:
+        try:
+            self._create(payload, queue_name)
+            logger.info("task published successfully")  # pragma: no cover
```
ajmaddaford, line 45: The positive log message should align with the exception message:
```suggestion
            logger.info("task created successfully")  # pragma: no cover
```

app/views/handlers/confirmation_email.py
JamesGardiner: Consumer has been updated to expect `display_address` in the payload https://github.com/ONSdigital/eq-submission-confirmation-consumer/pull/2/commits/6ce9def679f5d7b473a8a9c71ca7e6ee9039d0d9

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -0,0 +1,58 @@
...
+
+        self._parent = self._client.queue_path(
+            self._project_id, "europe-west2", queue_name
+        )
+
+        service_account_email = (
+            f"cloud-functions@{self._project_id}.iam.gserviceaccount.com"
+        )
+
+        url = f"https://europe-west2-{self._project_id}.cloudfunctions.net/{SUBMISSION_CONFIRMATION_CLOUD_FUNCTION_NAME}"
```
MebinAbraham, line 30: We need to validate `SUBMISSION_CONFIRMATION_CLOUD_FUNCTION_NAME` exists. Given we tie having cloud functions to a task client, this should be checked when we initialise the `CloudTaskPublisher`, alternatively given you can have multiple queues, you could have multiple functions for each queue, which means this should be passed in like the queue nam.

@berroar @ajmaddaford thoughts?

app/setup.py
```diff
@@ -342,6 +345,24 @@ def setup_submitter(application):
         raise Exception("Unknown EQ_SUBMISSION_BACKEND")
 
 
+def setup_task_client(application):
+    if application.config["EQ_SUBMISSION_CONFIRMATION_BACKEND"] == "cloud-tasks":
+        application.eq["task-client"] = CloudTaskPublisher()
+
+        if not application.config["EQ_SUBMISSION_CONFIRMATION_QUEUE"]:
```
MebinAbraham, line 352: The queue name is defaulted so this check isn't needed.

app/setup.py
```diff
@@ -342,6 +345,24 @@ def setup_submitter(application):
...
+            raise Exception("Setting EQ_SUBMISSION_CONFIRMATION_QUEUE Missing")
+
+        if not application.config["SUBMISSION_CONFIRMATION_CLOUD_FUNCTION_NAME"]:
```
MebinAbraham, line 355: Given `EQ_SUBMISSION_CONFIRMATION_QUEUE` is defaulted, I think `SUBMISSION_CONFIRMATION_CLOUD_FUNCTION_NAME` should also be defaulted and removed fromt this check.

app/setup.py
berroar: As discussed I've renamed this but have also moved the task creation to the `_publish_fulfilment_request` method

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -54,5 +51,5 @@ def create_task(self, payload: bytes, queue_name: str) -> None:
 
 class LogCloudTaskPublisher:
     @staticmethod
-    def create_task(payload: bytes, queue_name: str) -> None:
-        logger.info("creating cloud task", payload=payload, queue_name=queue_name)
+    def create_task(task: dict, queue_name: str) -> None:
```
JamesGardiner, line 54: don't use `dict` for type hints where possible - the typing library provides `Mapping` or `MutableMapping`

app/settings.py
```diff
@@ -59,9 +59,10 @@ def utcoffset_or_fail(date_value, key):
 
 EQ_PUBLISHER_BACKEND = os.getenv("EQ_PUBLISHER_BACKEND")
 EQ_FULFILMENT_TOPIC_ID = os.getenv("EQ_FULFILMENT_TOPIC_ID", "eq-fulfilment-topic")
-EQ_SUBMISSION_CONFIRMATION_TOPIC_ID = os.getenv(
-    "EQ_SUBMISSION_CONFIRMATION_TOPIC_ID", "eq-submission-confirmation-topic"
+EQ_SUBMISSION_CONFIRMATION_QUEUE = os.getenv(
+    "EQ_SUBMISSION_CONFIRMATION_QUEUE", "eq-submission-confirmation"
 )
+EQ_SUBMISSION_CONFIRMATION_BACKEND = os.getenv("EQ_SUBMISSION_CONFIRMATION_BACKEND")
```
MebinAbraham, line 65: This needs to be added to the Helm deployment yaml as it is currently failing due to not backend.

app/views/handlers/confirmation_email.py
```diff
@@ -63,16 +65,29 @@ def get_page_title(self):
...
+            "http_request": {
+                "http_method": HttpMethod.POST,
+                "url": "",
+                "oidc_token": {"service_account_email": ""},
+                "headers": {
+                    "Content-type": "application/json",
+                },
+                "body": fulfilment_request.message,
+            },
+        }
```
MebinAbraham, line 82: How come this was extracted out?. The struct is very specific to task creation. I think the url and email being done in create_task and the rest here isn't very clear.
I think we should extract the whole things out or none of it.|

Thoughts?

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+
+        url = f"https://europe-west2-{self._project_id}.cloudfunctions.net/{SUBMISSION_CONFIRMATION_CLOUD_FUNCTION_NAME}"
+
+        keys = {
+            "url": url,
+            "oidc_token": {"service_account_email": service_account_email},
+        }
+
+        task.update(keys)
```
MebinAbraham, line 37: Currently, these are at the wrong level and fails with `Protocol message Task has no "url" field.` Which also indicate we are missing a test for the task payload structure.
```py
{
'http_request': {}
'url': '', 
'oidc_token': {}
}
```
`url` and `odic_token` should be nested under `http_request`.

Worth review my other comment of where the task definition should live first.

app/views/handlers/confirmation_email.py
berroar: As discussed I've moved the additional bits out, using `current_app.eq["cloud_tasks"].project_id` to get the project id for now but can change this pending further discussions

app/settings.py
berroar: Added this to the helm chart

app/views/handlers/confirmation_email.py
berroar: As discussed, was moved back to the cloud task publisher file

app/settings.py
```diff
@@ -132,3 +133,7 @@ def utcoffset_or_fail(date_value, key):
 ADDRESS_LOOKUP_API_AUTH_TOKEN_LEEWAY_IN_SECONDS = int(
     os.getenv("ADDRESS_LOOKUP_API_AUTH_TOKEN_LEEWAY_IN_SECONDS", "300")
 )
+
+SUBMISSION_CONFIRMATION_CLOUD_FUNCTION_NAME = os.getenv(
```
ajmaddaford, line 137: I think this should be prefixed with `EQ_` to align with `EQ_SUBMISSION_CONFIRMATION_QUEUE`. I also think they should appear next to each other in this file.

app/views/handlers/confirmation_email.py
```diff
@@ -92,7 +98,7 @@ def is_enabled(schema: QuestionnaireSchema):
 
 
 @dataclass
-class ConfirmationEmailFulfilmentRequest(FulfilmentRequest):
+class ConfirmationEmailTask(FulfilmentRequest):
```
ajmaddaford, line 101: The name of this should be reverted to `ConfirmationEmailFulfilmentRequest` as this is the fulfilment request that goes into the task.

k8s/helm/values.yaml
```diff
@@ -63,3 +63,5 @@ addressLookupApi:
 
 confirmationEmailLimit: 10
 feedbackLimit: 10
+
+eqSubmissionConfirmationBackend: "cloud_tasks"
```
ajmaddaford, line 67: I don't think we prefix properties with `eq` (even though it's in the environment variable name).

ajmaddaford (changes requested):

app/setup.py
```diff
@@ -342,6 +345,15 @@ def setup_submitter(application):
         raise Exception("Unknown EQ_SUBMISSION_BACKEND")
 
 
+def setup_task_client(application):
+    if application.config["EQ_SUBMISSION_CONFIRMATION_BACKEND"] == "cloud_tasks":
```
MebinAbraham, line 349: Can we make these `cloud-tasks` instead of the underscore?

app/views/handlers/confirmation_email.py
```diff
@@ -21,7 +25,7 @@ class ConfirmationEmailLimitReached(Exception):
     pass
 
 
-class ConfirmationEmailFulfilmentRequestPublicationFailed(Exception):
+class ConfirmationEmailTaskCreationFailed(Exception):
```
MebinAbraham, line 28: The class name should be changed back since we decided for it to be a fulfilment request.

MebinAbraham:
Looks good. Functionally working as expected.

MebinAbraham (approved):
👍

ajmaddaford (approved):
👍

PR #509, opened 2021-01-27
Use non-root pod and container users

JamesGardiner:
Updates three separate security concerns:
  - Uses a non-root user to run processes inside the Docker container
  - Add a pod security context that specifies runAsUser and fsGroup
  - Add a container security context that specifies runAsUser and allowPrivilegeEscalation

### How to review 
Build the runner image, and run the container. Shell into the container and test that the user id and username are as follows:

```shell
$ whoami
> appuser
$ echo $UID
> 9000
```

Deploy to GKE using the updated deployment spec. Check that the YAML definition for both the runner pods and runner containers contains the additional securityContext definitions


petechd (approved):
[Screenshot 2021-01-27 at 11 17 12
](url)👍

ajmaddaford (approved):
Tested local docker build and GCP deployment ✅

PR #513, opened 2021-02-01
Fix error messages on dates

petechd:
This fixes error messages when dates entered are negative, it now displays correct invalid date text for day and month incorrect numbers. This problem is described on [following Trello card](https://trello.com/c/ebDRRBVT).

### How to review 
Try `test_dates`, two new unit tests were added for negative number of days and months.

### Checklist

### DIFF
--- a/app/forms/validators.py
+++ b/app/forms/validators.py
@@ -222,25 +222,15 @@ class DateCheck:
         self.message = message or error_messages["INVALID_DATE"]
 
     def __call__(self, form, field):
-
-        if not form.data or not re.match(r"\d{4}$", str(form.year.data)):
-            raise validators.StopValidation(self.message)
-        if re.match(r"^\d{4}-", str(form.data)) and not re.match(
-            r"\d{1,2}", str(form.month.data)
-        ):
-            raise validators.StopValidation(self.message)
-        if re.match(r"^\d{4}-\d{1,2}-", str(form.data)) and not re.match(
-            r"\d{1,2}", str(form.day.data)
-        ):
+        if not form.data or not hasattr(form, "year"):
             raise validators.StopValidation(self.message)
 
         try:
-            substrings = form.data.split("-")
-            if len(substrings) == 3:
+            if hasattr(form, "day"):
                 datetime.strptime(form.data, "%Y-%m-%d")
-            if len(substrings) == 2:
+            elif hasattr(form, "month"):
                 datetime.strptime(form.data, "%Y-%m")
-            if len(substrings) == 1:
+            else:
                 datetime.strptime(form.data, "%Y")
         except ValueError:
             raise validators.StopValidation(self.message)
--- a/app/forms/validators.py
+++ b/app/forms/validators.py
@@ -225,6 +225,14 @@ class DateCheck:
 
         if not form.data or not re.match(r"\d{4}$", str(form.year.data)):
             raise validators.StopValidation(self.message)
+        if re.match(r"^\d{4}-", str(form.data)) and not re.match(
+            r"\d{1,2}", str(form.month.data)
+        ):
+            raise validators.StopValidation(self.message)
+        if re.match(r"^\d{4}-\d{1,2}-", str(form.data)) and not re.match(
+            r"\d{1,2}", str(form.day.data)
+        ):
+            raise validators.StopValidation(self.message)
 
         try:
             substrings = form.data.split("-")
--- a/app/forms/validators.py
+++ b/app/forms/validators.py
@@ -222,7 +222,7 @@ class DateCheck:
         self.message = message or error_messages["INVALID_DATE"]
 
     def __call__(self, form, field):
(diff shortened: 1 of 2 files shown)


petechd:
[Review error messages on dates for negative number entry](https://trello.com/c/ebDRRBVT/4502-review-error-messages-on-dates-for-negative-number-entry)

JamesGardiner (approved):

app/forms/validators.py
```diff
@@ -225,6 +225,14 @@ def __call__(self, form, field):
...
         if not form.data or not re.match(r"\d{4}$", str(form.year.data)):
             raise validators.StopValidation(self.message)
+        if re.match(r"^\d{4}-", str(form.data)) and not re.match(
+            r"\d{1,2}", str(form.month.data)
+        ):
+            raise validators.StopValidation(self.message)
+        if re.match(r"^\d{4}-\d{1,2}-", str(form.data)) and not re.match(
+            r"\d{1,2}", str(form.day.data)
+        ):
+            raise validators.StopValidation(self.message)
```
ajmaddaford, line 235: I don't think we need to check `form.data` each time, so this can be simplified to:
```python
        if (
            not form.data
            or not re.match(r"\d{4}$", str(form.year.data))
            or not re.match(r"\d{1,2}", str(form.month.data))
            or not re.match(r"\d{1,2}", str(form.day.data))
        ):
            raise validators.StopValidation(self.message)
```

tests/app/forms/validation/test_date_check_validator.py
```diff
@@ -24,9 +24,11 @@ def test_date_type_validator_empty_string(self):
     def test_date_type_validator_missing_day(self):
         validator = DateCheck()
 
-        mock_form = Mock()
+        mock_form = MagicMock()
```
ajmaddaford, line 27: Why have these changed from `Mock` to `MagicMock`?

petechd: From what I saw in debug we cannot iterate through elements of `Mock` which breaks our code. In `MagicMock` we don't have to declare extra methods we want to use.

app/forms/validators.py
petechd: Some of the forms we use don't have `day` and/or `month` attributes and it breaks runner this way (when looking for `form.month.data` and `form.day.data`).

tests/app/forms/validation/test_date_check_validator.py
ajmaddaford: Any ones in particular? I tried changing some instances of this back to `Mock` and they worked.

app/forms/validators.py
ajmaddaford: Ah ok. In that case we should explicitly check for those attributes, something like:
```python
        if (
            not form.data
            or not re.match(r"\d{4}$", str(form.year.data))
            or (hasattr(form, "month") and not re.match(r"\d{1,2}", str(form.month.data)))
            or (hasattr(form, "day") and not re.match(r"\d{1,2}", str(form.day.data)))
        ):
            raise validators.StopValidation(self.message)
```

MebinAbraham: Correct me if I am wrong but could we not replace this whole method with:
```py
        if not form.data or not hasattr(form, "year"):
            raise validators.StopValidation(self.message)

        try:
            if hasattr(form, "day"):
                datetime.strptime(form.data, "%Y-%m-%d")
            elif hasattr(form, "month"):
                datetime.strptime(form.data, "%Y-%m")
            else:
                datetime.strptime(form.data, "%Y")
        except ValueError:
            raise validators.StopValidation(self.message)
```

tests/app/forms/validation/test_date_check_validator.py
petechd: `MagicMock` is removed now, after applying Mebin's fix it's not needed anymore. Thanks👍

app/forms/validators.py
petechd: Refactored, so it's using this solution above now, thanks👍.

app/forms/validators.py
```diff
@@ -222,17 +222,15 @@ def __init__(self, message=None):
         self.message = message or error_messages["INVALID_DATE"]
 
     def __call__(self, form, field):
-
-        if not form.data or not re.match(r"\d{4}$", str(form.year.data)):
+        if not form.data or not hasattr(form, "year"):
```
ajmaddaford, line 225: The `year` attribute doesn't need to be checked here, as it will always be present in a date form. 
```suggestion
        if not form.data:
```

app/forms/validators.py
ajmaddaford: Nice 👌

app/forms/validators.py
petechd: Removed, thanks!

ajmaddaford (approved):

MebinAbraham (approved):

PR #527, opened 2021-02-04
Override skip to content to allow translation

JamesGardiner:
Overrides the onsSkiptToContent macro to allow translation of the link text.

### How to review 
Check the macro still renders correctly and the translation templates are updatd correctly

### Checklist


LJBabbage (approved):
It works after updating the po file

MebinAbraham (approved):

PR #550, opened 2021-03-02
Add pause resume cloud task queue yaml

LJBabbage:
This PR introduces yaml tasks to allow the pausing/resuming of a cloud task queue in GCP

### How to review 
In a test GCP environment with a cloud task queue (eq-submission-confirmation) check the scripts pause/resume the queue

### Checklist


ajmaddaford:
Not sure why, but I thought the task files would live in the eq-submission-confirmation-consumer repo.

ajmaddaford (approved):

MebinAbraham:
I don't think it should live in runner as runner doesn't create it and makes no difference to runner whether it is paused or not. The reason to pause it is to stop the consumption of tasks, therefore I think the consumer repo is better suited.

LJBabbage:
> I don't think it should live in runner as runner doesn't create it and makes no difference to runner whether it is paused or not. The reason to pause it is to stop the consumption of tasks, therefore I think the consumer repo is better suited.

Its generic queue pause/resume script, the repo being name eq-submission-confirmation-consumer doesn't really lend itself to a 2nd queue being added. If the script was move there I am not sure why we would need the queue name to be passed in

ci/README.md
```diff
@@ -75,3 +75,26 @@ There are defaults for both DATAFLOW_TEMPLATE_VERSION and EXPIRATION_TIME_OFFSET
 
 - `DATAFLOW_TEMPLATE_VERSION` is the template version on GCP you want to use
 - `EXPIRATION_TIME_OFFSET_IN_SECONDS` is the offset in seconds expired sessions will be purged from (i.e 3600 would delete all sessions more than an hour old)
+
+
+## Pause cloud task queue
```
MebinAbraham, line 80:
```suggestion
## Pause Cloud Task queue
```

ci/README.md
```diff
@@ -75,3 +75,26 @@ There are defaults for both DATAFLOW_TEMPLATE_VERSION and EXPIRATION_TIME_OFFSET
...
+A cloud task queue can be paused by using the `pause_cloud_task_queue.yaml` task. This can be done via Concourse using the following command:
+
+```sh
+PROJECT_ID=<project_id> \
+QUEUE_NAME=<queue_name> \
+fly -t <target-concourse> execute \
+  --config ci/pause_cloud_task_queue.yaml
+```
+
+## Resume cloud task queue
```
MebinAbraham, line 91:
```suggestion
## Resume Cloud Task queue
```

MebinAbraham (approved):
> > I don't think it should live in runner as runner doesn't create it and makes no difference to runner whether it is paused or not. The reason to pause it is to stop the consumption of tasks, therefore I think the consumer repo is better suited.
> 
> Its generic queue pause/resume script, the repo being name eq-submission-confirmation-consumer doesn't really lend itself to a 2nd queue being added. If the script was moved there I am not sure why we would need the queue name to be passed in

The queue name is already hardcoded in all 3 repos. So technically it isn't needed anyway. It was simply to accommodate it in case the name ever changed. But won't block since it is generic enough.

PR #553, opened 2021-03-03
Fix mobile number validation

petechd:
This fixes a problem with mobile number validation described on [this incident card](https://trello.com/c/Zi0k1Oo5). Regex has been refactored so it checks for correct digit at the start (`^7[0-9]+$`).

### How to review 
Run `make test-unit`, make sure newly added unit test checks for numbers not starting with number `7`.

### Checklist


petechd:
[INC0012406 Mobile number validation](https://trello.com/c/Zi0k1Oo5/4606-inc0012406-mobile-number-validation)

tests/app/forms/validation/test_mobile_number_validation.py
```diff
@@ -59,6 +59,18 @@ def test_non_numeric_string_invalid(self):
 
         self.assertEqual(error_messages["INVALID_MOBILE_NUMBER"], str(ite.exception))
 
+    def test_number_starts_with_non_7(self):
```
JamesGardiner, line 62:
```suggestion
    def test_number_not_starting_with_7_raises_validation_error(self):
```

JamesGardiner (changes requested):
Can we add a test for a number that has no leading values before the 7 - e.g. 7654321012?

tests/app/forms/validation/test_mobile_number_validation.py
petechd: Thanks @JamesGardiner !

petechd:
> Can we add a test for a number that has no leading values before the 7 - e.g. 7654321012?

Added number with no leading value to `test_sanitise_mobile_number`. Thanks!

tests/app/forms/validation/test_mobile_number_validation.py
```diff
@@ -59,6 +60,18 @@ def test_non_numeric_string_invalid(self):
 
         self.assertEqual(error_messages["INVALID_MOBILE_NUMBER"], str(ite.exception))
 
+    def test_number_not_starting_with_7_raises_validation_error(self):
```
LJBabbage, line 63: Can we align the name with the others i.e _invalid rather than raises_validation_error, I also think it needs to highlight that its after the 0 gets stripped it needs to be 7

LJBabbage (changes requested):

tests/app/forms/validation/test_mobile_number_validation.py
LJBabbage: Just checked its called the prefix service type

LJBabbage: so something like test_prefix_service_type_invalid, you could always write a comment that says it should be 07

JamesGardiner (approved):

tests/app/forms/validation/test_mobile_number_validation.py
petechd: ok, renamed to test_prefix_service_type_invalid, thanks!

LJBabbage (approved):

PR #556, opened 2021-03-08
Handle confirmation param errors for mobile numbers and email when changed or removed

LJBabbage:
This PR changes a 500 errors into a 400 (BadRequest) when a respondent has removed the param (email/mobile number) or has changed the hash value

### How to review 
Using test_confirmation_email, complete the form and send an email, now on the confirmation page try changing the email hash in the url and enter, make sure it returns a 400 rather than a 500. Now remove the email param and value completely and make sure it returns a 400 as well

Using test_individual_response add 2 people. From the hub request an individual response by text message. Now on the confirm-number page try changing the mobile phone hash in the url and enter, make sure it returns a 400 rather than a 500. Now remove the mobile phone param and value completely and make sure it returns a 400 as well. Repeat for the confirmation page (the page after confirm-number)

### Checklist

### DIFF
--- a/app/routes/individual_response.py
+++ b/app/routes/individual_response.py
@@ -1,7 +1,9 @@
 from flask import Blueprint, g, redirect, request, url_for
 from flask_babel import lazy_gettext
 from flask_login import current_user, login_required
+from itsdangerous import BadSignature
 from structlog import get_logger
+from werkzeug.exceptions import BadRequest
 
 from app.authentication.no_questionnaire_state_exception import (
     NoQuestionnaireStateException,
@@ -248,7 +250,12 @@ def individual_response_text_message_confirmation(schema, questionnaire_store):
     if request.method == "POST":
         return redirect(url_for("questionnaire.get_questionnaire"))
 
-    mobile_number = url_safe_serializer().loads(request.args.get("mobile_number"))
+    if not (request_mobile_number := request.args.get("mobile_number")):
+        raise BadRequest
+    try:
+        mobile_number = url_safe_serializer().loads(request_mobile_number)
+    except BadSignature:
+        raise BadRequest
 
     return render_template(
         template="individual_response/confirmation-text-message",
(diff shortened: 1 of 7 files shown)


app/routes/individual_response.py
```diff
@@ -248,7 +250,12 @@ def individual_response_text_message_confirmation(schema, questionnaire_store):
     if request.method == "POST":
         return redirect(url_for("questionnaire.get_questionnaire"))
 
-    mobile_number = url_safe_serializer().loads(request.args.get("mobile_number"))
+    if not (request_mobile_number := request.args.get("mobile_number")):
+        raise BadRequest
+    try:
+        mobile_number = url_safe_serializer().loads(request_mobile_number)
+    except BadSignature:
```
LJBabbage, line 257: This could be BadData if we wanted to cover the base exception for itsdangerous, although I am not convinced that we need any of the others, and if we did they should be 500

ajmaddaford: I agree, I think if it's `BadData` it's more likely ti be something wrong on the server side than the client.

app/routes/individual_response.py
```diff
@@ -248,7 +250,12 @@ def individual_response_text_message_confirmation(schema, questionnaire_store):
...
+        raise BadRequest
```
ajmaddaford, line 258: This works without checking for the existence of the `mobile_number` request argument:
```suggestion
    try:
        mobile_number = url_safe_serializer().loads(request.args["mobile_number"])
    except BadSignature:
        raise BadRequest
```
A `werkzeug.exceptions.BadRequestKeyError` is raised and caught by the 400 error handler. For clarity we could add `BadRequestKeyError` to the 400 error handler.

This also has the advantage that the error trace logging makes a clear distinction between the key error and the bad signature.

ajmaddaford (approved):

JamesGardiner (approved):

PR #557, opened 2021-03-09
Add transactionId to Individual Response and Confirmation logging

JamesGardiner:
Adds the transactionId to log lines assosciated with the individual response and confirmation fulfilment requests

### How to review 
Check that the assosciated log lines are correctly emitted when requesting an IR and a confirmation email

### DIFF
--- a/app/cloud_tasks/cloud_task_publishers.py
+++ b/app/cloud_tasks/cloud_task_publishers.py
@@ -46,24 +46,27 @@ class CloudTaskPublisher:
         )
         return task
 
-    def create_task(self, body: bytes, queue_name: str, function_name: str, transactionId: str) -> Task:
+    def create_task(
+        self, body: bytes, queue_name: str, function_name: str, transactionId: str
+    ) -> Task:
 
         parent = self._client.queue_path(self._project_id, "europe-west2", queue_name)
         try:
             task = self._create_task_with_retry(body, function_name, parent)
-            logger.info("task created successfully", transactionId=transactionId)  # pragma: no cover
+            logger.info(
+                "task created successfully", transactionId=transactionId
+            )  # pragma: no cover
             return task
         except Exception as exc:
-            logger.exception(
-                "task creation failed",
-                transactionId=transactionId
-            )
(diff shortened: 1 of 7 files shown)


app/data_models/fulfilment_request.py
```diff
@@ -20,7 +25,7 @@ def message(self) -> bytes:
                 "source": "QUESTIONNAIRE_RUNNER",
                 "channel": "EQ",
                 "dateTime": datetime.now(tz=tzutc()).isoformat(),
-                "transactionId": str(uuid4()),
+                "transactionId": self.transactionId,
```
ajmaddaford, line 28: Other than this dict property, every other occurrence of `transactionId` should be `transaction_id`.

app/cloud_tasks/cloud_task_publishers.py
```diff
@@ -46,27 +46,29 @@ def _create_task_with_retry(
...
         try:
             task = self._create_task_with_retry(body, function_name, parent)
-            logger.info("task created successfully")  # pragma: no cover
+            logger.info("task created successfully", transactionId=transactionId)
             return task
         except Exception as exc:
-            logger.exception(
-                "task creation failed",
-            )
+            logger.exception("task creation failed", transactionId=transactionId)
```
ajmaddaford, line 59: Where the transaction id is referenced outside of a fulfilment request, I think we should consider qualifying it e.g. `fulfilment_request_transaction_id` to make a clear distinction between this and the `tx_id`.

MebinAbraham (approved):
Nothing else to add other than the ones already raised.

app/cloud_tasks/cloud_task_publishers.py
JamesGardiner: Makes sense 👍🏻

JamesGardiner: Done

app/data_models/fulfilment_request.py
JamesGardiner: Done

ajmaddaford (approved):

PR #559, opened 2021-03-11
Add additional log lines for email validation failures

berroar:

Adds logging to the thank you and send confirmation email pages to enable us to track email confirmation validation errors. 

The log messages have an event of `email validation error` and an additional property `error_message` set to the error message text. 

### How to review 

Use the `test_confirmation_email` schema and check the log messages are generated on validation failure.

### Checklist


berroar:
[Add logging to differentiate between empty and invalid confirmation emails](https://trello.com/c/qufdRJRu/4627-add-logging-to-differentiate-between-empty-and-invalid-confirmation-emails)

petechd:
Do we want to keep it as "info" and not "warning" or "error"?

berroar:
> Do we want to keep it as "info" and not "warning" or "error"?

Happy to have opinions on this...

app/routes/questionnaire.py
```diff
@@ -331,6 +335,10 @@ def send_confirmation_email(session_store, schema):
                 email=confirmation_email.get_url_safe_serialized_email(),
             )
         )
+    else:
+        logger.info(
+            "email validation error", error_message=confirmation_email.form.errors
+        )
```
MebinAbraham, line 341: This would be logged for all non POST requests. It should be:
```py
if method == "POST":
	if form_valid:
		return redirect()
	
	log_error()

handle_get()

File: app/routes/questionnaire.py [hunk 2]
```diff
@@ -295,6 +295,10 @@ def get_thank_you(schema, session_store):
                     email=confirmation_email.get_url_safe_serialized_email(),
                 )
             )
+        else:
```
ajmaddaford, line 298: The `else` is not required - if the previous `if` statement matches then the method returns.

app/routes/questionnaire.py
```diff
@@ -323,14 +327,19 @@ def send_confirmation_email(session_store, schema):
...
+    if request.method == "POST":
+        if confirmation_email.form.validate():
+            confirmation_email.handle_post()
+            return redirect(
+                url_for(
+                    ".get_confirmation_email_sent",
+                    email=confirmation_email.get_url_safe_serialized_email(),
+                )
+            )
+        else:
```
ajmaddaford, line 339: Same as above, the `else` is not required here.

app/routes/questionnaire.py
```diff
@@ -295,6 +295,10 @@ def get_thank_you(schema, session_store):
                     email=confirmation_email.get_url_safe_serialized_email(),
                 )
             )
+        else:
+            logger.info(
+                "email validation error", error_message=confirmation_email.form.errors
```
ajmaddaford, line 300: The message needs to be extracted from `errors`, otherwise the contents of `error_message` will be something like:
```
{'email': [l'Enter an email address in a valid format, for example name@example.com']}
```

We should be able to use something like `str(confirmation_email.form.errors["email"][0])`. The `str` converts the lazy string into a plain string.

ajmaddaford (approved):

MebinAbraham (approved):

PR #562, opened 2021-03-15
Add confirm email to confirmation email journey

LJBabbage:
This PR introduces a confirm email page in the confirmation email journey

### How to review 
Using test_confirmation_email fill complete the questionnaire and enter a email. Make sure it goes to the confirm email page and that each answer routes to the correct place, also test it errors as expected when no answer is provided or the email hash is doctored. Repeat the process for sending another email using the same schema

### Checklist


app/routes/questionnaire.py
```diff
@@ -340,6 +340,26 @@ def send_confirmation_email(session_store, schema):
     )
 
 
+@post_submission_blueprint.route("confirmation-email/confirm", methods=["GET", "POST"])
+@with_schema
+@with_session_store
+def confirm_confirmation_email(session_store, schema):
+    confirm_email = ConfirmEmail(
```
LJBabbage, line 347: I find the confirm/confirmation naming pattern confusing, but am restricted with the current design

app/views/handlers/confirmation_email.py
```diff
@@ -46,13 +49,20 @@ def __init__(
...
         self._page_title = page_title
+        self._serialised_email = serialised_email
 
     @property
     def page_title(self):
         return self._page_title or lazy_gettext("Confirmation email")
 
     @cached_property
     def form(self):
+        if self._serialised_email:
```
LJBabbage, line 60: We now need to pre-fill this page if no is answered on the confirm email page

app/routes/questionnaire.py
```diff
@@ -315,19 +316,18 @@ def get_thank_you(schema, session_store):
 @with_schema
 @with_session_store
 def send_confirmation_email(session_store, schema):
-    if not session_store.session_data.confirmation_email_count:
```
LJBabbage, line 318: This has been removed as the new design means you can go Thank you -> confirm (answer no) -> send_confirmation_email. This means there would be no count and therefore it now can't return a 404

app/views/handlers/confirm_email.py
```diff
@@ -0,0 +1,90 @@
...
+        try:
+            email = url_safe_serializer().loads(self._serialized_email)
+        except BadSignature:
+            raise BadRequest
+
+        self._schema = schema
+        self._session_store = session_store
+        self._form_data = form_data
+        self._email = email
+        self.page_title = "Confirm your email address"
```
MebinAbraham, line 37: Needs to be marked for translation.

app/views/handlers/confirm_email.py
```diff
@@ -0,0 +1,90 @@
...
+                    "id": "confirm-email",
+                    "mandatory": True,
+                    "options": [
+                        {
+                            "label": CONFIRM_EMAIL_YES_VALUE,
+                            "value": CONFIRM_EMAIL_YES_VALUE,
+                        },
+                        {
+                            "label": "No, I need to correct it",
+                            "value": "No, I need to correct it",
```
MebinAbraham, line 85: Needs to be marked for translation.

app/views/contexts/confirm_email_form_context.py
```diff
@@ -0,0 +1,13 @@
+from flask import url_for
```
JamesGardiner, line 1: I think this module would be better names `confirm_email_context.py` though I agree - confirm/confirmation is confusing here

app/views/handlers/confirm_email.py
```diff
@@ -0,0 +1,90 @@
...
+class ConfirmEmail:
+    def __init__(
+        self,
+        schema: QuestionnaireSchema,
+        session_store: SessionStore,
+        serialized_email,
+        form_data: Mapping,
+    ):
+        self._serialized_email = serialized_email
+
```
MebinAbraham, line 27: This class should have:
```py

        if self.is_limit_reached(session_store.session_data):
            raise ConfirmationEmailLimitReached
```

JamesGardiner (dismissed):
Only one minor change suggestion on top of what @MebinAbraham has pointed out

templates/confirm-email.html
```diff
@@ -0,0 +1,36 @@
...
+
+{% block main %}
+  {% set form = content.form %}
+  {% if form.mapped_errors %}
+    {% set error_title = _("There is a problem with this page") %}
+    {% include 'partials/error-panel.html' %}
+  {% endif %}
+
+  <h1 class="question__title u-fs-xxl" data-qa="confirm-email-title">{{ content.question.title }}</h1>
+  <legend class="fieldset__legend">{{ content.question.label}}</legend>
```
petechd, line 18: This needs to be `<p class="checkboxes_label">{{ content.question.label}}</p>` to get same styling as in prototype.

app/routes/questionnaire.py
```diff
@@ -343,6 +342,30 @@ def send_confirmation_email(session_store, schema):
...
+        confirm_email = ConfirmEmail(
+            schema, session_store, request.args["email"], form_data=request.form
+        )
+    except (ConfirmationEmailLimitReached, ConfirmationEmailNotEnabled):
+        return redirect(url_for(".get_thank_you"))
+
+    if request.method == "POST" and confirm_email.form.validate():
+        confirm_email.handle_post()
+        next_location_url = confirm_email.get_next_location_url()
+        return redirect(next_location_url)
```
MebinAbraham, line 359: This needs to be similar to IR and need to take into account the `YES` value as it is sending a fulfilment request on every post.

LJBabbage: Good spot, probably wouldn't go down that path as that naturally starts getting into rendered block. I don't see any issue with using
```
if self._form_data.get("confirm-email") == CONFIRM_EMAIL_YES_VALUE:
```

MebinAbraham: Ya checking form data is fine, IR example was just for info. The value check is done in get_next_location_url which could be removed if the redirect is done in `handle_post` like in IR, but I don't mind what route you go with for that. If it is done in `handle_post` it avoids the double evaluation or alternatively you could use `@cached_property`.

LJBabbage: You might have missed the line above which mentioned extracted it to a @cached_property

MebinAbraham: Ah ya I did. 👍

tests/functional/spec/confirmation_email.spec.js
```diff
@@ -33,9 +34,23 @@ describe("Email confirmation", () => {
...
+    });
+
+    it("When I submit the confirm email page without providing an answer, Then I get an error message", () => {
+      $(ConfirmEmailPage.submit()).click();
+      expect(browser.getUrl()).to.contain("confirmation-email/confirm");
+      expect($(ConfirmEmailPage.errorPanel()).isExisting()).to.be.true;
+      expect($(ConfirmEmailPage.errorPanel()).getText()).to.contain("Select an answer");
+    });
+
+    it("When I answering 'Yes' and submit the confirm email page, Then I go to email sent page", () => {
```
MebinAbraham, line 51:
```suggestion
    it("When I answer 'Yes' and submit the confirm email page, Then I go to email sent page", () => {
```

tests/functional/spec/confirmation_email.spec.js
```diff
@@ -56,13 +71,30 @@ describe("Email confirmation", () => {
...
+      $(ConfirmEmailPage.submit()).click();
       expect(browser.getUrl()).to.contain("confirmation-email/sent");
       expect($(ConfirmationEmailSentPage.confirmationText()).getText()).to.equal("A confirmation email has been sent to name@example.com");
     });
   });
+  describe("Given I launch the test email confirmation survey", () => {
+    before(() => {
+      browser.openQuestionnaire("test_confirmation_email.json");
+    });
+    it("When I enter an email and answering 'No' on the confirm email page, Then I go the confirmation send page with the email pre-filled", () => {
```
MebinAbraham, line 87:
```suggestion
    it("When I enter an email and answer 'No' on the confirm email page, Then I go the confirmation send page with the email pre-filled", () => {
```

tests/integration/routes/test_confirmation_email.py
```diff
@@ -123,41 +152,77 @@ def test_default_themed_schema_with_confirmation_email_not_set(self):
...
+        self.get("/submitted/confirmation-email/send/")
+        self.post({"email": "email@example.com"})
+        confirm_email_url = self.last_url
+        self.post({"confirm-email": "Yes, send the confirmation email"})
+
+        # When I try to access the confirm email page
+        self.get(confirm_email_url)
+
+        # Then I get routed to the thank you page
+        self.assertInUrl("/submitted/thank-you/")
```
MebinAbraham, line 212: This should be done for "/send" also. It could just be a subtest with the urls being the params. So when visiting `/send` or `/confirm`, they both should redirect to the thank you page.

MebinAbraham (approved):
Functionally seems fine and flow seems as expected so ✅. Would have liked more time but since it is a high priority lets get it moving, but just need to ensure some thorough business testing is done. Hopefully nothing was missed.

ajmaddaford (commented):
I've tried to be pragmatic and only highlight the things that I think need to change. There is some refactoring that if given more time should be done, but functionally this works as expected 👍

petechd (approved):

berroar (approved):
Tested this - flow is fine and can't see any missed scenarios 👍

app/views/handlers/confirm_email.py
```diff
@@ -0,0 +1,158 @@
...
+    ConfirmationEmailLimitReached,
+    ConfirmationEmailNotEnabled,
+)
+
+
+class ConfirmationEmailFulfilmentRequestPublicationFailed(Exception):
+    pass
+
+
+CONFIRM_EMAIL_YES_VALUE = lazy_gettext("Yes, send the confirmation email")
```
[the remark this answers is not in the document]
ajmaddaford: As discussed, leaving this as is because it is consistent with the feedback handler.

ajmaddaford (approved):

PR #564, opened 2021-03-15
Retry Redis ConnectionErrors once

JamesGardiner:
When a `redis.exceptions.ConnectionEror`is raised during a put, get or delete operation, the operation will be retried once, in case the issue was transient.

`retry_on_timeout`is also now set to `True`when instantiating the Redis client, which will retry a TimeoutConnection once.


JamesGardiner:
[Redis disconnection errors](https://trello.com/c/cnFTTtes/4614-redis-disconnection-errors)

LJBabbage (commented):
Any advice on how to test?

JamesGardiner:
> Any advice on how to test?

I don't think there will be an easy way to manually test this to be honest - if you really want to try it, you'd need to set up some kind of mock service to trigger a redis connection error

app/storage/redis.py
```diff
@@ -27,16 +31,26 @@ def put(self, model, overwrite=True):
...
 
-        record_created = self.client.set(
-            name=key_value, value=value, ex=expires_in, nx=not overwrite
-        )
+        try:
+            record_created = self.client.set(
+                name=key_value, value=value, ex=expires_in, nx=not overwrite
+            )
+        except RedisConnectionError:
+            logger.info("retrying redis set command")
```
ajmaddaford, line 39: Can we make the log event consistent across the three commands and make the command a property? That way we can easily search for all retries regardless of command, and still see the command that failed.

```suggestion
            logger.info("retrying redis command", command="set")
```

JamesGardiner: Yeah sure thing - I guess there's no value in seeing that a certain type of command failed?

app/storage/redis.py
```diff
@@ -36,7 +40,7 @@ def put(self, model, overwrite=True):
                 name=key_value, value=value, ex=expires_in, nx=not overwrite
             )
         except RedisConnectionError:
-            logger.info("retrying redis set command")
+            logger.info(self.command_failed_log_message, command="set")
```
ajmaddaford, line 43: Seeing as the log message has been factored out into a method, the method could do the logging. The log lines would then be something like `log_retry(command="set")`.

JamesGardiner: Done 👍🏻

ajmaddaford:
@LJBabbage I tested this locally by:
- stopping my Redis container
- adding breakpoints in PyCharm on the Redis commands
- launching a test survey

Now that logging has been added though you could look for those lines in the logs rather than adding breakpoints.

ajmaddaford (approved):
Tested locally by stopping the Redis container. Also tested that the second command works by adding breakpoints and starting Redis again between the two commands 👍

LJBabbage (approved):
Fires as expected when redis stopped

PR #573, opened 2021-03-22
Handle HTTP HEAD requests

ajmaddaford:
Some errors were observed when HTTP HEAD requests were sent to questionnaire endpoints. It was identified that this was because a HEAD request was following the expected path of a POST request and calling form validation. In most instances this doesn't result in an error, but in the case of an optional `Date` type answer it does. 

This pull request changes the route logic so that HEAD requests follow the same path as a GET request.

### How to review 
- Check that the changes fix the issue
- Are there any other scenarios we should consider?

### Checklist

### DIFF
--- a/tests/integration/routes/test_questionnaire.py
+++ b/tests/integration/routes/test_questionnaire.py
@@ -1,24 +1,31 @@
-import simplejson as json
-from mock import Mock
-
-from app.data_models import QuestionnaireStore
 from tests.integration.integration_test_case import IntegrationTestCase
 
 
 class TestQuestionnaire(IntegrationTestCase):
-    def setUp(self):
-        super().setUp()
-        self._application_context = self._application.app_context()
-        self._application_context.push()
-
-        storage = Mock()
-        data = {"METADATA": "test", "ANSWERS": [], "PROGRESS": []}
-        storage.get_user_data = Mock(
-            return_value=(json.dumps(data), QuestionnaireStore.LATEST_VERSION)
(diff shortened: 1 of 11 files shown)


tests/integration/routes/test_questionnaire.py
```diff
@@ -1,24 +1,31 @@
...
-    def setUp(self):
-        super().setUp()
-        self._application_context = self._application.app_context()
-        self._application_context.push()
-
-        storage = Mock()
-        data = {"METADATA": "test", "ANSWERS": [], "PROGRESS": []}
-        storage.get_user_data = Mock(
-            return_value=(json.dumps(data), QuestionnaireStore.LATEST_VERSION)
+    def test_head_request_on_optional_date(self):
```
MebinAbraham, line 5: Happy to keep as is, but this test would only ensure `block` route is valid. Ideally, we should make a HEAD request for each of our route definitions.

MebinAbraham (commented):
OPTIONS is also a valid: `Starting with Flask 0.6, OPTIONS is implicitly added and handled by the standard request handling.` So we should accept that or reject it at an app level.

tests/integration/routes/test_questionnaire.py
ajmaddaford: I've added tests for the additional questionnaire endpoints. This highlighted that there was a gap in the logic for the session endpoints, so I've fixed that and added tests.

MebinAbraham (approved):

JamesGardiner (approved):

PR #574, opened 2021-03-22
Clear radio fields in repeating sections

petechd:
This fixes voluntary radio clear action in repeating sections when JS is not enabled. Requirement is described on [this Trello](https://trello.com/c/NeAFJLd1).

### How to review 
New schema and new test was added to integration tests (`test_questionnaire_radio_voluntary.py`). Check new test schema with JS disabled (`test_repeating_section_with_radio_voluntary.json`).

### Checklist


app/questionnaire/questionnaire_store_updater.py
```diff
@@ -97,6 +97,10 @@ def remove_answers(self, answer_ids: List):
         for answer_id in answer_ids:
             self._answer_store.remove_answer(answer_id)
 
+    def remove_answers_with_list_item_id(self, answer_ids: List, list_item_id: str):
+        for answer_id in answer_ids:
+            self._answer_store.remove_answer(answer_id, list_item_id)
+
```
MebinAbraham, line 103: A new method is not needed, I think remove answer should just be updated to also take a list_item_id which default to None like:
```py
def remove_answers(self, answer_ids: List, list_item_id: str = None):
    for answer_id in answer_ids:
        self._answer_store.remove_answer(answer_id, list_item_id)
```

petechd: Removed extra method now, thanks!

app/views/handlers/question.py
```diff
@@ -211,5 +211,7 @@ def clear_radio_answers(self):
                 answer_ids_to_remove.append(answer["id"])
 
         if answer_ids_to_remove:
-            self.questionnaire_store_updater.remove_answers(answer_ids_to_remove)
+            self.questionnaire_store_updater.remove_answers(
+                answer_ids_to_remove, self.current_location.list_item_id
+            )
```
MebinAbraham, line 216: Just realized, questionnaire store updater already has current_location as an attribute, so we don't need to pass it in this case. We do `list_item_id = list_item_id or self._current_location.list_item_id` in other places to make it flexible.

petechd:
[Clear radio fields in repeating sections](https://trello.com/c/NeAFJLd1/4647-clear-radio-fields-in-repeating-sections)

JamesGardiner (commented):
This could do with an additional test in `test_answer_store.py`

petechd:
> This could do with an additional test in `test_answer_store.py`

Added new unit test for remove answer without list item id 👍

JamesGardiner (approved):

tests/integration/questionnaire/test_questionnaire_radio_voluntary.py
```diff
@@ -12,3 +12,24 @@ def test_radio_voluntary(self):
         self.post(action="clear_radios")
         self.post()
         self.assertNotInBody("Coffee")
+
+
+class TestQuestionnaireRepeatingSectionRadioVoluntary(IntegrationTestCase):
```
LJBabbage, line 17: I realise that it matches the pattern above, but out of context and without reading the test I can't tell what it is testing. It is just clearing radio options? With the class name this could be as simple as test_clear_radios. I bit of space between the actions and the asserts also makes it a bit easy to digest

LJBabbage (changes requested):
Works as expect, just one minor comment

tests/integration/questionnaire/test_questionnaire_radio_voluntary.py
petechd: Renamed and added some spacing, thanks!

LJBabbage (approved):

MebinAbraham (approved):

PR #577, opened 2021-03-23
Upgrade to design system version 32.1.3-census

berroar:

Upgrades to design system version `32.1.3-census`.

Includes a tweak to the way the GTM is loaded as this design system release includes a breaking change. [Design system release here](https://github.com/ONSdigital/design-system/releases/tag/32.1.0-census).

### How to review 

Check that nothing is broken by this release and that the Google Tag manager integration still functions as expected.

### Checklist


berroar:
[Upgrade Runner to DS 32.1.0-census](https://trello.com/c/ACcn5uLd/4636-upgrade-to-ds-3203-census)

app/helpers/template_helpers.py
```diff
@@ -212,8 +212,10 @@ def render_template(template: str, **kwargs: Mapping) -> str:
 
 
 def get_google_tag_manager_context() -> Mapping:
-    cookie = request.cookies.get("ons_cookie_policy")
-    if cookie and "'usage':true" in cookie:
+    if (
+        current_app.config["EQ_GOOGLE_TAG_MANAGER_ID"]
+        and current_app.config["EQ_GOOGLE_TAG_MANAGER_AUTH"]
+    ):
```
MebinAbraham, line 218: Could make use of the walrus operator:
```py
    if google_tag_manager_id := current_app.config["EQ_GOOGLE_TAG_MANAGER_ID"] and (
        google_tag_manager_auth := current_app.config["EQ_GOOGLE_TAG_MANAGER_AUTH"]
    ):
        return {
            "google_tag_manager_id": google_tag_manager_id,
            "google_tag_manager_auth": google_tag_manager_auth,
        }
    return {}
```

templates/layouts/_base.html
```diff
@@ -83,12 +83,19 @@
...
+        j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+        'https://www.googletagmanager.com/gtm.js?id='+i+dl+ '&gtm_auth={{ google_tag_manager_auth }}&gtm_cookies_win=x';var n=d.querySelector('[nonce]');
+        n&&j.setAttribute('nonce',n.nonce||n.getAttribute('nonce'));f.parentNode.insertBefore(j,f);
+        })(window,document,'script','dataLayer','{{ google_tag_manager_id }}');
+     }
+
+     var a = /^(.*)?\s*'usage':true\s*[^;]+(.*)?$/;
+        if (document.cookie.match(a)) {
+            loadGTM();
+        }
```
MebinAbraham, line 98: The indentation is slightly off, and I think the var should be meaningful.
```suggestion
      function loadGTM() {
        (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
        new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
        j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
        'https://www.googletagmanager.com/gtm.js?id='+i+dl+ '&gtm_auth={{ google_tag_manager_auth }}&gtm_cookies_win=x';var n=d.querySelector('[nonce]');
        n&&j.setAttribute('nonce',n.nonce||n.getAttribute('nonce'));f.parentNode.insertBefore(j,f);
        })(window,document,'script','dataLayer','{{ google_tag_manager_id }}');
      }

      var cookie_usage_regex = /^(.*)?\s*'usage':true\s*[^;]+(.*)?$/;
      if (document.cookie.match(cookie_usage_regex)) {
          loadGTM();
      }
```

MebinAbraham:
Verified that:
- gtm details is in source when enabled
- the function is called on clicking "Accept all cookies"
- the function is not called on initial page load

I did notice that if you click "Accept all cookies" then reload the page, our if statement check for the cookie is not true and the loadGTM is not called, but on a subsequent refresh the if statement passes and the `loadGTM` is called.

app/helpers/template_helpers.py
berroar: Looks ok to me, but when I update this the ide complains that `google_tag_manager_auth` is might now be referenced before it's assignment

MebinAbraham: If the py version of your IDE is not set to py 3.8 then it will complain as walrus only got introduced in 3.8

ajmaddaford (commented):
Works as expected, except for the issue Mebin found. Also, when evaluating the cookie we should be checking the specific `ons_cookie_policy` cookie rather than any cookie.

MebinAbraham:
So Andrew and I had a look at the issue and in fact there is a bug in the DS which means on the first request after clicking `Accept all cookies`, the GTM scripts are not loaded. Have made Matt aware of the bug who will fix the issue in the DS but until then, I think this should be on hold.

ajmaddaford:
Cookies now working as expected 👍

MebinAbraham (approved):

ajmaddaford (approved):

PR #578, opened 2021-03-25
Exit before_request early when request is `OPTIONS`

JamesGardiner:
Requests of type `OPTIONS` should not get schema data or metadata for the request, as an empty 200 response is returned by default.

This adds an early test to the `before_request` method for questionnaire and post-submission routes so that unnecessary calls are avoided.


JamesGardiner:
[Handle OPTIONS requests a GET request in runner](https://trello.com/c/9OEaqSm7/4654-handle-options-requests-a-get-request-in-runner)

tests/integration/routes/test_questionnaire.py
```diff
@@ -44,3 +44,12 @@ def test_head_request_on_block_with_optional_date_answer(self):
         )
         self.head("/questionnaire/date-non-mandatory-block/")
         self.assertStatusOK()
+
+    def test_options_request_exits_before_request_early(self):
+        self.launchSurvey("test_hub_and_spoke")
+        with self.assertLogs() as cm:
```
ajmaddaford, line 50: I think the naming here could be improved, `cm` is a bit vague - maybe `logs`?

ajmaddaford (commented):
The endpoints in the `session_blueprint` also need to be looked at.

tests/integration/routes/test_questionnaire.py
JamesGardiner: Done ✅

tests/integration/routes/test_questionnaire.py
```diff
@@ -44,3 +44,12 @@ def test_head_request_on_block_with_optional_date_answer(self):
         )
         self.head("/questionnaire/date-non-mandatory-block/")
         self.assertStatusOK()
+
+    def test_options_request_exits_before_request_early(self):
```
LJBabbage, line 48: Very minor but the 'early' confused me for a second, not sure it is needed

tests/integration/routes/test_thank_you.py
```diff
@@ -62,3 +62,15 @@ def test_head_request_on_thank_you(self):
         self.post()
         self.head("/submitted/thank-you")
         self.assertStatusOK()
+
+    def test_options_request_on_post_submission_exits_before_request_early(self):
```
LJBabbage, line 66: As above

LJBabbage (changes requested):
Does IR also need doing?

JamesGardiner:
> Does IR also need doing?

Yes 👍🏻

tests/integration/routes/test_questionnaire.py
JamesGardiner: Done

tests/integration/routes/test_thank_you.py
JamesGardiner: Done

LJBabbage (approved):
Looks good to me, although missing a 'then' in one of the tests

JamesGardiner:
> The endpoints in the `session_blueprint` also need to be looked at.

I've been looking at the flask code for options requests. If `OPTIONS` isn't set specifically in the route decorator, Flask doesn't enter that routes function, and instead returns a default response of code 200, with no content and sets the allowed header for the response correctly.

https://github.com/pallets/flask/blob/510c38cc63fd1d88772d6680713c38dd83e277b4/src/flask/app.py#L1435

Given this, it looks like we don't need the additional tests, as they would be testing Flask code.

ajmaddaford (approved):

PR #581, opened 2021-04-07
Update Python version to 3.9.4

JamesGardiner:
Updates Python to 3.9


Pipfile
```diff
@@ -14,7 +14,7 @@ pylint-quotes = "*"
 "beautifulsoup4" = "*"
 httmock = "*"
 "flake8" = "*"
-"flake8-blind-except" = "*"
+flake8-blind-except = "==v0.1.1"
```
ajmaddaford, line 17: From https://github.com/elijahandrews/flake8-blind-except/pull/12 and https://github.com/elijahandrews/flake8-blind-except/issues/11 - it looks like the functionality of this version is now included in flake8 as E722. We should consider dropping this dependency rather than fix on this version.

JamesGardiner: Done 👍🏻

petechd:
Can we upgrade ChromeDriver to 89 as a part of this PR?

petechd (commented):
Can we upgrade ChromeDriver to 89 as a part of this PR?

MebinAbraham (approved):

JamesGardiner:
> Can we upgrade ChromeDriver to 89 as a part of this PR?

happy to take views on this from other reviewers. @ajmaddaford @MebinAbraham ?

ajmaddaford (approved):

MebinAbraham:
> > Can we upgrade ChromeDriver to 89 as a part of this PR?
> 
> happy to take views on this from other reviewers. @ajmaddaford @MebinAbraham ?

I would say it should be separate as it is a JS dependency and not python. This is explicitly upgrading python and it's packages.

PR #582, opened 2021-04-07
Return None from get_session_store if no session data

ajmaddaford:
Changes the `get_session_store` method to only return a `SessionStore` object if a session exists. Prior to this, when a session didn't exist an 'empty' `SessionStore` object was returned with no `session_data`.

### How to review 
- Are the changes appropriate?
- Are there nay gaps?

### Checklist

### DIFF
--- a/app/authentication/authenticator.py
+++ b/app/authentication/authenticator.py
@@ -86,7 +86,7 @@ def load_user():
     """
     session_store = get_session_store()
 
-    if session_store and session_store.user_id and _is_session_valid(session_store):
+    if session_store and _is_session_valid(session_store):
         logger.debug("session exists")
 
         user_id = session_store.user_id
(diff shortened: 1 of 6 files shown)


MebinAbraham (commented):
The setup_babel's `get_locale` method no longer needs to check for session data.
https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/setup.py#L452-L459

```py
    @application.babel.localeselector
    def get_locale():  # pylint: disable=unused-variable
        session = get_session_store()
        return session.session_data.language_code if session else None
```

JamesGardiner (approved):
LGTM - aside from the get_locale change @MebinAbraham pointed out, I don't see any other usages of the session_data attr

MebinAbraham (approved):

PR #591, opened 2021-05-05
Add Python type hinting style guide

ajmaddaford:
As a team we have committed to adding type hints throughout the Python code. This pull request defines our approach to type hinting, to ensure consistency in the codebase and avoid repeating discussions about how we apply typing.

### How to review 
- Review the document
- Is it clear? 
- Has anything been missed?

### Checklist

### DIFF
--- a/doc/python-type-hinting.md
+++ b/doc/python-type-hinting.md
@@ -7,7 +7,7 @@ As a team we have committed to adding type hints throughout the Python code. Thi
 Specify types for variables initialised with `None`:
 
 ```python
-description: str = None
+description: Optional[str] = None
 ```
 
 Specify types for empty collections:
--- a/doc/python-type-hinting.md
+++ b/doc/python-type-hinting.md
@@ -71,7 +71,7 @@ from __future__ import annotations
 
 class TestClass:
 
-  def test(self, var: list[TestClass]) -> None:
+  def test(self, var: Sequence[TestClass]) -> None:
 ```
 
 This import is not necessary in Python 3.10.
--- a/doc/python-type-hinting.md
+++ b/doc/python-type-hinting.md
@@ -1,0 +1,85 @@
+# Python Type Hinting
+
+As a team we have committed to adding type hints throughout the Python code. This document defines our approach to type hinting, to ensure consistency in the codebase and avoid repeating discussions about how we apply typing.
+
+## Variables
+
+Specify types for variables initialised with `None`:
+
+```python
+description: str = None
+```
+
+Specify types for empty collections:
+
+```python
+items: list[str] = []
+mappings: dict[str, int] = {}
+```
+
+In all other places, types for variables are optional.
+
+## Standard collections
+
+For standard collections, use lower case names e.g `list` rather than `List`:
+
+```python
+items: list[int] = [1]
+mappings: set[int] = {1, 2}
+```
+
+https://www.python.org/dev/peps/pep-0585/
+
+## Generic types
+
+Specify at least the top-level type parameters for all generic types:
+
+```python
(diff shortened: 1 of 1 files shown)


doc/python-type-hinting.md
```diff
@@ -0,0 +1,85 @@
...
+def get_objects_matching(ids: Sequence[str]) -> dict[int, dict]
+```
+
+## Optional 
+
+When something is a single type and is optional use `Optional`:
+
+```python
+def test(self, var: Optional[int]) -> None:
+```
```
MebinAbraham, line 47: I think a point should be made between optional in which a `NoneType` value is valid, vs Optional where you have a default value. i.e a var which has a default value of 0, is optional in the sense that the caller does not need to provide a value, but it must be of one type. eg.:
```python
def test(self, increment_by: int = 0) -> None:
```
In the example above, the var `increment_by` is "optional" but must be of type int if provided, where as the `Optional[int]` should only be used if `None` is allowed, because we may evaluate like such `if var is None` or like `if var` etc.

doc/python-type-hinting.md
```diff
@@ -0,0 +1,85 @@
...
+
+When something can be multiple types and is optional use `Union` with `None`:
+
+```python
+def test(self, var: Union[None, int, str]) -> None:
```
MebinAbraham, line 52: Any specific reason for this over `Optional[Union[int,str]]` since `Optional[int]` is just syntatic sugar for `Union[None, int]`?.

MebinAbraham (commented):
Looks good. Just a few thoughts.

doc/python-type-hinting.md
ajmaddaford: Personal preference I guess, I think it reads more clearly without the nested `[]`.

doc/python-type-hinting.md
ajmaddaford: Good point, I have conflated optional arguments with arguments that can have the value `None`. I will reword this.

ajmaddaford: Updated.

doc/python-type-hinting.md
MebinAbraham: Fair enough. I do think Union without Optional is better syntactically, but I do think "Optional" makes the easier to understand. Also, IDE like PyCharm suggests converting Union with None to Optional Union. 

The recommendation in the image is slightly wrong as the Union shouldn't include `None` but it does execute it successfully to `Optional[Union[str, int]]`. The inspection rule:
```
Inspection info: This inspection reports when the Union[None, ...] annotation was used instead Optional[Union[...]]. 
Union[object, None] -> Optional[object] 
Union[int, string, None] -> Optional[Union[int, string]] 
Union[None] -> None
```
image

Happy to keep as is, if that is the preference by the majority.

JamesGardiner: I prefer the flat structure of `Union[None, int, str]`

JamesGardiner (approved):

doc/python-type-hinting.md
```diff
@@ -0,0 +1,87 @@
...
+## Standard collections
+
+For standard collections, use lower case names e.g `list` rather than `List`:
+
+```python
+items: list[int] = [1]
+mappings: set[int] = {1, 2}
+```
+
+https://www.python.org/dev/peps/pep-0585/
```
JamesGardiner, line 31: I think this could be expanded to include a brief summary of why while also linking to the PEP

ajmaddaford: It was a conscious thing to not explain the why, to keep the guide brief and to the point.

MebinAbraham (approved):

PR #593, opened 2021-05-06
Add types to QuestionnaireSchema

ajmaddaford:
- Adds types to QuestionnaireSchema
- Adds a mypy config to enforce types for all definitions in QuestionnaireSchema

### How to review 
- Are the types appropriate?
- Does it meet the [style guide](https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/doc/python-type-hinting.md)?
- Does anything need to be added to the style guide?

### Checklist


app/views/contexts/hub_context.py
```diff
@@ -45,7 +45,7 @@ def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
         rows = self._get_rows(enabled_section_ids)
 
         if survey_complete:
-            submission_schema = self._schema.get_submission()
+            submission_schema: dict = self._schema.get_submission() or {}
```
JamesGardiner, line 48: Doesn't the `get_submission` method default to a `dict`?

app/views/contexts/section_summary_context.py
```diff
@@ -234,4 +234,6 @@ def _add_link(self, summary, list_collector_block):
             )
 
     def _get_safe_page_title(self, title):
-        return safe_content(self._schema.get_single_string_value(title))
+        return (
+            safe_content(self._schema.get_single_string_value(title)) if title else ""
+        )
```
MebinAbraham, line 239: Don't see why this changed?

MebinAbraham (commented):
🎉

app/questionnaire/questionnaire_schema.py
```diff
@@ -202,132 +211,159 @@ def _section_ids_associated_to_list_name(self, list_name: str) -> List:
         return section_ids
 
     @staticmethod
-    def get_blocks_for_section(section):
+    def get_blocks_for_section(section: Mapping) -> Iterator[ImmutableDict]:
```
MebinAbraham, line 214: Since we said we are making the return types as specific as possible, this should return a `Generator`?. Few other occurrences also.

app/questionnaire/questionnaire_schema.py
```diff
@@ -439,28 +487,37 @@ def get_relationship_collectors_by_list_name(self, list_name: str):
...
+            ]
+            return plural_placeholder_string
+
+        placeholder_string: str = schema_object["text"]
+        return placeholder_string
 
     @staticmethod
-    def get_all_questions_for_block(block):
+    def get_all_questions_for_block(block: Mapping) -> list[ImmutableDict]:
         all_questions = []
```
MebinAbraham, line 521: We should specify types for empty collections?

app/views/handlers/block.py
```diff
@@ -33,7 +33,8 @@ def __init__(
         self._request_args = request_args or {}
         self._form_data = form_data
 
-        self.block = self._schema.get_block(current_location.block_id)
+        if self._current_location.block_id:
+            self.block = self._schema.get_block(self._current_location.block_id)
```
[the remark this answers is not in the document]
ajmaddaford: At this point the block id will never be not set on the location, but the location class itself defines it as optional. The real issue is that the location class should define `block_id` as a required attribute (which isn't done as in a handful of places we use Location objects without block ids).

app/views/contexts/hub_context.py
```diff
@@ -45,7 +45,7 @@ def get_context(self, survey_complete, enabled_section_ids) -> Mapping:
         rows = self._get_rows(enabled_section_ids)
 
         if survey_complete:
-            submission_schema = self._schema.get_submission()
+            submission_schema: Mapping = self._schema.get_submission() or {}
```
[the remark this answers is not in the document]
ajmaddaford: This was added to stop mypy failing. This should be looked at in detail when we apply typing properly to this file.

app/questionnaire/questionnaire_schema.py
```diff
@@ -23,9 +23,11 @@
 
 
 class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
-    def __init__(self, questionnaire_json, language_code=DEFAULT_LANGUAGE_CODE):
-        self._parent_id_map = {}
-        self._list_name_to_section_map = {}
+    def __init__(
+        self, questionnaire_json: Mapping, language_code: str = DEFAULT_LANGUAGE_CODE
```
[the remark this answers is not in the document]
ajmaddaford: I don't think the extra types add value here.

app/questionnaire/questionnaire_schema.py
ajmaddaford: Good spot, this was done before we agreed on the return types and I hadn't gone back to it. I have just tried changing it and there is one downside, the return type for these when using `Generator` is `Generator[ImmutableDict, None, None]`. Thoughts?

For reference the my docs can be found at https://mypy.readthedocs.io/en/stable/kinds_of_types.html#generators

app/questionnaire/questionnaire_schema.py
ajmaddaford: Do you mean the return type? Is that not covered by the existing type?

app/questionnaire/questionnaire_schema.py
MebinAbraham: I'm okay with that, just how it has to be but does make it specific.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
MebinAbraham: Just to make it slightly more specific, but I guess we serialize it and return `Any` so the guess it doesn't matter.

app/questionnaire/questionnaire_schema.py
ajmaddaford: Ok, done.

app/questionnaire/questionnaire_schema.py
ajmaddaford: Good spot, I have added types for all the variables initialised with an empty collection.

JamesGardiner (approved):

MebinAbraham (approved):

PR #598, opened 2021-05-10
Replace simplejson with rapidjson

LJBabbage:
This PR replaces simplejson with rapidjson

### How to review 
Make sure everything works and nothing has been missed

### Checklist

### DIFF
--- a/app/data_models/questionnaire_store.py
+++ b/app/data_models/questionnaire_store.py
@@ -1,10 +1,11 @@
 from types import MappingProxyType
 
-import simplejson as json
+import rapidjson as json
 
 from app.data_models.answer_store import AnswerStore
 from app.data_models.list_store import ListStore
 from app.data_models.progress_store import ProgressStore
+from app.utilities.for_json import for_json
 
 
 class QuestionnaireStore:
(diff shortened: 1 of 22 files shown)


app/utilities/for_json.py
```diff
@@ -0,0 +1,7 @@
+def for_json(object):
```
LJBabbage, line 1: This could feasibly be in helpers

app/data_models/questionnaire_store.py
```diff
@@ -58,7 +59,7 @@ def serialize(self):
             "PROGRESS": self.progress_store.serialize(),
             "RESPONSE_METADATA": self.response_metadata,
         }
-        return json.dumps(data, for_json=True)
+        return json.dumps(data, default=for_json, number_mode=json.NM_DECIMAL)
```
ajmaddaford, line 62: I think we should consider introducing a wrapper method for loading and dumping JSON that handles the `default` and `number_mode`. Something like:

```python
def load_json(data):
    return json.loads(data, number_mode=json.NM_DECIMAL)

def dump_json(data):
    return json.dumps(data, default=for_json, number_mode=json.NM_DECIMAL)
```

This would also abstract away the JSON library, so it could be easily swapped out.

petechd (approved):
Other then what @ajmaddaford mentioned on the card and in the session, this looks fine in tests!

PR #601, opened 2021-05-12
Refactor template_helpers.py

JamesGardiner:
Creates a new settings env var (`SURVEY_TYPE`) and refactors the `template_helpers` module.

Supports 6 survey types: [ `business` | `social` | `census` | `health` | `census-nisra` | `default`].

The `default` survey type has only been included to maintain support for current test schema definitions.

The default survey type if one is not set is `census`, this will change to `business` in future definitions.

### How to review 
Test functionality works. Set and unset the `SURVEY_TYPE` env var.

### Checklist


app/helpers/template_helpers.py
```diff
@@ -41,116 +160,69 @@ def get_page_header_context(language: str, theme: str) -> Dict[str, Any]:
...
-            **default_context,
-        },
-        "northernireland": {
-            **default_context,
-        },
-        "census": {
-            **default_context,
-        },
-        "census-nisra": {
+        context = {
```
berroar, line 168:
```suggestion
        return {
            **default_context,
            "copyrightDeclaration": {
                "copyright": lazy_gettext(
                    "Crown copyright and database rights 2021 NIMA MOU577.501."
                ),
                "text": lazy_gettext(
                    "Use of address data is subject to the terms and conditions."
                ),
            },
            "lang": "en",
            "poweredBy": {
                "logo": "nisra-logo-black-en",
                "alt": "NISRA - Northern Ireland Statistics and Research Agency",
            },
        }
```

app/settings.py
```diff
@@ -137,3 +137,5 @@ def utcoffset_or_fail(date_value, key):
 ADDRESS_LOOKUP_API_AUTH_TOKEN_LEEWAY_IN_SECONDS = int(
     os.getenv("ADDRESS_LOOKUP_API_AUTH_TOKEN_LEEWAY_IN_SECONDS", "300")
 )
+
+SURVEY_TYPE = os.getenv("SURVEY_TYPE", "census")
```
petechd, line 141: I'm struggling to get this working in my local runner after adding env variable when I use test schemas, they have hardcoded default themes. Do you know in which schemas should I test this?

app/helpers/template_helpers.py
```diff
@@ -14,25 +14,144 @@
...
+            )
+        else:
+            help_path = "help/how-to-answer-questions/online-questions-help/"
+            cookies_path = "cookies/"
+            accessibility_statement_path = "accessibility-statement/"
+            privacy_and_data_protection_path = "privacy-and-data-protection/"
+            terms_and_conditions_path = "terms-and-conditions/"
+            contact_us = "contact-us/"
+            languages = "help/languages-and-accessibility/languages/"
+            bsl_and_audio_videos = "help/languages-and-accessibility/accessibility/accessible-videos-with-bsl/"
```
ajmaddaford, line 47: These are Census specific urls.

app/helpers/template_helpers.py
```diff
@@ -14,25 +14,144 @@
...
+
+
+class CensusNISRAContextHelper(ContextHelper):
+    def __init__(self, language: str, base_url: str):
+        super().__init__(language, base_url)
+        for entry in ("languages", "bsl_and_audio_videos"):
+            self.static_content_urls.pop(entry)
+        self.static_content_urls[
+            "help"
+        ] = f"{base_url}help/help-with-the-questions/online-questions-help/"
```
ajmaddaford, line 150: I think this is wrong. If `languages` and `bsl_and_audio_videos` don't apply to all contexts then they shouldn't be set in the common class.

app/helpers/template_helpers.py
```diff
@@ -41,116 +160,69 @@ def get_page_header_context(language: str, theme: str) -> Dict[str, Any]:
...
+        context = {
             **default_context,
+            "copyrightDeclaration": {
+                "copyright": lazy_gettext(
+                    "Crown copyright and database rights 2021 NIMA MOU577.501."
+                ),
+                "text": lazy_gettext(
+                    "Use of address data is subject to the terms and conditions."
+                ),
+            },
```
ajmaddaford, line 177: I think the structure required by rendering should be returned in one place, the common Context class. We shouldn't be creating that rendering context and then mutating it in the sub classes.

app/helpers/template_helpers.py
JamesGardiner: This has changed slightly, so an early return is no longer an option

app/settings.py
JamesGardiner: I'd try pulling in one of the census schemas manually @petechd - apologies for it being a pain

petechd (commented):
The only time I can get this env var `SURVEY_TYPE` based theme working is when in session.py file on line 84 I change:
`cookie_session["theme"] = g.schema.json["theme"]`
to:
`cookie_session["theme"] = current_app.config["SURVEY_TYPE"]`
shouldn’t that be changed?

tests/app/helpers/test_template_helpers.py
```diff
@@ -19,42 +19,42 @@ def test_get_footer_context_nisra_theme(app):
...
+                        Link(
+                            text="Help",
+                            url="https://census.gov.uk/ni/help/help-with-the-questions/online-questions-help/",
+                            target="_blank",
+                        ),
+                        Link(
+                            text="Contact us",
+                            url="https://census.gov.uk/ni/contact-us/",
+                            target="_blank",
+                        ),
```
ajmaddaford, line 31: Python objects shouldn't be being passed to the templates, this should return as it did previously.

app/helpers/template_helpers.py
```diff
@@ -14,215 +14,299 @@
...
+        ]
+    )
 
 
-def _map_theme(theme: str) -> str:
-    """Maps a survey schema theme to a design system theme
+@dataclass
+class CensusContextOptions(ContextOptions):
+    title_logo: str = "census-logo-en"
+    title_logo_alt: str = lazy_gettext("Census 2021")
```
ajmaddaford, line 214: I can't see where the links are defined for Census?

app/helpers/template_helpers.py
```diff
@@ -14,215 +14,299 @@
...
-                ],
-            },
-        ],
-    }
+            "cookie_settings_url": self._cookie_settings_url,
+            "page_header": self.page_header_context,
+            "footer": self.footer_context,
+            "languages": get_languages_context(),
+            "theme": self._map_theme(self._theme),
+            "language_code": get_locale().language,
```
ajmaddaford, line 92: language should be passed in, as we already call `get_locale().language` prior to this. We should also consider passing it to `get_languages_context` as that would mean the language helper also wouldn't need to call `get_locale().language`.

app/helpers/template_helpers.py
```diff
@@ -14,215 +14,299 @@
...
+            "address_lookup_api_url": self._address_lookup_api,
+            "data_layer": get_data_layer(self._theme),
+            "include_csrf_token": (
+                request.url_rule and "POST" in request.url_rule.methods
+            ),
+            "google_tag_manager_id": self._google_tag_manager_id,
+            "google_tag_manager_auth": self._google_tag_manager_auth,
+        }
+
+    @cached_property
```
ajmaddaford, line 105: Is this method ever called more than once per request?

app/helpers/template_helpers.py
```diff
@@ -14,215 +14,299 @@
...
+    base_url: str,
+    language: str,
+) -> Dict[str, Any]:
+    context_options = {
+        "business": ContextOptions,
+        "health": ContextOptions,
+        "social": ContextOptions if language == "en" else ContextOptionsCymraeg,
+        "census": (
+            CensusContextOptions if language == "en" else CensusContextOptionsCymraeg
+        ),
```
ajmaddaford, line 278: The logic here should be reversed as the default should always be `en`.

ajmaddaford (commented):
Previously the page header context was cached outside of the request context (`@lru_cache(maxsize=None)`). I think we should aim for a similar level of caching. As most of the context doesn't change per request we should separate out the things that do e.g. the footer warning and including the csrf token so that most of the generated context can live in a long term cache.

app/data_models/answer_store.py
```diff
@@ -41,7 +41,7 @@ def __eq__(self, other):
 
     @staticmethod
     def _build_map(answers: List[Dict]):
-        """ Builds the answer_store's data structure from a list of answer dictionaries"""
+        """Builds the answer_store's data structure from a list of answer dictionaries"""
```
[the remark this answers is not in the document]
JamesGardiner: Done

app/helpers/template_helpers.py
```diff
@@ -165,33 +237,30 @@ def _map_theme(theme: str) -> str:
...
     include_csrf_token = request.url_rule and "POST" in request.url_rule.methods
     account_service_url = cookie_session.get(
         "account_service_url", f"{CENSUS_EN_BASE_URL}en/start"
     )
     sign_out_url = url_for("session.get_sign_out")
-    static_content_urls = get_static_content_urls(language_code, base_url, theme)
-    footer_context = get_footer_context(
-        language_code,
-        static_content_urls,
+    footer_context = context_helper.footer_context(
```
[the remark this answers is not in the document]
JamesGardiner: Done

app/helpers/template_helpers.py
JamesGardiner: Done

app/helpers/template_helpers.py
JamesGardiner: Done

app/helpers/template_helpers.py
JamesGardiner: Done

tests/app/helpers/test_template_helpers.py
JamesGardiner: What's the reason for not passing Python objects, especially as what we are passing already are objects?

app/helpers/template_helpers.py
JamesGardiner: Done

tests/app/helpers/test_template_helpers.py
ajmaddaford: We have previously had complexity caused by complex python objects being passed to and used by templates, so we have tried to not. For that one it's also that it's using the class for two different purposes - one to configure the links for a given runner mode, and then to serialise the link context. It works as it the two use the same property names, but that ties them together and isn't clear.

It also just so happens to work for that scenario as the property names are single words, in other instances the template property names are camel cased which we wouldn't have as python prroperty names.

app/helpers/template_helpers.py
```diff
@@ -14,215 +14,299 @@
...
+        theme: str,
+        base_url: str,
+        context_options: ContextOptions = ContextOptions(),
+    ):
+        self._base_url = base_url
+        self.context_options = context_options or ContextOptions()
+        self._theme = theme
+        self._sign_out_url = url_for("session.get_sign_out")
+        self._account_service_url = cookie_session.get(
+            "account_service_url", f"{self._base_url}en/start"
```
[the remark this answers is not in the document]
JamesGardiner: I have updated this so that account_service_url is an attribute of the ContextOptions class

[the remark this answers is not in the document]
JamesGardiner: 3e8ca0c98b8

app/helpers/template_helpers.py
```diff
@@ -14,215 +14,299 @@
...
-        ],
-    }
+            "cookie_settings_url": self._cookie_settings_url,
+            "page_header": self.page_header_context,
+            "footer": self.footer_context,
+            "languages": get_languages_context(),
+            "theme": self._map_theme(self._theme),
+            "language_code": get_locale().language,
+            "schema_theme": self._theme,
+            "survey_title": lazy_gettext("Census 2021"),
```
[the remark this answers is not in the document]
JamesGardiner: Done

app/helpers/template_helpers.py
```diff
@@ -232,73 +316,11 @@ def get_census_base_url(schema_theme: str, language_code: str) -> str:
...
-        # Replace piping with ellipsis
-        content = re.sub(r"{.*?}", "…", content)
-        # Strip HTML Tags
-        content = re.sub(r"</?[^>]+>", "", content)
-    return content
-
-
-def get_data_layer(schema_theme: str) -> List:
-    if schema_theme == "census-nisra":
+def get_data_layer(theme: str) -> List:
```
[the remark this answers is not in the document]
JamesGardiner: It makes sense for it to be set on the context options - it's not used anywhere else apart from the single usage in this module

app/helpers/template_helpers.py
JamesGardiner: No - I'll remove the cache

app/helpers/template_helpers.py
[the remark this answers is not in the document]
JamesGardiner: Updated so that it's now set on the ContextOptions class. Note that it needs to be set on the `__post_init__` methods for Census classes so that it overrides the mutable default value of `[]`

app/helpers/template_helpers.py
```diff
@@ -14,215 +14,299 @@
...
+    @staticmethod
+    def _map_theme(theme: str) -> str:
+        """Maps a survey schema theme to a design system theme
+
+        :param theme: A schema defined theme
+        :returns: A design system theme
+        """
+        if theme and theme not in ["census", "census-nisra"]:
+            return "main"
+        return "census"
```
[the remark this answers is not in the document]
JamesGardiner: Done

app/helpers/template_helpers.py
JamesGardiner: Done 0e1ecab1258

tests/app/helpers/test_template_helpers.py
JamesGardiner: Done

app/helpers/template_helpers.py
```diff
@@ -14,215 +14,299 @@
...
+
+
+@dataclass
+class ContextOptionsCymraeg(ContextOptions):
```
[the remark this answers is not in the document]
JamesGardiner: Done 3142831e7

app/helpers/template_helpers.py
JamesGardiner: Added in 3142831e7

app/helpers/template_helpers.py
```diff
@@ -14,215 +14,299 @@
...
+
+    return ContextHelper(
+        theme,
+        base_url,
+        context_options[theme](),
+    ).context
 
+
+def render_template(template: str, **kwargs: Mapping) -> str:
+    theme = cookie_session.get("theme", current_app.config["SURVEY_TYPE"])
```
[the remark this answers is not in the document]
JamesGardiner: As discussed at tech session, this is acceptable while the business define for us the differentiation and requirements of survey type and theme

app/helpers/template_helpers.py
JamesGardiner: Done

MebinAbraham (commented):
Just a few minor comments.

app/helpers/template_helpers.py
```diff
@@ -14,291 +14,343 @@
...
+    text: str
+    url: str
+    target: Optional[str] = "_blank"
+
+
+@dataclass
+class ContextOptions:
+    """Valid options for defining context."""
+
+    page_header_logo: Optional[str] = lazy_gettext("ons-logo-pos-en")
```
MebinAbraham, line 30: Don't need the `lazy_gettext` cast.

app/helpers/template_helpers.py
```diff
@@ -14,291 +14,343 @@
...
+                "accessibility-statement/",
+            ).__dict__,
+            Link(
+                lazy_gettext("Privacy and data protection"),
+                "privacy-and-data-protection/",
+            ).__dict__,
+            Link(
+                lazy_gettext("Terms and conditions"),
+                "terms-and-conditions/",
+            ).__dict__,
```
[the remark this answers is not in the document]
ajmaddaford: I'm not sure that would work with the script that extracts the strings to be translated from code.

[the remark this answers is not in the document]
MebinAbraham: Whoops, I edited my comment as you posted. Shame, would have been nice to reduce the duplication.

app/helpers/template_helpers.py
```diff
@@ -14,291 +14,343 @@
...
+    )
 
-def _map_theme(theme: str) -> str:
-    """Maps a survey schema theme to a design system theme
+    def __post_init__(self) -> None:
+        self.data_layer = [{"nisra": False}]
+
+
+@dataclass
+class CensusContextOptionsCymraeg(CensusContextOptions):
```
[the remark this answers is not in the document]
JamesGardiner: Because the language code we use is 'cy'

app/helpers/template_helpers.py
```diff
@@ -14,291 +14,343 @@
...
+    title_logo_alt: str = lazy_gettext("Census 2021")
+    account_service_url = f"{CENSUS_EN_BASE_URL}/en/start"
+    design_system_theme = "census"
+    footer_links: Iterable[Mapping] = field(
+        default_factory=lambda: [
+            Link(
+                lazy_gettext("Help"),
+                "help/help-with-the-questions/online-questions-help/",
+            ).__dict__,
+            Link(lazy_gettext("Contact us"), "contact-us/").__dict__,
```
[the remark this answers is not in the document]
JamesGardiner: What's your reason for not using the dunder method?

app/helpers/template_helpers.py
```diff
@@ -14,291 +14,343 @@
...
+    page_header_logo: Optional[str] = lazy_gettext("ons-logo-pos-en")
+    page_header_logo_alt: Optional[str] = lazy_gettext(
+        "Office for National Statistics logo"
+    )
+    copyright_declaration: Optional[str] = lazy_gettext(
+        "Crown copyright and database rights 2020 OS 100019153."
+    )
+    copyright_text: Optional[str] = lazy_gettext(
+        "Use of address data is subject to the terms and conditions."
+    )
```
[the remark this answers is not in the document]
JamesGardiner: as in `from flask_babel.speaklater import LazyString`?

app/helpers/template_helpers.py
```diff
@@ -14,291 +14,343 @@
...
     )
+    powered_by_logo: str = "nisra-logo-black-en"
+    powered_by_logo_alt: str = "NISRA - Northern Ireland Statistics and Research Agency"
+    account_service_url = CENSUS_NIR_BASE_URL
+
+    def __post_init__(self) -> None:
+        self.data_layer = [{"nisra": True}]
+
+
+def generate_context(
```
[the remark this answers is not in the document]
JamesGardiner: This breaks almost all of our tests - with a validation error.

```
flask_wtf.csrf.CSRFError: 400 Bad Request: The CSRF token is missing.
```

app/helpers/template_helpers.py
```diff
@@ -14,291 +14,343 @@
...
+        context_options[theme](),
+    ).context
+
 
+def render_template(template: str, **kwargs: Mapping) -> str:
+    # The fallback to assigning SURVEY_TYPE to theme is only being added until
+    # business feedback on the differentiation between theme and SURVEY_TYPE.
+    theme = cookie_session.get("theme", current_app.config["SURVEY_TYPE"])
+    language = get_locale().language
+    base_url = get_base_url("theme", language)
```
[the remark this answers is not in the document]
JamesGardiner: Yeah - the gap is around testing for `census-nisra` (that's the only theme check we do when getting the base url, otherwise it defaults to `CENSUS_EN_BASE_URL`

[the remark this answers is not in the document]
JamesGardiner: and more specifically, testing the render_template method, which it doesn't seem is tested anywhere

app/helpers/template_helpers.py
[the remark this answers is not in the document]
JamesGardiner: Changed these to the above `LazyString` and mypy checks fine 👍🏻

app/helpers/template_helpers.py
```diff
@@ -14,291 +14,343 @@
...
+            "data_layer": self.context_options.data_layer,
+            "include_csrf_token": (
+                request.url_rule and "POST" in request.url_rule.methods
+            ),
+            "google_tag_manager_id": self._google_tag_manager_id,
+            "google_tag_manager_auth": self._google_tag_manager_auth,
+        }
+
+    @property
+    def page_header_context(self) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
JamesGardiner: Done

app/helpers/template_helpers.py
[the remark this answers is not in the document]
MebinAbraham: Which is an official language code?

app/helpers/template_helpers.py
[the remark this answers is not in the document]
MebinAbraham: Developer preference I guess, but I don't think it's very readable to use dunders directly. I would say it is more of an internal representation which should have a wrapper and not invoked directly outside the object. Plus it is hard without checking code if there is our own implementation to support `dict()` on the object. The other objects we have, all use a wrapper method so thought it would also make that consistent. If we are keeping `__dict__` then there is one reference to `asdict()` which should be updated to use the dunder for consistency.

app/helpers/template_helpers.py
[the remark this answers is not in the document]
MebinAbraham: Guess we could close the gap by making the schema theme and language code like static known types instead of any arbitary string?. Prob for a future card.

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,378 @@
...
+    title_logo_alt: LazyString = lazy_gettext("Census 2021")
+    base_url: str = CENSUS_EN_BASE_URL
+    account_service_url = f"{base_url}/en/start"
+    design_system_theme = "census"
+    footer_links: Iterable[MutableMapping] = field(
+        default_factory=lambda: [
+            Link(
+                lazy_gettext("Help"),
+                f"{CENSUS_EN_BASE_URL}/help/help-with-the-questions/online-questions-help/",
+            ).__dict__,
```
MebinAbraham, line 183: The previous logic was:
```py
		help_path = (
            "help/help-with-the-questions/online-questions-help/"
            if is_nisra_theme
            else "help/how-to-answer-questions/online-questions-help/"
        )
```

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,378 @@
...
-                ],
-            },
-        ],
-    }
+            "google_tag_manager_id": self._google_tag_manager_id,
+            "google_tag_manager_auth": self._google_tag_manager_auth,
+        }
+
+    @property
+    def page_header_context(self) -> dict[str, Any]:
```
MebinAbraham, line 112: The header background seems to be misconfigured. For census, it is no longer purple:
image

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,378 @@
...
+        ):
+            context["poweredBy"] = {
+                "logo": self.context_options.powered_by_logo,
+                "alt": self.context_options.powered_by_logo_alt,
+            }
+
+        return context
+
+    @cached_property
+    def _footer_warning(self) -> Union[str, None]:
```
MebinAbraham, line 162: Style guide was to only use Union with None, when it is more than 1 type excluding None.
```suggestion
    def _footer_warning(self) -> Optional[str]:
```

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,378 @@
...
-    :returns (str): Modified string.
-    """
-    if content is not None:
-        # Replace piping with ellipsis
-        content = re.sub(r"{.*?}", "…", content)
-        # Strip HTML Tags
-        content = re.sub(r"</?[^>]+>", "", content)
-    return content
 
+def render_template(template: str, **kwargs: Union[str, Mapping]) -> Text:
```
MebinAbraham, line 362: How come `Text` is used instead of `str`?

MebinAbraham (commented):
Last few minor comments from me. Looks good.

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,378 @@
...
-            else lazy_gettext("Crown copyright and database rights 2020 OS 100019153."),
-            "text": lazy_gettext(
-                "Use of address data is subject to the terms and conditions."
+    @property
+    def context(self) -> dict[str, Any]:
+        return {
+            "account_service_url": self._account_service_url,
+            "account_service_log_out_url": self._account_service_log_out_url,
+            "contact_us_url": Link(
+                "Contact us", f"{self.context_options.base_url}/contact-us/"
```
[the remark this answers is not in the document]
JamesGardiner: We don't translate urls do we?

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,378 @@
...
+from app.settings import CENSUS_CY_BASE_URL, CENSUS_EN_BASE_URL, CENSUS_NIR_BASE_URL
 
-CENSUS_EN_BASE_URL = "https://census.gov.uk/"
-CENSUS_CY_BASE_URL = "https://cyfrifiad.gov.uk/"
-CENSUS_NIR_BASE_URL = f"{CENSUS_EN_BASE_URL}ni/"
-DEFAULT_THEME = "census"
 
+@dataclass
+class Link:
+    text: Union[str, LazyString]
```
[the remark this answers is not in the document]
JamesGardiner: It can be, I wanted to be permissive with the interface. Given we aren't using any in this PR, I'll make it more restrictive

app/helpers/template_helpers.py
JamesGardiner: It shouldn't be - `str` is correct in this scenario

app/helpers/template_helpers.py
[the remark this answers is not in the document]
JamesGardiner: ah you are referring to `Contact us` - yes this should be wraped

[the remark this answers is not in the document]
MebinAbraham: Ya the `Contact us`.

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,379 @@
...
+    title_logo: Optional[str] = None
+    title_logo_alt: Optional[str] = None
+    header_logo: Optional[str] = None
+    mobile_logo: Optional[str] = None
+    powered_by_logo: Optional[str] = None
+    powered_by_logo_alt: Optional[str] = None
+    crest: bool = True
+    footer_links: Optional[Iterable[MutableMapping]] = None
+    footer_legal_links: Optional[Iterable[Mapping]] = None
+    survey_title: Optional[LazyString] = lazy_gettext("Census 2021")
```
ajmaddaford, line 49: This shouldn't have a default of `Census 2021`, it should come from the derived options.

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,379 @@
...
 
-    :param theme: A schema defined theme
-    :returns: A design system theme
-    """
-    if theme and theme not in ["census", "census-nisra"]:
-        return "main"
-    return "census"
+@dataclass
+class CensusContextOptionsCymraeg(
+    CensusContextOptions,
```
ajmaddaford, line 224: Does this use any properties defined in `CensusContextOptionsCymraeg`? Even if it does it might be clearer to decide that everything should inherit directly from `ContextOptions`.

Also, the class name should end with `ContextOptions`, so should be `CensusCymraegContextOptions` or `CymraegCensusContextOptions`.

app/helpers/template_helpers.py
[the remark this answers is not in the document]
JamesGardiner: Done

app/helpers/template_helpers.py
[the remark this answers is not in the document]
JamesGardiner: Done

app/routes/session.py
```diff
@@ -133,17 +129,18 @@ def get_sign_out():
     Signs the user out of eQ and redirects to the log out url.
     """
     if not cookie_session:
-        log_out_url = get_census_base_url(None, None)
+        log_out_url = get_base_url(None, None)
 
     elif (
-        "census" in (theme := cookie_session.get("theme", DEFAULT_THEME))
+        "census"
+        in (theme := cookie_session.get("theme", current_app.config["SURVEY_TYPE"]))
```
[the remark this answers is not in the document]
JamesGardiner: Done 0211eeada

app/helpers/language_helper.py
```diff
@@ -34,26 +34,28 @@ def handle_language():
             session_store.save()
 
 
-def get_languages_context():
+def get_languages_context(current_language: str) -> Union[dict[str, Any], None]:
```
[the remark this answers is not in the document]
JamesGardiner: Done c34718c7e

app/storage/metadata_parser.py
```diff
@@ -167,6 +169,6 @@ def validate_questionnaire_claims(claims, questionnaire_specific_metadata):
 
 
 def validate_runner_claims(claims: Dict):
-    """ Validate claims required for runner to function"""
+    """Validate claims required for runner to function"""
```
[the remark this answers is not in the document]
JamesGardiner: argh - this is vs code's work 😆  It keeps autoformatting every file I touch. May need to open a separate PR for this to preserve my sanity

app/views/contexts/submit_questionnaire_context.py
```diff
@@ -52,8 +52,8 @@ def _get_summary_context(
             "summary_type": "Summary",
         }
 
-    def _build_all_groups(self) -> Generator[dict, None, None]:
-        """ NB: Does not support repeating sections """
+    def _build_all_groups(self):
+        """NB: Does not support repeating sections"""
```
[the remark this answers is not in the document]
JamesGardiner: No - reverting it in the next push

app/storage/metadata_parser.py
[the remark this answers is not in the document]
JamesGardiner: Done ddca78d2c

app/views/contexts/submit_questionnaire_context.py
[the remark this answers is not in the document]
JamesGardiner: Done ddca78d2c

app/helpers/template_helpers.py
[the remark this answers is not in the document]
JamesGardiner: Yes

app/helpers/template_helpers.py
JamesGardiner: Done

app/helpers/template_helpers.py
JamesGardiner: Done

app/helpers/__init__.py
```diff
@@ -1,10 +1,22 @@
...
 
+
+def get_base_url(schema_theme: str, language_code: str) -> str:
+    if language_code == "cy":
+        return CENSUS_CY_BASE_URL
+
+    if schema_theme == "census-nisra":
+        return CENSUS_NIR_BASE_URL
+
+    return CENSUS_EN_BASE_URL
```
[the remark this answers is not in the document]
JamesGardiner: It won't. But at the same time, we have no requirements for how to select a base url for anything other than census (that I am aware of). When we do, this will need updating.

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,379 @@
...
-    page_header_context.update({"title": survey_title})
-    google_tag_manager_context = get_google_tag_manager_context()
-    cdn_url = f'{current_app.config["CDN_URL"]}{current_app.config["CDN_ASSETS_PATH"]}'
-    base_url = get_census_base_url(theme, language_code)
-    include_csrf_token = request.url_rule and "POST" in request.url_rule.methods
-    account_service_url = cookie_session.get(
-        "account_service_url", f"{CENSUS_EN_BASE_URL}en/start"
+@dataclass
+class CensusNISRAContextOptions(
+    CensusContextOptions,
```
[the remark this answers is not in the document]
JamesGardiner: Yes - it uses the design_system_theme, title_logo, title_logo_alt and survey_title

app/helpers/template_helpers.py
JamesGardiner: As discussed IRL, this was a result of an incorrect assignment in the dataclass, which has now been fixed

app/helpers/template_helpers.py
JamesGardiner: It uses title_logo_alt, design_system_theme, and survey_title

app/helpers/template_helpers.py
JamesGardiner: Done

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,379 @@
...
+            "languages": get_languages_context(self._language),
+            "theme": self.context_options.design_system_theme,
+            "language_code": self._language,
+            "schema_theme": self._theme,
+            "survey_title": self.context_options.survey_title,
+            "cdn_url": self._cdn_url,
+            "address_lookup_api_url": self._address_lookup_api,
+            "data_layer": self.context_options.data_layer,
+            "include_csrf_token": (
+                request.url_rule and "POST" in request.url_rule.methods
```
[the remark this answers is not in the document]
JamesGardiner: Done

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,379 @@
...
+        context_options: ContextOptions = ContextOptions(),
+    ) -> None:
+        self._theme = theme
+        self._language = language
+        self._is_post_submission = is_post_submission
+        self._include_csrf_token = include_csrf_token
+        self.context_options = context_options
+        self._sign_out_url = url_for("session.get_sign_out")
+        self._account_service_url = cookie_session.get(
+            "account_service_url", f"{self.context_options.base_url}en/start"
```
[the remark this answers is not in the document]
JamesGardiner: Done

app/helpers/__init__.py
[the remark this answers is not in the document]
JamesGardiner: Now removed this logic in favour of retrieving it from the SurveyConfig

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,376 @@
...
+                lazy_gettext("Contact us"),
+                f"{self.context_options.base_url}/contact-us/",
+            ).__dict__,
+            "cookie_settings_url": self._cookie_settings_url,
+            "page_header": self.page_header_context,
+            "footer": self.footer_context,
+            "languages": get_languages_context(self._language),
+            "theme": self.context_options.design_system_theme,
+            "language_code": self._language,
+            "schema_theme": self._theme,
```
ajmaddaford, line 100: I don't think this is used anywhere and can be removed, both the context property and the `_theme` class property.

app/helpers/__init__.py
```diff
@@ -1,10 +1,15 @@
 from .address_lookup_api_helper import get_address_lookup_api_auth_token
 from .header_helpers import get_span_and_trace
-from .template_helpers import get_census_base_url
+from .template_helpers import context_options
 from .url_safe_serializer import url_safe_serializer
 
+
+def get_base_url(schema_theme: str, language_code: str) -> str:
+    return context_options(schema_theme, language_code).base_url
```
ajmaddaford, line 8: It's a bit odd that this method accepts a `language_code` parameter but only ever gets passed `None`. 

I think we should instead have a helper method for getting the survey config, that does something like this:

```python
def get_survey_config():
    theme = cookie_session.get("theme", current_app.config["SURVEY_TYPE"])
    language = get_locale().language
    return context_options(theme, language)
```

This could then be called directly from the sign out page and the `render_template` code could be refactored to use it.

app/settings.py
```diff
@@ -137,3 +137,10 @@ def utcoffset_or_fail(date_value, key):
...
+
+
+CENSUS_EN_BASE_URL = "https://census.gov.uk"
+CENSUS_CY_BASE_URL = "https://cyfrifiad.gov.uk"
+CENSUS_NIR_BASE_URL = f"{CENSUS_EN_BASE_URL}/ni"
```
ajmaddaford, line 146: I don't think these need to be here anymore and could just be properties in the survey configs?

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,376 @@
...
+        "business": SurveyConfig,
+        "health": SurveyConfig,
+        "social": SurveyConfig,
+        "census": (
+            CymraegCensusSurveyConfig if language == "cy" else CensusSurveyConfig
+        ),
+        "default": SurveyConfig,
+        "census-nisra": CensusNISRASurveyConfig,
+    }
+    return context_options_mapping[theme]()
```
ajmaddaford, line 357: Now that `ContextOptions` has been renamed `SurveyConfig, the naming needs be looked at throughout this file e.g. the function name and variables here should be survey config rather than context options.

app/helpers/__init__.py
JamesGardiner: Done

app/settings.py
JamesGardiner: Done

app/helpers/template_helpers.py
JamesGardiner: Done

app/storage/metadata_parser.py
```diff
@@ -132,7 +134,7 @@ def convert_schema_name(self, data, **kwargs):
 
 
 def validate_questionnaire_claims(claims, questionnaire_specific_metadata):
-    """ Validate any survey specific claims required for a questionnaire"""
+    """Validate any survey specific claims required for a questionnaire"""
```
ajmaddaford, line 137: These changes look unrelated to the pull request and should be removed.

app/views/contexts/submit_questionnaire_context.py
```diff
@@ -53,7 +53,7 @@ def _get_summary_context(
         }
 
     def _build_all_groups(self) -> Generator[dict, None, None]:
-        """ NB: Does not support repeating sections """
+        """ NB: Does not support repeating sections"""
```
ajmaddaford, line 56: This change looks unrelated to the pull request and should be removed.

app/views/contexts/summary/block.py
```diff
@@ -37,7 +37,7 @@ def _build_link(self, block_id, return_to):
     def get_question(
         block_schema, answer_store, list_store, metadata, schema, location
     ):
-        """ Taking question variants into account, return the question which was displayed to the user """
+        """Taking question variants into account, return the question which was displayed to the user"""
```
ajmaddaford, line 40: This change looks unrelated to the pull request and should be removed.

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+filterwarnings =
+    ignore::DeprecationWarning:google
```
ajmaddaford, line 3: This change looks unrelated to the pull request and should be removed.

tests/app/helpers/test_template_helpers.py
```diff
@@ -1,12 +1,86 @@
...
+                        },
+                    ]
+                }
+            ],
+        }
+
+        survey_config = get_survey_config(
+            theme="census",
+            language="en",
+        )
```
ajmaddaford, line 71: `get_survey_config` should be tested independently of `ContextHelper`. For this test `ContextHelper` can be passed a direct instantiation of the `SurveyConfig` class. This applies to the other tests that do this as well.

MebinAbraham (commented):
Can we add this to mypy.ini?. Since template helpers have been rewritten, I think it makes sense to enforce types.
```
[mypy-app.helpers.template_helpers]
disallow_untyped_defs = True
warn_return_any = True
```

tests/app/helpers/test_template_helpers.py
```diff
@@ -64,114 +138,87 @@ def test_get_footer_context_nisra_theme(app):
...
 
-    assert expected == result
+        result = ContextHelper(
+            language="en",
+            is_post_submission=False,
+            include_csrf_token=True,
+            survey_config=survey_config,
+        ).context["page_header"]
+    assert result == expected
```
ajmaddaford, line 224: Are there tests for context other than the `page_header` or `page_footer`?

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,371 @@
...
+def get_survey_config(
+    theme: Optional[str] = None, language: Optional[str] = None
+) -> SurveyConfig:
+    if not language:
+        language = get_locale().language
+    if not theme:
+        theme = cookie_session.get("theme", current_app.config["SURVEY_TYPE"])
+    survey_config = survey_config_mapping(theme, language)
+
+    return survey_config
```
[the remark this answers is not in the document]
JamesGardiner: I did consider this, but I preferred the explicit assignment from accessing the survey_config_mapping. I'll let @ajmaddaford be the deciding vote

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,371 @@
...
+            context["titleLogoAlt"] = f"{self.survey_config.title_logo_alt}"
+        if self.survey_config.header_logo:
+            context["customHeaderLogo"] = self.survey_config.header_logo
+        if self.survey_config.mobile_logo:
+            context["mobileLogo"] = self.survey_config.mobile_logo
+
+        return context
+
+    @property
+    def footer_context(self):
```
ajmaddaford, line 129: The return type is missing here.

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,371 @@
...
+    if not theme:
+        theme = cookie_session.get("theme", current_app.config["SURVEY_TYPE"])
+    survey_config = survey_config_mapping(theme, language)
+
+    return survey_config
+
+
+def render_template(template: str, **kwargs: Union[str, Mapping]) -> str:
+    # The fallback to assigning SURVEY_TYPE to theme is only being added until
+    # business feedback on the differentiation between theme and SURVEY_TYPE.
```
ajmaddaford, line 355: This comment needs to move into the `get_survey_config` function.

app/helpers/template_helpers.py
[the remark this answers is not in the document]
ajmaddaford: I think with the return value type set it's probably clear enough without the explicit assignment.

app/views/contexts/submit_questionnaire_context.py
JamesGardiner: Done - apologies - this was VS Code using the incorrect formatter after an update

pytest.ini
JamesGardiner: Done

templates/signed-out.html
```diff
@@ -1,6 +1,6 @@
 {% extends 'layouts/_base.html' %}
 
-{% set page_title = _("Signed out") ~ " - " ~ _("Census 2021") %}
+{% set page_title = _("Signed out") %}
```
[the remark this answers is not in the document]
JamesGardiner: should be rendered by base?

tests/functional/spec/custom_page_titles.spec.js
```diff
@@ -18,22 +18,22 @@ describe("Feature: Custom Page Titles", () => {
     it("When I navigate to the list collector page, Then I should see the custom page title", () => {
       $(HubPage.submit()).click();
       const expectedPageTitle = browser.getTitle();
-      expect(expectedPageTitle).to.equal("Custom page title - Census 2021");
+      expect(expectedPageTitle).to.equal("Custom page title");
```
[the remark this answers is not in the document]
JamesGardiner: These won't render the census page title part as they are default surveys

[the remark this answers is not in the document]
JamesGardiner: take a look at ir schema as that still renders it

app/storage/metadata_parser.py
JamesGardiner: Done

tests/app/helpers/test_template_helpers.py
JamesGardiner: 👍🏻

JamesGardiner: Done

app/helpers/template_helpers.py
JamesGardiner: Done

app/helpers/template_helpers.py
JamesGardiner: Done

app/routes/session.py
```diff
@@ -132,23 +128,13 @@ def get_sign_out():
...
-
-    else:
-        log_out_url = cookie_session.get(
+    log_out_url = (
+        get_survey_config().base_url
+        if not cookie_session
+        else cookie_session.get(
             "account_service_log_out_url", url_for(".get_signed_out")
         )
+    )
```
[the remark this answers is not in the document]
JamesGardiner: Done

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,371 @@
...
+def survey_config_mapping(theme: str, language: str) -> SurveyConfig:
+    return {
+        "business": SurveyConfig,
+        "health": SurveyConfig,
+        "social": SurveyConfig,
+        "census": (
+            CymraegCensusSurveyConfig if language == "cy" else CensusSurveyConfig
+        ),
+        "default": SurveyConfig,
+        "census-nisra": CensusNISRASurveyConfig,
```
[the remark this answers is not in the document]
JamesGardiner: As discussed offline, this is supported in V2, so we will maintain support in this PR

app/helpers/template_helpers.py
```diff
@@ -1,304 +1,371 @@
...
+            "cdn_url": self._cdn_url,
+            "address_lookup_api_url": self._address_lookup_api,
+            "data_layer": self.survey_config.data_layer,
+            "include_csrf_token": self._include_csrf_token,
+            "google_tag_manager_id": self._google_tag_manager_id,
+            "google_tag_manager_auth": self._google_tag_manager_auth,
+        }
 
+    @property
+    def page_header_context(self) -> dict[str, Any]:
```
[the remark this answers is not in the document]
JamesGardiner: Done

tests/app/helpers/test_template_helpers.py
JamesGardiner: there aren't, but there is no logic attached to other parts of the context either

ajmaddaford: There are, for example:

```python
self._account_service_url = cookie_session.get(
    "account_service_url", self.context_options.account_service_url
)
```

or 

```python
self._cdn_url = (
    f'{current_app.config["CDN_URL"]}{current_app.config["CDN_ASSETS_PATH"]}'
)
```

tests/integration/session/test_sign_out_and_exit.py
```diff
@@ -126,39 +86,3 @@ def test_redirects_to_account_service_log_out_url_when_present(self):
         )
         self.get(SIGN_OUT_URL_PATH)
         self.assertInUrl(ACCOUNT_SERVICE_LOG_OUT_URL_PATH)
-
-
-class TestExitPostSubmissionWithHubCensusTheme(TestExitPostSubmissionTestCase):
```
MebinAbraham, line 131: This should be repurposed to test the logic:

```
    log_out_url = (
        cookie_session.get("account_service_log_out_url", url_for(".get_signed_out"))
        if cookie_session
        else get_survey_config().base_url
    )
```

Is that behaviour tested elsewhere?

tests/app/helpers/test_template_helpers.py
JamesGardiner: cdn_url is tested in test_app_create.py, though the other is a valid point, as I don't believe that is tested anywhere else. I'll add a test for this and for the contact_us_url as that also has some logic attached to it

JamesGardiner: Done

tests/integration/session/test_sign_out_and_exit.py
JamesGardiner: I've added tests to test_session_sign_out.py

MebinAbraham (approved):
👍

tests/app/helpers/test_template_helpers.py
ajmaddaford: The testing in test_app_create tests that the `CDN_URL` environment variable is added to the CSP headers i.e. outside of templating. The CDN url that we put in the context here should be also be tested (this also happens to concatenate the cdn url with the assets path).

tests/app/helpers/test_template_helpers.py
```diff
@@ -64,114 +161,299 @@ def test_get_footer_context_nisra_theme(app):
...
+        ("business", "en", "main"),
+        ("health", "en", "main"),
+        ("social", "en", "main"),
+        ("northernireland", "en", "main"),
+        ("census", "en", "census"),
+        ("census", "cy", "census"),
+        ("census-nisra", "en", "census"),
+    ],
+)
+def test_correct_theme_in_config(app: Flask, theme: str, language: str, expected: str):
```
ajmaddaford, line 379: This is checking the theme in the context rather than the config.
```suggestion
def test_correct_theme_in_context(app: Flask, theme: str, language: str, expected: str):
```

tests/app/helpers/test_template_helpers.py
```diff
@@ -64,114 +161,299 @@ def test_get_footer_context_nisra_theme(app):
...
+        ("business", "en", None),
+        ("health", "en", None),
+        ("social", "en", None),
+        ("northernireland", "en", None),
+        ("census", "en", "Census 2021"),
+        ("census", "cy", "Census 2021"),
+        ("census-nisra", "en", "Census 2021"),
+    ],
+)
+def test_correct_survey_title_in_config(
```
ajmaddaford, line 404:
```suggestion
def test_correct_survey_title_in_context(
```

tests/app/helpers/test_template_helpers.py
```diff
@@ -64,114 +161,299 @@ def test_get_footer_context_nisra_theme(app):
...
+        ("business", "en", []),
+        ("health", "en", []),
+        ("social", "en", []),
+        ("northernireland", "en", []),
+        ("census", "en", [{"nisra": False}]),
+        ("census", "cy", [{"nisra": False}]),
+        ("census-nisra", "en", [{"nisra": True}]),
+    ],
+)
+def test_correct_data_layer_in_config(
```
ajmaddaford, line 431:
```suggestion
def test_correct_data_layer_in_context(
```

tests/app/helpers/test_template_helpers.py
```diff
@@ -1,12 +1,109 @@
...
+        survey_config = CensusSurveyConfig()
+
+        result = ContextHelper(
+            language="en",
+            is_post_submission=True,
+            include_csrf_token=True,
+            survey_config=survey_config,
+        ).context["footer"]["footerWarning"]
+
+    assert result == expected
```
ajmaddaford, line 103: The opposite test should also be done - the warning should not be present if `is_post_submission` is `False`.

tests/app/helpers/test_template_helpers.py
```diff
@@ -64,114 +161,299 @@ def test_get_footer_context_nisra_theme(app):
...
 
-def test_get_census_base_url_ga():
-    expected = "https://census.gov.uk/"
-    result = get_census_base_url(schema_theme="census", language_code="ga")
+        result = ContextHelper(
+            language="en",
+            is_post_submission=False,
+            include_csrf_token=False,
+            survey_config=survey_config,
+        ).context["include_csrf_token"]
```
ajmaddaford, line 457: We should really test `include_csrf_token` being `True` here as well.

JamesGardiner: Yep - I thought it was tested above, good spot

tests/app/helpers/test_template_helpers.py
JamesGardiner: It's not tested explicitly, but it is tested in test_footer_context_census_theme indirectly, e.g. the comparison would fail if footerWarning was present. Obviously that isn't clear enough, so it would be better to have an additional test here for it

ajmaddaford (approved):

PR #606, opened 2021-05-17
Remove Summary and Confirmation blocks + Introduce questionnaire flow

MebinAbraham:
As per [design doc](https://github.com/ONSdigital/eq-questionnaire-validator/pull/90), the Summary and Confirmation block types have been removed. The questionnaire flow property has been introduced, which is used to determine the type whether the hub/summary/confirmation page is used.

**__Decisions__**
- Prior to this change, the presence of a previous link was determined by whether the summary/confirmation block was part of the last section vs in its own section. The link was not shown if summary/confirmation was in its own section. With this change, a previous link is always shown on the submit page. This links to the last completed location in the questionnaire instead of the last block in the schema (existing behaviour).
- To keep the journey consistent, visiting the submit page before completing the questionnaire will redirect you to the first incomplete location instead of the initial proposal of a 404. Redirect to first incomplete is consistent for hub route and section route, and it seems to be the most appropriate choice in terms of UX since people are likely to bookmark summary pages. Visiting the `/submit` endpoint for a questionnaire that is non-Linear, a `404` is returned.
- The Confirmation html template inherits the hub template to reduce duplication as currently the only difference between the two visually is that the hub has rows (spokes).
- The mechanism of routing to the end of a questionnaire is highlighted in on the [validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/95).

**__Out of scope__**
- Review the use of handlers. Our handlers are becoming more and more diverged. We should review how we use and share the handlers and how we could make block and non-block handlers consistent.
- Introduce a handler for the Hub.
- Handle submission of the Hub via `/submit` and rationalising the questionnaire flow classes.

### How to review 
Ensure:
- existing features such as the Hub/Summary/Confirmation still work as expected.
- the `/submit` endpoint is used to display the final submit page. 
- you can route the final Summary/Confirmation.
- no features we regressed?

### Checklist


test_schemas/en/test_skipping_to_questionnaire_end_multiple_sections.json
```diff
@@ -0,0 +1,150 @@
+{
```
LJBabbage, line 1: Isn't this routing to end rather than skipping? I couldn't see _skip_conditions_

test_schemas/en/test_skipping_to_questionnaire_end_single_section.json
```diff
@@ -0,0 +1,133 @@
+{
```
LJBabbage, line 1: Same as above

LJBabbage (changes requested):
Only a first pass as the PR is quite large. It works as expected. only a few minor comments at present

app/views/contexts/submit_context.py
```diff
@@ -25,21 +29,30 @@ def __call__(self, collapsible=True, answers_are_editable=True):
...
-                "collapsible": collapsible,
-                "summary_type": "Summary",
-            },
             "title": title,
             "guidance": guidance,
             "warning": warning,
             "submit_button": submit_button,
         }
+        if include_summary:
+            context.update(self._get_summary_context(collapsible, answers_are_editable))
```
ajmaddaford, line 38: If `_get_summary_context` returned a dict without the `summary` key, this could be written as:
```suggestion
            context["summary"] = self._get_summary_context(collapsible, answers_are_editable)
```

app/questionnaire/router.py
```diff
@@ -305,10 +278,16 @@ def get_next_block_url(location, routing_path):
...
 
-    def get_last_location_in_survey(self):
-        last_section_id = self._schema.get_section_ids()[-1]
-        last_block_id = self._schema.get_last_block_id_for_section(last_section_id)
-        return Location(section_id=last_section_id, block_id=last_block_id)
+    def get_last_location_in_questionnaire(self) -> Location:
+        section_id, list_item_id = self._get_last_complete_section_key()
+        last_complete_block = self._progress_store.get_completed_block_ids(
+            section_id=section_id, list_item_id=list_item_id
+        )[-1]
```
ajmaddaford, line 285: I don't think this will always work as a block can be completed, but not on the routing path.

templates/confirmation.html
```diff
@@ -1,39 +1 @@
...
-          "attributes": {
-            "data-qa": "btn-submit",
-            "data-loading-msg": _("Submitting") + "&hellip;"
-          }
-        })
-      }}
-    </p>
-  {% endif %}
-{% endblock %}
+{% extends 'hub.html' %}
```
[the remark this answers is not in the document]
ajmaddaford: I agree, maybe there should be a common template they both use.

templates/summary.html
```diff
@@ -1,11 +1,45 @@
...
+          "attributes": {
+            "data-qa": "btn-submit",
+            "data-loading-msg": _("Submitting") + "&hellip;"
+          }
+        })
+      }}
+    </p>
+  {% endif %}
+{% endblock %}
+
```
ajmaddaford, line 42: I think the summary should be separated from the calculated summary template. We should consider the commonality between the hub/confirmation/submit when doing this.

app/routes/questionnaire.py
```diff
@@ -104,34 +111,68 @@ def get_questionnaire(schema, questionnaire_store):
...
+        content=context,
+        page_title=context["title"],
     )
 
-    return render_template("hub", content=hub_context, page_title=hub_context["title"])
+
+@questionnaire_blueprint.route("submit/", methods=["GET", "POST"])
+@with_questionnaire_store
+@with_schema
+def submit(
```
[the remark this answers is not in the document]
ajmaddaford: I agree. The same applies to the context naming, so `SubmitQuestionnaireContext` rather than `SubmitContext`.

app/questionnaire/questionnaire_schema.py
```diff
@@ -161,24 +161,30 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
 
         return answers_by_id
 
-    def get_hub(self) -> Optional[ImmutableDict]:
-        hub_schema: Optional[ImmutableDict] = self.json.get("hub")
-        return hub_schema
+    @cached_property
+    def _questionnaire_flow(self) -> ImmutableDict[str, Any]:
```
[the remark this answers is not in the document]
MebinAbraham: Not possible for questionnaire schemas as we load in arbitrary JSON. That's why atm most return Any or an untyped `ImmutableDict` which defaults to `ImmutableDict[Any, Any]`. It came up when Andrew's initial type hint PR went in. In future, we could potentially load the expected types by making using of the schemas in validator which says what type things should be.

app/questionnaire/questionnaire_schema.py
```diff
@@ -161,24 +161,30 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
...
+    def questionnaire_flow_options(self) -> ImmutableDict[str, Any]:
+        options: ImmutableDict[str, Any] = self._questionnaire_flow["options"]
+        return options
 
-    def is_hub_enabled(self) -> Optional[bool]:
-        if hub := self.get_hub():
-            return hub.get("enabled")
+    @cached_property
+    def is_questionnaire_flow_hub(self) -> bool:
+        return bool(self._questionnaire_flow["type"] == "Hub")
```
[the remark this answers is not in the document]
MebinAbraham: I think referencing `hub` is fine since that is the `type`. We could also have more than 2 types in future. I don't think calling it `non-linear` is clear, and I can't think of anything else that is better.

app/routes/questionnaire.py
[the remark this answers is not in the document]
MebinAbraham: Chose it to be consistent with other routes since it is under the questionnaire blueprint. More than happy to change if make it clearer.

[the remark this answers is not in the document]
MebinAbraham: Have also updated the class names to make it clearer.

app/questionnaire/router.py
```diff
@@ -142,7 +132,16 @@ def get_first_incomplete_location_in_survey_url(self):
             )
             return self.get_section_resume_url(section_routing_path)
 
-        return self.get_last_location_in_survey().url()
+        return self.get_next_location_url_for_questionnaire_flow()
+
+    def get_next_location_url_for_questionnaire_flow(self) -> str:
+        if self._schema.is_questionnaire_flow_hub and self.can_access_hub():
+            return url_for("questionnaire.get_questionnaire")
```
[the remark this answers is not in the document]
MebinAbraham: Correct, but it is needed on the endpoint as a user can explicitly visit it. It is needed here to prevent an unnecessary redirect which would go to the hub then redirect to the first incomplete location.

mypy.ini
```diff
@@ -1,6 +1,7 @@
 # Global options:
 
 [mypy]
+ignore_missing_imports = True
```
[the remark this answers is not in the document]
MebinAbraham: Ya, it is existing and this is something we can't prevent right now anyway as this is used to disable errors because of untyped libraries we use.

templates/confirmation.html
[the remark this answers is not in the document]
MebinAbraham: I agree, I just didn't know what to call the common template. Any suggestions?

test_schemas/en/test_skipping_to_questionnaire_end_multiple_sections.json
MebinAbraham: Schemas for skip conditions are prefixed with `skip_conditions`. I used `skipping` as that is what the existing schema was called for `skipping` to section end. I started renaming this to routing, but then it didn't sound right. It is about when we `skip`/`jump` to a questionnaire end that just route to it even if the underlying use is `routing rules`. 

Open to other suggestions.

app/routes/questionnaire.py
```diff
@@ -104,34 +111,68 @@ def get_questionnaire(schema, questionnaire_store):
...
+        )
+
+    if request.method == "POST":
+        submit_handler.handle_post()
+        return redirect(url_for("post_submission.get_thank_you"))
+
+    context = submit_handler.get_context()
+    return render_template(
+        submit_handler.template,
+        content=context,
```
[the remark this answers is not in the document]
MebinAbraham: This wasn't originally done because the `page_title` comes from the context object and I wanted to avoid making the call twice.
```py
page_title=context["title"]
```

app/views/contexts/submit_context.py
MebinAbraham: Done

app/questionnaire/router.py
```diff
@@ -142,7 +132,16 @@ def get_first_incomplete_location_in_survey_url(self):
             )
             return self.get_section_resume_url(section_routing_path)
 
-        return self.get_last_location_in_survey().url()
+        return self.get_next_location_url_for_questionnaire_flow()
+
+    def get_next_location_url_for_questionnaire_flow(self) -> str:
```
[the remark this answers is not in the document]
MebinAbraham: Ya fair enough, happy with the suggestion so committed.

app/questionnaire/router.py
MebinAbraham: 🙈 Good spot. Will refactor.

MebinAbraham: This now returns the last block on the path by generating a routing path.

app/views/handlers/submit.py
```diff
@@ -0,0 +1,63 @@
...
+        self,
+        schema: QuestionnaireSchema,
+        questionnaire_store: QuestionnaireStore,
+        language: str,
+    ):
+        self._schema = schema
+        self._questionnaire_store = questionnaire_store
+        self._language = language
+
+        if not self._schema.is_questionnaire_flow_linear:
```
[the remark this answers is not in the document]
MebinAbraham: Done

templates/summary.html
MebinAbraham: I think a small discussion is worthwhile before I dive into refactoring this.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
LJBabbage: It would be nice for them to both be adjectives, but it was more a observation rather than a change request

templates/confirmation.html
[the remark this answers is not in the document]
MebinAbraham: I have done some rationalisation of the templates. `Hub`, `Confirmation` and `Final summary` as well as `Calculated Summary` all now use a common template. If we want to extract Calculated summary out by accepting some duplication, we could do that but with the changes I've done, it felt appropriate to share the same common template. Let me know your thoughts.

The submit button is something which could be rationalised a bit more between these templates.

test_schemas/en/test_skipping_to_questionnaire_end_multiple_sections.json
MebinAbraham: I think it is clearer to name it after the behaviour than the technical details, the tests etc just didn't read quite right to me when it said routing, maybe that's just me. I'm not too fussed so have updated all references to `routing` instead of `skipping` for this and the existing schema.

templates/summary.html
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/606#discussion_r635458345

tests/app/questionnaire/test_router.py
```diff
@@ -206,17 +227,50 @@ def test_return_to_final_summary_next_location_url(self):
             schema, self.answer_store, self.list_store, progress_store, self.metadata
         )
         current_location = Location(section_id="default-section", block_id="name-block")
+        routing_path = RoutingPath(["name-block"], section_id="default-section")
+        next_location = router.get_next_location_url(
+            current_location, routing_path, return_to="final-summary"
+        )
+
+        self.assertEqual(url_for("questionnaire.submit_questionnaire"), next_location)
```
ajmaddaford, line 235: I think a different schema with more than one block should be used for this test. Currently the `next_location` is `questionnaire.submit_questionnaire` with or without the `return_to` being set.

tests/app/questionnaire/test_router.py
```diff
@@ -189,7 +186,31 @@ def test_return_to_summary_next_location_url(self):
             "/questionnaire/sections/property-details-section/", next_location
         )
 
-    def test_return_to_final_summary_next_location_url(self):
+    def test_linear_questionnaire_get_next_location_url_routes_to_submit_page_when_questionnaire_completed(
```
ajmaddaford, line 189: I can't see what this is testing that is different to `test_last_block_no_section_summary_next_location_url_is_submit_page`?

tests/functional/base_pages/summary.page.js
```diff
@@ -1,13 +1,17 @@
 import QuestionPage from "./question.page";
 
-class ThankYouPage extends QuestionPage {
+class SummmaryPage extends QuestionPage {
```
ajmaddaford, line 3: Is this used?

tests/functional/generate_pages.py
```diff
@@ -409,12 +409,47 @@ def process_calculated_summary(answers, page_spec):
...
 
-def process_final_summary(schema_data, page_spec, collapsible, section_summary=False):
-    for section in schema_data["sections"]:
-        write_summary_spec(collapsible, page_spec, section, section_summary)
+def process_final_summary(
+    schema_data, require_path, dir_out, spec_file, collapsible, section_summary=False
+):
+    logger.debug("Processing questionnaire flow")
+
+    page_filename = f"summary.page.js"
```
ajmaddaford, line 417: I think we need to be clear on the naming. The page is the `submit` page, that can include a (final) summary.

tests/functional/generate_pages.py
```diff
@@ -409,12 +409,47 @@ def process_calculated_summary(answers, page_spec):
...
+    page_path = os.path.join(dir_out, page_filename)
+
+    logger.info("creating %s...", page_path)
 
-    if collapsible:
-        page_spec.write(COLLAPSIBLE_SUMMARY_GETTER)
+    with open(page_path, "w") as page_spec:
+        page_name = "FinalSummary"
+        base_page = "QuestionPage"
+        base_page_file = "question.page"
```
ajmaddaford, line 425: Inheriting from QuestionPage seems wrong - it has a lot of methods that aren't relevant.

ajmaddaford:
I think there are gaps in the test_router tests, and the mix of long function names makes it difficult to see where these are. We need to think about how we can organise these tests more clearly, possibly splitting the next and previous location tests into their own class. Within that we need to think about how we make clear the hub vs linear questionnaire tests.

Some of the missing tests I think we have:
- Linear
  - First block in the questionnaire, previous location should be `None`
  - First block in a second section, previous location should be either be:
    - the last block in the previous section
    - the section summary for the previous section
  - next location when returning from a summary and the routing path changes

app/routes/questionnaire.py
```diff
@@ -1,23 +1,30 @@
+from __future__ import annotations
+
+from typing import Union
+
 import flask_babel
 from flask import Blueprint, g, redirect, request, url_for
 from flask_login import current_user, login_required
 from itsdangerous import BadSignature
 from structlog import get_logger
+from werkzeug import Response
```
JamesGardiner, line 10:
```suggestion
from flask import Response
```

app/views/contexts/__init__.py
```diff
@@ -2,13 +2,13 @@
...
 from .list_context import ListContext
-from .questionnaire_summary_context import QuestionnaireSummaryContext
 from .section_summary_context import SectionSummaryContext
+from .submit_questionnaire_context import SubmitQuestionnaireContext
 
 __all__ = [
     "CalculatedSummaryContext",
     "Context",
-    "QuestionnaireSummaryContext",
+    "SubmitQuestionnaireContext",
```
JamesGardiner, line 11: This be more consistent if it was named `QuestionnaireSubmissionContext`

app/questionnaire/questionnaire_schema.py
```diff
@@ -161,24 +161,33 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
 
         return answers_by_id
 
-    def get_hub(self) -> Optional[ImmutableDict]:
-        hub_schema: Optional[ImmutableDict] = self.json.get("hub")
-        return hub_schema
+    @cached_property
+    def _questionnaire_flow(self) -> ImmutableDict[str, Any]:
+        questionnaire_flow: ImmutableDict = self.json["questionnaire_flow"]
+        return questionnaire_flow
```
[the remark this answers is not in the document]
ajmaddaford: `self.json["questionnaire_flow"]` has type `Any`, which doesn't match the return type expected.

app/views/contexts/__init__.py
ajmaddaford: We use Submission to mean the payload that is submitted downstream, this is the context for the Submit page.

app/questionnaire/questionnaire_schema.py
```diff
@@ -161,24 +161,33 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
...
-    def is_hub_enabled(self) -> Optional[bool]:
-        if hub := self.get_hub():
-            return hub.get("enabled")
+    @cached_property
+    def is_questionnaire_flow_hub(self) -> bool:
+        return bool(self._questionnaire_flow["type"] == "Hub")
+
+    @cached_property
+    def is_questionnaire_flow_linear(self) -> bool:
+        return bool(self._questionnaire_flow["type"] == "Linear")
```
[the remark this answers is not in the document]
MebinAbraham: It's because typing can't infer this returns a `bool` because `self._questionnaire_flow["type"]` has type `Any`.

MebinAbraham:
WIP:
- refactoring all router tests + adding some missing tests
- template naming changes for readability
- `Confirmation` will be treated as `Submit` and Summary as `Submit with Summary`
- some further clean up of generate pages.

app/views/handlers/submit_questionnaire.py
```diff
@@ -0,0 +1,63 @@
...
+class SubmitQuestionnaireHandler:
+    def __init__(
+        self,
+        schema: QuestionnaireSchema,
+        questionnaire_store: QuestionnaireStore,
+        language: str,
+    ):
+        if not schema.is_questionnaire_flow_linear:
+            raise InvalidLocationException(
+                "Submit page not enabled or questionnaire is not complete"
```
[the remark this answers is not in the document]
MebinAbraham: Updated.

test_schemas/en/test_skipping_to_questionnaire_end_multiple_sections.json
```diff
@@ -0,0 +1,150 @@
...
+            "type": "string"
+        }
+    ],
+    "questionnaire_flow": {
+        "type": "Linear",
+        "options": { "include_summary": true }
+    },
+    "sections": [
+        {
+            "id": "test-skipping-section",
```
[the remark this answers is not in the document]
MebinAbraham: Removed. Also, the schema now references `routing` instead of `skipping.

tests/app/questionnaire/test_router.py
```diff
@@ -341,20 +387,49 @@ def test_previous_location_url(self):
...
             current_location, routing_path
         )
         expected_location_url = Location(
-            section_id="default-section", block_id="name-block"
+            section_id="default-section", block_id="mandatory-checkbox"
         ).url()
 
         self.assertEqual(previous_location_url, expected_location_url)
 
+    def test_previous_location_on_first_block_with_hub_enabled(self):
```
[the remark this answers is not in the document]
MebinAbraham: Removed the old one.

tests/app/questionnaire/test_router.py
MebinAbraham: Now uses a schema with multiple blocks.

tests/app/questionnaire/test_router.py
MebinAbraham: Have been refactored.

tests/functional/base_pages/confirmation.page.js
```diff
@@ -0,0 +1,37 @@
+import BasePage from "./base.page";
```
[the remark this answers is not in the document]
MebinAbraham: Updated. We now have a `SubmitPage` which handles summary also.

tests/functional/base_pages/summary.page.js
MebinAbraham: Was only for 1 test but has been removed as part of the base pages refactor.

tests/functional/generate_pages.py
MebinAbraham: Have updated.

tests/functional/generate_pages.py
MebinAbraham: Base pages have been slightly refactored and this now inherits from `SubmitBasePage`.

MebinAbraham:
> I think there are gaps in the test_router tests, and the mix of long function names makes it difficult to see where these are. We need to think about how we can organise these tests more clearly, possibly splitting the next and previous location tests into their own class. Within that we need to think about how we make clear the hub vs linear questionnaire tests.
> 
> Some of the missing tests I think we have:
> 
> * Linear
>   
>   * First block in the questionnaire, previous location should be `None`
>   * First block in a second section, previous location should be either be:
>     
>     * the last block in the previous section
>     * the section summary for the previous section
>   * next location when returning from a summary and the routing path changes

Have made some refactors to the router tests. There prob needs to be further clean up. Let me know your thoughts.

Have also added extra tests to cover more scenarios. I think there are more scenarios to potentially consider, but I think that should be separate, and we should look at refactoring router.

app/questionnaire/questionnaire_schema.py
```diff
@@ -161,24 +161,33 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
...
+    @cached_property
+    def questionnaire_flow_options(self) -> ImmutableDict[str, Any]:
+        options: ImmutableDict[str, Any] = self._questionnaire_flow["options"]
+        return options
 
-    def is_hub_enabled(self) -> Optional[bool]:
-        if hub := self.get_hub():
-            return hub.get("enabled")
+    @cached_property
+    def is_questionnaire_flow_hub(self) -> bool:
```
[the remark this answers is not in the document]
MebinAbraham: Updated in [`6ba688f` (#606)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/606/commits/6ba688f2348665e4d885073b68e6229d08602956)

app/questionnaire/questionnaire_schema.py
```diff
@@ -161,24 +161,33 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
 
         return answers_by_id
 
-    def get_hub(self) -> Optional[ImmutableDict]:
-        hub_schema: Optional[ImmutableDict] = self.json.get("hub")
-        return hub_schema
+    @cached_property
+    def _questionnaire_flow(self) -> ImmutableDict[str, Any]:
```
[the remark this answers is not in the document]
MebinAbraham: Updated in [`6ba688f` (#606)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/606/commits/6ba688f2348665e4d885073b68e6229d08602956)

app/questionnaire/questionnaire_schema.py
```diff
@@ -161,24 +161,33 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
...
 
-    def is_hub_enabled(self) -> Optional[bool]:
-        if hub := self.get_hub():
-            return hub.get("enabled")
+    @cached_property
+    def is_questionnaire_flow_hub(self) -> bool:
+        return bool(self._questionnaire_flow["type"] == "Hub")
+
+    @cached_property
+    def is_questionnaire_flow_linear(self) -> bool:
```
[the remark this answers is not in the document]
MebinAbraham: Updated in [`6ba688f` (#606)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/606/commits/6ba688f2348665e4d885073b68e6229d08602956)

app/views/handlers/submit_questionnaire.py
```diff
@@ -0,0 +1,62 @@
...
+    def router(self) -> Router:
+        return Router(
+            schema=self._schema,
+            answer_store=self._questionnaire_store.answer_store,
+            list_store=self._questionnaire_store.list_store,
+            progress_store=self._questionnaire_store.progress_store,
+            metadata=self._questionnaire_store.metadata,
+        )
+
+    def get_context(self) -> dict[str, Union[str, dict]]:
```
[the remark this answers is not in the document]
MebinAbraham: I kept it as `get_context` to be consistent with the other handlers and didn't want to change it all in this PR.

app/routes/questionnaire.py
MebinAbraham: Using `redirect` returns `Response` from `werkzeug`. Mypy complains when using the flask class.

test_schemas/en/test_skipping_to_questionnaire_end_multiple_sections.json
```diff
@@ -0,0 +1,150 @@
+{
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "137",
+    "theme": "default",
+    "title": "Test Skipping To Section End",
```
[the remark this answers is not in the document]
MebinAbraham: [`7c380bd` (#606)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/606/commits/7c380bd6c4ebf435ef8112d9716b3954e670deb0)

test_schemas/en/test_skipping_to_questionnaire_end_single_section.json
```diff
@@ -0,0 +1,133 @@
...
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "137",
+    "theme": "default",
+    "title": "Test Skipping To Section End",
```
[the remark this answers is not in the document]
MebinAbraham: [`7c380bd` (#606)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/606/commits/7c380bd6c4ebf435ef8112d9716b3954e670deb0)

tests/app/questionnaire/test_router.py
```diff
@@ -719,3 +793,46 @@ def test_full_routing_path_with_repeating_sections(self):
...
+                {"answer_id": "test-routing-answer", "value": "No"},
+                {
+                    "answer_id": "test-routing-optional-answer",
+                    "value": "I am a completionist",
+                },
+            ]
+        )
+        section_id = "test-routing-section"
+        last_block_on_path = "test-routing-forced"
+        last_completed_block = "test-routing-optional"
```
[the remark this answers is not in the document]
MebinAbraham: [`d456ec2` (#606)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/606/commits/d456ec2d952fa0593c1db41d86a97e1dc7879b88)

app/views/contexts/__init__.py
JamesGardiner: The issue is that it reads like a verb which isn't ideal

templates/submit-with-summary.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'layouts/_submit.html' %}
```
LJBabbage, line 1: This is slightly inconsistent when you look at calculatedsummary and hub which also extends submit. Following that pattern this should just be summary.

test_schemas/en/test_routing_to_section_end.json
```diff
@@ -23,21 +23,29 @@
...
+        "options": {
+            "summary": {
+                "collapsible": false
+            }
+        }
+    },
     "sections": [
         {
-            "id": "test-skipping-section",
+            "id": "test-section",
```
LJBabbage, line 36: Probably doesn't need test prefix for all ids

LJBabbage (approved):
This looks good to me. There is a few odd tweaks which could be made, but they are minor and shouldn't stop the merging of this PR

tests/app/views/contexts/test_submit_context.py
```diff
@@ -0,0 +1,165 @@
...
+                self.assertTrue("title" in block["question"])
+                self.assertTrue("answers" in block["question"])
+                for answer in block["question"]["answers"]:
+                    self.assertTrue("id" in answer)
+                    self.assertTrue("value" in answer)
+                    self.assertTrue("type" in answer)
+
+    @patch("app.jinja_filters.flask_babel.get_locale", Mock(return_value="en_GB"))
+    def test_submit_context_with_default_submission_content(self):
+        schemas = ["test_summary", "test_instructions"]
```
ajmaddaford, line 59: I don't think we need to test this with two different schemas.

tests/app/views/contexts/test_submit_context.py
```diff
@@ -0,0 +1,165 @@
...
+
+from app.data_models.answer_store import AnswerStore
+from app.data_models.list_store import ListStore
+from app.data_models.progress_store import ProgressStore
+from app.utilities.schema import load_schema_from_name
+from app.views.contexts import SubmitQuestionnaireContext
+from tests.app.app_context_test_case import AppContextTestCase
+
+
+class TestStandardSummaryContext(AppContextTestCase):
```
ajmaddaford, line 11: I don't think the split of classes and tests is right here. The `SubmitQuestionnaireContext` class sets the submission guidance regardless of whether a summary is included or not. I think we should have a `TestSubmitContext` class that tests for what is common, and a `TestSubmitContextWithSummary` class to test the summary.

test_schemas/en/test_summary_with_submission_text.json
```diff
@@ -21,6 +21,14 @@
             "type": "string"
         }
     ],
+    "questionnaire_flow": {
```
ajmaddaford, line 24: This file should be renamed to make it clear it's testing submit e.g. `test_submit_with_summary_custom_submission_text.json`

test_schemas/en/test_summary.json
```diff
@@ -21,6 +21,14 @@
             "type": "string"
         }
     ],
+    "questionnaire_flow": {
```
ajmaddaford, line 24: This schema should be named `test_submit_with_summary.json`

test_schemas/en/test_submit_page.json
```diff
@@ -4,17 +4,12 @@
     "schema_version": "0.0.1",
```
ajmaddaford, line 4: The file should be named `test_submit_with_custom_submission_text.json` to be consistent with the equivalent schema that has a summary.

tests/functional/spec/features/submit_page/submit_page_collapsible_summary.spec.js
```diff
@@ -22,9 +22,9 @@ describe("Collapsible Summary", () => {
       $(HouseholdCountSectionSummaryPage.submit()).click();
     });
 
-    it("When I am on the Final Summary, Then the group titles should be displayed", () => {
-      expect($(QuestionnaireSummaryPage.collapsibleSummary()).getText()).to.contain("Property Details");
-      expect($(QuestionnaireSummaryPage.collapsibleSummary()).getText()).to.contain("House Details");
+    it("When I am on the submit page, Then the group titles should be displayed", () => {
+      expect($(SubmitPage.collapsibleSummary()).getText()).to.contain("Property Details");
+      expect($(SubmitPage.collapsibleSummary()).getText()).to.contain("House Details");
```
ajmaddaford, line 27: An existing issue with the test, but it's not clear how this is testing the "collapsible" bit.

tests/functional/spec/features/submit_page/submit_page_custom_submission_text.spec.js
```diff
@@ -0,0 +1,17 @@
+import DessertBlockPage from "../../../generated_pages/summary_with_submission_text/dessert-block.page.js";
+import SubmitPage from "../../../generated_pages/summary_with_submission_text/submit.page.js";
+
+describe("Summary Screen", () => {
+  beforeEach("Load the survey", () => {
+    browser.openQuestionnaire("test_summary_with_submission_text.json");
+  });
+
+  it("Given a survey with summary has been completed, when submission content has been set in the schema, then the correct content should be displayed", () => {
```
ajmaddaford, line 9:
```suggestion
  beforeEach("Load the questionnaire", () => {
    browser.openQuestionnaire("test_summary_with_submission_text.json");
  });

  it("Given a questionnaire with a summary and custom submission content has been completed, then the correct submission content should be displayed", () => {
```

tests/functional/spec/features/submit_page/submit_page_summary.spec.js
```diff
@@ -0,0 +1,100 @@
...
+import NumbersPage from "../../../generated_pages/summary/numbers.page.js";
+import RadioPage from "../../../generated_pages/summary/radio.page.js";
+import SubmitPage from "../../../generated_pages/summary/submit.page.js";
+
+describe("Submit Page with Summary", () => {
+  beforeEach("Load the survey", () => {
+    browser.openQuestionnaire("test_summary.json");
+  });
+
+  it("Given a survey with a summary has been completed when the submit page is displayed, then it should contain a summary of all answers", () => {
```
ajmaddaford, line 12: I realise this was copied from another spec file, but "survey" should be "questionnaire". Happy for a card to be raised to do this across everything.

tests/functional/spec/features/submit_page/submit_page_summary.spec.js
```diff
@@ -0,0 +1,100 @@
...
+  it("Given a number value of zero is entered when on the submit page then formatted 0 should be displayed on the summary", () => {
+    $(RadioPage.submit()).click();
+    $(DessertPage.answer()).setValue("Cake");
+    $(DessertPage.submit()).click();
+    $(DessertConfirmationPage.yes()).click();
+    $(DessertConfirmationPage.submit()).click();
+    $(NumbersPage.currency()).setValue("0");
+    $(NumbersPage.submit()).click();
+    expect($(SubmitPage.numbersCurrencyAnswer()).getText()).to.contain("£0.00");
+  });
```
ajmaddaford, line 67: This should really be a test for the currency answer type.

app/questionnaire/router.py
```diff
@@ -154,15 +153,10 @@ def get_section_resume_url(self, routing_path):
 
         return self.get_first_location_in_section(routing_path).url()
 
-    def is_survey_complete(self):
+    @cached_property
```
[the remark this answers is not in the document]
MebinAbraham: It seems fine for our current use, but I guess it could happen. I wanted to avoid scenarios in which it would evaluate more than once. Maybe we should look at passing around the value where appropriate.

app/questionnaire/router.py
```diff
@@ -305,10 +278,13 @@ def get_next_block_url(location, routing_path):
             list_item_id=routing_path.list_item_id,
         )
 
-    def get_last_location_in_survey(self):
-        last_section_id = self._schema.get_section_ids()[-1]
-        last_block_id = self._schema.get_last_block_id_for_section(last_section_id)
-        return Location(section_id=last_section_id, block_id=last_block_id)
+    def get_last_location_in_questionnaire(self) -> Location:
+        routing_path = self.routing_path(*self._get_last_complete_section_key())
```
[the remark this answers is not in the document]
MebinAbraham: I didn't want to call `_get_last_complete_section_key` in the handler hence why it was done in router. When working at the questionnaire level, we do generate a path in-place such as in `get_first_incomplete_location_in_questionnaire_url`. Therefore, I would say it is consistent with that. For now, I have reused the `get_last_location_in_section` within this method. Let me know your thoughts.

templates/calculatedsummary.html
```diff
@@ -1 +1,34 @@
-{% extends 'summary.html' %}
+{% extends 'layouts/_submit.html' %}
```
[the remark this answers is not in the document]
MebinAbraham: Calculated summary has its own template now and doesn't share with submit. It does also extend from `_questionnaire.html`. It turned out to be cleaner.

templates/submit-with-summary.html
MebinAbraham: Calculated summary has its own template now and doesn't share with submit.

test_schemas/en/test_routing_to_section_end.json
MebinAbraham: I guess but don't really want to spend more time on ids. I would say we need a style guide and/or a linter. This is an issue with a **lot** of our schemas and I think it makes sense to address it all together and have a style guide.

templates/layouts/_submit.html
```diff
@@ -0,0 +1,47 @@
...
+  {% if content.guidance %}
+    <div class="u-mt-l">
+      <p class="u-mb-no" data-qa="guidance">{{ content.guidance }}</p>
+    </div>
+  {% endif %}
+
+  {% if content.individual_response_url %}
+    {% set title = _("If you can’t answer someone else’s questions") %}
+    {% include 'partials/individual-response-guidance.html' %}
+  {% endif %}
```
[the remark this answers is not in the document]
MebinAbraham: Due to the way I inherit `_submit.html`, I would still need an empty block in `_submit.html` to define this in `hub.html`. Thoughts?.

templates/submit-with-summary.html
```diff
@@ -0,0 +1,23 @@
...
+        "submitType": 'timer',
+        "classes": "btn--loader js-btn-submit",
+        "attributes": {
+          "data-qa": "btn-submit",
+          "data-loading-msg": _("Submitting") + "&hellip;"
+        }
+      })
+    }}
+  </p>
+{% endblock submit_button %}
```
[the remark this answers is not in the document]
MebinAbraham: This was wrong actually, it should have been different for `hub` vs `submit`. However, I now use the same config for all of them. I used the hub's definition since it took into account displaying a guidance above the button which now applies to submit pages

tests/app/questionnaire/test_router.py
```diff
@@ -49,66 +189,106 @@ def test_cant_access_location(self):
...
+        self.list_store = ListStore(
+            [{"items": ["abc123"], "name": "people", "primary_person": "abc123"}]
+        )
+
+        is_questionnaire_complete = self.router.is_questionnaire_complete
+
+        self.assertTrue(is_questionnaire_complete)
+
+
+class TestRouterLocationAccessibility(TestRouterBaseTestCase):
```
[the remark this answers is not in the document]
MebinAbraham: Went with `TestRouterLocationValidity` since that is what I called it first before changing it.

tests/app/questionnaire/test_router.py
```diff
@@ -119,17 +299,20 @@ def test_cant_access_location_not_on_allowable_path(self):
...
 
-        can_access_location = router.can_access_location(current_location, routing_path)
         self.assertFalse(can_access_location)
 
-    def test_next_location_url(self):
-        schema = load_schema_from_name("test_checkbox")
-        progress_store = ProgressStore(
+
+class TestRouterNextLocation(TestRouterBaseTestCase):
+    def test_section_in_progress(self):
```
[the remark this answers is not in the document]
MebinAbraham: I was trying to avoid the use of `next_location` etc in the test names when the class is named after a feature. However, I can see it can cause some ambiguities, however, I made some changes to the existing names which should be clearer. Let me know.

tests/app/questionnaire/test_router.py
```diff
@@ -171,59 +353,26 @@ def test_return_to_summary_next_location_url(self):
...
-            current_location, routing_path, return_to="final-summary"
-        )
-        expected_location = Location(
-            section_id="default-section", block_id="summary"
-        ).url()
-
-        self.assertEqual(next_location, expected_location)
-
-    def test_return_to_first_incomplete_location_when_last_block_in_section_in_progress(
+    def test_get_first_incomplete_location_url_section_in_progress(
```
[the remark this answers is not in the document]
MebinAbraham: I have renamed this. The purpose of this was specifically around backwards routing. So testing next location returns the first incomplete location in the section when the current block is the last block on the path, but the section is not complete due to backwards journey.

tests/app/questionnaire/test_router.py
```diff
@@ -278,294 +426,266 @@ def test_last_block_section_summary_on_completion_false_next_location_url(self):
...
 
-        router = Router(
-            schema,
-            self.answer_store,
-            self.list_store,
-            self.progress_store,
-            self.metadata,
-        )
+class TestRouterPreviousLocation(TestRouterBaseTestCase):
+    def test_url_on_question_page(self):
```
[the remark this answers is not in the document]
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/606#discussion_r640033204

tests/app/questionnaire/test_router.py
```diff
@@ -278,294 +426,266 @@ def test_last_block_section_summary_on_completion_false_next_location_url(self):
...
 
-        self.assertEqual(previous_location_url, expected_location_url)
+        self.assertEqual(expected_location_url, previous_location_url)
 
-    def test_is_path_complete(self):
-        schema = load_schema_from_name("test_textfield")
-        progress_store = ProgressStore(
+
+class TestRouterPreviousLocationLinearFlow(TestRouterBaseTestCase):
+    def test_no_url_on_first_block_single_section(self):
```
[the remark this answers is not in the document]
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/606#discussion_r640033204

tests/app/questionnaire/test_router.py
```diff
@@ -278,294 +426,266 @@ def test_last_block_section_summary_on_completion_false_next_location_url(self):
...
+        self.assertEqual(
+            url_for("questionnaire.get_questionnaire"), previous_location_url
+        )
 
-        self.assertTrue(is_survey_complete)
 
-    def test_get_first_incomplete_location_url_in_section(self):
-        schema = load_schema_from_name("test_section_summary")
+class TestRouterSectionResume(TestRouterBaseTestCase):
+    def test_section_complete_returns_url_for_first_incomplete_location(self):
```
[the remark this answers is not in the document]
MebinAbraham: This and the one below was named in correctly. I have corrected this now.
This should have been `test_section_in_progress_returns_url_for_first_incomplete_location`.

tests/app/questionnaire/test_router.py
```diff
@@ -576,146 +696,69 @@ def test_get_first_incomplete_location_url_in_section(self):
...
-            self.list_store,
-            self.progress_store,
-            self.metadata,
+        self.schema = load_schema_from_name("test_hub_complete_sections")
+        self.progress_store = ProgressStore(
+            [
+                {
+                    "section_id": "employment-section",
+                    "block_ids": ["employment-status", "employment-type"],
+                    "status": CompletionStatus.COMPLETED,
```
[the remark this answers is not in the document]
MebinAbraham: This was incorrectly named, it should have been `test_section_complete_returns_url_for_first_location`. Has been updated.

tests/app/views/contexts/test_submit_context.py
MebinAbraham: It was to test when a summary was used and when a summary wasn't but as you pointed out, since that is done regardless of whether a summary exists or not, I have removed one redundant schema.

tests/app/views/contexts/test_submit_context.py
MebinAbraham: Have cleaned up the classes to with clear separation of Submit with and without summary.

tests/functional/spec/features/submit_page/submit_page_collapsible_summary.spec.js
MebinAbraham: I guess it was because `collapsibleSummary` didn't fail. But ya it's not great. While I am here, I made a slight extension to these tests to be a bit more meaningful. I haven't done much as I didn't want to go on a tangent.

tests/functional/spec/features/submit_page/submit_page_summary.spec.js
MebinAbraham: Ya I think we need to clean up as a whole as there is a lot of references to `survey` throughout the code base. But I have changed the ones in these spec files.

tests/functional/spec/features/submit_page/submit_page_summary.spec.js
MebinAbraham: I have left it as is for now. A currency spec didn't exist and didn't want to introduce a new currency spec with just this.

app/questionnaire/router.py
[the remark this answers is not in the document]
ajmaddaford: Fair enough. I think the `get_first_incomplete_location_in_questionnaire_url`, `is_questionnaire_complete` and `get_last_location_in_questionnaire` methods should be grouped together in that class. Also the method should return a url (and be renamed `get_last_location_in_questionnaire_url`) to be consistent with the other methods in the class.

templates/layouts/_submit.html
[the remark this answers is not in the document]
ajmaddaford: I think an empty block in `_submit.html` as a placeholder for the individual response content is fine. It should be named something generic like `pre_submit_button_content`.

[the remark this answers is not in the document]
MebinAbraham: Done

app/questionnaire/router.py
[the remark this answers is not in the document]
MebinAbraham: 👍 Have grouped together, but I moved it to the top which might look a bit out of place, however I think `@propery` should be grouped together.

tests/app/questionnaire/test_router.py
```diff
@@ -278,444 +427,314 @@ def test_last_block_section_summary_on_completion_false_next_location_url(self):
...
-                    "items": ["abc123", "123abc"],
-                    "name": "people",
-                    "primary_person": "abc123",
                 }
             ]
         )
-
-        router = Router(
-            schema, self.answer_store, list_store, progress_store, self.metadata
+        previous_location_url = self.router.get_last_location_in_questionnaire_url()
```
ajmaddaford, line 634:
```suggestion
        last_location_url = self.router.get_last_location_in_questionnaire_url()
```

ajmaddaford (commented):
Nice one! Changes look good 👍

ajmaddaford (approved):

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=introduce-questionnaire-flow
```
MebinAbraham, line 2: Update before merge.
```suggestion
tag=v2.0.0
```

JamesGardiner (approved):

PR #607, opened 2021-05-19
Remove questionnaire id

LJBabbage:
Questionnaire ID is census specific and can be remove

### How to review 
Check everything works as before and that questionnaire id has been removed from everywhere

### Checklist

### DIFF
--- a/app/routes/session.py
+++ b/app/routes/session.py
@@ -75,13 +75,11 @@ def login():
     schema_name = claims["schema_name"]
     tx_id = claims["tx_id"]
     ru_ref = claims["ru_ref"]
-    questionnaire_id = claims["questionnaire_id"]
 
     logger.bind(
         schema_name=schema_name,
         tx_id=tx_id,
         ru_ref=ru_ref,
-        questionnaire_id=questionnaire_id,
     )
     logger.info("decrypted token and parsed metadata")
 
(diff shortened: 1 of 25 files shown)


berroar (approved):
All appears to be working as expected

app/routes/session.py
```diff
@@ -75,13 +75,11 @@ def login():
     schema_name = claims["schema_name"]
     tx_id = claims["tx_id"]
     ru_ref = claims["ru_ref"]
-    questionnaire_id = claims["questionnaire_id"]
 
     logger.bind(
         schema_name=schema_name,
         tx_id=tx_id,
         ru_ref=ru_ref,
-        questionnaire_id=questionnaire_id,
```
MebinAbraham, line 84: From card:
```
ensure `case_id` remains log bound and used as a first class business attribute
```
Currently it isn't bound here.

MebinAbraham (commented):
Working as expected, just that comment about `case_id`.

app/routes/session.py
LJBabbage: I reckon we need a new story for that as there are other questions that need answering. Is case_id now mandatory being the main one, because using the term first class business attribute feels like it should be. If that is the case we naturally need to re-examine how we use case_id across not only this repo, but launcher and benchmark

MebinAbraham: Working updating the card.

MebinAbraham (approved):

PR #608, opened 2021-05-24
Make case_id mandatory

LJBabbage:
For census case_id was made optional, this PR reverts it back to mandatory for business surveys

### How to review 
Make sure everything still works, that case_id is now mandatory and tests have been changed. Also check that condition statements and defaults involving case_id have all been updated and nothing has been missed

### Checklist

### DIFF
--- a/app/authentication/authenticator.py
+++ b/app/authentication/authenticator.py
@@ -121,7 +121,7 @@ def _create_session_data_from_metadata(metadata):
         ru_name=metadata.get("ru_name"),
         ru_ref=metadata.get("ru_ref"),
         response_id=metadata.get("response_id"),
-        case_id=metadata.get("case_id"),
+        case_id=metadata["case_id"],
         case_ref=metadata.get("case_ref"),
         trad_as=metadata.get("trad_as"),
         account_service_url=metadata.get("account_service_url"),
--- a/app/data_models/session_data.py
+++ b/app/data_models/session_data.py
@@ -10,7 +10,7 @@ class SessionData:
         ru_name,
         ru_ref,
         response_id,
-        case_id=None,
+        case_id,
         case_ref=None,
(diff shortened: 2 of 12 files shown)


MebinAbraham (commented):
Can now be bound to the logger in `session.login`.

LJBabbage:
> Can now be bound to the logger in `session.login`.

Completely forgot about that. Another card :laughing ?

MebinAbraham (commented):
Case ID can be accessed directly instead of calling `.get("case_id")` since we know it is mandatory. It removes the ambiguity of whether it might not exist at some point in the request journey.

LJBabbage:
> Case ID can be accessed directly instead of calling `.get("case_id")` since we know it is mandatory. It removes the ambiguity of whether it might not exist at some point in the request journey.

We oddly do that for a number of mandatory fields (submission, flush and authenticator) which we might want to do the same.

JamesGardiner:
> > Case ID can be accessed directly instead of calling `.get("case_id")` since we know it is mandatory. It removes the ambiguity of whether it might not exist at some point in the request journey.
> 
> We oddly do that for a number of mandatory fields (submission, flush and authenticator) which we might want to do the same.

Also, there is a mixture of using `.get` and `[]` notation throughout Runner. If we are going to change it we should make it consistent everywhere, though that should potentially be a different PR

JamesGardiner (approved):
Changes look good - I'd like to update all the instances of where `case_id` is accessed using `.get` as it implies that it's optional, when it's not. This may need an update to our runner style guide to some degree, to say that optional keys should be accessed using `.get` whilst mandatory keys should be  square bracket notation, but happy to move that to another PR if that is the group consensus?

MebinAbraham (approved):

PR #609, opened 2021-05-28
Fix calculated summary rows

petechd:
Fixes issue with calculated summaries that was reported by Author team. [Link to trello](https://trello.com/c/C0fNatq8). Calculated summary question nesting had to be changed in order for this to be displayed only once on summary page.

### How to review 
Check one of the test schemas that covers this (test_calculated_summary.json).


petechd:
[Investigate Calculated Summaries](https://trello.com/c/C0fNatq8/4727-investigate-calculated-summaries)

tests/app/test_jinja_filters.py
```diff
@@ -442,3 +443,60 @@ def test_other_config_visibility(
 
     other = OtherConfig(Mock(), answer_schema_textfield)
     assert other.open is expected_visibility
+
+
+@patch("app.jinja_filters.flask_babel.get_locale", Mock(return_value="en_GB"))
+def test_single_calculated_summary_question():
```
LJBabbage, line 449: I am not sure what the single in the name refers to, is it needed?

tests/app/test_jinja_filters.py
```diff
@@ -442,3 +443,60 @@ def test_other_config_visibility(
...
+        no_answer_provided="No answer Provided",
+        edit_link_text="Change",
+        edit_link_aria_label="Change your answer for",
+        calculated_question={
+            "title": "Grand total of previous values",
+            "id": "calculated-summary-question",
+            "answers": [{"id": "calculated-summary-answer", "value": "£2.00"}],
+        },
+    )
+    assert len(rows) == 3
```
LJBabbage, line 502: I think the assert should be more specific than this. I reckon you need to build SummaryRows and SummaryRowItems of what is expected, then assert against that.

LJBabbage (changes requested):
This calculation looks like it works, but I need to check what is expected on the summary page. Using test_calculated_summary.json Question titles appear if there are multiple answers, but not if there are single. Additionally, not everything questions appears in the summary which looking at the schema I couldn't tell if that was expected

tests/app/test_jinja_filters.py
petechd: Renamed it to just `test_calculated_summary_question`

tests/app/test_jinja_filters.py
petechd: I think this has to stay, extra rows are not added when SummaryRow class object is being created, only at `map_summary_item_config` level.

tests/app/test_jinja_filters.py
```diff
@@ -442,3 +443,60 @@ def test_other_config_visibility(
...
```
MebinAbraham, line 502: To assert that this is not 3 rows with Grand total, I would just add something like:
```py
assert len(rows) == 3
assert rows[0].rowTitle == "First Number Question Title"
assert rows[1].rowTitle == "Second Number Question Title"
assert rows[2].rowTitle == "Grand total of previous values"
```

I would update the name to something more specific, like `test_calculated_summary_total_is_not_duplicated` since it isn't test all of calculated summary.

petechd: Extra assertions added, test renamed, thanks!

ajmaddaford: I don't think we currently have a test for the rest of the generated context. Is there a reason not to go a step further and assert the full content of rows?

petechd: Grand total is not added to question rows, it has its own rows. So before the change it would be something like 
```py
>>>  len(rows)
5
>>>  rows[0].rowTitle
>>>  "First Number Question Title"
>>>  rows[1].rowTitle
>>>  "Grand total of previous values"
>>>  rows[2].rowTitle
>>>  "Second Number Question Title"
>>>  rows[3].rowTitle
>>>  "Grand total of previous values"
>>>  rows[4].rowTitle
>>>  "Grand total of previous values"
```

ajmaddaford: I understand that this tests the defect, but there is still a gap in the tests for the rest of the context e.g. the edit link that should appear for the answer rows but not the totals.

MebinAbraham: So we do have a `TestCalculatedSummaryContext` but there are gaps in that, so that could be updated. I think other than that, we only have some functional tests. The jinja filters wraps a layer on top of that in its own format which should really be tested also.

petechd: Should I add some assertions? Something like:
```py
    assert (
        rows[0].rowItems[0].actions[0].ariaLabel
        == "Change your answer for First answer label"
    )
    assert (
        rows[0].rowItems[0].actions[0].attributes["data-qa"]
        == "first-number-answer-edit"
    )
    assert rows[0].rowItems[0].actions[0].attributes["data-ga"] == "click"
    assert rows[0].rowItems[0].actions[0].attributes["data-ga-category"] == "Summary"
    assert rows[0].rowItems[0].actions[0].attributes["data-ga-action"] == "Edit click"
    assert rows[0].rowItems[0].actions[0].text == "Change"
    hasattr(rows[2].rowItems[0], 'actions')
    assert len(rows[2].rowItems[0].attributes) == 1
    assert rows[2].rowItems[0].attributes["data-qa"] == "calculated-summary-answer"
```

petechd: Whole context is tested now, thanks👍

app/jinja_filters.py
```diff
@@ -409,6 +409,9 @@ def __init__(self, text, other=None):
         if other or other == 0:
             self.other = other
 
+    def __eq__(self, other):
+        return self.text == other.text
```
MebinAbraham, line 413: It is a good practice to check the instance is of same type. Also, you could compare the `__dict__` dunder methods since these are simple object, and it would keep them up to date if new attributes got added. In your definition, two `SummaryRowItemValue` objects with same text would return True even if the value of `self.other` was different.
```py
    def __eq__(self, other):
        return (
            isinstance(other, SummaryRowItemValue) and self.__dict__ == other.__dict__
        )
```
Applies to the other objects also.

MebinAbraham (commented):
Functionally works as expected 👍

app/jinja_filters.py
ajmaddaford: Are these `__eq__` methods needed? The `map_summary_item_config_processor` method casts the return from `map_summary_item_config` as a [dict](https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/jinja_filters.py#L572), could the same be done in the test?

petechd: > Are these `__eq__` methods needed? The `map_summary_item_config_processor` method casts the return from `map_summary_item_config` as a [dict](https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/jinja_filters.py#L572), could the same be done in the test?

I tested this and whenever `SummaryRowItem` and `SummaryRow` extends some class/classes inside it, it fails the unit test when I cast return as `dict`, so `dict(expected) == dict(result)` always fails. I get type errors.

petechd: > It is a good practice to check the instance is of same type. Also, you could compare the `__dict__` dunder methods since these are simple object, and it would keep them up to date if new attributes got added. In your definition, two `SummaryRowItemValue` objects with same text would return True even if the value of `self.other` was different.
> 
> ```python
>     def __eq__(self, other):
>         return (
>             isinstance(other, SummaryRowItemValue) and self.__dict__ == other.__dict__
>         )
> ```
> 
> Applies to the other objects also.

I could only compare using `isinstance()` when I try `__dict__` on classes that extend some other classes it fails..

petechd: > It is a good practice to check the instance is of same type. Also, you could compare the `__dict__` dunder methods since these are simple object, and it would keep them up to date if new attributes got added. In your definition, two `SummaryRowItemValue` objects with same text would return True even if the value of `self.other` was different.
> 
> ```python
>     def __eq__(self, other):
>         return (
>             isinstance(other, SummaryRowItemValue) and self.__dict__ == other.__dict__
>         )
> ```
> 
> Applies to the other objects also.

I added this everywhere except `SummaryRowItem` which fails tests if I add `self.__dict__ == other.__dict__` to its `__eq__` method.

MebinAbraham: @adammahmood I wasn't aware you can call `map_summary_item_config_processor` directly and get a dict back of all objects, thought that was a Flask specific thing that did the conversion to dict. If it can, then we don't need `__eq__`.

MebinAbraham (approved):

ajmaddaford (approved):

tests/app/test_jinja_filters.py
petechd: We build SummaryRows and SummaryRowItems now in unit-test and assertion is specific so this is resolved.

tests/app/test_jinja_filters.py
petechd: The test name and assertions has changed so this is resolved now.

PR #611, opened 2021-06-04
Set submission_language_code to -9 on flush

LJBabbage:
On flushing the submission_language_code should be set to -9

### How to review 
Check that the submission_language_code is -9 when flushing

### Checklist

### DIFF
--- a/app/routes/flush.py
+++ b/app/routes/flush.py
@@ -62,12 +62,12 @@ def _submit_data(user):
         router = Router(schema, answer_store, list_store, progress_store, metadata)
         full_routing_path = router.full_routing_path()
 
-        message = json.dumps(
-            convert_answers(
-                schema, questionnaire_store, full_routing_path, flushed=True
-            ),
-            for_json=True,
+        payload = convert_answers(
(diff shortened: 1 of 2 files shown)


app/routes/flush.py
```diff
@@ -62,12 +62,12 @@ def _submit_data(user):
...
 
-        message = json.dumps(
-            convert_answers(
-                schema, questionnaire_store, full_routing_path, flushed=True
-            ),
-            for_json=True,
+        payload = convert_answers(
+            schema, questionnaire_store, full_routing_path, flushed=True
         )
+        payload["submission_language_code"] = "-9"
```
LJBabbage, line 68: This might need checking as I am assuming it's string as language codes usually are

MebinAbraham (commented):
Happy with the change, will tick once string vs int is decided.

ajmaddaford (approved):

MebinAbraham (approved):

app/routes/flush.py
MebinAbraham: String has been agreed as appropriate.

PR #617, opened 2021-06-24
Remove placeholder identifier list

LJBabbage:
Currently we allow a list of identifiers in placeholders, this PR replaces them. The changes affect first_non_empty_item and concatenate_list

### How to review 
Check that no placeholder have a list of identifier and that everything works as before

### Checklist


app/questionnaire/placeholder_parser.py
```diff
@@ -97,8 +82,11 @@ def _parse_transforms(self, transform_list: Sequence[Mapping]):
 
         for transform in transform_list:
             transform_args: Dict[str, Union[None, str, List[str]]] = {}
+
             for arg_key, arg_value in transform["arguments"].items():
-                if not isinstance(arg_value, dict):
+                if isinstance(arg_value, list):
+                    transform_args[arg_key] = self._resolve_value_source_list(arg_value)
+                elif not isinstance(arg_value, dict):
```
ajmaddaford, line 89: I think it would be clearer if we made this more explicit by removing the `not`, something like:

```python
if isinstance(arg_value, list):
    transform_args[arg_key] = self._resolve_value_source_list(arg_value)
elif isinstance(arg_value, dict):
    if "value" in arg_value:
        transform_args[arg_key] = arg_value["value"]
    elif arg_value["source"] == "previous_transform":
        transform_args[arg_key] = transformed_value
    else:
        transform_args[arg_key] = self._resolve_value_source(arg_value)
else:
    transform_args[arg_key] = arg_value

```

app/questionnaire/placeholder_parser.py
```diff
@@ -113,6 +101,16 @@ def _parse_transforms(self, transform_list: Sequence[Mapping]):
...
 
+    def _resolve_value_source_list(self, value_source_list: list[Dict]) -> list[str]:
+        answers: list[str] = []
+        for value_source in value_source_list:
+            answer = self._resolve_value_source(value_source)
+            if isinstance(answer, list):
+                answers.extend(answer)
+            else:
+                answers.append(answer)
+        return answers
```
ajmaddaford, line 112: - This is building up a list of `values` rather than `answers`, as the value source could be metadata.
  ```suggestion
    def _resolve_value_source_list(self, value_source_list: list[Dict]) -> list[str]:
        values: list[str] = []
        for value_source in value_source_list:
            value = self._resolve_value_source(value_source)
            if isinstance(value, list):
                values.extend(value)
            else:
                values.append(value)
        return values
  ```

- The types need to be looked at as the values could be numbers (they could also be dicts e.g. for duration answers, but that wouldn't work here anyway).
- I've had a look but can't see it, is there a test that demonstrates using an answer value source that resolves to a list?

LJBabbage: Test no, schema yes, https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/test_schemas/en/test_question_title_in_error.json#L105, so perhaps a specific test is appropriate. However I do wonder if we should have a separate transformation for checkboxes concatenating lists. In this instance the transform does not need to be a list, which the story dictates. I can't see why we would combine checkboxes or checkboxes and strings

app/questionnaire/placeholder_parser.py
```diff
@@ -96,23 +81,41 @@ def _parse_transforms(self, transform_list: Sequence[Mapping]):
         transformed_value = None
 
         for transform in transform_list:
-            transform_args: Dict[str, Union[None, str, List[str]]] = {}
+            transform_args: Dict[
+                str, Union[None, str, List[Union[str, int, float]]]
```
ajmaddaford, line 85: I think answers can either be `str`, `int`, `Decimal`, `dict` (duration and address answers), or `list[str]` (checkbox answers).
```suggestion
                str, Union[None, str, int, Decimal, dict, list[str]]
```

We should also be using the built in types rather than the ones from `typing` e.g. `list` rather than `List`. I can see that this matches what was already there, but as long as it doesn't cause any problems we should update it.

app/questionnaire/placeholder_parser.py
ajmaddaford: I think the checkbox unit test that you've added is enough. I think we should either leave it as it is, or change the argument to allow an object (for a value source that resolved to a list i.e. a checkbox answer) or an array (for a list of value sources).

app/questionnaire/placeholder_parser.py
```diff
@@ -96,23 +82,39 @@ def _parse_transforms(self, transform_list: Sequence[Mapping]):
...
-                    transform_args[arg_key] = self._resolve_value_source(arg_value)
+                    transform_args[arg_key] = arg_value
 
             transformed_value = getattr(self._transformer, transform["transform"])(
                 **transform_args
             )
 
         return transformed_value
 
+    def _resolve_value_source_list(self, value_source_list: List[Dict]) -> list[str]:
```
ajmaddaford, line 108: The built-in types should be used.
```suggestion
    def _resolve_value_source_list(self, value_source_list: list[dict]) -> list[str]:
```

I think all types in this file should be changed to use the built-in types.

berroar (approved):
Tested this, couldn't find any issues beyond what had already been commented  👍

ajmaddaford (approved):

PR #621, opened 2021-07-06
Add survey type config for business

LJBabbage:
This PR adds a business config for business templates. I have added some dummy links taken from https://surveys.ons.gov.uk/sign-in/ but I am unsure that is what we want so haven't tested it bar the one contact_us link.

N.B I have used # instead of links as we currently don't have them and the story says so.

### How to review 
Make sure the new theme works as expected. test_business_theme.json can be used

### Checklist


test_schemas/en/test_theme_business.json
```diff
@@ -0,0 +1,76 @@
+{
```
LJBabbage, line 1: I have added this as we have schemas for a couple of the other themes (however health is missing), however we don't use them in any tests. It does raises the question with what we want to do with theme schemas and testing.

tests/functional/spec/census_exit.spec.js
```diff
@@ -14,7 +12,7 @@ describe("Post submission exit", () => {
     $(HubPage.submit()).click();
     browser.deleteAllCookies();
     $(CensusThankYouPage.exit()).click();
-    expect(browser.getUrl()).to.equal(BASE_URL);
+    expect(browser.getUrl()).to.equal("https://surveys.ons.gov.uk/sign-in/");
```
LJBabbage, line 15: As discussed in Slack, this looks odd, but is as expect now surveys.ons.gov.uk is default when session is removed

app/helpers/template_helpers.py
```diff
@@ -327,11 +328,47 @@ class CensusNISRASurveyConfig(
...
+    footer_legal_links: Iterable[Mapping] = field(
+        default_factory=lambda: [
+            Link(lazy_gettext("Cookies"), f"{BUSINESS_BASE_URL}/cookies/").__dict__,
+            Link(
+                lazy_gettext("Privacy and data protection"),
+                f"{BUSINESS_BASE_URL}/privacy-and-data-protection/",
+            ).__dict__,
+        ],
+        compare=False,
+    )
```
MebinAbraham, line 364: I think the links should be placeholders since we don't have confirmed locations yet. Some of the content from RAS might move also. If we decided to keep it, the base URL need to be updated to `"https:www.//ons.gov.uk"`.

LJBabbage: Problem with half baked stories, all our test schemas would use this and links would be broken/placeholders/might be wrong. I am more incline to remove the links completely and not have loose ends, which might be confused as already being implemented

LJBabbage: Changed to # :frowning:

ajmaddaford (commented):
We agreed we would separate the business and census configs into their own files.
> introduce separate config class files for `business` and `census`

test_schemas/en/test_theme_business.json
MebinAbraham: I think we should just remove them if they don't have tests. They could be added in when needed with the appropriate tests.

app/helpers/template_config/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+from .survey_config import SurveyConfig
+
+__all__ = [
+    "SurveyConfig",
+    "CensusSurveyConfig",
+    "CensusNISRASurveyConfig",
+    "WelshCensusSurveyConfig",
+    "BusinessSurveyConfig",
+    "Link",
+]
```
ajmaddaford, line 17: I think as the folder contains survey configs, it should be named `survey_configs`. We should also consider moving the folder to the top level as it's not a helper.

app/helpers/template_config/business_config.py
```diff
@@ -0,0 +1,41 @@
+from dataclasses import dataclass, field
+from typing import Iterable, Mapping, MutableMapping
+
+from flask_babel import lazy_gettext
+
+from app.helpers.template_config.link import Link
+from app.helpers.template_config.survey_config import SurveyConfig
+
+BUSINESS_BASE_URL = "https://surveys.ons.gov.uk"
```
ajmaddaford, line 9: Now that this is in it's own file this could be shortened to `BASE_URL`.

app/helpers/template_config/census_config.py
```diff
@@ -0,0 +1,174 @@
...
+
+from flask_babel import lazy_gettext
+from flask_babel.speaklater import LazyString
+
+from app.helpers.template_config.link import Link
+from app.helpers.template_config.survey_config import SurveyConfig
+
+CENSUS_EN_BASE_URL = "https://census.gov.uk"
+CENSUS_CY_BASE_URL = "https://cyfrifiad.gov.uk"
+CENSUS_NIR_BASE_URL = f"{CENSUS_EN_BASE_URL}/ni"
```
ajmaddaford, line 12: Now that these are in their own file they could be shortened to `EN_BASE_URL` etc.

ajmaddaford (approved):

PR #626, opened 2021-07-14
Add email and telephone transforms

LJBabbage:
This PR adds telephone number and email link transforms for placeholders

### How to review 
Using test_introduction.json make the email and telephone links appear and look as expected

### Checklist


tests/functional/spec/introduction.spec.js
```diff
@@ -12,8 +12,5 @@ describe("Introduction page", () => {
     expect($(IntroductionPage.introDescription()).getText()).to.contain(
       "To take part, all you need to do is check that you have the information you need to answer the survey questions."
     );
-    expect($(IntroductionPage.introTitleDescription()).getText()).to.contain(
```
LJBabbage, line 15: Oddly name, but as we have absorbed this now into content the data-qa doesn't exist. You could argue we should still test for this, but after a quick look around we don't seem to test like this anywhere else for placeholders, just integration

tests/integration/questionnaire/test_questionnaire_placeholders.py
```diff
@@ -34,6 +34,11 @@ def test_title_placeholders_rendered_in_summary(self):
 
     def test_placeholders_rendered_in_pages(self):
         self.launchSurvey("test_placeholder_transform")
+        self.assertInBody(
+            'If the company details or structure have changed contact us on <a href="tel:03001234931">0300 1234 931</a> '
+            'or email <a href="mailto:surveys@ons.gov.uk?subject=Change%20of%20details%20reference%20123456789012A">surveys@ons.gov.uk</a>'
+        )
```
ajmaddaford, line 40: It seems unnecessary to duplicate the schema from `test_introduction.json` into `test_placeholder_transform.json` just so it can be tested here. Can a test be added that uses `test_introduction.json` instead?

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -229,3 +229,30 @@ def test_format_name(self):
...
+            self.transforms.email_link("test@email.com", "test subject")
+            == '<a href="mailto:test@email.com?subject=test%20subject">test@email.com</a>'
+        )
+
+    def test_email_link_without_subject(self):
+
+        assert (
+            self.transforms.email_link("test@email.com")
+            == '<a href="mailto:test@email.com">test@email.com</a>'
+        )
```
ajmaddaford, line 251: I think it would be better to build the tests up as:
- `test_email_link`
- `test_email_link_with_subject`
- `test_email_link_with_subject_and_reference`

tests/integration/questionnaire/test_questionnaire_placeholders.py
LJBabbage: We perhaps need to have a rethink on test_placeholder_transform.json as I instinctively thought all our transforms would live in there, perhaps they should

ajmaddaford: Yeah, it's not clear, as there is also the `test_placeholder_full` schema that also tests placeholder transforms. I think as it already exists in the `test_introduction` it isn't needed in either of these schemas.

ajmaddaford (approved):

petechd (approved):
Reminder to change validator branch back to `latest`.

PR #628, opened 2021-07-22
Add view_submitted_response to submission handler

LJBabbage:
This PR reintroduces view_submitted_response, which if true means the question store is not deleted on submission

### How to review 
I haven't added an integration/functional test or a schema, but I have tested the handler. My theory is until we have the feature that allows the visualisation of answers on post submission, there isn't a great deal to test integration/functional wise. We already have a test that checks the store is deleted so don't need to check that again, just that it is being called or not

### Checklist

### DIFF
--- a/tests/app/views/handlers/test_submission_handler.py
+++ b/tests/app/views/handlers/test_submission_handler.py
@@ -27,26 +27,79 @@ class TestSubmissionPayload(AppContextTestCase):
             ru_ref="ru_ref",
             case_id="0123456789000000",
         )
-        self.session_store = SessionStore("user_ik", "pepper", "eq_session_id")
         self.expires_at = datetime.now(tzutc()) + timedelta(seconds=5)
-
-    def test_submission_language_code_in_payload(self):
-        session_store = self.session_store.create(
+        self.session_store = SessionStore("user_ik", "pepper", "eq_session_id").create(
             "eq_session_id", "user_id", self.session_data, self.expires_at
         )
-        storage = Mock()
-        storage.get_user_data = Mock(return_value=("{}", 1))
 
+    def test_submission_language_code_in_payload(self):
         with patch(
             "app.views.handlers.submission.get_session_store",
-            return_value=session_store,
+            return_value=self.session_store,
         ):
             with patch(
                 "app.views.handlers.submission.convert_answers", return_value={}
             ):
                 submission_handler = SubmissionHandler(
-                    QuestionnaireSchema({}), QuestionnaireStore(storage), {}
+                    QuestionnaireSchema({}), self.questionnaire_store_mock(), {}
                 )
                 assert (
                     submission_handler.get_payload()["submission_language_code"] == "cy"
                 )
+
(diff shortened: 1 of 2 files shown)


tests/app/views/handlers/test_submission_handler.py
```diff
@@ -27,26 +27,79 @@ def setUp(self):
...
-                    QuestionnaireSchema({}), QuestionnaireStore(storage), {}
+                    QuestionnaireSchema({}), self.questionnaire_store_mock(), {}
                 )
                 assert (
                     submission_handler.get_payload()["submission_language_code"] == "cy"
                 )
+
+    def test_questionnaire_store_delete_called(self):
+        questionnaire_store = self.questionnaire_store_mock()
+        self.questionnaire_store_delete_mock(questionnaire_store)
```
MebinAbraham, line 52: This method doesn't add much. I think it should just be inline for clarity or made it part of `self.questionnaire_store_mock()`.

tests/app/views/handlers/test_submission_handler.py
```diff
@@ -27,26 +27,79 @@ def setUp(self):
...
+                    submission_handler = SubmissionHandler(
+                        QuestionnaireSchema(
+                            {"submission": {"view_submitted_response": "True"}}
+                        ),
+                        questionnaire_store,
+                        {},
+                    )
+                    submission_handler.submit_questionnaire()
+
+                    assert not questionnaire_store.delete.called
```
MebinAbraham, line 93: The nesting makes it hard to read. I think `with patch` should only be used if you need to patch and return something you have in scope i.e `self.session_store`, for anything else it can be done using the decorator which also removes 1 level of nesting making it easier on the eyes. Assigning schema to a variable makes the formatting nicer on the eyes also.
```py
@patch(
    "app.views.handlers.submission.SubmissionHandler.get_payload",
    Mock(return_value={}),
)
def test_view_submitted_response_true_questionnaire_store_delete_not_called(self):
    questionnaire_store = self.questionnaire_store_mock()

    with self.app_request_context():
        with patch(
            "app.views.handlers.submission.get_session_store",
            return_value=self.session_store,
        ):
            schema = QuestionnaireSchema(
                {"submission": {"view_submitted_response": "True"}}
            )
            submission_handler = SubmissionHandler(
                schema, questionnaire_store, full_routing_path=[]
            )
            submission_handler.submit_questionnaire()

            assert not questionnaire_store.delete.called
```
In this example, I put `full_routing_path=[]` because it was being passed `{}` when it should be a list. (would get picked up if we had type hints). But usually a good practice to add keyword args for empty containers to make it obvious what it is being passed to.

LJBabbage: I had considered that, but aren't a huge fan of mixing decorators and then having patches in the code, I actually find that harder on the eyes,  but am happy to change it for some green.

MebinAbraham (approved):

berroar (approved):

PR #629, opened 2021-07-22
Add support for new routing rules

MebinAbraham:
Added support for the new when rules schema design introduced by https://github.com/ONSdigital/eq-questionnaire-validator/pull/96 

- This PR adds the interface to support the new operator and operations. This class is not yet used to evaluate routing for a questionnaire, that will be done in the next piece of work. This introduces the main interface and unit tests to go with it. However, because of this, it is hard to know if any edge cases have been missed that might only be captured by an integration test.
- Normal functions have been used for the operations instead of anonymous `lambda` functions to take advantage of type hints. Even for the simple ones like `not`, `lambda` has not been used to keep it consistent.

This PR depends on: https://github.com/ONSdigital/eq-questionnaire-runner/pull/630

### How to review 
Ensure the changes makes sense and that there are adequate unit tests in place.

### Checklist


LJBabbage (commented):
I reckon we should take this opportunity to document the rules and usages. Looking at the tests I find it hard to determine if anything is missing and see some instances where it looks like duplication, is [(0.5, 0.5), True] not the same as [(3, 3), True] ? If we did document the rules, we could use it as contract, which we could write tests against. It would also make it easier for author and anyone else when looking at our rules. The flip side would be ones we don't mention are untested.

MebinAbraham:
> I reckon we should take this opportunity to document the rules and usages. Looking at the tests I find it hard to determine if anything is missing and see some instances where it looks like duplication, is [(0.5, 0.5), True] not the same as [(3, 3), True] ? If we did document the rules (even a simple grid might help), we could use it as contract, which we write tests against. It would also make it easier for author and anyone else when looking at our rules. The flip side would be ones we don't mention are untested.

Document in what sense?. We have the initial doc for routing rules in runner but now also the schemas itself in validator which is the source of truth?. Not quite sure what other documentation you mean.
So for why [(0.5, 0.5), True] is tested as well as [(3, 3), True] is to check diff types, to ensure for example floats werent being casted for ints etc. I could make the tests clearer if I did the params inline but would cause a lot of duplication. But happy to consider either option.

LJBabbage:
If I come to the repo afresh, there is no guide or at least a sign post to what rules we have and which ones are supported, which has been raised as an issue before. Even simple comments next to the parametrize list would help

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -0,0 +1,62 @@
...
+        operator = Operator(next(iter(rule)))
+        operands = rule[operator.name]
+
+        if not isinstance(operands, (list, tuple)):
+            raise TypeError(
+                f"The rule is invalid. Got a non list/tuple for operands - {type(operands)}"
+            )
+
+        operands = (self._evaluate(operand) for operand in operands)
+        return operator.evaluate(operands)
```
ajmaddaford, line 59: I find it confusing that `_evaluate` is called when the operand is a value rather than a rule, only to return the value immediately. I expected this to work like the validator code i.e. checking that something is a rule before recursively calling the method.

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -0,0 +1,283 @@
...
+now_as_yyyy_mm_dd = now.strftime("%Y-%m-%d")
+now_as_yyyy_mm = now.strftime("%Y-%m")
+now_as_yyyy = now.strftime("%Y")
+
+test_data_equals_operation_numeric_and_date_matching_values = [
+    [(0.5, 0.5), True],
+    [(1.0, 1), True],
+    [(3, 3), True],
+    [(now, now), True],
+]
```
ajmaddaford, line 20: I think the same comment made on the value sources pr applies here too. The test data is too far removed from the tests and it makes them hard to read. We should consider some duplication to make the tests more readable.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -0,0 +1,993 @@
+from datetime import datetime
```
ajmaddaford, line 1: We need to think about what needs to be tested, as these tests are hard to digest. I don't think we need to test every operation with every combination of value source. A value source will resolve to a value. As long as we have tests to ensure the value source resolution works (covered by the value source resolver tests), we should only need to test that the rules work with value sources i.e. not all of the values that value sources can return.

app/questionnaire/routing/when_rule_evaluator.py
MebinAbraham: I did that to reduce nesting and have a true recursive nature since the operations are not expensive, plus it returns a generator. Happy to change it to something like this, which only calls evaluate when it needs to resolve an operation and still benefits from a generator.
```py
    def _evaluate(self, rule: answer_value_types):
        operator = Operator(next(iter(rule)))
        operands = rule[operator.name]

        if not isinstance(operands, list):
            raise TypeError(
                f"The rule is invalid. Got a non list type for operands - {type(operands)}"
            )

        resolved_operands = self.get_resolved_operands(operands)
        return operator.evaluate(resolved_operands)

    def get_resolved_operands(self, operands: list):
        for operand in operands:
            if isinstance(operand, dict) and not self._is_date_offset(operand):
                if "source" in operand:
                    yield self.value_source_resolver.resolve(operand)
                else:
                    yield self._evaluate(operand)
            else:
                yield operand
```

ajmaddaford: This is much more readable 👍 

> I didn't explicitly check for operator types, as that isn't strictly necessary.

My concern with this is:
- logic about a specific operator has leaked into this code
- as more operators get added this could overloaded with checks

app/questionnaire/routing/operations.py
```diff
@@ -0,0 +1,85 @@
...
+    return lhs == rhs
+
+
+@casefold
+def evaluate_not_equal(lhs: answer_types, rhs: answer_types) -> bool:
+    return lhs != rhs
+
+
+def evaluate_greater_than(lhs: answer_types, rhs: answer_types) -> bool:
+    return _are_non_none_operands(lhs, rhs) and lhs > rhs  # type: ignore
```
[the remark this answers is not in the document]
MebinAbraham: Ah nice one didn't think of that. So this gets us mostly there, but I would have to check that the operands are not None inline for each operation for it to work. As it can't work out we are check for None in `_are_non_none_operands`. So:
```py
def evaluate_greater_than(lhs: AnyValue, rhs: AnyValue) -> bool:
    return lhs is not None and rhs is not None and lhs > rhs
```

Thoughts?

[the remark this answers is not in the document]
ajmaddaford: Another option could be doing the check before the method is called? (for this we would need to create a list of operations that don't work with `None` values).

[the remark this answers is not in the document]
MebinAbraham: I did have that initially but then thought each operation should be self-contained, but will revisit given it makes typing better.

app/questionnaire/routing/helpers.py
```diff
@@ -0,0 +1,30 @@
...
+
+from app.questionnaire.value_source_resolver import answer_value_types
+
+
+def _casefold(value: answer_value_types) -> answer_value_types:
+    try:
+        return (
+            [_casefold(v) for v in value]
+            if isinstance(value, (list, tuple))
+            else value.casefold()  # type: ignore
```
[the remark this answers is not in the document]
MebinAbraham: Ya, this was because it was doing EAFP but will make to LBYL to make typing better.

app/questionnaire/routing/operations.py
```diff
@@ -0,0 +1,85 @@
...
+
+answer_types = Union[bool, str, int, float, None, datetime]
+
+
+def _are_non_none_operands(*operands: Union[Sequence, answer_types]) -> bool:
+    return all(operand is not None for operand in operands)
+
+
+@casefold
+def evaluate_equal(lhs: answer_types, rhs: answer_types) -> bool:
```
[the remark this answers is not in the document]
MebinAbraham: This has been updated as a result of https://github.com/ONSdigital/eq-questionnaire-runner/pull/629#discussion_r677254775

app/questionnaire/routing/operations.py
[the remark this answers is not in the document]
MebinAbraham: Operations that require non NoneType values are now explicitly checked. That, with the use of `TypeVar` the mypy ignore comment have been removed.

app/questionnaire/routing/helpers.py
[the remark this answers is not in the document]
MebinAbraham: Fixed by [`ca00c07` (#629)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/629/commits/ca00c077b76f6d0dd4c564cdcb45083483c71073)

app/questionnaire/routing/operations.py
```diff
@@ -0,0 +1,85 @@
...
+
+from dateutil.relativedelta import relativedelta
+
+from app.questionnaire.routing.helpers import casefold, datetime_as_midnight
+from app.questionnaire.rules import convert_to_datetime
+
+answer_types = Union[bool, str, int, float, None, datetime]
+
+
+def _are_non_none_operands(*operands: Union[Sequence, answer_types]) -> bool:
```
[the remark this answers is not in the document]
MebinAbraham: Renamed to `_operands_not_none` and now only done once upfront in `Operator.py`

app/questionnaire/routing/when_rule_evaluator.py
MebinAbraham: Updated to the above snippet while also explicitly checking Operator types for the reasons you highlighted above.

tests/app/questionnaire/routing/test_operators.py
MebinAbraham: Duplicated test date to improve readability in both this file and the rules file also.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
MebinAbraham: I have reduced the number of combinations that run and made all the test data in line to improve readability at the cost of duplication.

For date, I decided to keep the tests with multiple operators, as that is not tested in `test_operators` as a comparison only from the perspective that it resolves values.

app/questionnaire/routing/when_rule_evaluator.py
ajmaddaford: I can't see the change to explicitly check operator types.

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -0,0 +1,72 @@
...
+        self, operands: list[ValueSourceTypes]
+    ) -> Generator[Union[bool, Optional[datetime], ValueSourceTypes], None, None]:
+        for operand in operands:
+            if isinstance(operand, dict) and not self._is_date_offset(operand):
+                if "source" in operand:
+                    yield self.value_source_resolver.resolve(operand)
+                elif any(operator in operand for operator in OPERATIONS):
+                    yield self._evaluate(operand)
+                else:
+                    raise NotImplementedError("Got an unsupported operand")
```
ajmaddaford, line 67: The checking for a date offset should be removed as this still doesn't address the concerns:
- logic about a specific operator (date offset) has leaked into this code
- as more operators get added this could become overloaded with checks

If the dict is neither a value source or operator, it should be considered an operand. 

Also, for most occurrences of a dict it will be a value source or operator, so it should check for those first.

MebinAbraham: Misunderstood what you initially meant. Have updated to address this now.

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -0,0 +1,291 @@
...
+from app.questionnaire.routing.helpers import datetime_as_midnight
+from app.questionnaire.routing.operator import Operator
+from app.questionnaire.rules import convert_to_datetime
+
+now = datetime.utcnow()
+now_as_yyyy_mm_dd = now.strftime("%Y-%m-%d")
+now_as_yyyy_mm = now.strftime("%Y-%m")
+now_as_yyyy = now.strftime("%Y")
+
+equals_operations = [
```
LJBabbage, line 15: This would be easy if ordered, either by type (string, dates, bool etc) my preference, or result (True/False) it currently looks random to my eyes and difficult to see the gaps

app/questionnaire/routing/helpers.py
```diff
@@ -0,0 +1,30 @@
...
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        casefolded_args = [_casefold(arg) for arg in args]
+        casefolded_kwargs = {k: _casefold(v) for k, v in kwargs.items()}
+        return func(*casefolded_args, **casefolded_kwargs)
+
+    return wrapper
+
+
+def datetime_as_midnight(date: datetime) -> datetime:
```
LJBabbage, line 29: Never a massive fan of the term midnight as can be ambiguous to some (today/tomorrow) although not pretty datetime_as_today_start at least removes some confusion

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -0,0 +1,63 @@
...
+            use_default_answer=True,
+        )
+
+    def _evaluate(self, rule: dict[str, list]) -> Union[bool, Optional[datetime]]:
+        operator = Operator(next(iter(rule)))
+        operands = rule[operator.name]
+
+        if not isinstance(operands, list):
+            raise TypeError(
+                f"The rule is invalid. Got a non list type for operands - {type(operands)}"
```
LJBabbage, line 43: "The rule is invalid. Got a non list type for operands" -> "The rule is invalid, operands should be list type and not {type(operands)}"

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -0,0 +1,291 @@
...
+def test_operation_not_equal(operands, expected_result):
+    operator = Operator(Operator.NOT_EQUAL)
+    assert operator.evaluate(operands) is not expected_result
+
+
+greater_than_and_less_than_operations = [
+    [(0.7, 0.5), True],
+    [(2, 1.0), True],
+    [(7, 3), True],
+    [(datetime.utcnow(), now), True],
```
LJBabbage, line 56: This is interesting as now = datetime.utcnow() I think we need to be a careful with this one, and do a proper offset

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -0,0 +1,291 @@
...
+    [(0.5, 0.7), False],
+    [(1.0, 2), False],
+    [(3, 7), False],
+    [(now, datetime.utcnow()), False],
```
LJBabbage, line 60: As above

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -0,0 +1,291 @@
...
+)
+def test_operation_less_than(operands, expected_result):
+    operator = Operator(Operator.LESS_THAN)
+    assert operator.evaluate(operands) is not expected_result
+
+
+@pytest.mark.parametrize(
+    "operands, expected_result",
+    [
+        [(0.7, 0.5), False],
```
LJBabbage, line 85: A better order, but still would prefer each type true, then false.

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -0,0 +1,291 @@
...
+)
+def test_operation_less_than_or_equal(operands, expected_result):
+    operator = Operator(Operator.LESS_THAN_OR_EQUAL)
+    assert operator.evaluate(operands) is expected_result
+
+
+@pytest.mark.parametrize(
+    "operands, expected_result",
+    [
+        [(0.7, 0.5), True],
```
LJBabbage, line 107: As above, I don't see order in this list

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -0,0 +1,965 @@
...
+        list_store=list_store,
+        location=location,
+        routing_path_block_ids=routing_path_block_ids,
+    )
+
+
+@pytest.mark.parametrize(
+    "rule, expected_result",
+    [
+        ({Operator.NOT: [False]}, True),
```
LJBabbage, line 64: Oh, I like this, much easier to follow

tests/app/questionnaire/routing/test_operators.py
MebinAbraham: So `now` is the time the test suite started, so a newly generated `datetime.utcnow()` would always be greater unless frozen. Have now assigned a future date using a delta to a variable for clarity.

tests/app/questionnaire/routing/test_operators.py
MebinAbraham: Forgot to order when I got rid of some tests. I went with ordering by result, then by type.

tests/app/questionnaire/routing/test_operators.py
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/629#discussion_r681055227

tests/app/questionnaire/routing/test_operators.py
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/629#discussion_r681055227

app/questionnaire/routing/when_rule_evaluator.py
MebinAbraham: Updated with a slight change in [`ed7e6c4` (#629)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/629/commits/ed7e6c44d22662e34cf62426400e5b7d9a315da1).

tests/app/questionnaire/routing/test_operators.py
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/629#discussion_r681147775

app/questionnaire/routing/helpers.py
MebinAbraham: See what others think, as I'm not a big fan of `datetime_as_today_start`. I thought midnight was clear since it for a given datetime. Happy to consider other schemes if it is prone to tripping people up.

tests/app/questionnaire/routing/test_operators.py
LJBabbage: I realised that it would be higher, just prefer it to be explicit as to a less seasoned eye it does look a bit odd

app/questionnaire/routing/helpers.py
LJBabbage: No neither do I, I just dislike the term midnight. Straw poll of the one person around me reckoned midnight was tomorrow after closing one eye and screwing up nose.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -0,0 +1,965 @@
...
+        ({Operator.GREATER_THAN: [1, 3]}, False),
+        ({Operator.GREATER_THAN_OR_EQUAL: [0.5, 1]}, False),
+        ({Operator.LESS_THAN: [1, 0.5]}, False),
+        ({Operator.LESS_THAN_OR_EQUAL: [1, 0.5]}, False),
+        ({Operator.IN: ["No", ["Yes"]]}, False),
+        ({Operator.ANY_IN: [["Maybe"], ["Yes", "No"]]}, False),
+        ({Operator.ALL_IN: [["Yes", "Maybe"], ["Yes", "No"]]}, False),
+    ],
+)
+def test_all_boolean_operators_as_rule(rule, expected_result):
```
LJBabbage, line 91: Probably don't need the all, (the other tests don't have this), specially as all does mean something else in this context (i.e all_in)

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -0,0 +1,965 @@
...
+                        {"source": "metadata", "identifier": "region_code"},
+                        ["GB-NIR"],
+                    ]
+                },
+            ],
+            False,
+        ),
+    ],
+)
+def test_logic_and(operator, operands, expected_result):
```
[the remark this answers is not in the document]
LJBabbage: I agree. Have you put this in to show an expanded and operator with 2 rules, as I think the combination of ({Operator.AND: [True, True]}, True), ({Operator.EQUAL: ["Yes", "Yes"]}, True), and ({Operator.IN: ["Yes", ["Yes"]]}, True) in test_all_boolean_operators_as_rule although separate is good enough

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -0,0 +1,63 @@
...
+from app.questionnaire.routing.operator import OPERATIONS, Operator
+from app.questionnaire.value_source_resolver import (
+    ValueSourceResolver,
+    ValueSourceTypes,
+)
+
+
+@dataclass
+class WhenRuleEvaluator:
+    rule: dict
```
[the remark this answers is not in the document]
MebinAbraham: Good idea. 👍

app/questionnaire/routing/operations.py
```diff
@@ -0,0 +1,83 @@
...
+    return all(iter(values))
+
+
+def evaluate_or(values: Iterable[bool]) -> bool:
+    return any(iter(values))
+
+
+@casefold
+def evaluate_in(lhs: NonArrayPrimitiveTypes, rhs: Sequence) -> bool:
+    return rhs is not None and lhs in rhs
```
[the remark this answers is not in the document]
MebinAbraham: Was done inline as `lhs` is allowed to be `None`, as we could do `None in [None, "Yes"]` as a shortcut to not answered or answer equal "Yes".

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -0,0 +1,965 @@
...
+                {
+                    Operator.IN: [
+                        {"source": "metadata", "identifier": "region_code"},
+                        ["GB-ENG", "GB-WLS"],
+                    ]
+                },
+            ],
+            True,
+        ),
+        # Test inverse
```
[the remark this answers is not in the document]
MebinAbraham: Sectioned into `Test True` and `Test False`.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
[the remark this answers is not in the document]
MebinAbraham: Removed. This was done first and was simply left behind after introducing the `test_all_boolean_operators_as_rule `.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -0,0 +1,965 @@
...
+                        {"source": "answers", "identifier": "answer-1"},
+                        "No, I do not",
+                    ]
+                },
+            ],
+            False,
+        ),
+    ],
+)
+def test_logic_or(operator, operands, expected_result):
```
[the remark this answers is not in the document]
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/629#discussion_r681919980

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -0,0 +1,965 @@
...
+        ),
+    )
+
+    assert when_rule_evaluator.evaluate() is expected_result
+
+
+@pytest.mark.parametrize(
+    "answer_value, expected_result", [("No", True), ("Yes", False)]
+)
+def test_logic_not(answer_value, expected_result):
```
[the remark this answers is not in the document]
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/629#discussion_r681919980

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -0,0 +1,297 @@
...
+    assert operator.evaluate(operands) is expected_result
+
+
+@pytest.mark.parametrize(
+    "operands, expected_result",
+    [
+        [(True, True), True],
+        [(True, True, True, True), True],
+        [(True, False), True],
+        [(False, False, False, True), True],
```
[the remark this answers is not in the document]
MebinAbraham: Whoops 🙈. Added, thanks.

app/questionnaire/routing/operator.py
```diff
@@ -0,0 +1,84 @@
...
+        self, operands: Union[Generator, Iterable]
+    ) -> Union[bool, Optional[datetime]]:
+        if self._ensure_operands_not_none:
+            operands = list(operands)
+            if self._operands_not_none(*operands) is False:
+                return False
+
+        value: Union[bool, Optional[datetime]] = (
+            self._operation(operands)
+            if self._short_circuit
```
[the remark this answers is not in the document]
MebinAbraham: Updated manually as there was multiple references.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -0,0 +1,965 @@
...
+
+@pytest.mark.parametrize(
+    "is_repeating_answer, list_item_id_for_answer, expected_result",
+    [
+        (True, None, False),
+        (True, "item-2", False),
+        (True, "item-1", True),
+        (False, None, True),
+        (False, "item-2", False),
+        (False, "item-1", False),
```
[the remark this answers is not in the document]
MebinAbraham: I guess the confusion is from the way it is used. The idea is that it is just the `list_item_id_for_answer` used in the `AnswerStore`. So that when not in a repeat, the answer store might still have an answer but with list_item_id, and this ensures it doesn't fetch that and alternatively when answer is repeating, and answer store doesn't have an answer with matching list item id, it returns False. The `list_item_id_for_answer` is used for the Answer in the answer store and not the location itself, that might be where the confusion is. Happy to swap it around so it is the location that changes, if it helps.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -0,0 +1,965 @@
...
+                            ]
+                        },
+                    ]
+                },
+            ],
+            True,
+        ),
+    ],
+)
+def test_nested_rules(operator, operands, expected_result):
```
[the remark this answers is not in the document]
MebinAbraham: It was to test nesting with just more than one operator and where the second operator `OR` only returns True on 1 side. Happy to drop one, if we think it isn't needed.

app/questionnaire/routing/when_rule_evaluator.py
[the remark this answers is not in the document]
MebinAbraham: Done.

app/questionnaire/routing/helpers.py
MebinAbraham: @ajmaddaford got a preference or any other suggestions?

tests/app/questionnaire/routing/test_when_rule_evaluator.py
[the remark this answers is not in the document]
ajmaddaford: I think one should be dropped as we test the operators individually, and this is just to test the nesting works.

app/questionnaire/routing/operations.py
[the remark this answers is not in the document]
ajmaddaford: Ah, ok, I thought there would be a reason. We should add something along the lines of what you've written in a comment above the return line.

app/questionnaire/routing/helpers.py
ajmaddaford: `datetime_at_start_of_day`?

LJBabbage (approved):

tests/app/questionnaire/routing/test_when_rule_evaluator.py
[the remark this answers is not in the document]
MebinAbraham: As discussed, this has been simplified and now is called `test_rule_access_answer_outside_of_repeating_section`. The other scenarios are already tested between the existing tests.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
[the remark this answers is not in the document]
MebinAbraham: Dropped one of them.

app/questionnaire/routing/operations.py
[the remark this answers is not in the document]
MebinAbraham: Added a comment.

app/questionnaire/routing/helpers.py
MebinAbraham: This method has now been dropped because `date` is now returned instead of `datetime`.

ajmaddaford: Resolving this as the method has been removed.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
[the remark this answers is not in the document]
ajmaddaford: We now have have some similar tests with inconsistent names:
```
test_rule_access_answer_outside_of_repeating_section
test_answer_source_with_answer_not_on_path_non_repeating_section
test_answer_source_with_answer_not_on_path_repeating_section
```

Maybe:
```
test_answer_source_outside_of_repeating_section
test_answer_source_not_on_path_non_repeating_section
test_answer_source_not_on_path_repeating_section
```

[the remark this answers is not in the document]
MebinAbraham: Done. Since these also existed in value source resolver tests in terms of resolving, I have updated those also.

ajmaddaford (approved):

PR #631, opened 2021-07-27
Add submitted_at to Questionnaire State

LJBabbage:
This PR adds submitted_at to Questionnaire State so State can be deleted at a later date

### How to review 
I have added view_submitted_response to test_submit_with_summary_custom_submission_text.json so it can be tested on GCP if required. If so, deploy this branch using a pipeline and launch  test_submit_with_summary_custom_submission_text, open up datastore and entities selecting questionnaire-state, check the submitted_at is intialized as null and stays as so until the questionnaire is submitted.

N.B This piece and the view_submitted_response story has introduced a bug. If view_submitted_response is set and the survey is submitted, if you were to resume the survey with the correct Response ID it will take you to the submission page even though you have already submitted, because of this you can submitted again and the date is updated. This probably needs a change in router, but feels like it needs to be discussed. We don't use view_submitted_response and have no way to visualise it, so I can't see why this should delay this story.

### Checklist


test_schemas/en/test_submit_with_summary_custom_submission_text.json
```diff
@@ -33,7 +33,8 @@
         "button": "Submission button",
         "guidance": "Submission guidance",
         "title": "Submission title",
-        "warning": "Submission warning"
+        "warning": "Submission warning",
+        "view_submitted_response": true
```
LJBabbage, line 37: Purely for testing on GCP, will be removed on merge

app/views/handlers/submission.py
```diff
@@ -42,7 +42,10 @@ def submit_questionnaire(self):
         self._store_submitted_time_and_display_address_in_session()
         submission_schema: Mapping = self._schema.get_submission() or {}
 
-        if not submission_schema.get("view_submitted_response"):
+        if submission_schema.get("view_submitted_response"):
+            self._questionnaire_store.submitted_at = datetime.now(tz=tzutc())
```
MebinAbraham, line 46: This shouldn't be generated here as it won't always match the `submitted_at` value in the payload. So need to either generate the datetime upfront and pass to convert answer and this or just reuse the value from the payload here.

app/views/handlers/submission.py
```diff
@@ -1,6 +1,7 @@
 from datetime import datetime
```
MebinAbraham, line 1: - A test needs to be added to ensure that on submission the `submitted_at` property is indeed populated.
- Another test should to be added to ensure that the submitted time in the payload and questionnaire store matches.

MebinAbraham (commented):
Property is populated and indexed as expected. Just a couple of tweaks needed.

app/views/handlers/submission.py
LJBabbage: Does it need to? There is no requirement to find submitted by the payload. In fact this was purely being used for deletion after 24 hours so we would never be in a position to find it in time anyway

app/views/handlers/submission.py
LJBabbage: > test_store_and_get_submitted_at. does this
> as mentioned above there is no need to connect the 2

app/views/handlers/submission.py
LJBabbage: However I will change it

app/views/handlers/submission.py
```diff
@@ -57,5 +67,5 @@ def get_payload(self):
     def _store_submitted_time_and_display_address_in_session(self):
         session_data = self._session_store.session_data
         session_data.display_address = self._metadata.get("display_address")
-        session_data.submitted_time = datetime.utcnow().isoformat()
+        session_data.submitted_time = self.submitted_at.isoformat()
```
MebinAbraham, line 70: As discussed, we should look to drop `submitted_time` from session data and use the `submitted_at` from the questionnaire state directly.

app/views/handlers/submission.py
MebinAbraham: Ya `test_submit_view_submitted_response_true_submitted_at_set` covers that it is populated. I missed that. But we should test that values are aligned now that they are.

LJBabbage: I reckon that's not needed, its create in one place and no transformation is done to it, so it's not going to be different. It's an argument you could use for any var across are app if used more than once

app/routes/questionnaire.py
```diff
@@ -57,7 +57,7 @@ def before_questionnaire_request():
 
     metadata = get_metadata(current_user)
     if not metadata:
-        raise NoQuestionnaireStateException(401)
+        raise NoQuestionnaireStateException(401)  # pragma: no cover
```
LJBabbage, line 60: As discussed this has been removed until issue of state deletion are fixed

tests/integration/session/test_multiple_login.py
```diff
@@ -73,6 +75,7 @@ def setUp(self):
 
         self.cache = {self.client_a: {}, self.client_b: {}}
 
+    @pytest.mark.xfail
```
LJBabbage, line 78: As above, because state is still there and router has a bug which will redirect to submission page, this has been marked as failing. Will be fixed in another card

app/routes/questionnaire.py
```diff
@@ -77,13 +77,14 @@ def before_questionnaire_request():
...
+@with_questionnaire_store
+def before_post_submission_request(questionnaire_store):
     if request.method == "OPTIONS":
         return None
-
     session_store = get_session_store()
     session_data = session_store.session_data
-    if not session_data.submitted_time:
+
+    if not questionnaire_store.submitted_at:
```
MebinAbraham, line 87: The decorator shouldn't be used here, it needs to be fetched using `get_questionnaire_store` after the `OPTIONS` check. If you make an `OPTIONS` requests without a session, this will generate a 500.

Also, the call to `get_session_store()` should be moved below the `submitted_at` check and only called when needed.

app/routes/questionnaire.py
```diff
@@ -310,9 +311,10 @@ def relationships(
 
 @post_submission_blueprint.route("thank-you/", methods=["GET", "POST"])
 @with_session_store
+@with_questionnaire_store
 @with_schema
-def get_thank_you(schema, session_store):
-    thank_you = ThankYou(schema, session_store)
+def get_thank_you(schema, questionnaire_store, session_store):
+    thank_you = ThankYou(schema, session_store, questionnaire_store.submitted_at)
```
MebinAbraham, line 317: Would be more readable if the order of the arguments for the route matched Thank You so:
```
@with_questionnaire_store
@with_session_store
@with_schema
def get_thank_you(schema, session_store, questionnaire_store):
    thank_you = ThankYou(schema, session_store, questionnaire_store.submitted_at)
```

tests/app/storage/test_encrypted_questionnaire_storage.py
```diff
@@ -37,7 +41,19 @@ def test_store_and_get(self):
...
+    def test_store_and_get_submitted_at(self):
+        user_id = "1"
+        user_ik = "2"
+        encrypted = EncryptedQuestionnaireStorage(user_id, user_ik, "pepper")
+        data = "test"
+        encrypted.save(data, datetime.now(tz=tzutc()))
+        self.assertEqual(
+            ("test", QuestionnaireStore.LATEST_VERSION, datetime.now(tz=tzutc())),
+            encrypted.get_user_data(),
         )
```
MebinAbraham, line 57: The vars don't provide any more value over keyword args especially since they are not reused. Plus, I think it makes it more readable.
```py
    def test_store_and_get_submitted_at(self):
        encrypted = EncryptedQuestionnaireStorage(
            user_id="1", user_ik="2", pepper="pepper"
        )
        encrypted.save(data="test", submitted_at=datetime.now(tz=tzutc()))

        self.assertEqual(
            ("test", QuestionnaireStore.LATEST_VERSION, datetime.now(tz=tzutc())),
            encrypted.get_user_data(),
        )
```

I think the test above should be cleaned up the same way and I think the tests need to be renamed to maybe something like:
```
- test_store_and_get_without_submiited_at
- test_store_and_get_with_submiited_at
```

tests/integration/questionnaire/test_questionnaire_endpoints.py
```diff
@@ -73,20 +73,6 @@ def test_get_thank_you_raises_404_when_questionnaire_is_not_complete(self):
         # Then we are shown a 404 page
         self.assertStatusNotFound()
 
-    def test_data_is_deleted_on_submission(self):
```
MebinAbraham, line 76: This could check data is retained on submission.

app/submitter/converter.py
```diff
@@ -52,6 +52,7 @@ def convert_answers(schema, questionnaire_store, routing_path, flushed=False):
         schema: QuestionnaireSchema instance with populated schema json
         questionnaire_store: EncryptedQuestionnaireStorage instance for accessing current questionnaire data
         routing_path: The full routing path followed by the user when answering the questionnaire
+        submitted_at: The iso date/time that the questionnaire is submitted
```
MebinAbraham, line 55:
```suggestion
        submitted_at: The date and time in ISO 8601 of the submission
```

app/views/handlers/submission.py
MebinAbraham: > its create in one place and no transformation is done to it

How can we know without a test?. At the moment, it relies on cached property, if someone changed it to `propery` in future, it wouldn't necessarily match and no tests would break. Happy to let it go since it is not an official requirement.

tests/integration/questionnaire/test_questionnaire_endpoints.py
LJBabbage: I don't think we need to test something hasn't been deleted and is still there

tests/app/storage/test_datastore.py
```diff
@@ -23,7 +23,7 @@ def setUp(self):
         self.ds = Datastore(self.mock_client)
 
     def test_get_by_key(self):
-        model = QuestionnaireState("someuser", "data", 1)
+        model = QuestionnaireState("someuser", "data", 1, None)
```
ajmaddaford, line 26: I think it would make sense if `QuestionnaireState` could be initialised without a `submitted_at` i.e. default to `None`.

tests/app/storage/test_encrypted_questionnaire_storage.py
```diff
@@ -29,15 +33,26 @@ def test_encrypted_storage_requires_user_ik(self):
...
+    @freeze_time(datetime.now(tzutc()).replace(second=0, microsecond=0))
+    def test_store_and_get_with_submitted_at(self):
+
+        encrypted = EncryptedQuestionnaireStorage(
+            user_id="1", user_ik="2", pepper="pepper"
+        )
+        encrypted.save(data="test", submitted_at=datetime.now(tz=tzutc()))
+        self.assertEqual(
+            ("test", QuestionnaireStore.LATEST_VERSION, datetime.now(tz=tzutc())),
+            encrypted.get_user_data(),
```
ajmaddaford, line 55: `freeze_time` seems unnecessary given that we're setting and getting the time.
```suggestion
    def test_store_and_get_with_submitted_at(self):
		now = datetime.now(timezone.utc)
        encrypted = EncryptedQuestionnaireStorage(
            user_id="1", user_ik="2", pepper="pepper"
        )
        encrypted.save(data="test", submitted_at=now)
        
        self.assertEqual(
            ("test", QuestionnaireStore.LATEST_VERSION, now),
            encrypted.get_user_data(),
```

tests/app/test_jinja_filters.py
```diff
@@ -90,6 +92,20 @@ def test_format_date_time_in_gmt(self):
...
+        loc_dt = london_timezone.localize(datetime(2021, 8, 3, 00, 15, 0))
+
+        # When
+        with self.app_request_context("/"):
+            format_value = format_datetime(self.autoescape_context, loc_dt)
+
+        # Then
+        self.assertEqual(
+            format_value, "<span class='date'>3 August 2021 at 00:15</span>"
+        )
```
MebinAbraham, line 107: This test is incorrect as it is passing a London date time to format datetime instead of UTC datetime. Since the existing two tests have the date in the output, we should:
- update those to test date **and** time
- The date we pass should be UTC since that is the actual behaviour. Calling `datetime()` would not have timezone info by default.
- For testing GMT, I think it makes sense to do two dates, as I found it confusing without.

We should update the existing tests to something like this, which test both the date and time in terms of BST/GMT.
```py
    def test_format_date_time_in_bst(self):
        # Given a date after BST started
        date_time = datetime(2018, 3, 29, 23, 59, 0).replace(tzinfo=timezone.utc)

        # When
        with self.app_request_context("/"):
            format_value = format_datetime(self.autoescape_context, date_time)

        # Then
        self.assertEqual(
            format_value, "<span class='date'>30 March 2018 at 00:59</span>"
        )

    def test_format_date_time_in_gmt(self):
        # Given
        test_data = {
            datetime(2018, 10, 28, 00, 15, 0): "28 October 2018 at 01:15",
            # Clocks go back on 29th Oct 2018
            datetime(2018, 10, 29, 00, 15, 0): "29 October 2018 at 00:15"
        }
        for date_time, expected_value in test_data.items():
            with self.subTest(
                date_time=date_time, expected_valye=expected_value,
            ):
                date_time = date_time.replace(tzinfo=timezone.utc)
                # When
                with self.app_request_context("/"):
                    format_value = format_datetime(self.autoescape_context, date_time)

                # Then
                self.assertEqual(
                    format_value, f"<span class='date'>{expected_value}</span>"
                )
```

tests/app/views/contexts/test_completed_metadata_context.py
```diff
@@ -1,10 +1,18 @@
...
+
 import pytest
 
 from app.data_models.session_data import SessionData
 from app.views.contexts.thank_you_context import build_default_thank_you_context
 
 
+@cache
+def submitted_at():
+    return datetime.now(timezone.utc)
```
MebinAbraham, line 13: This should be a constant, like the other tests.

tests/app/views/handlers/test_confirmation_email_fulfilment_request.py
```diff
@@ -26,7 +26,7 @@ def session_data():
         survey_url=None,
         ru_name=None,
         ru_ref=None,
-        submitted_time=datetime.now(tzutc()).isoformat(),
+        submitted_at=datetime.now(tzutc()).isoformat(),
```
ajmaddaford, line 29: The code used to get the current time should be consistent throughout i.e. `datetime.now(timezone.utc)`.

tests/app/test_jinja_filters.py
```diff
@@ -90,6 +92,20 @@ def test_format_date_time_in_gmt(self):
             format_value, "<span class='date'>28 October 2018 at 11:59</span>"
         )
 
+    def test_format_date_using_london_timezone(self):
+        # Given a London timezone
+        london_timezone = timezone("Europe/London")
+        loc_dt = london_timezone.localize(datetime(2021, 8, 3, 00, 15, 0))
```
ajmaddaford, line 98: Variable names shouldn't be shortened, it's also clear from reading the test that it will be a localised datetime, so this could just be named something like `date_time`.

app/data_models/app_models.py
```diff
@@ -56,12 +57,14 @@ class QuestionnaireStateSchema(Schema, DateTimeSchemaMixin):
...
     version = fields.Integer()
+    submitted_at = fields.DateTime(allow_none=True)
 
     @post_load
     def make_model(self, data, **kwargs):
         created_at = data.pop("created_at", None)
         updated_at = data.pop("updated_at", None)
-        model = QuestionnaireState(**data)
+        submitted_at = data.pop("submitted_at", None)
+        model = QuestionnaireState(**data, submitted_at=submitted_at)
```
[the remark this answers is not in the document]
LJBabbage: Good, question. I wondered that myself when I looked at it again

tests/app/test_jinja_filters.py
LJBabbage: I lifted this from their docs :laughing:

ajmaddaford (approved):

MebinAbraham (approved):

berroar (approved):

PR #633, opened 2021-07-29
 Adds type hints for the authentication module

berroar:

Adds type hints for the `authentication` module.

### How to review 

Check the type hints added are correct

### Checklist


berroar:
[Introduce Type Hints to Authentication module (2 SP)](https://trello.com/c/XFSXwbRF/4799-introduce-type-hints-to-authentication-module-2-sp)

mypy.ini
```diff
@@ -21,3 +21,31 @@ warn_return_any = True
...
+disallow_untyped_defs = True
+warn_return_any = True
+
+[mypy-app.authentication.user]
+disallow_untyped_defs = True
+warn_return_any = True
+
+[mypy-app.authentication.user_id_generator]
+disallow_untyped_defs = True
+warn_return_any = True
```
MebinAbraham, line 51: When doing the entire module, this should just be:
```suggestion
[mypy-app.authentication.*]
disallow_untyped_defs = True
warn_return_any = True

```

Any file that can't be done, can be ignored at a file level.

app/authentication/authenticator.py
```diff
@@ -22,27 +24,31 @@
 
 
 @login_manager.user_loader
-def user_loader(user_id):
+def user_loader(user_id: str) -> Union[str, None]:
```
MebinAbraham, line 27: If something can be `None` and one other type, then we should use `Optional` so `Optional[str]`.
```suggestion
def user_loader(user_id: str) -> Optional[str]:
```

app/authentication/authenticator.py
```diff
@@ -22,27 +24,31 @@
...
     logger.debug("loading user", user_id=user_id)
     return load_user()
 
 
 @login_manager.request_loader
-def request_load_user(request):  # pylint: disable=unused-argument
+def request_load_user(
+    request: Any,
```
MebinAbraham, line 34: We need to minimise the use of `Any`. It should not be used unless it can't be avoided. Here this is a flask request so `Request` from flask.

app/authentication/authenticator.py
```diff
@@ -22,27 +24,31 @@
...
+) -> Optional[User]:  # pylint: disable=unused-argument
     logger.debug("load user")
     return load_user()
 
 
 @user_logged_out.connect_via(ANY)
-def when_user_logged_out(sender_app, user):  # pylint: disable=unused-argument
+def when_user_logged_out(
+    sender_app: Any, user: str
```
MebinAbraham, line 42: This is an instance of the flask app so `Flask`.

app/authentication/authenticator.py
```diff
@@ -160,14 +163,14 @@ def store_session(metadata):
...
         raise NoTokenException("Please provide a token")
 
     logger.debug("decrypting token")
     decrypted_token = decrypt(
         token=encrypted_token,
-        key_store=current_app.eq["key_store"],
+        key_store=current_app.eq["key_store"],  # type: ignore
         key_purpose=KEY_PURPOSE_AUTHENTICATION,
         leeway=current_app.config["EQ_JWT_LEEWAY_IN_SECONDS"],
     )
```
MebinAbraham, line 176: The type of decrypted token is a dict. When MyPy says expected some type but got `Any` that most times means the method you called doesn't have type hints, i.e decrypt in our case, but for those, you can specify the type inline. So this would become:

```py
def decrypt_token(encrypted_token: str) -> dict[str, Any]:
    if not encrypted_token:
        raise NoTokenException("Please provide a token")

    logger.debug("decrypting token")
    decrypted_token: dict[str, Any] = decrypt(
        token=encrypted_token,
        key_store=current_app.eq["key_store"],  # type: ignore
        key_purpose=KEY_PURPOSE_AUTHENTICATION,
        leeway=current_app.config["EQ_JWT_LEEWAY_IN_SECONDS"],
    )

    logger.debug("token decrypted")
    return decrypted_token
```

app/authentication/no_questionnaire_state_exception.py
```diff
@@ -1,7 +1,10 @@
...
 class NoQuestionnaireStateException(Exception):
-    def __init__(self, value):
+    def __init__(self, value: Union[str, int]) -> None:
         super().__init__()
         self.value = value
 
-    def __str__(self):
-        return self.value
+    def __str__(self) -> str:
+        return self.value  # type: ignore
```
MebinAbraham, line 10: We always raise this with an int (status code), but `str` comes from the fact that this inherits from base exception. The type for `__str__` just needs to be casted, then you don't need to ignore it.

```suggestion
    def __str__(self) -> str:
        return str(self.value)
```

app/authentication/no_token_exception.py
```diff
@@ -1,7 +1,10 @@
...
 class NoTokenException(Exception):
-    def __init__(self, value):
+    def __init__(self, value: Union[str, int]) -> None:
         super().__init__()
         self.value = value
 
-    def __str__(self):
-        return self.value
+    def __str__(self) -> str:
+        return self.value  # type: ignore
```
MebinAbraham, line 10: Same comment as the one for `NoQuestionnaireStateException`.

app/authentication/roles.py
```diff
@@ -1,12 +1,13 @@
...
+from typing import Any
 
 from flask_login import current_user
 from werkzeug.exceptions import Forbidden
 
 from app.globals import get_metadata
 
 
-def role_required(role):
+def role_required(role: str) -> Any:
```
MebinAbraham, line 10: Decorators don't have a great support for typing. They should get better in 3.10.
But for now, this can be a little more specific.
```suggestion
def role_required(role: str) -> Callable:
```

app/authentication/roles.py
```diff
@@ -25,11 +26,11 @@ def dump():
     :param role: The role required by the function being decorated
     """
 
-    def role_required_decorator(func):
+    def role_required_decorator(func: Any) -> Any:
```
MebinAbraham, line 29:
```suggestion
    def role_required_decorator(func: Callable) -> Callable:
```

MebinAbraham (commented):
Only the `cookie_session.py` is now missing type.

app/authentication/cookie_session.py
```diff
@@ -5,5 +5,5 @@
 
 class SHA256SecureCookieSessionInterface(SecureCookieSessionInterface):
     @staticmethod
-    def digest_method():
+    def digest_method():  # type: ignore
```
MebinAbraham, line 8: Is it not possible to type this?

app/authentication/authenticator.py
```diff
@@ -160,14 +163,14 @@ def store_session(metadata):
     logger.info("user authenticated")
 
 
-def decrypt_token(encrypted_token):
+def decrypt_token(encrypted_token: str) -> dict[str, Any]:
```
LJBabbage, line 166: I wonder if this could be a bit tighter, I had a quick look at what is returned with a decrypted_token and it only looks like str, int and list

app/authentication/user.py
```diff
@@ -1,8 +1,10 @@
+from typing import Any, Optional
+
 from flask_login import UserMixin
 
 
 class User(UserMixin):
-    def __init__(self, user_id, user_ik):
+    def __init__(self, user_id: str, user_ik: Optional[Any]) -> None:
```
LJBabbage, line 7: Is user_ik not a str?

berroar: Yeah I initially had it set as `str` but it generates the following mypy error `"User" has incompatible type "Optional[Any]"; expected "str"`.

LJBabbage: Mine works with Optional[str], maybe need someone else to try

berroar: It works for me so have updated it

MebinAbraham:
The only thing I can think of that we prob need to sort is, where possible argument types should be abstract and return types should be explicit, as that is what we have defined in the style guide. For example in `store_session`, the metadata should be a `Mapping` instead of `dict`.

LJBabbage (approved):

app/authentication/authenticator.py
```diff
@@ -160,14 +165,14 @@ def store_session(metadata):
...
 
 
-def decrypt_token(encrypted_token):
+def decrypt_token(encrypted_token: str) -> dict[str, Union[str, list, int]]:
     if not encrypted_token:
         raise NoTokenException("Please provide a token")
 
     logger.debug("decrypting token")
-    decrypted_token = decrypt(
+    decrypted_token: dict[str, Any] = decrypt(
```
MebinAbraham, line 173: This type hint needs to match the return type of the function.

app/authentication/user.py
```diff
@@ -1,8 +1,10 @@
+from typing import Optional
+
 from flask_login import UserMixin
 
 
 class User(UserMixin):
-    def __init__(self, user_id, user_ik):
+    def __init__(self, user_id: str, user_ik: Optional[str]) -> None:
```
MebinAbraham, line 7: `user_id` is also optional as `session_store.user_id` default value is `None` and the check below also covers against it.

MebinAbraham (approved):

PR #640, opened 2021-08-17
Display answer labels when using multiple answers

MebinAbraham:
- Updates the multiple answer's schema to demonstrate a question with multiple answers while ensure the labels are displayed as expected.
- Ensures that the label/legend is displayed when using multiple answers. Some exception to this:
  - Checkbox label has not been done as it need tweaking which will be done in https://trello.com/c/D9VWhami/4838-add-instruction-property-to-checkbox-answers
  - Year only date and single duration answer labels have not been checked as it need tweaks in the DS which Matt is aware of. At the moment, they are not consistent and do not use the proper HTML label element. Those will need to be tested once the changes have been made in the DS.

### How to review 
- Ensure when a question has multiple answers that the label is rendered unless:
  - the answer is a checkbox with a single option
  - of type `YearDate`
  - of type single `Duration` 

### Checklist


LJBabbage (commented):
Is this a new pattern, having multiple answers in a question? I haven't seen this for a long time and was at least expecting multiple questions in a block if we were doing that.

MebinAbraham:
As discussed, the pattern of having multiple answers already exists in runner, however it has not been documented in the DS about the allowed combinations and its scenarios. This schema was updated with the different types to demonstrate that answers are shown on the summary as expected when using multiple answers, and that answers use the correct label/legend. The single question was used since there is no doc around the allowed combinations.

tests/functional/spec/multiple_answers.spec.js
```diff
@@ -1,23 +1,87 @@
...
-    $(PersonalDetailPage.surname()).setValue("SOLO");
-    $(PersonalDetailPage.submit()).click();
-    expect($(SubmitPage.firstNameAnswer()).getText()).to.contain("HAN");
-    expect($(SubmitPage.surnameAnswer()).getText()).to.contain("SOLO");
+  describe("Given I have launched a questionnaire that has multiple answers per question", () => {
+    beforeEach("Load the questionnaire", () => {
+      browser.openQuestionnaire("test_multiple_answers.json");
+    });
+
+    it("When I am on the question page, Then all answers should have a label/legend", () => {
```
LJBabbage, line 71: This is strictly not quite right at the moment as checkboxAnswer doesn't. It might in the future, but we don't really want to get in the habit of doing that.

LJBabbage (changes requested):

tests/functional/spec/multiple_answers.spec.js
```diff
@@ -1,23 +1,87 @@
...
+      expect($(SubmitPage.numberAnswer()).getText()).to.equal("5");
+      expect($(SubmitPage.percentageAnswer()).getText()).to.equal("3%");
+      expect($(SubmitPage.mobileNumberAnswer()).getText()).to.equal("07700900111");
+      expect($(SubmitPage.textareaAnswer()).getText()).to.equal("Fuel type petrol");
+
+      expect($(SubmitPage.ageAnswer()).getText()).to.equal("10");
+      expect($(SubmitPage.ageEstimateAnswer()).getText()).to.equal("This age is an estimate");
+    });
+
+    it("When I click 'Change' an answer, Then I should be taken to the correct page and be able to edit my answer", () => {
```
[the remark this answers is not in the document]
MebinAbraham: This now mentions that it should be in focus.

tests/functional/spec/multiple_answers.spec.js
MebinAbraham: Added a comment and also added it to the card which will fix checkbox legend.

LJBabbage (approved):

ajmaddaford (approved):

PR #641, opened 2021-08-18
Update submitted_at to timestamp

LJBabbage:
After issues with purging questionnaire state using datetime this PR changes submitted_at to a timestamp. As our current timestamp doesn't use milliseconds and submitted_at is also included in the payload it was decided to remove  the milliseconds on submission.py

### How to review 
Check the change works, locally or on GCP if you like (I did do both)

### Checklist

### DIFF
--- a/app/data_models/app_models.py
+++ b/app/data_models/app_models.py
@@ -57,7 +57,7 @@ class QuestionnaireStateSchema(Schema, DateTimeSchemaMixin):
     user_id = fields.Str()
     state_data = fields.Str()
     version = fields.Integer()
-    submitted_at = fields.DateTime(allow_none=True)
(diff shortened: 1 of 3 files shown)


MebinAbraham (approved):
Seems to be as expected, but have now realised the query we were doing for the purge will need to also now check the values are not null otherwise we will purge everything.

LJBabbage:
> will need to also now check the values are not null otherwise we will purge everything.

That was always the case

MebinAbraham:
> > will need to also now check the values are not null otherwise we will purge everything.
> 
> That was always the case

Yup, but that was not done in Rhys' changes.

petechd (approved):

PR #643, opened 2021-08-19
Add new rules engine for routing rules (equal & not equal)

petechd:
This adds evaluation of `equal` and `not equal` rules from the new routing rules engine introduced in [this PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/629). It makes use of new `WhenRuleEvaluator` data class. Relevant routing schemas were added along with section end test schema. New "when" rule evaluator was added to `path_finder.py`  rather than `rules.py` to make deprecation of it easier.

### How to review 
Run new integration tests, check new schemas.

### Checklist


petechd:
[Use new rules engine for routing rules (3 SP)](https://trello.com/c/gBZNSyam/4829-use-new-rules-engine-for-routing-rules-3-sp)

app/questionnaire/path_finder.py
```diff
@@ -134,35 +135,56 @@ def _build_routing_path_block_ids(self, blocks, current_location):
...
+                if should_goto:
+                    if rule.get("section") == "End":
+                        return None
+                    next_block_id = rule["block"]
+                    next_block_index = PathFinder._block_index_for_block_id(
+                        blocks, next_block_id
+                    )
 
-                return next_block_index
+                    return next_block_index
```
ajmaddaford, line 187: The logic within `should_goto` needs to be the same for both. I think we can handle both in the same loop with something like:
```python
        for rule in routing_rules:
            if "goto" in rule:
                rule = rule["goto"]
                should_goto = evaluate_goto(
                    rule,
                    self.schema,
                    self.metadata,
                    self.answer_store,
                    self.list_store,
                    current_location=this_location,
                )
            else:
                should_goto = should_goto_new(
                    rule,
                    self.schema,
                    self.answer_store,
                    self.list_store,
                    self.metadata,
                    this_location,
                    routing_path_block_ids,
                )

            if should_goto:
                if rule.get("section") == "End":
                    return None

                next_block_id = self._get_next_block_id(rule)
                next_block_index = PathFinder._block_index_for_block_id(
                    blocks, next_block_id
                )
                next_precedes_current = (
                    next_block_index is not None and next_block_index < block_index
                )

                if next_precedes_current:
                    self._remove_rule_answers(rule, this_location)
                    routing_path_block_ids.append(next_block_id)
                    return None

                return next_block_index
```

app/questionnaire/path_finder.py
```diff
@@ -178,3 +200,23 @@ def _remove_rule_answers(self, goto_rule, this_location):
...
+    rule, schema, answer_store, list_store, metadata, location, routing_path_block_ids
+):
+    when_rule_evaluator = WhenRuleEvaluator(
+        schema,
+        answer_store,
+        list_store,
+        metadata,
+        location,
+        routing_path_block_ids,
+    )
```
ajmaddaford, line 215: The `WhenRuleEvaluator` should be initialised once in the init method.

app/questionnaire/path_finder.py
```diff
@@ -178,3 +200,23 @@ def _remove_rule_answers(self, goto_rule, this_location):
...
+    if when_rule := rule.get("when", None):
+        operator = list(when_rule.keys())[0]
+        return when_rule_evaluator.evaluate(
+            rule={operator: [when_rule[operator][1], when_rule[operator][0]]}
+        )
```
ajmaddaford, line 220: This shouldn't be necessary, it should work by passing the rule directly to the when rule evaluator i.e. `when_rule_evaluator.evaluate(when_rule)`. That doesn't work because when we deserialise schemas we convert lists to Tuples and there is a type check in the `evaluate` method that expects a list. That type check should allow a list or Tuple.

app/questionnaire/path_finder.py
```diff
@@ -178,3 +200,23 @@ def _remove_rule_answers(self, goto_rule, this_location):
...
+):
+    when_rule_evaluator = WhenRuleEvaluator(
+        schema,
+        answer_store,
+        list_store,
+        metadata,
+        location,
+        routing_path_block_ids,
+    )
+    if when_rule := rule.get("when", None):
```
ajmaddaford, line 216:
```suggestion
    if when_rule := rule.get("when"):
```

test_schemas/en/test_new_routing_to_section_end.json
```diff
@@ -0,0 +1,200 @@
+{
```
ajmaddaford, line 1: If the other new schemas that have been added route to the section end, why is this schema needed?

app/questionnaire/path_finder.py
petechd: This one is done now, `evaluate` allows tuple and only `when_rule` is passed, thanks!

app/questionnaire/path_finder.py
petechd: I moved it to top level of `_evaluate_routing_rules` method because `__init__` of PathFinder doesn't have access to arguments like `location` and `routing_path_block_ids`. Consulted this with @MebinAbraham, he said to keep it simple for now.

app/questionnaire/path_finder.py
petechd: All this above added now so it's the same loop and same code reused. Thanks!

test_schemas/en/test_new_routing_to_section_end.json
petechd: That was needed when loop existed, after adding changes from your comments this is redundant and removed, thanks!

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -33,11 +33,13 @@ def __post_init__(self) -> None:
             use_default_answer=True,
         )
 
-    def _evaluate(self, rule: dict[str, list]) -> Union[bool, Optional[date]]:
+    def _evaluate(
+        self, rule: dict[str, Union[list, tuple]]
+    ) -> Union[bool, Optional[date]]:
```
MebinAbraham, line 38: Opening it up to list/tuple muddies the type check, I think we should update it to allow any `Sequence` then list/tuple would work fine.
```suggestion
    def _evaluate(self, rule: dict[str, Sequence]) -> Union[bool, Optional[date]]:
```

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -33,11 +33,13 @@ def __post_init__(self) -> None:
...
         operator = Operator(next(iter(rule)))
         operands = rule[operator.name]
 
-        if not isinstance(operands, list):
+        if not isinstance(operands, list) and not isinstance(operands, tuple):
             raise TypeError(
                 f"The rule is invalid, operands should be of type list and not {type(operands)}"
             )
```
MebinAbraham, line 45:
```suggestion
        if not isinstance(operands, Sequence):
            raise TypeError(
                f"The rule is invalid, operands should be of type Sequence and not {type(operands)}"
            )
```

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -46,7 +48,7 @@ def _evaluate(self, rule: dict[str, list]) -> Union[bool, Optional[date]]:
         return operator.evaluate(resolved_operands)
 
     def get_resolved_operands(
-        self, operands: list[ValueSourceTypes]
+        self, operands: Union[list[ValueSourceTypes], tuple]
```
MebinAbraham, line 51:
```suggestion
        self, operands: Sequence[ValueSourceTypes]
```

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -58,5 +60,7 @@ def get_resolved_operands(
             else:
                 yield operand
 
-    def evaluate(self, rule: dict[str, list]) -> Union[bool, Optional[date]]:
+    def evaluate(
+        self, rule: dict[str, Union[list, tuple]]
+    ) -> Union[bool, Optional[date]]:
```
MebinAbraham, line 65:
```suggestion
    def evaluate(self, rule: dict[str, Sequence]) -> Union[bool, Optional[date]]:
```

test_schemas/en/test_new_routing_number_equals.json
```diff
@@ -0,0 +1,129 @@
...
+                        {
+                            "type": "Question",
+                            "id": "number-question",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "answer",
+                                        "mandatory": true,
+                                        "type": "Number",
+                                        "label": "123"
```
MebinAbraham, line 47: I know it is copied from the existing schema, but this label provides no value.

MebinAbraham (commented):
Looks good, just a few minor comments.

MebinAbraham (commented):
Other thing to consider is that the existing routing schemas have functional tests.

test_schemas/en/test_new_routing_number_equals.json
petechd: I changed it to "Enter 123", cannot be removed as it's required by answer type number..

petechd:
> Other thing to consider is that the existing routing schemas have functional tests.

Should I make functional tests use new schemas, build duplicates of these tests or keep them as they are and wait till we remove old engine?

MebinAbraham (approved):
We can discuss the functional tests for these routing rules in the tech session. Probably makes sense to introduce them in groups and speed up the introduction of each operator in schema.

ajmaddaford (approved):

PR #639, opened 2021-08-12
Add submitted response page

LJBabbage:
### What is the context of this PR?
This PR introduces the page view-response, which allows a respondent to view their answers after submission.

N.B Some elements of the EPIC have still not been fully designed. I have made comments in the code where I haven't been able to fully implemented or have had to put in a temporary fix

### How to review 
Using test_view_submitted_response.json complete the questionnaire and submit, then in the browser go to /submitted/view-response and make sure the page is displaying as expected

### Checklist

* [ ] New static content marked up for translation
* [ ] Newly defined schema content included in eq-translations repo

PR #644, opened 2021-08-19
Update and add a test to make sure after submission a respondent is redirect to thank you

LJBabbage:
When we stopped deleting QuestionnaireState, we created an issue when a survey was submitted, it could be submitted again, if by the same respondent or another. Work on other stories has corrected the issue, however we still had a test mark as failing that needed updating and could do with another test adding

### How to review 
Make sure the tests make sense. You can also test the behaviour by taking note of the Response ID on launcher, completing and submitting, then launching again with the same Response ID. It should take you to thank you page the 2nd time. 

For multiple users I used 2 different browsers. I took note of the Response ID, launched the questionnaire in one browser and answered the first question, I then launched the same questionnaire with the same Response ID in a different browser and completed the questionnaire and submitted. At this point I reverted back to the first browser and tried to submit, it should take you to the thank you page

### Checklist


MebinAbraham (changes requested):
The IR route is missing the check, hence it allows you to access after submission.

MebinAbraham:
While we are here, we should also look to refactor the `get_questionnaire_store` method as per https://github.com/ONSdigital/eq-questionnaire-runner/pull/639#discussion_r690497241

PR #645, opened 2021-08-20
Add lint rule for Python UTC date objects

petechd:

### How to review 

### Checklist


petechd:
[Add lint rule for Python UTC date objects (1 SP)](https://trello.com/c/cARKKgcD/4815-add-lint-rule-for-python-utc-date-objects-1-sp)

tox.ini
```diff
@@ -2,4 +2,4 @@
 # Ignore node_modules and cloned repos when not in a virtual environment
 exclude = node_modules/*,tests/*,src/*
 max-line-length = 160
-ignore = C815,C816,W503,E203
+ignore = C815,C816,W503,E203,DTZ007,DTZ005,DTZ004
```
ajmaddaford, line 5: Is there any reason to disable the these rules? The changes required when they're enabled look minimal.

petechd: I left it out but can fix them, changes are minimal as you said but didn't want to jump the gun, so I waited for reviewer's opinion first.

ajmaddaford: That fair enough. I think we should fix them, as they relate to similar recommendations in the Python docs.

petechd: Ok, added all recommendations now, probably they can still be refactored slightly. Thanks!

app/data_models/app_models.py
```diff
@@ -19,28 +18,30 @@ def __init__(self, eq_session_id, user_id, expires_at, session_data):
...
-            return int(value.replace(tzinfo=tzutc()).strftime("%s"))
+            return int(datetime.timestamp(value))
 
     def _deserialize(self, value, attr, data, **kwargs):
         if value:
             # Timestamp to timezone aware datetime
-            return datetime.utcfromtimestamp(value).replace(tzinfo=tzutc())
+            return datetime.fromtimestamp(value, tz=timezone.utc).replace(
+                tzinfo=timezone.utc
+            )
```
MebinAbraham, line 44: You don't need the `replace` since the `tz` is set on initialisation now.

app/forms/questionnaire_form.py
```diff
@@ -255,7 +255,7 @@ def _has_min_and_max_single_dates(date_from, date_to):
 
     @staticmethod
     def _get_offset_value(period_object):
-        now = datetime.now()
+        now = datetime.now(tz=timezone.utc)
```
MebinAbraham, line 258: Need to check whether this was done on purpose so that when do offset it uses system locale. We prob want to pass London timezone in if that is the case. For routing offset it uses `utc` but it also questions whether it should. Worth checking with Donna.

app/forms/validators.py
```diff
@@ -232,11 +232,11 @@ def __call__(self, form, field):
 
         try:
             if hasattr(form, "day"):
-                datetime.strptime(form.data, "%Y-%m-%d")
+                datetime.strptime(form.data, "%Y-%m-%d").replace(tzinfo=timezone.utc)
```
MebinAbraham, line 235: It might be worth making a helper method for this as it is done in a lot of places. Thoughts?

app/routes/session.py
```diff
@@ -99,10 +98,10 @@ def login():
 
 
 def validate_jti(decrypted_token):
-    expires_at = datetime.utcfromtimestamp(decrypted_token["exp"]).replace(
-        tzinfo=tzutc()
-    )
-    jwt_expired = expires_at < datetime.now(tz=tzutc())
+    expires_at = datetime.fromtimestamp(
+        decrypted_token["exp"], tz=timezone.utc
+    ).replace(tzinfo=timezone.utc)
```
MebinAbraham, line 103: Again, shouldn't need to replace when you pass in `tz` already.

app/views/handlers/block.py
```diff
@@ -120,7 +120,7 @@ def _update_section_completeness(
     def _set_started_at_metadata(self):
         response_metadata = self._questionnaire_store.response_metadata
         if not response_metadata.get("started_at"):
-            started_at = datetime.utcnow().isoformat()
+            started_at = datetime.now(timezone.utc).replace(tzinfo=None).isoformat()
```
MebinAbraham, line 123: Replace can be removed.

MebinAbraham (commented):
There are still timezone naive usages of datetime in tests, we should look to see why they are not picked up. It doesn't seem to be caused by https://trello.com/c/zYZWNuS1/4374-fix-python-linting-to-lint-all-files

app/data_models/app_models.py
```diff
@@ -19,28 +18,30 @@ def __init__(self, eq_session_id, user_id, expires_at, session_data):
...
+        self.expires_at = expires_at.replace(tzinfo=timezone.utc)
 
 
 # pylint: disable=no-self-use
 class Timestamp(fields.Field):
     def _serialize(self, value, attr, obj, **kwargs):
         if value:
             # Timezone aware datetime to timestamp
-            return int(value.replace(tzinfo=tzutc()).strftime("%s"))
+            return int(datetime.timestamp(value))
```
MebinAbraham, line 37: How come the timezone was dropped?

petechd:
> There are still timezone naive usages of datetime in tests, we should look to see why they are not picked up. It doesn't seem to be caused by https://trello.com/c/zYZWNuS1/4374-fix-python-linting-to-lint-all-files

They were not picked up because `tox.ini` file exluded them from flake8 checks. I fixed it now and got rid of naive timezone usages in test files.

app/data_models/app_models.py
petechd: Removed, thanks!

app/routes/session.py
petechd: Removed, thanks!

app/views/handlers/block.py
petechd: I think it needs to stay as it removes "UTC offset to corresponding time zone" after usage of `.isofromat()` method which returns:
- date
- time
- UTC offset
when we convert `datetime.utcnow()` to iso utc offset is trimmed.

app/data_models/app_models.py
petechd: This one is reverted back to what it was, I had some flake8 issues and had to change it but now it's fixed.

app/forms/questionnaire_form.py
petechd: I changed it to London, BST timezone as the offset needs to be at least one hour (one timezone) since this method returns minimum offset.

app/forms/validators.py
petechd: It is used five times in runner files and three times in tests, I would like to know what @ajmaddaford thinks about this.

MebinAbraham: I can see like 9 usages in runner outside tests.

app/views/handlers/block.py
MebinAbraham: I don't see a reason to not keep UTC offset, it is a valid ISO. Worth confirming with Mark as we may need to update the docs around started at in the ons docs.

app/forms/questionnaire_form.py
MebinAbraham: I don't follow. This will only offset if we have specified one.

app/data_models/app_models.py
MebinAbraham: The usage of `tzutc()` should be dropped for `timezone.utc`.

app/views/handlers/block.py
pricem14pc: We specify in the docs that "all datetimes are expressed using ISO_8601 and are assumed to be normalised to UTC _unless_ a timezone identifier is given". As Mebin points out the inclusion of the offset is a valid ISO date. This implies that downstream _should_ expect to handle offsets in the value if they are included.

Having said that, convention has dictated that we haven't previously included the offset (as we've never needed to). While the doc doesn't say if its UTC (GMT +0) then we _won't_ include the offset, we would probably need to communicate any change to include the offset to both SDX and processing.

app/forms/questionnaire_form.py
petechd: I confirmed with Donna that we are supposed to use utc time as this is too much of an edge case.

PR #652, opened 2021-08-26
Add View Submission Feature to the Thank You Page

berroar:

Adds the `view-submssion` feature to the `thank-you` page.

### How to review 

Check that the behaviour matches the scenarios attached to the trello card.

### Checklist


app/views/contexts/thank_you_context.py
```diff
@@ -36,6 +47,11 @@ def build_thank_you_context(
         "submission_text": submission_text,
         "metadata": metadata,
         "guidance": guidance_content,
+        "view_submitted_response": {
+            "enabled": post_submission_schema.get("view_response", False),
+            "expired": view_submitted_response_expired,
```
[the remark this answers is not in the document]
berroar: Updated to use a method to build `view_submitted_response`

app/globals.py
```diff
@@ -101,3 +105,9 @@ def get_metadata(user: User) -> Union[None, MappingProxyType]:
 def get_answer_store(user: User) -> AnswerStore:
     questionnaire_store = get_questionnaire_store(user.user_id, user.user_ik)
     return questionnaire_store.answer_store
+
+
+def get_view_submitted_response_expired(submitted_at: datetime) -> bool:
```
ajmaddaford, line 110: As this method return a boolean, it would be clearer if it was named `is_` rather than `get_`
```suggestion
def is_view_submitted_response_expired(submitted_at: datetime) -> bool:
```

Something else we should consider, probably outside of this pull request, is where we put shared code like this. We should review the use of globals.py and the helpers, libs and utlities namespaces.

app/views/contexts/thank_you_context.py
```diff
@@ -36,9 +39,25 @@ def build_thank_you_context(
         "submission_text": submission_text,
         "metadata": metadata,
         "guidance": guidance_content,
+        "view_submitted_response": build_view_submitted_response(schema, submitted_at),
     }
 
 
+def build_view_submitted_response(schema, submitted_at):
```
ajmaddaford, line 46: The method name should end with `_context` to be consistent with the other method names. 

Arguably they shouldn't start with `build_`, but that is a wider naming issue that should be addressed in a different pull request.

app/views/contexts/thank_you_context.py
```diff
@@ -15,6 +17,7 @@ def build_thank_you_context(
     session_data: SessionData,
     submitted_at: datetime,
     survey_type: str,
+    schema: QuestionnaireSchema,
```
ajmaddaford, line 20: `schema` should be passed as one of the first arguments to `build_thank_you_context` to be consistent with how it is passed around elsewhere.

templates/thank-you.html
```diff
@@ -32,4 +32,24 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
     <a href="https://www.ons.gov.uk/surveys">https://www.ons.gov.uk/surveys</a></p>
   {% endif %}
 
+  {% if content.view_submitted_response.enabled %}
+    <h1 data-qa="view-answers-title">{{ _("Get a copy of your answers") }}</h1>
+    {% if content.view_submitted_response.expired == false %}
+      <p class="u-mb-s">{{ _("You can <a href='{url}' id='view-answers-link'>save or print your answers</a> for your records").format(url = content.view_submitted_response.url) }}</p>
```
ajmaddaford, line 38: The screenshots on the card have a full stop
```suggestion
      <p class="u-mb-s">{{ _("You can <a href='{url}' id='view-answers-link'>save or print your answers</a> for your records.").format(url = content.view_submitted_response.url) }}</p>
```

templates/thank-you.html
```diff
@@ -32,4 +32,24 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
     <a href="https://www.ons.gov.uk/surveys">https://www.ons.gov.uk/surveys</a></p>
   {% endif %}
 
+  {% if content.view_submitted_response.enabled %}
+    <h1 data-qa="view-answers-title">{{ _("Get a copy of your answers") }}</h1>
+    {% if content.view_submitted_response.expired == false %}
```
ajmaddaford, line 37:
```suggestion
    {% if not content.view_submitted_response.expired %}
```

tests/app/views/contexts/test_submitted_response_context.py
```diff
@@ -82,6 +83,21 @@ def test_build_view_submitted_response_context_submitted_text_with_trad_as(app:
         )
 
 
+def test_view_submitted_response_expired_is_true_when_submitted_at_greater_than_the_expiration_time(
+    app: Flask,
+):
+    with app.app_context():
+        questionnaire_store = fake_questionnaire_store()
+        questionnaire_store.submitted_at = datetime.now(timezone.utc) - timedelta(
+            minutes=46
```
ajmaddaford, line 92: It would be better if the VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS value with something added was used here.

tests/app/views/contexts/test_thank_you_context.py
```diff
@@ -52,12 +54,62 @@ def test_default_survey_context(fake_session_data, app: Flask):
...
+
+    with app.app_context():
+
+        context = build_thank_you_context(
+            fake_session_data, submitted_at, SURVEY_TYPE_DEFAULT, SCHEMA
+        )
+
+        assert context["view_submitted_response"]["expired"] is True
+
+
```
ajmaddaford, line 106: These tests have been added between other tests, they should be moved to after the `test_default_survey_context_with_trad_as` method.

tests/app/views/contexts/test_thank_you_context.py
```diff
@@ -52,12 +54,62 @@ def test_default_survey_context(fake_session_data, app: Flask):
...
+    with app.app_context():
+        context = build_thank_you_context(
+            fake_session_data, SUBMITTED_AT, SURVEY_TYPE_DEFAULT, SCHEMA
+        )
+
+        assert context["view_submitted_response"]["expired"] is False
+        assert context["view_submitted_response"]["url"] == "/submitted/view-response/"
+
+
+def test_view_submitted_response_expired_is_true_when_submitted_at_greater_than_expiry_time(
```
ajmaddaford, line 93: The method names are possibly unnecessarily long. Instead of:
```
test_view_submitted_response_enabled_is_true_when_set_in_schema
test_view_submitted_response_enabled_is_false_when_not_enabled_in_schema
test_view_submitted_response_expired_is_false_when_submitted_at_less_than_expiry_time
test_view_submitted_response_expired_is_true_when_submitted_at_greater_than_expiry_time
```
they could be:
```
test_view_submitted_response_enabled
test_view_submitted_response_not_enabled
test_view_submitted_response_not_expired
test_view_submitted_response_expired
```

tests/functional/spec/thank_you.spec.js
```diff
@@ -39,3 +41,22 @@ describe("Thank You Default", () => {
     });
   });
 });
+
+describe("Thank You Default View Response Enabled", () => {
+  describe("Given I launch a questionnaire where view response is enabled", () => {
+    beforeEach(() => {
+      browser.openQuestionnaire("test_thank_you.json");
```
ajmaddaford, line 48: Is there a reason for adding this to the thank you page schema and tests, rather than the view submitted response schema and tests?

berroar: I added it here as I wanted to test that the `view-answers` link etc is available on the thank you page, I can move it if you think it's better with the view submitted response tests?

templates/thank-you.html
```diff
@@ -32,4 +32,24 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
     <a href="https://www.ons.gov.uk/surveys">https://www.ons.gov.uk/surveys</a></p>
```
LJBabbage, line 32: I would probably line these up to avoid any confusion that there is view-answers and a view-submitted-response feature so view-submitted-response-title rather than view-answers-title. And I would go view-submitted-response-warning etc as well

LJBabbage (approved):
One minor suggestion, but its my preference and can be ignored if you think otherwise

app/views/contexts/thank_you_context.py
```diff
@@ -15,9 +17,16 @@ def build_thank_you_context(
...
     guidance_content: Optional[dict] = None,
 ) -> Mapping:
 
+    submission_schema: Mapping = schema.get_post_submission() or {}
+    view_answers = int((datetime.now(tz=tzutc()) - submitted_at).total_seconds()) < 2700
```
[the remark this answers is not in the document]
LJBabbage: In Mebins absence I am happy this has been resolved

tests/functional/spec/thank_you.spec.js
ajmaddaford: Ah ok, I think that does make sense in this test, but raises questions about what the view submitted response schema is for. Given we have a card to review and rationalise the schemas, I think this is fine for now.

ajmaddaford (approved):

templates/thank-you.html
berroar: Have aligned these 👍

PR #655, opened 2021-09-06
Refactor Field handlers to make use of ValueSourceResolver

LJBabbage:
Reuse ValueSourceResolver for field_handlers

### How to review 
Check everything still works as expected

### Checklist


app/forms/field_handlers/field_handler.py
```diff
@@ -18,23 +21,38 @@ class FieldHandler(ABC):
...
+
+        self.value_source_resolver = ValueSourceResolver(
+            answer_store=self.answer_store,
+            list_store=self.list_store,
+            metadata=self.metadata,
+            schema=self.schema,
+            location=self.location,
+            list_item_id=list_item_id,
+            escape_answer_values=True,
+        )
```
MebinAbraham, line 55: This is currently initialised for every answer and for every option that has a detail answer which is wasteful. We should make this a cached property then it will only be initialised once per answer and only when it is required. 

Even with the cached property we are still initialising it per answer when needed instead of per block but changing that behaviour requires further refactoring i.e not passing schema in but a value source resolver to handlers, therefore that is something to look at in future.

app/forms/field_handlers/field_handler.py
```diff
@@ -66,20 +84,14 @@ def get_mandatory_validator(self) -> Union[ResponseRequired, Optional[Any]]:
             )
         return validators.Optional()
 
-    def get_schema_value(self, schema_element: dict) -> Optional[AnswerValueTypes]:
+    def get_schema_value(
+        self, schema_element: dict
+    ) -> Union[ValueSourceEscapedTypes, ValueSourceTypes]:
```
MebinAbraham, line 89: Not a blocker, but we could make an alias for this called something like `ResolvedValueSourceTypes`. This is used in multiple places and in future we will use it in other places and since we are using a custom alias already, using one would make it clearer in my opinion.

app/questionnaire/value_source_resolver.py
```diff
@@ -23,7 +23,7 @@ class ValueSourceResolver:
     list_store: ListStore
     metadata: dict
     schema: QuestionnaireSchema
-    location: Union[Location, RelationshipLocation]
+    location: Union[Optional[Location], RelationshipLocation]
```
MebinAbraham, line 26: When we have more than one type, we should use `None` instead of `Optional` as per our type hinting style guide.

MebinAbraham: If `location` can be `None` then the resolver code needs to be tweaked to handle location not existing, it currently doesn't and would fail if it didn't have a location when trying to access location's attribute.

app/forms/field_handlers/field_handler.py
LJBabbage: :laughing: You introduced it

app/questionnaire/value_source_resolver.py
LJBabbage: This is a slightly odd situation. I am thinking that post_submission would be where location is none, but at that point get_schema_value would not be called from a field_handler??

app/questionnaire/value_source_resolver.py
```diff
@@ -115,7 +121,7 @@ def resolve(
         if source == "answers":
```
LJBabbage, line 121: I can't extend this condition due to default answer line 49

app/forms/field_handlers/field_handler.py
```diff
@@ -18,24 +21,41 @@ class FieldHandler(ABC):
     def __init__(
         self,
         answer_schema: dict,
-        error_messages: dict = None,
+        schema: QuestionnaireSchema,
         answer_store: AnswerStore = None,
+        list_store: ListStore = None,
         metadata: dict = None,
```
MebinAbraham, line 27: I thought we agreed we will now always pass everything to generate_form > field handlers. Field handler should now always get passed the stores, and metadata. Atm, the Value source resolver still doesn't get what it needs as an empty store is initialised when the value is None which is just hiding the problem, we should pass the actual stores in through from the route.

LJBabbage: Yeah, remember that bit when I said I was confused what we were doing, and it was never really cleared up.

app/forms/questionnaire_form.py
```diff
@@ -328,28 +341,32 @@ def _option_value_in_data(answer, option, data):
...
     for answer in question.get("answers", []):
         for option in answer.get("options", []):
             if "detail_answer" in option:
                 disable_validation = not _option_value_in_data(answer, option, data)
                 detail_answer = option["detail_answer"]
+
                 answer_fields[option["detail_answer"]["id"]] = get_field_handler(
                     detail_answer,
-                    error_messages,
+                    schema,
```
ajmaddaford, line 357: I think we should consider initialising the `ValueSourceResolver` before looping over the answers, and passing it to the `get_field_handler` method. It looks like that would avoid the need to pass schema and the various stores (answers, lists, metadata). 

It looks like the only use of a store outside of resolving value sources is for calculations (https://github.com/ONSdigital/eq-questionnaire-runner/blob/refactor-field-handler/app/forms/questionnaire_form.py#L110), which could be modified to use the value source resolver. For now this could be done in code rather than modifying the `calculations` schema and another card raised to make the schema changes.

app/views/handlers/confirm_email.py
```diff
@@ -65,8 +72,9 @@ def form(self):
         return generate_form(
             schema=self._schema,
             question_schema=self.question_schema,
-            answer_store=None,
-            metadata=None,
+            answer_store=self._questionnaire_store.answer_store,
+            list_store=self._questionnaire_store.list_store,
+            metadata=self._questionnaire_store.metadata,
```
ajmaddaford, line 77: We should consider passing questionnaire_store instead of it's individual stores here.

LJBabbage: True, but this is done all over, IR, question, confirm_email, feedback, even block and submit could do it as well

app/forms/questionnaire_form.py
LJBabbage: :hear_no_evil: I would have to change those bloody tests again

app/views/handlers/confirm_email.py
ajmaddaford: I mentioned it because list store is being added. I agree we should be consistent, and a card should be raised for that elsewhere in the codebase, but as this is being changed now I think it makes sense to rationalise to using `questionnaire_store` rather than adding `list_store`.

app/forms/questionnaire_form.py
ajmaddaford: Yeah, not ideal, but a lot of them could be reverted, and would end up simpler.

app/forms/field_handlers/duration_handler.py
```diff
@@ -9,7 +9,9 @@ class DurationHandler(FieldHandler):
 
     def get_field(self) -> FormField:
         return FormField(
-            get_duration_form(self.answer_schema, self.error_messages),
+            get_duration_form(
+                self.answer_schema, self.value_source_resolver.schema.error_messages
+            ),
```
MebinAbraham, line 14: Not a fan of using attributes from one object for another use case than it was intended for. Normally we make most object attrs private that are needed to the internals of the class only. Doing this makes Value source resolver coupled with the handler even though `field` shouldn't rely on value source.

I personally think it should be changed, but I won't block. I think the reference should be updated to `self.error_messages` which you already have set up in the `__init__`.
```py
self.value_source_resolver = value_source_resolver
self.error_messages = self.value_source_resolver.schema.error_messages or {}
```

app/forms/questionnaire_form.py
```diff
@@ -225,9 +234,19 @@ def _validate_calculated_question(
...
+        list_item_id = self.location.list_item_id if self.location else None
+        value_source_resolver = ValueSourceResolver(
+            answer_store=self.answer_store,
+            list_store=self.list_store,
+            metadata=self.metadata,
+            schema=self.schema,
+            location=self.location,
+            list_item_id=list_item_id,
+            escape_answer_values=False,
         )
```
MebinAbraham, line 247: This and these same lines in `get_answer_fields` could be removed if `value_source_resolver` was a property of the `QuestionnaireForm`. Not sure if that is necessarily the right place or whether it should in block handler, but either way I think one of those approaches is more appropriate than initialising inline here and in `get_answer_fields`.

MebinAbraham (commented):
Much clearer now 👍 

There are still typing issues hidden since we don't type check `QuestionnaireForm` so we should look to get that typed. I'll raise a card for it.

app/forms/questionnaire_form.py
MebinAbraham: As discussed, `get_answer_fields` doesn't have access to the form yet, so this can't be done for that. It could be done for the date handler but since there is only a single usage at the moment, it decided it was okay to keep as is for now.

app/forms/field_handlers/number_handler.py
```diff
@@ -89,7 +85,7 @@ def get_field_references(
     def _get_number_field_validators(
         self,
     ) -> list[Union[NumberCheck, NumberRange, DecimalPlaces]]:
-        answer_errors = self.error_messages.copy()
+        answer_errors = self.value_source_resolver.schema.error_messages.copy()
```
MebinAbraham, line 88: This should be `self.error_messages`

app/forms/field_handlers/number_handler.py
```diff
@@ -70,7 +66,7 @@ def get_field(self) -> Union[DecimalField, IntegerField]:
 
     def get_field_references(
         self,
-    ) -> dict[str, Union[bool, Optional[AnswerValueTypes]]]:
+    ) -> dict[str, Union[bool, Union[ValueSourceEscapedTypes, ValueSourceTypes]]]:
```
MebinAbraham, line 69: The nested `Union` is redundant.
```suggestion
    ) -> dict[str, Union[bool, ValueSourceEscapedTypes, ValueSourceTypes]]:
```

app/settings.py
```diff
@@ -139,7 +140,7 @@ def utcoffset_or_fail(date_value, key):
 )
 
 VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS = int(
-    os.getenv("VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS", "2700")
+    os.getenv("VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS", "10")
```
MebinAbraham, line 143: This needs reverting.

app/setup.py
```diff
@@ -97,6 +97,7 @@ def __call__(self, environ, start_response):
         return self.app(environ, start_response)
 
 
+# pylint: disable=W1514
```
MebinAbraham, line 100: How come these have been applied?

app/settings.py
LJBabbage: Accidentally changed, was testing something else

app/setup.py
LJBabbage: Accidentally changed, was testing something else

MebinAbraham (approved):
Apart from the coupling of error messages with handlers, LGTM.

app/forms/field_handlers/field_handler.py
```diff
@@ -18,20 +18,17 @@ class FieldHandler(ABC):
...
         disable_validation: bool = False,
         question_title: str = None,
     ):
-        self.answer_schema = answer_schema
+        self.value_source_resolver = value_source_resolver
         self.error_messages = error_messages or {}
-        self.answer_store = answer_store or AnswerStore()
-        self.metadata = metadata or {}
-        self.location = location
+        self.answer_schema = answer_schema
```
ajmaddaford, line 28: Minor, but these should match the init param order.
```python
        self.answer_schema = answer_schema
        self.value_source_resolver = value_source_resolver
        self.error_messages = error_messages or {}
```

app/forms/questionnaire_form.py
```diff
@@ -10,16 +10,25 @@
...
+        answer_store,
+        list_store,
+        metadata,
+        location,
+        **kwargs,
     ):
         self.schema = schema
         self.question = question_schema
+        self.list_store = list_store
         self.answer_store = answer_store
```
ajmaddaford, line 32: The ordering here should match the init param order
```suggestion
        self.answer_store = answer_store
        self.list_store = list_store
```

tests/app/forms/field_handlers/test_date_handler.py
```diff
@@ -113,58 +115,53 @@ def test_get_referenced_offset_value_for_meta(app):
...
+    answer_store.add_or_update(test_answer_id)
+    value_source_resolver.answer_store = answer_store
+    value_source_resolver.location = location
+    value_source_resolver.list_item_id = list_item_id
+    answer = {"maximum": {"value": {"identifier": "date", "source": "answers"}}}
 
-    handler = DateHandler(answer, answer_store=answer_store, location=location)
+    handler = DateHandler(answer, value_source_resolver)
     maximum_date = handler.get_date_value("maximum")
+    maximum_date = handler.transform_date_by_offset(maximum_date, {"months": 1})
```
ajmaddaford, line 156: Why does `transform_date_by_offset` need to be called rather than being part of the answer schema as it was previously?

app/forms/field_handlers/field_handler.py
ajmaddaford: Resolved in https://github.com/ONSdigital/eq-questionnaire-runner/commit/fdc18d4747444751dbb2bea1da8180d1eea9d06e

tests/app/forms/field_handlers/test_date_handler.py
ajmaddaford: Resolved in https://github.com/ONSdigital/eq-questionnaire-runner/commit/fdc18d4747444751dbb2bea1da8180d1eea9d06e

ajmaddaford (approved):

PR #656, opened 2021-09-06
New routing rules for greater then and less than numbers

rmccar:
This PR is to add new schemas for greater than and less than numbers using the new routing rules engine.

Also:
- Fixes some typos
- Removes a duplicate routing test operators test
- Removes duplicate path finder test for old schema
- Adds tests for greater than and less than, testing for them to fail when number is equal
- Removes functional tests for non single condition routing

### How to review 
- Tests pass
- Check test changes make sense
- Check new schemas work as expected

### Checklist


rmccar:
[Numbers: Using the new routing rules >=, <=, <,> (3 SP)](https://trello.com/c/TJgLHJD8/4854-numbers-using-the-new-routing-rules-3-sp)

LJBabbage (changes requested):
Think you might have missed the functional test update (number.spec.js) on the card
```update existing functional tests to use new schema```

File: tests/app/questionnaire/routing/test_operators.py [hunk 1]
```diff
@@ -76,6 +79,24 @@ def test_operation_greater_than(operands, expected_result):
     assert operator.evaluate(operands) is expected_result
 
 
+@pytest.mark.parametrize(
+    "operands, expected_result",
+    greater_than_and_less_than_operations_equals_test,
+)
+def test_operation_greater_than(operands, expected_result):
```
berroar, line 86: There are three tests here with the same name, it's a bit confusing on what's being tested

berroar (changes requested):

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -76,6 +79,15 @@ def test_operation_greater_than(operands, expected_result):
     assert operator.evaluate(operands) is expected_result
 
 
+@pytest.mark.parametrize(
+    "operands, expected_result",
+    greater_than_and_less_than_operations_equals_test,
+)
+def test_operation_greater_than_same_number(operands, expected_result):
+    operator = Operator(Operator.GREATER_THAN)
+    assert operator.evaluate(operands) is expected_result
```
MebinAbraham, line 88: This is already being testing and is not needed.

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -85,6 +97,15 @@ def test_operation_less_than(operands, expected_result):
     assert operator.evaluate(operands) is not expected_result
 
 
+@pytest.mark.parametrize(
+    "operands, expected_result",
+    greater_than_and_less_than_operations_equals_test,
+)
+def test_operation_less_than_same_number(operands, expected_result):
+    operator = Operator(Operator.LESS_THAN)
+    assert operator.evaluate(operands) is expected_result
```
MebinAbraham, line 106: This is already being testing and is not needed.

tests/app/questionnaire/test_path_finder.py
```diff
@@ -100,29 +100,6 @@ def test_routing_path_with_conditional_path(self):
 
         self.assertEqual(routing_path, expected_path)
 
-    def test_routing_basic_and_conditional_path(self):
```
MebinAbraham, line 103: How come this was removed?. The naming isn't the best.

test_schemas/en/test_new_routing_number_greater_than_or_equal_single_condition.json
```diff
@@ -0,0 +1,133 @@
...
+                                    "goto": {
+                                        "block": "correct-answer",
+                                        "when": [
+                                            {
+                                                "id": "answer",
+                                                "condition": "greater than or equal to",
+                                                "value": 123
+                                            }
+                                        ]
+                                    }
```
MebinAbraham, line 65: This is the old routing structure, the new operator structure need to be used so here `>=`.

test_schemas/en/test_new_routing_number_less_than_or_equal_single_condition.json
```diff
@@ -0,0 +1,133 @@
...
+                                        "block": "correct-answer",
+                                        "when": [
+                                            {
+                                                "id": "answer",
+                                                "condition": "less than or equal to",
+                                                "value": 123
+                                            }
+                                        ]
+                                    }
+                                },
```
MebinAbraham, line 66: Same here, this is using the old structure instead of the new one.

test_schemas/en/test_new_routing_number_less_than_or_equal_single_condition.json
```diff
@@ -0,0 +1,133 @@
+{
```
MebinAbraham, line 1: The `_single_condition` bit should be dropped from the schema names, and the references need to be updated.

MebinAbraham (commented):
Maybe in future we should collate all routing schemas into one by groups i.e numbers/dates etc.

tests/app/questionnaire/routing/test_operators.py
rmccar: Ive added this test and the one below because for greater than and less than I realised that there was no test for if the number entered was the same as the number in the question. For example `
[(1, 1), False],`. This is only tested for currently on the `<=` and `>=` tests. I also needed to do them separately to the other tests because of the way the other tests were written, less than and greater than expect opposite results but in this case they would both be expecting "False"

tests/app/questionnaire/routing/test_operators.py
rmccar: See comment above

tests/app/questionnaire/test_path_finder.py
rmccar: Thought we had discussed in tech session not needing to test the old engine anymore?

rmccar: But I agree the naming could be clearer

MebinAbraham: Oh just realised the same tests is done below with the new schemas.

tests/functional/spec/features/routing/number.spec.js
```diff
@@ -57,66 +57,54 @@ describe("Feature: Routing on a Number", () => {
...
+  describe("Less Than", () => {
+    describe("Given I start number routing less than survey", () => {
       before(() => {
-        browser.openQuestionnaire("test_routing_number_greater_than_or_equal.json");
+        browser.openQuestionnaire("test_new_routing_number_less_than.json");
       });
 
-      it("When I enter a greater number, Then I should be routed to the correct page", () => {
-        $(NumberQuestionPage.answer()).setValue(555);
+      it("When I enter a smaller number, Then I should be routed to the correct page", () => {
```
MebinAbraham, line 82: We should make the tests explicit and consistent. For example, `When I enter a number less than 123, Then ...`

Some tests already do this, so can we just make sure it is consistent between these files.

tests/functional/spec/features/routing/number.spec.js
```diff
@@ -142,35 +130,9 @@ describe("Feature: Routing on a Number", () => {
...
-        $(NumberQuestionPage.submit()).click();
-        expect(browser.getUrl()).to.contain(IncorrectAnswerPage.pageName);
-      });
-    });
-
     describe("Given I have number routing with a less than or equal to single condition", () => {
       before(() => {
-        browser.openQuestionnaire("test_routing_number_less_than_or_equal_single_condition.json");
+        browser.openQuestionnaire("test_new_routing_number_less_than_or_equal.json");
       });
```
MebinAbraham, line 136: I think we should remove the single_condition references.

rmccar: I think Ive removed them all now, I cant find any more

tests/functional/spec/features/routing/number.spec.js
```diff
@@ -27,16 +27,16 @@ describe("Feature: Routing on a Number", () => {
...
 
-      it("When I enter a different number, Then I should be routed to the correct page", () => {
+      it("When I enter a number that isn't 123, Then I should be routed to the correct page", () => {
         $(NumberQuestionPage.answer()).setValue(987);
         $(NumberQuestionPage.submit()).click();
         expect(browser.getUrl()).to.contain(CorrectAnswerPage.pageName);
       });
 
-      it("When I enter a matching number, Then I should be routed to the incorrect page", () => {
+      it("When I enter the 123, Then I should be routed to the incorrect page", () => {
```
MebinAbraham, line 39:
```suggestion
      it("When I enter 123, Then I should be routed to the incorrect page", () => {
```

MebinAbraham (approved):

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -58,14 +58,18 @@ def test_operation_not_equal(operands, expected_result):
...
     [(next_week, current_date), True],
-    [(next_week, current_date), True],
     # Test False
     [(0.5, 0.7), False],
     [(1.0, 2), False],
     [(3, 7), False],
     [(current_date, next_week), False],
 ]
 
+greater_than_and_less_than_operations_equals_test = [
```
berroar, line 68: Might be clearer just as
```suggestion
greater_than_and_less_than_operations_equals = [
```

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -76,6 +80,15 @@ def test_operation_greater_than(operands, expected_result):
     assert operator.evaluate(operands) is expected_result
 
 
+@pytest.mark.parametrize(
+    "operands, expected_result",
+    greater_than_and_less_than_operations_equals_test,
+)
+def test_operation_greater_than_same_number(operands, expected_result):
```
LJBabbage, line 87: I don't see why we would have the same test repeated, if greater_than_and_less_than_operations_equals_test (or not_equals as suggested above) and greater_than_and_less_than_operations need to be kept separate for other test, why just not add them together for this one. Can be done before the parametrize or even in it like 

```@pytest.mark.parametrize(
    "operands, expected_result",
    greater_than_and_less_than_operations + not_equals,
)
def test_operation_greater_than(operands, expected_result):
    operator = Operator(Operator.GREATER_THAN)
    assert operator.evaluate(operands) is expected_result
```

LJBabbage (changes requested):

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -58,14 +58,18 @@ def test_operation_not_equal(operands, expected_result):
...
     [(next_week, current_date), True],
-    [(next_week, current_date), True],
     # Test False
     [(0.5, 0.7), False],
     [(1.0, 2), False],
     [(3, 7), False],
     [(current_date, next_week), False],
 ]
 
+greater_than_and_less_than_operations_equals = [
```
LJBabbage, line 68: Find the name confusing, isn't this just not_equals?

rmccar: I suppose you could view it either way, its testing to see if the numbers are equal. But its expecting them to not be equal

rmccar: Im happy to change it though if its clearer

LJBabbage: Although I do see we have an odd way of doing not_equals, which adding this in might make it even more confusing

LJBabbage (approved):

berroar (approved):

tests/app/questionnaire/routing/test_operators.py
LJBabbage: Although this works for the greater than, how less than is implement it wouldn't. For me the tests still look odd and have repetition, but it works, so will resolve

rmccar: I would do this but combining the tests like this would only work for the greater than ones. For the less than ones, one is expecting the result and one is not expecting the result so it wouldn't work combining them. I could still combine the greater than one but I just think doing one and not the other would be more confusing.

PR #657, opened 2021-09-07
Add print button to the view submitted response page

berroar:

Adds a `Print answers` button to the `view-submitted-response` page.

### How to review 

Use the `test_view_submitted_response` schema to test this change

### Checklist

### DIFF
--- a/templates/view-submitted-response.html
+++ b/templates/view-submitted-response.html
@@ -34,14 +34,14 @@
         "type": 'button',
         "text": 'Print answers',
         "buttonStyle": "print",
+        "classes": "btn--small",
         "attributes": {
           "data-qa": "btn-print"
         }
       })
     }}
-    <br><br>
     {% block summary %}
-      <div class="u-mb-m">
+      <div class="u-mt-l">
         {% include 'partials/summary/summary.html' %}
       </div>
     {% endblock summary %}
--- a/templates/view-submitted-response.html
+++ b/templates/view-submitted-response.html
@@ -41,7 +41,7 @@
       })
     }}
     {% block summary %}
-      <div class="u-mt-l">
+      <div class="u-mt-l u-mb-l">
         {% include 'partials/summary/summary.html' %}
       </div>
     {% endblock summary %}
--- a/templates/view-submitted-response.html
+++ b/templates/view-submitted-response.html
@@ -29,6 +29,17 @@
   </h1>
   {{ onsMetadata(content.metadata)}}
   {% if not content.view_submitted_response.expired %}
+    {{
+      onsButton({
+        "type": 'button',
+        "text": 'Print answers',
+        "buttonStyle": "print",
+        "attributes": {
+          "data-qa": "btn-print"
+        }
+      })
+    }}
+    <br><br>
     {% block summary %}
       <div class="u-mb-m">
         {% include 'partials/summary/summary.html' %}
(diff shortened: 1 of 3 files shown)


berroar:
[Ability to print (1 SP)](https://trello.com/c/LppdlOi1/4839-ability-to-print-1-sp)

templates/view-submitted-response.html
```diff
@@ -29,6 +29,17 @@ <h1 class="u-mt-m">
...
+      onsButton({
+        "type": 'button',
+        "text": 'Print answers',
+        "buttonStyle": "print",
+        "attributes": {
+          "data-qa": "btn-print"
+        }
+      })
+    }}
+    <br><br>
```
MebinAbraham, line 42: Instead of break tags, this should be done via CSS, the prototype set `u-mt-l` CSS classes on the summary.

templates/view-submitted-response.html
```diff
@@ -29,6 +29,17 @@ <h1 class="u-mt-m">
   </h1>
   {{ onsMetadata(content.metadata)}}
   {% if not content.view_submitted_response.expired %}
+    {{
+      onsButton({
+        "type": 'button',
+        "text": 'Print answers',
+        "buttonStyle": "print",
```
MebinAbraham, line 36: The button in the prototype is the small version, which can be done via `"classes": "btn--small",`

templates/view-submitted-response.html
```diff
@@ -29,8 +29,19 @@ <h1 class="u-mt-m">
...
+        "classes": "btn--small",
+        "attributes": {
+          "data-qa": "btn-print"
+        }
+      })
+    }}
     {% block summary %}
-      <div class="u-mb-m">
+      <div class="u-mt-l">
```
MebinAbraham, line 44: You still want to keep the margin bottom, also this should really be added via the ons summary component, but we need to do some refactoring and clean up as section summary doesn't set these. For now, I think it is fine to add here.

tests/functional/spec/features/view_submitted_response/view_submitted_response.js
```diff
@@ -15,6 +15,7 @@ describe("View Submitted Response", () => {
     $(AddressBlockPage.submit()).click();
     $(SubmitPage.submit()).click();
     browser.url("/submitted/view-response");
+    expect($(ViewSubmittedResponsePage.printButton()).getText()).to.equal("Print answers");
```
MebinAbraham, line 18: I think you should assert that it is rendered on the page via like `isDisplayed()`

MebinAbraham (approved):

LJBabbage (changes requested):
We seem to only have positive checks in this story and the one previous unless I missed it. We really should be checking that content, including the print button isn't displayed after 45 minutes (naturally we would override that in the test to 0)

MebinAbraham:
> button isn't displayed after 45 minutes

We don't support that yet, we return a 404 atm as discussed in tech session. It will handled in https://trello.com/c/nYmiT9i0/4846-view-submitted-response-page-after-timeout

LJBabbage:
> > button isn't displayed after 45 minutes
> 
> We don't support that yet, we return a 404 atm as discussed in tech session. It will handled in https://trello.com/c/nYmiT9i0/4846-view-submitted-response-page-after-timeout

Probably need an out of scope on the original cards as it's not obvious and reviewing should be dependent on tech session attendance

LJBabbage (approved):

PR #658, opened 2021-09-09
New routing dates

LJBabbage:
This PR adds schemas for the new routing rules and update the functional tests to use them

N.B New functional tests have been added for >= and <=, they don't need integration test as all operators for dates are tested at a higher level

### How to review 
Make sure the new schemas and changes in functional tests work

### Checklist

### DIFF
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=update-new-rules-date-types
 docker pull onsdigital/eq-questionnaire-validator:$tag
 docker run -d -p 5001:5000 "onsdigital/eq-questionnaire-validator:$tag"
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=update-new-rules-date-types
+tag=latest
 docker pull onsdigital/eq-questionnaire-validator:$tag
 docker run -d -p 5001:5000 "onsdigital/eq-questionnaire-validator:$tag"
--- a/test_schemas/en/test_new_routing_date_equals.json
+++ b/test_schemas/en/test_new_routing_date_equals.json
@@ -1,0 +1,240 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "001",
+    "title": "Test Routing Date Equals",
+    "theme": "default",
+    "description": "A test survey for routing based on equal dates",
+    "metadata": [
+        {
+            "name": "user_id",
+            "type": "string"
+        },
+        {
+            "name": "period_id",
+            "type": "string"
+        },
+        {
+            "name": "ru_name",
+            "type": "string"
+        }
+    ],
+    "questionnaire_flow": {
+        "type": "Linear",
+        "options": {
+            "summary": {
+                "collapsible": false
+            }
+        }
+    },
+    "sections": [
+        {
+            "id": "default-section",
+            "groups": [
+                {
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "comparison-date-block",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "comparison-date-answer",
+                                        "mandatory": true,
+                                        "q_code": "11",
+                                        "type": "Date"
+                                    }
+                                ],
+                                "id": "comparison-date-question",
+                                "title": "Title",
+                                "type": "General",
+                                "guidance": {
+                                    "contents": [
+                                        {
+                                            "title": "If you enter 31/03/2018 the following dates will be valid",
+                                            "list": [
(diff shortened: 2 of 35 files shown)


tests/functional/spec/features/routing/date.spec.js
```diff
@@ -118,22 +125,35 @@ describe("Feature: Routing on a Date", () => {
...
 
         const expectedUrl = browser.getUrl();
 
         expect(expectedUrl).to.contain(CorrectAnswerPage.pageName);
       });
 
-      it("When I enter a date less than or equal to March 2017, Then I should be routed to the incorrect page", () => {
-        $(DateGreaterThanQuestionPage.Month()).setValue(3);
-        $(DateGreaterThanQuestionPage.Year()).setValue(2017);
+      it("When I enter the 1st March 2017, Then I should be routed to the incorrect page", () => {
```
LJBabbage, line 142: Although this wasn't supposed to be changed I could leave an or condition in a test that wasn't exercising both parts, so I added an extra test. I also need to change the type as it was MonthYear

tests/functional/spec/features/routing/date.spec.js
```diff
@@ -143,33 +163,126 @@ describe("Feature: Routing on a Date", () => {
...
+        $(DateLessThanQuestionPage.month()).setValue(monthToday);
+        $(DateLessThanQuestionPage.year()).setValue(yearToday);
+        $(DateLessThanQuestionPage.submit()).click();
+
+        const browserUrl = browser.getUrl();
+
+        expect(browserUrl).to.contain(CorrectAnswerPage.pageName);
+      });
+
+      it("When I enter a date equal to today, Then I should be routed to the incorrect page", () => {
```
LJBabbage, line 224: As above

MebinAbraham (commented):
I think we should have additional examples or at least some of the existing examples should be updated to use `MonthYear` and `Year` date types since they are separate Answer types and we want to ensure those are also as expected.

tests/functional/spec/features/routing/date.spec.js
```diff
@@ -1,17 +1,24 @@
...
+import DateEqualsComparisonQuestionPage from "../../../generated_pages/new_routing_date_equals/comparison-date-block.page";
+import DateEqualsQuestionPage from "../../../generated_pages/new_routing_date_equals/date-question.page";
+import DateNotEqualsQuestionPage from "../../../generated_pages/new_routing_date_not_equals/date-question.page";
+import DateGreaterThanQuestionPage from "../../../generated_pages/new_routing_date_greater_than/date-question.page";
+import DateGreaterThanOrEqualsQuestionPage from "../../../generated_pages/new_routing_date_greater_than_or_equals/date-question.page";
+import DateLessThanQuestionPage from "../../../generated_pages/new_routing_date_less_than/date-question.page";
+import DateLessThanOrEqualsQuestionPage from "../../../generated_pages/new_routing_date_less_than_or_equals/date-question.page";
+
+const today = new Date();
+const dayToday = today.getDate(); // yesterday
```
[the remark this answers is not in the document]
LJBabbage: Must of missed it, there was a strange way it worked before and I got rid

LJBabbage:
> I think we should have additional examples or at least some of the existing examples should be updated to use `MonthYear` and `Year` date types since they are separate Answer types and we want to ensure those are also as expected.

I don't think we need to do that in functional tests. As long as they are covered in unit I think it's fine

test_schemas/en/test_new_routing_date_equals.json
```diff
@@ -0,0 +1,240 @@
...
+                                                }
+                                            ]
+                                        }
+                                    ]
+                                },
+                                "type": "General"
+                            },
+                            "routing_rules": [
+                                {
+                                    "block": "correct-answer",
```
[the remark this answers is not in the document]
LJBabbage: Not sure what this comment means? I just lifted the other schema and changed the rule

MebinAbraham (commented):
> > I think we should have additional examples or at least some of the existing examples should be updated to use `MonthYear` and `Year` date types since they are separate Answer types and we want to ensure those are also as expected.
> 
> I don't think we need to do that in functional tests. As long as they are covered in unit I think it's fine

The operator's unit tests are testing the operators itself, which is just that date can be compared not the process of mapping Date/MonthYear/Year to an actual datetime, it's not its responsibility to do that. The functional test is to showcase the feature of routing based on a date for which we have 3 types. This is also something the PM would use. I think the functional tests should have examples of the other two date types. I'm not saying that it needs to duplicate the entire thing for both the types, just that some example should use the other types instead of them all using Date, i.e swap some Date for MonthYear/Year.

LJBabbage:
> The operator's unit tests are testing the operators itself, which is just that date can be compared not the process of ...

This was a migration piece, not extension. This argument would suggest we have every rule for every date type. That is an expensive approach

berroar (approved):

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=update-new-rules-date-types
```

MebinAbraham (approved):
👍

PR #660, opened 2021-09-14
View submitted response after timeout

LJBabbage:
After view submitted response timeout has expired (assuming a session still exists)  view-response should now be available, but with the buttons removed and a security message instead. This PR changes it from what used to be 404 and updates the tests

### How to review 
You don't want to wait 45 minutes so update VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS in settings to something sensible and using test_view_submitted_response make sure it displays as expected, before and after the expiration

### Checklist


app/routes/questionnaire.py
```diff
@@ -385,7 +384,7 @@ def get_view_submitted_response(schema, questionnaire_store):
             flask_babel.get_locale().language,
         )
 
-    except (ViewSubmittedResponseNotEnabled, ViewSubmittedResponseExpired):
+    except (ViewSubmittedResponseNotEnabled):
```
MebinAbraham, line 387: The parenthesis is now redundant.
```suggestion
    except ViewSubmittedResponseNotEnabled:
```

MebinAbraham (approved):
Looks good. 

Only thing is this point on the card which is not implemented.
- `After 45 INACTIVE minutes, any action that results in the page reloading will take you to the timeout page`.

We currently go to the `submission-complete` page. But since the error page content will be updated soon, we could look to align this behaviour across the board and remove `submission-complete` then as it is not relevant to just post submission.

LJBabbage:
> Looks good.
> 
> Only thing is this point on the card which is not implemented.
> 
> * `After 45 INACTIVE minutes, any action that results in the page reloading will take you to the timeout page`.
> 
> We currently go to the `submission-complete` page. But since the error page content will be updated soon, we could look to align this behaviour across the board and remove `submission-complete` then as it is not relevant to just post submission.

I left it due to this 
``` 
New content of 'Inactivity session timeout' page out of scope (separate card)
```
Seemed like it needs more analysis so decided not to change what is default behaviour at present

LJBabbage:
> Looks good.
> 
> Only thing is this point on the card which is not implemented.
> 
> * `After 45 INACTIVE minutes, any action that results in the page reloading will take you to the timeout page`.
> 
> We currently go to the `submission-complete` page. But since the error page content will be updated soon, we could look to align this behaviour across the board and remove `submission-complete` then as it is not relevant to just post submission.

I just left it as it was, I didn't even notice it wasn't going "Submission Complete - Timeout test", but I would think that was right, thats what we said we wanted before, that the respondent wanted to know their submission was successful?

MebinAbraham:
> > Looks good.
> > Only thing is this point on the card which is not implemented.
> > 
> > * `After 45 INACTIVE minutes, any action that results in the page reloading will take you to the timeout page`.
> > 
> > We currently go to the `submission-complete` page. But since the error page content will be updated soon, we could look to align this behaviour across the board and remove `submission-complete` then as it is not relevant to just post submission.
> 
> I just left it as it was, I didn't even notice it wasn't going "Submission Complete - Timeout test", but I would think that was right, thats what we said we wanted before, that the respondent wanted to know their submission was successful?

Ya, it is being changed whereby they will have one page to handle them all. The page says why the users may have landed there, which could be session expired/already submitted etc.

tests/functional/spec/census_exit.spec.js
```diff
@@ -28,6 +28,6 @@ describe("Post submission exit", () => {
     $(CensusThankYouPage.exit()).click();
     browser.back();
     expect(browser.getUrl()).to.contain("submitted/thank-you");
-    expect($("body").getHTML()).to.contain("This page is no longer available");
+    expect($("body").getHTML()).to.contain("Your session has timed out due to inactivity");
```
LJBabbage, line 31: It doesn't make sense to introduce code to determine if something is census (which isn't used any more) based or not on the post_submission session expiration.

MebinAbraham: I think while we are here, we should rename the file to `exit.spec.js` since it is not tied to Census. This behaviour should be the same for all schemas.

LJBabbage: True

MebinAbraham (approved):

berroar (approved):

PR #662, opened 2021-09-15
Spike: Download PDF button

berroar:

This spike was to investigate adding the functionality to download the `view-submission` page as a PDF using `flask-weasyprint` (which is already being used in v2).

Unfortunately there were issues rendering the PDF correctly using `flask-weasyprint`. Having done some investigation , although we were not able to find the direct cause, we found that there were issues relating to css media queries which are used frequently in our design system that could potentially be causing issues, [see the issue here](https://github.com/Kozea/WeasyPrint/issues/1286). `flask-weasyprint` also requires system libraries `pango` and `libffi` to be installed in order to be able to run, which would need to be installed with our docker image.

As a result for this Spike I looked at alternative libraries for the PDF download and was able to get the download working with the `PDFKit` library. 

### Caveats

Unfortunately this library is also dependant on system libraries being installed, in order to run the code you will need to install `wkhtmltopdf` with brew ( `brew install wkhtmltopdf`). This will need to be added to the dockerfile.

As there is currently a design system PR open to add a download button to the Design system the styling for the `Download pdf` button is incorrect.

As this is a spike, I have not yet added tests, and am not currently installing `wkhtmltopdf` in the dockerfile.

There is a warning for the PDFKit pypi package that states that `Version in debian/ubuntu repos have reduced functionality`, so we would need to test this to see if we are impacted.

### How to review 

To run locally, ensure that `wkhtmltopdf` is installed via `brew install wkhtmltopdf`, and then `pipenv install --dev` to install the PDFKit library,

Use `test_view_submitted_response` to see the added pdf download button and generate the download.

Is this library appropriate - should we investigate further alternatives?

### Checklist


berroar:
[Spike Flask-WeasyPrint (1 day)](https://trello.com/c/LIYz7voN/4861-spike-flask-weasyprint-1-day)

MebinAbraham:
Given the statement `Version in debian/ubuntu repos have reduced functionality`, the Dockerfile should be updated so that we can test the spike under those conditions.

pricem14pc (commented):
> _and am not currently installing wkhtmltopdf in the dockerfile._

Is the PR "how to test" up to date @berroar as I'm seeing `wkhtmltopdf` installed in the Docker file. Given its fairly fundamental to its usage (and appears to be building a functional working image) I think it's worth the library usage and image build being part of the spike / PR review so we have sight of any issues with the libraries up front.

MebinAbraham:
The output locally looks differently to the output under our deployed image. The deployed image looks slight better and closer to the print stylesheet, however the style sheets needs tweaks as the cookie banner is rendered as well but should be hidden. The margin is also a bit narrow on the deployed version. Can we pass additional CSS or can it only be provided upfront or only read from a file?. Not sure if it is within the scope of the spike to look at those or just address them when implementing for real.

In general, the PDF created works mostly as expected, just a bit slow, unfortunately (I can see users clicking it multiple times due to the slowness). We should disable the logging outputs it generates.

As Mark surfaced, the docker image size does increase by over 50%, maybe we can look to slim down our image also.

app/routes/questionnaire.py
```diff
@@ -394,6 +395,32 @@ def get_view_submitted_response(schema, questionnaire_store):
...
 
 
+@post_submission_blueprint.route("download-pdf", methods=["GET"])
+@with_questionnaire_store
+@with_schema
+def download_pdf(schema, questionnaire_store):
+
+    metadata = get_metadata(user=current_user)
+
+    filename = "{}_{}.pdf".format(metadata.get("eq_id"), metadata.get("form_type"))
```
MebinAbraham, line 405: This would have to be based on something that always exists, i.e schema_name or the attrs that will exist for business.

pricem14pc:
> The output locally looks differently to the output under our deployed image

@MebinAbraham do you mean running Runner (e.g `pipenv run`) rather than a Docker container running locally?

app/routes/questionnaire.py
pricem14pc: > This would have to be based on something that always exists, i.e schema_name or the attrs that will exist for business.

Fine for the spike. In addition this is the requirements and v2 implementation. However, I am surprised at this as both values are internal and probably shouldn't be exposed to the user. Will require some business analysis when we take it forward.

MebinAbraham:
> > The output locally looks differently to the output under our deployed image
> 
> @MebinAbraham do you mean running Runner (e.g `pipenv run`) rather than a Docker container running locally?

Ya, i.e. diff between MacOS vs Debian.

berroar:
> > The output locally looks differently to the output under our deployed image
> 
> @MebinAbraham do you mean running Runner (e.g `pipenv run`) rather than a Docker container running locally?

I've added example outputs to the trello card now

PR #665, opened 2021-09-17
Add type hints for forms modules `address_form.py`, `email_form.py`, `questionnaire_form.py`

jamesnunn:
Adds type hints to `apps/forms/address_form.py`, `apps/forms/email_form.py`, `apps/forms/questionnaire_form.py` to ease development and identify issues with incompatible types earlier. 

Companion PR, adding type hints to the other modules in `forms`: https://github.com/ONSdigital/eq-questionnaire-runner/pull/668

### How to review 
Run unit tests (most importantly mypy) to show no breakage and that types used are all correct. Reviewers to check the more complicated types like nested dicts/lists if they happen to know the code already.

### Checklist

### DIFF
--- a/app/forms/address_form.py
+++ b/app/forms/address_form.py
@@ -1,12 +1,20 @@
+from __future__ import annotations
+
 import logging
 from functools import cached_property
+from typing import Any
 
 from wtforms import Form, HiddenField, StringField
+from wtforms.validators import InputRequired
 
 logger = logging.getLogger(__name__)
 
+AddressValidatorTypes = list[InputRequired]
 
-def get_address_form(validators):
+
+def get_address_form(
+    validators: AddressValidatorTypes,
+) -> Form:
     class AddressForm(Form):
         line1 = StringField(validators=validators)
         line2 = StringField()
@@ -15,7 +23,8 @@ def get_address_form(validators):
         uprn = HiddenField()
 
         @cached_property
-        def data(self):
-            return super().data
+        def data(self) -> dict[str, Any]:
+            data_: dict[str, Any] = super().data
+            return data_
 
     return AddressForm
--- a/app/forms/field_handlers/number_handler.py
+++ b/app/forms/field_handlers/number_handler.py
@@ -87,7 +87,7 @@ class NumberHandler(FieldHandler):
     def _get_number_field_validators(
         self,
     ) -> list[Union[NumberCheck, NumberRange, DecimalPlaces]]:
-        answer_errors = self.error_messages.copy()
+        answer_errors = dict(self.error_messages)
 
         for error_key in self.validation_messages.keys():
             answer_errors[error_key] = self.get_validation_message(error_key)
--- a/app/forms/questionnaire_form.py
+++ b/app/forms/questionnaire_form.py
@@ -1,31 +1,46 @@
+from __future__ import annotations
+
 import itertools
 import logging
+from collections.abc import Callable
 from datetime import datetime, timedelta, timezone
 from decimal import Decimal
+from typing import Any, Mapping, Optional, Sequence, Union
 
 from dateutil.relativedelta import relativedelta
 from flask_wtf import FlaskForm
 from werkzeug.datastructures import ImmutableMultiDict, MultiDict
 from wtforms import validators
 
+from app.data_models import AnswerStore, AnswerValueTypes, ListStore
 from app.forms import error_messages
-from app.forms.field_handlers import DateHandler, get_field_handler
+from app.forms.field_handlers import DateHandler, FieldHandler, get_field_handler
 from app.forms.validators import DateRangeCheck, MutuallyExclusiveCheck, SumCheck
+from app.questionnaire import Location, QuestionnaireSchema, QuestionSchema
+from app.questionnaire.relationship_location import RelationshipLocation
 from app.questionnaire.value_source_resolver import ValueSourceResolver
 
 logger = logging.getLogger(__name__)
+
+Calculation = Mapping[str, Any]
+QuestionnaireExtraValidators = Mapping[str, Sequence[Callable]]
+Period = Mapping[str, int]
+PeriodLimits = Mapping[str, Any]
+Error = Union[Mapping, Sequence]
+Errors = Mapping[str, Error]
+ErrorList = Sequence[tuple[str, str]]
 
 
 class QuestionnaireForm(FlaskForm):
     def __init__(
         self,
-        schema,
-        question_schema,
-        answer_store,
-        list_store,
-        metadata,
-        location,
-        **kwargs,
+        schema: QuestionnaireSchema,
+        question_schema: QuestionSchema,
+        answer_store: AnswerStore,
+        list_store: ListStore,
+        metadata: dict[str, Any],
+        location: Union[None, Location, RelationshipLocation],
+        **kwargs: Union[MultiDict[str, Any], Mapping[str, Any], None],
     ):
         self.schema = schema
         self.question = question_schema
@@ -33,18 +48,20 @@ class QuestionnaireForm(FlaskForm):
         self.list_store = list_store
         self.metadata = metadata
         self.location = location
-        self.question_errors = {}
-        self.options_with_detail_answer = {}
+        self.question_errors: dict[str, str] = {}
+        self.options_with_detail_answer: dict = {}
         self.question_title = self.question.get("title", "")
 
         super().__init__(**kwargs)
 
-    def validate(self, extra_validators=None):
+    def validate(
+        self, extra_validators: Optional[QuestionnaireExtraValidators] = None
+    ) -> bool:
         """
         Validate this form as usual and check for any form-level validation errors based on question type
         :return: boolean
         """
-        super(QuestionnaireForm, self).validate(extra_validators)
+        super(QuestionnaireForm, self).validate(extra_validators=extra_validators)
         valid_fields = FlaskForm.validate(self)
         valid_date_range_form = True
         valid_calculated_form = True
@@ -71,7 +88,7 @@ class QuestionnaireForm(FlaskForm):
             and valid_mutually_exclusive_form
         )
 
-    def validate_date_range_question(self, question):
+    def validate_date_range_question(self, question: QuestionSchema) -> bool:
         date_from = question["answers"][0]
         date_to = question["answers"][1]
         if self._has_min_and_max_single_dates(date_from, date_to):
@@ -109,25 +126,25 @@ class QuestionnaireForm(FlaskForm):
 
         return True
 
-    def validate_calculated_question(self, question):
+    def validate_calculated_question(self, question: QuestionSchema) -> bool:
         for calculation in question["calculations"]:
-            target_total, currency = self._get_target_total_and_currency(
-                calculation, question
-            )
-            if self.answers_all_valid(
-                calculation["answers_to_calculate"]
-            ) and self._validate_calculated_question(
-                calculation, question, target_total, currency
-            ):
-                # Remove any previous question errors if it passes this OR before returning True
-                if question["id"] in self.question_errors:
-                    self.question_errors.pop(question["id"])
-                return True
+            result = self._get_target_total_and_currency(calculation, question)
+            if result:
+                target_total, currency = result
+                if self.answers_all_valid(
+                    calculation["answers_to_calculate"]
+                ) and self._validate_calculated_question(
+                    calculation, question, target_total, currency
+                ):
+                    # Remove any previous question errors if it passes this OR before returning True
+                    if question["id"] in self.question_errors:
+                        self.question_errors.pop(question["id"])
+                    return True
 
         return False
 
-    def validate_mutually_exclusive_question(self, question):
-        is_mandatory = question.get("mandatory")
+    def validate_mutually_exclusive_question(self, question: QuestionSchema) -> bool:
+        is_mandatory: bool = question["mandatory"]
         messages = (
             question["validation"].get("messages") if "validation" in question else None
         )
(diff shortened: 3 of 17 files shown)


jamesnunn:
Errors remaining - these are harmless but I'll still see if I can reduce them down
```
app/forms/questionnaire_form.py:185: error: Item "None" of "Optional[Answer]" has no attribute "value"
app/forms/questionnaire_form.py:185: error: Incompatible types in assignment (expression has type "Union[str, int, Decimal, Dict[str, Union[int, str]], List[str], Any]", variable has type "Union[None, int, float]")
app/forms/questionnaire_form.py:296: error: Incompatible types in assignment (expression has type "Optional[datetime]", variable has type "datetime")
app/forms/questionnaire_form.py:297: error: Incompatible types in assignment (expression has type "Optional[datetime]", variable has type "datetime")
app/forms/questionnaire_form.py:300: error: Incompatible types in assignment (expression has type "Optional[datetime]", variable has type "datetime")
app/forms/questionnaire_form.py:301: error: Incompatible types in assignment (expression has type "Optional[datetime]", variable has type "datetime")
app/forms/questionnaire_form.py:458: error: Incompatible types in assignment (expression has type "Union[Mapping[Any, Any], Sequence[Any]]", variable has type "Dict[Any, Any]")
app/forms/questionnaire_form.py:464: error: Incompatible types in assignment (expression has type "Union[Mapping[Any, Any], Sequence[Any]]", variable has type "List[Any]")
app/forms/questionnaire_form.py:520: error: Argument 1 to "_clear_detail_answer_field" has incompatible type "Dict[str, Any]"; expected "MultiDict[Any, Any]"
app/forms/questionnaire_form.py:522: error: Incompatible types in assignment (expression has type "Optional[Dict[str, Any]]", variable has type "Dict[Any, Any]")
Found 10 errors in 1 file (checked 172 source files)
Mypy type check faile```

File: app/forms/questionnaire_form.py [hunk 1]
```diff
@@ -148,23 +165,39 @@ def validate_mutually_exclusive_question(self, question):
...
-            target_answer.get("currency"),
-        )
+        calculation_value = self.answer_store.get_answer(
+            calculation["answer_id"]
+        ).value  # type: ignore # expect not None
+        currency = target_answer.get("currency")
+
+        return calculation_value, currency
 
     def validate_date_range_with_period_limits_and_single_date_limits(
```
berroar, line 192: We can use `NoReturn` on some of these to address the `Function is missing a return type annotation` errors

app/forms/questionnaire_form.py
```diff
@@ -186,7 +221,12 @@ def validate_date_range_with_single_date_limits(question_id, period_range):
             raise Exception(exception)
 
     def _validate_date_range_question(
```
berroar, line 223: Missing a `bool` return type here?

app/forms/questionnaire_form.py
jamesnunn: Are there errors for this? Happy to add them but I didn't see the errors to address

berroar: Unless i'm missing something, when I ran mypy against `questionnaire_form` I got the following errors which aren't being ignored:
```
app/forms/questionnaire_form.py:34: error: Function is missing a type annotation for one or more arguments
app/forms/questionnaire_form.py:192: error: Function is missing a return type annotation
app/forms/questionnaire_form.py:211: error: Function is missing a return type annotation
app/forms/questionnaire_form.py:223: error: Function is missing a return type annotation
app/forms/questionnaire_form.py:368: error: Returning Any from function declared to return "Decimal"
app/forms/questionnaire_form.py:398: error: Returning Any from function declared to return "str"
app/forms/questionnaire_form.py:493: error: Function is missing a return type annotation
```

app/forms/questionnaire_form.py
```diff
@@ -247,33 +295,36 @@ def _get_period_range_for_single_date(self, date_from, date_to):
...
-        from_min_period_date = handler.get_date_value("minimum")
-        from_max_period_date = handler.get_date_value("maximum")
+        from_min_period_date: datetime = handler.get_date_value("minimum")  # type: ignore # expect not None
+        from_max_period_date: datetime = handler.get_date_value("maximum")  # type: ignore # expect not None
 
         handler.answer_schema = date_to
-        to_min_period_date = handler.get_date_value("minimum")
-        to_max_period_date = handler.get_date_value("maximum")
+        to_min_period_date: datetime = handler.get_date_value("minimum")  # type: ignore # expect not None
+        to_max_period_date: datetime = handler.get_date_value("maximum")  # type: ignore # expect not None
```
MebinAbraham, line 303: For things like this, this ignore is only needed because we are reusing a function that could return None or a datetime but we know in this instance they should exist, therefore you could refactor the existing method to always return a date.

```py
    def get_date_value_if_exists(self, key: str) -> Optional[datetime]:
        return self.get_date_value(key) if key in self.answer_schema else None

    def get_date_value(self, key:str) -> datetime:
        """
        Gets attributes within a minimum or maximum of a date field and validates that the entered date
        is valid.
        """
        date_value = self.get_referenced_date(key)

        if "offset_by" in self.answer_schema[key]:
            offset = self.answer_schema[key]["offset_by"]
            date_value = self.transform_date_by_offset(date_value, offset)

        return date_value
```
So we use `get_date_value_if_exists` when None is acceptable and the other when we expect to return a value. 

Thoughts?

app/forms/questionnaire_form.py
jamesnunn: They are different ... I don't get those! How do you run yours?

MebinAbraham (commented):
There are outstanding issues which aren't picked up by mypy which can be sorted by updating the mypy config to pick up these files, for example:

```ini
[mypy-app.forms.address_form]
disallow_untyped_defs = True
warn_return_any = True
```

You can't point to the entire module since Ash has done some files in a seperate PR, therefore, it can be rationalised when either of two PRs merges. Also, once one PR merges, it should surface some issues which are not visible at the moment not since the typing is split.

app/forms/questionnaire_form.py
jamesnunn: Sure, I was just avoiding it for the sake of pleasing a type hint

MebinAbraham: I was thinking we type ignore only minor things that say can be `Any` etc but if it can technically break at runtime, it is worth handling if it is only a minor change.

MebinAbraham (commented):
Looking good. Few minor comments.

app/questionnaire/rules.py
```diff
@@ -124,22 +126,21 @@ def get_date_match_value(date_comparison, answer_store, schema, metadata):
...
+                years=offset.get("years", 0),
+            )
 
     return match_value
 
 
-def convert_to_datetime(value: Optional[str]) -> Optional[datetime]:
-    if not value:
-        return None
+def convert_to_datetime(value: str) -> datetime:
```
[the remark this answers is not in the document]
jamesnunn: The typing led me there, but it bugged me the function does the falsy-value check rather than the check being done by the callers of the function. I did follow the code through and check usages and tests, and they all ran fine after the changes. I can revert and just overload if that's the preference

app/forms/questionnaire_form.py
```diff
@@ -247,33 +295,37 @@ def _get_period_range_for_single_date(self, date_from, date_to):
...
     @staticmethod
-    def is_date_form_populated(date_from, date_to):
+    def is_date_form_populated(date_from: Sequence, date_to: Sequence) -> bool:
         return all(field.data for field in itertools.chain(date_from, date_to))
 
     @staticmethod
-    def _has_min_and_max_single_dates(date_from, date_to):
+    def _has_min_and_max_single_dates(
+        date_from: Union[str, Mapping[str, Mapping]],
+        date_to: Union[str, Mapping[str, Mapping]],
```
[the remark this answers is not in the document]
jamesnunn: yeah I'm not sure why that ended up with a str there

app/forms/address_form.py
```diff
@@ -15,7 +23,8 @@ class AddressForm(Form):
         uprn = HiddenField()
 
         @cached_property
-        def data(self):
-            return super().data
+        def data(self) -> dict[str, Any]:
+            data_: dict[str, Any] = super().data
+            return data_
```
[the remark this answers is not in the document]
jamesnunn: He must have abandoned the change as it isn't in his PR

app/questionnaire/rules.py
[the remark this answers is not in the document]
MebinAbraham: Given the change has been done already and references refactored, happy to keep as is 👍

app/forms/field_handlers/number_handler.py
```diff
@@ -87,7 +87,7 @@ def get_field_references(
     def _get_number_field_validators(
         self,
     ) -> list[Union[NumberCheck, NumberRange, DecimalPlaces]]:
-        answer_errors = self.error_messages.copy()
+        answer_errors = dict(self.error_messages).copy()
```
MebinAbraham, line 90: You don't need the `.copy()`, calling `dict()` on a mapping will return a copy already.

app/forms/questionnaire_form.py
```diff
@@ -109,25 +126,25 @@ def validate_date_range_question(self, question):
...
+                target_total, currency = result
+                if self.answers_all_valid(
+                    calculation["answers_to_calculate"]
+                ) and self._validate_calculated_question(
+                    calculation, question, target_total, currency
+                ):
+                    # Remove any previous question errors if it passes this OR before returning True
+                    if question["id"] in self.question_errors:
+                        self.question_errors.pop(question["id"])
+                    return True
```
MebinAbraham, line 142: Not a blocker, but usually it is nice to exit early to avoid such nesting to improve readability.

app/forms/questionnaire_form.py
```diff
@@ -149,23 +166,39 @@ def validate_mutually_exclusive_question(self, question):
             return False
         return True
 
-    def _get_target_total_and_currency(self, calculation, question):
+    def _get_target_total_and_currency(
+        self,
+        calculation: Calculation,
+        question: QuestionSchema,
+    ) -> Union[tuple[Union[Calculation, AnswerValueTypes], Optional[str]], None]:
```
MebinAbraham, line 173: Not a blocker, this is one type, i.e `Optional[tuple[...]]`. Don't think one is necessarily any more readable IMO, therefore just following the style guide.
```suggestion
    ) -> Optional[tuple[Union[Calculation, AnswerValueTypes], Optional[str]]]:
```

app/forms/questionnaire_form.py
jamesnunn: The values possibly returned by all the functions called in this function (`get_referenced_date`, `transform_date_by_offset`) are all optional too, so `get_date_value` as you've refactored here is still have an optional return type. This little gathering of code has this unusual pattern of having an optional main argument which, if it's falsy, returns `None`. I'd normally expect to see callers of the functions do the checking of values before calling the function, which led to [this](https://github.com/ONSdigital/eq-questionnaire-runner/pull/665#discussion_r716990564) change, which did break stuff. I'm happy to delve deeper and refactor these but for the sake of this single type hint, is it worth it?

app/questionnaire/rules.py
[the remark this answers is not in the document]
jamesnunn: I reverted it as it happens - I had broken a functional test and see this (https://github.com/ONSdigital/eq-questionnaire-runner/pull/665#discussion_r721999745) it isn't trivial to fix

app/forms/questionnaire_form.py
MebinAbraham: The typing for `get_referenced_date` and `transform_date_by_offset` implies they both always return a datetime, so when would it be optional?. Have I misunderstood or is the typing not correct right now?.

jamesnunn: `get_referenced_date` returns: `return convert_to_datetime(value) if isinstance(value, str) else None`
`convert_to_datetime` returns None if the arg is falsy
`transform_date_by_offset` returns the original arg if it's falsy

MebinAbraham: Re: prior to your change, this is one of those where we shouldn't get `None` as validator should guard against it, but runner doesn't know that and technically should guard itself. We could simply raise an exception here if we get back `None` instead of a `datetime`, then the typing would behave as expected and can stay as `Optional[datetime]` and code covers NoneTypes. so that would be remove

jamesnunn: All done -raising an exception in the case the return type was None for either value.

app/forms/questionnaire_form.py
jamesnunn: I'm not sure we'd be exiting early from the loop because every calculation in the loop needs to be checked. I could do this to remove the larger nested block?
```
            if not result:
                continue
```

MebinAbraham: Sorry by exit, that was what I was referring to in the instance.

MebinAbraham (approved):
👍

petechd (approved):

app/forms/questionnaire_form.py
MebinAbraham: Happy to mark as resolved, change has been made to resolve remaining error and fix to ensure no uses of `NoReturn`.

ONS-Guilhem-Forey:
[Remove support for old when rules (3 SP)](https://trello.com/c/uCidXQd2/5295-remove-support-for-old-when-rules-3-sp)

ONS-Guilhem-Forey:
[Remove support for old when rules (3 SP)](https://trello.com/c/uCidXQd2/5295-remove-support-for-old-when-rules-3-sp)

PR #666, opened 2021-09-21
Add new routing and or and not operators

berroar:

Adds the new routing rules schemas and functional tests for the `and`, `or` and `not` operators.

### How to review 

Make sure the new schemas and functional tests work

### Checklist

### DIFF
--- a/test_schemas/en/test_new_routing_and.json
+++ b/test_schemas/en/test_new_routing_and.json
@@ -6,7 +6,7 @@
     "survey_id": "001",
     "title": "Test Routing And",
     "theme": "default",
-    "description": "A test survey for routing based on the and operator",
+    "description": "A test survey for routing based on the AND operator",
     "metadata": [
         {
             "name": "user_id",
--- a/test_schemas/en/test_new_routing_and.json
+++ b/test_schemas/en/test_new_routing_and.json
@@ -1,0 +1,175 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "001",
+    "title": "Test Routing And",
+    "theme": "default",
+    "description": "A test survey for routing based on the and operator",
+    "metadata": [
+        {
+            "name": "user_id",
+            "type": "string"
+        },
+        {
+            "name": "period_id",
+            "type": "string"
+        },
+        {
+            "name": "ru_name",
+            "type": "string"
+        }
+    ],
+    "questionnaire_flow": {
+        "type": "Linear",
+        "options": {
+            "summary": {
+                "collapsible": false
+            }
+        }
+    },
+    "sections": [
+        {
+            "id": "default-section",
+            "groups": [
+                {
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "number-question-1",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "answer-1",
+                                        "mandatory": true,
+                                        "type": "Number",
+                                        "label": "Enter 123"
+                                    }
+                                ],
+                                "id": "question",
+                                "title": "Enter the number 123",
+                                "type": "General"
+                            }
+                        },
+                        {
+                            "type": "Question",
+                            "id": "number-question-2",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "answer-2",
+                                        "mandatory": true,
+                                        "type": "Number",
+                                        "label": "Enter 321"
+                                    }
+                                ],
+                                "id": "question",
+                                "title": "Enter the number 321",
+                                "type": "General"
+                            },
+                            "routing_rules": [
+                                {
+                                    "block": "correct-answer",
+                                    "when": {
+                                        "and": [
+                                            {
+                                                "==": [
+                                                    {
+                                                        "source": "answers",
+                                                        "identifier": "answer-1"
+                                                    },
+                                                    123
+                                                ]
+                                            },
+                                            {
+                                                "==": [
+                                                    {
+                                                        "source": "answers",
+                                                        "identifier": "answer-2"
+                                                    },
+                                                    321
+                                                ]
+                                            }
+                                        ]
+                                    }
+                                },
+                                {
+                                    "block": "incorrect-answer"
+                                }
+                            ]
+                        },
+                        {
+                            "type": "Interstitial",
+                            "id": "incorrect-answer",
+                            "content": {
+                                "title": "You did not enter the correct answers",
+                                "contents": [
+                                    {
+                                        "description": {
+                                            "text": "You were asked to enter <em>123</em> and <em>321</em> but you actually entered <em>{answer-1}</em> and <em>{answer-2}</em>.",
+                                            "placeholders": [
+                                                {
+                                                    "placeholder": "answer-1",
+                                                    "value": {
+                                                        "source": "answers",
+                                                        "identifier": "answer-1"
+                                                    }
+                                                },
+                                                {
+                                                    "placeholder": "answer-2",
+                                                    "value": {
+                                                        "source": "answers",
+                                                        "identifier": "answer-2"
+                                                    }
+                                                }
+
+                                            ]
+                                        }
+                                    }
+                                ]
+                            },
+                            "routing_rules": [
+                                {
+                                    "section": "End"
+                                }
+                            ]
+                        },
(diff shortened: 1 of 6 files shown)


berroar:
[Use new routing rules logic operators: and/or/not (3 SP)](https://trello.com/c/93unLV4t/4876-use-new-routing-rules-logic-operators-and-or-not-3-sp)

test_schemas/en/test_new_routing_not.json
```diff
@@ -0,0 +1,177 @@
...
+                            },
+                            "routing_rules": [
+                                {
+                                    "block": "correct-answer",
+                                    "when": {
+                                        "not": [
+                                            {
+                                                "and": [
+                                                    {
+                                                        "==": [
```
MebinAbraham, line 79: Won't block since it technically demonstrates the feature, but I think we should have chosen a use case where `not` had to be used. Cases like this we want to have linting for in future, for example, instead of doing `not and ==`, we would do `or !=`. But use `not` for cases we cant avoid it, i.e `answer not in checkbox_answer_list`. In most cases, we would use `not` with an array operator. Although not technically not wrong to do `not x > 2` we should always use `x < 3` etc.

test_schemas/en/test_new_routing_and.json
```diff
@@ -0,0 +1,173 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "001",
+    "title": "Test Routing And",
+    "theme": "default",
+    "description": "A test survey for routing based on the and operator",
```
MebinAbraham, line 9: I would emphasis the operators i.e `and`. The same applies to the other schemas.

tests/functional/spec/features/routing/and.spec.js
```diff
@@ -0,0 +1,38 @@
...
+import IncorrectAnswerPage from "../../../generated_pages/new_routing_and/incorrect-answer.page";
+
+describe("Feature: Routing - And Operator", () => {
+  describe("Equals", () => {
+    describe("Given I start the and operator routing survey", () => {
+      beforeEach(() => {
+        browser.openQuestionnaire("test_new_routing_and.json");
+      });
+
+      it("When I enter both answers correctly, Then I should be routed to the correct page", () => {
```
MebinAbraham, line 13: I am not a huge for of the generic `When`, `Then`, Explicit is better than implicit. I prefer if it was more like `When I enter 123 and 321, Then ...`. The outputs of the func test become useful then and can be replayed without looking, at the file for inputs. Maybe see if others agree. Although not a blocker, if these are being kept like this it might be worth rationalizing it, as it is almost identical in all 3 files. (this might change but on my comment about the `not` operator`)

MebinAbraham (commented):
Technically does what it needs to do but just a few comments for consideration, otherwise LGTM.

test_schemas/en/test_new_routing_not.json
berroar: I've changed this now so am using testing not with `in` rather than `and`

tests/functional/spec/features/routing/and.spec.js
berroar: I've made them more specific now and the not tests have been updated.

tests/functional/spec/features/routing/and.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+      it("When I enter 555 and 321, Then I should be routed to the incorrect page", () => {
+        $(FirstNumberQuestionPage.answer1()).setValue(555);
+        $(FirstNumberQuestionPage.submit()).click();
+        $(SecondNumberQuestionPage.answer2()).setValue(321);
+        $(SecondNumberQuestionPage.submit()).click();
+        expect(browser.getUrl()).to.contain(IncorrectAnswerPage.pageName);
+      });
+
+      it("When I enter 555 and 444, Then I should be routed to the incorrect page", () => {
```
MebinAbraham, line 29: Maybe these should be `do not enter 123 ..` as to me, it reads like you need to enter those exactly number for it to be incorrect.

MebinAbraham (approved):

test_schemas/en/test_new_routing_not.json
```diff
@@ -0,0 +1,111 @@
...
+            "groups": [
+                {
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "country-checkbox",
+                            "question": {
+                                "id": "country-checkbox-question",
+                                "title": "Have you visited any of the following countries?",
+                                "type": "General",
```
MebinAbraham, line 44: Some description would help, i.e `Select/don't select India`, then it is obvious to test with.

LJBabbage (changes requested):
Works as expected, but I struggled with the schemas specially the Not, it's not obvious what it is doing. I selected India and I ended up on the submission page not knowing if that was right or wrong, I think this one needs another interstitial page. I also think each schemas (UI) needs a simple description, something like, enter both values as requested to route to the correct page etc

tests/functional/spec/features/routing/or.spec.js
```diff
@@ -0,0 +1,38 @@
...
+import IncorrectAnswerPage from "../../../generated_pages/new_routing_or/incorrect-answer.page";
+
+describe("Feature: Routing - OR Operator", () => {
+  describe("Equals", () => {
+    describe("Given I start the or operator routing survey", () => {
+      beforeEach(() => {
+        browser.openQuestionnaire("test_new_routing_or.json");
+      });
+
+      it("When I enter 123 and 321, Then I should be routed to the correct page", () => {
```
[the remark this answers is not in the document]
berroar: I changed these based on @MebinAbraham comments above to be more explicit, previously used things like `When I enter both answers incorrectly` etc. So not sure which approach to take here now

tests/functional/spec/features/routing/or.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+      it("When I enter 555 and 321, Then I should be routed to the correct page", () => {
+        $(FirstNumberQuestionPage.answer1()).setValue(555);
+        $(FirstNumberQuestionPage.submit()).click();
+        $(SecondNumberQuestionPage.answer2()).setValue(321);
+        $(SecondNumberQuestionPage.submit()).click();
+        expect(browser.getUrl()).to.contain(CorrectAnswerPage.pageName);
+      });
+
+      it("When I do not enter 123 or 321, Then I should be routed to the incorrect page", () => {
```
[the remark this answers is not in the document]
berroar: Added an additional test to each now

berroar:
> Works as expected, but I struggled with the schemas specially the Not, it's not obvious what it is doing. I selected India and I ended up on the submission page not knowing if that was right or wrong, I think this one needs another interstitial page. I also think each schemas (UI) needs a simple description, something like, enter both values as requested to route to the correct page etc

Added a new interstitial page to the `not` schema and some additional guidance to the other schemas to make it clearer

tests/functional/spec/features/routing/or.spec.js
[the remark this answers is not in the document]
LJBabbage: The problem is I have no idea if 123 and 321 is correct from this statement, that is the most important thing, not what the number is, in this number it's arbitrary, just needs to be right/expected or wrong/not expected. Now because Mebin will argue till my bedtime, lets find a middle ground. 
"When I enter 123 and 321 which are both correct, Then I should be routed to the correct page". Anything like that works. "both true", "both expected"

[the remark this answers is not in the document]
berroar: alright cool i will update it 👍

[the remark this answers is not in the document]
berroar: Have updated these now

test_schemas/en/test_new_routing_or.json
```diff
@@ -56,6 +56,13 @@
                             "type": "Question",
                             "id": "number-question-2",
                             "question": {
+                                "guidance": {
+                                    "contents": [
+                                        {
+                                            "description": "Enter 123 here OR 321 on the next question to route to the “correct” page otherwise you will be routed to the “incorrect” page"
```
LJBabbage, line 62: This description is on the wrong question

LJBabbage (changes requested):
Only a description on the wrong question, everything else looks good

LJBabbage (approved):

PR #668, opened 2021-09-22
Add type hint form module - validtors.py and duration_form.py

parimooa:

- Add type hints to `app/forms/validators.py` , `app/forms/duration_form.py` 
- `error_messages` are not optional to `FieldHandler` class
-  Updated unit test to pass `error_messages` `dict`
- Add unit tests to reflect code changes and have 100% code coverage due to typing

companion PR adding type hints to the other modules in `forms` is https://github.com/ONSdigital/eq-questionnaire-runner/pull/665
### How to review 
Run the python linter and unit tests to ensure nothing is broken. Verify where code has changed to accommodate the typing errors(raised by mypy) and previous types were updated.

### Checklist

### DIFF
--- a/app/forms/duration_form.py
+++ b/app/forms/duration_form.py
@@ -1,10 +1,18 @@
+from __future__ import annotations
+
+from typing import Callable, Mapping, Optional, Type
+
 from wtforms import Form
 
 from app.forms.fields import IntegerFieldWithSeparator
 
+ErrorMessageType = dict[str, str]
+
 
 class DurationForm(Form):
-    def validate(self, extra_validators=None):
+    def validate(
+        self, extra_validators: Optional[dict[str, list[Callable]]] = None
+    ) -> bool:
         super(DurationForm, self).validate(extra_validators)
 
         if all(not field.raw_data[0] for field in self._fields.values()):
@@ -30,18 +38,20 @@ class DurationForm(Form):
 
         return True
 
-    def _set_error(self, key):
+    def _set_error(self, key: str) -> None:
         list(self._fields.values())[0].errors = [self.answer_errors[key]]
 
     @property
-    def data(self):
-        data = super().data
+    def data(self) -> Optional[dict[str, Optional[str]]]:
+        data: dict[str, Optional[str]] = super().data
         if all(value is None for value in data.values()):
             return None
         return data
 
 
-def get_duration_form(answer, error_messages):
+def get_duration_form(
+    answer: Mapping, error_messages: ErrorMessageType
+) -> Type[DurationForm]:
     class CustomDurationForm(DurationForm):
         mandatory = answer["mandatory"]
         units = answer["units"]
@@ -56,7 +66,9 @@ def get_duration_form(answer, error_messages):
     return CustomDurationForm
 
 
-def _get_answer_errors(answer, error_messages):
+def _get_answer_errors(
+    answer: Mapping, error_messages: ErrorMessageType
+) -> dict[str, str]:
     answer_errors = error_messages.copy()
 
     if "validation" in answer and "messages" in answer["validation"]:
--- a/app/forms/field_handlers/field_handler.py
+++ b/app/forms/field_handlers/field_handler.py
@@ -1,8 +1,9 @@
 from abc import ABC
 from functools import cached_property
-from typing import Any, Optional, Union
+from typing import Mapping, Optional, Union
 
 from wtforms import Field, validators
+from wtforms.validators import Optional as OptionalValidator
 
 from app.forms.validators import ResponseRequired, format_message_with_title
 from app.questionnaire.value_source_resolver import (
@@ -19,19 +20,19 @@ class FieldHandler(ABC):
         self,
         answer_schema: dict,
         value_source_resolver: ValueSourceResolver,
-        error_messages: dict = None,
+        error_messages: dict[str, str],
         disable_validation: bool = False,
         question_title: str = None,
     ):
         self.answer_schema = answer_schema
         self.value_source_resolver = value_source_resolver
-        self.error_messages = error_messages or {}
+        self.error_messages = error_messages
         self.disable_validation = disable_validation
         self.question_title = str(question_title)
 
-        self.validation_messages = self.answer_schema.get("validation", {}).get(
-            "messages", {}
-        )
+        self.validation_messages: Mapping[str, str] = self.answer_schema.get(
+            "validation", {}
+        ).get("messages", {})
 
     @cached_property
     def validators(self) -> list[validators.Optional]:
@@ -47,12 +48,13 @@ class FieldHandler(ABC):
     def guidance(self) -> str:
         return self.answer_schema.get("guidance", "")
 
-    def get_validation_message(self, message_key: str) -> Optional[str]:
-        return self.validation_messages.get(message_key) or self.error_messages.get(
-            message_key
+    def get_validation_message(self, message_key: str) -> str:
+        return (
+            self.validation_messages.get(message_key)
+            or self.error_messages[message_key]
         )
 
-    def get_mandatory_validator(self) -> Union[ResponseRequired, Optional[Any]]:
+    def get_mandatory_validator(self) -> Union[ResponseRequired, OptionalValidator]:
         if self.answer_schema["mandatory"] is True:
             mandatory_message = self.get_validation_message(self.MANDATORY_MESSAGE_KEY)
 
--- a/app/forms/validators.py
+++ b/app/forms/validators.py
@@ -1,19 +1,31 @@
+from __future__ import annotations
+
 import re
 from datetime import datetime, timezone
 from decimal import Decimal, InvalidOperation
+from typing import TYPE_CHECKING, Iterable, List, Mapping, Optional, Sequence, Union
 
 import flask_babel
 from babel import numbers
 from dateutil.relativedelta import relativedelta
 from flask_babel import ngettext
+from flask_wtf import FlaskForm
 from structlog import get_logger
-from wtforms import validators
+from wtforms import Field, StringField, validators
 from wtforms.compat import string_types
 
 from app.forms import error_messages
+from app.forms.fields import (
+    DateField,
+    DecimalFieldWithSeparator,
+    IntegerFieldWithSeparator,
+)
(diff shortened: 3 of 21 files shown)


berroar:
When I run mypy against the validators.py I still get the following errors:

app/forms/validators.py
```diff
@@ -52,15 +61,15 @@ class ResponseRequired:
...
+    def __init__(self, message: str, strip_whitespace: bool = True):
         self.message = message
 
         if strip_whitespace:
             self.string_check = lambda s: s.strip()
         else:
             self.string_check = lambda s: s
 
-    def __call__(self, form, field):
+    def __call__(self, form: FlaskForm, field: StringField) -> NoReturn:
```
MebinAbraham, line 72: Field could be many different types that string field?.

MebinAbraham (commented):
There are outstanding issues which aren't picked up by mypy which can be sorted by updating the mypy.ini config to pick up these files, for example:

```ini
[mypy-app.forms.validators]
disallow_untyped_defs = True
warn_return_any = True
```

Will do a full pass once this has been added as it is hiding some issues I think.

app/forms/validators.py
```diff
@@ -22,12 +27,16 @@
...
+PeriodType = Mapping[str, int]
+
 
 class NumberCheck:
-    def __init__(self, message=None):
+    def __init__(self, message: OptionalMessage = None):
         self.message = message or error_messages["INVALID_NUMBER"]
 
-    def __call__(self, form, field):
+    def __call__(self, form: FlaskForm, field: IntegerField) -> NoReturn:
```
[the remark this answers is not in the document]
parimooa: Yes, that throws error as circular reference hence had to use FlaskForm instead  
> Using `QuestionnaireForm` would be more appropriate, since we inherit from `FlaskForm` in that. Did you try questionnaire form?

app/forms/duration_form.py
```diff
@@ -30,18 +40,20 @@ def validate(self, extra_validators=None):
 
         return True
 
-    def _set_error(self, key):
+    def _set_error(self, key: str):
         list(self._fields.values())[0].errors = [self.answer_errors[key]]
 
     @property
-    def data(self):
+    def data(self) -> Union[str, None]:
```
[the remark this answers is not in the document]
parimooa: good spot i missed that out

app/forms/validators.py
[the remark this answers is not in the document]
MebinAbraham: That is a common issue when typing, you can get it working by importing questionnaire form when typing checking only. So to use `QuestionnaireForm` you would need to import `TYPE_CHECKING` from `typing. Then you can reference it using quotes.

```py

from typing import TYPE_CHECKING
...

if TYPE_CHECKING:
    from app.forms.questionnaire_form import QuestionnaireForm
    
def __call__(self, form: "QuestionnaireForm", field: IntegerField) -> None:
	...
```

app/forms/field_handlers/field_handler.py
```diff
@@ -47,12 +48,12 @@ def label(self) -> Optional[str]:
     def guidance(self) -> str:
         return self.answer_schema.get("guidance", "")
 
-    def get_validation_message(self, message_key: str) -> Optional[str]:
+    def get_validation_message(self, message_key: str) -> Union[str, Any]:
         return self.validation_messages.get(message_key) or self.error_messages.get(
             message_key
         )
```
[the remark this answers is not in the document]
parimooa: i tried that but it break whole loads of tests in `field handler` because of this piece of code in here    https://github.com/ONSdigital/eq-questionnaire-runner/blob/877ca2e025e56f6d59cee2e868cf4d611e695036/app/forms/field_handlers/number_handler.py#L90
with error as ` AttributeError: 'NoneType' object has no attribute 'copy'
as it is inherited from.`    
`FieldHandler`  class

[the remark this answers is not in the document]
MebinAbraham: That will be because the test will be passing `None` when it no longer should. They will need updating to pass an mapping in.

app/forms/validators.py
```diff
@@ -244,20 +259,20 @@ def __call__(self, form, field):
...
         self.minimum_date = minimum_date
         self.maximum_date = maximum_date
         self.date_format = date_format
 
-    def __call__(self, form, field):
+    def __call__(self, form: "QuestionnaireForm", field: DateField) -> None:
         date = convert_to_datetime(form.data)
 
-        if self.minimum_date:
+        if self.minimum_date and date:
```
berroar, line 275: The nested if statements here can be merged

app/forms/validators.py
```diff
@@ -280,22 +295,31 @@ def __call__(self, form, field):
...
-    def __call__(self, form, from_field, to_field):
+    def __call__(
+        self, form: "QuestionnaireForm", from_field: DateField, to_field: DateField
+    ) -> None:
         from_date = convert_to_datetime(from_field.data)
         to_date = convert_to_datetime(to_field.data)
 
-        if from_date >= to_date:
-            raise validators.ValidationError(self.messages["INVALID_DATE_RANGE"])
+        if from_date and to_date:
```
berroar, line 320: And can be merged here too

app/forms/validators.py
```diff
@@ -1,33 +1,45 @@
...
+PeriodType = Mapping[str, int]
+
 
 class NumberCheck:
-    def __init__(self, message=None):
+    def __init__(self, message: str = None):
         self.message = message or error_messages["INVALID_NUMBER"]
 
-    def __call__(self, form, field):
+    def __call__(self, form: FlaskForm, field: IntegerField) -> None:
```
MebinAbraham, line 42: It could be a `DecimalField` also, but worth using our classes if it causes no issues.
```suggestion
    def __call__(self, form: FlaskForm, field: Union[DecimalFieldWithSeparator, IntegerFieldWithSeparator]) -> None:
```

app/forms/validators.py
```diff
@@ -86,57 +98,58 @@ class NumberRange:
 
     def __init__(
         self,
-        minimum=None,
-        minimum_exclusive=False,
-        maximum=None,
-        maximum_exclusive=False,
-        messages=None,
-        currency=None,
+        minimum: NumType = None,
```
MebinAbraham, line 101: Minor but as per our style guide, we should wrap any types initialised as `None` in `Optional`. There are a couple of occurrence like this which we should look to make consistent. https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/doc/python-type-hinting.md#variables

app/forms/validators.py
```diff
@@ -86,57 +98,58 @@ class NumberRange:
...
         self.maximum_exclusive = maximum_exclusive
-        self.messages = {**error_messages, **(messages or {})}
+        self.messages: Mapping[str, str] = {**error_messages, **(messages or {})}
         self.currency = currency
 
-    def __call__(self, form, field):
-        value = field.data
-        error_message = None
+    def __call__(
+        self, form: "QuestionnaireForm", field: DecimalFieldWithSeparator
```
MebinAbraham, line 116: Should this also not be both types?
```suggestion
        self, form: "QuestionnaireForm", field: Union[DecimalFieldWithSeparator, IntegerFieldWithSeparator]
```

app/forms/validators.py
```diff
@@ -86,57 +98,58 @@ class NumberRange:
...
+    ) -> None:
+        value: Union[int, Decimal] = field.data
         if value is not None:
-            if self.minimum is not None:
-                error_message = self.validate_minimum(value)
-            if error_message is None and self.maximum is not None:
+            error_message = self.validate_minimum(value)
+            if error_message is None:
                 error_message = self.validate_maximum(value)
```
MebinAbraham, line 122: This could now be simplified:
```suggestion
            error_message = self.validate_minimum(value) or self.validate_maximum(value)
```

app/forms/validators.py
```diff
@@ -223,10 +238,10 @@ def __call__(self, form, field):
 
 
 class DateCheck:
-    def __init__(self, message=None):
+    def __init__(self, message: Optional[str] = None):
         self.message = message or error_messages["INVALID_DATE"]
 
-    def __call__(self, form, field):
+    def __call__(self, form: "QuestionnaireForm", field: DateField) -> None:
```
MebinAbraham, line 244: Field for these dates are actually `StringField` instead of our custom date fields, seems to be due to the way the date field is set up.
https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/forms/fields/month_year_date_field.py#L9-L12

app/forms/validators.py
```diff
@@ -244,58 +259,65 @@ def __call__(self, form, field):
...
+        minimum_date: Optional[datetime] = None,
+        maximum_date: Optional[datetime] = None,
     ):
         self.messages = {**error_messages, **(messages or {})}
         self.minimum_date = minimum_date
         self.maximum_date = maximum_date
         self.date_format = date_format
 
-    def __call__(self, form, field):
+    def __call__(self, form: "QuestionnaireForm", field: DateField) -> None:
```
MebinAbraham, line 272: I believe this is also a `StringField`.

app/forms/validators.py
```diff
@@ -244,58 +259,65 @@ def __call__(self, form, field):
...
+        period_min: Optional[dict[str, int]] = None,
+        period_max: Optional[dict[str, int]] = None,
+    ):
         self.messages = {**error_messages, **(messages or {})}
         self.period_min = period_min
         self.period_max = period_max
 
-    def __call__(self, form, from_field, to_field):
+    def __call__(
+        self, form: "QuestionnaireForm", from_field: DateField, to_field: DateField
```
MebinAbraham, line 313: `DateField`s here seem to be correct.

MebinAbraham (commented):
Just a few minor issues, otherwise 👍

app/forms/duration_form.py
```diff
@@ -1,10 +1,19 @@
+from __future__ import annotations
+
+from collections.abc import Callable
```
berroar, line 3: Minor - but any reason why we're aren't using typing Callable here?

berroar (approved):

app/forms/duration_form.py
parimooa: Updated  `callable` is now called from `typing`.

PR #670, opened 2021-09-22
Routing on checkbox count

rmccar:
Adds the count operator and schema and tests to go alongside it.
Also changes to validator [here](https://github.com/ONSdigital/eq-questionnaire-validator/pull/110)

### How to review 
- All tests pass
- Changes make sense
- Launch new schema and see that the functionality works

### Checklist


rmccar:
[Routing on count of checkbox answer (2 SP)](https://trello.com/c/q9ngIbPM/4796-routing-on-count-of-checkbox-answer-2-sp)

rmccar:
Reminder to change validator branch back to "latest"

rmccar (commented):
Reminder to change validator branch back to "latest"

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=routing-on-checkbox-count
```
rmccar, line 2:
```suggestion
tag=latest
```

test_schemas/en/test_routing_checkbox_count.json
```diff
@@ -0,0 +1,155 @@
+{
```
MebinAbraham, line 1: Worth prefixing with `test_new_routing` to make deprecation of the old one easier.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -811,3 +811,35 @@ def test_raises_exception_when_bad_operand_type():
...
+        ),
+    ],
+)
+def test_answer_source_count(rule, expected_result):
+    when_rule_evaluator = get_when_rule_evaluator(
+        answer_store=AnswerStore(
+            [
+                {
+                    "answer_id": "some-answer",
+                    "value": (["array element 1", "array element 2"]),
```
MebinAbraham, line 840: This shouldn't be nested.
```suggestion
                    "value": ["array element 1", "array element 2"],
```

tests/functional/spec/features/routing/checkbox_count_routing.spec.js
```diff
@@ -0,0 +1,39 @@
+import ToppingCheckboxPage from "../../../generated_pages/routing_checkbox_count/topping-checkbox.page.js";
+import CorrectAnswerPage from "../../../generated_pages/routing_checkbox_count/correct-answer.page";
+import IncorrectAnswerPage from "../../../generated_pages/routing_checkbox_count/incorrect-answer.page";
+
+describe("Test routing using number of checkboxes checked", () => {
+  beforeEach(() => {
+    browser.openQuestionnaire("test_routing_checkbox_count.json");
+  });
+
+  it("Given a user selects 2 checkboxes, they should be routed to the correct page", () => {
```
MebinAbraham, line 10: The pattern should be `Given, When, Then`.

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -330,3 +330,12 @@ def test_nonetype_operands_for_comparison_operators(operator_name, operands):
 def test_nonetype_operands_for_array_operators(operator_name, operands):
     operator = Operator(operator_name)
     assert operator.evaluate(operands) is False
+
+
+@pytest.mark.parametrize(
+    "operands, expected_result",
+    [([("tuple element 1", "tuple element 2")], 2)],
```
MebinAbraham, line 337: As discussed, this should test a list as well. List covers answers, the tuple covers raw schema lists.

test_schemas/en/test_routing_checkbox_count.json
```diff
@@ -0,0 +1,155 @@
...
+                                        "description": {
+                                            "text": "You were asked to select <em>2</em> toppings and you selected <em>{answer}</em>.",
+                                            "placeholders": [
+                                                {
+                                                    "placeholder": "answer",
+                                                    "value": {
+                                                        "source": "answers",
+                                                        "identifier": "topping-checkbox-answer"
+                                                    }
+                                                }
```
MebinAbraham, line 142: Piping checkbox without any formatting looks odd, it outputs the Markup etc. Since we don't have count for placeholder, you can use `format_list` placeholder to make it look more appropriate.
```suggestion
                                                {
                                                    "placeholder": "answer",
                                                    "transforms": [
                                                        {
                                                            "transform": "format_list",
                                                            "arguments": {
                                                                "list_to_format": {
                                                                    "source": "answers",
                                                                    "identifier": "topping-checkbox-answer"
                                                                }
                                                            }
                                                        }
                                                    ]
                                                }
```

test_schemas/en/test_routing_checkbox_count.json
```diff
@@ -0,0 +1,155 @@
...
+                                        "description": {
+                                            "text": "You were asked to select <em>2</em> toppings but you actually selected <em>{answer}</em>.",
+                                            "placeholders": [
+                                                {
+                                                    "placeholder": "answer",
+                                                    "value": {
+                                                        "source": "answers",
+                                                        "identifier": "topping-checkbox-answer"
+                                                    }
+                                                }
```
MebinAbraham, line 114: Piping checkbox without any formatting looks odd, it outputs the Markup etc. Since we don't have count for placeholder, you can use `format_list` placeholder to make it look more appropriate.
```suggestion
                                                {
                                                    "placeholder": "answer",
                                                    "transforms": [
                                                        {
                                                            "transform": "format_list",
                                                            "arguments": {
                                                                "list_to_format": {
                                                                    "source": "answers",
                                                                    "identifier": "topping-checkbox-answer"
                                                                }
                                                            }
                                                        }
                                                    ]
                                                }
```

test_schemas/en/test_routing_checkbox_count.json
rmccar: Resolving because its a duplicate

app/questionnaire/routing/operations.py
```diff
@@ -49,6 +49,10 @@ def evaluate_or(values: Iterable[bool]) -> bool:
     return any(iter(values))
 
 
+def evaluate_count(values: Optional[Sized]) -> int:
+    return len(values) if values else 0
```
LJBabbage, line 53: We don't seem to exercise the else in any unit tests. This is strange as line coverage is 100%

LJBabbage (changes requested):

LJBabbage:
Should try to avoid force-pushed if at all possible, it can causes issue when reviewing

rmccar:
> Should try to avoid force-pushed if at all possible, it can causes issue when reviewing

Ok I was doing that so I could squash my commits

LJBabbage:
> > Should try to avoid force-pushed if at all possible, it can causes issue when reviewing
> 
> Ok I was doing that so I could squash my commits

We always do it at the end. No-one cares if there are few extra. Keep in mind that James once had one with 130+ commits, that always helps me if I have done a few random ones

tests/functional/spec/features/routing/checkbox_count_routing.spec.js
```diff
@@ -0,0 +1,39 @@
...
+
+  it("Given a user selects 2 checkboxes, When they submit, Then they should be routed to the correct page", () => {
+    $(ToppingCheckboxPage.cheese()).click();
+    $(ToppingCheckboxPage.ham()).click();
+    $(ToppingCheckboxPage.submit()).click();
+
+    expect(browser.getUrl()).to.contain(CorrectAnswerPage.pageName);
+  });
+
+  it("Given a user selects no checkboxes, When they submit, Then they should be routed to the incorrect page", () => {
```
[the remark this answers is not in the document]
LJBabbage: Ah, I know why, because you are only checking the url, which would still be the same, but with the "Sorry, there is a problem with this service" on the page instead

test_schemas/en/test_new_routing_checkbox_count.json
```diff
@@ -0,0 +1,162 @@
...
+                                "contents": [
+                                    {
+                                        "description": {
+                                            "text": "You were asked to select <em>2</em> toppings but you actually selected <em>{answer}</em>.",
+                                            "placeholders": [
+                                                {
+                                                    "placeholder": "answer",
+                                                    "transforms": [
+                                                        {
+                                                            "transform": "format_list",
```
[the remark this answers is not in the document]
rmccar: Added an if to the "format_list" function to deal with this and a test to go alongside it

tests/functional/spec/features/routing/checkbox_count_routing.spec.js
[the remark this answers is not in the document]
rmccar: Added another transform "list_item_count" to fix this and show the number of the items selected instead of a list. I feel like this fits better with the test too.

app/questionnaire/placeholder_transforms.py
```diff
@@ -32,12 +32,15 @@ def format_date(self, date_to_format, date_format):
...
+        if list_to_format is not None:
+            formatted_list = "<ul>"
+            for item in list_to_format:
+                formatted_list += f"<li>{item}</li>"
+            formatted_list += "</ul>"
 
-        return formatted_list
+            return formatted_list
+        else:
+            return ""
```
LJBabbage, line 43: When there is a return in the if then you don't need the else
```
if list_to_format is not None:
     formatted_list = "<ul>"
     for item in list_to_format:
          formatted_list += f"<li>{item}</li>"
     formatted_list += "</ul>"
     return formatted_list
return ""

File: app/questionnaire/placeholder_transforms.py [hunk 12]
```diff
@@ -32,12 +32,15 @@ def format_date(self, date_to_format, date_format):
 
     @staticmethod
     def format_list(list_to_format):
```
LJBabbage, line 34: The var and return is missing typing

app/questionnaire/placeholder_transforms.py
```diff
@@ -227,3 +230,10 @@ def format_name(first_name, middle_names, last_name, include_middle_names=False)
     @staticmethod
     def _create_hyperlink(href: str, link_text: str) -> str:
         return f'<a href="{href}">{link_text}</a>'
+
+    @staticmethod
+    def list_item_count(list_to_count) -> int:
```
LJBabbage, line 235: This can go on one line and be simplified. The var is also missing typing 
```
return len(list_to_count) if list_to_count else 0
```

tests/functional/spec/features/routing/checkbox_count_routing.spec.js
```diff
@@ -0,0 +1,39 @@
...
+    $(ToppingCheckboxPage.submit()).click();
+
+    expect(browser.getUrl()).to.contain(IncorrectAnswerPage.pageName);
```
LJBabbage, line 21: We probably need to verify some content on the page, just to make sure it's not a 500 being presented on the same url

app/questionnaire/placeholder_transforms.py
LJBabbage: You don't need the if, it's in the return

app/questionnaire/placeholder_transforms.py
```diff
@@ -39,12 +39,13 @@ def format_date(self, date_to_format: str, date_format: str) -> str:
...
-        formatted_list += "</ul>"
-
-        return formatted_list
+        if list_to_format is not None:
+            formatted_list = "<ul>"
+            for item in list_to_format:
+                formatted_list += f"<li>{item}</li>"
+            formatted_list += "</ul>"
+            return formatted_list
+        return ""
```
MebinAbraham, line 48: I think this should be reverted, I can't see a real use case where we would ask to format an empty list and accept an empty string as being valid. Given there are no requirements around it and given it is not used, I don't think it should be changed. I think the list count you added is fine for the purposes of demoing the routing behaviour.

app/questionnaire/placeholder_transforms.py
```diff
@@ -243,3 +244,8 @@ def format_name(
     @staticmethod
     def _create_hyperlink(href: str, link_text: str) -> str:
         return f'<a href="{href}">{link_text}</a>'
+
+    @staticmethod
+    def list_item_count(list_to_count) -> int:
+        if list_to_count is not None:
+            return len(list_to_count) if list_to_count else 0
```
MebinAbraham, line 251: This should always return an `int`. The first if is not needed, but this could be simplified with the need for `if`.
```suggestion
    def list_item_count(list_to_count: Optional[Sized]) -> int:
        return len(list_to_count or [])
```

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -331,3 +331,16 @@ def test_nonetype_operands_for_comparison_operators(operator_name, operands):
...
     operator = Operator(operator_name)
     assert operator.evaluate(operands) is False
+
+
+@pytest.mark.parametrize(
+    "operands, expected_result",
+    [
+        ([("tuple element 1", "tuple element 2")], 2),
+        ([["list element 1", "list element 2"]], 2),
+        ([None], 0),
```
MebinAbraham, line 341: I think it is worth adding an empty collection also.
```suggestion
        ([("tuple element 1", "tuple element 2")], 2),
        ([["list element 1", "list element 2"]], 2),
        ([[]], 0),
        ([None], 0),```

**Review by MebinAbraham — COMMENTED** [review]
Looking good, just few minor comments.

**Review by LJBabbage — APPROVED** [review]

File: tests/app/questionnaire/test_placeholder_transforms.py [hunk 18]
```diff
@@ -260,3 +261,13 @@ def test_telephone_number_link(self):
...
+
+    def test_list_item_count(self):
+
+        assert (
+            self.transforms.list_item_count(
+                ["Alice Aardvark", "Bob Berty Brown", "Dave Dixon Davies"]
+            )
+            == 3
+        )
+        assert self.transforms.list_item_count([]) == 0
```
MebinAbraham, line 273: Worth asserting `NoneType` as well, as unanswered will be None.
```py
assert self.transforms.list_item_count(None) == 0
```

tests/functional/spec/features/routing/checkbox_count_routing.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    browser.openQuestionnaire("test_new_routing_checkbox_count.json");
+  });
+
+  it("Given a user selects 2 checkboxes, When they submit, Then they should be routed to the correct page", () => {
+    $(ToppingCheckboxPage.cheese()).click();
+    $(ToppingCheckboxPage.ham()).click();
+    $(ToppingCheckboxPage.submit()).click();
+
+    expect(browser.getUrl()).to.contain(CorrectAnswerPage.pageName);
+    expect($("body").getText()).to.have.string("You were asked to select 2 toppings and you selected 2.");
```
MebinAbraham, line 16: I think fetching fetching from body should be avoiding, we should use a selector we have. If you don't have the selector for description, you can use the `questionTitle()` selector. You will need to update the title for `Correct` to like `You selected 2 toppings`. Then this would be like:
```js
expect($(CorrectAnswerPage.questionText()).getText()).to.have.string("You selected 2 toppings");
```

tests/functional/spec/features/routing/checkbox_count_routing.spec.js
```diff
@@ -0,0 +1,43 @@
+import ToppingCheckboxPage from "../../../generated_pages/new_routing_checkbox_count/topping-checkbox.page.js";
```
MebinAbraham, line 1: `_routing` can be dropped from the name since it is in the `routing` folder.

rmccar: Taking `_routing` away makes it not conform with other schemas using the same naming convention. I think this is something we should deal with in the mop up card

MebinAbraham: I only see 2 spec files with `_routing` one has already been dropped in my PR, and the other could be dropped in Ash's PR given he is making changes there. Given there is only one outstanding, I think it makes sense to update.

app/questionnaire/routing/operations.py
```diff
@@ -49,6 +49,10 @@ def evaluate_or(values: Iterable[bool]) -> bool:
     return any(iter(values))
 
 
+def evaluate_count(values: Optional[Sized]) -> int:
+    return len(values or [])
+
```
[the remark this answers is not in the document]
MebinAbraham: @rmccar For this, you would want to use `evaluate_count` from `operations.py` in `placeholder_transforms.py` as operations is the longer-term goal.

PR #671, opened 2021-09-28
Add dynamic date placeholder transform

jamesnunn:
Adds 2 transforms: 
- Generate start and end dates of a range given a reference date, number of weeks to offset and number of days to include in the range.
- Format a pair of dates representing a date range, variably including the month/year portions of the first date of the range.

### How to review 

### Checklist

### DIFF
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -1,5 +1,4 @@
 from datetime import datetime, timedelta, timezone
-from typing import Optional
 from urllib.parse import quote
 
 from babel.dates import format_datetime
@@ -36,7 +35,7 @@ class PlaceholderTransforms:
     def format_currency(self, number=None, currency="GBP"):
         return format_currency(number, currency, locale=self.locale)
 
-    def format_date(self, date_to_format, date_format):
+    def format_date(self, date_to_format: str, date_format: str) -> str:
         date_to_format = datetime.strptime(
             date_to_format, self.input_date_format
         ).replace(tzinfo=timezone.utc)
@@ -146,7 +145,7 @@ class PlaceholderTransforms:
         offset_full_weeks: int,
         days_in_range: int,
         first_day_of_week: str = "monday",
-    ):
+    ) -> tuple[datetime, datetime]:
         """Generate a start and end date for a date range given a reference date,
         weeks prior and number of days in range.
 
@@ -182,6 +181,34 @@ class PlaceholderTransforms:
         )
 
         return tuple(sorted([first_day_of_prior_full_week, last_day_of_range]))
+
+    @staticmethod
+    def format_date_range_pair(date_pair: tuple[datetime, datetime]) -> str:
+        """Format a pair of dates as a string, clarifying differences in month or year.
+
+        E.g.
+            Monday 1 to Sunday 8 September 2021
+            Monday 29 September to Sunday 6 October 2021
+            Monday 29 December 2021 to Sunday 6 January 2022
+
+        :param date_pair: Pair of datetime objects representing a date range.
+        :type date_pair: tuple[datetime, datetime]
+        :return: String containing the date range as text.
+        :rtype: str
+        """
+        date1, date2 = date_pair
+        date1_format = "%A %d".replace(" 0", "")
+        date2_format = "%A %d %B %Y"
+
+        if date1.year != date2.year:
+            date1_format += " %B %Y"
+        elif date1.month != date2.month:
+            date1_format += " %B"
+
+        date1_formatted = date1.strftime(date1_format)
+        date2_formatted = date2.strftime(date2_format)
+
+        return f"{date1_formatted} to {date2_formatted}"
 
     @staticmethod
     def parse_date(date):
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -1,4 +1,5 @@
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
+from typing import Optional
 from urllib.parse import quote
 
 from babel.dates import format_datetime
@@ -7,6 +8,17 @@ from dateutil.relativedelta import relativedelta
 from flask_babel import ngettext
 
 from app.settings import DEFAULT_LOCALE
+
+
+DAYS = {
+    "monday": 0,
+    "tuesday": 1,
+    "wednesday": 2,
+    "thursday": 3,
+    "friday": 4,
+    "saturday": 5,
+    "sunday": 6,
+}
 
 
 class PlaceholderTransforms:
@@ -127,6 +139,43 @@ class PlaceholderTransforms:
             "{number_of_days} day", "{number_of_days} days", time.days
         )
         return day_string.format(number_of_days=time.days)
+
+    @staticmethod
+    def generate_date_range(
+        reference_date: str,
+        number_of_weeks_prior: int,
+        days_in_range: int,
+        first_day_of_week: Optional[str] = "monday",
+    ):
+        """Generate a start and end date for a date range given a reference date,
+        weeks prior and number of days in range.
+
+        :param reference_date: The date to reference the date range from, can be YYY&-MM-DD, YY-MM or 'now'.
+        :type reference_date: datetime
+        :param number_of_weeks_prior: Number of weeks prior to the refence date to start the date range.
+        :type number_of_weeks_prior: int
+        :param days_in_range: Number of days in the range, including the resolved start date.
+        :type days_in_range: int
+        :param first_day_of_week: Which day of the week should be considered the first (default 'monday').
+            This is the day of the week which is selected as the start of the range.
+        :type first_day_of_week: str
+        :return: The start and end datetime objects of the range.
+        :rtype: Tuple[datetime, datetime]
+        """
+        reference_date = PlaceholderTransforms.parse_date(reference_date)
+        first_day_of_week = DAYS[first_day_of_week]
+
+        first_day_of_last_partial_week = reference_date - timedelta(
+            days=reference_date.weekday() + first_day_of_week
+        )
+        first_day_of_prior_full_week = first_day_of_last_partial_week - timedelta(
+            days=number_of_weeks_prior * 7
+        )
+        last_day_of_range = first_day_of_prior_full_week + timedelta(
+            days=days_in_range - 1
+        )
+
+        return (first_day_of_prior_full_week, last_day_of_range)
 
     @staticmethod
     def parse_date(date):
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -143,18 +143,18 @@ class PlaceholderTransforms:
     @staticmethod
     def generate_date_range(
         reference_date: str,
-        number_of_weeks_prior: int,
+        offset_full_weeks: int,
         days_in_range: int,
-        first_day_of_week: Optional[str] = "monday",
+        first_day_of_week: str = "monday",
     ):
         """Generate a start and end date for a date range given a reference date,
         weeks prior and number of days in range.
 
         :param reference_date: The date to reference the date range from, can be YYY&-MM-DD, YY-MM or 'now'.
         :type reference_date: datetime
-        :param number_of_weeks_prior: Number of weeks prior to the refence date to start the date range.
-        :type number_of_weeks_prior: int
-        :param days_in_range: Number of days in the range, including the resolved start date.
+        :param offset_full_weeks: Number of full weeks to offset from the reference date to start the date range.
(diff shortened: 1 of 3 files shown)


tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -260,3 +260,141 @@ def test_telephone_number_link(self):
...
+                first_day_of_week,
+                (date_from, date_to),
+            ) in enumerate(params):
+                expected = (date_from, date_to)
+                actual = self.transforms.generate_date_range(
+                    ref_date, weeks_prior, day_range, first_day_of_week
+                )
+                self.assertEqual(expected, actual, f"FAIL: Item {i}")
+
+    def test_generate_date_range_all_weekdays(self):
```
petechd, line 279: I don't think we test the `lower()` method you added, so can we add some more (or refactor) test scenarios? Something like:
`("2021-09-27", -1, 7, "Monday", ("2021-09-20", "2021-09-26")),`
 ` ("2021-09-27", -2, 7, "MoNdAy", ("2021-09-13", "2021-09-19")),`

app/questionnaire/placeholder_transforms.py
```diff
@@ -139,6 +149,79 @@ def calculate_date_difference(first_date: str, second_date: str) -> str:
...
+        :rtype: str
+        """
+        start_date, end_date = list(map(PlaceholderTransforms.parse_date, date_range))
+        start_date_format = "%A %d"
+        end_date_format = "%A %d %B %Y"
+
+        if start_date.year != end_date.year:
+            start_date_format += " %B %Y"
+        elif start_date.month != end_date.month:
+            start_date_format += " %B"
```
MebinAbraham, line 218: On the card checklist, it says `Date format to be <day-of-the-week> <date-of-month> <month> <year> for both from and to`, so is this bit of code relevant anymore?. Worth confirm with Donna what is on the card is still correct or whether they still want this behaviour.

app/questionnaire/placeholder_transforms.py
```diff
@@ -139,6 +149,79 @@ def calculate_date_difference(first_date: str, second_date: str) -> str:
...
+
+        start_date_formatted = start_date.strftime(start_date_format).replace(" 0", " ")
```
MebinAbraham, line 220: I think a comment is useful here as it is not obvious, why the replace is being done. Python support `%-d` to strip leading zero but looks like it doesn't work on Windows.

app/questionnaire/placeholder_transforms.py
```diff
@@ -139,6 +149,79 @@ def calculate_date_difference(first_date: str, second_date: str) -> str:
...
+        end_date_formatted = end_date.strftime(end_date_format).replace(" 0", " ")
```
MebinAbraham, line 221: We should still use `flask_babel.format_date`, as regardless of what happens with translations, that would have to be kept as is. (Don't need to test the translations)

test_schemas/en/test_generate_date_range.json
```diff
@@ -0,0 +1,223 @@
+{
```
MebinAbraham, line 1: We group schemas with a prefix relevant to the concept tested in the schema, so for this it would be like `test_placeholder_date_range` etc.

test_schemas/en/test_generate_date_range.json
```diff
@@ -0,0 +1,223 @@
...
+                                "answers": [
+                                    {
+                                        "id": "date-range-month-answer",
+                                        "label": "How many days did you work during this period?",
+                                        "mandatory": false,
+                                        "type": "Number"
+                                    }
+                                ]
+                            }
+                        },
```
MebinAbraham, line 183: I don't think these should live in schema, these should be just unit tested.

test_schemas/en/test_generate_date_range.json
```diff
@@ -0,0 +1,223 @@
...
+                                            "transforms": [
+                                                {
+                                                    "transform": "format_date_range_pair",
+                                                    "arguments": {
+                                                        "date_range": {
+                                                            "value": ["2019-05-01", "2020-05-01"]
+                                                        }
+                                                    }
+                                                }
+                                            ]
```
MebinAbraham, line 203: For this, if we should drive it based on 2 dynamic inputs, could be 2 answers or you could use metadata dates we have in scope.
```json
{
    "source": "metadata",
    "identifier": "ref_p_start_date"
}
```
There is also an `ref_p_end_date`. To get more info on how it is used in a schema, have a look at `test_date_validation_yyyy_combined.json`.

app/questionnaire/placeholder_transforms.py
```diff
@@ -139,6 +149,79 @@ def calculate_date_difference(first_date: str, second_date: str) -> str:
...
+        reference_datetime = PlaceholderTransforms.parse_date(reference_date)
+        try:
+            first_day_of_week_idx = DAYS[first_day_of_week.lower()]
+        except KeyError as err:
+            raise KeyError(f"'{first_day_of_week}' is not a valid weekday") from err
+
+        if days_in_range < 1:
+            raise ValueError("'days_in_range' must be a positive integer")
+
+        first_day_of_last_partial_week = reference_datetime - timedelta(
```
MebinAbraham, line 183: Is this now always the `current_week` of the reference date?. It is not the last week?.

MebinAbraham (commented):
Initial pass, haven't looked at tests yet. Will re-review based on any changes to validator as that would change runner's implementation.

app/questionnaire/placeholder_transforms.py
```diff
@@ -139,6 +149,79 @@ def calculate_date_difference(first_date: str, second_date: str) -> str:
...
+        :type offset_full_weeks: int
+        :param days_in_range: Number of days in the range, including the start date, must be a positive integer.
+        :type days_in_range: int
+        :param first_day_of_week: Which day of the week should be considered the first (default 'monday').
+            This is the day of the week which is selected as the start of the range.
+        :type first_day_of_week: str
+        :return: The start and end datetime objects of the range.
+        :rtype: Tuple[datetime, datetime]
+        """
+        reference_datetime = PlaceholderTransforms.parse_date(reference_date)
```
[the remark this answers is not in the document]
jamesnunn: Shouldn't this suggestion be part of that card to support `started_at` then?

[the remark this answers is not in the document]
MebinAbraham: Ya, we can make it a task on that card, given this card no longer is responsible for the "started_at" value source. The work has already started, so we could just have a mop up card.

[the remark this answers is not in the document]
MebinAbraham: Have added a card to the backlog to add support for placeholders with ISO date strings.

app/questionnaire/placeholder_transforms.py
jamesnunn: Yes it is the current week - current week is the same as last_partial_week. I can rename to clarify

app/questionnaire/placeholder_transforms.py
```diff
@@ -139,6 +149,79 @@ def calculate_date_difference(first_date: str, second_date: str) -> str:
...
+            Monday 29 December 2021 to Sunday 6 January 2022
+
+        :param date_range: Pair of date strings representing a date range.
+        :type date_range: tuple[str, str]
+        :return: String containing the date range as text.
+        :rtype: str
+        """
+        start_date, end_date = list(map(PlaceholderTransforms.parse_date, date_range))
+        start_date_format = "%A %d"
+        end_date_format = "%A %d %B %Y"
```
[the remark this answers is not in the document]
MebinAbraham: We should try to make it generic if we can to support other surveys but as you point out the tricky bit is the format needing to change, therefore I am happy to keep it as is and expand it when needed. Confirmed with Donna that this single format should be enough for now.

app/questionnaire/placeholder_transforms.py
MebinAbraham: Could you update the card accordingly, as that checklist item is misleading?.

test_schemas/en/test_placeholder_transform_date_range_bounds.json
```diff
@@ -0,0 +1,174 @@
...
+                            "id": "date-question",
+                            "question": {
+                                "id": "primary-date-question",
+                                "title": "Please enter a reference date to generate a date range",
+                                "type": "General",
+                                "answers": [
+                                    {
+                                        "id": "reference-date",
+                                        "label": "Reference Date",
+                                        "description": "This scenario uses 2 prior full week offset, 15 day range, first day of the week monday",
```
MebinAbraham, line 51: The actual offset is set to `-1`.

app/questionnaire/placeholder_transforms.py
```diff
@@ -139,6 +149,80 @@ def calculate_date_difference(first_date: str, second_date: str) -> str:
         )
         return day_string.format(number_of_days=time.days)
 
+    def date_range_bounds(
+        self,
+        reference_date: str,
+        offset_full_weeks: int,
+        days_in_range: int,
+        first_day_of_week: str = "monday",
```
MebinAbraham, line 157:
```suggestion
        first_day_of_week: str = "MONDAY",
```

app/questionnaire/placeholder_transforms.py
```diff
@@ -139,6 +149,80 @@ def calculate_date_difference(first_date: str, second_date: str) -> str:
...
+        :param days_in_range: Number of days in the range, including the start date, must be a positive integer.
+        :type days_in_range: int
+        :param first_day_of_week: Which day of the week should be considered the first (default 'monday').
+            This is the day of the week which is selected as the start of the range.
+        :type first_day_of_week: str
+        :return: The start and end datetime objects of the range.
+        :rtype: Tuple[datetime, datetime]
+        """
+        try:
+            first_day_of_week_idx = WEEKDAYS[first_day_of_week.upper()]
```
MebinAbraham, line 175: Is the upper needed given validation ensures it is uppercase already?

scripts/run_validator.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-tag=latest
+tag=generate_date_range-and-format_date_range_pair
```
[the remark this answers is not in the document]
jamesnunn: I'll sort this once the other branch is merged

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -262,6 +262,149 @@ def test_telephone_number_link(self):
...
+            ("2021-09-28", -1, 7, "MONDAY", ("2021-09-20", "2021-09-26")),
+            ("2021-09-29", -1, 7, "MONDAY", ("2021-09-20", "2021-09-26")),
+            ("2021-09-30", -1, 7, "MONDAY", ("2021-09-20", "2021-09-26")),
+            ("2021-10-01", -1, 7, "MONDAY", ("2021-09-20", "2021-09-26")),
+            ("2021-10-02", -1, 7, "MONDAY", ("2021-09-20", "2021-09-26")),
+            ("2021-10-03", -1, 7, "MONDAY", ("2021-09-20", "2021-09-26")),
+            ("2021-10-04", -1, 7, "MONDAY", ("2021-09-27", "2021-10-03")),
+        ]
+
+        self.date_range_bounds(params)
```
MebinAbraham, line 293: Thought of renaming `self.date_range_bounds(params)` to like `self.assert_date_range_bounds(params)`, just to give it some context on what it is doing.

My concern with these tests is that it seems a mix of patterns from pytest/subtest, i.e it isn't clear what the params are until you check the `date_range_bounds`, but that wouldn't be the case with pytest paramaterize. We should good to move away and just use pytest. Keep it as is for now I guess. I was going to suggest maybe a comment on what each param was but would need to duplicate for each test.

tests/functional/spec/date_ranges.spec.js
```diff
@@ -0,0 +1,42 @@
+import DateQuestionPage from "../generated_pages/placeholder_transform_date_range_bounds/date-question.page";
+import DaysQuestionBlockPage from "../generated_pages/placeholder_transform_date_range_bounds/days-question-block.page";
+import Block0Page from "../generated_pages/placeholder_transform_date_range_bounds/block0.page";
+import RangeQuestionBlockPage from "../generated_pages/placeholder_transform_date_range_bounds/range-question-block.page";
+
+describe("Date checks", () => {
```
MebinAbraham, line 6: The schema name should be like `placeholder_date_ranges.spec` for clarity.

tests/app/questionnaire/test_placeholder_transforms.py
MebinAbraham: For context for other, the new test have been moved to `pytest` for clarity. We can migrate existing one as a separate piece.

petechd (approved):

PR #672, opened 2021-09-28
Fix RoutingPath in data converter 0.0.1

pricem14pc:
The RoutingPath class implementation has changed to support a tuple of IDs, and at submission/conversion time a list of 'full' routing paths is now expected.

The original 0.0.1 converter requires updating to align with this pattern. The 0.0.3 converter had previously been updated.

Check that the existing unit tests pass given the structural changes to the RoutingPath

NOTE: It is acknowledged that gaps exist in the Runner features/datatypes this converter can support (e.g. lists are not supported) and this ticket only addresses those defined in the existing unit tests.


app/submitter/convert_payload_0_0_1.py
```diff
@@ -7,61 +7,68 @@
...
-    metadata, answer_store, list_store, schema, routing_path
+    metadata, answer_store, list_store, schema, full_routing_path
 ):
     """
     Convert answers into the data format below
+    Answers for list items are not currently supported
     'data': {
           '001': '01-01-2016',
           '002': '30-03-2016'
         }
```
LJBabbage, line 18: Not sure about this comment, or its position. I assume your are talking about list_items in connection to relationships, as list items could be deemed as a list of items like a checkbox ["Eggs", "Bacon", "Sausages"]

app/submitter/convert_payload_0_0_1.py
```diff
@@ -7,61 +7,68 @@
...
     :param metadata: questionnaire metadata
     :param answer_store: questionnaire answers
+    :param list_store: list store
     :param schema: QuestionnaireSchema class with popuated schema json
-    :param routing_path: the path followed in the questionnaire
+    :param full_routing_path: the path followed in the questionnaire
```
LJBabbage, line 23: The description might need a tweak

LJBabbage (approved):
This matches what we have for 0_0_3 and as far as I can see is just adding a for loop to cope with a list of RoutingPaths and the required tests. A couple of minor comments, which shouldn't blcok

MebinAbraham (approved):
There are no tests with multiple routing paths, there prob should be. However, won't block on it.

app/submitter/convert_payload_0_0_1.py
pricem14pc: Well, not specifically relationships (although relationship answer data isn't used in Business surveys). This was referring to any repeating section answers for a given list_item_id. I'll update this comment to `list_item_id` for clarification. The comment itself is intended to be visible at the top of the docstring.

app/submitter/convert_payload_0_0_1.py
pricem14pc: @LJBabbage How about

> _a list of section routing paths followed in the questionnaire_

PR #677, opened 2021-10-04
Update routing schema with new rules

parimooa:
Updated routing test schema with new routing rules

### How to review 
Launch survey with updated schemas and check if surveys can be completed without any issues and  ensure that correct ruling is followed during the course of surveys

**Schemas:**

- test_routing_answer_comparison.json
- test_routing_case_insensitive_text_field.json
- test_routing_group.json
- test_routing_to_questionnaire_end_multiple_sections.json
- test_routing_to_questionnaire_end_single_section.json
- test_routing_to_section_end.json

### Checklist


test_schemas/en/test_new_routing_case_insensitive_text_field.json
```diff
@@ -64,33 +64,31 @@
                             },
```
MebinAbraham, line 64: Some text in the schema is now redundant and should be changed, it references terms like `Equals Any` which is now not the case, so should be updated with the new changes.

test_schemas/en/test_new_routing_to_section_end.json
```diff
@@ -35,7 +35,9 @@
         {
```
MebinAbraham, line 35: This schema still has a routing rule based on the old structure.

test_schemas/en/test_routing_not_affected_by_answers_not_on_path.json
```diff
@@ -62,21 +62,19 @@
                             },
```
MebinAbraham, line 62: This is in the out of scope on the card?. The change is fine, skip condition can be migrated another time.

jamesnunn:
LGTM 👍

test_schemas/en/test_new_routing_case_insensitive_text_field.json
parimooa: updated description

MebinAbraham (approved):

PR #679, opened 2021-10-07
Introduce response metadata as value source

petechd:

### How to review 

### Checklist

### DIFF
--- a/app/questionnaire/path_finder.py
+++ b/app/questionnaire/path_finder.py
@@ -18,9 +18,11 @@ class PathFinder:
         list_store: ListStore,
         progress_store: ProgressStore,
         metadata: Mapping,
+        response_metadata: dict,
     ):
         self.answer_store = answer_store
         self.metadata = metadata
+        self.response_metadata = response_metadata
         self.schema = schema
         self.progress_store = progress_store
         self.list_store = list_store
@@ -136,6 +138,7 @@ class PathFinder:
             self.answer_store,
             self.list_store,
             self.metadata,
+            self.response_metadata,
             location=this_location,
             routing_path_block_ids=routing_path_block_ids,
         )
--- a/app/questionnaire/placeholder_parser.py
+++ b/app/questionnaire/placeholder_parser.py
@@ -28,6 +28,7 @@ class PlaceholderParser:
         answer_store: AnswerStore,
         list_store: ListStore,
         metadata: ImmutableDict,
+        response_metadata: dict,
         schema: QuestionnaireSchema,
         list_item_id: Optional[str] = None,
         location: Location = None,
@@ -36,6 +37,7 @@ class PlaceholderParser:
         self._answer_store = answer_store
         self._list_store = list_store
         self._metadata = metadata
+        self._response_metadata = response_metadata
         self._schema = schema
         self._list_item_id = list_item_id
         self._location = location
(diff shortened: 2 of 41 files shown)


app/questionnaire/path_finder.py
```diff
@@ -18,9 +18,11 @@ def __init__(
         list_store: ListStore,
         progress_store: ProgressStore,
         metadata: Mapping,
+        response_metadata: dict,
```
MebinAbraham, line 21: We have some inconsistencies in where we used dict vs mapping, for these I think we should change to `Mapping`. You will to change the typing in value source from `dict` to `Mapping` as well.

app/questionnaire/placeholder_parser.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         answer_store: AnswerStore,
         list_store: ListStore,
         metadata: ImmutableDict,
+        response_metadata: dict,
```
MebinAbraham, line 31: Same with placeholders renderer/parser. We should change to `Mapping`.

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -49,6 +49,7 @@ def get_when_rule_evaluator(
     return WhenRuleEvaluator(
```
MebinAbraham, line 49: This file has tests for each of the value source, so a new one should be added for `response_metadata`.

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -35,6 +35,7 @@ def get_value_source_resolver(
     answer_store: AnswerStore = AnswerStore(),
```
MebinAbraham, line 35: This file has tests for each of the value source, so a new one should be added for `response_metadata`.

app/questionnaire/placeholder_parser.py
petechd: Changed to `Mapping`, thanks!

tests/app/questionnaire/routing/test_when_rule_evaluator.py
```diff
@@ -247,6 +249,31 @@ def test_metadata_source(metadata_value, expected_result):
...
+    )
+
+    assert (
+        when_rule_evaluator.evaluate(
+            rule={
+                Operator.EQUAL: [
+                    {"source": "response_metadata", "identifier": "started_at"},
+                    "2021-10-11T09:40:11.220038+00:00",
+                ]
+            },
```
MebinAbraham, line 271: Normally wouldn't use an operator like this for response metadata, but given we don't support ISO format dates in operators or placeholder yet, I think it is fine for now.

tests/app/questionnaire/test_placeholder_parser.py
```diff
@@ -95,6 +97,39 @@ def test_metadata_transform_placeholder():
...
         metadata={"ref_p_start_date": "2019-02-11"},
+        response_metadata={},
+        schema=QuestionnaireSchema({}),
+    )
+    placeholders = parser(placeholder_list)
+
+    assert placeholders["start_date"] == "Monday 11 February 2019"
+
+
+def test_response_metadata_transform_placeholder():
```
MebinAbraham, line 108: Caveat that this is a point in time test as we don't support ISO format dates yet.

MebinAbraham (approved):

tests/app/questionnaire/test_placeholder_parser.py
petechd: Added a comment in that test for future reference.

PR #680, opened 2021-10-08
Fix python lint all files

parimooa:

- Updated old string formatting to new style using `f-string`
- Removed unused imports

### How to review 
Run all the tests and pick some surveys to check manually 

### Checklist

### DIFF
--- a/app/data_models/list_store.py
+++ b/app/data_models/list_store.py
@@ -67,10 +67,10 @@ class ListModel:
     def first(self):
         try:
             return self.items[0]
-        except IndexError:
+        except IndexError as first_item_missing:
             raise IndexError(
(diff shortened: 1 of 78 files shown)


app/data_models/list_store.py
```diff
@@ -67,10 +67,10 @@ def __repr__(self):
     def first(self):
         try:
             return self.items[0]
-        except IndexError:
+        except IndexError as first_item_missing:
             raise IndexError(
                 f"unable to access first item in list, list '{self.name}' is empty"
-            )
+            ) from first_item_missing
```
MebinAbraham, line 73: I prefer to keep the reference as `as e:`, `as ex:`, `as exc:` as that is common pattern in Python. I guess developer preference.

PR #682, opened 2021-10-11
Upgrade to DS v42

rmccar:
Upgrade to Design System to version 42.0.0

Fixes breaking changes in buttons and panels, removes 3 panel imports that are not being used and removes js-btn-save which isn't actually being set because its getting overwritten in header.

### How to review 
- Tests pass

### Checklist


rmccar:
[Update EQ Runner design-system version to 42.0.0 (3 SP)](https://trello.com/c/Vf2vw2d2/4910-update-eq-runner-design-system-version-to-4200-3-sp)

templates/census-thank-you.html
```diff
@@ -13,7 +13,7 @@
   {% call onsPanel({
     "type": "success",
     "classes": "ons-u-mb-s",
-    "icon": "check",
+    "iconType": "check",
```
MebinAbraham, line 16: There are other uses of `icon` such as on the hub that need updating. Currently, the icons on the hub is broken. 

Jinja filters also have references to the old way which needs updating.

templates/layouts/configs/_save-sign-out-button.html
```diff
@@ -17,7 +17,8 @@
                 "data-ga-category": "Navigation",
                 "data-ga-action": "Save and sign out click"
             },
-            "buttonStyle": "exit"
```
MebinAbraham, line 20: Is `buttomStyle` still a valid use in the DS?

rmccar: It is for some button types but not for exit anymore. For exit we just add the icon, icon position and variant now

PR #684, opened 2021-10-12
Browser back button from thank you page

LJBabbage:

### How to review 

### Checklist

### DIFF
--- a/templates/errors/previously-submitted.html
+++ b/templates/errors/previously-submitted.html
@@ -1,0 +1,9 @@
+{% extends 'layouts/_base.html' %}
+
+{% set page_title = _("Submission Complete") %}
+
+{% block main %}
+  <h1>{{ _("This page is no longer available") }}</h1>
+  <p>{{ _("Your survey has been submitted") }}</p>
+  <p>{{ _("<a href='{url}'>Return to previous page</a>").format(url = thank_you_url) }}</p>
+{% endblock %}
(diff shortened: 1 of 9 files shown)


templates/errors/previously-submitted.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'layouts/_base.html' %}
+
+{% set page_title = _("Submission Complete") %}
```
LJBabbage, line 3: This wasn't on the card, but I check with Donna that this is correct

tests/integration/test_no_questionnaire_state.py
```diff
@@ -15,21 +15,6 @@ def test_questionnaire_route_before_questionnaire_submitted(self):
             # Then
             self.assertStatusUnauthorised()
 
-    def test_questionnaire_route_after_questionnaire_submitted(self):
```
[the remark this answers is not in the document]
MebinAbraham: It is possible to have a cookie but not have submitted in the cookie if another user submits the questionnaire first, or the user relaunches in a new browser or after closing the browser. But the test we have for `resuming` should cover that, so seem okay to remove.

tests/app/submitter/test_previously_submitted_exception.py
```diff
@@ -0,0 +1,13 @@
+import unittest
```
parimooa, line 1: Shouldn't the new unit tests be written in `pytest` ?

PR #685, opened 2021-10-13
Ability to save answers as a PDF

MebinAbraham:

- Adds the ability to save the answers as a PDF on the view response page.
- Since the print style sheet need to live in runner, I have added a script to pull the latest style sheet from the CDN when loading the design system, otherwise, the styles may become out of date and the rendered PDF may not look as expected.

### How to review
Use `test_view_submitted_response.json` schema to ensure you are able to download the answers as a PDF. 

- Deploy this to GCP or run it via Docker to ensure it works as expected under Debian.

### Checklist


tests/app/views/handlers/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,48 @@
...
+
+        self.schema = QuestionnaireSchema({"post_submission": {"view_response": True}})
+        self.language = "en"
+
+    def set_submitted_at(self, submitted_at):
+        self.storage = Mock()
+        self.storage.get_user_data = Mock(return_value=("{}", 1, submitted_at))
+
+    def test_pdf_not_downloadable(self):
+        submitted_at = datetime.now(timezone.utc) - timedelta(minutes=60)
```
jamesnunn, line 25: Only splitting hairs but would you not test a boundary condition like -46 mins?

tests/app/views/handlers/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,48 @@
...
+        self.questionnaire_store = QuestionnaireStore(self.storage)
+
+        self.questionnaire_store.set_metadata(
+            {"schema_name": "test_view_submitted_response"}
+        )
+        pdf = ViewSubmittedResponsePDF(
+            self.schema, self.questionnaire_store, self.language
+        )
+
+        self.assertEqual(pdf.filename, f"test_view_submitted_response.pdf")
```
jamesnunn, line 48: f string has no replacement targets here so I guess it's not needed?

tests/app/views/handlers/test_view_submitted_response_pdf.py
MebinAbraham: Fair enough, updated.

app/questionnaire/questionnaire_schema.py
```diff
@@ -209,6 +209,12 @@ def get_post_submission(self) -> ImmutableDict:
         schema: ImmutableDict = self.json.get("post_submission", ImmutableDict({}))
         return schema
 
+    @property
+    def is_view_submitted_response_enabled(self) -> bool:
```
parimooa, line 213: why is this  `property` instead of method?

app/views/handlers/view_submitted_response_pdf.py
```diff
@@ -0,0 +1,44 @@
...
+        return f"{self._metadata['schema_name']}.pdf"
+
+    def get_pdf(self) -> io.BytesIO:
+        content_as_bytes = pdfkit.from_string(
+            input=self.get_rendered_html(),
+            output_path=None,
+            css=f'{current_app.config["PRINT_STYLE_SHEET_FILE_PATH"]}/print.css',
+            options={
+                "quiet": "",
+                "dpi": 365,
```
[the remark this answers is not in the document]
MebinAbraham: Do you mean the `dpi` or the literal options?.

tests/app/views/handlers/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,48 @@
+from datetime import datetime, timedelta, timezone
```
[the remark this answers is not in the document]
MebinAbraham: To keep it consistent with the existing `test_view_submitted_response.py`, which, this has overlap with and was written using unit tests. Would be happy to the new one and existing to pytest.

app/questionnaire/questionnaire_schema.py
MebinAbraham: Why does it need to be a method?. I see it as a property since we do not pass any arguments, which is consistent with our other `is_` functions on questionnaire schema class. It uses method when you have to pass args, and property when you don't.

tests/app/views/handlers/test_view_submitted_response_pdf.py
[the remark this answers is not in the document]
MebinAbraham: Actually my bad, I was looking at the wrong file, I was looking at the integration test not this, however even for this the existing one is unittest but given it is only 1 existing test, ill migrate them to pytest so it leaves us in a better shape.

app/questionnaire/questionnaire_schema.py
parimooa: Fair enough ? could it be a `@cached_property` like other properties to be consistent ? i am not sure.

MebinAbraham: Good point, changed to `@cached_property`.

app/views/handlers/view_submitted_response_pdf.py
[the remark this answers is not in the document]
MebinAbraham: Made options a class attribute.

tests/app/views/handlers/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,35 @@
...
+from unittest.mock import Mock
```
parimooa, line 2: Dont see `Mock` getting used in here ?

tests/app/views/handlers/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,35 @@
...
+
+import pytest
+
+from app.data_models import QuestionnaireStore
+from app.questionnaire.questionnaire_schema import QuestionnaireSchema
+from app.views.handlers.view_submitted_response import ViewSubmittedResponseExpired
+from app.views.handlers.view_submitted_response_pdf import ViewSubmittedResponsePDF
+from tests.app.views.handlers.conftest import set_storage_data
```
parimooa, line 10: this could be `from .conftest import set_storage_data` ?

tests/app/views/handlers/conftest.py
```diff
@@ -0,0 +1,33 @@
+from unittest.mock import Mock
```
parimooa, line 1: Consider using [monkeypatch](https://docs.pytest.org/en/6.2.x/monkeypatch.html), i will leave it up to you.

tests/app/views/handlers/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,35 @@
+from datetime import datetime, timedelta, timezone
+from unittest.mock import Mock
+
+import pytest
+
+from app.data_models import QuestionnaireStore
+from app.questionnaire.questionnaire_schema import QuestionnaireSchema
```
parimooa, line 7: Also `QuestionnaireSchema` is not being used anywhere

tests/app/views/handlers/test_view_submitted_response_pdf.py
MebinAbraham: Removed, we really need to get those pylint cards in, as lint should fail for these but don't atm 😞 .

tests/app/views/handlers/conftest.py
MebinAbraham: Haven't actually used `monkeypatch` for test yet, but we could consider that, atm I used `Mock` as that is what we do across runner.

tests/app/views/handlers/test_view_submitted_response_pdf.py
parimooa: coming from AWS lambda , it was a pain to have relative imports. very weird. If we use relative across let it be like that. TBH iam not fussed about it.

tests/integration/routes/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,66 @@
...
+        self.post()
+
+        # When I try to download my answer
+        self.get("/submitted/download-pdf")
+
+        # Then a 404 page is returned
+        self.assertStatusNotFound()
+
+    def test_download_when_submitted_response_enabled_but_not_expired(self):
+        # Given I launch and complete a questionnaire that does not have view-submitted-response enabled
```
parimooa, line 23: this comment contradicts the test , is `view-submitted-response` enabled or disabled ?

tests/integration/routes/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,66 @@
...
+        self._launch_and_complete_questionnaire()
+
+        # When I try to download my answer
+        self.get(self.VIEW_RESPONSE_PAGE_URL)
+        self.get(self.get_download_button().attrs["href"])
+
+        # Then a the download is successful
```
parimooa, line 30: typo ?

app/routes/questionnaire.py
```diff
@@ -393,14 +395,34 @@ def get_view_submitted_response(schema, questionnaire_store):
...
-        template="view-submitted-response",
-        content=view_submitted_response.get_context(),
-        page_title=view_submitted_response.get_page_title(),
+    return view_submitted_response.get_rendered_html()
+
+
+@post_submission_blueprint.route("download-pdf", methods=["GET"])
+@with_questionnaire_store
+@with_schema
+def get_view_submitted_response_pdf(
```
parimooa, line 407: consider doc strings

app/routes/questionnaire.py
```diff
@@ -393,14 +395,34 @@ def get_view_submitted_response(schema, questionnaire_store):
...
+    schema: QuestionnaireSchema, questionnaire_store: QuestionnaireStore
+):
```
parimooa, line 409: return type missing

app/views/handlers/view_submitted_response_pdf.py
```diff
@@ -0,0 +1,49 @@
...
+from app.data_models import QuestionnaireStore
+from app.questionnaire import QuestionnaireSchema
+from app.views.handlers.view_submitted_response import (
+    ViewSubmittedResponse,
+    ViewSubmittedResponseExpired,
+)
+
+
+class ViewSubmittedResponsePDF(ViewSubmittedResponse):
+    mimetype = "application/pdf"
```
parimooa, line 15: consider doc strings

tests/integration/routes/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,66 @@
...
+
+        # Check filename is set as expected
+        assert (
+            "filename=test_view_submitted_response.pdf"
+            in self.last_response_headers["Content-Disposition"]
+        )
+
+        # Check content length is reasonable.
+        # This is given some leeway as it can change with DS changes.
+        self.assertGreater(self.last_response.content_length, 16500)
```
parimooa, line 47: why not `assert self.last_response.content_length > 16500 ,"PDF content is less than 16500"`

tests/integration/routes/test_view_submitted_response_pdf.py
```diff
@@ -0,0 +1,66 @@
...
+
+        # When I try to download my answer
+        self.get(self.VIEW_RESPONSE_PAGE_URL)
+        self.get(self.get_download_button().attrs["href"])
+
+        # Then a the download is successful
+        self.assertStatusOK()
+
+        # Check response content type is PDF
+        self.assertEqual(self.last_response.content_type, "application/pdf")
```
parimooa, line 34:
```suggestion
       assert self.last_response.content_type == "application/pdf" ," Response type is not of 'PDF' "
```

app/routes/questionnaire.py
MebinAbraham: Have added doc string. We should have a wider convo with the type on the style guide for doc strings and when we should have them etc. The doc strings we have mostly conform to `Sphinx` format, but I personally prefer the Google's style guide and it not as visually dense. We should discuss the options in a tech session and put together or follow a style guide.

tests/integration/routes/test_view_submitted_response_pdf.py
MebinAbraham: https://github.com/ONSdigital/eq-questionnaire-runner/pull/685#discussion_r731835329

PR #686, opened 2021-10-13
Add new rules engine for skip conditions

petechd:
Adds new rules engine for skip conditions, for the first pass only `test_skip_condition_group.json` and `test_skip_condition_group.json` schemas were migrated and renamed with the prefix `test_new`, as the `test_skip_condition.json` is showcasing skip conditions functionality in question, which is no longer supported. Skip conditions are still backward compatible as required on [the card](https://trello.com/c/nkAW6L7j).

### How to review 
Two new tests added fro group and block skip condition testing.

### Checklist


petechd:
[Runner use new rules engine for skip conditions (2 SP)](https://trello.com/c/nkAW6L7j/4903-runner-use-new-rules-engine-for-skip-conditions-2-sp)

app/questionnaire/path_finder.py
```diff
@@ -34,7 +34,7 @@ def routing_path(
         Visits all the blocks in a section and returns a path given a list of answers.
         """
         blocks: List[Mapping] = []
-        routing_path_block_ids = []
+        routing_path_block_ids: List = []
```
MebinAbraham, line 37: Collections should define what type they hold.

app/questionnaire/path_finder.py
```diff
@@ -44,7 +44,21 @@ def routing_path(
...
+                    if isinstance(group.get("skip_conditions"), dict):
+                        when_rule_evaluator = WhenRuleEvaluator(
+                            self.schema,
+                            self.answer_store,
+                            self.list_store,
+                            self.metadata,
+                            self.response_metadata,
+                            location=current_location,
+                            routing_path_block_ids=routing_path_block_ids,
+                        )
```
MebinAbraham, line 56: We should create this outside the loop so that it isn't initialised multiple times. It would be temporarily we would create with regardless of if it is new/old, but I think that is fine as the overhead is negligible and makes it consistent with routing rules.

app/questionnaire/path_finder.py
```diff
@@ -80,16 +94,30 @@ def _build_routing_path_block_ids(self, blocks, current_location):
...
+            else:
+                is_skipping = block.get("skip_conditions") and evaluate_skip_conditions(
+                    block["skip_conditions"],
+                    self.schema,
+                    self.metadata,
+                    self.answer_store,
+                    self.list_store,
+                    current_location=current_location,
+                    routing_path_block_ids=routing_path_block_ids,
+                )
```
MebinAbraham, line 120: We should assign `block.get("skip_conditions")` to a var since it is referenced quite a few times.

app/questionnaire/path_finder.py
```diff
@@ -80,16 +94,30 @@ def _build_routing_path_block_ids(self, blocks, current_location):
...
+                    self.list_store,
+                    self.metadata,
+                    self.response_metadata,
+                    location=current_location,
+                    routing_path_block_ids=routing_path_block_ids,
+                )
+                when_rule = block["skip_conditions"].get("when")
+                is_skipping = block["skip_conditions"] and when_rule_evaluator.evaluate(
+                    when_rule
+                )
```
MebinAbraham, line 110: Can be inline, since it has to existing at this point so `.get()` isn't needed. You also don't need the first check to see if you have skip conditions as you should at this point. Empty dict would be an invalid schema.

```suggestion
                is_skipping = when_rule_evaluator.evaluate(
                    skip_conditions["when"]
                )
```

app/questionnaire/path_finder.py
```diff
@@ -44,7 +44,21 @@ def routing_path(
...
+                        when_rule_evaluator = WhenRuleEvaluator(
+                            self.schema,
+                            self.answer_store,
+                            self.list_store,
+                            self.metadata,
+                            self.response_metadata,
+                            location=current_location,
+                            routing_path_block_ids=routing_path_block_ids,
+                        )
+                        when_rule = group.get("skip_conditions").get("when")
```
MebinAbraham, line 57: Same as the comment for block, worth assigning skip conditions to a var.

app/questionnaire/path_finder.py
petechd: Added type to Collection, thanks!

app/questionnaire/path_finder.py
petechd: Moved it above the `for` loop, I hope that is what you meant, thanks!

app/questionnaire/path_finder.py
petechd: Assigned, thanks!

petechd:
> Looks good to me and works as expected, The schema could do with a tweak to make it look at bit better (i.e Do you want to skip? being both question title and answer title) and changing the functional tests to Given, When, Then

I fixed the functional tests so they follow "Given->When->Then" pattern and cleaned up test schemas so no title-label repetition and changed some titles to make more sense.

app/questionnaire/path_finder.py
```diff
@@ -80,16 +96,28 @@ def _build_routing_path_block_ids(self, blocks, current_location):
...
+            else:
+                is_skipping = skip_conditions and evaluate_skip_conditions(
+                    skip_conditions,
+                    self.schema,
+                    self.metadata,
+                    self.answer_store,
+                    self.list_store,
+                    current_location=current_location,
+                    routing_path_block_ids=routing_path_block_ids,
+                )
```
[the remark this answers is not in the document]
petechd: Repetition removed, I need you to check it first (if it's ok to be in path finder) before I rewrite the tests though..

app/questionnaire/path_finder.py
```diff
@@ -169,6 +169,11 @@ def _evaluate_routing_rules(
     def _evaluate_skip_conditions(
         self, this_location, routing_path_block_ids, skip_conditions
     ):
+        no_skip_condition = skip_conditions is None or len(skip_conditions) == 0
+
+        if no_skip_condition:
```
MebinAbraham, line 174: Can this not be?
```suggestion
        if not skip_conditions:
```

test_schemas/en/test_new_skip_condition_block.json
```diff
@@ -0,0 +1,99 @@
...
+                {
+                    "id": "do-you-want-to-skip-group",
+                    "title": "Do you want to skip the next block?",
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "do-you-want-to-skip",
+                            "question": {
+                                "id": "do-you-want-to-skip-question",
+                                "title": "Do you want to skip?",
```
MebinAbraham, line 44: I think we need to make the schema a bit clearer.

The question title should be like : `Do you want to skip the next question?`.

Then some guidance like `Select "Yes" to skip the next question and go straight to the summary`.

test_schemas/en/test_new_skip_condition_block.json
```diff
@@ -0,0 +1,99 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "0",
+    "title": "Skip group",
```
MebinAbraham, line 7: This isn't the group schema.

test_schemas/en/test_new_skip_condition_group.json
```diff
@@ -0,0 +1,105 @@
...
+                    "id": "do-you-want-to-skip-group",
+                    "title": "Do you want to skip the next group?",
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "do-you-want-to-skip",
+                            "question": {
+                                "id": "do-you-want-to-skip-question",
+                                "title": "Do you want to skip?",
+                                "type": "General",
```
MebinAbraham, line 45: Same here, we need to make it clearer. The title of the group is a question, which is confusing when you get to the summary. We should set the title of the group to just like `Group 1`, `Group 2 (Skippable)` etc so on the summary it is clear you skipped/not skipped a group.

MebinAbraham (commented):
Looks good, just some clarity in schema is needed I think.

app/questionnaire/path_finder.py
```diff
@@ -176,6 +166,39 @@ def _evaluate_routing_rules(
...
+            self.answer_store,
+            self.list_store,
+            self.metadata,
+            self.response_metadata,
+            location=this_location,
+            routing_path_block_ids=routing_path_block_ids,
+        )
+
+        if isinstance(skip_conditions, dict):
+            return when_rule_evaluator.evaluate(skip_conditions["when"])
```
MebinAbraham, line 186: Can be moved down to only initialise when needed.
```suggestion
        if isinstance(skip_conditions, dict):
            when_rule_evaluator = WhenRuleEvaluator(
                self.schema,
                self.answer_store,
                self.list_store,
                self.metadata,
                self.response_metadata,
                location=this_location,
                routing_path_block_ids=routing_path_block_ids,
            )

            return when_rule_evaluator.evaluate(skip_conditions["when"])
```

test_schemas/en/test_new_skip_condition_group.json
petechd: Fixed, thanks!

app/questionnaire/path_finder.py
```diff
@@ -44,13 +44,9 @@ def routing_path(
...
                 if "skip_conditions" in group:
-                    if evaluate_skip_conditions(
-                        group["skip_conditions"],
-                        self.schema,
-                        self.metadata,
-                        self.answer_store,
-                        self.list_store,
-                        current_location=current_location,
+                    skip_conditions = group.get("skip_conditions")
+                    if self._evaluate_skip_conditions(
```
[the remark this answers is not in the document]
petechd: Ok, done, thanks!

LJBabbage (approved):

PR #689, opened 2021-10-14
Update the feedback payload object structure

berroar:

Updates the payload object structure for feedback to match the desired structure by SDC.

### How to review 

Check that the updated structure matches the described structure on the trello card.

### Checklist

### DIFF
--- a/app/helpers/template_helpers.py
+++ b/app/helpers/template_helpers.py
@@ -38,6 +38,7 @@ class ContextHelper:
         self._account_service_url = cookie_session.get(
             "account_service_url", self._survey_config.account_service_url
         )
+        self._account_service_surveys_url = f"{self._account_service_url}{self._survey_config.account_service_surveys_path}"
         self._account_service_log_out_url = cookie_session.get(
             "account_service_log_out_url"
         )
@@ -56,6 +57,7 @@ class ContextHelper:
         return {
             "account_service_url": self._account_service_url,
             "account_service_log_out_url": self._account_service_log_out_url,
+            "account_service_surveys_url": self._account_service_surveys_url,
             "contact_us_url": Link(
                 lazy_gettext("Contact us"),
                 f"{self._survey_config.base_url}/contact-us/",
--- a/app/storage/metadata_parser.py
+++ b/app/storage/metadata_parser.py
@@ -78,7 +78,7 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
     tx_id = VALIDATORS["uuid"]()  # type:ignore
     response_id = VALIDATORS["string"](validate=validate.Length(min=1))  # type:ignore
 
-    account_service_url = VALIDATORS["url"](required=False)  # type:ignore
+    account_service_url = VALIDATORS["url"](required=True)  # type:ignore
     case_id = VALIDATORS["uuid"]()  # type:ignore
     account_service_log_out_url = VALIDATORS["url"](required=False)  # type:ignore
     roles = fields.List(fields.String(), required=False)
--- a/app/survey_config/business_config.py
+++ b/app/survey_config/business_config.py
@@ -11,6 +11,8 @@ from app.survey_config.survey_config import SurveyConfig
 class BusinessSurveyConfig(
     SurveyConfig,
 ):
+
+    account_service_surveys_path: str = "/surveys/todo"
     survey_title: str = "ONS Business Surveys"
 
     footer_links: Iterable[MutableMapping] = field(
--- a/app/survey_config/survey_config.py
+++ b/app/survey_config/survey_config.py
@@ -23,6 +23,7 @@ class SurveyConfig:
     )
     base_url: str = ACCOUNT_SERVICE_BASE_URL
     account_service_url: Optional[str] = None
+    account_service_surveys_path: Optional[str] = None
     title_logo: Optional[str] = None
     title_logo_alt: Optional[str] = None
     header_logo: Optional[str] = None
--- a/app/translations/messages.pot
+++ b/app/translations/messages.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2021-10-18 09:39+0100\n"
+"POT-Creation-Date: 2021-10-20 15:17+0100\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -203,17 +203,17 @@ msgstr[1] ""
 msgid "Not a valid choice"
 msgstr ""
 
-#: app/helpers/template_helpers.py:60 app/survey_config/business_config.py:23
+#: app/helpers/template_helpers.py:62 app/survey_config/business_config.py:24
 #: app/survey_config/census_config.py:30 app/survey_config/census_config.py:79
 #: app/survey_config/census_config.py:139
 msgid "Contact us"
 msgstr ""
 
-#: app/helpers/template_helpers.py:104
+#: app/helpers/template_helpers.py:106
 msgid "The following links open in a new tab"
 msgstr ""
 
-#: app/helpers/template_helpers.py:134
+#: app/helpers/template_helpers.py:136
 msgid ""
 "Make sure you <a href='{sign_out_url}'>leave this page</a> or close your "
 "browser if using a shared device"
@@ -297,21 +297,21 @@ msgstr ""
 msgid "An individual access code has been sent by text"
 msgstr ""
 
-#: app/survey_config/business_config.py:22
+#: app/survey_config/business_config.py:23
 msgid "What we do"
 msgstr ""
 
-#: app/survey_config/business_config.py:25
+#: app/survey_config/business_config.py:26
 msgid "Accessibility"
 msgstr ""
 
-#: app/survey_config/business_config.py:35
+#: app/survey_config/business_config.py:36
 #: app/survey_config/census_config.py:44 app/survey_config/census_config.py:94
 #: app/survey_config/census_config.py:146
 msgid "Cookies"
 msgstr ""
 
-#: app/survey_config/business_config.py:37
+#: app/survey_config/business_config.py:38
 #: app/survey_config/census_config.py:50 app/survey_config/census_config.py:100
 #: app/survey_config/census_config.py:152
 msgid "Privacy and data protection"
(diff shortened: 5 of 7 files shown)


berroar:
[Update feedback object structure (2 SP)](https://trello.com/c/YOVTInBi/4899-update-feedback-object-structure-2-sp)

app/views/handlers/feedback.py
```diff
@@ -226,17 +234,48 @@ def __call__(self) -> Mapping:
...
+            "launch_language_code": self.metadata.get(
+                "language_code", DEFAULT_LANGUAGE_CODE
+            ),
+            "version": "0.0.3",
+        }
+
+        if self.metadata.get("form_type"):
+            payload["form_type"] = self.metadata["form_type"]
+
+        payload["data"] = {
```
parimooa, line 271: Rather than adding key and then deleting it later , why not conditionally update `payload` ?
```suggestion
        payload["data"] = {
            "feedback_text": self.feedback_text,
            "feedback_type": self.feedback_type,
            "feedback_count": self.feedback_count,
            }
          if self.feedback_type_question_category:
                    payload["data"].update({"feedback_type_question_category": self.feedback_type_question_category})
          return payload
```

app/views/handlers/feedback.py
```diff
@@ -226,17 +234,48 @@ def __call__(self) -> Mapping:
...
+        metadata,
+        session_store,
+        schema,
+        feedback_count,
         feedback_text,
         feedback_type,
         feedback_type_question_category=None,
     ):
+        self.metadata = metadata
+        self.session_store = session_store
```
parimooa, line 246: how about adding type hints ?

app/submitter/converter.py
```diff
@@ -73,8 +73,8 @@ def convert_answers(
         "survey_id": survey_id,
         "flushed": flushed,
         "submitted_at": submitted_at.isoformat(),
-        "collection": _build_collection(metadata),
-        "metadata": _build_metadata(metadata),
+        "collection": build_collection(metadata),
```
parimooa, line 76: you need to update doc strings for `convert_answers` method  since the payload structure has changed?

berroar: Added now

app/views/handlers/feedback.py
```diff
@@ -224,19 +232,94 @@ def __call__(self) -> Mapping:
...
+            "submitted_at": SubmissionHandler.submitted_at,
+            "collection": build_collection(self.metadata),
+            "metadata": build_metadata(self.metadata),
+            "survey_id": self.schema.json["survey_id"],
+            "tx_id": self.metadata["tx_id"],
+            "type": "uk.gov.ons.edc.eq:feedback",
+            "launch_language_code": self.metadata.get(
+                "language_code", DEFAULT_LANGUAGE_CODE
+            ),
+            "version": "0.0.3",
```
pricem14pc, line 308: The version for payloads is intending to convey the data structure. Feedback conforms to the `0.0.1` data structure (essentially key:value strings). 0.0.3 is used for representing survey answer objects.

parimooa (approved):
LGTM

tests/app/views/handlers/test_feedback_upload.py
```diff
@@ -1,41 +1,172 @@
...
-def test_feedback_payload_with_feedback_type_question_category():
-    feedback_payload = FeedbackPayload(
-        "Feedback text", "Feedback type", "Feedback type question category"
-    )
-    expected_payload = {
-        "feedback_text": "Feedback text",
-        "feedback_type": "Feedback type",
-        "feedback_type_question_category": "Feedback type question category",
-    }
+class TestFeedbackUpload(AppContextTestCase):  # pylint: disable=too-many-public-methods
```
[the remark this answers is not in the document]
berroar: Updated file to use pytest

app/views/handlers/feedback.py
```diff
@@ -224,19 +233,106 @@ def __call__(self) -> Mapping:
...
+    """
+
     def __init__(
         self,
-        feedback_text,
-        feedback_type,
-        feedback_type_question_category=None,
+        metadata: Mapping[str, Union[str, int, list]],
+        started_at: str,
+        submitted_at: datetime,
```
MebinAbraham, line 288: This is currently being passed in from `self._questionnaire_store.submitted_at` but it should be the `submitted_at` time for the feedback, not the questionnaire?.

app/views/handlers/feedback.py
```diff
@@ -224,19 +233,106 @@ def __call__(self) -> Mapping:
...
+    :return payload: Feedback payload object
+    """
+
     def __init__(
         self,
-        feedback_text,
-        feedback_type,
-        feedback_type_question_category=None,
+        metadata: Mapping[str, Union[str, int, list]],
+        started_at: str,
```
MebinAbraham, line 287: I know we are conforming to 0.0.1 but seems odd to include the questionnaire started_at in the feedback payload. Even some other keys just seems odd as part of feedback but guess it is what it is.

tests/app/views/handlers/test_feedback_upload.py
```diff
@@ -1,41 +1,196 @@
...
+    }
+
+
+def test_feedback_payload_with_feedback_type_question_category(
+    session_data, schema, metadata
+):
+
+    session_data = session_data
+    schema = schema
+    metadata = metadata
```
MebinAbraham, line 80: These are redundant and some be removed.

tests/app/views/handlers/test_feedback_upload.py
```diff
@@ -1,41 +1,196 @@
...
+    assert expected_payload == feedback_payload()
+
+
+def test_feedback_payload_without_feedback_type_question_category(
+    session_data, schema, metadata
+):
 
+    session_data = session_data
+    schema = schema
+    metadata = metadata
```
MebinAbraham, line 135: Same here, don't need to reassign.

app/views/handlers/feedback.py
berroar: Updated this now

app/views/handlers/feedback.py
```diff
@@ -1,6 +1,6 @@
 from datetime import datetime, timezone
```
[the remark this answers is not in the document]
berroar: Have updated this now 👍

berroar:
> The `form_type` and `submission_language_code` seems to be missing from the payload. Other than that, it matches the schema definitions doc.

Added `submission_language_code` , `form_type` is there but is only set if it's available like the answer payload.

pricem14pc:
> > The `form_type` and `submission_language_code` seems to be missing from the payload. Other than that, it matches the schema definitions doc.
> 
> Added `submission_language_code` , `form_type` is there but is only set if it's available like the answer payload.

Yes, some of these are optional. It's not particularly well understood as to when things like `form_type` or `case_ref` will be included. A common payload generator for `0.0.1` attributes to be used for both feedback and surveyresposne would be preferable but we can revisit that if we ever need to. I don't think there will be a great deal of iteration on 0.0.1. If feedback for social is progressed then we'll look to align that with 0.0.3 and there will be new requirements around all the other properties we will need to meet.

app/views/handlers/feedback.py
```diff
@@ -211,32 +217,117 @@ def is_enabled(schema: QuestionnaireSchema) -> bool:
...
+            "tx_id": self.metadata["tx_id"],
+            "type": "uk.gov.ons.edc.eq:feedback",
+            "launch_language_code": self.metadata.get(
+                "language_code", DEFAULT_LANGUAGE_CODE
+            ),
+            "version": "0.0.1",
+        }
+
+        if form_type := self.metadata.get("form_type"):
+            payload["form_type"] = form_type
```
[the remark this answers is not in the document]
MebinAbraham: I think this is the issue with not having a common converter, for answers we have all these:
```py
    if metadata.get("channel"):
        payload["channel"] = metadata["channel"]
    if metadata.get("case_type"):
        payload["case_type"] = metadata["case_type"]
    if metadata.get("form_type"):
        payload["form_type"] = metadata["form_type"]
    if metadata.get("region_code"):
        payload["region_code"] = metadata["region_code"]
    if response_metadata.get("started_at"):
        payload["started_at"] = response_metadata["started_at"]
    if metadata.get("case_ref"):
        payload["case_ref"] = metadata["case_ref"]
```

But for feedback, not all these have been added but the document Mark has updated doesn't state that none of these won't exist for feedback, even though the example don't have all these. So it is worth confirming if all these should be in feedback also, if not, we need to update the doc to highlight which keys are not present for feedback.

app/views/handlers/feedback.py
```diff
@@ -72,14 +79,17 @@ def handle_post(self):
...
             session_data.tx_id,
         )
 
         feedback_message = FeedbackPayload(
+            self._questionnaire_store.metadata,
+            self._questionnaire_store.response_metadata,
+            session_data.case_id,
+            self._schema,
+            session_data.feedback_count,
+            session_data.language_code,
```
pricem14pc, line 92: If python does this positionally I think the arguments for feedback_count and language_code need to be swapped.

app/submitter/converter.py
```diff
@@ -128,3 +118,22 @@ def _build_metadata(metadata):
...
+    if case_type := metadata.get("case_type"):
+        payload["case_type"] = case_type
+    if form_type := metadata.get("form_type"):
+        payload["form_type"] = form_type
+    if region_code := metadata.get("region_code"):
+        payload["region_code"] = region_code
+    if started_at := response_metadata.get("started_at"):
+        payload["started_at"] = started_at
+    if case_ref := metadata.get("case_ref"):
+        payload["case_ref"] = case_ref
```
jamesnunn, line 137: Maybe personal preference for DRY but this reduces the blockiness of the code for me
```suggestion
    for key in ["channel", "case_type", "form_type", "region_code", "started_at", "case_ref"]:
        if value := metadata.get(key):
            payload[key] = value
```

app/views/handlers/feedback.py
```diff
@@ -10,8 +10,15 @@
...
 from app.forms.questionnaire_form import generate_form
 from app.questionnaire.questionnaire_schema import QuestionnaireSchema
+from app.submitter.converter import (
+    build_collection,
+    build_metadata,
+    get_optional_payload_properties,
+)
 from app.views.contexts.feedback_form_context import build_feedback_context
 
+DEFAULT_LANGUAGE_CODE = "en"
```
jamesnunn, line 20: Is there not a default language code for the entire session?

app/views/handlers/feedback.py
```diff
@@ -59,42 +68,47 @@ def form(self):
...
 
-    def get_context(self):
+    def get_context(self) -> Mapping[str, Union[str, bool, dict]]:
         return build_feedback_context(self.question_schema, self.form)
 
-    def get_page_title(self):
+    def get_page_title(self) -> Union[str, Any]:
         if self.form.errors:
             return gettext("Error: {page_title}").format(page_title=self.PAGE_TITLE)
         return self.PAGE_TITLE
```
MebinAbraham, line 77: The `Any` makes the `str` redundant, the `Any` is just cause of `gettext`, we can make it explicit.
```suggestion
    def get_page_title(self) -> str:
        if self.form.errors:
            title: str = gettext("Error: {page_title}").format(
                page_title=self.PAGE_TITLE
            )
            return title
        return self.PAGE_TITLE
```

app/views/handlers/feedback.py
```diff
@@ -123,7 +137,7 @@ def question_schema(self):
 
         options = (
             {"label": value, "value": value}
-            for value in detail_answers_option_map.get(self._schema.form_type or "H")
+            for value in detail_answers_option_map.get(self._schema.form_type or "H")  # type: ignore
```
MebinAbraham, line 140: This ignore can be removed, it is complaining cause it is a `.get()`. It should be key access. `self._schema.form_type or "H"` will always resolve to an existing key.
```suggestion
            for value in detail_answers_option_map[self._schema.form_type or "H"]
```

app/views/handlers/feedback.py
```diff
@@ -211,32 +225,118 @@ def is_enabled(schema: QuestionnaireSchema) -> bool:
...
-        self.feedback_count = feedback_count
-        self.feedback_submission_date = datetime.now(timezone.utc).strftime("%Y-%m-%d")
-        self.form_type = form_type
-        self.language_code = language_code
-        self.region_code = region_code
+    def __init__(self, case_id: str, tx_id: str):
+        self.case_id = case_id
         self.tx_id = tx_id
 
     def __call__(self) -> Mapping:
```
MebinAbraham, line 232: Worth updating this to be explicit:
```suggestion
    def __call__(self) -> dict[str, str]
```

app/views/handlers/feedback.py
```diff
@@ -211,32 +225,118 @@ def is_enabled(schema: QuestionnaireSchema) -> bool:
...
+        self.feedback_count = feedback_count
         self.feedback_text = feedback_text
         self.feedback_type = feedback_type
         self.feedback_type_question_category = feedback_type_question_category
 
-    def __call__(self) -> Mapping:
-        payload = vars(self)
-        if not self.feedback_type_question_category:
-            del payload["feedback_type_question_category"]
+    def __call__(self) -> Mapping[str, Any]:
```
MebinAbraham, line 309:
```suggestion
    def __call__(self) -> dict[str, Any]:
```

PR #695, opened 2021-10-20
Add back to surveys link to the thank you page

berroar:

Adds a `Back to surveys` link to the `Thank You` page. This link directs to a RAS page outside of eQ that lists the respondents surveys which are left to do.

### How to review 

Check the functionality matches that detailed on the Trello card.

### Checklist

### DIFF
--- a/test_schemas/en/test_new_routing_not_affected_by_answers_not_on_path.json
+++ b/test_schemas/en/test_new_routing_not_affected_by_answers_not_on_path.json
@@ -152,16 +152,17 @@
                             "content": {
                                 "title": "This page should have been skipped!"
                             },
-                            "skip_conditions": [
-                                {
-                                    "when": [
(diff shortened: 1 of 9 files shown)


LJBabbage (commented):
Goes a bit wonky when run locally as it tries to go to http://localhost:8000/surveys/todo, but I am thinking Petes PR fixes that?

LJBabbage:
> > Goes a bit wonky when run locally as it tries to go to http://localhost:8000/surveys/todo, but I am thinking Petes PR fixes that?
> 
> Only when no cookie is set, it would still be going to that 404 page when you have a session as launcher sets the url, but that is something we have to accept and why we decided not to set the default url to launcher in staging/test.

Of course launcher sets that. but realistically if it is the default we should cater for it on launcher. Even if it is in a dev environment we should strive to make things work.

PR #696, opened 2021-10-21
Add session expiry end point

LJBabbage:
This PR adds the session-expiry end point, which returns expires_at in response object with the application/json mimetype

### How to review 
Launch a questionnaire and then visit /session-expiry to check it works and changes with time

### Checklist

### DIFF
--- a/app/routes/session.py
+++ b/app/routes/session.py
@@ -121,9 +121,9 @@ def get_session_expired():
     return render_template("errors/401")
 
 
-@session_blueprint.route("/session-expiry", methods=["GET"])
+@session_blueprint.route("/session-expiry", methods=["GET", "PATCH"])
 @login_required
-def get_session_expiry():
+def session_expiry():
     return jsonify(expires_at=get_session_store().expiration_time.isoformat())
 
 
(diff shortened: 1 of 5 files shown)


app/routes/session.py
```diff
@@ -121,6 +121,12 @@ def get_session_expired():
     return render_template("errors/401")
 
 
+@session_blueprint.route("/session-expiry", methods=["GET"])
```
MebinAbraham, line 124: Was there a reason this wasn't `session/expiry`?. That is what was intended by the card. I think it should live off the session route, I think the same for the existing endpoint in this file, but that don't need to change here.

MebinAbraham (changes requested):

app/routes/session.py
LJBabbage: You're 18 days late on this request and what are you doing going through old stories :laughing:. It was just aligned with the other endpoints. Being only a spike and already being worked on by Matt I don't think we should change it at the moment, but can before it gets merged

MebinAbraham: PR still open... would have asked if I wasn't on holiday. Doesn't need to change while Matt is developing, just a simple question to see if there was a reason it doesn't match the card or just an oversight, maybe a covo that happened that I wasn't aware off... no need for the snarky comment 🤦

LJBabbage: Not snarky, I used the laughing emoji.

PR #703, opened 2021-11-04
Add collection exercise sid to state

LJBabbage:
Add collection_exercise_sid to state

### How to review 
Using this branch locally and/or on GCP make sure collection-exercise-sid is added to state

### Checklist

### DIFF
--- a/app/data_models/questionnaire_store.py
+++ b/app/data_models/questionnaire_store.py
@@ -22,7 +22,12 @@ class QuestionnaireStore:
         self.answer_store = AnswerStore()
         self.progress_store = ProgressStore()
 
-        raw_data, version, self.submitted_at = self._storage.get_user_data()
+        (
+            raw_data,
+            self.collection_exercise_sid,
+            version,
+            self.submitted_at,
+        ) = self._storage.get_user_data()
 
         if raw_data:
             self._deserialize(raw_data)
@@ -69,4 +74,11 @@ class QuestionnaireStore:
 
     def save(self):
         data = self.serialize()
-        self._storage.save(data=data, submitted_at=self.submitted_at)
+        collection_exercise_sid = (
+            self.collection_exercise_sid or self._metadata["collection_exercise_sid"]
+        )
+        self._storage.save(
+            data=data,
+            collection_exercise_sid=collection_exercise_sid,
+            submitted_at=self.submitted_at,
+        )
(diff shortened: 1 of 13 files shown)


app/data_models/questionnaire_store.py
```diff
@@ -69,4 +74,11 @@ def delete(self):
 
     def save(self):
         data = self.serialize()
-        self._storage.save(data=data, submitted_at=self.submitted_at)
+        collection_exercise_sid = (
```
LJBabbage, line 77: This isn't the best. I would prefer to do this one when the state/store is created, but it's stuck in 

```
def get_questionnaire_store(user_id: str, user_ik: str) -> QuestionnaireStore:
    # Sets up a single QuestionnaireStore instance per request.
    store = g.get("_questionnaire_store")
    if store is None:
        secret_store = current_app.eq["secret_store"]  # type: ignore
        pepper = secret_store.get_secret_by_name(
            "EQ_SERVER_SIDE_STORAGE_ENCRYPTION_USER_PEPPER"
        )
        storage = EncryptedQuestionnaireStorage(user_id, user_ik, pepper)
        # pylint: disable=assigning-non-slot
        store = g._questionnaire_store = QuestionnaireStore(storage)

    return store
```
Unfortunately the metadata is not added in authenticator until after the state/store is created.  I could pass it into  get_questionnaire_store but we use that in lots of places just to get the store, effectively knowing it will be there, so passing in metadata isn't really appropriate. We perhaps need to consider separating out  the getting and the creating of state/store. Open to other ideas if anyone has them

tests/app/storage/test_encrypted_questionnaire_storage.py
```diff
@@ -1,23 +1,10 @@
...
-from flask import current_app
-
 from app.data_models import QuestionnaireStore
-from app.data_models.app_models import QuestionnaireState
 from app.storage.encrypted_questionnaire_storage import EncryptedQuestionnaireStorage
-from app.storage.storage_encryption import StorageEncryption
 from tests.app.app_context_test_case import AppContextTestCase
 
 
-def _save_state_data(user_id, data, state_version=QuestionnaireStore.LATEST_VERSION):
```
LJBabbage, line 12: This wasn't being used

app/data_models/questionnaire_store.py
MebinAbraham: Given ce id isn't needed to identify the state, I think this is fine for where we are.

MebinAbraham (approved):
LGTM

berroar (approved):
Tested in a staging env and the `collection_exercise_sid` is being exposed 👍

jamesnunn:
Tested, sid is there

jamesnunn (approved):

PR #704, opened 2021-11-08
Add new rules engine for section enabled rules

petechd:
This adds new rules engine for the section enabled element of schemas, based on element type it will now evaluate `when` rules using old or new when rule evaluation. All schemas that use that used old type of when rules in section enabled has been updated and added to schemas with `new` prefix.

### How to review 
Run new test schemas manually or run unit tests.

### Checklist

### DIFF
--- a/app/questionnaire/router.py
+++ b/app/questionnaire/router.py
@@ -1,4 +1,4 @@
-from typing import Optional, Sequence, Tuple
+from typing import Optional, Tuple
 
 from flask import url_for
 
@@ -272,17 +272,7 @@ class Router:
             return True
 
         enabled = section["enabled"]
-        if isinstance(enabled, Sequence):
-            for condition in enabled:
-                if evaluate_when_rules(
-                    condition["when"],
-                    self._schema,
-                    self._metadata,
-                    self._answer_store,
-                    self._list_store,
-                ):
-                    return True
-        else:
+        if isinstance(enabled, dict):
             location = Location(section_id=section["id"])
             when_rule_evaluator = WhenRuleEvaluator(
                 self._schema,
@@ -294,7 +284,18 @@ class Router:
                 routing_path_block_ids=None,
             )
 
-            return when_rule_evaluator.evaluate(enabled.get("when"))
+            return when_rule_evaluator.evaluate(enabled["when"])
+
+        else:
+            for condition in enabled:
+                if evaluate_when_rules(
+                    condition["when"],
+                    self._schema,
+                    self._metadata,
+                    self._answer_store,
+                    self._list_store,
+                ):
+                    return True
 
         return False
 
--- a/app/questionnaire/router.py
+++ b/app/questionnaire/router.py
@@ -273,14 +273,13 @@ class Router:
 
         enabled = section["enabled"]
         if isinstance(enabled, dict):
-            location = Location(section_id=section["id"])
             when_rule_evaluator = WhenRuleEvaluator(
                 self._schema,
                 self._answer_store,
                 self._list_store,
                 self._metadata,
                 self._response_metadata,
-                location=location,
+                location=None,
                 routing_path_block_ids=None,
             )
 
--- a/app/questionnaire/router.py
+++ b/app/questionnaire/router.py
@@ -1,9 +1,10 @@
-from typing import Optional, Tuple
+from typing import Optional, Sequence, Tuple
 
 from flask import url_for
 
 from app.questionnaire.location import Location
 from app.questionnaire.path_finder import PathFinder
+from app.questionnaire.routing.when_rule_evaluator import WhenRuleEvaluator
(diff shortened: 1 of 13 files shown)


petechd:
[Runner to use new rules engine for section enabled rules (3 SP)](https://trello.com/c/JKht4nws/4948-runner-to-use-new-rules-engine-for-section-enabled-rules-3-sp)

app/questionnaire/router.py
```diff
@@ -270,15 +271,31 @@ def _is_section_enabled(self, section):
         if "enabled" not in section:
             return True
 
-        for condition in section["enabled"]:
-            if evaluate_when_rules(
-                condition["when"],
+        enabled = section["enabled"]
+        if isinstance(enabled, Sequence):
```
MebinAbraham, line 275: To be consistent with pathfinder, I think we should flip this logic. So check if it is a dict first then handle that, then handle list/tuple which is the old after.

app/questionnaire/router.py
```diff
@@ -270,15 +271,31 @@ def _is_section_enabled(self, section):
...
                 self._list_store,
-            ):
-                return True
+                self._metadata,
+                self._response_metadata,
+                location=location,
+                routing_path_block_ids=None,
+            )
+
+            return when_rule_evaluator.evaluate(enabled.get("when"))
```
MebinAbraham, line 297: The `when` key has to exist so don't need `.get()`.

app/questionnaire/router.py
```diff
@@ -270,15 +271,31 @@ def _is_section_enabled(self, section):
...
+                if evaluate_when_rules(
+                    condition["when"],
+                    self._schema,
+                    self._metadata,
+                    self._answer_store,
+                    self._list_store,
+                ):
+                    return True
+        else:
+            location = Location(section_id=section["id"])
```
MebinAbraham, line 286: I don't follow why this has been done?.

test_schemas/en/test_new_section_enabled_checkbox.json
```diff
@@ -0,0 +1,168 @@
...
+                {
+                    "when": [
+                        {
+                            "id": "section-1-answer",
+                            "condition": "contains",
+                            "value": "Section 3"
+                        }
+                    ]
+                }
+            ],
```
MebinAbraham, line 139: This is the old rules structure.

MebinAbraham (commented):
Not sure where we stand anymore with tests, but we need to update the references in test to the new schema or duplicate them as the new ones are being tested.

test_schemas/en/test_new_section_enabled_checkbox.json
petechd: Apologies, somehow it got into the new schema, fixed now, thanks!

app/questionnaire/router.py
petechd: Removed unnecessary `.get()`, thanks!

app/questionnaire/router.py
petechd: Logic flipped now, so first checking for dict and then `else`, thanks!

app/questionnaire/router.py
petechd: So I removed it from here and made `location` optional in `WhenRuleEvaluator` but that lead to new line of code, assigning `list_item_id` a conditional value as it is part of Location class (and that is done in some other place in runner as well):
`list_item_id = self.location.list_item_id if self.location else None`

petechd:
All the tests are referencing new schemas now except one that I had to keep in order to get 100% coverage (old evaluator engine code was missing tests).

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -19,19 +19,20 @@ class WhenRuleEvaluator:
     list_store: ListStore
     metadata: Mapping
     response_metadata: Mapping
-    location: Union[Location, RelationshipLocation]
+    location: Optional[Union[Location, RelationshipLocation]]
```
MebinAbraham, line 22: As per the style guide, optional when we have one type. 
```suggestion
    location: Union[None, Location, RelationshipLocation]
```

MebinAbraham (approved):

app/questionnaire/router.py
```diff
@@ -270,16 +271,30 @@ def _is_section_enabled(self, section):
...
                 self._schema,
                 self._metadata,
                 self._answer_store,
                 self._list_store,
-            ):
-                return True
-        return False
+            )
+            for condition in enabled
+        )
```
MebinAbraham, line 297: Nice use of `any`.

petechd: I was not sure if you'll prefer simplicity over readability here but "simple is better than complex".

LJBabbage (approved):

PR #707, opened 2021-11-09
Add type hinting for submitter

petechd:
This adds type hints for `submitter.py` as it's one of the key areas of the application and we recently had a bug that could be easily avoided if we used them.

### How to review 
Run `./scripts/run_mypy.sh` and check if all the type hints are correct.

### Checklist

### DIFF
--- a/app/submitter/submitter.py
+++ b/app/submitter/submitter.py
@@ -155,7 +155,7 @@ class GCSFeedbackSubmitter:
 
 class LogFeedbackSubmitter:
     @staticmethod
-    def upload(metadata: MetadataType, payload: dict) -> bool:
+    def upload(metadata: MetadataType, payload: str) -> bool:
         logger.info("uploading feedback")
         logger.info(
             "feedback message",
--- a/app/submitter/submitter.py
+++ b/app/submitter/submitter.py
@@ -98,7 +98,7 @@ class RabbitMQSubmitter:
         except AMQPError as e:
             logger.error("unable to close connection", exc_info=e, category="rabbitmq")
 
-    def send_message(self, message: str, tx_id: str, case_id: str) -> Union[None, bool]:
+    def send_message(self, message: str, tx_id: str, case_id: str) -> Optional[bool]:
         """
         Sends a message to rabbit mq and returns a true or false depending on if it was successful
         :param message: The message to send to the rabbit mq queue
--- a/app/submitter/submitter.py
+++ b/app/submitter/submitter.py
@@ -143,7 +143,7 @@ class GCSFeedbackSubmitter:
         client = storage.Client()
         self.bucket = client.get_bucket(bucket_name)
 
-    def upload(self, metadata, payload):
+    def upload(self, metadata: Mapping[str, str], payload: str) -> bool:
         blob = self.bucket.blob(str(uuid4()))
         blob.metadata = metadata
         blob.upload_from_string(payload.encode("utf8"))
--- a/app/submitter/submitter.py
+++ b/app/submitter/submitter.py
@@ -1,4 +1,4 @@
-from typing import Mapping, Optional, Union
+from typing import Mapping, Optional
 from uuid import uuid4
 
 from google.cloud import storage  # type: ignore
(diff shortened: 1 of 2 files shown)


petechd:
[Type hinting for submitter (2SP)](https://trello.com/c/7R6IibV1/4958-type-hinting-for-submitter-2sp)

app/submitter/submitter.py
```diff
@@ -149,7 +160,7 @@ def upload(self, metadata, payload):
 
 class LogFeedbackSubmitter:
     @staticmethod
-    def upload(metadata, payload):
+    def upload(metadata: Mapping[str, Union[str, int, list]], payload: dict) -> bool:
```
parimooa, line 163: Consider making MetadataType as constant with `MetadataType=Mapping[str, Union[str, int, list]]` since it i used few times in code.

app/submitter/submitter.py
```diff
@@ -81,15 +90,15 @@ def _connect(self):
...
+    def _disconnect(connection: Optional[BlockingConnection]) -> None:
         try:
             if connection:
                 logger.info("attempt to close connection", category="rabbitmq")
                 connection.close()
         except AMQPError as e:
             logger.error("unable to close connection", exc_info=e, category="rabbitmq")
 
-    def send_message(self, message, tx_id, case_id):
+    def send_message(self, message: str, tx_id: str, case_id: str) -> Union[None, bool]:
```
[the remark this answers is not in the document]
MebinAbraham: When does this return `None`?.

app/submitter/submitter.py
```diff
@@ -132,11 +141,13 @@ def send_message(self, message, tx_id, case_id):
...
 
 class GCSFeedbackSubmitter:
-    def __init__(self, bucket_name):
+    def __init__(self, bucket_name: str) -> None:
         client = storage.Client()
         self.bucket = client.get_bucket(bucket_name)
 
-    def upload(self, metadata, payload):
+    def upload(
+        self, metadata: Mapping[str, Union[str, int, list]], payload: dict
```
MebinAbraham, line 149: FeedbackMetadata's `__call__` return type is `dict[str, str]`, so this should be `Mapping[str, str]`?

If Lewis' PR merges first, the type for `payload` will change to a `str`.

app/submitter/submitter.py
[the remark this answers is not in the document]
petechd: Since we use `try` that doesn't `return` anything I gave it `None` but changed it to `Optional`, as @parimooa suggested.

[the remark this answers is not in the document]
MebinAbraham: But it always returns a bool, the try doesn't escape out any other way?. That methods will always return a bool or raises an exception?.

app/submitter/submitter.py
petechd: Changed to `Mapping[str, str]`, keeping payload:dict for now as Lewis's PR didn't merge yet.

app/submitter/submitter.py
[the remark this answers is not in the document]
petechd: Fixed, I don't always get how type hints work yet, thanks!

app/submitter/submitter.py
petechd: It's only used once once now so I won't set a constant, but would do otherwise, thanks!

MebinAbraham: The type for the metadata should match between the GCS/Log submitter.

petechd: So I made them a const as @parimooa suggested and re-use it now in both submitters, thanks!

parimooa (approved):

app/submitter/submitter.py
```diff
@@ -144,7 +155,7 @@ def upload(self, metadata, payload):
 
 class LogFeedbackSubmitter:
     @staticmethod
-    def upload(metadata, payload):
+    def upload(metadata: MetadataType, payload: dict) -> bool:
```
MebinAbraham, line 158:
```suggestion
    def upload(metadata: MetadataType, payload: str) -> bool:
```

MebinAbraham (approved):

PR #708, opened 2021-11-09
Move feedback metadata update before encryption

LJBabbage:
Feedback metadata was being updated after encryption meaning this would break using GCSFeedbackSubmitter, this moves if before encryption

### How to review 
Check locally and on GCP feedback now works

### Checklist

### DIFF
--- a/app/views/handlers/feedback.py
+++ b/app/views/handlers/feedback.py
@@ -21,7 +21,6 @@ from app.submitter.converter import (
     build_metadata,
     get_optional_payload_properties,
 )
-from app.utilities.json import json_dumps
 from app.views.contexts.feedback_form_context import build_feedback_context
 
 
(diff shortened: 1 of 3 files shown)


app/views/handlers/feedback.py
```diff
@@ -105,8 +104,8 @@ def handle_post(self) -> None:
             feedback_type=self.form.data.get("feedback-type"),
         )
 
-        message = json_dumps(feedback_message())
-
+        message = feedback_message()
+        message.update(feedback_metadata())
```
MebinAbraham, line 108: `feedback_metadata()` should be assigned to a var and reused. It is being created twice at the moment, once here and once on upload.

MebinAbraham (approved):
Working as expected in GCS.

PR #709, opened 2021-11-09
Conditionally extend session from PATCH/GET /session-expiry

jamesnunn:
The /session-expiry endpoints needed to be provided for PATCH and GET; PATCH will extend the session and return the updated expires_at value, GET will not extend the session and return the current expires_at value.

### How to review 

- Launch a survey
- Within 1 min, go to http://localhost:5000/session-expiry to show the GET request returns the initial expires_at timestamp
- Also within 1 min, run this command from chrome dev tools to show the PATCH request returns the inital expires_at timestamp 
```
fetch('http://localhost:5000/session-expiry', {
  method: 'PATCH',
  headers: {
    'Content-type': 'application/json; charset=UTF-8'
  }
})
.then(res => res.json())
.then(console.log)
```
- Wait for longer than 1 min
- Go to http://localhost:5000/session-expiry again to show the GET endpoint doesn't update the expires_at time
- Send the PATCH request again and check the expires_at time has been updated to 45mins from now
- Go to http://localhost:5000/session-expiry yet again to show the GET response reflects the time set by the PATCH request.

### Checklist

### DIFF
--- a/app/authentication/authenticator.py
+++ b/app/authentication/authenticator.py
@@ -32,7 +32,7 @@ def user_loader(user_id: str) -> Optional[str]:
 def request_load_user(request: Request) -> Optional[User]:
     logger.debug("load user")
 
-    if request.endpoint:
+    if request and request.endpoint:
         extend_session = not (
             request.endpoint == "session.session_expiry" and request.method == "GET"
         )
--- a/app/authentication/authenticator.py
+++ b/app/authentication/authenticator.py
@@ -30,16 +30,22 @@ def user_loader(user_id: str) -> Optional[str]:
 
 @login_manager.request_loader
 def request_load_user(
-    request: Request,
-) -> Optional[User]:  # pylint: disable=unused-argument
+    request: Request,  # pylint: disable=unused-argument
+) -> Optional[User]:
     logger.debug("load user")
-    return load_user()
+    if request:
+        extend_session = not (
+            "session-expiry" in request.path and request.method == "GET"
+        )
+    else:
+        extend_session = True
+    return load_user(extend_session)
 
 
 @user_logged_out.connect_via(ANY)
 def when_user_logged_out(
-    sender_app: Flask, user: str
-) -> None:  # pylint: disable=unused-argument
+    sender_app: Flask, user: str  # pylint: disable=unused-argument
+) -> None:
     logger.debug("log out user")
     session_store = get_session_store()
     if session_store:
@@ -82,7 +88,7 @@ def _is_session_valid(session_store: SessionStore) -> bool:
     )
 
 
-def load_user() -> Optional[User]:
+def load_user(extend_session=True) -> Optional[User]:
     """
     Checks for the present of the JWT in the users sessions
     :return: A user object if a JWT token is available in the session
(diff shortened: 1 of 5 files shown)


jamesnunn:
[Prevent session-expiry end point extending the session expiry (1SP)](https://trello.com/c/5BT1W6om/4957-prevent-session-expiry-end-point-extending-the-session-expiry-1sp)

MebinAbraham (commented):
Functionally, as expected.

app/authentication/authenticator.py
```diff
@@ -30,16 +30,24 @@ def user_loader(user_id: str) -> Optional[str]:
...
 @login_manager.request_loader
 def request_load_user(
-    request: Request,
-) -> Optional[User]:  # pylint: disable=unused-argument
+    request: Request,  # pylint: disable=unused-argument
+) -> Optional[User]:
     logger.debug("load user")
-    return load_user()
+
+    if request:
```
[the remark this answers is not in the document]
jamesnunn: Running tests, there was at least one case where the request was `None` so I assumed so

tests/integration/routes/test_session.py
```diff
@@ -63,18 +63,39 @@ def test_head_request_on_session_signed_out(self):
...
+    def test_get_session_expiry_doesnt_extend_session(self):
         self.launchSurvey()
-        self.get("/session-expiry")
+        # Advance time by 20 mins...
+        with freeze_time(TIME_TO_FREEZE + timedelta(seconds=20 * 60)):
+            self.get("/session-expiry")
+            response = self.getResponseData()
+            parsed_json = json_loads(response)
+            # ... check that the session expiry time is not affected by
+            # the request, and is still 45mins from the start time
```
LJBabbage, line 74: It's fine, but naturally this might not be 45 minutes in the future as it can be set.

LJBabbage (approved):
Looks and works as expected, a couple of minor comments, but not worth blocking over. Particularly enjoyed the code snippet in the PR which made things quicker to test

tests/integration/routes/test_session.py
jamesnunn: True, I'll set it in the setup to fix it

app/authentication/authenticator.py
```diff
@@ -29,17 +29,23 @@ def user_loader(user_id: str) -> Optional[str]:
...
-def request_load_user(
-    request: Request,
-) -> Optional[User]:  # pylint: disable=unused-argument
+def request_load_user(request: Request) -> Optional[User]:
     logger.debug("load user")
-    return load_user()
+
+    if request.endpoint:
+        extend_session = not (
+            "session_expiry" in request.endpoint and request.method == "GET"
```
MebinAbraham, line 37: I think this should be checked explicitly.
```suggestion
            "session.session_expiry" == request.endpoint and request.method == "GET"
```

app/authentication/authenticator.py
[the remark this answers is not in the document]
MebinAbraham: Can we remove this then?

MebinAbraham (approved):

PR #710, opened 2021-11-11
Add type hints to the storage module

MebinAbraham:
- Added type hints to the `app/storage` module.
- Moved `app/storage/metadata_parser.py` to `app/utilities/` as I am not sure why it is in the storage module, hence also not added type hints for that file.

### How to review 

Ensure:
- the type hints are valid
- app works as expected

### Checklist


app/storage/storage.py
```diff
@@ -29,50 +51,53 @@ class StorageModel:
...
     @cached_property
-    def index_fields(self):
+    def index_fields(self) -> list[str]:
         return self._config.get("index_fields", [])
 
     @cached_property
-    def table_name(self):
-        return current_app.config[self._config["table_name_key"]]
+    def table_name(self) -> str:
+        table_name: str = current_app.config[self._config["table_name_key"]]
```
parimooa, line 77: function and variable name is same , can it be just table ?

app/storage/dynamodb.py
```diff
@@ -20,14 +26,15 @@ def put(self, model, overwrite=True):
...
+            succeeded: bool = response == 200
+            return succeeded
         except ClientError as e:
             if e.response["Error"]["Code"] == "ConditionalCheckFailedException":
                 raise ItemAlreadyExistsError() from e
 
             raise  # pragma: no cover
 
-    def get(self, model_type, key_value):
+    def get(self, model_type: Type[ModelTypes], key_value: str) -> Optional[ModelTypes]:
```
parimooa, line 37: why `Type` ?

parimooa (approved):
LGTM

app/storage/dynamodb.py
MebinAbraham: Because we use the type to look u the model. It is not an instance here, e.g:
https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/data_models/session_store.py#L87

berroar (approved):

PR #712, opened 2021-11-12
Add type hints to remaining submitter module files

petechd:
- Added remaining type hints to the `app/submitter` module (main `submitter.py` file has been refactor in earlier PR).
- Added two `# type: ignore` where refactoring makes the changes out of scope and topic of another card (eq. serious refactoring of `Relatioship Store`).

### How to review 

Ensure:
- the type hints are valid
- app works as expected

### Checklist


rmccar:
[ons-questionnaire-schema Github Actions (2 SP)](https://trello.com/c/oL3M4o8X/4960-ons-questionnaire-schema-github-actions-2-sp)

petechd:
[Add remaining type hints to the runner's submitter module (1 SP)](https://trello.com/c/k3UbGpZn/4965-add-remaining-type-hints-to-the-runners-submitter-module-1-sp)

app/submitter/convert_payload_0_0_1.py
```diff
@@ -73,48 +82,55 @@ def convert_answers_to_payload_0_0_1(
...
-    if answer_type == "Date":
-        return (
-            datetime.strptime(answer_value, "%Y-%m-%d")
-            .replace(tzinfo=timezone.utc)
-            .strftime("%d/%m/%Y")
-        )
+def _format_downstream_answer(
+    answer_type: str, answer_value: AnswerValueTypes, answer_data: str
+) -> str:
+    if isinstance(answer_value, str):
```
LJBabbage, line 88: This is interesting as I don't think Date and MonthYearDate answer_value can be anything but string however due to the generic nature of _format_downstream_answer and type hinting I can see why this has to be done

LJBabbage (approved):
LGTM

app/submitter/convert_payload_0_0_1.py
petechd: Some of these need some serious refactoring..

app/submitter/convert_payload_0_0_3.py
```diff
@@ -85,23 +95,29 @@ def add_list_collector_answers(
     list_item_ids = list_store[list_name].items
 
     for list_item_id in list_item_ids:
-        for answer_id in answers_ids_in_add_block:
-            answer = answer_store.get_answer(answer_id, list_item_id)
-            if answer:
-                answers_payload.add_or_update(answer)
+        if answers_ids_in_add_block:
```
MebinAbraham, line 98: This should be outside the loop. We can just return early if it doesn't exist.

app/submitter/convert_payload_0_0_1.py
```diff
@@ -1,14 +1,23 @@
...
 # pylint: disable=too-many-locals,too-many-nested-blocks
 def convert_answers_to_payload_0_0_1(
-    metadata, answer_store, list_store, schema, full_routing_path
-):
+    metadata: Mapping[str, Union[str, int, list]],
+    answer_store: AnswerStore,
+    list_store: ListStore,
+    schema: QuestionnaireSchema,
+    full_routing_path: RoutingPath,
+) -> OrderedDict:
```
MebinAbraham, line 20:
```suggestion
) -> OrderedDict[str, Any]:
```

app/submitter/convert_payload_0_0_1.py
```diff
@@ -52,7 +61,7 @@ def convert_answers_to_payload_0_0_1(
                     if answer["id"] == answer_in_block.answer_id:
                         answer_schema = answer
 
-                value = answer_in_block.value
+                value: AnswerValueTypes = answer_in_block.value
```
MebinAbraham, line 64: Not sure this annotation is needed as it should already know that if it knows it is an answer?.

app/submitter/convert_payload_0_0_1.py
```diff
@@ -73,48 +82,55 @@ def convert_answers_to_payload_0_0_1(
...
-                if option["value"] == user_answer
-            ),
-            None,
-        )
-
-        if option:
-            if "detail_answer" in option:
-                detail_answer = answer_store.get_answer(option["detail_answer"]["id"])
+    if isinstance(value, Iterable) and value:
+        for user_answer in value:
```
MebinAbraham, line 111: I don't think we should modify the code in this way for this function, as this is only ever called when we know it is a checkbox. Therefore, I would update the type for value to `ListAnswer` and add a type ignore to callers line. That way, code changes are not needed to please type hints as the `if` would always be true.

We need to refactor the type for answer value and set it to `Any` so we can override it. That is why we have this problem, but one for another card.

app/submitter/convert_payload_0_0_1.py
```diff
@@ -73,48 +82,55 @@ def convert_answers_to_payload_0_0_1(
...
+                .strftime("%m/%Y")
+            )
 
     return answer_data
 
 
-def _get_checkbox_answer_data(answer_store, answer_schema, value):
+def _get_checkbox_answer_data(
+    answer_store: AnswerStore, answer_schema: Mapping, value: AnswerValueTypes
+) -> dict:
```
MebinAbraham, line 108: This should be more specific. `dict[str, str]` should be correct as q_code and option values are strings.

MebinAbraham (commented):
Just a few minor issues. Does highlight we need to change the way answer values are typed. It should be typed as `Any` so it can be overridden (seperate card).

app/submitter/convert_payload_0_0_1.py
```diff
@@ -1,14 +1,23 @@
...
 from app.questionnaire.location import Location
+from app.questionnaire.routing_path import RoutingPath
 from app.questionnaire.schema_utils import choose_question_to_display
 
 
 # pylint: disable=too-many-locals,too-many-nested-blocks
 def convert_answers_to_payload_0_0_1(
-    metadata, answer_store, list_store, schema, full_routing_path
-):
+    metadata: Mapping[str, Union[str, int, list]],
```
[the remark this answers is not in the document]
petechd: So this one was somehow hard to implement, the `MetadataType` in `submitter.py` has different types (`Mapping[str, str]`) when in this file and `converter.py` the type has to be `Mapping[str, Union[str, int, list]]`. I tried to at least import `MetadataType` const from `converter.py` to `convert_payload_0_0_1.py` but that gave me `circular import` error (both files would import from each other). So I gave up and defined `MetadatType` again in `convert_payload_0_0_1.py` as it's used twice in this file.

app/submitter/converter.py
```diff
@@ -1,26 +1,38 @@
...
+    DEFAULT_LANGUAGE_CODE,
+    QuestionnaireSchema,
+)
+from app.questionnaire.routing_path import RoutingPath
 from app.submitter.convert_payload_0_0_1 import convert_answers_to_payload_0_0_1
 from app.submitter.convert_payload_0_0_3 import convert_answers_to_payload_0_0_3
 
 logger = get_logger()
 
+MetadataType = Mapping[str, Union[str, int, list]]
```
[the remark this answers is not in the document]
petechd: See above

app/submitter/converter.py
```diff
@@ -1,26 +1,38 @@
...
 
 def convert_answers(
-    schema, questionnaire_store, routing_path, submitted_at, flushed=False
-):
+    schema: QuestionnaireSchema,
+    questionnaire_store: QuestionnaireStore,
+    routing_path: RoutingPath,
+    submitted_at: datetime,
+    flushed: bool = False,
+) -> dict[str, Mapping[str, str]]:
```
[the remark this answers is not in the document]
petechd: Added, thanks!

app/submitter/convert_payload_0_0_3.py
petechd: Moved it outside the loop, thanks!

app/submitter/convert_payload_0_0_1.py
petechd: Removed, and it was unnecessary indeed, thanks!

app/submitter/convert_payload_0_0_1.py
petechd: This whole `if` conditional and nesting removed, ignore added to `value` where it's being called as the type is `ListAnswer` now, see below.

app/submitter/convert_payload_0_0_1.py
petechd: More specific now, thanks!

app/submitter/convert_payload_0_0_1.py
```diff
@@ -107,14 +123,13 @@ def _get_checkbox_answer_data(answer_store, answer_schema, value):
...
-
                 # if the user has selected an option with a detail answer we need to find the detail answer value it refers to.
                 # the detail answer value can be empty, in this case we just use the main value (e.g. other)
-                user_answer = detail_answer.value or user_answer
+                user_answer = detail_answer.value or user_answer  # type: ignore
 
             qcodes_and_values.append((option.get("q_code"), user_answer))
 
-    checkbox_answer_data = OrderedDict()
+    checkbox_answer_data: dict = OrderedDict()
```
MebinAbraham, line 132: Is this needed?. If so, it should match the return type.
```suggestion
    checkbox_answer_data: dict[str, str] = OrderedDict()
```

app/submitter/convert_payload_0_0_3.py
```diff
@@ -83,25 +93,30 @@ def add_list_collector_answers(
     )
     list_name = list_collector_block["for_list"]
     list_item_ids = list_store[list_name].items
-
-    for list_item_id in list_item_ids:
-        for answer_id in answers_ids_in_add_block:
-            answer = answer_store.get_answer(answer_id, list_item_id)
-            if answer:
-                answers_payload.add_or_update(answer)
+    if answers_ids_in_add_block:
```
MebinAbraham, line 96: Ideally this should return early instead of nesting, but not a blocker.

MebinAbraham (approved):

app/submitter/convert_payload_0_0_1.py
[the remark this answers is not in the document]
MebinAbraham: Ya, it is just confusing to have multiple metadata types of different types. The files will both have to import from a separate file, or you will indeed get that circular import. But ya, keep it as is.

app/submitter/convert_payload_0_0_3.py
petechd: Yeah, I tried that and got coverage error, so would have to write bunch of tests for this if we return early so I just did the nesting as it's less intrusive.

PR #713, opened 2021-11-15
Prevent non-test schemas from being part of acceptance test

petechd:
Prevents non-test schemas from being part of acceptance tests (and pages generation/linting process) in runner, which solves our problem with social schema. This would fix the build process issue in Social demo PR.

### How to review 
In the build check if non-test schemas are being skipped (for example `mbs_0253.json`.) and the appropriate message is being displayed.

### Checklist

### DIFF
--- a/tests/functional/generate_pages.py
+++ b/tests/functional/generate_pages.py
@@ -893,15 +893,19 @@ if __name__ == "__main__":
             for root, dirs, files in os.walk(args.SCHEMA):
                 for file in [os.path.join(root, file) for file in files]:
                     filename = os.path.basename(file)
-                    logger.info("File %s", filename)
-                    if filename[0] == ".":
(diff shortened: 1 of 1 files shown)


petechd:
[Prevent non-test schemas from being part of acceptance tests](https://trello.com/c/tnsLPuKB/4970-prevent-non-test-schemas-from-being-part-of-acceptance-tests)

LJBabbage (commented):
I am not sure if we need to do this. There is a card in flight that makes a distinction between test schemas and business/social by putting them in different folders (https://trello.com/c/rxXnVcTg/4961-load-schemas-in-runner-build-and-expose-in-schemas-endpoint-2-sp. Waiting for that, then pointing at the correct folder feels like a neater and better solution

PR #715, opened 2021-11-15
Adds `response_expires_at` claim parser and validator

jamesnunn:
Adds support for the presence of an optional JWT claim for `response_expires_at`, the format for the datetime being defined [here](https://github.com/ONSdigital/ons-schema-definitions/pull/63/files#diff-a6a25ce0ba7f597b0b758d885e64b9199e649414b317bee97e910e177cc774d7R119)

See also https://github.com/ONSdigital/eq-questionnaire-launcher/pull/34

### How to review 
Check unit tests defined to ensure the claim is correctly interrogated.

### Checklist


jamesnunn:
[Runner to accept response_expires_at launch claim (2 SP)](https://trello.com/c/NRdaA9wK/4968-runner-to-accept-responseexpiresat-launch-claim-2-sp)

app/utilities/metadata_parser.py
```diff
@@ -41,10 +43,27 @@ class DateString(fields.DateTime):
...
+
+        if "%z" in self._out_format:
+            dt, tz = date.split("+")
+            # Add a colon separating the timezone offset hours and minutes
+            if ":" not in date.split("+")[-1]:
+                new_tz = f"{tz[:2]}:{tz[2:]}"
+
+            date = f"{dt}+{new_tz}"
+
+        return date
```
MebinAbraham, line 66: I don't think most of this is needed. Marshmallow has support for ISO 8601 already. Replacing this with the below work as expected for me and the new tests also pass.

```py
class DateString(fields.DateTime):
    """Currently, runner cannot handle Date objects in metadata
    Since all metadata is serialized and deserialized to JSON.
    This custom field deserializes Dates to strings.
    """

    def _deserialize(self, *args, **kwargs):  # pylint: disable=arguments-differ
        date = super()._deserialize(*args, **kwargs)

        if self.format == "iso8601":
            return date.isoformat()

        return date.strftime(self.format)

VALIDATORS = {
    "date": functools.partial(DateString, format="%Y-%m-%d", required=True),
    "uuid": functools.partial(UUIDString, required=True),
    "boolean": functools.partial(fields.Boolean, required=True),
    "string": functools.partial(fields.String, required=True),
    "url": functools.partial(fields.Url, required=True),
    "iso_8601_date": functools.partial(DateString, format="iso8601", required=True),
}
```

- The format arg isn't needed as that is the default, but I included it for clarity. 
- Format can be `iso` or `iso8601`, used `iso8601` to be explicit.
- I think the name should be `iso_8601_date` or `_date_string` for clarity, as it is not a datetime object.

tests/app/parser/test_metadata_parser.py
```diff
@@ -144,6 +144,28 @@ def test_deserialisation_iso_8601_dates(fake_metadata_runner):
     assert isinstance(claims["birthday"], str)
 
 
+def test_deserialisation_iso_8601_datetime(fake_metadata_runner):
+    claims = validate_runner_claims(fake_metadata_runner)
+
+    assert claims["response_expires_at"] == "2021-11-22T15:34:54+00:00"
```
MebinAbraham, line 150: I think we should also test with Zulu time i.e `2021-11-22T15:34:54Z` since that is a valid iso8601 format and Marshmallow support it.

jamesnunn: It does work with `Z` but I specifically avoided testing all possible incidentally supported formats because Mark agreed on the single specific format it should be, which is `2021-11-22T15:34:54+00:00`

app/utilities/metadata_parser.py
jamesnunn: Funny that, I tried all sorts of combinations and it just kept on returning not quite the right format. ta

tests/app/parser/test_metadata_parser.py
MebinAbraham: I get that but given the schema definition doc that got merged says `An ISO_8601 formatted date-time` I think Zulu should be tested otherwise even if it works we won't have confidence in it. They may not use Zulu but if the spec is 8601 then I think it should be tested given there is no caveat made in the doc about not using zulu. I don't think there is anything else to test, not sure what you mean by `all possible incidentally supported formats`?. Zulu is distinctly different to the offset, so just 2 tests is all I see as needed.

jamesnunn: Mark appears to have gone to pains to pin down the exact single format we would support: https://surveydatacollection.slack.com/archives/CLR17KQKU/p1636726964118200?thread_ts=1636726041.118000&cid=CLR17KQKU

MebinAbraham: Had a discussion with @pricem14pc about this and came to the decision for us to add a test for Zulu instead of updating the document specification to says we don't support Zulu.

app/utilities/metadata_parser.py
```diff
@@ -41,10 +43,18 @@ class DateString(fields.DateTime):
     """Currently, runner cannot handle Date objects in metadata
     Since all metadata is serialized and deserialized to JSON.
     This custom field deserializes Dates to strings.
+
+    :param out_format: Output date string format. If falsy, the output
+        format is the same as the input.
```
MebinAbraham, line 48:
```suggestion
```

tests/app/parser/test_metadata_parser.py
```diff
@@ -144,6 +144,35 @@ def test_deserialisation_iso_8601_dates(fake_metadata_runner):
...
+
+
+def test_deserialisation_iso_8601_datetime_with_zulu(fake_metadata_runner):
+    claims = validate_runner_claims(fake_metadata_runner)
+    fake_metadata_runner["response_expires_at"] = "1900-11-22T15:34:54Z"
+
+    assert claims["response_expires_at"] == "2021-11-22T15:34:54+00:00"
```
MebinAbraham, line 157: This doesn't make sense to me. It needs to assigned before you validate, and also need to be a date in the future?

tests/app/parser/conftest.py
```diff
@@ -16,6 +16,7 @@ def fake_metadata_runner():
         "response_id": str(uuid.uuid4()),
         "account_service_url": "https://ras.ons.gov.uk",
         "case_id": str(uuid.uuid4()),
+        "response_expires_at": "2021-11-22T15:34:54+00:00",
```
MebinAbraham, line 19: This and the test data will need to use a dynamically generated date in the future (now + 7 days) or you need to use `freeze_time` from `freezegun`, otherwise after 15:34 on 22nd Nov, these tests will start failing.

parimooa (approved):
LGTM

tests/app/parser/test_metadata_parser.py
jamesnunn: Yep I mucked this one up, should be sorted now.

tests/app/parser/conftest.py
jamesnunn: Done

app/utilities/metadata_parser.py
```diff
@@ -88,6 +98,11 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
         required=False, validate=validate.Length(min=1)
     )  # type:ignore
     case_type = VALIDATORS["string"](required=False)  # type:ignore
+    response_expires_at = VALIDATORS["iso_8601_date_string"](
+        required=False,
+        validate=lambda x: datetime.strptime(x, ISO_8601_DATETIME).astimezone()
```
MebinAbraham, line 103: `astimezone()` without a timezone will use the systems' timezone which we can't garuntee is UTC. So we need to either use `.replace(tzinfo=timezone.utc)` or `.astimezone(tz=timezone.utc)`, we use `.replace()` elsewhere prob cause the time data shouldn't change.

tests/app/parser/test_metadata_parser.py
```diff
@@ -144,6 +145,39 @@ def test_deserialisation_iso_8601_dates(fake_metadata_runner):
...
+
+
+@freeze_time("2021-11-15T15:34:54+00:00")
+def test_deserialisation_iso_8601_datetime_with_zulu(fake_metadata_runner):
+    fake_metadata_runner["response_expires_at"] = "2021-11-22T15:34:54Z"
+    claims = validate_runner_claims(fake_metadata_runner)
+
+    assert claims["response_expires_at"] == "2021-11-22T15:34:54+00:00"
```
MebinAbraham, line 161: These can be parametrized. I think the test is clear enough that we don't need two.
```suggestion
@freeze_time("2021-11-15T15:34:54+00:00")
@pytest.mark.parametrize(
    "date_string",
    ["2021-11-22T15:34:54+00:00", "2021-11-22T15:34:54Z"],
)
def test_deserialisation_iso_8601_date(date_string, fake_metadata_runner):
    fake_metadata_runner["response_expires_at"] = date_string
    claims = validate_runner_claims(fake_metadata_runner)

    assert claims["response_expires_at"] == "2021-11-22T15:34:54+00:00"
```

jamesnunn: Implemented these

MebinAbraham (approved):

PR #717, opened 2021-11-16
Add expires at to state

jamesnunn:
Uses the `response_expires_at` value from https://github.com/ONSdigital/eq-questionnaire-runner/pull/715 and sets the datastore `questionnaire-state.expires_at` to this value.

### How to review 
Complete a survey launched from a GCP environment and check Datastore for the object and that it contains an `expires_at` property, which should be 7 days from now.

### Checklist


jamesnunn:
[Add expires_at to questionnaire_state (2 SP)](https://trello.com/c/6R9ZXgl2/4969-add-expiresat-to-questionnairestate-2-sp)

app/data_models/questionnaire_store.py
```diff
@@ -27,6 +29,7 @@ def __init__(self, storage, version=None):
             self.collection_exercise_sid,
             version,
             self.submitted_at,
+            self.expires_at,
```
LJBabbage, line 32: Is this used? If we are always updating from the metadata, would we ever need to access it from questionnaire_state?

app/data_models/questionnaire_store.py
```diff
@@ -77,8 +80,12 @@ def save(self):
         collection_exercise_sid = (
             self.collection_exercise_sid or self._metadata["collection_exercise_sid"]
         )
+        response_expires_at = self._metadata.get("response_expires_at")
         self._storage.save(
             data=data,
             collection_exercise_sid=collection_exercise_sid,
             submitted_at=self.submitted_at,
+            expires_at=dateutil.parser.parse(response_expires_at)
```
MebinAbraham, line 88: This should be consistent with `metadata_parser.py`, which uses `datetime.strptime()`. I prefer this dateutil.

tests/app/data_model/test_questionnaire_store.py
```diff
@@ -5,6 +5,7 @@
 from app.data_models.answer_store import AnswerStore
```
MebinAbraham, line 5: We should have a test to show that it always uses the metadata value, so state value changes if metadata value changes.

app/data_models/questionnaire_store.py
jamesnunn: So you think it's valid to be returned from get_user_data but just not set as a class attribute?

app/data_models/questionnaire_store.py
jamesnunn: True, and I would prefer dateutil usually too, but in this case I think I'll use strptime to make sure we still have a properly formatted ISO timestamp rather than allowing dateutil to parse any valid timestamp

app/data_models/questionnaire_store.py
LJBabbage: Does it need to be returned by get_user_data, do we ever use it? I could only see it being called from the one place in questionnaire_store.py and tests

app/data_models/questionnaire_store.py
MebinAbraham: Ya, fair enough, but it wouldn't ever get to this if it wasn't ISO 8601. But happy with either.

tests/app/data_model/test_questionnaire_store.py
jamesnunn: I think this isn't needed now, Lewis rightly pointed out the code I added for QuestionnaireStore wasn't being used so I just removed the attribute

app/data_models/questionnaire_store.py
jamesnunn: Yeah not needed for the card so I removed it.

LJBabbage (approved):

MebinAbraham (approved):

PR #718, opened 2021-11-17
Add VPC network  reference in purge

parimooa:
Updated Dataflow purge job parameters including custom VPC `eq` and subnetworks `eq`
Companion [Eq-Terraform-PR](https://github.com/ONSdigital/eq-terraform-gcp/pull/66)

**Background**: Dataflow job uses `default` network  when no network name is provided. So purge job for existing query will fail If `default` network has already been deleted. After introducing custom VPC network, purge job should work without errors as network name is explicitly passed as parameter

Default Network gets created when TF infrastructure code enables 'compute engine api' 

### How to review 
**Pre-requisite**: Create infrastructure with custom VPC ( https://github.com/ONSdigital/eq-terraform-gcp/pull/66)
Delete the `default` network manually 
Run Purge job to ensure expired session and questionnaire state are getting deleted successfully without any errors

### Checklist


ci/purge_expired_questionnaire_state.yaml
```diff
@@ -28,6 +28,8 @@ run:
             datastoreReadGqlQuery="SELECT * FROM \`questionnaire-state\` WHERE submitted_at < $((`date --utc +%s` - EXPIRATION_TIME_OFFSET_IN_SECONDS)) AND submitted_at > NULL",\
             datastoreReadProjectId="${PROJECT_ID}",\
             datastoreDeleteProjectId="${PROJECT_ID}" \
+            --network redis  \
+            --subnetwork regions/europe-west2/subnetworks/redis \
```
MebinAbraham, line 32: Happy with this once it is signed off from Mark. Worth mentioning that this `VPC` isn't tied to Redis, it was just named after it as that is the only thing that would use it, but could be renamed to like `eq`.

Also, according to the docs, you don't need the `network` when the `subnetwork` is provided.

https://cloud.google.com/dataflow/docs/guides/specifying-networks
> If you specify a subnetwork, Dataflow chooses the network for you. Therefore, when specifying a subnetwork, you can omit the network parameter.

MebinAbraham:
Will wait for conflicts to be resolved as with Lewis' change, we are now using a different template.

MebinAbraham (approved):

petechd (approved):
Tested deploy/destroy with these changes few times and it works as expected

PR #719, opened 2021-11-17
Add type hints to the router

petechd:
- Added type hints to the `app/router` module.

### How to review 

Ensure:
- the type hints are valid
- app works as expected

### Checklist

### DIFF
--- a/app/questionnaire/router.py
+++ b/app/questionnaire/router.py
@@ -1,22 +1,27 @@
-from typing import Optional, Tuple
+from datetime import date
+from typing import Generator, List, Mapping, Optional, Sequence, Tuple, Union
 
 from flask import url_for
+from werkzeug.datastructures import ImmutableDict
 
+from app.data_models import AnswerStore, ListStore, ProgressStore
+from app.questionnaire import QuestionnaireSchema
 from app.questionnaire.location import Location
 from app.questionnaire.path_finder import PathFinder
 from app.questionnaire.routing.when_rule_evaluator import WhenRuleEvaluator
+from app.questionnaire.routing_path import RoutingPath
 from app.questionnaire.rules import evaluate_when_rules
 
 
 class Router:
     def __init__(
         self,
-        schema,
-        answer_store,
-        list_store,
-        progress_store,
-        metadata,
-        response_metadata,
+        schema: QuestionnaireSchema,
+        answer_store: AnswerStore,
+        list_store: ListStore,
+        progress_store: ProgressStore,
+        metadata: Mapping[str, Union[str, int, list]],
+        response_metadata: Mapping,
     ):
         self._schema = schema
         self._answer_store = answer_store
@@ -35,7 +40,7 @@ class Router:
         )
 
     @property
-    def enabled_section_ids(self):
+    def enabled_section_ids(self) -> List:
         return [
             section["id"]
             for section in self._schema.get_sections()
@@ -64,10 +69,14 @@ class Router:
         routing_path = self.routing_path(*self._get_last_complete_section_key())
         return self.get_last_location_in_section(routing_path).url()
 
-    def is_list_item_in_list_store(self, list_item_id, list_name):
+    def is_list_item_in_list_store(
+        self, list_item_id: str, list_name: Optional[str]
+    ) -> bool:
         return list_item_id in self._list_store[list_name]
 
-    def can_access_location(self, location: Location, routing_path):
+    def can_access_location(
+        self, location: Location, routing_path: RoutingPath
+    ) -> bool:
         """
         Checks whether the location is valid and accessible.
         :return: boolean
@@ -82,22 +91,26 @@ class Router:
 
         return location.block_id in self._get_allowable_path(routing_path)
 
-    def can_access_hub(self):
+    def can_access_hub(self) -> bool:
         return self._schema.is_flow_hub and all(
             self._progress_store.is_section_complete(section_id)
             for section_id in self._schema.get_section_ids_required_for_hub()
             if section_id in self.enabled_section_ids
         )
 
-    def can_display_section_summary(self, section_id, list_item_id=None):
+    def can_display_section_summary(
+        self, section_id: str, list_item_id: str = None
+    ) -> Union[ImmutableDict, None, bool]:
         return self._schema.get_summary_for_section(
             section_id
         ) and self._progress_store.is_section_complete(section_id, list_item_id)
 
-    def routing_path(self, section_id, list_item_id=None):
+    def routing_path(self, section_id: str, list_item_id: str = None) -> RoutingPath:
         return self._path_finder.routing_path(section_id, list_item_id)
 
-    def get_next_location_url(self, location, routing_path, return_to=None):
+    def get_next_location_url(
+        self, location: Location, routing_path: RoutingPath, return_to: str = None
+    ) -> str:
         """
         Get the next location in the section. If the section is complete, determine where to go next,
         whether it be a summary, the hub or the next incomplete location.
@@ -121,13 +134,17 @@ class Router:
 
         return self.get_next_block_url(location, routing_path)
 
-    def _get_next_location_url_for_last_block_in_section(self, location):
+    def _get_next_location_url_for_last_block_in_section(
+        self, location: Location
+    ) -> str:
         if self._schema.show_summary_on_completion_for_section(location.section_id):
             return self._get_section_url(location)
 
         return self.get_next_location_url_for_end_of_section()
 
-    def get_previous_location_url(self, location, routing_path):
+    def get_previous_location_url(
+        self, location: Location, routing_path: RoutingPath
+    ) -> Optional[str]:
         """
         Returns the previous 'location' to visit given a set of user answers
         """
@@ -136,7 +153,7 @@ class Router:
         if block_id_index != 0:
             previous_block_id = routing_path[block_id_index - 1]
             previous_block = self._schema.get_block(previous_block_id)
-            if previous_block["type"] == "RelationshipCollector":
+            if previous_block["type"] == "RelationshipCollector":  # type: ignore
                 return url_for(
                     "questionnaire.relationships",
                     last=True,
@@ -162,7 +179,7 @@ class Router:
 
         return self.get_first_incomplete_location_in_questionnaire_url()
 
-    def get_section_resume_url(self, routing_path):
+    def get_section_resume_url(self, routing_path: RoutingPath) -> str:
         section_key = (routing_path.section_id, routing_path.list_item_id)
 
         if section_key in self._progress_store:
@@ -172,11 +189,11 @@ class Router:
 
         return self.get_first_location_in_section(routing_path).url()
 
-    def is_path_complete(self, routing_path):
+    def is_path_complete(self, routing_path: RoutingPath) -> bool:
         return not bool(self._get_first_incomplete_location_in_section(routing_path))
 
     @staticmethod
-    def get_first_location_in_section(routing_path) -> Location:
+    def get_first_location_in_section(routing_path: RoutingPath) -> Location:
         return Location(
             block_id=routing_path[0],
             section_id=routing_path.section_id,
@@ -185,7 +202,7 @@ class Router:
         )
 
     @staticmethod
-    def get_last_location_in_section(routing_path) -> Location:
+    def get_last_location_in_section(routing_path: RoutingPath) -> Location:
         return Location(
             block_id=routing_path[-1],
             section_id=routing_path.section_id,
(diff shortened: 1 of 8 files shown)


app/questionnaire/router.py
```diff
@@ -35,7 +40,7 @@ def __init__(
         )
 
     @property
-    def enabled_section_ids(self):
+    def enabled_section_ids(self) -> List:
```
MebinAbraham, line 43: Should be more specific for container types, also we should use the built-in `list` over `typing.List`.

app/questionnaire/router.py
```diff
@@ -64,10 +69,14 @@ def get_last_location_in_questionnaire_url(self) -> str:
         routing_path = self.routing_path(*self._get_last_complete_section_key())
         return self.get_last_location_in_section(routing_path).url()
 
-    def is_list_item_in_list_store(self, list_item_id, list_name):
+    def is_list_item_in_list_store(
+        self, list_item_id: str, list_name: Optional[str]
+    ) -> bool:
         return list_item_id in self._list_store[list_name]
```
MebinAbraham, line 75: If we didn't have a list name, this would error?

app/questionnaire/router.py
```diff
@@ -82,22 +91,26 @@ def can_access_location(self, location: Location, routing_path):
...
             if section_id in self.enabled_section_ids
         )
 
-    def can_display_section_summary(self, section_id, list_item_id=None):
+    def can_display_section_summary(
+        self, section_id: str, list_item_id: str = None
+    ) -> Union[ImmutableDict, None, bool]:
         return self._schema.get_summary_for_section(
             section_id
         ) and self._progress_store.is_section_complete(section_id, list_item_id)
```
MebinAbraham, line 106: This should always return a bool?

app/questionnaire/router.py
```diff
@@ -82,22 +91,26 @@ def can_access_location(self, location: Location, routing_path):
...
 
-    def routing_path(self, section_id, list_item_id=None):
+    def routing_path(self, section_id: str, list_item_id: str = None) -> RoutingPath:
```
MebinAbraham, line 108: We should use explicit optional, I think there is an implict-optional flag we could disable. So if it is `None` mark it as `Optional[str]` etc.

app/questionnaire/router.py
```diff
@@ -82,22 +91,26 @@ def can_access_location(self, location: Location, routing_path):
...
         return self._path_finder.routing_path(section_id, list_item_id)
 
-    def get_next_location_url(self, location, routing_path, return_to=None):
+    def get_next_location_url(
+        self, location: Location, routing_path: RoutingPath, return_to: str = None
```
MebinAbraham, line 112: Same here for `return_to`, we should use `OptionL[str]`.

app/questionnaire/router.py
```diff
@@ -185,15 +202,15 @@ def get_first_location_in_section(routing_path) -> Location:
...
+    def get_last_location_in_section(routing_path: RoutingPath) -> Location:
         return Location(
             block_id=routing_path[-1],
             section_id=routing_path.section_id,
             list_name=routing_path.list_name,
             list_item_id=routing_path.list_item_id,
         )
 
-    def full_routing_path(self):
+    def full_routing_path(self) -> List[RoutingPath]:
```
MebinAbraham, line 213:
```suggestion
    def full_routing_path(self) -> list[RoutingPath]:
```

app/questionnaire/router.py
```diff
@@ -228,7 +249,7 @@ def _get_first_incomplete_location_in_section(self, routing_path):
                     list_name=routing_path.list_name,
                 )
 
-    def _get_allowable_path(self, routing_path):
+    def _get_allowable_path(self, routing_path: RoutingPath) -> List:
```
MebinAbraham, line 252: We should use `list` and we should return the types.

app/questionnaire/router.py
```diff
@@ -245,7 +266,7 @@ def _get_allowable_path(self, routing_path):
 
         return allowable_path
 
-    def get_enabled_section_keys(self):
+    def get_enabled_section_keys(self) -> Generator:
```
MebinAbraham, line 269: We should specify the yield type. Should be tuple of two strings.

app/questionnaire/router.py
```diff
@@ -257,7 +278,7 @@ def get_enabled_section_keys(self):
                 section_key = (section_id, None)
                 yield section_key
 
-    def _get_first_incomplete_section_key(self):
+    def _get_first_incomplete_section_key(self) -> Sequence:
         for section_id, list_item_id in self.get_enabled_section_keys():
             if not self._progress_store.is_section_complete(section_id, list_item_id):
                 return section_id, list_item_id
```
MebinAbraham, line 284: This returns an optional tuple of strings.

app/questionnaire/router.py
```diff
@@ -267,7 +288,7 @@ def _get_last_complete_section_key(self) -> Tuple[str, Optional[str]]:
             if self._progress_store.is_section_complete(section_id, list_item_id):
                 return section_id, list_item_id
```
MebinAbraham, line 290: We should replace the type for this function from `Tuple` to `tuple`. We should use built-in types over types from mypy.

app/questionnaire/router.py
```diff
@@ -267,7 +288,7 @@ def _get_last_complete_section_key(self) -> Tuple[str, Optional[str]]:
...
 
-    def _is_section_enabled(self, section):
+    def _is_section_enabled(self, section: Mapping) -> Union[bool, date, None]:
```
MebinAbraham, line 291: `date`?. This should always return a bool?.

MebinAbraham (commented):
Some issues are hard to spot until we type the files that depend on Router, so will prob have some issues to tackle when that happens.

app/questionnaire/router.py
petechd: Changed to `list`, thanks!

app/questionnaire/router.py
petechd: `Optional[str] ` now, thanks!

app/questionnaire/router.py
petechd: Added, thanks!

app/questionnaire/router.py
petechd: Added return type and `list` thanks!

app/questionnaire/router.py
petechd: The first yeld returns tuple of [Any, Any] the second tuple of [Any, None] so I left it without type of tuple contents, so `Generator[tuple, None, None]` for now..

app/questionnaire/router.py
petechd: Done, thanks!

app/questionnaire/router.py
petechd: Done, thanks!

app/questionnaire/router.py
petechd: This one depends on `WhenRuleEvaluator` methods which don't return `bool` hence these types in router. I'll try to fix'em in evaluator.

app/questionnaire/router.py
petechd: This one is Optional because `list_name' is an optional attribute of Location class (which we use in here) so it's not always there potentially🤦‍♂️.

`list_name: Optional[str] = None`

app/questionnaire/router.py
petechd: It doesn't always return `bool` because we don't have a fallback option in here:
Screenshot 2021-11-18 at 09 03 39
so it returns `None` sometimes.

petechd: it would work for router if we did something like this:

Screenshot 2021-11-18 at 09 16 07

... but that would break `questionnaire_schema.py` types in many places.

app/questionnaire/router.py
MebinAbraham: The type the container returns needs to be specified, so we know what sort of list we get.

app/questionnaire/router.py
MebinAbraham: It just needs to be `str` for `list_name` and the before this is called, it needs to check you have a `list_name`.

app/questionnaire/router.py
MebinAbraham: This is because we are simply using the existence of that dict object to see if we can access the summary. So you just need to cast it to a `bool`. You can either cast to bool, or check the condition:

```py
    def can_display_section_summary(
        self, section_id: str, list_item_id: str = None
    ) -> bool:
        return bool(
            self._schema.get_summary_for_section(section_id)
        ) and self._progress_store.is_section_complete(section_id, list_item_id)
```

or

```py
    def can_display_section_summary(
        self, section_id: str, list_item_id: str = None
    ) -> bool:
        return self._schema.get_summary_for_section(
            section_id
        ) is not None and self._progress_store.is_section_complete(
            section_id, list_item_id
        )
```

app/questionnaire/router.py
MebinAbraham: This wouldn't be `tuple[Any, str]` but `tuple[str, str]` once https://github.com/ONSdigital/eq-questionnaire-runner/pull/719#discussion_r753114992 is fixed.

app/questionnaire/router.py
MebinAbraham: Ya this is a consequence of the evaluator supporting boolean and value operators, but I think we should make code changes so that the typing is explicit otherwise we are just making it harder for ourselves. We know it should only be a `bool` if the schema is valid.

So 2 options, either check if is an instance of a bool, and raise exception if not, or cast to a bool which will always give true or false (it shouldn't ever be not a bool) but this will help us have strict typing. Casting is easier and doesn't require any tests or ignore line coverage.

app/questionnaire/router.py
petechd: So it's not always `tuple[str, str]` as the Generator returns a tuple without `list_item_id` sometimes so I had to make it `Generator[Union[tuple[str, None], tuple[str, str]], None, None]` and in other places that call this method: `tuple[str, Optional[str]]`. Thanks!

app/questionnaire/router.py
petechd: I used the first approach, thanks!

app/questionnaire/router.py
petechd: Added, thanks!

app/questionnaire/router.py
petechd: Added `str` type for `list_name` and extra conditional in parent method to check if `list_name` exists which lead to coverage error, so had to introduce extra test in `test_router.py`. Thanks!

app/questionnaire/router.py
MebinAbraham: That's fine, it would be clearer to use `Optional` so `Generator[tuple[str, Optiona[str]], None, None]`

petechd: Changed to `Generator[tuple[str, Optiona[str]], None, None]`, thanks!

app/questionnaire/router.py
petechd: I've added casting to bool in `_is_section_enabled` as other approach would would mean refactoring old `when rule evaluator` and don't we want to get rid of it soon anyway? Thanks!

app/questionnaire/router.py
```diff
@@ -64,40 +67,55 @@ def get_last_location_in_questionnaire_url(self) -> str:
...
             self._progress_store.is_section_complete(section_id)
             for section_id in self._schema.get_section_ids_required_for_hub()
             if section_id in self.enabled_section_ids
         )
 
-    def can_display_section_summary(self, section_id, list_item_id=None):
-        return self._schema.get_summary_for_section(
-            section_id
+    def can_display_section_summary(
+        self, section_id: str, list_item_id: str = None
```
MebinAbraham, line 102:
```suggestion
        self, section_id: str, list_item_id: Optional[str] = None
```

MebinAbraham (approved):

app/questionnaire/router.py
```diff
@@ -136,7 +158,7 @@ def get_previous_location_url(self, location, routing_path):
         if block_id_index != 0:
             previous_block_id = routing_path[block_id_index - 1]
             previous_block = self._schema.get_block(previous_block_id)
-            if previous_block["type"] == "RelationshipCollector":
+            if previous_block["type"] == "RelationshipCollector":  # type: ignore
```
MebinAbraham, line 161: This doesn't need to be ignored, I guess it is complaining you could get a `None` if so it should be:
```py
if previous_block and previous_block["type"] == "RelationshipCollector":
```

LJBabbage (approved):

app/questionnaire/router.py
petechd: Added, thanks!

PR #720, opened 2021-11-19
Refactor operations.py into class

LJBabbage:
Make operations into a class, and move instantiated higher. The instantiation has been moved higher to allow access to schema etc, but at present that is not implemented and on a different card

### How to review 
Make sure everything still works

### Checklist

### DIFF
--- a/app/questionnaire/routing/when_rule_evaluator.py
+++ b/app/questionnaire/routing/when_rule_evaluator.py
@@ -12,25 +12,6 @@ from app.questionnaire.value_source_resolver import (
     ValueSourceTypes,
 )
 
-operations = Operations()
-
-OPERATIONS_MAPPINGS: dict[str, Callable] = {
-    Operator.NOT: operations.evaluate_not,
-    Operator.AND: operations.evaluate_and,
-    Operator.OR: operations.evaluate_or,
-    Operator.EQUAL: operations.evaluate_equal,
-    Operator.NOT_EQUAL: operations.evaluate_not_equal,
-    Operator.GREATER_THAN: operations.evaluate_greater_than,
-    Operator.LESS_THAN: operations.evaluate_less_than,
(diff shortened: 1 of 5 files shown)


tests/app/questionnaire/routing/test_operators.py
```diff
@@ -343,5 +369,27 @@ def test_nonetype_operands_for_array_operators(operator_name, operands):
     ],
 )
 def test_operation_count(operands, expected_result):
-    operator = Operator(Operator.COUNT)
+    operation = get_operation(Operator.COUNT)
+    operator = Operator(Operator.COUNT, operation)
     assert operator.evaluate(operands) is expected_result
+
+
+def get_operation(operator):
```
LJBabbage, line 377: This is a pain, but as operations will eventually be initialized with schema, locale etc and the function names don't match up I have had to put a map here, however I realise this is duplicated in when_rule_evaluator.

app/questionnaire/routing/when_rule_evaluator.py
```diff
@@ -36,12 +37,30 @@ def __post_init__(self) -> None:
...
 
     def _evaluate(self, rule: dict[str, Sequence]) -> Union[bool, Optional[date]]:
-        operator = Operator(next(iter(rule)))
-        operands = rule[operator.name]
+        next_rule = next(iter(rule))
+        operator = Operator(next_rule, self.operation_mapping[next_rule])
+        operands = rule[next_rule]
 
-        if not isinstance(operands, Sequence):
+        if not isinstance(rule[next_rule], Sequence):
```
petechd, line 63: I think you forgot about using previously assigned `operands` here?

LJBabbage: Oh yeah, I changed it, then realised it was used elsewhere, but must of missed changing it back, thanks

berroar (approved):
LGTM beyond the comments above 👍

petechd (approved):
Looks good!

PR #725, opened 2021-11-24
Add support for day_of_week date offset

MebinAbraham:
Added support for the `day_of_week` date offset. Follows the same pattern as the existing `date_range_bounds` placeholder transform.

### How to review 
Ensure the changes are appropriate.

### Checklist


app/questionnaire/placeholder_transforms.py
```diff
@@ -223,29 +206,6 @@ def format_date_range(self, date_range: tuple[str, str]) -> str:
 
         return f"{start_date_formatted} to {end_date_formatted}"
 
-    @staticmethod
-    def parse_date(date: str) -> datetime:
```
MebinAbraham, line 227: Moved this to a top level function in the utils file. Could be a static method on operation but type hints and overloading does not work with static methods. https://github.com/python/mypy/issues/7781

app/questionnaire/rules.py
```diff
@@ -138,17 +138,7 @@ def get_date_match_value(date_comparison, answer_store, schema, metadata):
...
-
-    if re.match(r"\d{4}-\d{2}-\d{2}", value):
-        date_format = "%Y-%m-%d"
-    elif re.match(r"\d{4}$", value):
-        date_format = "%Y"
-    else:
-        date_format = "%Y-%m"
-
-    return datetime.strptime(value, date_format).replace(tzinfo=timezone.utc)
+    return parse_datetime(value)
```
MebinAbraham, line 141: Didn't update calls to `parse_datetime` to keep changes to a minimum, but happy to remove this function.

app/questionnaire/routing/operations.py
```diff
@@ -81,18 +98,39 @@ def evaluate_any_in(lhs: Sequence, rhs: Sequence) -> bool:
 
     @staticmethod
     def resolve_date_from_string(
-        date_string: Optional[str], offset: Optional[dict[str, int]] = None
+        date_string: Optional[str],
+        offset: Optional[DateOffset] = None,
+        offset_by_full_weeks: bool = False,
```
MebinAbraham, line 103: This flag is needed cause the placeholder date bound works slightly differently, as it always uses a full week offset, so it can do things like "Last week Thursday" if the reference date is less than 1 week ago. This new function should allow us to do everything we need to do.

app/questionnaire/routing/operations.py
```diff
@@ -81,18 +98,39 @@ def evaluate_any_in(lhs: Sequence, rhs: Sequence) -> bool:
...
+                    raise ValueError(
+                        "Negative days offset must be less than or equal to -7 when used with `day_of_week` offset"
+                    )
+
+                days_difference = (
+                    value_as_date.weekday() - DAYS_OF_WEEK[day_of_week_offset]
+                )
+                days_to_reduce = days_difference % 7
+
+                if not offset_by_full_weeks and days_offset < 0 and days_difference < 0:
```
LJBabbage, line 125: This sentence is too complicate for me, we also shouldn't be adjusting the date, we should be changing the days_offset and then taking that off the date once it is right.  
```
                if not offset_by_full_weeks and days_offset < 0 and days_difference < 0:
                    # If we have a negative offset and date difference is less than 6 days,
                    # then the days we reduce the offset must be adjusted by 7
                    days_to_reduce -= 7
                days_offset -= days_to_reduce
```
I still haven't got the comment right

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -314,6 +315,64 @@ def test_operation_date(date_string: str, offset):
     assert operator.evaluate(operands) == expected_result
 
 
+@pytest.mark.parametrize(
+    "offset, expected_result",
+    [
+        # Last Thursday (Yesterday from reference date)
```
LJBabbage, line 321: I find these comments confusing. I think we need to rationalise reference date and Initial date (used in doc string) and move it to the top, because as soon as I see the first comment "Yesterday from reference date", I am looking for what it is. Yesterday is also a funny term used in this context, it should be "day before reference date". As a soon as you say yesterday I am thinking today.

app/questionnaire/routing/operations.py
MebinAbraham: Ya, had it like this initially but changed it to avoid the double negative. Don't mind changing it back if it is not confusing. 

As for the comment, ya, it is a tricky one to get your head around. Maybe an example would help if it was inline with the comments?. Ill come back to the comments once more have looked at it.

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -314,6 +315,64 @@ def test_operation_date(date_string: str, offset):
...
+            datetime(year=2020, month=12, day=26),
+        ),
+        # Next Thursday / Next Week Thursday
+        ({"day_of_week": "THURSDAY", "days": 7}, datetime(year=2021, month=1, day=7)),
+        # Next Saturday (Tomorrow from reference date)
+        ({"day_of_week": "SATURDAY", "days": 7}, datetime(year=2021, month=1, day=2)),
+        # Next Week Saturday
+        ({"day_of_week": "SATURDAY", "days": 14}, datetime(year=2021, month=1, day=9)),
+        # 1 full week Thursday
+        ({"day_of_week": "THURSDAY", "days": 14}, datetime(year=2021, month=1, day=14)),
```
LJBabbage, line 342: Do we need all these combination? I find this a bit strange "Next Thursday / Next Week Thursday" perhaps just one, because reference date is Friday that makes it New week Thursday I guess

app/questionnaire/routing/operations.py
MebinAbraham: I've redone the comment. Let me know if it is clearer, and what else we may want to add to it.

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -314,6 +315,64 @@ def test_operation_date(date_string: str, offset):
...
+            {"day_of_week": "SATURDAY", "days": -7},
+            datetime(year=2020, month=12, day=26),
+        ),
+        # Next Thursday / Next Week Thursday
+        ({"day_of_week": "THURSDAY", "days": 7}, datetime(year=2021, month=1, day=7)),
+        # Next Saturday (Tomorrow from reference date)
+        ({"day_of_week": "SATURDAY", "days": 7}, datetime(year=2021, month=1, day=2)),
+        # Next Week Saturday
+        ({"day_of_week": "SATURDAY", "days": 14}, datetime(year=2021, month=1, day=9)),
+        # 1 full week Thursday
```
LJBabbage, line 341: 1 full week from Thursday but an offset of 14 days, unfortunately I also find this confusing

tests/app/questionnaire/routing/test_operators.py
```diff
@@ -314,6 +315,64 @@ def test_operation_date(date_string: str, offset):
...
+        ({"day_of_week": "THURSDAY", "days": 14}, datetime(year=2021, month=1, day=14)),
+    ],
+)
+def test_operation_date_offsets_literal_date(offset, expected_result):
+    """
+    This is to test a literal date instead of dynamically generating the expected date.
```
LJBabbage, line 347: The doc strings are a bit inconsistent, some have them some do not. I don't think we need this one, the name tells me what I need to know

app/questionnaire/routing/operations.py
LJBabbage: I actually understand what it is doing, but it is difficult to write a comment, I failed, so can't complain whatever you write

LJBabbage: Double negatives are fine, coming from someone with a Maths degree

tests/app/questionnaire/routing/test_operators.py
MebinAbraham: `Next Thursday / Next Week Thursday` added both to make it clearer, they are not different for this test case where we are on a Friday.

tests/app/questionnaire/routing/test_operators.py
MebinAbraham: Ya, that is a consequence of us always get the "past" date first. We always resolve the last "day of the week", so we will be 1 week behind. No easy solution to this, whatever we do, this will exist in some form. If we didn't resolve to the past day always that brings its own complications also (initial did that).

LJBabbage (approved):
It works, although I still find it a bit difficult to comprehend, and wonder how easy this will be to implement and explain in Author. However this is not the fault of this PR and does what it is supposed to

petechd (approved):

app/questionnaire/routing/operations.py
jamesnunn: Ugh I can't comment on this line and make it visible unless I unresolve the conversation - but it's nothing to do with the convo.

Could you just help out a future person reading this by parenthesising the terms - I'm pretty sure logic is left associative but I always find myself having to double check when maintaining, especially when mixing negatives and ands/ors

app/questionnaire/rules.py
jamesnunn: Do you mean calls to `convert_to_datetime`? I think if it's a simple refactor it may add noise but it's easy to see what's happened

MebinAbraham: Ya, it was just to reduce noise, the refactor **should** be a drop-in replacement. 🤞

app/questionnaire/routing/operations.py
MebinAbraham: Do you mean line 125 @jamesnunn?. I didn't add parentheses, since it would short circuit + it is ordered based on precedence. Happy to change if you think it will help:
```py
                if not offset_by_full_weeks and (
                    days_offset < 0 and days_difference < 0
                ):
```
Happy with this? ☝️

app/questionnaire/rules.py
MebinAbraham: @LJBabbage @petechd Since you already ticked and this was left here beforehand, happy for me to change?

app/questionnaire/routing/operations.py
jamesnunn: Yeah that's what I'm after. How does it short circuit though?

app/questionnaire/rules.py
petechd: @MebinAbraham yeah happy with that!

LJBabbage: Yep

app/questionnaire/routing/operations.py
MebinAbraham: `and` in python short circuits, so if `not offset_by_full_weeks` is `True`, `days_offset < 0 and days_difference < 0` wouldn't even be evaluated. Is that what you were asking or?

app/questionnaire/rules.py
MebinAbraham: Done in [`d1f2892` (#725)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/725/commits/d1f28929e0cdf1cbdc52f334475871b288adfcde)
@jamesnunn @petechd @LJBabbage

app/questionnaire/routing/operations.py
jamesnunn: I'm not sure that's a problem because they're all `and`s and you'd be testing the result of that - it either gets to that block or it doesn't - all of the conditions are eventually evaluated. Anyway, it's fine in my opinion as it is!

MebinAbraham: Not sure I follow. `all of the conditions are eventually evaluated`, only if leading conditions are True?. Anyhow, I decided to commit it as I don't mind, and it doesn't make it any less clearer. Plus, I already committed it for my other branch already 😄

jamesnunn: Yeah no worries we're splitting hairs again on this one. if you're happy with it then I am too

tests/app/questionnaire/routing/test_utils.py
```diff
@@ -0,0 +1,32 @@
...
+@freeze_time(datetime.now(timezone.utc))
+def test_parse_datetime_now():
+    assert parse_datetime("now") == datetime.now(timezone.utc)
+
+
+def test_parse_date_exception():
+    with pytest.raises(ValueError):
+        parse_datetime("2021--10")
+        parse_datetime("2021-10-229")
+        parse_datetime("2021-10-29T10:53:41.511833+00:0000")
```
jamesnunn, line 32: I think this doesn't ever reach lines 31+32 unless `pytest.raises` does something magic that I don't know about

tests/app/questionnaire/routing/test_utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+from app.questionnaire.routing.utils import parse_datetime
+
+
+@pytest.mark.parametrize(
+    "date_string, date_format",
+    [
+        ("2021-10", "%Y-%m"),
+        ("2021-10-29", "%Y-%m-%d"),
+        ("2021-10-29T10:53:41.511833+00:00", "%Y-%m-%dT%H:%M:%S.%f%z"),
```
jamesnunn, line 14: Maybe add the Zulu variant for testing?

tests/app/questionnaire/routing/test_utils.py
MebinAbraham: No, you're right, it doesn't. Just copied the existing test but didn't clock it. Have fixed.

tests/app/questionnaire/routing/test_utils.py
```diff
@@ -25,8 +25,16 @@ def test_parse_datetime_now():
...
 
 
-def test_parse_date_exception():
+@pytest.mark.parametrize(
+    "date_string",
+    [
+        "2021--10",
+        "2021-10-229",
+        "2021-10-29T10:53:41.511833+00:0000",
+        "2021-10-29T10:53:41.511833+00:0000",
```
jamesnunn, line 34: Got a dupe here, harmless though

tests/app/questionnaire/routing/test_utils.py
MebinAbraham: The current version of `parse_datetime` doesn't support Zulu or ISO without milliseconds. I have changed the implementation in https://github.com/ONSdigital/eq-questionnaire-runner/pull/, so will address there.

jamesnunn (approved):

PR #728, opened 2021-11-30
Add map operator

MebinAbraham:
Added support for the `map` operator which will be used by dynamic options for both its `values` and `transform`.

Note: nested map operators are not supported.

### How to review 
Ensure the changes make sense.

Use the individual commits to your advantage. The main changes are in the first commit, the rest are just naming refactors.

### Checklist


app/questionnaire/rules/utils.py
```diff
@@ -1,6 +1,12 @@
 from datetime import datetime, timezone
 from typing import Optional, overload
 
+ISO_8601_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
+
+
+def parse_iso_8601_datetime(iso_8601_datetime: str) -> datetime:
+    return datetime.fromisoformat(iso_8601_datetime).replace(tzinfo=timezone.utc)
```
petechd, line 8: Don't we want to use `dateutil.parser.isoparse` (from `dateutil` package) here which is a more fully featured ISO-8601 format parser and would give us some more flexibility in the future?

jamesnunn (approved):
Hard for me to traverse the context of the code so just reviewed the code at face value, and the tests

app/questionnaire/rules/rule_evaluator.py
```diff
@@ -42,30 +42,43 @@ def __post_init__(self) -> None:
...
         if not isinstance(operands, Sequence):
             raise TypeError(
                 f"The rule is invalid, operands should be of type Sequence and not {type(operands)}"
             )
 
-        resolved_operands = self.get_resolved_operands(operands)
+        resolved_operands: Iterable[Union[bool, Optional[date], ValueSourceTypes]]
+
+        if operator_name == Operator.MAP:
+            resolved_iterables = self._resolved_operand(operands[1])
```
parimooa, line 57: `resolved_iterables ` can be `bool, Optional[date], ValueSourceTypes` , how come it is iterable ? consider renaming

parimooa (commented):
couple of minor comments , otherwise looks good

app/questionnaire/rules/operations.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class Operations:
+    """
+    A class to group the operations
+    """
+
+    def __init__(self, language: str):
+        self._language = language
+        self._locale = DEFAULT_LOCALE if language in ["en", "eo"] else language
```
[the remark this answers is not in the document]
MebinAbraham: This is something we should make consistent, currently it is not. List is used when they don't need to be, mainly because it is negligible. I have changed since it is a better approach to use set. Have changed the one in placeholder transforms also.

petechd (approved):

berroar (approved):

parimooa (approved):

PR #729, opened 2021-12-01
Updated skip condition to new rules

parimooa:
Update skip condition for `List` as value source
update condition to check for old and new `skip` rules for `List`
Related [Validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/129)
### How to review 
Run against schema validator , it should pass without any errors

### Checklist


rmccar (dismissed):

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,6 +215,12 @@ def get_post_submission(self) -> ImmutableDict:
         schema: ImmutableDict = self.json.get("post_submission", ImmutableDict({}))
         return schema
 
+    @staticmethod
+    def _check_list_exists(rule: Union[dict, tuple], list_name: str) -> bool:
+        return isinstance(rule, dict) and (
+            rule.get("list") == list_name or rule.get("identifier") == list_name
+        )
```
MebinAbraham, line 222: Which would this be a tuple?. Wouldn't the existing code error if this was a tuple and not a mapping?

parimooa: as discussed yesterday, you were right, it wasn't picking up nested rule. I have fixed it now.

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,19 +215,40 @@ def get_post_submission(self) -> ImmutableDict:
...
+                rule.get("list") == list_name and "list" in rule for rule in rules
+            )
+
+        # New rules
+        if (
+            isinstance(rules, dict)
+            and rules.get("source") == "list"
+            and rules.get("identifier") == list_name
+        ):
+            return True
```
MebinAbraham, line 231: Once we know the source is a `list`, we need to return a bool, otherwise it would error as we are at the final dict and list name didn't match for any number of reasons, such as it is a rule for a diff list name, when this happens, it would currently try to continue iterating and would then fall over.

The test doesn't pick this up because one rule is using the new format, the other rule is using the old format. We should extend it, so you have 4 blocks, 2 that matches the list name and 2 that doesn't. You would use old format for 2 blocks, and new format for the other 2.

```suggestion
        if isinstance(rules, dict) and rules.get("source") == "list":
            return rules["identifier"] == list_name
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,19 +215,36 @@ def get_post_submission(self) -> ImmutableDict:
         schema: ImmutableDict = self.json.get("post_submission", ImmutableDict({}))
         return schema
 
+    def _check_list_exists(self, rules: Union[dict, Sequence], list_name: str) -> bool:
```
MebinAbraham, line 218: I think we should rename it to something like, `_is_list_name_in_rule`?. `_check_list_exists` to me suggests tell me if list exists in the schema.

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,19 +215,36 @@ def get_post_submission(self) -> ImmutableDict:
...
+        # old rules
+        if isinstance(rules, Sequence):
+            return any(
+                rule.get("list") == list_name and "list" in rule for rule in rules
+            )
+
+        # New rules
+        if isinstance(rules, dict) and rules.get("source") == "list":
+            return rules.get("identifier") == list_name
```
MebinAbraham, line 229: This would actually still fail, as if source is not list, we still don't need to iterate any further. Should be something like below but may need rework based on the comment below.

```py
        if isinstance(rules, dict) and (source := rules.get("source")):
            return source == "list" and rules.get("identifier") == list_name
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,19 +215,36 @@ def get_post_submission(self) -> ImmutableDict:
...
+
+        # for nested new rules
+        if isinstance(rules, dict):
+            operator = next(iter(rules))
+            operand = rules[operator]
+            return self._check_list_exists(operand[0], list_name)
```
MebinAbraham: - I think this needs some more rework, I think it will be easier to use the logic used in the rules' evaluator on how it does recursion. This now assumes all dicts are rules, but we can have dicts that are not rules, such as date offsets. Similar to rules evaluator, we want to know it is a dict and a valid rule before doing recursion.
- Also that last line is only checking the first operand `operand[0]`, it needs to check all operands. Right now, if you swap the test operands around for the one where you expect to find a list it would fail because it isn't checking the 2nd operand which has the list rule.

MebinAbraham (changes requested):

app/questionnaire/questionnaire_schema.py
parimooa: Updated this method to handle all nested cases (hopefully :-) )

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,20 +216,52 @@ def get_post_submission(self) -> ImmutableDict:
...
+                if isinstance(rule, dict) and any(
+                    operator in rule for operator in OPERATION_MAPPING
+                ):
+                    return self._is_list_name_in_rule(rule, list_name)
+
+    @staticmethod
+    def get_operand(rules: dict) -> list:
+        operator = next(iter(rules))
+        operand: list = rules[operator]
+        return operand
```
MebinAbraham, line 253: Minor but could be plural since most have more than 1 operand.

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,20 +216,52 @@ def get_post_submission(self) -> ImmutableDict:
...
+                    return (
+                        rule.get("source") == "list"
+                        and rule.get("identifier") == list_name
+                    )
+
+                # Nested rules
+                if isinstance(rule, dict) and any(
+                    operator in rule for operator in OPERATION_MAPPING
+                ):
+                    return self._is_list_name_in_rule(rule, list_name)
```
MebinAbraham, line 247: I don't think this is instance sequence check is needed?. If it is the old rules, you will have a list of rules, or if it is the new rules, then the operands would be a list as well?.

Also, could continue early and avoid the multiple isisntance check.

```suggestion
        for rule in rules:
            if not isinstance(rule, dict):
                continue

            # old rules
            if "list" in rule:
                return rule.get("list") == list_name

            # New rules
            if "source" in rule:
                return (
                    rule.get("source") == "list"
                    and rule.get("identifier") == list_name
                )

            # Nested rules
            if any(
                operator in rule for operator in OPERATION_MAPPING
            ):
                return self._is_list_name_in_rule(rule, list_name)
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,20 +216,52 @@ def get_post_submission(self) -> ImmutableDict:
         schema: ImmutableDict = self.json.get("post_submission", ImmutableDict({}))
         return schema
 
+    def _is_list_name_in_rule(
+        self, when_rules: Union[dict, Sequence], list_name: str
+    ) -> bool:
```
MebinAbraham, line 221: I think it is worth dropping the `when`, just `rule` is fine as it can be used outside of a `when`.

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,20 +216,52 @@ def get_post_submission(self) -> ImmutableDict:
         schema: ImmutableDict = self.json.get("post_submission", ImmutableDict({}))
         return schema
 
+    def _is_list_name_in_rule(
+        self, when_rules: Union[dict, Sequence], list_name: str
```
LJBabbage, line 220: Should probably be just rule

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,20 +216,52 @@ def get_post_submission(self) -> ImmutableDict:
...
+    ) -> bool:
+        rules: Union[dict, Sequence]
+        if isinstance(when_rules, dict) and any(
+            operator in when_rules for operator in OPERATION_MAPPING
+        ):
+            rules = self.get_operand(when_rules)
+        else:
+            rules = when_rules
+
+        if isinstance(rules, Sequence):
```
LJBabbage, line 230: When would it be a dict here?

LJBabbage (commented):
Looks good and works as expected

LJBabbage:
Mebin gets the bonus points for quickest finger, spent an age trying to work out if the terms were right in _is_list_name_in_rule to decide I couldn't suggest anything better, and just make 2 comments. In the meantime Mebin has nipped in and already mentioned them 😞

app/questionnaire/questionnaire_schema.py
parimooa: It will be always `sequence` , if it is a `dict` then we work out( before) if it has `operator` in  then we get operator values which could be a rule or another operator with rules as below

```
                 "when": { 
                            ">": [
                                        {
                                            "count": [
                                                {
                                                    "source": "list",
                                                    "identifier": "people"
                                                }
                                            ]
                                        },
                                        1
                                    ]
                                }
                            },
```

app/questionnaire/questionnaire_schema.py
parimooa: updated to `rules_dict` , `rules` is being used many times

app/questionnaire/questionnaire_schema.py
parimooa: https://github.com/ONSdigital/eq-questionnaire-runner/pull/729#discussion_r768867516

LJBabbage (approved):
Looks good, very very minor but the comments on _is_list_name_in_rule are indented wrong, or maybe that was deliberate

app/questionnaire/questionnaire_schema.py
```diff
@@ -215,20 +216,49 @@ def get_post_submission(self) -> ImmutableDict:
...
+            if "list" in rule:
+                return rule.get("list") == list_name
+
+                # New rules
+            if "source" in rule:
+                return (
+                    rule.get("source") == "list" and rule.get("identifier") == list_name
+                )
+
+                # Nested rules
```
MebinAbraham, line 242: Minor, but the indentation is off.
```suggestion
            # Old rules
            if "list" in rule:
                return rule.get("list") == list_name

            # New rules
            if "source" in rule:
                return (
                    rule.get("source") == "list" and rule.get("identifier") == list_name
                )

            # Nested rules
```

MebinAbraham (approved):

PR #730, opened 2021-12-02
Add test benchmark business schema

LJBabbage:
This PR adds the business test schema

### How to review 
There isn't really much to test here, bar launching it and making sure it has a feedback option after submission, as the checks below do the rest. The plan was to take the JSON from Author and do the minimal amount of changes to make it work, validate and add feedback, so we can use it for benchmarking. The simple fact is there are multiple things that  could be changed, ids, spelling, rules etc (but we are not reviewing the JSON as such, just it doesn't blow up ) and to quote Shaggy the content **wasn't me** anyway

### Checklist


rmccar (approved):

petechd (approved):
Schema works as expected, gets to submission and feedback etc.

PR #732, opened 2021-12-09
Ability to provide a file prefix for datastore backup and restore jobs

MebinAbraham:
Added the ability to provide a file prefix when carrying out the backup and restore jobs. Chose the term file prefix over backup prefix as that is what it really is and how it is referenced by Google.

### How to review 
Ensure the changes are appropriate and that they work by fly executing a destroy/restore task or by using eq-pipelines.

### Checklist


ci/backup_questionnaire_state.yaml
```diff
@@ -8,6 +8,7 @@ params:
   SERVICE_ACCOUNT_JSON: ((gcp.service_account_json))
   PROJECT_ID:
   BUCKET_NAME:
+  FILE_PREFIX:
```
MebinAbraham, line 11: Decided not to add a default in the end, as it wouldn't be correct when we come to support multiple survey types.

ci/README.md
```diff
@@ -32,11 +32,13 @@ Questionnaire state can be backed up using the `backup_questionnaire_state.yaml`
 ```sh
 PROJECT_ID=<project_id> \
 BUCKET_NAME=<bucket_name> \
+FILE_PREFIX=<file_prefix> \
```
LJBabbage, line 35: Not sure about this name. If I read the docs it's a folder not a file prefix and it's only used for backups. Wouldn't BACKUP_FOLDER_NAME or similar be more appropriate?

LJBabbage: The card says BACKUP_PREFIX which makes me even more confused

berroar (approved):

LJBabbage (commented):
All works as expected. But I do find the naming of the vars odd. I think the readme kind of highlights this as the description doesn't match the var.
```
BACKUP_NAME  is the timestamped folder name. 
FILE_PREFIX is the top level folder
```
Looking at the vars I am thinking that the file it generates is <FILE_PREFIX>-<BACKUP_NAME> but it's not. BACKUP_NAME and FILE_PREFIX are just folders unless I have missed something.

MebinAbraham:
> All works as expected. But I do find the naming of the vars odd. I think the readme kind of highlights this as the description doesn't match the var.
> 
> ```
> BACKUP_NAME  is the timestamped folder name. 
> FILE_PREFIX is the top level folder
> ```
> 
> Looking at the vars I am thinking that the file it generates is <FILE_PREFIX>-<BACKUP_NAME> but it's not. BACKUP_NAME and FILE_PREFIX are just folders unless I have missed something.

Ya, they are folders. Google called it a file prefix because it could contain more than one name separated by slashed for one name. What are your suggestions?.

Maybe we could make them something like:
```
TOP_LEVEL_FOLDER
BACKUP_FOLDER
```

LJBabbage:
> > All works as expected. But I do find the naming of the vars odd. I think the readme kind of highlights this as the description doesn't match the var.
> > ```
> > BACKUP_NAME  is the timestamped folder name. 
> > FILE_PREFIX is the top level folder
> > ```
> > 

> > Looking at the vars I am thinking that the file it generates is <FILE_PREFIX>-<BACKUP_NAME> but it's not. BACKUP_NAME and FILE_PREFIX are just folders unless I have missed something.
> 
> Ya, they are folders. Google called it a file prefix because it could contain more than one folder based if whether you use slashes in the prefix. What are your suggestions?.
> 
> Maybe we could make them something like:
> 
> ```
> TOP_LEVEL_FOLDER/ROOT_FOLDER
> BACKUP_FOLDER
> ```

 I look at the vars in the pipeline (specially restore as it has both) and thinking outside the context of GCP what intuitively tells me what the value for each should be. For Me the combination BACKUP_FOLDER (or even BACKUP_NAME) and BACKUP_TIMESTAMP (not sure you need the folder here, but you could) would get me where I wanted to go.

pricem14pc:
> Looking at the vars I am thinking that the file it generates is <FILE_PREFIX>-<BACKUP_NAME> but it's not. BACKUP_NAME and FILE_PREFIX are just folders unless I have missed something.

They appear as folders visually but aren't really folders. You can't traverse them as they're just part of the filename.

LJBabbage (approved):
There seems to be a little bit of confusion on this one (mainly from GCP visually calling them and showing them as folders) However fundamental this works as expected and shouldn't be held up

PR #734, opened 2021-12-10
Adds a transform to get the label for an answer value

jamesnunn:
Adds a transform to get the label for an answer value

### How to review 

Single answer (placeholder)
1. Load the the survey `test_placeholder_option_label_from_value.json`
2. Select a single radio answer
3. The first option of radio is a placeholder piped from metadata
3. Check that the question text on the next page contains the label of the question you selected (If you selected `ESSENTIAL SERVICES LTD` the text should contain `ESSENTIAL SERVICES LTD. (piped)`, which is a placeholder for the label in the schema). Submit.
4. Check that the summary page contains the single checkbox answer and the recovery time answer. 

### Checklist


jamesnunn:
[Using answer option labels as placeholders (3 SP)](https://trello.com/c/YBSWn2q1/4867-using-answer-option-labels-as-placeholders-3-sp)

app/questionnaire/placeholder_transforms.py
```diff
@@ -292,3 +298,33 @@ def _create_hyperlink(href: str, link_text: str) -> str:
...
+
+    def option_label_from_value(self, value: str, answer_id: str) -> str:
+
+        answers = self.schema.get_answers_by_answer_id(answer_id)
+        option_label: str = next(
+            (
+                options["label"]
+                for answer in answers
+                for options in answer["options"]
+                if value in options["value"]
```
MebinAbraham, line 310: We should match the value exactly. `in` would mean it could be a partial match.

berroar (commented):
I notice on the card it says `If placeholder doesn't resolve to anything, we will write out an empty string` however I think at the moment this will return the first option regardless, is this intended?

app/questionnaire/placeholder_transforms.py
```diff
@@ -292,3 +298,33 @@ def _create_hyperlink(href: str, link_text: str) -> str:
...
+        if isinstance(option_label, str):
+            label = option_label
+
+        elif isinstance(option_label, dict):
+            parser = placeholder_parser.PlaceholderParser(
+                language=self.locale,
+                answer_store=AnswerStore(),
+                list_store=ListStore(),
+                metadata=ImmutableDict({}),
+                response_metadata={},
```
MebinAbraham, line 324: This won't work because it is not given any of the actual stores. We shouldn't just be initializing the stores here. You will need to have the actual store we create, so placeholders have answer data etc to use when rendering. You want the `PlaceholderRenderer` which will use the parser.

app/questionnaire/placeholder_transforms.py
```diff
@@ -292,3 +298,33 @@ def _create_hyperlink(href: str, link_text: str) -> str:
...
+                schema=self.schema,
+            )
+            placeholder_text = option_label["text"].replace("{", "").replace("}", "")
```
MebinAbraham, line 327: We shouldn't need to do anything like this. I think what you need is `PlaceholderRenderer` which is responsible for getting the rendered text. It should to be initialised at the class level to prevent initialising it multiple times within the same block. It might be cleaner to pass the class an initialised placeholder renderer than pass it all the stores and then initialise itself.

`RuleEvaluator` can create an instance of `PlaceholderRender` and pass it to `Operations`, you will still need to do the same here and have access to `PlaceholderRender` from `PlaceholderTransforms`

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -378,3 +391,18 @@ def test_date_range_bounds_kwarg_default_monday(placeholder_transform):
...
+
+
+@pytest.mark.parametrize(
+    "answer_id,value,expected",
+    [
+        ("mandatory-checkbox-answer", "Head", "Head label"),
+        ("mandatory-checkbox-answer", "Body", "Body label"),
+    ],
+)
+def test_option_label_from_value_with_placeholder_label(
```
MebinAbraham, line 403: We should unit test most of it in operation.py. We also need to unit test with labels that contain placeholders. The unit test here could just be for coverage assuming it is reusing the function from operations.py.

MebinAbraham (changes requested):

app/questionnaire/placeholder_transforms.py
```diff
@@ -292,3 +298,33 @@ def _create_hyperlink(href: str, link_text: str) -> str:
 
     def list_item_count(self, list_to_count: Optional[Sized]) -> int:
         return self._operations.evaluate_count(list_to_count)
+
+    def option_label_from_value(self, value: str, answer_id: str) -> str:
```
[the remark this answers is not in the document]
parimooa: Due to circular reference issue , Placeholder parser can't be used in `operations.py`.

app/questionnaire/rules/operations.py
```diff
@@ -214,3 +229,24 @@ def evaluate_map(
...
+        answers = self.schema.get_answers_by_answer_id(answer_id)
+        label_options: str = next(
+            (
+                options["label"]
+                for answer in answers
+                for options in answer["options"]
+                if value == options["value"]
+            ),
+            "",
+        )
```
MebinAbraham, line 244: Personally, I think a list/dict comphrenshion would be more readable than a generator with `next`. However, not a blocker. Performance diff is negligible.

app/questionnaire/rules/operations.py
```diff
@@ -214,3 +229,24 @@ def evaluate_map(
...
+
+        if isinstance(label_options, str):
+            label = label_options
+
+        elif isinstance(label_options, dict):
```
MebinAbraham, line 249: Could be an `else`, otherwise it would be an invalid schema anyway.

tests/app/questionnaire/conftest.py
```diff
@@ -815,6 +818,194 @@ def section_with_repeating_list():
     }
 
 
+@pytest.fixture
+def section_with_labels():
```
MebinAbraham, line 822: The fixture name isn't very clear. Maybe just `labels_schema_with_placeholders`  would be clearer.

app/questionnaire/rules/operations.py
```diff
@@ -49,9 +57,16 @@ class Operations:
...
-    def __init__(self, language: str) -> None:
+    def __init__(
+        self,
+        language: str,
+        schema: QuestionnaireSchema,
+        renderer: Optional["PlaceholderRenderer"] = None,
+    ) -> None:
         self._language = language
         self._locale = DEFAULT_LOCALE if language in {"en", "eo"} else language
+        self.renderer = renderer or None
```
MebinAbraham, line 68: We should always pass in a renderer. Anything operations that depend on it will error at run time otherwise. A lot on placeholder parser shouldn't be optional either, but for anything new we added, we should always have them available, hence should pass it. We should accept that tests needs updating, as we shouldn't avoid because of that.

tests/app/questionnaire/conftest.py
```diff
@@ -831,5 +1022,33 @@ def mock_schema():
...
+    )
+    renderer = PlaceholderRenderer(
+        language="en",
+        answer_store=answer_id,
+        list_store=ListStore(),
+        metadata=ImmutableDict({}),
+        response_metadata={},
+        schema=schema,
+    )
+    return schema, renderer
```
MebinAbraham, line 1047: We should make the schema its own fixture. I think it's confusing to return unrelated things from one fixture

tests/app/questionnaire/conftest.py
```diff
@@ -831,5 +1022,33 @@ def mock_schema():
...
+
+
+@pytest.fixture
+def operation_helper(schema_placeholder_renderer):
+    schema, renderer = schema_placeholder_renderer
+    ops_helper = OperationHelper(language="en", schema=schema, renderer=renderer)
+    return ops_helper
```
MebinAbraham, line 1054: If we are using this in only one test file, we should move it there, as it makes the test more readable.

tests/app/questionnaire/conftest.py
MebinAbraham: To be honest, I think this should be done directly in the test it is used in. It is being set up for that specific test. I think some duplication in test placeholder transform is fine, as at the moment, it is hard to follow as the main set-up for the test is done in the fixture instead of the test body.

tests/app/questionnaire/conftest.py
```diff
@@ -831,5 +1022,33 @@ def mock_schema():
...
-    return PlaceholderTransforms(language="en")
+def placeholder_transform(schema_placeholder_renderer):
+    schema, renderer = schema_placeholder_renderer
+    return PlaceholderTransforms(language="en", schema=schema, renderer=renderer)
+
+
+@pytest.fixture
+def schema_placeholder_renderer(section_with_labels):
+    schema = QuestionnaireSchema(section_with_labels)
+    answer_id = AnswerStore(
```
MebinAbraham, line 1033: minor, variable doesn't match assignment.

MebinAbraham (commented):
I think the schema needs to be simplified, it is not very clear on how to use it. However, I think it would be better to just create the scenario highlighted on the card, the one in overflow. We did also post in the card's description a while back `new schema to showcase the 3 question overflow scenario of this requirement`.  That would cover all scenarios, including piping from radios as well as checkbox.

app/questionnaire/rules/operations.py
parimooa: the intention was to assign default `str` if `value` is not found , not the performance.  In list comprehension , you have to convert `list` to str and i don't think you can assign default using list/dict comprehension.

tests/app/questionnaire/conftest.py
parimooa: They are related as this renderer and schema is specific to placeholder test. However, rest of the tests now use `mock_schema` and `mock_renderer`. fixtures Also as suggested , Placeholder is instantiated directly for labels tests

tests/app/questionnaire/conftest.py
parimooa: I prefer `fixtures` in `conftest` to keep separation between `tests` and `fixtures`.  there are couple of tests using this.

app/questionnaire/rules/operations.py
MebinAbraham: Value should never not be found if it is a valid schema. Only reason it might not be found is for dynamic options which we won't support here anyway.

> In list comprehension , you have to convert list to str

Not sure I follow. You are already creating a generator, a list comp would be similar. i.e.

```py
        label_options: str = [
            options["label"]
            for answer in answers
            for options in answer["options"]
            if value == options["value"]
        ][0]
```

or:

```py
        label_options: str = {
            options["label"]: options["value"]
            for answer in answers
            for options in answer["options"]
        }[value]
```

parimooa: In my implementation, it assigns empty string `""` to `label_options`  if  generator raises `StopIteration` error , guarding against any error. 
Your suggestion may need another `if` condition in order to assign empty string `label_option = ""`  when `value` and `option['label']` don't match.
Happy to change if you think list/dict comp makes more sense

MebinAbraham: I don't have anything against the gen, I just think it is clearer with list/dict comp personally. In both cases we would get an error anyway, list would get index error and dict would get key error if value didn't exist, so that is covered, and we don't need to assign the label to an empty str as we should expect to find a value.

parimooa: I have seen instances where we return empty string as default if condition fails hence i used generator to assign empty. i agree list/dict comp is more readable. will update it to list comp

tests/app/questionnaire/conftest.py
```diff
@@ -831,5 +1059,49 @@ def mock_schema():
...
+    )
+    renderer = PlaceholderRenderer(
+        language="en",
+        answer_store=answer_store,
+        list_store=ListStore(),
+        metadata=ImmutableDict({}),
+        response_metadata={},
+        schema=schema,
+    )
+    return schema, renderer
```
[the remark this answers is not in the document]
MebinAbraham: On second thought, we should remove `labels_schema_with_placeholders` fixture schema which is essentially a duplicate of the actual test schema and instead just load it in via 
```py
load_schema_from_name("test_option_label_from_value")`
````
Then that fixture should be named explicitly to something like `option_label_from_value_schema`. If you do this approach, be sure to update the unit test asserts, as the label/value is currently mismatching.

The other option is to just mock the schema methods like `get_answers_by_answer_id`. Either way, we should look to prevent the duplication.

File: tests/functional/spec/option_label_from_value.spec.js [hunk 14]
```diff
@@ -0,0 +1,37 @@
...
+import RecoveryQuestionCheckboxBlockPage from "../generated_pages/option_label_from_value/recovery-question-checkbox-block.page";
+import MandatoryRadioPage from "../generated_pages/option_label_from_value/mandatory-radio.page";
+import RecoveryQuestionRadioBlockPage from "../generated_pages/option_label_from_value/recovery-question-radio-block.page";
+
+describe("Option label value check", () => {
+  beforeEach("Load the survey", () => {
+    browser.openQuestionnaire("test_option_label_from_value.json");
+  });
+
+  it("Given multiple checkbox options are provided, when I select a single answer and go to the next page, then the placeholder contains the label text of the answer I selected", () => {
```
MebinAbraham, line 11:
```suggestion
  it("Given multiple checkbox options are provided, when I select a single answer and go to the next page, then the question title contains the label text of the answer I selected", () => {
```

tests/functional/spec/option_label_from_value.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it("Given multiple checkbox options are provided, when I select more than one answer and go to the next page, then I am asked to select only one answer from a list", () => {
+    $(MandatoryCheckboxPage.head()).click();
+    $(MandatoryCheckboxPage.body()).click();
+    $(MandatoryCheckboxPage.submit()).click();
+    expect($(MandatoryRadioPage.questionText()).getText()).to.contain(
+      "If you suffered any one injury from the options below, please select the most serious one"
+    );
+  });
```
MebinAbraham, line 26: This isn't needed as this is testing routing which is covered elsewhere and not the purpose of this spec.

tests/functional/spec/option_label_from_value.spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+  it("Given multiple checkbox options are provided and I subsequently select a single answer from the next displayed list, when I select one answer and go to the next page, then the placeholder contains the label text of the answer I selected", () => {
+    $(MandatoryCheckboxPage.head()).click();
+    $(MandatoryCheckboxPage.body()).click();
+    $(MandatoryCheckboxPage.submit()).click();
+    $(MandatoryRadioPage.nose()).click();
+    $(MandatoryRadioPage.submit()).click();
+
+    expect($(RecoveryQuestionRadioBlockPage.questionText()).getText()).to.contain("Nose");
```
MebinAbraham, line 35: This should be two expects, or we should have 2 tests. One to assert the label from the checkbox is piped in the radio option. One to assert the label from that radio is then piped into the question title.

tests/app/questionnaire/rules/test_operators.py
```diff
@@ -40,7 +39,7 @@
     "operands, expected_result",
     equals_operations,
 )
-def test_operation_equal(operands, expected_result):
+def test_operation_equal(operands, expected_result, get_operator):
     operator = get_operator(Operator.EQUAL)
```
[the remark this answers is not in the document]
parimooa: Fixture has to be a class instance with params, this seems to be the only way to do it.  Not sure why you find it odd. If you think there is a better way, happy to discuss. Calling fixture in function is deprecated ( https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly)

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=answer-labels-as-placeholders
```
parimooa, line 2: change this to latest when `validator` is merged

schemas/test/en/test_placeholder_option_label_from_value.json
```diff
@@ -0,0 +1,266 @@
...
+                                            },
+                                            {
+                                                "label": "Right Arm",
+                                                "value": "Right Arm",
+                                                "q_code": "2"
+                                            },
+                                            {
+                                                "label": "Left Arm",
+                                                "value": "Left Arm",
+                                                "q_code": "3"
```
MebinAbraham, line 64: q_codes can be removed, they are not needed for data version 0.0.3.

schemas/test/en/test_placeholder_option_label_from_value.json
```diff
@@ -0,0 +1,266 @@
...
+        {
+            "id": "checkbox-section",
+            "groups": [
+                {
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "mandatory-checkbox",
+                            "question": {
+                                "answers": [
```
MebinAbraham, line 42: We should add some guidance on this question page on how to use this schema. We do this with a couple of schemas. It will make it easier to know how to test. So something simple guidance to state, scenario one, choose 1 option which will be piped into the question title, and scenario two, pick more than 1, and the first selected answer will be piped into a radio etc.

schemas/test/en/test_placeholder_option_label_from_value.json
```diff
@@ -0,0 +1,266 @@
...
+                                "type": "General"
+                            }
+                        },
+                        {
+                            "type": "Question",
+                            "id": "recovery-question-checkbox-block",
+                            "question": {
+                                "id": "recovery-question-checkbox",
+                                "title": {
+                                    "text": "How long did it take to recover from the injury to your {body_part}?",
```
MebinAbraham, line 81: Just for consistency as the other routes has it highlighted.
```suggestion
                                    "text": "How long did it take to recover from the injury to your <em>{body_part}</em>?",
```

MebinAbraham (commented):
Looks good, minor issues.

tests/functional/spec/placeholder_option_label_from_value.spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+describe("Option label value check", () => {
+  beforeEach("Load the survey", () => {
+    browser.openQuestionnaire("test_placeholder_option_label_from_value.json");
+  });
+
+  it("Given multiple checkbox options are provided, when I select a single answer and go to the next page, then the question title contains the label text of the answer I selected", () => {
+    $(MandatoryCheckboxPage.head()).click();
+    $(MandatoryCheckboxPage.submit()).click();
+    expect($(RecoveryQuestionCheckboxBlockPage.questionText()).getText()).to.contain("Head");
```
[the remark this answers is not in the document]
parimooa: This is not piped , it is the only option selected so it should be as it is in Question title . It is only when we select more than one ?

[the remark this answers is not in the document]
MebinAbraham: It is still piped from the radio option?. Radio also pipes in its answer.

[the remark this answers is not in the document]
MebinAbraham: As discussed, this is for when selecting the single checkbox, my mistake.

tests/functional/spec/placeholder_option_label_from_value.spec.js
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  it("Given multiple checkbox options are provided, when I select more than one answer and go to the next page, then first option in radio is piped from checkbox", () => {
+    $(MandatoryCheckboxPage.head()).click();
+    $(MandatoryCheckboxPage.body()).click();
+    $(MandatoryCheckboxPage.submit()).click();
+    const answerPipedOption = $(MandatoryRadioPage.answerBodyPartPipedLabel()).selector;
+    expect($(answerPipedOption).getText()).to.contain("Head (piped)");
```
MebinAbraham, line 22: You don't need to get the selector specifically.
```suggestion
    expect($(MandatoryRadioPage.answerBodyPartPipedLabel()).getText()).to.contain("Head (piped)");
```

MebinAbraham (approved):
👍

LJBabbage (changes requested):
This part on the card doesn't seem to work for me 
```
Given I am on my checkbox answer options
And I have previously entered more than 1 answer into the checkbox question
When I have then selected an answer from the subsequent radio answer options
And I change one of those radio answers
Then I can see my selected radio box answer as part of the question
And The answer will be cleared and the respondent will need to answer it again 
```

I initially changed my answers on the radio and it just sent me back to the summary. Then I went back to the checkboxes and changed from 1st and 2nd to 3rd and 4th and again it just sent me back to the summary

tests/functional/spec/placeholder_option_label_from_value.spec.js
```diff
@@ -0,0 +1,33 @@
+import MandatoryCheckboxPage from "../generated_pages/placeholder_option_label_from_value/mandatory-checkbox.page";
```
MebinAbraham, line 1: https://github.com/ONSdigital/eq-questionnaire-runner/pull/756 introduced a `test/functional/spec/features/placeholder/` folder. We should move this to that.

tests/functional/spec/features/placeholder/placeholder_option_label_from_value.spec.js
```diff
@@ -0,0 +1,21 @@
+import MandatoryRadioPage from "../../../generated_pages/placeholder_option_label_from_value/mandatory-radio.page";
+import ConfirmationQuestionRadioBlockPage from "../../../generated_pages/placeholder_option_label_from_value/confirmation-question-radio-block.page";
+
+describe("Option label value check", () => {
+  beforeEach("Load the survey", () => {
+    browser.openQuestionnaire("test_placeholder_option_label_from_value.json");
+  });
+
+  it("Given radio options are provided and I subsequently select a single answer from the displayed list, when I select first answer (piped from metadata) and go to the next page, then the question title contains the label text of the answer I selected", () => {
```
LJBabbage, line 9: Not sure you need the single answer bit, just because it is a radio. So Given radio options, when I ...

LJBabbage (approved):

MebinAbraham (approved):

tests/app/questionnaire/rules/test_operation_helper.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+from app.questionnaire.rules.operations import DateOffset
+
+
+def test_string_to_datetime(operation_helper):
+    offset = DateOffset(days=-1 * 7, day_of_week="MONDAY")
```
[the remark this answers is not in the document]
parimooa: Copy-pasted :-)  for coverage as it is duplicate test from `operations.py` and  It happens to be -1 x 7 :-).  We had to get around the circular reference issue in operations.py so ended up having introduce `operation_helper.py` which just calls `operation` methods.

PR #735, opened 2021-12-13
Adds GAR-specific push task

jamesnunn:
Updates the workflow to upload the runner image to Google Artifact Registry

### How to review 
Create a branch with no changes, push and create a PR to trigger the workflow. When workflow actions are complete, check that the image is in the image hosting project and is correctly tagged.
Run docker pull to check the image is publicly accessible.

### Checklist

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -173,7 +173,7 @@ jobs:
       - name: Build
         run: >
           docker build -t onsdigital/eq-questionnaire-runner:$TAG
-          -t eu.gcr.io/${{ secrets.GCR_PROJECT_ID }}/eq-questionnaire-runner:$TAG -t ${{ secrets.GAR_LOCATION }}/${{ secrets.GAR_PROJECT_ID }}/eq-questionnaire-runner/eq-questionnaire-runner:$TAG .
+          -t eu.gcr.io/${{ secrets.GCR_PROJECT_ID }}/eq-questionnaire-runner:$TAG -t ${{ secrets.GAR_LOCATION }}/${{ secrets.GAR_PROJECT_ID }}/docker-images/eq-questionnaire-runner:$TAG .
       - name: Push to Docker Hub
         run: |
           echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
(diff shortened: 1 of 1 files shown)


jamesnunn:
[Use GAR in sandbox and push Runner via Github Actions (3 SP)](https://trello.com/c/11y1a33L/5025-use-gar-in-sandbox-and-push-runner-via-github-actions-3-sp)

MebinAbraham (commented):
So it looks like the images are stored under its own repo already, hence the repeated registry name, "europe-west2-docker.pkg.dev/ons-eq-ci/**eq-questionnaire-runner/eq-questionnaire-runner**/".

Can we just make a GCR repository be called like `docker-images`?. Then we can use the single GCR repository for all our git repos, which will be separated by the git repo name anyway. All our repos share the same permissions, so shouldn't cause any issues. This is also one of the documented ways of using GAR, obv by repo isn't an issue, I just don't think we need it plus it means we don't need to create a repo for each of our git repos.

So instead of:
- `europe-west2-docker.pkg.dev/ons-eq-ci/eq-questionnaire-runner/eq-questionnaire-runner@sha256:123...`
- `europe-west2-docker.pkg.dev/ons-eq-ci/eq-questionnaire-launcher/eq-questionnaire-launcher@sha256:123...`

We would have:
- `europe-west2-docker.pkg.dev/ons-eq-ci/docker-images/eq-questionnaire-runner@sha256:123...`
- `europe-west2-docker.pkg.dev/ons-eq-ci/docker-images/eq-questionnaire-launcher@sha256:123...`

Thoughts?

jamesnunn:
I'm not _strongly_ against it, the only thing that wouldn't work like you mention is if the permissions had to be altered on any single repo which might be unlikely but the changes are non-trivial. All we'd be gaining is the small convenience of not having to specify the name twice and only having the upfront cost, albeit low, of creating a single docker repo.

MebinAbraham:
> I'm not _strongly_ against it, the only thing that wouldn't work like you mention is if the permissions had to be altered on any single repo which might be unlikely but the changes are non-trivial. All we'd be gaining is the small convenience of not having to specify the name twice and only having the upfront cost, albeit low, of creating a single docker repo.

Given we never needed perm per repo, I think it would be better to move to that single repo to more closing align with what GCR looks like. It's one of the ways to use GAR, we just need to choose a preference. @pricem14pc do you have a preference?

pricem14pc:
> > I'm not _strongly_ against it, the only thing that wouldn't work like you mention is if the permissions had to be altered on any single repo which might be unlikely but the changes are non-trivial. All we'd be gaining is the small convenience of not having to specify the name twice and only having the upfront cost, albeit low, of creating a single docker repo.
> 
> Given we never needed perm per repo, I think it would be better to move to that single repo to more closely align with what GCR looks like. It's one of the ways to use GAR, we just need to choose a preference. @pricem14pc do you have a preference?

pricem14pc:
Sigh...

MebinAbraham (approved):
👍

pricem14pc (approved):
Successful push to GAR, and gcloud authenticated image pulled and run locally as expected.

PR #738, opened 2021-12-16
Update previous behaviour after coming from summary page

berroar:

Updates the behaviour so that if you go back to a question page from the summary page, when you click the `previous` button, then you return back to the summary. 

### How to review 

Use the `test_section_summary` or `test_submit_with_summary` schema and
 - Complete the survey
 - Use the `change` link on the `Summary` page to edit a question
 - On the `Question` page hit `previous`
 - Check that you are returned to the Summary page

### Checklist


berroar:
[[AA] Behaviour on using previous after coming from summary (3 SP)](https://trello.com/c/YCRzVRBr/4006-aa-behaviour-on-using-previous-after-coming-from-summary-3-sp)

tests/app/questionnaire/test_router.py
```diff
@@ -561,6 +561,33 @@ def test_within_section(self):
...
+
+        current_location = Location(
+            section_id="property-details-section", block_id="insurance-type"
+        )
+        routing_path = RoutingPath(
+            ["insurance-type", "insurance-address"], section_id="default-section"
+        )
+        previous_location = self.router.get_previous_location_url(
+            current_location, routing_path, return_to="section-summary"
+        )
```
MebinAbraham, line 585: We should have one for final summary also.

tests/functional/spec/features/submit_page/submit_page_summary.spec.js
```diff
@@ -36,6 +36,15 @@ describe("Submit Page with Summary", () => {
...
   });
 
+  it("Given a questionnaire with a summary has been completed and a summary page edit link is clicked, when I click previous, then it should return to the summary", () => {
+    completeAllQuestions();
+
+    $(SubmitPage.radioAnswerEdit()).click();
+    $(RadioPage.previous()).click();
+
+    expect(browser.getUrl()).to.contain(SubmitPage.pageName);
+  });
```
MebinAbraham, line 46: We should also test that this only happens on the first page you land on. I don't think we have a test to assert the next page's previous link doesn't go to summary, but just the previous question, essentially asserting the return_to isn't persisted. You need to change path to test this. Not sure if this change allow for path changes, if not the `test_section_summary` one does.

MebinAbraham (commented):
Functionally as expected. Just a few things to consider.

app/questionnaire/router.py
```diff
@@ -124,15 +124,12 @@ def get_next_location_url(
...
-                return self._get_section_url(location)
-
-            if return_to == "final-summary" and self.is_questionnaire_complete:
-                return url_for("questionnaire.submit_questionnaire")
-
             if is_last_block_in_section:
                 return self._get_next_location_url_for_last_block_in_section(location)
 
+            if return_to:
+                return self._get_return_to_location_url(location, return_to)
```
MebinAbraham, line 131: How come the order changes?. Also `_get_return_to_location_url` may not return anything, so this needs to check you have a return to and are able to get a return to url.

eg:
```py
if return_to and (return_to_url := self._get_return_to_location_url(location, return_to)):
	return return_to_url

File: app/questionnaire/router.py [hunk 4]
```diff
@@ -148,11 +145,20 @@ def _get_next_location_url_for_last_block_in_section(
...
+        return_to: Optional[str] = None,
     ) -> Optional[str]:
         """
         Returns the previous 'location' to visit given a set of user answers
         """
+
+        if return_to and self._progress_store.is_section_complete(
+            location.section_id, location.list_item_id
+        ):
+            return self._get_return_to_location_url(location, return_to)
```
MebinAbraham, line 160: Same here, it should only return if you have a return to url.

app/questionnaire/router.py
berroar: Reverted order

app/questionnaire/router.py
```diff
@@ -175,6 +186,13 @@ def get_previous_location_url(
 
         return None
 
+    def _get_return_to_location_url(self, location: Location, return_to: str) -> str:
+        if return_to == "section-summary":
+            return self._get_section_url(location)
+
+        if return_to == "final-summary" and self.is_questionnaire_complete:
+            return url_for("questionnaire.submit_questionnaire")
```
MebinAbraham, line 194: Since previous URLs are rendered on the page and is clicked on by the users instead of us redirecting, the anchor link is no longer persisted when going back to the summary, so we will need to keep that in mind and think of a solution ad it will be needed for the upcoming return to summary journey.

MebinAbraham (commented):
We do have other overridden previous location methods, however, they all seem sensible at the moment.

Looks good, minor issues left.

MebinAbraham (approved):

app/questionnaire/router.py
```diff
@@ -148,11 +147,23 @@ def _get_next_location_url_for_last_block_in_section(
...
 
     def get_previous_location_url(
-        self, location: Location, routing_path: RoutingPath
+        self,
+        location: Location,
+        routing_path: RoutingPath,
+        return_to: Optional[str] = None,
     ) -> Optional[str]:
         """
         Returns the previous 'location' to visit given a set of user answers
```
LJBabbage, line 156: Might need a tweak to this now

tests/functional/spec/features/section_summary/section_summary.spec.js
```diff
@@ -84,6 +84,15 @@ describe("Section Summary", () => {
       expect(browser.getUrl()).to.contain(AddressDurationPage.pageName);
     });
 
+    it("When I select edit from Final Summary and change an answer and then go to the next question and click previous, Then I should return to the question I originally edited", () => {
```
LJBabbage, line 87: This wasn't the test I was expecting, I thought it would be the same as the summary one, i.e you go back from the section summary then click previous to get back to the section summary. Am I missing something, or do we have that already?

LJBabbage (changes requested):
Couple of minor things, but works as expected

tests/functional/spec/features/section_summary/section_summary.spec.js
berroar: This was an additional test in order to cover the scenario Mebin commented on here: https://github.com/ONSdigital/eq-questionnaire-runner/pull/738#discussion_r770393792

berroar: Perhaps the title needs to be updated to make it clearer, but difficult to come up with a good given when, then

LJBabbage: Do we have one that covers a return to the section summary by not changing any answers and using previous?

berroar: Not at the moment, I will add one, it is only covered in the unit test

berroar: Added another test now 👍

LJBabbage (approved):

PR #739, opened 2021-12-16
Runner to use new rules engine for variants

LJBabbage:
This PR allows the new rules engine to be used with question and content variants 

### How to review 
Using test_variants_question.json, test_variants_content.json, test_variants_first_item_in_list.json Make sure everything works as expected

### Checklist


app/questionnaire/questionnaire_schema.py
```diff
@@ -603,6 +602,24 @@ def has_address_lookup_answer(question: Mapping) -> bool:
             if answer["type"] == "Address" and "lookup_options" in answer
         )
 
+    def _get_values_for_key(
```
LJBabbage, line 605: This was only being used in questionnaire_schema and was causing a dependancy loop in schema_utils when RuleEvaluator was added. I have also tweaked it as ignore_keys was not been passed on recursion, which it should have been

schemas/test/en/test_new_variants_content.json
```diff
@@ -0,0 +1,106 @@
+{
```
LJBabbage, line 1: These new schemas are just copies of the old ones with the when's updates, I have not necessarily reviewed them for content, although I did spot an error on one

tests/app/questionnaire/test_schema_utils.py
```diff
@@ -209,20 +208,3 @@ def test_choose_question_to_display(question_variant_schema):
     )
 
     assert question_to_display["title"] == "Question 1, No"
-
-
-def test_get_values_for_key_ignores_variants():
```
LJBabbage, line 214: Due to the moving of get_values_for_key I have extended tests/app/questionnaire/conftest.py to cover these. They were also not working quite right, giving fake line coverage and hiding a bug. We never initially pass blocks into get_values_for_key only sections, only on the recursion of get_values_for_key did a block get passed, at which point it didn't pass the ignore_keys.

tests/integration/questionnaire/test_questionnaire_question_variants.py
```diff
@@ -39,6 +49,7 @@ def test_summaries_proxy(self):
         self.assertInBody("What is your age")
 
     def complete_first_section(self, proxy=False):
+
```
LJBabbage, line 52: Will remove the extra line, when reviews come in

app/views/contexts/summary/block.py
```diff
@@ -38,16 +39,23 @@ def get_question(
         block_schema, answer_store, list_store, metadata, schema, location
     ):
         """ Taking question variants into account, return the question which was displayed to the user """
+
```
LJBabbage, line 42: Will remove the extra line, when reviews come in

MebinAbraham (commented):
There is a bug in `PathFinder._remove_rule_answers` when changing old answer, as it is expecting old rules structure.

Steps to reproduce:
1. Launch a new variant question schema
2. Who is this questionnaire about?: Enter some name
3. Are you `<name>`?: Yes
4. Age?: 20 then save and continue
5. Click browser back until `Are you <name>?`
6. Change answer to `No, I am answering on their behalf`
7. `<name> is over 16?`: No
You should now get a 500 error.

tests/app/questionnaire/test_schema_utils.py
LJBabbage: They were lame duck tests and were only put in to hit this line. https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/questionnaire/schema_utils.py#L138. If the ignore_keys was corrected they weren't needed anyway, assuming the conftest was extended. The rest of the method is already tested with this https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/tests/app/questionnaire/test_questionnaire_schema.py#L203, although you could argue it isn't explicit. However now it is private it is difficult to get at, without an ignore. We could do it, but I would argue it should be extended to cover all scenarios

app/views/contexts/summary/block.py
```diff
@@ -38,16 +39,23 @@ def get_question(
...
+                display_variant = when_rule_evaluator.evaluate(variant.get("when"))
+            else:
+                display_variant = evaluate_when_rules(
+                    variant.get("when"),
+                    schema,
+                    metadata,
+                    answer_store,
+                    list_store,
+                    location,
+                )
```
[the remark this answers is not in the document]
LJBabbage: Good point, I blame the original developer 😆

LJBabbage:
> There is a bug in `PathFinder._remove_rule_answers` when changing old answer, as it is expecting old rules structure.
> 
> Steps to reproduce:
> 
> 1. Launch a new variant question schema
> 2. Who is this questionnaire about?: Enter some name
> 3. Are you `<name>`?: Yes
> 4. Age?: 20 then save and continue
> 5. Click browser back until `Are you <name>?`
> 6. Change answer to `No, I am answering on their behalf`
> 7. `<name> is over 16?`: No
>    You should now get a 500 error.

Good spot, must be a gap in our tests

LJBabbage:
> > There is a bug in `PathFinder._remove_rule_answers` when changing old answer, as it is expecting old rules structure.
> > Steps to reproduce:
> > 
> > 1. Launch a new variant question schema
> > 2. Who is this questionnaire about?: Enter some name
> > 3. Are you `<name>`?: Yes
> > 4. Age?: 20 then save and continue
> > 5. Click browser back until `Are you <name>?`
> > 6. Change answer to `No, I am answering on their behalf`
> > 7. `<name> is over 16?`: No
> >    You should now get a 500 error.
> 
> Good spot, although is this specific to variants?

As discussed, this isn't directly related to variants but has highlighted an issue that has been generate by the new routing engine. I new card will be raised

schemas/test/en/test_new_variants_content.json
petechd: Some minor changes possibly needed, answer label shouldn't be a question. Should be something like the label for age answer in the next schema.

schemas/test/en/test_new_variants_question.json
```diff
@@ -0,0 +1,621 @@
+{
```
petechd, line 1: Section title probably missing in this one, will display as 'None' on summary page.

app/questionnaire/schema_utils.py
```diff
@@ -39,10 +40,22 @@ def _choose_variant(
...
         when_rules = variant.get("when", [])
-        if evaluate_when_rules(
+
+        if isinstance(when_rules, dict):
+            when_rule_evaluator = RuleEvaluator(
+                schema,
+                answer_store,
+                list_store,
+                metadata,
+                {},
```
MebinAbraham, line 52: Since variants can now used the new rules, we should pass in response metadata. It needs to be passed into `transform_variants` and down to this method. Where `transform_variants` is called, `response_metadata` data is available.

MebinAbraham (approved):

petechd (approved):

PR #742, opened 2021-12-17
Upgrade python packages

jamesnunn:
Upgrades packages

### How to review 
Check changes make sense.

### Checklist


app/forms/questionnaire_form.py
```diff
@@ -57,6 +57,17 @@ def __init__(
...
+    def data(self):
+        d = super().data
+        try:
+            if d["csrf_token"] is None:
+                d["csrf_token"] = ""
+        except KeyError:
+            pass
+
+        return d
+
```
jamesnunn, line 70: `csrf_token` was being returned as `None` so have to convert specifically. Can't convert all cases of `None`s because some are tested as `None`....

app/forms/fields/select_field_with_detail_answer.py
```diff
@@ -28,8 +28,9 @@ def iter_choices(self):
             yield value, label, self.coerce(value) == self.data, detail_answer_id
 
     def pre_validate(self, form):
-        for value, _, _ in self.choices:
-            if value == self.data:
-                break
-        else:
-            raise ValueError(self.gettext("Not a valid choice"))
+        if self.data is not None:
```
jamesnunn, line 31: This is a nuclear option to get tests working so I can see where changes can be made

app/forms/fields/date_field.py
```diff
@@ -33,8 +34,8 @@ def __init__(self, validators, **kwargs):
         form_class = get_form_class(validators)
         super().__init__(form_class, **kwargs)
 
-    def process(self, formdata, data=None):
-        if data is not None:
+    def process(self, formdata, data=None, extra_filters=None):
+        if data is not None and not isinstance(data, UnsetValue):
```
jamesnunn, line 38: `data` is now an `UnsetValue`, not None, but unsure if it's necessary to remove the `not None` check

app/forms/questionnaire_form.py
```diff
@@ -57,6 +57,14 @@ def __init__(
 
         super().__init__(**kwargs)
 
+    @property
+    def data(self):
+        d = super().data
+        if d["csrf_token"] is None:
+            d["csrf_token"] = ""
+
+        return d
```
jamesnunn, line 66: `csrf_token` is now parsed as `None` if non existent, I'm not sure of the impact so I convert it back to blank string here to be safe but it is dirty...

jamesnunn:
[Upgrade runner Python packages](https://trello.com/c/3KMO5s8r/5035-upgrade-runner-python-packages)

app/forms/fields/multiple_select_field_with_detail_answer.py
```diff
@@ -24,6 +24,7 @@ def __iter__(self):
             yield opt
 
     def iter_choices(self):
-        for value, label, detail_answer_id in self.choices:
-            selected = self.data is not None and self.coerce(value) in self.data
-            yield value, label, selected, detail_answer_id
+        if self.choices is not None:
```
MebinAbraham, line 27: When/Why would choices be None?. Is there some PR that explains the changes?.

MebinAbraham (commented):
Looks good, just a question about the choices being None.

MebinAbraham (commented):
When was the Pipfile lock generated?. If it has been a few weeks, i.e since original testing, it is worth updating it again to ensure you got the most up-to-date packages on merge.

app/forms/fields/multiple_select_field_with_detail_answer.py
jamesnunn: Good spot, it's unnecessary, I added the check as a workaround to another now-not-needed workaround I had introduced. Removed now

jamesnunn:
> When was the Pipfile lock generated?. If it has been a few weeks, i.e since original testing, it is worth updating it again to ensure you got the most up-to-date packages on merge.

Updated that too

app/forms/fields/select_field_with_detail_answer.py
```diff
@@ -28,8 +29,12 @@ def iter_choices(self):
             yield value, label, self.coerce(value) == self.data, detail_answer_id
 
     def pre_validate(self, form):
-        for value, _, _ in self.choices:
-            if value == self.data:
+
+        if not self.choices:
```
LJBabbage, line 33: As mentioned in a different part I don't think this can happen https://github.com/ONSdigital/eq-questionnaire-validator/blob/2e3948aacf6f3c892919649481f2ca45e0036b18/app/validators/answers/option_answer_validator.py#L21

app/forms/fields/select_field_with_detail_answer.py
```diff
@@ -28,8 +29,12 @@ def iter_choices(self):
...
+            raise TypeError(self.gettext("Not a valid choice"))
+
```
LJBabbage, line 35: I probably would want to align with the library now validate_choice is a thing https://github.com/wtforms/wtforms/blob/563b781bb0262ff71dda94f094b5d5413a87af7a/src/wtforms/fields/choices.py#L138

tests/app/forms/field_handlers/test_select_handler.py
```diff
@@ -54,3 +56,41 @@ class TestForm(Form):
     assert form.test_field.label.text == radio_json["label"]
     assert form.test_field.description == radio_json["guidance"]
     assert form.test_field.choices == expected_choices
+
+
+def test_get_field_with_bad_choices(value_source_resolver):
+    radio_json = {
+        "guidance": "",
```
LJBabbage, line 63: I can see this is a copy and paste, so understand why it is like this, but I don't think this is would validate.  Reckon you should slimline out the stuff we don't need and get rid of guidance (the bit I don't think we use like that anymore), description, q_code and validation message.

tests/app/forms/field_handlers/test_select_handler.py
```diff
@@ -54,3 +56,41 @@ class TestForm(Form):
...
+        SelectHandler, "build_choices_with_detail_answer_ids", return_value=None
+    ):
+        handler = SelectHandler(radio_json, value_source_resolver, error_messages)
+
+        class TestForm(Form):
+            test_field = handler.get_field()
+
+        form = TestForm()
+
+        with pytest.raises(TypeError):
```
LJBabbage, line 95: I was under the impression from the code above that TypeError was raised on empty choices and this should be ValidationError unless I have missed something
```        
if not self.choices:
      raise TypeError(self.gettext("Not a valid choice"))
...
 else:
      raise ValidationError(self.gettext("Not a valid choice"))

````

**Review by LJBabbage — CHANGES_REQUESTED** [review]

File: app/forms/fields/select_field_with_detail_answer.py (hunk 7 above)
Comment (jamesnunn) [reply app/forms/fields/select_field_with_detail_answer.py:35]: the default behaviour is to validate - so you think we should now support the ability to not validate now too? How deeply integrated should this be, considering it's an upgrade?

File: tests/app/forms/field_handlers/test_select_handler.py (hunk 8 above)
Comment (jamesnunn) [reply tests/app/forms/field_handlers/test_select_handler.py:63]: So you mean it shouldn't validate with those parts you said to remove? Or it's invalid despite those superfluous bits.

File: app/forms/fields/select_field_with_detail_answer.py (hunk 7 above)
Comment (LJBabbage) [reply app/forms/fields/select_field_with_detail_answer.py:32]: You are right it can be left as is, after all the validate_choice is a new feature we didn't have before and that is the default now. I have a tenancy when overwriting to align as much as possible, but in this instance it can be ignored until we need it.

File: tests/app/forms/field_handlers/test_select_handler.py (hunk 8 above)
Comment (LJBabbage) [reply tests/app/forms/field_handlers/test_select_handler.py]: I can see you have already updated, it was just about reducing the amount of JSON and removing keys which aren't used anymore.

File: tests/app/forms/field_handlers/test_select_handler.py [hunk 10]
```diff
@@ -54,3 +56,35 @@ class TestForm(Form):
     assert form.test_field.label.text == radio_json["label"]
     assert form.test_field.description == radio_json["guidance"]
     assert form.test_field.choices == expected_choices
+
+
+def test_get_field_with_bad_choices(value_source_resolver):
```
MebinAbraham, line 61: Is this still needed given the if choices check has been removed?. This shouldn't happen?.

jamesnunn: No this is necessary for  an empty choices list.

tests/app/forms/field_handlers/test_select_handler.py
jamesnunn: Yeah the TypeError was actually being raised from trying to iterate over a `None` object which was from a dodgy mock which is why the test was passing. Turns out I actually need to test presence of `form.errors` instead.

tests/app/forms/field_handlers/test_select_handler.py
MebinAbraham: What does that mean?. We should have had tests for choices that don't match the available choices?. I ask because the test is basically using an invalid schema.

jamesnunn: Lewis mentioned the same yet the schema apparently validates in the existing test using it - I used the same schema. So is the existing test bad?

I figured the new code for `SelectFieldWithDetailAnswer.pre_validate` needs coverage but as it happens I ran the unittests without that test and it still says 100% coverage so I guess that's alright

jamesnunn: Unless of course the schema isn't validated in the test

MebinAbraham (approved):

LJBabbage (approved):

PR #743, opened 2021-12-20
Fix Design System CDN url file path 

parimooa:
Fix errors on browsers developer console. Currently the app tries to fetch the `favicons` and `manifest.json` from a non-existent file path. This issue happens due to the wrong value of `theme` which makes the file path for these files.

**Note:** This issue is only visible through chrome's developers console. Safari doesn't report any error

### How to review 
Launch any survey and observer there is no `favicons` or `manisfest.json` file not found errors

### Checklist


LJBabbage (approved):
Manually test by changing one of the schema to Census theme, it worked. We perhaps should try to separate out the use of term theme as we have 2 uses (schema and design_system) I immediately looked at the schema to see a theme as default, only to realise later that the var was something different (i.e design_system_theme), but it is pre-existing. Would be good to test this somehow, but not sure how to

templates/layouts/_base.html
```diff
@@ -46,6 +45,10 @@
   "cspNonce": csp_nonce
 } %}
 
+{% if theme %}
+  {% do pageConfig.update({ "theme":theme}) %}
```
LJBabbage, line 49: A little extra padding?

berroar (approved):

PR #744, opened 2021-12-20
Update tooltips unit label

berroar:

Updates runner so that the tooltip displays a full unit label when a unit is being highlighted.

### How to review 

Use the `test_unit_patterns` schema to check the change. The full unit description should be displayed in the tooltip when highlighted.

### Checklist


berroar:
[Remove tooltips from runner (1 SP)](https://trello.com/c/aEF37ouR/4753-remove-tooltips-from-runner-1-sp)

tests/app/test_jinja_filters.py
```diff
@@ -139,7 +139,7 @@ def test_format_unit_welsh(self):
         self.assertEqual(format_unit("duration-year", 100, "long"), "100 mlynedd")
 
     @patch("app.jinja_filters.flask_babel.get_locale", Mock(return_value="en_GB"))
-    def test_format_unit_input_label(self):
+    def test_format_unit_input_label_short(self):
```
petechd, line 142: We could add a functional test using something we already used in our tests, look through html since the tooltip element is in body of the document.

tests/functional/spec/features/units.spec.js
```diff
@@ -32,4 +32,13 @@ describe("Units", () => {
     expect($(SubmitPage.durationHour()).getText()).to.equal("6 awr");
     expect($(SubmitPage.durationYear()).getText()).to.equal("20 mlynedd");
   });
+
+
+  it("Given we open a questionnaire with unit labels, when the label is highlighted by the tooltip, then the full unit label should be displayed.", () => {
```
LJBabbage, line 36: long rather than full, just to align.

LJBabbage (approved):
Works as expected, the smallest of comments. Did find the component a bit odd, suffix, title and text and how they are rendered is not intuitive at all, but not us

tests/app/test_jinja_filters.py
berroar: Had to update the labels in the test schema in order for the functional test to work correctly.

petechd (approved):

tests/functional/spec/features/units.spec.js
```diff
@@ -32,4 +32,12 @@ describe("Units", () => {
     expect($(SubmitPage.durationHour()).getText()).to.equal("6 awr");
     expect($(SubmitPage.durationYear()).getText()).to.equal("20 mlynedd");
   });
+
+  it("Given we open a questionnaire with unit labels, when the label is highlighted by the tooltip, then the long unit label should be displayed.", () => {
+    browser.openQuestionnaire("test_unit_patterns.json", { language: "en" });
+    expect($("body").getHTML()).to.contain('title="centimeters"');
+    expect($("body").getHTML()).to.contain('title="meters"');
+    expect($("body").getHTML()).to.contain('title="kilometers"');
+    expect($("body").getHTML()).to.contain('title="miles"');
```
MebinAbraham, line 41: Minor, but we should avoid checking the entire html where possible. For this, we can get the value of the title explicitly, e.g:
```js
    expect($(SetLengthUnitsBlockPage.centimetres()).getAttribute("title")).to.equal("centimeters");
```

MebinAbraham (changes requested):

MebinAbraham (approved):

PR #745, opened 2021-12-20
Change static content links to point to RAS pages

petechd:
This PR removes `cookie_settings_url` from runner, moves the concatenation of the `contact_us_url` from `template_helpers` to the `survey_config` and introduce concatenation of the URL locations for `/cookie_settings_url` and `/privacy_and_data_protection` into `survey_config`.

### How to review 
Run new unit tests, visually check schema for links, check if this docker image deploys fine in pipelines using other related PRs: https://github.com/ONSdigital/eq-pipelines/pull/316 and https://github.com/ONSdigital/eq-terraform-gcp-runner/pull/15 .

[Deployed pipeline using this PR](https://concourse.gcp.onsdigital.uk/teams/eq-dev/pipelines/pete-test).

### Checklist


petechd:
[Static content in runner (3 SP)](https://trello.com/c/whNPzosV/4752-static-content-in-runner-3-sp)

tests/app/helpers/test_template_helpers.py
```diff
@@ -268,30 +329,6 @@ def test_get_page_header_context_census_nisra(app: Flask):
                 "target": "_blank",
             },
         ),
-        (
```
petechd, line 271: `contact-us` link generation has changed and I didn't see any value in updating/adding census tests since we don't use it anymore.

app/survey_config/survey_config.py
```diff
@@ -26,6 +28,16 @@ class SurveyConfig:
     account_service_surveys_path: Optional[str] = None
     title_logo: Optional[str] = None
     title_logo_alt: Optional[str] = None
+    contact_us_url: tuple[dict[str, Any]] = (
+        Link(lazy_gettext("Contact us"), f"{base_url}/contact-us/").__dict__,
+    )
+    cookie_settings_url: Optional[str] = f"{base_url}/cookies"
+    privacy_and_data_protection: tuple[dict[str, Any]] = (
```
berroar, line 35: Perhaps should be `privacy_and_data_protection_url` for consistency

petechd: Added, thanks!

app/survey_config/survey_config.py
```diff
@@ -1,10 +1,12 @@
+# pylint: disable=R1707
```
berroar, line 1: I think this can be removed

petechd: Removed👍

app/survey_config/survey_config.py
```diff
@@ -26,6 +27,16 @@ class SurveyConfig:
...
+    contact_us_url: tuple[dict[str, Any]] = (
+        Link(lazy_gettext("Contact us"), f"{base_url}/contact-us/").__dict__,
+    )
+    cookie_settings_url: Optional[str] = f"{base_url}/cookies"
+    privacy_and_data_protection_url: tuple[dict[str, Any]] = (
+        Link(
+            lazy_gettext("Privacy and data protection"),
+            f"{base_url}/privacy-and-data-protection",
+        ).__dict__,
+    )
```
MebinAbraham, line 39: This is a dict not tuple. It comes out as a tuple due to the trailing comma. However, that is superseeded by the fact that this should be the string urls, not `Link`, this is already wrong on master, but we should sort it now since it is being updated. We are actually sending a dict to the view as the url, instead of the str. If you generate a 500 error on submission, you can see the contact us link is mangled.

So in the init, we create the string urls which can be made into `Link` object for the footer portion.

MebinAbraham: We should also be consistent with the trailing `/`. Since runner default to including it, we can include for these also.

app/survey_config/business_config.py
```diff
@@ -18,7 +21,10 @@ class BusinessSurveyConfig(
     footer_links: Iterable[MutableMapping] = field(
         default_factory=lambda: [
             Link(lazy_gettext("What we do"), "#").__dict__,
-            Link(lazy_gettext("Contact us"), "#").__dict__,
+            Link(
+                lazy_gettext("Contact us"),
+                f"{base_url}/contact-us/",
```
MebinAbraham, line 26: We should reuse the link from the object, so `self.contact_us_url`.

app/survey_config/business_config.py
```diff
@@ -30,10 +36,10 @@ class BusinessSurveyConfig(
 
     footer_legal_links: Iterable[Mapping] = field(
         default_factory=lambda: [
-            Link(lazy_gettext("Cookies"), "#").__dict__,
+            Link(lazy_gettext("Cookies"), f"{base_url}/cookies").__dict__,
             Link(
                 lazy_gettext("Privacy and data protection"),
-                "#",
+                f"{base_url}/privacy-and-data-protection",
```
MebinAbraham, line 42: Same here, the URL should come from the property we already created instead of reconstructing.

MebinAbraham (changes requested):

tests/app/helpers/test_template_helpers.py
MebinAbraham: Highlights that our current approach won't supported translated link, if they were to have diff link for diff languages. One to keep in mind

app/survey_config/business_config.py
petechd: Changed, thanks!

app/survey_config/business_config.py
petechd: Reusing the survey config one now, thanks.

app/survey_config/survey_config.py
petechd: So all the Link dataclasses references are gone from this file, it creates strings now, which business config uses.

app/survey_config/business_config.py
```diff
@@ -18,7 +21,7 @@ class BusinessSurveyConfig(
     footer_links: Iterable[MutableMapping] = field(
         default_factory=lambda: [
             Link(lazy_gettext("What we do"), "#").__dict__,
-            Link(lazy_gettext("Contact us"), "#").__dict__,
+            Link(lazy_gettext("Contact us"), SurveyConfig.contact_us_url).__dict__,
```
MebinAbraham, line 24: Needs to the property from `self`, for which you need to make use of `__post_init__`. The same applies to the footer legal links.

app/survey_config/business_config.py
```diff
@@ -3,9 +3,12 @@
 
 from flask_babel import lazy_gettext
 
+from app.settings import ACCOUNT_SERVICE_BASE_URL
 from app.survey_config.link import Link
 from app.survey_config.survey_config import SurveyConfig
 
+base_url: str = ACCOUNT_SERVICE_BASE_URL
```
MebinAbraham, line 10: Unused?

petechd: Removed, thanks.

MebinAbraham (approved):

berroar (approved):

PR #746, opened 2021-12-24
Fix Remove Rule Answers Bug for old rules and introduce for new

LJBabbage:
This PR fixes the bug with remove answers when routing backwards. It also introduces it for the new when rules. Personally I don't think path_finder.py should be responsible for removing answers, but without a refactor I couldn't see an easy way to do it. 

### How to review 
With both test_new_confirmation_question and test_confirmation_question, make sure when you route backwards that the answer and block is removed and progress is updated to in progress. You can do this as follows

1. Launch the survey
2. For test_new_confirmation_question answer "yes" to the "Do you want the confirmation question to route backwards" question
3. Enter "0" in Total number of employees
4. Answer yes to the next question (the are slightly different for each one, but are both confirmation questions)
5. Check the JSON http://localhost:5000/dump/debug and make sure confirm-zero-employees-answer is in the answers and the block (confirm-zero-employees-block) is in the progress store. Confirm the section is COMPLETE
6. Go back, by using previous or by the browser
7. Click No or equivalent on the confirmation question
8. Now check the JSON again and make sure confirm-zero-employees-answer and confirm-zero-employees-block is gone, and the progress is IN_PROGRESS

### Checklist


app/questionnaire/questionnaire_store_updater.py
```diff
@@ -193,9 +193,9 @@ def remove_relationship_answers_for_list_item_id(
             self._answer_store.add_or_update(answer)
 
     def add_completed_location(self, location: Optional[Location] = None):
-        location = location or self._current_location
-
-        self._progress_store.add_completed_location(location)
+        if not self._progress_store.is_routing_backwards:
```
LJBabbage, line 196: Due to how our current architecture is structured the block was being removed, but because this came after routing it was immediate being put. Nearly send me nuts as unit tests confirmed it was being removed, but in UI it was still there, argggg.

schemas/test/en/test_new_confirmation_question.json
```diff
@@ -0,0 +1,175 @@
...
+                                        "mandatory": true
+                                    }
+                                ],
+                                "id": "confirm-zero-employees-question",
+                                "title": "Did you answer the last question correctly?"
+                            },
+                            "routing_rules": [
+                                {
+                                    "block": "number-of-employees-total-block",
+                                    "when": {
```
LJBabbage, line 145: You could argue we need a single rule in here as that is how it would normally be used. Naturally this exercise the recursive element

app/questionnaire/path_finder.py
```diff
@@ -204,15 +207,42 @@ def _get_next_block_id(self, rule):
...
+        self.progress_store.update_section_status(
+            CompletionStatus.IN_PROGRESS, this_location.section_id
+        )
+
+    def _remove_current_blocks_answers_for_new_backwards_routing(
+        self, rules, this_location, answer_ids_for_current_block
+    ):
+        rules = self.schema.get_operands(rules)
+        for rule in rules:
+            if isinstance(rule, dict):
```
berroar, line 236: Can merge the nested if here:
```suggestion
            if isinstance(rule, dict) and (
                "identifier" in rule
                and rule["identifier"] in answer_ids_for_current_block
            ):
```

berroar (approved):
Tested and both old and new are working functionally 👍

app/questionnaire/path_finder.py
```diff
@@ -204,15 +207,42 @@ def _get_next_block_id(self, rule):
...
-            for condition in goto_rule["when"]:
-                if "meta" not in condition.keys():
-                    self.answer_store.remove_answer(condition["id"])
+    def _remove_current_blocks_answers_for_backwards_routing(
+        self, rules, this_location
+    ):
+        answer_ids_for_current_block = self.schema.get_answer_ids_for_block(
+            this_location.block_id
+        )
+        if "when" in rules.keys():
```
parimooa, line 216: Don't need to use `keys`, it is faster without using it 
```suggestion
        if "when" in rules:
```

app/questionnaire/path_finder.py
```diff
@@ -204,15 +207,42 @@ def _get_next_block_id(self, rule):
...
+                    if "id" in rule and rule["id"] in answer_ids_for_current_block:
+                        self.answer_store.remove_answer(rule["id"])
 
-        self.progress_store.remove_completed_location(location=this_location)
+        self.progress_store.remove_location_for_backwards_routing(this_location)
+        self.progress_store.update_section_status(
+            CompletionStatus.IN_PROGRESS, this_location.section_id
+        )
+
+    def _remove_current_blocks_answers_for_new_backwards_routing(
```
parimooa, line 231: missing type hints

app/questionnaire/path_finder.py
```diff
@@ -204,15 +207,42 @@ def _get_next_block_id(self, rule):
...
 
-    def _remove_rule_answers(self, goto_rule, this_location):
-        # We're jumping backwards, so need to delete all answers from which
-        # route is derived. Need to filter out conditions that don't use answers
-        if "when" in goto_rule.keys():
-            for condition in goto_rule["when"]:
-                if "meta" not in condition.keys():
-                    self.answer_store.remove_answer(condition["id"])
+    def _remove_current_blocks_answers_for_backwards_routing(
+        self, rules, this_location
```
parimooa, line 211: missing type hints

parimooa (approved):
Tested it as per description , it seems to work. few minor comments

PR #752, opened 2022-01-10
Fix spacing on list collectors

rmccar:
Extra space being added to list collectors with and without items.
- For ones without items an empty summary is being added which is removed in this PR
- For ones with items a double margin is being added (to the top of the summary and bottom of the question) which is also removed in this PR

This PR has also fixed a inconsistency in the size of the indentation in the sectionsummary.html file 

### How to review 
- Using the test_list_collector schema compare the differences between master and this branch and see that the gaps have been reduced in both instances

### Checklist

### DIFF
--- a/templates/partials/question.html
+++ b/templates/partials/question.html
@@ -93,9 +93,11 @@
 }) %}
   {%- if content.list -%}
     {% set list = content.list %}
-    <div class="ons-u-mb-m">
-       {% include 'partials/summary/list-summary.html' %}
-    </div>
+    {%- if content.list.list_items | length > 0 -%}
+      <div class="ons-u-mb-m">
+        {% include 'partials/summary/list-summary.html' %}
+      </div>
+    {% endif %}
   {% endif %}
   {{ question_warning }}
   {{ individual_response_guidance }}
--- a/templates/partials/question.html
+++ b/templates/partials/question.html
@@ -91,13 +91,11 @@
   "instruction": question_instruction,
   "legendIsQuestionTitle": should_wrap_with_fieldset(question)
 }) %}
-  {%- if content.list -%}
+  {%- if content.list and content.list.list_items | length > 0 -%}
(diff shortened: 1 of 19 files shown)


rmccar:
[Fix spacing on list collectors](https://trello.com/c/jgnc6TQP/5057-fix-spacing-on-list-collectors)

petechd (approved):

templates/partials/question.html
```diff
@@ -93,9 +93,11 @@
 }) %}
   {%- if content.list -%}
     {% set list = content.list %}
-    <div class="ons-u-mb-m">
-       {% include 'partials/summary/list-summary.html' %}
-    </div>
+    {%- if content.list.list_items | length > 0 -%}
```
LJBabbage, line 96: Not sure why you would have 2 separate conditions, I don't see the benefit of the set if there are no list_items. Empty list items are also considered false so you could just have this
```  
{%- if content.list and content.list.list_items -%}
   {% set list = content.list %}
    <div class="ons-u-mb-m">
    </div>
{% endif %}
```

LJBabbage (changes requested):

templates/partials/question.html
rmccar: Yeah I agree, I've fixed that. I originally tried having one condition in a slightly different way and it fell over but that seems to work

templates/partials/question.html
```diff
@@ -91,10 +91,10 @@
   "instruction": question_instruction,
   "legendIsQuestionTitle": should_wrap_with_fieldset(question)
 }) %}
-  {%- if content.list -%}
+  {%- if content.list and content.list.list_items and content.list.list_items | length > 0 -%}
```
LJBabbage, line 94: Not sure why you have checked for the existence of content.list.list_items, then its length? content.list.list_items would just return false if empty so you don't need the length check

PR #753, opened 2022-01-10
Return to summary if new question is generated

LJBabbage:
If a Complete section is visited from the summary (section or final) and due to a change to that answer means more questions are generated and thus making the section in_progress, this update will take the respondent back to the summary once the extra questions have been answered

### How to review 
Using test_section_summary.json answer the first question with both and fill in the address, there is a skip on the 3rd question so you should end up on the section summary. Now click on the change your answer link for the first question and change it to either of the other options, you will still get the address next (it's iteration 1), but if you click save and continue again, you now get another question, before going to the section summary. Repeat but do it all the way to the final summary before changing the answer, in this instance it should return to the final summary.

The key element when testing is the url, make sure on all pages after the change that the ?return_to=<summary-type> is being passed, we are effectively exercising the return_to functionality, so the only test that is really needed is the persistence of it from one block to the next.  

### Checklist


app/questionnaire/router.py
```diff
@@ -129,18 +128,18 @@ def get_next_location_url(
             ):
                 return return_to_url
 
-            if is_last_block_in_section:
-                return self._get_next_location_url_for_last_block_in_section(location)
+            return self._get_next_location_url_for_complete_section(location)
 
         # Due to backwards routing, you can be on the last block without the section being complete
+        is_last_block_in_section = routing_path[-1] == location.block_id
```
LJBabbage, line 134: This could be removed, wasn't convinced routing_path[-1] == location.block_id was obvious what it was doing, the var helps.

schemas/test/en/test_list_change_evaluates_sections.json
```diff
@@ -27,6 +27,18 @@
     },
     "sections": [
         {
+            "summary": {
```
LJBabbage, line 30: All sections should now have summaries now, this one didn't. So it was removed the test as you can now just can't skip through a section not changing anything when complete or add a summary to allow a value to be changed

schemas/test/en/test_repeating_sections_with_hub_and_spoke.json
```diff
@@ -27,6 +27,25 @@
     },
     "sections": [
         {
+            "summary": {
```
LJBabbage, line 30: As above

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
```diff
@@ -155,27 +156,19 @@ describe("Feature: Repeating Sections with Hub and Spoke", () => {
 
       // Start section for first visitor
       $(HubPage.summaryRowLink("section")).click();
-      $(PrimaryPersonPage.submit()).click();
```
LJBabbage, line 158: Behaviour has changed, once a section is complete you will now need to use the section summary to navigate, you can't linearly step from one block to the next if that section was already completed

tests/functional/wdio.conf.js
```diff
@@ -52,7 +52,7 @@ exports.config = {
       "goog:chromeOptions": {
         args: [
           process.env.EQ_RUN_FUNCTIONAL_TESTS_HEADLESS ? "--headless" : "--start-maximized",
-          "--window-size=1280,1090",
+          "--window-size=1280,1100",
```
LJBabbage, line 55: Had to up this again, looks like if an element is outside the immediate window it doesn't work

jamesnunn:
I've tested the behaviour as described and the query parameter is populated as expected

tests/app/questionnaire/test_router.py
```diff
@@ -510,6 +547,31 @@ def test_return_to_final_summary_questionnaire_is_complete(self):
...
+        )
+        routing_path = RoutingPath(
+            ["radio", "dessert", "dessert-confirmation", "numbers"],
+            section_id="default-section",
+        )
+        next_location = self.router.get_next_location_url(
+            current_location, routing_path, return_to="final-summary"
+        )
+
+        self.assertIn("/questionnaire/numbers/?return_to=final-summary", next_location)
```
MebinAbraham, line 573: I'm a bit torn on these, although these test are correct and perfectly valid, I don't think it is enough to validate that the query string is persisted. These tests are essentially unit testing that, if you provide a `return_to` to `get_next_location_url`, it behaves as expected rather than it is persisted between requests. 

If the block handler was not passed the `return_to`, these unit tests would still pass, but the URL wouldn't have the query string at run time. Therefore, we should have an integration/functional test to assert that it is indeed persisted between requests. I couldn't see any integration/function tests for it.

app/questionnaire/router.py
```diff
@@ -346,13 +345,16 @@ def _is_section_enabled(self, section: Mapping) -> bool:
         )
 
     @staticmethod
-    def get_next_block_url(location: Location, routing_path: RoutingPath) -> str:
+    def get_next_block_url(
+        location: Location, routing_path: RoutingPath, **kwargs: str
+    ) -> str:
```
MebinAbraham, line 350: We also need to support the persistence of the query param for `get_previous_location_url`. At the moment, if you open up say 3 new questions, if you click `Previous` the query string is lost and when you subsequently go forward and complete the new questions, we won't jump back to the summary. 

Have confirmed with Donna that even if a user clicks `Previous`, soon as the section is re-complete, we should jump back to where we came from. We already have `return_to` in scope for `get_previous_location_url`, it is just not persisted.

MebinAbraham (commented):
Looking good. I'm sure there are probably some gotchas with the new behaviours that we'll find further down the line.

schemas/test/en/test_list_change_evaluates_sections.json
LJBabbage: Agreed, this pattern doesn't really lend itself to social current, and I do wonder about relationships as they naturally don't have summaries at present.

tests/app/questionnaire/test_router.py
LJBabbage: I had a similar thought pattern, but came down on the side of not having them, as it was harnessing something that we already had (i.e the return_to ). However it was only by a nano smidgen so will stick some extra tests in

app/questionnaire/router.py
LJBabbage: True, shouldn't be difficult

schemas/test/en/test_section_summary.json
```diff
@@ -135,6 +135,38 @@
                             ]
                         }
                     ]
+                },
+                {
+                    "id": "address-listed",
```
LJBabbage, line 140: This has been added so there is now a block between the new question that would appear on a routing change and the section summary. With the new question being last in the section, it could have been routing normally to the section summary rather than using the return_to and as such failing

MebinAbraham (approved):
👍

jamesnunn (approved):

PR #754, opened 2022-01-10
Introduce timeout modal

petechd:
This adds new [timeout modal element](https://ons-design-system.netlify.app/components/timeout-modal/) to Runner as per requirement described [on this card](https://trello.com/c/k5ET9O3C). It is based on previous work by @LJBabbage and @jamesnunn (adding session expiry endpoint and extending session from PATCH/GET). New test schema and functional test was added as well as previous `timeout.json` schema test had to be removed because of present timeout behaviour. Some test had to be modified because of the way they select elements of body of the document.

### How to review 
To test manually change the value of `EQ_SESSION_TIMEOUT_SECONDS` to a smaller number, remember that `onsTimeoutModal` has a `"showModalTimeInSeconds"` field that further customises the behaviour of the element.
Go to `http://localhost:5000/session-expiry/` to check current expiry time.
Check new functional test using `test_timeout_modal.json` schema that has "session_timeout_in_seconds" field.

### Checklist


petechd:
[Timeout modal (3 SP)](https://trello.com/c/k5ET9O3C/4498-timeout-modal-3-sp)

jamesnunn:
I've given this a try and find the following behaviour, happy to walk through the steps I'm taking which might indeed prove me testing the wrong thing.

Run runner with `EQ_SESSION_TIMEOUT_SECONDS=65 make run`
Launch a survey and wait 5 seconds for the modal to appear

Things that reset the timer in the modal window:
    Lose focus of tab and refocusing
    Lose focus of browser window and refocusing
    Clicking 'Continue survey'

However, resetting the timer with any of these methods doesn't appear to update the expiry time, and regardless of the what the timer shows in the modal, the browser signs out at the time indicated at the expiry time endpoint, it is never extended.

tests/functional/spec/timeout.spec.js
```diff
@@ -1,33 +1,21 @@
...
     browser.openQuestionnaire("test_timeout.json");
   });
 
-  it("Given I am completing an electronic questionnaire, when I have been inactive for the timeout period and attempt to submit data, then I will be redirected to a page confirming my session has timed out", () => {
-    browser.pause(6000);
-    $(TimeoutBlockPage.submit()).click();
-    const expectedPageTitle = browser.getTitle();
-    expect(expectedPageTitle).to.equal("Page is not available - Timeout test");
-    assertTimeoutText();
-  });
```
MebinAbraham, line 24: Is this tested via integration test?. If not, we need to as this is still a valid scenario when JS is disabled.

schemas/test/en/test_timeout.json
```diff
@@ -4,7 +4,7 @@
     "schema_version": "0.0.1",
     "data_version": "0.0.3",
     "survey_id": "0",
-    "session_timeout_in_seconds": 5,
+    "session_timeout_in_seconds": 120,
```
MebinAbraham, line 7: Does it not just need to be slightly above 60?.

tests/functional/spec/timeout.spec.js
MebinAbraham, line 24: Is this tested via integration test?. If not, we need to ensure it is covered, as this is still a valid scenario when JS is disabled.

schemas/test/en/test_timeout.json
MebinAbraham, line 7: Can this not be slightly above 60?.

pricem14pc:
Can we double check/confirm the Cloud Armor policy config in each env (sandbox/formal) that handles our public endpoints to make sure the new endpoints aren't going to get blocked.

templates/layouts/_questionnaire.html
```diff
@@ -39,4 +42,22 @@
     {% endif %}
   {% endblock %}
 
+  {{
+    onsTimeoutModal({
+      "showModalTimeInSeconds": 60,
+      "serverSessionExpiryEndpoint": "/session-expiry",
+      "sessionExpiresAt": "",
+      "redirectUrl": "/signed-out",
```
MebinAbraham, line 50: We should use `url_for` to fetch this and avoid hardcoding.

templates/layouts/_questionnaire.html
```diff
@@ -39,4 +42,22 @@
     {% endif %}
   {% endblock %}
 
+  {{
+    onsTimeoutModal({
+      "showModalTimeInSeconds": 60,
+      "serverSessionExpiryEndpoint": "/session-expiry",
+      "sessionExpiresAt": "",
```
MebinAbraham, line 49: The expires at will need to be exposed to the templates and will need passing in here.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,11 @@
+describe("Timeout Modal", () => {
+  beforeEach("Open Survey", () => {
+    browser.openQuestionnaire("test_timeout_modal.json");
+  });
+
+  it("Given I am completing the survey, when session time out is set to same amount as the showModalTimeInSeconds of timeout modal I will be able to see timeout modal with the option to extend the session instantly", () => {
```
MebinAbraham, line 6: The focus of the test here should be that, there is only 60 seconds until session expiry. We may want to set it to like 75 seconds and ensure there is no popup until we hit the 60-second mark. Since it's client side, you will have to be estimating the time left as it won't be an exact science.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,11 @@
...
+    expect($("body").getHTML()).to.include(
+      '<p class="ons-js-timeout-timer" aria-hidden="true" aria-relevant="additions">To protect your information, your progress will be saved and you will be signed out in <span class="ons-u-fw-b">1 minute</span>'
```
MebinAbraham, line 8: Since we have JS available, can we not assert the actual modal and it's contents instead of the HTML?.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,11 @@
+describe("Timeout Modal", () => {
```
MebinAbraham, line 1: I think we should cover the happy paths, such as:
- clicking `Continue` extends the session.
- ensure it works on post submission pages

tests/functional/spec/timeout_modal.spec.js
MebinAbraham: To properly test, the timeout needs to be like at least > 120 seconds, as we only store the updated session if it has been at least 60 seconds. Otherwise, if you were to use say 75 second, then after 15 seconds you would get the pop-up modal, but you can't extend the session as the session is only 15 seconds old. So clicking `Continue` on the modal would only work when the countdown is at like 14 seconds left, at which point our session is now >60 seconds old.

schemas/test/en/test_timeout.json
petechd: Yeah it can, I was not sure about the value I should use.

templates/layouts/_questionnaire.html
petechd: So I made it part of cookies as "expires_at" and pass it to this bit of template as part of `session`. It gets updated whenever session is extended. Thanks!

templates/layouts/_questionnaire.html
petechd: `url_for` now used for both elements, so no hardcoded endpoints, thanks!

petechd:
> I've given this a try and find the following behaviour, happy to walk through the steps I'm taking which might indeed prove me testing the wrong thing.
> 
> Run runner with `EQ_SESSION_TIMEOUT_SECONDS=65 make run` Launch a survey and wait 5 seconds for the modal to appear
> 
> Things that reset the timer in the modal window:
> 
> * Lose focus of tab and refocusing
> * Lose focus of browser window and refocusing
> * Clicking 'Continue survey'
> 
> However, resetting the timer with any of these methods doesn't appear to update the expiry time, and regardless of the what the timer shows in the modal, the browser signs out at the time indicated at the expiry time endpoint, it is never extended.
> 
> Clicking 'Continue survey' doesn't cause the modal to disappear.

Is this because you can only extend the session after a minute or more (since it was last extended/initialised)?

tests/functional/spec/timeout.spec.js
petechd: I think it is already tested in`/tests/integration/session/test_timeout.py`, even the contents: `self.assertInBody("Sorry, you need to sign in again")`

tests/functional/spec/timeout_modal.spec.js
petechd: Added tests of `Continue` button, modal on post submission page and extending session when the page receives focus (by returning from another tab). Thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: Changed to assert the actual modal elements after adding base page template. Thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: I've extended the timeout to exactly 120 seconds, which works in new functional tests, all the scenarios we discussed are covered (`Continue` button, post submission, page back in focus).

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    it("I will be able to see the timeout modal with the option to extend the session after 60 seconds", () => {
+      checkTimeoutModal();
+    });
+  });
+
+  describe("Given I am completing the survey, when I click on “Continue survey” button of the timeout modal", () => {
+    it("I will be able to extend the session", () => {
+      checkTimeoutModal();
+      $(TimeoutModalPage.submit()).click();
+      expect($(TimeoutModalPage.timer()).getText()).to.equal("");
```
MebinAbraham, line 21: We should assert that:
- the modal is not displayed, then wait like 15 second and ensure it is we are still on the same page (i.e redirect didn't happen)
- to ensure it did actually extend the session, assert that if we refresh, we are still on the same page (i.e it hasn't timed us out in the background).

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    });
+  });
+  describe("Given I am completing the survey, when I open a new window and then focus on the modal window ", () => {
+    it("I will extend the session", () => {
+      checkTimeoutModal();
+      browser.newWindow("");
+      browser.switchWindow("http://localhost:5000/questionnaire/timeout-modal-interstitial/");
+      expect($(TimeoutModalPage.timer()).getText()).to.equal(
+        "To protect your information, your progress will be saved and you will be signed out in 1 minute."
+      );
```
MebinAbraham, line 31: This check if the modal is displayed. As above, we need to check it has actually extended as expected.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,47 @@
+import TimeoutInterstitialPage from "../generated_pages/timeout_modal/timeout-modal-interstitial.page";
```
MebinAbraham, line 1: Need one to capture, the modal displaying an us not doing anything to ensure after the countdown, the modal is not displayed and that we have been redirected to the expected page.

templates/layouts/_base.html
```diff
@@ -139,3 +140,25 @@
...
 {% endblock %}
+
+{% block preFooter %}
+  {% if session_expires_at %}
+    {{
+      onsTimeoutModal({
+        "showModalTimeInSeconds": 60,
+        "serverSessionExpiryEndpoint": url_for('session.session_expiry'),
+        "sessionExpiresAt": session_expires_at,
+        "redirectUrl": url_for('session.get_signed_out'),
```
MebinAbraham, line 151: Need to confirm with Donna/Sarah, but I think the content on this page needs updating. This didn't match the content you get on a 401.html, which includes the bullet point about being timed out etc.

MebinAbraham (commented):
For visibility: Outside the scope of this PR and more related to the modal itself. But we do need to surface any valid ones.
- It looks like the modal is making two requests each time instead of one?. I am seeing two patch requests concurrently.
- Also, I was under the impression, user activity didn't extend the session when the modal is displayed. I had behaviours where the modal did disappear when I was moving the mouse.
- I will double check this, but I seem to get the modal before I expected it, almost like I didn't extend the session.

schemas/test/en/test_timeout.json
```diff
@@ -4,7 +4,7 @@
     "schema_version": "0.0.1",
     "data_version": "0.0.3",
     "survey_id": "0",
-    "session_timeout_in_seconds": 5,
+    "session_timeout_in_seconds": 70,
```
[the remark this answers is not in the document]
petechd: Changed to 5, added mention about non JS test, removed the spec file, it will be tested now as part of `schemas/test/en/test_timeout_modal.json` now, thanks!

schemas/test/en/test_timeout_modal.json
```diff
@@ -0,0 +1,58 @@
...
+                    "id": "timeout-group",
+                    "title": "Timeout modal",
+                    "blocks": [
+                        {
+                            "id": "timeout-modal-interstitial",
+                            "content": {
+                                "title": "Timeout modal interstitial",
+                                "contents": [
+                                    {
+                                        "description": "Timeout modal will appear on this page after 60 seconds as the session timeout is set to 125 seconds, pressing “Continue survey” button on the modal will extend the session by that amount, “showModalTimeInSeconds” field is set to 60"
```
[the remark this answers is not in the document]
petechd: Changed the wording to 65 and 125, thanks!

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,47 @@
...
+
+  describe("Given I am completing the survey, when I click on “Continue survey” button of the timeout modal", () => {
+    it("I will be able to extend the session", () => {
+      checkTimeoutModal();
+      $(TimeoutModalPage.submit()).click();
+      expect($(TimeoutModalPage.timer()).getText()).to.equal("");
+    });
+  });
+  describe("Given I am completing the survey, when I open a new window and then focus on the modal window ", () => {
+    it("I will extend the session", () => {
```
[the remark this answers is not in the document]
petechd: Changed the wording to `It will` in each instance, thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: So we ensure now that redirect didn't happen (waiting 15 seconds, had to extend the timeout to 80000). I don't think we can refresh the window using webdriver.

tests/functional/spec/timeout_modal.spec.js
petechd: So I added a 10s wait to check if the session got extended (we assert if we didn't get redirected).

tests/functional/spec/timeout_modal.spec.js
petechd: So I used the 5s session expiry schema for this, otherwise we would have to set timeout to more than 125 seconds.

templates/layouts/_base.html
petechd: Checked with Donna and it's the text we've been given, so no changes.

MebinAbraham: Have clarified this should go to `session.get_session_expired` which has the right content since it more that it has expired vs they signed out.

petechd: Ok, refactored the test to reflect the changes, thanks!

jamesnunn (approved):
Retested with the suggested settings, behaves as expected

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,69 @@
...
+import TimeoutSubmitPage from "../generated_pages/timeout_modal/submit.page";
+import { TimeoutModalPage } from "../base_pages/timeout-modal.page.js";
+
+describe("Timeout", () => {
+  describe("Given I am completing the survey, when the session timeout is set to 5 seconds", () => {
+    before(() => {
+      browser.openQuestionnaire("test_timeout.json");
+    });
+
+    it("It will redirect to signed-out page after session expires", () => {
```
MebinAbraham, line 11: We should update all these test to `Given` `When` `Then`.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      expect($(TimeoutModalPage.timer()).getText()).to.equal("");
+      browser.pause(15000);
+      expect(browser.getUrl()).to.contain(TimeoutInterstitialPage.pageName);
+    });
+  });
+  describe("Given I am completing the survey, when I open a new window and then focus on the modal window", () => {
+    it("It will extend the session", () => {
+      checkTimeoutModal();
+      browser.newWindow("");
+      browser.switchWindow("http://localhost:5000/questionnaire/timeout-modal-interstitial/");
```
MebinAbraham, line 51: We should be able to get the url using the page object?.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      browser.pause(10000);
+      expect(browser.getUrl()).to.contain(TimeoutInterstitialPage.pageName);
+    });
+  });
+
+  describe("Given I am completing the survey, when I get to post submission page", () => {
+    it("It will make the timeout modal visible 60 seconds before session expiry", () => {
```
MebinAbraham, line 58: Could we look at repeating the entire test case for post submission?. As anything we test on question pages should also work post submission?.

MebinAbraham (commented):
Just some minor comments about the test

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+describe("Timeout", () => {
+  describe("Given I am completing the survey, when the session timeout is set to 5 seconds", () => {
+    before(() => {
+      browser.openQuestionnaire("test_timeout.json");
+    });
+
+    it("It will redirect to signed-out page after session expires", () => {
+      browser.pause(25000);
+      expect(browser.getUrl()).to.equal("http://localhost:5000/session-expired");
```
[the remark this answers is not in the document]
petechd: Got rid of the localhost and assert "to.contain("/session-expired/") now, thanks!

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,69 @@
...
+import { TimeoutModalPage } from "../base_pages/timeout-modal.page.js";
+
+describe("Timeout", () => {
+  describe("Given I am completing the survey, when the session timeout is set to 5 seconds", () => {
+    before(() => {
+      browser.openQuestionnaire("test_timeout.json");
+    });
+
+    it("It will redirect to signed-out page after session expires", () => {
+      browser.pause(25000);
```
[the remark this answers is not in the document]
petechd: I tried locally and this one is minimum leeway we need, even if we change schema expiry to 1s it still needs 10+s leeway.

tests/functional/spec/timeout_modal.spec.js
petechd: Updated to use `Given`, `When`, `Then`. I used one of the possible styles we already have present in tests, I hope it's the one you wanted.

tests/functional/spec/timeout_modal.spec.js
petechd: Using `.pageName` for the switch now, which miraculously works!

tests/functional/spec/timeout_modal.spec.js
petechd: Repeated the whole test case for post submission, thanks!

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,62 @@
...
+      browser.openQuestionnaire("test_timeout_modal.json");
+      $(TimeoutInterstitialPage.submit()).click();
+      $(TimeoutSubmitPage.submit()).click();
+    });
+    testSet(ThankYouPage);
+  });
+});
+
+function testSet(page) {
+  it("When the session timeout is set to 125 seconds, Then it will make the timeout modal with the option to extend the session visible after 65 seconds and it will redirect after session expires", () => {
```
MebinAbraham, line 27: The timeout seconds is not important, and this is specially about the redirect, so maybe it should be:
- When the timeout modal is displayed, and I do not extend my session
- Then I will be redirected to the session expired page

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    checkTimeoutModal();
+    browser.pause(65000);
```
MebinAbraham, line 29: Worth adding a comment why. i.e We are waiting for the session to expire.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,62 @@
...
+        "Sorry, you need to sign in again",
+        "This is because you have either:",
+        "been inactive for 45 minutes and your session has timed out to protect your information",
+        "followed a link to a page you are not signed in to",
+        "followed a link to a survey that has already been submitted"
+      )
+      .to.not.include("To protect your information, your progress will be saved and you will be signed out in");
+  });
+
+  it("When I click “Continue survey” button of the timeout modal, Then it will extend the session and modal won‘t reappear in the next 15 seconds, no redirect will happen", () => {
```
MebinAbraham, line 42: - When the timeout modal is displayed, and I click the “Continue survey” button
- Then my session will be extended

The `modal won‘t reappear in the next 15 seconds, no redirect will happen` is just a sanity check from our end.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    checkTimeoutModal();
+    $(TimeoutModalPage.submit()).click();
+    expect($(TimeoutModalPage.timer()).getText()).to.equal("");
+    browser.pause(15000);
+    expect(browser.getUrl()).to.contain(page.pageName);
```
MebinAbraham, line 47: Maybe add the comment here about, waiting 15 seconds to sanity check it hasn't redirected shortly after.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  it("When I open a new window and then focus on the modal window, Then it will extend the session", () => {
+    checkTimeoutModal();
+    browser.newWindow("");
+    browser.switchWindow(page.pageName);
+    browser.pause(10000);
+    expect(browser.getUrl()).to.contain(page.pageName);
```
MebinAbraham, line 55: Same here. Comment about why, as it might not be clear when coming back to then test in future.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+  it("When I click “Continue survey” button of the timeout modal, Then it will extend the session and modal won‘t reappear in the next 15 seconds, no redirect will happen", () => {
+    checkTimeoutModal();
+    $(TimeoutModalPage.submit()).click();
+    expect($(TimeoutModalPage.timer()).getText()).to.equal("");
+    browser.pause(15000);
+    expect(browser.getUrl()).to.contain(page.pageName);
+  });
+
+  it("When I open a new window and then focus on the modal window, Then it will extend the session", () => {
```
MebinAbraham, line 50: Maybe?

- When the timeout modal is displayed, but I open a new window and then focus back on the timeout modal window
- Then my session will be extended

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    beforeEach(() => {
+      browser.openQuestionnaire("test_timeout_modal.json");
+      $(TimeoutInterstitialPage.submit()).click();
+      $(TimeoutSubmitPage.submit()).click();
+    });
+    testSet(ThankYouPage);
+  });
+});
+
+function testSet(page) {
```
MebinAbraham, line 26: testSet > testCases?. Also, could use arrow func.

tests/functional/spec/timeout_modal.spec.js
MebinAbraham, line 47: We will need to do a page refresh to ensure the session has been extended. If modal didn't extend session but disappeared, this would still page. So refresh the page and ensure the page doesn't say `"Sorry, you need to sign in again",`. So for the next time that asserts session is extended.

tests/functional/spec/timeout_modal.spec.js
petechd: Added, thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: Done, thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: Added, thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: Added a comment, thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: Comment added, thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: Reworded, thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: Changed to an arrow func and renamed, thanks!

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    expect($(TimeoutModalPage.timer()).getText()).to.equal("");
+    browser.pause(15000); // Waiting 15 seconds to sanity check it hasn't redirected
+    browser.refresh();
+    expect(browser.getUrl()).to.contain(page.pageName);
+    expect($("body").getHTML()).to.not.include(
+      "Sorry, you need to sign in again",
+      "This is because you have either:",
+      "been inactive for 45 minutes and your session has timed out to protect your information",
+      "followed a link to a page you are not signed in to",
+      "followed a link to a survey that has already been submitted"
```
MebinAbraham, line 54: I don't think this entire thing is needed, just the main title (Sorry, you need to sign in again). If you did the whole content and this broke and context was slightly different, it would pass.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    );
+  });
+
+  it("When the timeout modal is displayed, but I open a new window and then focus back on the timeout modal window, Then my session will be extended", () => {
+    checkTimeoutModal();
+    browser.newWindow("");
+    browser.switchWindow(page.pageName);
+    browser.pause(10000); // Waiting 10 seconds to sanity check it hasn't redirected
```
MebinAbraham, line 62: Need to test with the refresh here as well to ensure session has extended.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,70 @@
...
+        "followed a link to a survey that has already been submitted"
+      )
+      .to.not.include("To protect your information, your progress will be saved and you will be signed out in");
+  });
+
+  it("When the timeout modal is displayed, and I click the “Continue survey” button, Then my session will be extended", () => {
+    checkTimeoutModal();
+    $(TimeoutModalPage.submit()).click();
+    expect($(TimeoutModalPage.timer()).getText()).to.equal("");
+    browser.pause(15000); // Waiting 15 seconds to sanity check it hasn't redirected
```
MebinAbraham, line 46: With this pause and the one below, the pause has to be long enough to where the session would have expired. i.e timeout modal pops up at 60 seconds, if we then wait 15 second, we still have 45 seconds until the session times out hence we cannot know it extended. So we need to wait ~65 seconds, then refresh.

tests/functional/spec/timeout_modal.spec.js
petechd: Added, thanks!

tests/functional/spec/timeout_modal.spec.js
petechd: Added refresh step here, tests are passing!

tests/functional/spec/timeout_modal.spec.js
petechd: Extended this to 65 seconds now, luckily doesn't time out (our wdio.conf setting is 135 seconds).

tests/functional/spec/timeout_modal.spec.js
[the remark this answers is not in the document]
petechd: This is outdated and irrelevant now.

tests/functional/spec/timeout_modal.spec.js
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  it("When the timeout modal is displayed, but I open a new window and then focus back on the timeout modal window, Then my session will be extended", () => {
+    checkTimeoutModal();
+    browser.newWindow("");
+    browser.switchWindow(page.pageName);
+    browser.refresh();
+    browser.pause(65000); // Waiting 65 seconds to sanity check that it hasn’t expired
+    expect(browser.getUrl()).to.contain(page.pageName);
+  });
```
MebinAbraham, line 59: It's possible to set the timeout per spec, therefore we should revert the global mocha timeout to 60 seconds.

```suggestion
  }).timeout(140000);

  it("When the timeout modal is displayed, and I click the “Continue survey” button, Then my session will be extended", () => {
    checkTimeoutModal();
    $(TimeoutModalPage.submit()).click();
    expect($(TimeoutModalPage.timer()).getText()).to.equal("");
    browser.pause(65000); // Waiting 65 seconds to sanity check that it hasn’t expired
    browser.refresh();
    expect(browser.getUrl()).to.contain(page.pageName);
    expect($("body").getHTML()).to.not.include("Sorry, you need to sign in again");
  }).timeout(140000);

  it("When the timeout modal is displayed, but I open a new window and then focus back on the timeout modal window, Then my session will be extended", () => {
    checkTimeoutModal();
    browser.newWindow("");
    browser.switchWindow(page.pageName);
    browser.refresh();
    browser.pause(65000); // Waiting 65 seconds to sanity check that it hasn’t expired
    expect(browser.getUrl()).to.contain(page.pageName);
  }).timeout(140000);
```

PR #756, opened 2022-01-13
Add missing functional tests batch 1

petechd:
This is the first batch of missing functional tests, some of the schemas are currently unused i.e they are not referenced in any unit/integration or front end test.

### How to review 
Run `./node_modules/.bin/wdio tests/functional/wdio.conf.js --spec tests/functional/spec/difference.spec.js` and/or test manually.

### Checklist


petechd:
[Create functional tests batch 1 (2 SP)](https://trello.com/c/OEGuo7SR/5053-create-functional-tests-batch-1-2-sp)

petechd:
[Timeout modal (3 SP)](https://trello.com/c/k5ET9O3C/4498-timeout-modal-3-sp)

tests/functional/spec/difference.spec.js
```diff
@@ -0,0 +1,73 @@
+import AgeBlockYearPage from "../generated_pages/difference_in_years/age-block.page";
```
MebinAbraham, line 1: The spec name is not clear. I think we make a folder called placeholder and have placeholder related spec in there.

tests/functional/spec/difference.spec.js
```diff
@@ -0,0 +1,73 @@
...
+import AgeBlockMonthYearPage from "../generated_pages/difference_in_years_month_year/age-block.page";
+import AgeBlockDayMonthYearRangePage from "../generated_pages/difference_in_years_range/date-block.page";
+import AgeBlockMonthYearRangePage from "../generated_pages/difference_in_years_month_year_range/date-block.page";
+
+describe("Difference check (years)", () => {
+  before("Load the survey", () => {
+    browser.openQuestionnaire("test_difference_in_years.json");
```
MebinAbraham, line 8: I think while we are here, we should rename the schema to group them by feature like we do already. So i.e `test_placeholder_difference_in_years.json` or using actual name of the transform `test_placeholder_calculate_date_difference.json`

tests/functional/spec/difference.spec.js
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  it("Given a day, month and year answer is provided for a date question then the age in years should be calculated and displayed on the page ", () => {
+    $(AgeBlockYearPage.day()).setValue(1);
+    $(AgeBlockYearPage.month()).setValue(1);
+    $(AgeBlockYearPage.year()).setValue(1990);
+    $(AgeBlockYearPage.submit()).click();
+
+    const expectedPageTitle = browser.getTitle();
+    expect(expectedPageTitle).to.equal("You are … old. Is this correct? - Difference between two dates");
```
MebinAbraham, line 18: Any reason we are asserting the title?. We should assert the question title with the actual placeholder content.

tests/functional/spec/difference.spec.js
```diff
@@ -0,0 +1,73 @@
...
+
+  it("Given a month and year answers 'from' and 'to' are provided for a date question then the difference in years should be calculated and displayed on the page ", () => {
+    $(AgeBlockMonthYearRangePage.periodFromMonth()).setValue(1);
+    $(AgeBlockMonthYearRangePage.periodFromYear()).setValue(1990);
+    $(AgeBlockMonthYearRangePage.periodToMonth()).setValue(1);
+    $(AgeBlockMonthYearRangePage.periodToYear()).setValue(1991);
+
+    $(AgeBlockMonthYearRangePage.submit()).click();
+
+    expect($("body").getText()).to.have.string("You were out of the UK for 1 year. Is this correct?");
```
MebinAbraham, line 51: We should use our selector, we should have a selector to get the question title.

tests/functional/spec/difference.spec.js
petechd: The reason is we would have to change the assertion next year?

tests/functional/spec/difference.spec.js
petechd: Using selectors now, thanks!

tests/functional/spec/difference.spec.js
petechd: Changed the spec names so they are more specific and descriptive, created `placeholder` folder in `features`, moved placeholder schemas specs in there.

tests/functional/spec/difference.spec.js
petechd: Changed them to option 1, so `test_placeholder_difference_in_years.json` etc.

tests/functional/spec/difference.spec.js
MebinAbraham: We should work out the age in JS and assert that, otherwise the test is redundant as it does check the placeholder pipes in correct number.

petechd: I've added a function using Date API that generates the number of years, hope that solves this problem?

LJBabbage (approved):
Looks good to me

tests/functional/spec/difference.spec.js
MebinAbraham: Should be okay given we are only working with years difference. Worst case it fails once a year, so no biggie.

MebinAbraham (approved):

PR #757, opened 2022-01-13
Dynamic answer options driven by operator functions

MebinAbraham:
- Added support for dynamic answer options for Radio, Dropdowns and Checkbox, including mutually exclusive checkbox. This support is for dynamic options driven by operator functions such as `date-range` using the `map` operator. Subsequent cards will add support for answer and list sources.
- The field handlers for Radio/Dropdown/Checkbox now share the same base handler since they share common behaviour.
- The jinja filter config class has also been refactored slightly, now Radio/Checkbox/RelationshipRadio share the same config class. Dropdown doesn't share this class since the model required in the view is different to Radio/Checkbox.
- A new class has been created to handle the generation of the dynamic answer options. This is required both in the field handlers and the summary context.
- Two schemas are used to demo the behaviour, as it needs to with dynamic options on its own, as well as with additional static options.
- For unit tests, the schemas are shared to prevent a lot of duplication, although it makes it slightly less readable, I do think it is fine given it prevents a lot of duplication. https://github.com/ONSdigital/eq-questionnaire-runner/blob/6dd137e287bdeb2c3191caa74f0559003570c2bd/tests/app/forms/field_handlers/test_dropdown_handler.py#L27-L38
- For functional tests, to prevent duplicating the test cases for the two schema, the common spec is ran against both schemas. https://github.com/ONSdigital/eq-questionnaire-runner/blob/aeff47e1f6d92ca3ebc7cdd97a8f2cd389bb9c81/tests/functional/spec/features/dynamic_answer_options/function_driven.js#L8-L33

**Caveats**:
- Will only work with data version 0.0.3. Data version 0.0.1 is out of scope.
- Dynamic options cannot be used with Relationships answer type
- Backwards journey is not supported. This means if you change a previously selected dynamic option, the dependant questions aren't cleared from the answer store and progress is not updated. This will be done when we come to handle answer dependencies.

### How to review 
Use the PR commits to help breakdown the reviews. Each commit is a complete piece of work and don't overlap with the other commits. There are not a lot of code changes, most are schema/tests.

- Ensure the changes make sense and all required use cases are covered.

### Checklist


app/forms/field_handlers/select_handlers.py
```diff
@@ -1,13 +1,60 @@
...
+    def _build_dynamic_choices(self) -> list[ChoiceWithDetailAnswer]:
+        if not self.dynamic_options_schema:
+            return []
+
+        dynamic_options = DynamicAnswerOptions(
+            dynamic_options_schema=self.dynamic_options_schema,
+            rule_evaluator=self.rule_evaluator,
+        )
+
+        choices = [
```
petechd, line 37: Would it make it unreadable if we remove `choices` and do immediate return?
```python
return [
            ChoiceWithDetailAnswer(option["value"], option["label"], None)
            for option in dynamic_options.evaluate()
        ]
```

schemas/test/en/test_dynamic_answer_options_function_driven.json
```diff
@@ -0,0 +1,347 @@
...
+                                                    "EEEE d MMMM yyyy"
+                                                ]
+                                            }
+                                        }
+                                    },
+                                    {
+                                        "id": "dynamic-mutually-exclusive-static-answer",
+                                        "mandatory": false,
+                                        "type": "Checkbox",
+                                        "label": "OOr",
```
petechd, line 330: Typo? Do we even need label here?

schemas/test/en/test_dynamic_answer_options_function_driven_with_static_options.json
```diff
@@ -0,0 +1,371 @@
...
+                                                "label": "None of the above",
+                                                "value": "None of the above"
+                                            }
+                                        ]
+                                    },
+                                    {
+                                        "id": "dynamic-mutually-exclusive-static-answer",
+                                        "mandatory": false,
+                                        "type": "Checkbox",
+                                        "label": "OOr",
```
petechd, line 354: Same here, this label can be removed.

app/jinja_filters.py
```diff
@@ -349,7 +323,9 @@ def __init__(self, detail_answer_field, detail_answer_schema):
 def map_checkbox_config(form, answer):
     options = form["fields"][answer["id"]]
 
-    return [CheckboxConfig(option, i, form, answer) for i, option in enumerate(options)]
+    return [
+        SelectConfig(option, idx, answer, form) for idx, option in enumerate(options)
```
petechd, line 327: Do we want to keep it consistent and call it `index` instead of `idx` since SelectConfig takes `index` as an argument? Same on line 341.

app/jinja_filters.py
```diff
@@ -247,66 +247,38 @@ def __init__(self, _for, text, description=None):
...
         self.value = option.data
         self.checked = option.checked
 
         label_description = None
-        answer_option = answer["options"][index]
 
-        if answer_option and "description" in answer_option:
-            label_description = answer_option["description"]
+        try:
+            self._answer_option = answer.get("options", [])[index]
```
LJBabbage, line 260: This is interesting. I might need an example, is it trying to do 2 things at once. 
1. If options doesn't exist if defaults to an empty list which will fall over because it has no index 
2. If there are options, but an index is supplied which doesn't exist it also falls over

Also interested why we then assign an empty dict

app/questionnaire/dynamic_answer_options.py
```diff
@@ -0,0 +1,34 @@
...
+
+@dataclass
+class DynamicAnswerOptions:
+    dynamic_options_schema: Mapping[str, Any]
+    rule_evaluator: RuleEvaluator
+
+    def evaluate(self) -> tuple[dict[str, str], ...]:
+        values = self.dynamic_options_schema["values"]
+        if "source" in values:  # pylint: disable=no-else-raise
+            # :TODO: Implement value sources support
```
LJBabbage, line 16: Never a fan of TODO, rather have a story

tests/app/forms/field_handlers/conftest.py
```diff
@@ -2,45 +2,142 @@
...
+def get_mock_response_metadata():
+    return {"started_at": "2021-01-01T09:00:00.220038+00:00"}
+
+
+def get_mock_location():
+    return None
+
+
+def get_mock_list_item_id():
+    return None
```
LJBabbage, line 43: Slightly odd, would probably have just done this inline. Or at least make the mock more representative, i.e get_mock_location_none.

tests/functional/spec/features/dynamic_answer_options/function_driven.js
```diff
@@ -0,0 +1,205 @@
...
+        it("When I click a selected option, then it should be deselected", () => {
+          for (let i = 0; i < testCase.answerOptionCount; i++) {
+            $(DynamicCheckboxPage.answerByIndex(i)).click();
+            expect($(DynamicCheckboxPage.answerByIndex(i)).isSelected()).to.be.false;
+          }
+        });
+
+        it("When I submit the page, then I should be taken to the next page", () => {
+          $(DynamicCheckboxPage.submit()).click();
+          expect(browser.getUrl()).to.contain(DynamicRadioPage.pageName);
```
LJBabbage, line 57: Is this needed, isn't this just selecting and deselecting a checkbox? I see a few others below as well.

LJBabbage (changes requested):
This looks good and works as expected. As we are combining dynamic and static options, we do need to check validation at some stage. I think all of your questions bar the feeding one were optional, would like to see a mandatory one with test for each. Bar that I can see very little else, although I think we can slim line the functional tests a touch

app/forms/field_handlers/select_handlers.py
MebinAbraham: I don't think it does, so done 👍

schemas/test/en/test_dynamic_answer_options_function_driven.json
MebinAbraham: Removed

schemas/test/en/test_dynamic_answer_options_function_driven_with_static_options.json
MebinAbraham: Removed

app/jinja_filters.py
MebinAbraham: Updated

app/jinja_filters.py
MebinAbraham: In either case, we want to get the item based on an index, which won't exist for dynamic answers. 

The empty dict assignment is so that the type of the variable is a single type (dict) but nothing stopping us making it an optional dict and assigning None. I'm easy either way.

app/questionnaire/dynamic_answer_options.py
MebinAbraham: We do have a story, but I think it's fine personally, as it gives an indication to the next dev where to start. Also, most IDE's support `TODO` and a good way to seen what's missing. Not precious about it, but I do think it is fine.

tests/app/forms/field_handlers/conftest.py
MebinAbraham: Was to ensure the rule evaluator and value source resolver stayed in sync, but have now removed these and made it inline for all those that return a simple default value.

tests/functional/spec/features/dynamic_answer_options/function_driven.js
MebinAbraham: This was to ensure the dynamic options were in fact rendered properly and that I am able to select/deselect them. If I were to only select the first option, we can't guarantee that the rest of the dynamic options were rendered properly.

MebinAbraham:
> As we are combining dynamic and static options, we do need to check validation at some stage. I think all of your questions bar the feeding one were optional, would like to see a mandatory one with test.

Added a new schema for mandatory dynamic options along with the necessary functional tests.

tests/functional/spec/features/dynamic_answer_options/function_driven.js
LJBabbage: Still think it's a bit overkill, but not going to argue

LJBabbage (approved):

app/forms/field_handlers/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 from app.questionnaire.value_source_resolver import ValueSourceResolver
 
+from ...questionnaire.rules.rule_evaluator import RuleEvaluator
```
jamesnunn, line 5: Any reason not to use absolute import here?

app/forms/field_handlers/number_handler.py
```diff
@@ -84,6 +50,20 @@ def get_field_references(
...
         }
 
+    @cached_property
+    def max_decimals(self) -> int:
+        return self.answer_schema.get("decimal_places", 0)
+
+    def get_field(self) -> Union[DecimalField, IntegerField]:
+        field_type = (
+            DecimalFieldWithSeparator
+            if self.max_decimals > 0
```
jamesnunn, line 60: Very minor but I'd suggest a property that evaluates this, with the same logic but called something like `is_decimal_field`

app/forms/field_handlers/__init__.py
MebinAbraham: Think IDE auto imported following the rest below, which are all relative. Have changed all to be absolute.

app/forms/field_handlers/number_handler.py
MebinAbraham: Was existing, hence I didn't bother with any refactor here, but have changed while I'm here 👍

petechd (approved):

jamesnunn (approved):

app/views/contexts/question.py
```diff
@@ -18,7 +18,7 @@ def build_question_context(rendered_block, form):
         answer_ids.append(answer["id"])
 
         if answer["type"] in ("Checkbox", "Radio"):
```
parimooa, line 20: for my understanding why there is no `Dropdown` here ?

app/views/contexts/summary/block.py
```diff
@@ -21,14 +23,24 @@ def __init__(
         self.title = block_schema.get("title")
         self.number = block_schema.get("number")
         self.link = self._build_link(block_schema["id"], return_to)
+
+        self._rule_evaluator = RuleEvaluator(
```
parimooa, line 27: Missing type hints

app/jinja_filters.py
```diff
@@ -383,21 +363,21 @@ def map_relationships_config_processor():
...
+        self.value, self.text = option.value, option.label
         self.selected = select.data == self.value
         self.disabled = self.value == "" and select.flags.required
 
 
 @blueprint.app_template_filter()
-def map_select_config(select):
-    return [SelectOptionConfig(choice, select) for choice in select.choices]
+def map_dropdown_config(select):
+    return [DropdownConfig(choice, select) for choice in select.choices]
```
parimooa, line 375: choices have `value=''` for `label= select an answer` , that is not one of the options/choices to select from  , why is it under choices ?

parimooa (changes requested):
Manually tested works functionally ok, have few minor comments

app/jinja_filters.py
```diff
@@ -361,7 +337,9 @@ def map_checkbox_config_processor():
 def map_radio_config(form, answer):
     options = form["fields"][answer["id"]]
 
-    return [RadioConfig(option, i, form, answer) for i, option in enumerate(options)]
+    return [
```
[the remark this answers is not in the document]
MebinAbraham: Yup, have refactored to remove the duplication.

app/views/contexts/question.py
MebinAbraham: This is for types that support the details answer options which is essentially a nested field within a radio/checkbox option, which dropdowns don't support.
image

app/views/contexts/summary/block.py
MebinAbraham: Not enforced in this file, so won't be validated.

app/jinja_filters.py
MebinAbraham: If I understood your question correctly, then that is cause dropdown have a default value to chose from which is `disabled`, it's more of a placeholder.

parimooa (approved):

PR #760, opened 2022-01-17
Update Authentication tests to pytest

LJBabbage:
This PR updates the tests in the folder Authentication from unittest to pytest

### How to review 
Make sure the new tests work and they make sense

### Checklist


tests/app/authentication/test_authenticator.py
```diff
@@ -1,167 +1,153 @@
...
-            period_str="period_str",
-            language_code=None,
-            launch_language_code=None,
-            survey_url=None,
-            ru_name="ru_name",
-            ru_ref="ru_ref",
-            case_id="case_id",
+
+
+def test_check_session_with_user_id_in_session(
```
MebinAbraham, line 10: I think this is only passing in actions as it doesn't have authenticated google credentials. I am not able to get this this and `test_check_session_with_no_user_id_in_session`  passing locally as it is trying to connect to an actual Datastore and Redis instances. Those are no patched here but are on master viahttps://github.com/ONSdigital/eq-questionnaire-runner/blob/master/tests/app/app_context_test_case.py#L51-L55

tests/app/authentication/test_authenticator.py
```diff
@@ -1,167 +1,153 @@
...
+    app,
+    session_store,
+    session_data,
+    expires_at,
+    mock_get_session_store,
```
MebinAbraham, line 15: Any reason not to do the patch of session store inline?. I find this confusing as we have 2 session stores here, but one is just to patch. But it does save a bit of duplication, but I think clarity is important. I think calling it `patch_get_session_store` might make it clearer on what it does.

MebinAbraham (commented):
Looks good. Just an issue with two tests that create the app and interact with the DB.

MebinAbraham (approved):

petechd (approved):

PR #761, opened 2022-01-18
Adds query param and anchor for return_to_answer_id

jamesnunn:
Adds query parameter `return_to_answer_id` to allow the appropriate anchor to be added to `Previous` links when editing answers from `section-summary`.

### How to review 
Launch `test_section_summary.json` and complete questions up to the first section summary. Check that all `Previous` links do not contain an anchor or `return_to` or `return_to_answer` query params.

From the first section summary, check that any/all `Change` links for answers provided contain: `?return_to=section-summary&return_to_answer_id=<appropriate-answer-id>#<appropriate-answer-id>`

Click `Change` on any answer and on the edit page, check that the `Previous` link contains `#<appropriate-answer-id>`.

Click `Previous` and check that the section summary scrolls to the relevant answer. You may need to shrink the window vertically to see this behaviour 

### Checklist


jamesnunn:
[Return to specific place on summary with 'Previous' link (3 SP)](https://trello.com/c/ZSS6YQuQ/5065-return-to-specific-place-on-summary-with-previous-link-3-sp)

jamesnunn:
Depends on https://github.com/ONSdigital/eq-questionnaire-runner/pull/762 being resolved

tests/functional/spec/features/section_summary/section_summary.spec.js
```diff
@@ -15,6 +15,12 @@ describe("Section Summary", () => {
       browser.openQuestionnaire("test_section_summary.json");
```
jamesnunn, line 15: Need to test scrolling behaviour after https://github.com/ONSdigital/eq-questionnaire-runner/pull/762

tests/functional/spec/features/anchor_scrolling.spec.js
```diff
@@ -0,0 +1,63 @@
...
+import InsuranceTypePage from "../../generated_pages/section_summary/insurance-type.page.js";
+import ListedPage from "../../generated_pages/section_summary/listed.page.js";
+import PropertyDetailsSummaryPage from "../../generated_pages/section_summary/property-details-section-summary.page.js";
+import HouseType from "../../generated_pages/section_summary/house-type.page.js";
+import HouseholdCountSectionSummaryPage from "../../generated_pages/section_summary/household-count-section-summary.page.js";
+import HouseholdDetailsSummaryPage from "../../generated_pages/section_summary/house-details-section-summary.page.js";
+import NumberOfPeoplePage from "../../generated_pages/section_summary/number-of-people.page.js";
+import SubmitPage from "../../generated_pages/section_summary/submit.page.js";
+
+describe("Section Summary", () => {
```
LJBabbage, line 11: Could probably benefit with a name change to the thing it's testing (i.e Summary anchor scrolling)

tests/functional/spec/features/anchor_scrolling.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  describe("Given I start a Test Section Summary survey", () => {
+    beforeEach(() => {
+      browser.openQuestionnaire("test_section_summary.json");
+      $(InsuranceTypePage.both()).click();
+      $(InsuranceTypePage.submit()).click();
+    });
+
+    it("When I have provided an answer and click through to the next question, the Previous button url shouldn't yet contain any anchors or reference to return_to or return_to_answer_id", () => {
```
LJBabbage, line 19: We usually use the pattern Given / When / Then, so these statements need a tweak

LJBabbage, line 19: Not sure the "yet" adds much and probably previous page link rather than button

tests/functional/spec/features/anchor_scrolling.spec.js
```diff
@@ -0,0 +1,63 @@
...
+      expect($(InsuranceAddressPage.previous()).getAttribute("href")).not.to.contain("#");
+      expect($(InsuranceAddressPage.previous()).getAttribute("href")).not.to.contain("return_to");
+      expect($(InsuranceAddressPage.previous()).getAttribute("href")).not.to.contain("return_to_answer_id");
+    });
+
+    it("When I reach the summary page, the Change button url should contain return_to, return_to_answer_id query params", () => {
```
LJBabbage, line 25: Probably needs to be section summary and change answer link

tests/functional/spec/features/anchor_scrolling.spec.js
```diff
@@ -0,0 +1,63 @@
...
+      $(InsuranceAddressPage.submit()).click();
+      $(ListedPage.submit()).click();
+      expect($(PropertyDetailsSummaryPage.insuranceAddressAnswerEdit()).getAttribute("href")).to.contain(
+        "/questionnaire/insurance-address/?return_to=section-summary&return_to_answer_id=insurance-address-answer#insurance-address-answer"
+      );
+    });
+
+    it("When I edit an answer from the summary page, the Previous button url should contain an anchor referencing the answer id of the answer I am changing and when clicked, the browser url contains the same", () => {
```
LJBabbage, line 33: I would prefer to avoid multiple whens and split into 2 tests.

app/views/contexts/summary/block.py
```diff
@@ -12,15 +13,18 @@ def __init__(
         list_store,
         metadata,
         response_metadata,
-        schema,
+        schema: QuestionnaireSchema,
```
LJBabbage, line 16: Any reason for type hinting for just schema? I was thinking none or all

LJBabbage (changes requested):
This works per the card, although I do think the functional tests need another look, they work but I do find them hard to follow and sometimes trying to do too much.

app/views/contexts/summary/block.py
jamesnunn: was just using it for getting hold of method names from `schema`, will remove

app/views/contexts/summary/block.py
```diff
@@ -12,15 +13,18 @@ def __init__(
...
     ):
         self.id = block_schema["id"]
         self.location = location
         self.title = block_schema.get("title")
         self.number = block_schema.get("number")
-        self.link = self._build_link(block_schema["id"], return_to)
+        first_answer_id_for_block = schema.get_answer_ids_for_block(self.id)[0]
+        self.link = self._build_link(
+            block_schema["id"], return_to, first_answer_id_for_block
+        )
```
MebinAbraham, line 27: Actually, this won't be quite right when questions have more than one answer. The `_build_link` method needs to be moved to `app/views/contexts/summary/answer.py`, as the link needs to depend on the specific answer. 

To cover this in a test, you just need to add two answers to one block and ensure, both change links have the appropriate answer ID for anchors/query params.

schemas/test/en/test_anchor_scrolling.json
```diff
@@ -0,0 +1,291 @@
+{
```
LJBabbage, line 1: I don't think we need to add a new schema here, we have a number that have multiple answers, test_numbers would be a good one to use. If you really did want to keep the schema, it should be stripped back to the basics of the test (i.e remove the skip conditions/extra blocks etc) The description would also need updating

jamesnunn: I don't think test_numbers would be suitable as there's no final summary. I'll reuse the one I have

app/views/contexts/summary/answer.py
```diff
@@ -17,4 +29,16 @@ def serialize(self):
             "unit": self.unit,
             "unit_length": self.unit_length,
             "currency": self.currency,
+            "link": self.link,
         }
+
+    def _build_link(self, block_id, list_name, list_item_id, return_to):
```
MebinAbraham, line 35: Minor but I think we should make these keyword only args, then it would be easy to follow and less chance of messing up the order.

app/views/contexts/summary/question.py
```diff
@@ -18,22 +27,43 @@ def __init__(
...
         self.rule_evaluator = rule_evaluator
 
-        self.answers = self._build_answers(answer_store, question_schema)
+        self.answers = self._build_answers(
+            answer_store,
+            question_schema,
+            block_id,
+            location.list_name if location else None,
+            return_to,
+        )
```
MebinAbraham, line 39: With many things being passed, keyword args would help with readability and to ensure order isn't wrong.

app/views/contexts/summary/question.py
```diff
@@ -44,7 +74,14 @@ def _build_answers(self, answer_store, question_schema):
...
             )
 
-            summary_answer = Answer(answer_schema, answer).serialize()
+            summary_answer = Answer(
+                answer_schema,
+                answer,
+                block_id,
+                list_name,
+                self.list_item_id,
+                return_to,
```
MebinAbraham, line 83: Same here, I think it would be beneficial to use keyword positional arguments.

schemas/test/en/test_anchor_scrolling.json
MebinAbraham: Had a look at the latest iteration, and I'm still not convinced, we need this. Can we not reuse, `test_section_summary`, this seems almost identical?. You just need to do the same edit and add an extra answer in, which shouldn't cause much issue with existing tests?.

tests/app/views/contexts/summary/test_answer.py
```diff
@@ -1,29 +1,39 @@
...
+        )
 
     def test_date_answer_type(self):
         # Given
         answer_schema = {"id": "answer-id", "label": "", "type": "date"}
         user_answer = None
 
         # When
-        answer = Answer(answer_schema, user_answer)
+        answer = Answer(answer_schema, user_answer, "1", "2", "3", "section-summary")
```
MebinAbraham, line 36: Input should be more meaningful. Keyword arg would help also.

tests/app/views/contexts/summary/test_question.py
```diff
@@ -78,8 +78,10 @@ def test_create_question(self):
             question_schema,
             answer_store=self.answer_store,
             schema=self.schema,
-            list_item_id=None,
             rule_evaluator=self.get_rule_evaluator(),
+            location=None,
+            block_id="",
```
MebinAbraham, line 83: These should be realistic, as block ID would never not exist at this point. Same for the rest.

tests/functional/spec/features/anchor_scrolling.spec.js
```diff
@@ -0,0 +1,71 @@
...
+      expect($(InsuranceAddressPage.previous()).getAttribute("href")).not.to.contain("#");
+      expect($(InsuranceAddressPage.previous()).getAttribute("href")).not.to.contain("return_to");
+      expect($(InsuranceAddressPage.previous()).getAttribute("href")).not.to.contain("return_to_answer_id");
+    });
+
+    it("When I reach the section summary page, Then the Change link url should contain return_to, return_to_answer_id query params", () => {
+      $(InsuranceAddressPage.submit()).click();
+      $(AddressDurationPage.submit()).click();
+      expect($(PropertyDetailsSummaryPage.insuranceAddressAnswer2Edit()).getAttribute("href")).to.contain(
+        "/questionnaire/insurance-address/?return_to=section-summary&return_to_answer_id=insurance-address-answer2#insurance-address-answer2"
```
MebinAbraham, line 27: I think for these tests, the asserts could be slimmed down, to just check everything from `?`, as the URL leading to it isn't important. That should be tested already, if not should be separate which ensures the previous link is working. This to me is testing the query param + anchor.

tests/app/views/contexts/summary/test_answer.py
```diff
@@ -1,29 +1,39 @@
...
+            "section-summary",
+        )
 
         # Then
         self.assertEqual(answer.id, "answer-id")
         self.assertEqual(answer.label, "Answer Label")
         self.assertEqual(answer.value, user_answer)
+        self.assertEqual(
+            answer.link,
+            "http://test.localdomain/questionnaire/house-type/?return_to=section-summary&return_to_answer_id=answer-id#answer-id",
```
MebinAbraham, line 27: I think it's worth having this test separate and ensuring it works fine with a list name as well. I don't think we had a test for that.

schemas/test/en/test_anchor_scrolling.json
jamesnunn: It is almost identical, but it obviously would mean the tests include a lot of elements aren't needed for the test. This way it's at least streamlined.

tests/app/views/contexts/summary/test_answer.py
jamesnunn: I'll include all elements and test the order of them all

schemas/test/en/test_anchor_scrolling.json
MebinAbraham: Okay fair enough.

app/questionnaire/router.py
```diff
@@ -356,9 +372,12 @@ def get_next_block_url(
...
     @staticmethod
-    def _get_section_url(location: Location) -> str:
+    def _get_section_url(
+        location: Location, return_to_answer_id: Optional[str] = None
+    ) -> str:
         return url_for(
             "questionnaire.get_section",
             section_id=location.section_id,
             list_item_id=location.list_item_id,
+            _anchor=return_to_answer_id,
```
MebinAbraham, line 382: Section URL does use the anchor. i.e we are never at the hub with an anchor.

schemas/test/en/test_anchor_scrolling.json
```diff
@@ -0,0 +1,195 @@
...
```
MebinAbraham, line 1: I'm not sure about the schema name. Maybe something like `test_submit_with_summary_return_to_anchor`

tests/app/views/contexts/summary/test_question.py
```diff
@@ -78,8 +78,10 @@ def test_create_question(self):
             question_schema,
             answer_store=self.answer_store,
             schema=self.schema,
-            list_item_id=None,
             rule_evaluator=self.get_rule_evaluator(),
+            location=None,
+            block_id="house-type",
+            return_to="",
```
MebinAbraham, line 84: Would be None if it didn't exist?.

tests/functional/spec/features/anchor_scrolling.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      expect($(InsuranceAddressPage.previous()).getAttribute("href")).to.contain("/questionnaire/sections/property-details-section/#insurance-address-answer2");
+    });
+
+    it("When I edit an answer from the section summary page and click the Previous link, Then the browser url should contain an anchor referencing the answer id of the answer I am changing", () => {
+      $(InsuranceAddressPage.submit()).click();
+      $(AddressDurationPage.submit()).click();
+      $(PropertyDetailsSummaryPage.insuranceAddressAnswer2Edit()).click();
+      $(InsuranceAddressPage.previous()).click();
+      expect(browser.getUrl()).to.contain("/questionnaire/sections/property-details-section/#insurance-address-answer2");
+    });
```
MebinAbraham, line 44: I think we should test this for the submit button click also, as it should both have the same behaviour.

MebinAbraham (commented):
Last few minor issues. Otherwise LGTM.

app/questionnaire/router.py
MebinAbraham: As discussed, this is fine, I got confused between this same method in hub.py

tests/functional/spec/features/submit_with_summary_return_to_answer.spec.js
```diff
@@ -0,0 +1,77 @@
...
+
+    it("When I edit an answer from the section summary page and click the Previous link, Then the browser url should contain an anchor referencing the answer id of the answer I am changing", () => {
+      $(InsuranceAddressPage.submit()).click();
+      $(AddressDurationPage.submit()).click();
+      $(PropertyDetailsSummaryPage.insuranceAddressAnswer2Edit()).click();
+      $(InsuranceAddressPage.previous()).click();
+      expect(browser.getUrl()).to.contain("property-details-section/#insurance-address-answer2");
+    });
+
+    it("When I edit an answer from the section summary page and click the Submit button, Then the browser url should contain an anchor referencing the answer id of the answer I am changing", () => {
```
MebinAbraham, line 46: Maybe for these to, the `Then` should be like `Then I am taken to the summary and the browser url contains ...`

app/views/contexts/summary/question.py
```diff
@@ -18,22 +27,43 @@ def __init__(
...
     def _get_answer(self, answer_store, answer_id):
         return answer_store.get_escaped_answer_value(answer_id, self.list_item_id)
 
-    def _build_answers(self, answer_store, question_schema):
+    def _build_answers(
+        self, *, answer_store, question_schema, block_id, list_name, return_to
+    ):
 
         if self.summary:
+            answer_id = f"{self.id}-concatenated-answer"
```
MebinAbraham, line 49: I think this needs a functional test to ensure anchor uses this. As it is explicitly different to other answer types. It does work functionally, but I don't think we have any tests to validate this?

tests/functional/spec/features/submit_with_summary_return_to_answer.spec.js
```diff
@@ -43,15 +53,15 @@ describe("Summary Anchor Scrolling", () => {
       expect(browser.getUrl()).to.contain("property-details-section/#insurance-address-answer2");
     });
 
-    it("When I edit an answer from the section summary page and click the Submit button, Then the browser url should contain an anchor referencing the answer id of the answer I am changing", () => {
+    it("When I edit an answer from the section summary page and click the Submit button, Then I am take to the summary page and the browser url should contain an anchor referencing the answer id of the answer I am changing", () => {
```
MebinAbraham, line 56: `take` > `taken`

MebinAbraham (approved):

LJBabbage (approved):

PR #762, opened 2022-01-18
Update to DS version 45.1.4 and pass in new summary ids

rmccar:
This PR will update Runner to DS version 45.1.4 which some other pieces of work require.

Also as part of this ChromeDriver needed updating to get functional tests to pass and the `titleTag` param no longer needs to be set to `H2` because that is now default so I have removed some instances where this is set.

### How to review 
Tests pass and eQ works as expected

### Checklist


rmccar:
[Introduce anchor link to summary page tables (2 SP)](https://trello.com/c/aNrekImJ/5064-introduce-anchor-link-to-summary-page-tables-2-sp)

app/jinja_filters.py
```diff
@@ -453,9 +453,11 @@ def __init__(  # noqa: C901, R0912 pylint: disable=too-complex, too-many-branche
             and answer["label"]
         ):
             self.rowTitle = answer["label"]
+            self.id = answer["id"] + "-label"
             self.rowTitleAttributes = {"data-qa": answer["id"] + "-label"}
         else:
             self.rowTitle = strip_tags(question["title"])
+            self.id = question["id"] + "-row-item"
```
MebinAbraham, line 460: The IDs shouldn't be suffixed. Also in both cases, we want to use the answer id but we only really need to set the answer id, if it is a row that can be edited, so:
```py
        if edit_link_text:
            self.id = answer["id"]
```

MebinAbraham (changes requested):
The jinja filter test needs updating to ensure `id` is present in the output. Integration test can be done by the feature that uses this (so James' PR)

EDIT: Ignore, just seen the changes, didn't spot it.

LJBabbage (approved):

jamesnunn (approved):

MebinAbraham (approved):

PR #765, opened 2022-01-20
Assign response_id from RAS claims

parimooa:
Use `response_id` from RAS , if `response_id` is not in claims, create response_id by concatenating metadata fields `ru_ref`,`collection_exercise_sid`,`eq_id` and `form_type`

### How to review 
On launcher
**Scenario 1**

1. Choose business schema
2. Remove 'response_id' data ( keep it blank)
3. Observe that `metadata['response_id']` is concatenation of `ru_ref`,`collection_exercise_sid`,`eq_id` and `form_type`

**Scenario 2** 

1. Choose business schema
2. Note down 'response_id' value 
3. Observe that `metadata['response_id'] `is  same as in step 2

### Checklist


app/utilities/metadata_parser.py
```diff
@@ -146,6 +146,27 @@ def convert_schema_name(self, data, **kwargs):
             )
         return data
 
+    @post_load
+    def metadata_response_id(
```
LJBabbage, line 150: Don't think we need the metadata bit, probably needs to be something like update_response_id (although I am not 100% on that but can't think of anything else at present)

app/utilities/metadata_parser.py
```diff
@@ -146,6 +146,27 @@ def convert_schema_name(self, data, **kwargs):
...
+        self, data, **kwargs
+    ):  # pylint: disable=no-self-use, unused-argument
+        """
+        If response_id is not present : Build response_id from ru_ref,collection_exercise_sid,eq_id and form_type
```
LJBabbage, line 154: Would swap these over for code flow as the first return is response_id is present

app/utilities/metadata_parser.py
```diff
@@ -146,6 +146,27 @@ def convert_schema_name(self, data, **kwargs):
...
+        If response_id is present : return as it is
+
+        """
+        if data.get("response_id"):
+            return data
+        ru_ref = data.get("ru_ref")
+        collection_exercise_sid = data.get("collection_exercise_sid")
+        eq_id = data.get("eq_id")
+        form_type = data.get("form_type")
+        if ru_ref and collection_exercise_sid and eq_id and form_type:
```
LJBabbage, line 165: Would be half tempted to move the 2 optional ones first (ie. eq_id and form_type). Also could consider walrus operators, although the extra brackets make it look a big fugly

tests/app/parser/test_metadata_parser.py
```diff
@@ -177,3 +177,20 @@ def test_business_params_without_schema_name(fake_business_metadata_runner):
     claims = validate_runner_claims(fake_business_metadata_runner)
 
     assert claims["schema_name"] == "mbs_0253"
+
+
+def test_when_response_id_is_missing(fake_business_metadata_runner):
```
LJBabbage, line 182: Maybe a test for when eq_id/form_type is missing as well

app/utilities/metadata_parser.py
MebinAbraham, line 165: This should raise a `ValidationError` for the `else` case.

app/utilities/metadata_parser.py
MebinAbraham: Maybe `convert_response_id` to keep it consistent with the schema_name one. Or `update` but should also then update `convert_schema_name` to `update_schema_name` for consistency. I don't mind either, as long as it is consistent.

app/utilities/metadata_parser.py
parimooa: moved optional up , don't think walrus operator is going to do much ere

parimooa: It is not needed `marshmallow` handles it as part of `validation`

MebinAbraham: Could also just check you got `eq_id` and `form_type`, the other two are mandatory and must exist anyway.

LJBabbage: I agree with Walrus I removed the comment as the brackets made it look weird

MebinAbraham: Although true, I don't think that is explicit and replies on the order and the fact that `schema_name` is optional. If schema_name became mandatory, then our test wouldn't fail, but we may not have eq_id/form_type.

parimooa: The else part doesn't get executed , it is caught before that in [here](https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/utilities/metadata_parser.py#L131-L133)

parimooa: ignore , above one. i tweaked test to make sure `else` part `ValidationError` gets raised

MebinAbraham: Yes, I understand that, but that solely dependents on the logic for `schema_name`, if the schema name existed but eq_id or form_type didn't, this wouldn't raise a validation error.

app/utilities/metadata_parser.py
```diff
@@ -146,6 +146,30 @@ def convert_schema_name(self, data, **kwargs):
...
+                                        and updates metadata with response_id
+
+
+        """
+        if data.get("response_id"):
+            return data
+        eq_id = data.get("eq_id")
+        form_type = data.get("form_type")
+        ru_ref = data.get("ru_ref")
+        collection_exercise_sid = data.get("collection_exercise_sid")
```
LJBabbage, line 165: Very minor but these could be inside the if now.

LJBabbage (approved):

app/utilities/metadata_parser.py
```diff
@@ -146,6 +146,30 @@ def convert_schema_name(self, data, **kwargs):
...
+        if eq_id and form_type:
+            response_id = f"{ru_ref}{collection_exercise_sid}{eq_id}{form_type}"
+            data["response_id"] = response_id
+            return data
+
+        raise ValidationError("Either eq_id or form_type is missing")
```
MebinAbraham, line 171: Minor, can we quote the claim name like the schema name one. Also, we should say why we need it as it is useful for debugging and upstream. As in, it is missing but required when response_id isn't provided. Something like that?.

app/utilities/metadata_parser.py
MebinAbraham, line 165: Minor, and non-blocker but these could be bracket access since they are mandatory.
```py
ru_ref = data.["ru_ref"[
collection_exercise_sid = data["collection_exercise_sid"]]
```

tests/app/parser/test_metadata_parser.py
```diff
@@ -177,3 +177,31 @@ def test_business_params_without_schema_name(fake_business_metadata_runner):
...
+    expected = (
+        f"{fake_business_metadata_runner['ru_ref']}"
+        f"{fake_business_metadata_runner['collection_exercise_sid']}"
+        f"{fake_business_metadata_runner['eq_id']}"
+        f"{fake_business_metadata_runner['form_type']}"
+    )
+    del fake_business_metadata_runner["response_id"]
```
MebinAbraham, line 189: We should do a version of this where it is set to a falsely value like `""`, upstream have done this in the past and I guess is why Mark asked us to guard key not existing or empty value.

app/utilities/metadata_parser.py
```diff
@@ -146,6 +146,32 @@ def convert_schema_name(self, data, **kwargs):
...
+    ):  # pylint: disable=no-self-use, unused-argument
+        """
+        If response_id is present : return as it is
+        If response_id is not present : Build response_id from ru_ref,collection_exercise_sid,eq_id and form_type
+                                        and updates metadata with response_id
+
+
+        """
+        if data.get("response_id"):
+            return data
```
MebinAbraham, line 161: We should add a log line here to say something like "respose_id exists in claim, skipping generation". That way, it is easy to view in logs which path it took.

MebinAbraham (commented):
Looks good, working as expected. Minor suggestion about logging.

MebinAbraham (approved):

app/utilities/metadata_parser.py
pricem14pc: But don't write out the `response_id` anywhere in the log. Even though this is now going to be using known values in the interim, we should continue to treat it as a secret.

PR #767, opened 2022-01-21
Fix period limits when minimum not set

LJBabbage:
This PR fixes period limits when minimum is not set. We were checking an object which could be None and therefore it was falling over. Our code just needed a tweak and min_offset = self._get_offset_value moved inside the initial part of the if

### How to review 
Make sure I have fixed it properly. You could lift the json from the test and put it in a schema (test_date_range is probably easiest) and try it if you like. You could also try lcree_0010.json if you want to use the schema that fell over because of it. It will be available in launcher (assuming you have .schemas-version >=v3.0.2)

### Checklist

### DIFF
--- a/app/forms/questionnaire_form.py
+++ b/app/forms/questionnaire_form.py
@@ -203,14 +203,11 @@ class QuestionnaireForm(FlaskForm):
         period_limits_item = self._get_period_limits(period_limits)
         period_min = period_limits_item[0]
 
-        if period_min:
-            min_offset = self._get_offset_value(period_min)
+        # Exception to be raised if range available is smaller than minimum range allowed
+        if period_min and period_range < self._get_offset_value(period_min):
+            exception = f"The schema has invalid period_limits for {question_id}"
 
-            # Exception to be raised if range available is smaller than minimum range allowed
-            if period_range < min_offset:
-                exception = f"The schema has invalid period_limits for {question_id}"
-
-                raise Exception(exception)
+            raise Exception(exception)
 
     @staticmethod
     def validate_date_range_with_single_date_limits(
--- a/app/forms/questionnaire_form.py
+++ b/app/forms/questionnaire_form.py
@@ -200,15 +200,17 @@ class QuestionnaireForm(FlaskForm):
         period_range: timedelta,
     ) -> None:
         # Get period_limits from question
-        period_limits_item: tuple = self._get_period_limits(period_limits)
-        period_min: Mapping[str, int] = period_limits_item[0]
-        min_offset = self._get_offset_value(period_min)
+        period_limits_item = self._get_period_limits(period_limits)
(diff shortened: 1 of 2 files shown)


app/forms/questionnaire_form.py
```diff
@@ -200,15 +200,17 @@ def validate_date_range_with_period_limits_and_single_date_limits(
...
-            exception = f"The schema has invalid period_limits for {question_id}"
+        if period_min:
+            min_offset = self._get_offset_value(period_min)
 
-            raise Exception(exception)
+            # Exception to be raised if range available is smaller than minimum range allowed
+            if period_range < min_offset:
+                exception = f"The schema has invalid period_limits for {question_id}"
+
+                raise Exception(exception)
```
MebinAbraham, line 213: Minor, but the check could be done inline to avoid the nesting. The value isn't reused either.

```py
        # Exception to be raised if range available is smaller than minimum range allowed
        if period_min and period_range < self._get_offset_value(period_min):
            exception = f"The schema has invalid period_limits for {question_id}"

            raise Exception(exception)
```

MebinAbraham (approved):
👍

app/forms/questionnaire_form.py
LJBabbage: Thought about it, but went the other way, just for readability, but equally not bothered. Next reviewer wins, thumbs up/down

tests/app/forms/test_questionnaire_form.py
```diff
@@ -881,6 +881,60 @@ def test_invalid_maximum_period_limit_and_single_date_periods(self):
...
+                    "date-range-to-year": "2018",
+                }
+            )
+
+            form = generate_form(
+                schema,
+                question_schema,
+                AnswerStore(),
+                ListStore(),
+                metadata=None,
```
parimooa, line 926: is not `{}` instead of `None`

app/forms/questionnaire_form.py
```diff
@@ -200,15 +200,17 @@ def validate_date_range_with_period_limits_and_single_date_limits(
...
 
-        # Exception to be raised if range available is smaller than minimum range allowed
-        if period_min and period_range < min_offset:
-            exception = f"The schema has invalid period_limits for {question_id}"
+        if period_min:
+            min_offset = self._get_offset_value(period_min)
 
-            raise Exception(exception)
+            # Exception to be raised if range available is smaller than minimum range allowed
+            if period_range < min_offset:
```
parimooa: What if they are equal ?

parimooa (commented):
Looks good , few minor comments

app/forms/questionnaire_form.py
LJBabbage: Its the minimum allowed so is fine as doesn't need an exception

parimooa (approved):

PR #768, opened 2022-01-21
Add links to the header

MebinAbraham:
- Added the `My account` and `Sign out` links to the header
- Refactored the use of `account_service_url`. `account_service_url` from the claims is now assigned as `account_service_base_url` and usages of the existing `acount_service_url` in the templates are now `account_service_log_out_url`

### How to review 
Ensure:
- links in the header are working as expected
- existing links on pages such as session expired still are as expected

### Checklist


app/survey_config/survey_config.py
```diff
@@ -41,3 +39,22 @@ class SurveyConfig:
...
+    sign_out_button_text: str = lazy_gettext("Save and sign out")
+    contact_us_url: str = field(init=False)
+    cookie_settings_url: str = field(init=False)
+    privacy_and_data_protection_url: str = field(init=False)
+
+    def __post_init__(self):
+        self.contact_us_url: str = f"{self.base_url}/contact-us/"
+        self.cookie_settings_url: str = f"{self.base_url}/cookies/"
+        self.privacy_and_data_protection_url: str = (
+            f"{self.base_url}/privacy-and-data-protection/"
```
LJBabbage, line 51: :thinking:

tests/functional/spec/save_sign_out.spec.js
```diff
@@ -51,18 +52,12 @@ describe("SaveSignOut", () => {
     expect($(IntroThankYouPagePage.exitButton()).isExisting()).to.be.false;
   });
 
-  it("Given a logout url is not set, when I navigate the questionnaire, then I see the correct sign out buttons", () => {
-    browser.openQuestionnaire("test_introduction.json", { includeLogoutUrl: false });
+  it("Given a Census questionnaire, when I navigate the questionnaire, then I see the correct sign out buttons", () => {
```
LJBabbage, line 55: Updated to Census?

tests/integration/session/test_sign_out_and_exit.py
```diff
@@ -72,17 +70,31 @@ def _launch_and_submit_questionnaire(self, schema, **kwargs):
...
+        self.assertInRedirect(ACCOUNT_SERVICE_LOG_OUT_URL)
 
-    def test_redirects_to_account_service_log_out_url_when_present(self):
-        self._launch_and_submit_questionnaire(
-            schema="test_new_hub_section_required_and_enabled",
-            account_service_log_out_url=ACCOUNT_SERVICE_LOG_OUT_URL,
+
+class TestCensusSignOut(IntegrationTestCase):
+    def setUp(self):
+        # Perform setup steps
```
LJBabbage, line 84: I hope everyone inside a setup knows there are steps 😆 . I don't think we need the comment

LJBabbage (commented):
Looks good, bar an odd comment. I however do not like the sign-out link and save and sign out button, which do the same thing, but have different text and sit a cm away from each other. Very odd pattern. However I realise it is nothing to do with the story

app/survey_config/business_config.py
```diff
@@ -36,3 +42,21 @@ def __post_init__(self):
...
+                HeaderLink(
+                    lazy_gettext("My account"),
+                    self.account_service_my_account_url,
+                    id="header-link-my-account",
+                ).__dict__,
+                HeaderLink(
+                    lazy_gettext("Sign out"), sign_out_url, id="header-link-sign-out"
+                ).__dict__,
+            ]
+            if is_authenticated
```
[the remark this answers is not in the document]
MebinAbraham: Ya possibly. It depends on if we want the ability to have any service links when not authenticated. Not need at the moment but would allow that. I'm easy either way, so happy to change if we think it's better to check is_authenticated in template helpers.

app/survey_config/survey_config.py
MebinAbraham: Due to https://surveydatacollection.slack.com/archives/CLR17KQKU/p1642770634085600

tests/functional/spec/save_sign_out.spec.js
MebinAbraham: Ya, previously the button text depended on us getting a log out url in the claims, which was specific to census and no longer the case any more. It is now done by survey type. It should say `complete later` for surveys where the user does not have an account, so social/census.

tests/integration/session/test_sign_out_and_exit.py
MebinAbraham: Copy and pasted existing test 🤷

app/survey_config/business_config.py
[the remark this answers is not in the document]
LJBabbage: I can't visualise any situation where we would want to provide service links to anyone not authenticated, and think IA would see it that way as well. As this card is high priority I won't block.

app/helpers/template_helpers.py
```diff
@@ -139,29 +146,39 @@ def _footer_warning(self) -> Optional[str]:
...
         "health": SurveyConfig,
         "social": SurveyConfig,
         "northernireland": SurveyConfig,
         "census": (WelshCensusSurveyConfig if language == "cy" else CensusSurveyConfig),
         "census-nisra": CensusNISRASurveyConfig,
-    }[theme]()
+    }
+
+    return survey_type_to_config[theme](
+        base_url=base_url,
```
petechd, line 161: Is this `base_url` argument needed if `SurveyConfig` class takes `base_url` from env var?

app/survey_config/business_config.py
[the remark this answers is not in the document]
MebinAbraham: Ya, depends on if they ever want things like `Exit` links.

app/helpers/template_helpers.py
MebinAbraham: It needs to be passed in from the value we get in the claims, as the env var is a fallback default for when we don't have a cookie.

petechd: Got it, thanks!

parimooa: This doesn't seem right , it is being flagged in pycharm as _unused argument_

parimooa (commented):
I noticed **my account** link goes to surveys  goes to _https://surveys.ons.gov.uk/sign-in/_ where as **`sign-out`** is  _https://localhost:8000/sign-in/logout_ , shouldn't the latter one be _https://surveys.ons.gov.uk/sign-in/logout_ , why is base url different ? in Trello card i can see that base url should be same

app/survey_config/business_config.py
```diff
@@ -36,3 +42,21 @@ def __post_init__(self):
...
+    ) -> Optional[list[dict]]:
+        return (
+            [
+                HeaderLink(
+                    lazy_gettext("My account"),
+                    self.account_service_my_account_url,
+                    id="header-link-my-account",
+                ).__dict__,
+                HeaderLink(
+                    lazy_gettext("Sign out"), sign_out_url, id="header-link-sign-out"
```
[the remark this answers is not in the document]
MebinAbraham: Nah, it should be our sign out endpoint, which will handle the logic on where it should go. As we need to sign them out first.

MebinAbraham:
> I noticed **my account** link goes to surveys goes to _[surveys.ons.gov.uk/sign-in](https://surveys.ons.gov.uk/sign-in/)_ where as **`sign-out`** is _[localhost:8000/sign-in/logout](https://localhost:8000/sign-in/logout)_ , shouldn't the latter one be _[surveys.ons.gov.uk/sign-in/logout](https://surveys.ons.gov.uk/sign-in/logout)_ , why is base url different ? in Trello card i can see that base url should be same

Good spot. 👍 It wasn't passing in the base URL in render templates call of `get_survey_config`. Have fixed the issue in the latest commit.

LJBabbage (approved):

parimooa (approved):

petechd (approved):

PR #769, opened 2022-01-24
Use wdio test suites for front end tests

berroar:

In order to improve the speed of our functional tests following the addition of the new timeout modal feature, I have split the functional tests into 4 suites: `features`, `components`, `general` and `timeout`.

This has reduced functional test running time from ~24m to ~17m.

### How to review 
Check all tests are still being run and are still passing.

### Checklist


berroar:
[Speed up runner’s functional tests (3 SP)](https://trello.com/c/eQKZ6beh/5082-speed-up-runners-functional-tests-3-sp)

MebinAbraham (commented):
So this is still running synchronously. The speed-up we see is because the timeout modal tests runs first and uses only 1 instance (since they are in the same suite), so not both instances are used by it, hence other tests can still run while timeout modal test is running. The improvement is good, however, what we wanted to do was actually run multiple instances of functional tests, so 2 lots. So the first lot could run just the timeout modal, the 2nd lot could run the rest. So you would have 2 run functional test jobs in actions.

The simplest approach is essentially duplicate most but just changing the name of the suite, however I think we should look at https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs, where the variable would be the suite name, then it would be easy to break it down into 2 or even more runs without duplication.

berroar:
> So this is still running synchronously. The speed-up we see is because the timeout modal tests runs first and uses only 1 instance (since they are in the same suite), so not both instances are used by it, hence other tests can still run while timeout modal test is running. The improvement is good, however, what we wanted to do was actually run multiple instances of functional tests, so 2 lots. So the first lot could run just the timeout modal, the 2nd lot could run the rest. So you would have 2 run functional test jobs in actions.
> 
> The simplest approach is essentially duplicate most but just changing the name of the suite, however I think we should look at https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs, where the variable would be the suite name, then it would be easy to break it down into 2 or even more runs without duplication.

Split this out into a matrix now 👍

petechd (commented):
Tested this in a fork and functional test passes fine, just need merging into master to remove main `test-functional` task without the matrix, https://github.com/petechd/eq-questionnaire-runner/runs/4937702992?check_suite_focus=true

MebinAbraham (commented):
Just for visibility, as discussed, we should split the timeout spec into two files, currently it is not making use of the 2 instance hence the 16 min run time.

tests/functional/wdio.conf.js
```diff
@@ -17,6 +17,12 @@ exports.config = {
   // directory is where your package.json resides, so `wdio` will be called from there.
   //
   specs: ["./tests/functional/spec/**/*.js"],
+  suites: {
+    timeout_modal: ["./tests/functional/spec/timeout/*.spec.js"],
+    components: ["./tests/functional/spec/components/**/*.js"],
+    features: ["./tests/functional/spec/features/**/*.js"],
+    general: ["./tests/functional/spec/general/*.js"],
```
MebinAbraham, line 24: I'm not against this, but the only issue is if someone put a new spec into `/tests/functional/spec/new_test.spec.js`, it wouldn't get run, and we wouldn't realise?. Maybe general could just be `/tests/functional/spec/*.spec.js`. Same issue if someone adds a new folder, but that is very unlikely as the folder we have now covered pretty much everything.

MebinAbraham:
Runtime looks much better 💪. We can look to break out the unit tests in future into, maybe, unit tests/integrations tests. We would also get benefit from speeding up the docker compose, as that alone takes over 3 minutes.

tests/functional/spec/timeout/timeout_modal.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  checkTimeoutModal() {
+    $(TimeoutModalPage.timer()).waitForDisplayed({ timeout: 70000 });
+    expect($(TimeoutModalPage.timer()).getText()).to.equal("To protect your information, your progress will be saved and you will be signed out in 1 minute.");
+  }
+}
+
+export default TestCase;
+export const TimeoutModalTestCase = new TestCase();
```
MebinAbraham, line 47: Is the first line needed given you have the second export?

berroar: Removed now

MebinAbraham (commented):
Also worth adding to the functional test part of the readme about running suites.

README.md
```diff
@@ -221,6 +221,12 @@ To run a single test, add `.only` into the name of any `describe` or `it` functi
 `describe.only('Skip Conditions', function() {...}` or
 `it.only('Given this is a test', function() {...}`
 
+In order to speed up our functional tests running in Github Actions on pull request, the tests have been split out into 4 suites which can be run concurrently. These suites are configured in the `wdio.conf.js` file.
+An individual test suite can be run using:
+
+`./scripts/run_tests_functional.sh <suite_name>` or
+`./node_modules/.bin/wdio tests/functional/wdio.conf.js --suite <suite_name>`
```
MebinAbraham, line 228: Could just be `yarn test_functional --suite <suite>`?

README.md
```diff
@@ -221,6 +221,12 @@ To run a single test, add `.only` into the name of any `describe` or `it` functi
 `describe.only('Skip Conditions', function() {...}` or
 `it.only('Given this is a test', function() {...}`
 
+In order to speed up our functional tests running in Github Actions on pull request, the tests have been split out into 4 suites which can be run concurrently. These suites are configured in the `wdio.conf.js` file.
+An individual test suite can be run using:
```
MebinAbraham, line 225: I don't think the GitHub actions bit is relevant to document here. It thinks it just needs to state that we have suites and you can run them using etc.

petechd (approved):

tests/functional/wdio.conf.js
```diff
@@ -17,6 +17,12 @@ exports.config = {
   // directory is where your package.json resides, so `wdio` will be called from there.
   //
   specs: ["./tests/functional/spec/**/*.js"],
+  suites: {
+    timeout_modal: ["./tests/functional/spec/timeout/*.spec.js"],
+    components: ["./tests/functional/spec/components/**/*.js"],
+    features: ["./tests/functional/spec/features/**/*.js"],
```
MebinAbraham, line 23: Oddly, I didn't think this would pick up file within `features` but only folders, but it does seem to work fine with files.

MebinAbraham (commented):
The lint checks are now duplicated in the matrix, they prob should be moved that out of the matrix and run beforehand however, I am not against keeping them given they take less than 20 seconds to run. Thoughts?

tests/functional/wdio.conf.js
berroar: Yeah think this picks. up both files and folders 👍

.github/workflows/pull_request.yml
```diff
@@ -106,6 +106,9 @@ jobs:
       - name: Running schema tests
         run: ./scripts/validate_test_schemas.sh
   test-functional:
+    strategy:
+      matrix:
+        suite: [ timeout_modal, features, general, components ]
     needs: [python-dependencies, node-dependencies]
```
parimooa, line 112: Why does test-functional need python dependencies ?

parimooa (commented):
quite an improvement , nearly half the time , looks good

MebinAbraham:
> The lint checks are now duplicated in the matrix, they prob should be moved that out of the matrix and run beforehand however, I am not against keeping them given they take less than 20 seconds to run. Thoughts?

@parimooa Got an opinion?. I'm happy to leave it in.

.github/workflows/pull_request.yml
MebinAbraham: Cause the generates pages are created by `generate_pages.py`. One day it won't be Py 😄

parimooa: can we not generate it as `artifacts` in previous steps and share with following functionals tests . We dont need dependencies

MebinAbraham: We definitely could, but personally don't think that should block this though, as that is an improvement external to the scope of this. Also, the python dependencies are cached, I don't believe they get reinstalled at this point. But ya, we can look to do further clean up.

parimooa:
> The lint checks are now duplicated in the matrix, they prob should be moved that out of the matrix and run beforehand however, I am not against keeping them given they take less than 20 seconds to run. Thoughts?

20 seconds is lot :-)

.github/workflows/pull_request.yml
parimooa: it could save 2 mins of docker-compose , probably need a card to do that so this can go in . We need this urgently.

MebinAbraham: As discussed, these deps are for actions and not docker. But we can still look to improve from a best practice point of view separately. Once we move generating the pages out into artifacts, this can go.

parimooa (approved):

MebinAbraham (approved):

PR #772, opened 2022-01-24
Default submission language code for submission/feedback

MebinAbraham:
This PR ensures that for submission and feedback payload, `submission_language_code` is set to our default of `en` if upstream does not send us a `language_code`.

### How to review 
- Ensure the changes are accurate and that `submission_language_code` is always present and defaulted to `en` when missing from the launch claims.

### Checklist


MebinAbraham:
Will override submission language on launch in Session data.

MebinAbraham:
Session data is not available on flush but have plans to get rid of session data in favour of questionnaire store metadata in future, therefore, sticking with this option for now.

tests/app/views/handlers/test_submission_handler.py
```diff
@@ -43,6 +43,25 @@ def test_submission_language_code_in_payload(self):
             )
             assert submission_handler.get_payload()["submission_language_code"] == "cy"
 
+    @patch("app.views.handlers.submission.convert_answers", Mock(return_value={}))
+    def test_submission_language_code_uses_default_language_if_session_data_language_not_present(
```
LJBabbage, line 47: prefer **language_code_is_none** rather than language_not_present

LJBabbage (approved):

tests/app/submitter/test_converter.py
```diff
@@ -182,6 +182,7 @@ def test_converter_language_code_not_set_in_payload(
         fake_questionnaire_schema, fake_questionnaire_store, {}, SUBMITTED_AT
     )
 
+    assert fake_questionnaire_store.metadata.get("launch_language_code") is None
```
jamesnunn, line 185: This might not be worth worrying about, but this injects behaviour that you might not want - the presence of `launch_language_code` as `None` is different to it not existing and adding some defaulting behaviour in the test. Maybe:
```suggestion
    with pytest.raises('KeyError'):
        fake_questionnaire_store.metadata["launch_language_code"]
```

MebinAbraham: Updated, but also realised key should be `language_code` so have changed that.

jamesnunn (approved):

PR #775, opened 2022-01-25
Adds works-either-way `stripped_base_url` property to BusinessConfig

jamesnunn:
Fixes the incorrect `account_service_url` value by removing `/surveys/todo` if it exists in the url. It will work either way, so should still function correctly when the correct url is eventually provided without an urgent fix.

### How to review 
Launch a survey with the `account_service_url` changed to include `/surveys/todo` (e.g. `http://localhost:8000/surveys/todo`). Submit a survey and check that the `Back to surveys` url contains only one instance of `/surveys/todo` and not `/surveys/todo/surveys/todo`.

Do the same but without passing the extra `/surveys/todo` in `account_service_url` and check that the url again is correct, proving that it will just work when the url is eventually fixed.

### Checklist

### DIFF
--- a/app/survey_config/business_config.py
+++ b/app/survey_config/business_config.py
@@ -1,3 +1,5 @@
+from warnings import warn
+
 from dataclasses import dataclass, field
 from typing import Iterable, Mapping, MutableMapping, Optional
 
@@ -19,13 +21,19 @@ class BusinessSurveyConfig(
         super().__post_init__()
 
         if not self.account_service_log_out_url:
-            self.account_service_log_out_url: str = f"{self.base_url}/sign-in/logout"
+            self.account_service_log_out_url: str = (
+                f"{self.stripped_base_url}/sign-in/logout"
+            )
 
         if not self.account_service_my_account_url:
-            self.account_service_my_account_url: str = f"{self.base_url}/my-account"
+            self.account_service_my_account_url: str = (
+                f"{self.stripped_base_url}/my-account"
+            )
 
         if not self.account_service_todo_url:
-            self.account_service_todo_url: str = f"{self.base_url}/surveys/todo"
+            self.account_service_todo_url: str = (
+                f"{self.stripped_base_url}/surveys/todo"
+            )
 
         self.footer_links = [
             Link(lazy_gettext("What we do"), self.what_we_do_url).__dict__,
@@ -60,3 +68,10 @@ class BusinessSurveyConfig(
             if is_authenticated
             else None
         )
+
+    @property
+    def stripped_base_url(self) -> str:
+        warn(
+            "base_url contains extra pathing which will eventually be corrected and this function will need to be removed"
+        )
+        return self.base_url.replace("/surveys/todo", "")


jamesnunn:
[Back to surveys link not working (2 SP)](https://trello.com/c/xRocDgdj/5087-back-to-surveys-link-not-working-2-sp)

app/survey_config/business_config.py
```diff
@@ -60,3 +67,10 @@ def get_service_links(
             if is_authenticated
             else None
         )
+
+    @property
+    def stripped_base_url(self) -> str:
+        warn(
+            "base_url contains extra pathing which will eventually be corrected and this function will need to be removed"
+        )
+        return self.base_url.replace("/surveys/todo", "")
```
MebinAbraham, line 76: Functionally, works as expected. Minor suggestion. Any reason not to overwrite the `base_url` with the stripped version as this approach doesn't prevent the next dev using `self.base_url` directly?.

tests/app/helpers/test_template_helpers.py
```diff
@@ -557,6 +557,30 @@ def test_survey_config_base_url_provided_used_in_links(
...
+def test_survey_config_base_url_correct_todo(app: Flask):
+    base_url = "http://localhost"
+    with app.app_context():
+        result = BusinessSurveyConfig(base_url=base_url)
+
+    assert result.base_url == "http://localhost"
+
+    assert result.account_service_log_out_url == "http://localhost/sign-in/logout"
+    assert result.account_service_my_account_url == "http://localhost/my-account"
+    assert result.account_service_todo_url == "http://localhost/surveys/todo"
```
MebinAbraham, line 581: Don't think this is needed. These are already tested.

app/survey_config/business_config.py
jamesnunn: The options for the change were: _This can be done either for the self.base_url or for each of the resulting 3 URLs that are constructed_

I went for the second option, I figured the implications of only updating the 3 urls were considered already, but I guess if the base_url is likely to be wrong for a while it is worth doing then

app/survey_config/business_config.py
```diff
@@ -19,20 +19,16 @@ class BusinessSurveyConfig(
     def __post_init__(self):
         super().__post_init__()
 
+        self.base_url = self._stripped_base_url
```
MebinAbraham, line 22: This needs to be called before calling `super().__post_init__()` as that makes use of base url. At the moment the links they generate are also broken. So tests should be updated to assert them as well.

```py
    def __post_init__(self):
        self.contact_us_url: str = f"{self.base_url}/contact-us/"
        self.cookie_settings_url: str = f"{self.base_url}/cookies/"
        self.privacy_and_data_protection_url: str = (
            f"{self.base_url}/privacy-and-data-protection/"
        )
 ```

File: app/survey_config/business_config.py (hunk 1 above)
Comment (MebinAbraham) [reply app/survey_config/business_config.py:72]: Ya, I think it's worth changing, we don't know when it will be fixed and turns as other urls were broken also.

File: app/survey_config/business_config.py [hunk 4]
```diff
@@ -18,6 +19,8 @@ class BusinessSurveyConfig(
...
```
parimooa, line 20: i think this should be suffice , no need to do property. i am not sure if it is anti-pattern , i hope not
`self.base_url=self.base_url.replace("/surveys/todo", "")` ?

jamesnunn: I just thought to keep it as a single thing that can be removed later, plus include the warning in the property.

jamesnunn: Not anymore, just need to remove `self.base_url = self._stripped_base_url` and the function now. Happy to with majority though if there is one

parimooa (approved):

MebinAbraham (approved):

PR #777, opened 2022-01-26
Fix exit button link on introduction page

MebinAbraham:
Fixes the missing exit button link on the introduction page.

### How to review 
- Launch `test_introduction.json`
- Ensure when you click `Exit` you are signed out and taken to the log-out URL.

### Checklist

### DIFF
--- a/app/translations/messages.pot
+++ b/app/translations/messages.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2022-01-25 12:33+0000\n"
+"POT-Creation-Date: 2022-01-26 15:54+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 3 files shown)


LJBabbage (commented):
This is a slight pain, as usually I would expect this to be done in context of the handler, in this case Content, and then match the pattern we have used in multiple other places in templates (i.e {% set sign_out_url = content.sign_out_url %})  However I am thinking because this is used on at least error pages, that because they don't pass any content on rendering they would need updating as well?

MebinAbraham:
> This is a slight pain, as usually I would expect this to be done in context of the handler, in this case Content, and then match the pattern we have used in multiple other places in templates (i.e {% set sign_out_url = content.sign_out_url %}) However I am thinking because this is used on at least error pages, that because they don't pass any content on rendering they would need updating as well?

I think we need to look to do this as a clean-up piece, we seem to only use content.sign_out_url in post submission pages. But we do have `url_for` littered though the templates. I prefer to pass to the templates the context it needs instead of `url_for`, but I think it needs a wider clean up to remove all the ones we can. We should be able to remove all, including error pages, as they still call render templates method, so we can pass it.

We also need to clean up the actual context passes, in question pages, we pass it as top level arguments, in post submission pages it is sent as `content.*` which is inconsistent. So if I were to pass this to template helpers now, it would be top level even though existing ones on post submission use `content.sign_out_url`.

parimooa (approved):
functionally working as expected

berroar (approved):

LJBabbage (approved):

PR #782, opened 2022-02-01
Update form validation dates to pytest and fix coverage

LJBabbage:
This PR updates the tests for forms validation for DateCheck, DateRequired and DateRangeCheck to pytest. Extra tests have been added as coverage was missing from the original and some have been removed as I thought they weren't overly useful

### How to review 
Firstly check coverage of just these new tests now has line coverage for the correct parts in validators.py. I quite like updating scripts/run_tests_unit.sh  (i.e py.test ...... -k test_date_check_validator) and then looking at htmlcov/index.html in a browser to see coverage. (there are other ways of course)

Once that has been done check that everything has been updated to pytest correctly.

### Checklist

### DIFF
--- a/tests/app/forms/validation/test_date_check_validator.py
+++ b/tests/app/forms/validation/test_date_check_validator.py
@@ -1,198 +1,90 @@
-import unittest
-from unittest.mock import Mock
-
+import pytest
 from wtforms.validators import StopValidation
 
 from app.forms import error_messages
-from app.forms.validators import DateCheck
 
 
-class TestDateCheckValidator(unittest.TestCase):
-    def test_date_type_validator_empty_string(self):
-        validator = DateCheck()
-
-        mock_form = Mock()
-        mock_form.data = ""
-
-        mock_field = Mock()
-
-        with self.assertRaises(StopValidation) as ite:
-            validator(mock_form, mock_field)
-
-        self.assertEqual(error_messages["INVALID_DATE"], str(ite.exception))
-
-    def test_date_type_validator_missing_day(self):
-        validator = DateCheck()
-
-        mock_form = Mock()
-        mock_form.data = "2016-12-"
-        mock_form.year.data = "2016"
-
-        mock_field = Mock()
-
-        with self.assertRaises(StopValidation) as ite:
-            validator(mock_form, mock_field)
-
-        self.assertEqual(error_messages["INVALID_DATE"], str(ite.exception))
-
-    def test_date_type_validator_missing_month(self):
-        validator = DateCheck()
-
-        mock_form = Mock()
-        mock_form.data = "2016--03"
-        mock_form.year.data = "2016"
-
-        mock_field = Mock()
-
-        with self.assertRaises(StopValidation) as ite:
-            validator(mock_form, mock_field)
-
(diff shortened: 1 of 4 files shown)


tests/app/forms/validation/test_date_check_validator.py
```diff
@@ -1,198 +1,90 @@
...
-        with self.assertRaises(StopValidation) as ite:
-            validator(mock_form, mock_field)
-
-        self.assertEqual(error_messages["INVALID_DATE"], str(ite.exception))
-
-    def test_date_type_validator_invalid_month_digits(self):
-        validator = DateCheck()
-
-        mock_form = Mock()
-        mock_form.data = "2020--2-20"
```
LJBabbage, line 125: I couldn't see the value of the 3 tests with invalid_digits, (i.e 2020--2-20) Naturally we could be here all day with different combinations (letters, characters, too long etc) So I just concentrated on missing parts and invalid dates, i.e a month with 13, a day with 32 , an incorrect leap year

tests/app/forms/validation/test_date_range_validator.py
```diff
@@ -1,193 +1,167 @@
...
 
-        self.assertEqual(
-            "Enter a reporting period greater than or equal to 20 days",
-            str(context.exception),
-        )
 
-    def test_date_range_with_too_large_period(self):
-        period_max = {"months": 1}
-        validator = DateRangeCheck(period_max=period_max)
+def test_message_combinations(
```
MebinAbraham, line 84: There is a lot going on in this test, I know the existing had it, but while we are here I think we should parametrize these, which has the period and expected message as params.

MebinAbraham, line 84: There is a lot going on in this test, I know the existing had it, but while we are here I think we should parametrize these, which has the period and expected message as params.

tests/app/forms/validation/test_date_range_validator.py
```diff
@@ -1,193 +1,167 @@
...
 
-        period_to = Mock()
-        period_to.data = "2017-01-05"
+    date_range_check = get_date_range_check()
 
-        mock_form = Mock()
-        validator(mock_form, period_from, period_to)
+    with pytest.raises(ValidationError) as context:
+        date_range_check(mock_form, mock_period_from, mock_period_to)
+    assert error_messages["INVALID_DATE_RANGE"] == str(context.value)
```
MebinAbraham, line 31: Minor, a new line above would make it more reasable.

MebinAbraham (commented):
Looks good. Just a minor issue with an existing test.

tests/app/forms/validation/test_date_range_validator.py
LJBabbage: This test isn't actually very good, going to split it up, name them properly and hit all combinations across the 2 new tests. Going to keep them separate due to var being passed in

MebinAbraham (approved):

tests/app/forms/validation/test_date_check_validator.py
petechd: Looks much better now!

petechd (approved):

PR #784, opened 2022-02-03
Add type hints to data_models

petechd:
Adds type hints to `data_models` modules in runner.

### How to review 
Check if new type hints are correct and adhere to the style guide.

### Checklist


petechd:
[Type annotations for Data Models (3 SP)](https://trello.com/c/bsiIhhYo/5103-type-annotations-for-data-models-3-sp)

app/data_models/app_models.py
```diff
@@ -34,19 +41,31 @@ def __init__(self, eq_session_id, user_id, expires_at, session_data):
...
             # Timezone aware datetime to timestamp
             return int(value.replace(tzinfo=timezone.utc).timestamp())
 
-    def _deserialize(self, value, *args, **kwargs):  # pylint: disable=unused-argument
+    # pylint: disable=unused-argument
+    def _deserialize(
+        self,
+        value: float,
+        *args: Optional[list],
+        **kwargs: Any,
```
jamesnunn, line 67: `**kwargs` should be `dict` I believe

app/data_models/app_models.py
```diff
@@ -56,8 +75,13 @@ class DateTimeSchemaMixin:
     created_at = fields.DateTime()
     updated_at = fields.DateTime()
 
+    # pylint: disable=unused-argument
```
jamesnunn, line 78: Is this function better off as a `@staticmethod` then?

app/data_models/app_models.py
```diff
@@ -34,19 +41,31 @@ def __init__(self, eq_session_id, user_id, expires_at, session_data):
...
 
 # pylint: disable=no-self-use
 class Timestamp(fields.Field):
-    def _serialize(self, value, *args, **kwargs):  # pylint: disable=unused-argument
+    # pylint: disable=unused-argument
+    def _serialize(
+        self,
+        value: datetime,
+        *args: Optional[list],
+        **kwargs: Any,
```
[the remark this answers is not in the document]
petechd: This causes mypy error probably because parent class (`Field`) in fields.py doesn't have type declared for **kwargs:
`def _serialize(self, value: typing.Any, attr: str, obj: typing.Any, **kwargs):`. Can't really change it cause it calls standard library class.

app/data_models/app_models.py
petechd: Same as above.

app/data_models/app_models.py
```diff
@@ -1,17 +1,18 @@
...
-        collection_exercise_sid,
-        version,
-        submitted_at=None,
-        expires_at=None,
+        user_id: str,
+        state_data: str,
+        collection_exercise_sid: str,
+        version: int,
+        submitted_at: datetime = None,
+        expires_at: datetime = None,
```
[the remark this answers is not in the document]
petechd: Missed that one, thanks!

app/data_models/list_store.py
```diff
@@ -10,46 +12,45 @@
...
+def random_string(length: int) -> str:
     return "".join(random.choice(ascii_letters) for _ in range(length))
 
 
 class ListModel:
     def __init__(
         self,
         name: str,
-        items: Optional[List] = None,
+        items: Optional[list] = None,
```
MebinAbraham, line 23: 👍

app/data_models/list_store.py
```diff
@@ -10,46 +12,45 @@
...
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
         if not isinstance(other, ListModel):
             return NotImplemented
         return self.items == other.items and self.primary_person == other.primary_person
 
-    def __iter__(self):
-        for list_item in self.items:
-            yield list_item
+    def __iter__(self) -> Iterator:
```
MebinAbraham, line 37: We should be more specific.

app/data_models/list_store.py
```diff
@@ -60,11 +61,11 @@ def serialize(self):
 
         return serialized
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"<ListModel name={self.name} items={self.items}, primary_person={self.primary_person}>"
 
     @property
-    def first(self):
+    def first(self) -> Any:
```
MebinAbraham, line 68: This returns a string.

app/data_models/list_store.py
```diff
@@ -120,54 +121,53 @@ def __init__(self, existing_items: Optional[List[Mapping]] = None):
...
         self._is_dirty = False
 
-    def __iter__(self):
-        for list_item in self._lists.values():
-            yield list_item
+    def __iter__(self) -> Iterator:
+        yield from self._lists.values()
 
-    def __getitem__(self, list_name):
+    def __getitem__(self, list_name: str) -> Union[ListModel, Any]:
```
MebinAbraham, line 127: Should just be a ListModel. FYI, any time `Any` is return, it makes the other types you return obsolete as Any will supersede all types.

app/data_models/list_store.py
```diff
@@ -120,54 +121,53 @@ def __init__(self, existing_items: Optional[List[Mapping]] = None):
...
-    def _generate_identifier(self):
+    def _generate_identifier(self) -> Optional[str]:
         """Generate an unused random 6 character string"""
         while True:
             candidate = random_string(EQ_LIST_ITEM_ID_LENGTH)
             if candidate not in self._list_item_ids():
                 return candidate
 
-    def _list_item_ids(self):
+    def _list_item_ids(self) -> list:
```
MebinAbraham, line 159:
```suggestion
    def _list_item_ids(self) -> list[str]:
```

app/data_models/list_store.py
```diff
@@ -181,7 +181,9 @@ def delete_list_item(self, list_name, item_id):
 
         self._is_dirty = True
 
-    def add_list_item(self, list_name, primary_person=False):
+    def add_list_item(
+        self, list_name: str, primary_person: bool = False
+    ) -> Optional[str]:
```
MebinAbraham, line 186: Should always return a string?

app/data_models/list_store.py
```diff
@@ -208,11 +210,11 @@ def add_list_item(self, list_name, primary_person=False):
 
         return list_item_id
 
-    def serialize(self):
+    def serialize(self) -> list:
         return [list_model.serialize() for list_model in self]
 
     @classmethod
-    def deserialize(cls, serialized: list):
+    def deserialize(cls, serialized: list) -> Any:
```
MebinAbraham, line 217: This returns a store.
```suggestion
    def deserialize(cls, serialized: list[Mapping]) -> ListStore:
```

app/data_models/progress_store.py
```diff
@@ -12,7 +22,7 @@ class CompletionStatus:
     NOT_STARTED: str = "NOT_STARTED"
     INDIVIDUAL_RESPONSE_REQUESTED: str = "INDIVIDUAL_RESPONSE_REQUESTED"
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator:
```
MebinAbraham, line 25: We should be more specific.

app/data_models/progress_store.py
```diff
@@ -34,7 +44,7 @@ def __init__(self, in_progress_sections: List[Mapping] = None) -> None:
             in_progress_sections or []
         )  # type: MutableMapping
 
-    def __contains__(self, section_key) -> bool:
+    def __contains__(self, section_key: Union[int, Sequence[int]]) -> bool:
```
MebinAbraham, line 47: Sections keys are tuples made up of the section_id and optional list item id.

app/data_models/progress_store.py
```diff
@@ -84,7 +94,7 @@ def is_section_complete(
 
     def section_keys(
         self, statuses: Iterable[str] = None, section_ids: Iterable[str] = None
-    ):
+    ) -> list:
```
MebinAbraham, line 97: As per type hinting guide, if we can, define the types for all collections.

app/data_models/progress_store.py
```diff
@@ -125,7 +135,7 @@ def update_section_status(
 
     def get_section_status(
         self, section_id: str, list_item_id: Optional[str] = None
-    ) -> str:
+    ) -> Union[str, Any]:
```
MebinAbraham, line 138: Always a str?

app/data_models/progress_store.py
```diff
@@ -134,7 +144,7 @@ def get_section_status(
 
     def get_completed_block_ids(
         self, section_id: str, list_item_id: Optional[str] = None
-    ) -> List[Optional[str]]:
+    ) -> Union[list[Optional[str]], Any]:
```
MebinAbraham, line 147: Always a list of strings?

app/data_models/questionnaire_store.py
```diff
@@ -10,18 +12,21 @@
 class QuestionnaireStore:
     LATEST_VERSION = 1
 
-    def __init__(self, storage, version=None):
+    def __init__(self, storage: Any, version: int = None):
+        # Storage type hint needs to be Any due to circular import
```
MebinAbraham, line 16: Circular import when typing can be avoided by using the type hinting flag.
```py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from something import stuff
```

app/data_models/questionnaire_store.py
```diff
@@ -10,18 +12,21 @@
...
         self._storage = storage
         if version is None:
             version = self.get_latest_version_number()
         self.version = version
-        self._metadata = {}
+        self._metadata: Any = {}
         # self.metadata is a read-only view over self._metadata
-        self.metadata = MappingProxyType(self._metadata)
-        self.response_metadata = {}
+        self.metadata: Any = MappingProxyType(self._metadata)
```
MebinAbraham, line 23: How come these are Any?

app/data_models/questionnaire_store.py
```diff
@@ -10,18 +12,21 @@
...
+        self.response_metadata: dict = {}
         self.list_store = ListStore()
         self.answer_store = AnswerStore()
         self.progress_store = ProgressStore()
+        self.submitted_at: Union[datetime, None]
```
MebinAbraham, line 28: Optional instead of `Union` since it is one other type?.

app/data_models/questionnaire_store.py
```diff
@@ -35,10 +40,10 @@ def __init__(self, storage, version=None):
         if version is not None:
             self.version = version
 
-    def get_latest_version_number(self):
+    def get_latest_version_number(self) -> int:
         return self.LATEST_VERSION
 
-    def set_metadata(self, to_set):
+    def set_metadata(self, to_set: Mapping[str, Union[str, int, list]]) -> Any:
```
MebinAbraham, line 46: It returns `self` so should be a questionnaire store instance.

app/data_models/questionnaire_store.py
```diff
@@ -48,15 +53,15 @@ def set_metadata(self, to_set):
...
+    def _deserialize(self, data: str) -> None:
         json_data = json_loads(data)
         self.progress_store = ProgressStore(json_data.get("PROGRESS"))
         self.set_metadata(json_data.get("METADATA", {}))
         self.answer_store = AnswerStore(json_data.get("ANSWERS"))
         self.list_store = ListStore.deserialize(json_data.get("LISTS"))
         self.response_metadata = json_data.get("RESPONSE_METADATA", {})
 
-    def serialize(self):
+    def serialize(self) -> Any:
```
MebinAbraham, line 64: Would be fixed by typing `json_dumps`

app/data_models/relationship_store.py
```diff
@@ -21,45 +21,47 @@ class RelationshipStore:
     Stores and updates relationships.
     """
 
-    def __init__(self, relationships: Optional[List[Mapping]] = None) -> None:
+    def __init__(self, relationships: dict = None) -> None:
         self._is_dirty = False
         self._relationships = self._build_map(relationships or [])
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator:
```
MebinAbraham, line 28: We should be more specific.

app/data_models/relationship_store.py
```diff
@@ -21,45 +21,47 @@ class RelationshipStore:
...
+    def is_dirty(self) -> bool:
         return self._is_dirty
 
-    def clear(self):
+    def clear(self) -> None:
         self._relationships.clear()
         self._is_dirty = True
 
-    def serialize(self):
+    def serialize(self) -> list:
```
MebinAbraham, line 48: We should be more specific.

app/data_models/relationship_store.py
```diff
@@ -21,45 +21,47 @@ class RelationshipStore:
...
         return [
             relationship.for_json() for relationship in self._relationships.values()
         ]
 
-    def get_relationship(self, list_item_id, to_list_item_id):
+    def get_relationship(
+        self, list_item_id: str, to_list_item_id: str
+    ) -> Union[Relationship, list]:
```
MebinAbraham, line 55: This should return an optional Relationship.

app/data_models/relationship_store.py
```diff
@@ -89,7 +91,7 @@ def remove_all_relationships_for_list_item_id(self, list_item_id: str):
             self._is_dirty = True
 
     @staticmethod
-    def _build_map(relationships):
+    def _build_map(relationships: Union[dict, Union[dict, list]]) -> dict:
```
MebinAbraham, line 94: relationships should be a mapping and return should be similar to the other store `dict[tuple[str, str], Relationship]`

app/data_models/session_data.py
```diff
@@ -1,24 +1,29 @@
...
+        response_id: METADATA_MAPPING_TYPE,
+        case_id: METADATA_MAPPING_TYPE,
+        case_ref: METADATA_MAPPING_TYPE = None,
+        account_service_base_url: METADATA_MAPPING_TYPE = None,
+        account_service_log_out_url: METADATA_MAPPING_TYPE = None,
+        trad_as: METADATA_MAPPING_TYPE = None,
+        display_address: str = None,
+        confirmation_email_count: int = 0,
+        feedback_count: int = 0,
+        **_: str,
```
MebinAbraham, line 26: Each property should have an explicit type. i.e tx_id would never be a list or int, etc.

MebinAbraham (commented):
Quite a few mistypes, but should be easy fixes. We need to be careful with using `Any`, in most cases, we don't need it. It should only be needed if we are interacting with untyped 3rd packages.

app/views/contexts/view_submitted_response_context.py
```diff
@@ -20,46 +20,49 @@ def build_view_submitted_response_context(
...
-        submitted_text = lazy_gettext("Answers submitted.")
-    elif trad_as := questionnaire_store.metadata.get("trad_as"):
-        submitted_text = lazy_gettext(
-            "Answers submitted for <span>{ru_name}</span> ({trad_as})"
-        ).format(ru_name=questionnaire_store.metadata["ru_name"], trad_as=trad_as)
-    else:
-        submitted_text = lazy_gettext(
-            "Answers submitted for <span>{ru_name}</span>"
-        ).format(ru_name=questionnaire_store.metadata["ru_name"])
+    if questionnaire_store.submitted_at:
```
MebinAbraham, line 23: Instead of this indentation, we should escape early. This shouldn't happen on the basis that the dev has set it up correctly, as far as typing goes, it is correctly highlight that we need to handle it, so I would just raise an exception therefore will need a test to prevent line cov failing.

app/views/handlers/feedback.py
```diff
@@ -48,8 +48,9 @@ def __init__(
     ):
         if not self.is_enabled(schema):
             raise FeedbackNotEnabled
-
-        if self.is_limit_reached(session_store.session_data):
+        if isinstance(
+            session_store.session_data, SessionData
+        ) and self.is_limit_reached(session_store.session_data):
```
MebinAbraham, line 53: As above.

app/views/handlers/feedback.py
```diff
@@ -85,38 +86,42 @@ def get_page_title(self) -> str:
...
-            feedback_text=self.form.data.get("feedback-text"),
-            feedback_type=self.form.data.get("feedback-type"),
-        )
-        message = feedback_message()
-        metadata = feedback_metadata()
-        message.update(metadata)
-        encrypted_message = encrypt(
-            message, current_app.eq["key_store"], KEY_PURPOSE_SUBMISSION  # type: ignore
-        )
+        if isinstance(session_data, SessionData):
```
MebinAbraham, line 89: We should avoid this indenting for the sake of type hinting. We should escape early but the check here should be needed if we escape early on _init_ if we don't have session data.

app/views/handlers/list_add_question.py
```diff
@@ -21,8 +21,9 @@ def handle_post(self):
...
-        list_length = len(
-            self._questionnaire_store.list_store[self._current_location.list_name]
-        )
+        if self._current_location.list_name:
+            list_length = len(
+                self._questionnaire_store.list_store[self._current_location.list_name]
+            )
 
-        return {"list_item_position": list_length + 1}
+            return {"list_item_position": list_length + 1}
```
MebinAbraham, line 29: Same here, list name should exist, if it didn't, currently runner would error. So we either need to handle that, or ignore type hinting and accept the risk that we should have configured it right.

app/views/handlers/view_submitted_response.py
```diff
@@ -36,9 +36,10 @@ def __init__(
 
     @property
     def has_expired(self) -> bool:
-        return has_view_submitted_response_expired(
-            self._questionnaire_store.submitted_at
-        )
+        if self._questionnaire_store.submitted_at:
```
MebinAbraham, line 39: the else bit should return a bool also.

app/data_models/app_models.py
petechd: I've made it static since `self` was not needed, extra test added as well, thanks!

app/authentication/authenticator.py
```diff
@@ -103,7 +103,10 @@ def load_user(extend_session: bool = True) -> Optional[User]:
         user_ik = cookie_session.get(USER_IK)
         user = User(user_id, user_ik)
 
-        if session_store.session_data.tx_id:
+        if (
+            isinstance(session_store.session_data, SessionData)
```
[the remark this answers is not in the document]
petechd: Because `session_data` had a type `Union[dict, SessionData]` in `globals.py`, I've modified it in globals (removed dict as a potential type), so it only can be optional SessionData and we check if it exists in this snippet above now instead of `isinstance()`. Thanks!

app/data_models/answer_store.py
```diff
@@ -28,35 +28,36 @@ def __init__(self, existing_answers: List[Dict] = None):
         self.answer_map = self._build_map(existing_answers or [])
         self._is_dirty = False
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator:
```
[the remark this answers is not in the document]
petechd: Iterator is more specific now (returns a tuple). Thanks!

app/data_models/relationship_store.py
petechd: More specific now. Thanks.

app/data_models/relationship_store.py
petechd: Added, thanks.

app/data_models/app_models.py
```diff
@@ -56,8 +75,13 @@ class DateTimeSchemaMixin:
...
     updated_at = fields.DateTime()
 
+    # pylint: disable=unused-argument
     @pre_dump
-    def set_date(self, data, **kwargs):  # pylint: disable=unused-argument
+    def set_date(
+        self,
+        data: QuestionnaireState,
+        **kwargs: dict,
+    ) -> QuestionnaireState:
```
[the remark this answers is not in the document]
petechd: Added union return of QuestionnaireState and EQSession. thanks.

app/data_models/session_store.py
```diff
@@ -11,18 +16,18 @@
...
 class SessionStore:
-    def __init__(self, user_ik, pepper, eq_session_id=None):
+    def __init__(self, user_ik: str, pepper: str, eq_session_id: str = None) -> None:
         self.eq_session_id = eq_session_id
-        self.user_id = None
+        self.user_id: Optional[str] = None
         self.user_ik = user_ik
-        self.session_data = None
-        self._eq_session = None
+        self.session_data: Optional[Union[dict, SessionData]] = None
```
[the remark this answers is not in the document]
petechd: Added as optional `SessionData`

app/data_models/list_store.py
```diff
@@ -120,54 +121,53 @@ def __init__(self, existing_items: Optional[List[Mapping]] = None):
 
         self._is_dirty = False
 
-    def __iter__(self):
-        for list_item in self._lists.values():
-            yield list_item
+    def __iter__(self) -> Iterator:
```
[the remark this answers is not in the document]
petechd: More specific now, thanks!

app/data_models/relationship_store.py
petechd: Returning optional Relationship now, thanks.

app/data_models/relationship_store.py
petechd: Ok, added, thanks.

app/storage/storage_encryption.py
```diff
@@ -51,7 +51,7 @@ def encrypt_data(self, data: Union[str, dict]) -> str:
         serialized_token: str = jwe_token.serialize(compact=True)
         return serialized_token
 
-    def decrypt_data(self, encrypted_token: str) -> bytes:
+    def decrypt_data(self, encrypted_token: str) -> Union[bytes, str]:
```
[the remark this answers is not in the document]
petechd: I think this needs to stay as it is as it touches standard library `builtins.py` types.

app/data_models/session_data.py
petechd: I don't think we are able to cast more specific type on these, because they are given more generic type in `authenticator.py`:
```
def _create_session_data_from_metadata(
    metadata: Mapping[str, Union[str, int, list]]
```
so they could be any of str, int, list types.

app/data_models/answer_store.py
```diff
@@ -28,35 +28,36 @@ def __init__(self, existing_answers: List[Dict] = None):
...
+    def _build_map(answers: list[dict]) -> dict:
         """Builds the answer_store's data structure from a list of answer dictionaries"""
         return {
             (answer["answer_id"], answer.get("list_item_id")): Answer.from_dict(answer)
             for answer in answers
         }
 
     @staticmethod
-    def _validate(answer):
+    def _validate(answer: Optional[Answer]) -> None:
```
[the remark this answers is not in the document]
petechd: Answer no longer optional, other type hint changes made if fixable.

app/data_models/answer_store.py
```diff
@@ -102,13 +103,13 @@ def get_answers_by_answer_id(
...
 
-    def clear(self):
+    def clear(self) -> None:
         """
         Clears answers *in place*
         """
         self.answer_map.clear()
 
-    def remove_answer(self, answer_id: str, list_item_id: str = None):
+    def remove_answer(self, answer_id: str, list_item_id: str = None) -> None:
```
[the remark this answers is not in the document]
petechd: Added `no_implicit_optional = True` to mypy.ini data models section and fixed all the occurrences in that directory. Thanks.

app/data_models/fulfilment_request.py
```diff
@@ -17,7 +17,7 @@ def transaction_id(self) -> str:
         return str(uuid4())
 
     @property
-    def message(self) -> bytes:
+    def message(self) -> Union[bytes, Any]:
```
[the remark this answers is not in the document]
petechd: It's because json_dumps uses `dumps()` from `simplejson` that doesn't have type hints so cannot return bytes.

app/views/handlers/feedback.py
```diff
@@ -254,8 +259,8 @@ def __init__(
         metadata: Mapping[str, Union[str, int, list]],
         response_metadata: Mapping[str, Union[str, int, list]],
         schema: QuestionnaireSchema,
-        case_id: str,
-        submission_language_code: Optional[str],
+        case_id: METADATA_MAPPING_TYPE,
+        submission_language_code: METADATA_MAPPING_TYPE,
```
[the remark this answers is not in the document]
petechd: Similar to previous comment, I don't think we are able to cast more specific type on these, because they are given more generic type in `authenticator.py`:
```
def _create_session_data_from_metadata(
    metadata: Mapping[str, Union[str, int, list]]
```
so they could be any of str, int, list types as they are extracted using `metadata.get()`.

app/views/contexts/thank_you_context.py
```diff
@@ -30,9 +30,10 @@ def build_thank_you_context(
         submission_text = lazy_gettext(
             "Your answers have been submitted for <span>{company_name}</span>"
         ).format(company_name=session_data.ru_name)
-    metadata = build_submission_metadata_context(
-        survey_type, submitted_at, session_data.tx_id
-    )
+    if isinstance(session_data.tx_id, str):
```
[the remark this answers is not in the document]
petechd: It's not but it may not be a string since tx_id could be any of `str, int, list`, since that's what is casted in `authenticator.py` on metadata.

app/data_models/list_store.py
```diff
@@ -10,46 +12,45 @@
...
     ):
         self.name = name
         self.items = items or []
         self.primary_person = primary_person
         self.same_name_items = same_name_items or []
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
         if not isinstance(other, ListModel):
             return NotImplemented
```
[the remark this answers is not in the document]
petechd: Added to answer store, thanks.

app/data_models/list_store.py
```diff
@@ -10,46 +12,45 @@
...
         return self.items == other.items and self.primary_person == other.primary_person
 
-    def __iter__(self):
-        for list_item in self.items:
-            yield list_item
+    def __iter__(self) -> Iterator:
+        yield from self.items
 
-    def __getitem__(self, list_item_index):
+    def __getitem__(self, list_item_index: int) -> Union[dict, Any]:
```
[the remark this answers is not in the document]
petechd: I think it checks standard library `builtins` `items` where the type is not define so has to stay `Any`.

app/data_models/list_store.py
petechd: Same as above, I think it confuses standard library `builtins` `items` where the type is not define with our var which has the same name, so it becomes `Any`.

app/data_models/list_store.py
```diff
@@ -120,54 +121,53 @@ def __init__(self, existing_items: Optional[List[Mapping]] = None):
...
-    def get(self, item: str):
+    def get(self, item: str) -> ListModel:
         return self.__getitem__(item)
 
-    def list_item_position(self, for_list, list_item_id) -> int:
+    def list_item_position(self, for_list: str, list_item_id: Union[dict, Any]) -> int:
         return self[for_list].index(list_item_id) + 1
 
-    def _generate_identifier(self):
+    def _generate_identifier(self) -> Optional[str]:
```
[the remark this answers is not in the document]
petechd: Always returns str now, thanks.

app/data_models/list_store.py
```diff
@@ -120,54 +121,53 @@ def __init__(self, existing_items: Optional[List[Mapping]] = None):
...
         return {
             list_model["name"]: ListModel(**list_model) for list_model in list_models
         }
 
-    def get(self, item: str):
+    def get(self, item: str) -> ListModel:
         return self.__getitem__(item)
 
-    def list_item_position(self, for_list, list_item_id) -> int:
+    def list_item_position(self, for_list: str, list_item_id: Union[dict, Any]) -> int:
```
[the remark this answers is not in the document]
petechd: Made it a string, thanks.

app/data_models/progress_store.py
petechd: More specific now, thanks!

app/data_models/progress_store.py
petechd: Defined, thanks.

app/data_models/questionnaire_store.py
petechd: Cannot be typed cause it uses standard library `dumps`.

app/data_models/questionnaire_store.py
petechd: Changed to QuestionnaireStore, cheers!

app/data_models/questionnaire_store.py
petechd: Optional datetime now, thanks.

app/data_models/list_store.py
petechd: Changed to ListModel as return.

app/data_models/progress_store.py
petechd: `status` in `Progress` class has type `status: Optional[str] = None` so `self._progress[section_key].status` not always returns a string.

app/data_models/progress_store.py
petechd: 'block_ids` are `List[Optional[str]]` in `Progress` class hence Any as return.

app/data_models/questionnaire_store.py
petechd: I was able get pass the circular import problem by just importing `EncryptedQuestionnaireStorage` type, possibly some other typing fixes helped. Thanks.

app/data_models/questionnaire_store.py
petechd: We use `clear()` method further down in the code on this `metadata` but we declare it here as empty `dict` hence the type hints confusion.

app/data_models/session_store.py
```diff
@@ -64,43 +75,44 @@ def save(self):
...
-            current_app.eq["storage"].delete(self._eq_session)
+            current_app.eq["storage"].delete(self._eq_session)  # type: ignore
 
             self._eq_session = None
             self.eq_session_id = None
             self.user_id = None
             self.session_data = None
 
-    def _load(self):
+    def _load(self) -> Optional[EQSession]:
```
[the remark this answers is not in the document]
petechd: Sorry, I am not sure if I understand this sentence😁 This method returns `EQSession` or throws an error so it's `Optional[EQSession]`.

app/views/handlers/confirm_email.py
```diff
@@ -49,8 +49,9 @@ def __init__(
 
         if not ConfirmationEmail.is_enabled(schema):
             raise ConfirmationEmailNotEnabled
-
-        if ConfirmationEmail.is_limit_reached(session_store.session_data):
+        if isinstance(
+            session_store.session_data, SessionData
+        ) and ConfirmationEmail.is_limit_reached(session_store.session_data):
```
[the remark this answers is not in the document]
petechd: I've added `ignore` to it, for now, as we probably need wider discussion about it.

app/views/handlers/confirmation_email.py
```diff
@@ -33,7 +33,9 @@ def __init__(
         if not self.is_enabled(schema):
             raise ConfirmationEmailNotEnabled
 
-        if self.is_limit_reached(session_store.session_data):
+        if isinstance(
+            session_store.session_data, SessionData
+        ) and self.is_limit_reached(session_store.session_data):
```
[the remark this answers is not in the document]
petechd: I've added `ignore` to it, for now, as we probably need wider discussion about it.

app/views/handlers/feedback.py
petechd: I've added `ignore` to it, for now, as we probably need wider discussion about it.

app/views/handlers/list_add_question.py
petechd: Added ignore to type hinting here, I hope you agree?

app/views/contexts/view_submitted_response_context.py
petechd: Escaping early with an exception, nesting removed, test added to give 100% cover, thanks!

app/views/handlers/feedback.py
petechd: So, as we discussed, this will get type ignore for now and we will revisit once we decide about the approach.

app/views/handlers/feedback.py
```diff
@@ -84,13 +83,16 @@ def get_page_title(self) -> str:
...
     def handle_post(self) -> None:
-        session_data = self._session_store.session_data
+        session_data: SessionData = self._session_store.session_data  # type: ignore
         session_data.feedback_count += 1
 
-        feedback_metadata = FeedbackMetadata(
-            session_data.case_id,
-            session_data.tx_id,
-        )
+        if isinstance(session_data.case_id, str) and isinstance(
```
MebinAbraham, line 89: Should be removed. Will be fixed by fixing the types on SessionData/Questionnaire metadata.

app/views/handlers/feedback.py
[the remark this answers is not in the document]
MebinAbraham: We need to make the initial metadata `Mapping[str, Any]` so that we can be explicit. We could also use typed dicts.

app/views/handlers/feedback.py
petechd: After changing metadata type to `Mapping[str, Any]` `tx_id and `case_id` are still optional `str` so I think this needs to stay.

petechd:
> Will be fixed when the comment against session data is done. tx_id can never be list etc, so we need to update our typing.

`tx_id` is still optional `str` so the `if isinstance(session_data.tx_id, str):` need to stay, I think.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -14,7 +14,7 @@ def convert_answers_to_payload_0_0_3(
     list_store: ListStore,
     schema: QuestionnaireSchema,
     full_routing_path: RoutingPath,
-) -> List[Dict]:
+) -> list[Answer]:
```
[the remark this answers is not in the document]
petechd: I had to keep it as it is, mypy is checking these files now (`[mypy-app.submitter.*]` rule). This is the only type it accepts 🤷‍♂️

petechd:
> It should be bytes. It is explicitly typed as bytes on L58, and that is what is returned.
I tried many times, but even if it has assigned `bytes` type, it still but it returns `str` somehow, issue with mypy?

app/data_models/session_store.py
```diff
@@ -64,43 +77,44 @@ def save(self):
...
             ).decrypt_data(encrypted_session_data)
 
-            session_data = session_data.decode()
-            # for backwards compatibility
-            # session data used to be base64 encoded before encryption
-            try:
-                session_data = base64url_decode(session_data).decode()
-            except ValueError:
-                pass
+            if isinstance(session_data, bytes):
```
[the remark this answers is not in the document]
petechd: I tried many times, but even if it has assigned `bytes` type, it still but it returns `str` somehow, so this nesting needs to stay. Potential issue with mypy?

app/data_models/list_store.py
```diff
@@ -208,11 +208,11 @@ def add_list_item(self, list_name, primary_person=False):
 
         return list_item_id
 
-    def serialize(self):
+    def serialize(self) -> list[dict]:
```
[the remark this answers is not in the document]
petechd: Ok, added. Thanks!

app/data_models/relationship_store.py
```diff
@@ -21,45 +21,47 @@ class RelationshipStore:
...
+    def is_dirty(self) -> bool:
         return self._is_dirty
 
-    def clear(self):
+    def clear(self) -> None:
         self._relationships.clear()
         self._is_dirty = True
 
-    def serialize(self):
+    def serialize(self) -> list[Mapping[str, str]]:
```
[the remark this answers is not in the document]
petechd: Added `TypedDict` as an extra class, similar to what we did with list store. Thanks!

app/submitter/convert_payload_0_0_3.py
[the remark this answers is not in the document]
MebinAbraham: You're correct, the doc string is confusing, as it looks JSON, but that is only happens when we call json serialize.

app/views/handlers/feedback.py
```diff
@@ -64,7 +63,7 @@ def form(self) -> QuestionnaireForm:
             question_schema=self.question_schema,
             answer_store=self._questionnaire_store.answer_store,
             list_store=self._questionnaire_store.list_store,
-            metadata=self._questionnaire_store.metadata,
+            metadata=self._questionnaire_store.metadata,  # type: ignore
```
MebinAbraham, line 66: This shouldn't be ignored. metadata on questionnaire should be typed as `Mapping[str, Any]`, then any typing of metadata such as the one in `generate_form` should be updated to match.

app/views/handlers/feedback.py
```diff
@@ -84,13 +83,14 @@ def get_page_title(self) -> str:
...
 
-        feedback_metadata = FeedbackMetadata(
-            session_data.case_id,
-            session_data.tx_id,
-        )
+        if session_data.case_id and session_data.tx_id:
+            feedback_metadata = FeedbackMetadata(
+                session_data.case_id,
+                session_data.tx_id,
+            )
```
MebinAbraham, line 93: Feedback metadata needs to exist, if this statement was false, this would error as `feedback_metadata` wouldn't exist. I think it is okay to ignore it for now, we need to make a design choice first, so no harm ignoring here now given we have ignores in few places here already. We also need to review the types for metadata as case_id and tx_id should be mandatory.

tests/app/views/contexts/test_submitted_response_context.py
```diff
@@ -100,6 +101,15 @@ def test_view_submitted_response_expired(
         assert "summary" not in context
 
 
+def test_build_view_submitted_response_exception(app: Flask):
```
MebinAbraham, line 104: Name isn't very clear, maybe `test_build_view_submitted_response_no_submitted_at`

tests/app/views/handlers/test_view_submitted_response.py
```diff
@@ -15,3 +15,14 @@ def test_not_enabled(storage, language):
 
     with pytest.raises(ViewSubmittedResponseNotEnabled):
         ViewSubmittedResponse(QuestionnaireSchema({}), questionnaire_store, language)
+
+
+def test_has_expired_returns_false(storage, language, app):
```
MebinAbraham, line 20: Maybe `test_has_expired_no_submitted_at_return_false`

MebinAbraham (commented):
Almost there. For all stores, we should use typed dicts and make them consistent. Looking good though. We do need to tackle the ignores soon.

app/data_models/progress_store.py
MebinAbraham: We should make this `str` and change the `Optional[str]` of `status` from `Progress`, this is because currently we set status to None, then update it in a subsequent call. It would make our lives easier if status is not optional, we can set it to `IN_PROGRESS` when we first create it, then the subsequent call which overwrites can stay as is. So on L162 in progress store we should pass in the status of in_progress when creating it and updating the type hint on `Progress.status` to `str`.

app/data_models/progress_store.py
MebinAbraham: Still outstanding, it should be ``list[str]` instead of list of optional strings.

app/authentication/authenticator.py
```diff
@@ -142,7 +140,7 @@ def _create_session_data_from_metadata(
     )
 
 
-def store_session(metadata: Mapping[str, Union[str, int, list]]) -> None:
+def store_session(metadata: dict[str, Any]) -> None:
```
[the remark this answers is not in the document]
petechd: That would mean we would have to change to Mapping the `self._metadata` in `QuestionnaireStore`:
`self._metadata: Mapping[str, Any] = {}`

app/data_models/answer_store.py
```diff
@@ -28,35 +28,39 @@ def __init__(self, existing_answers: List[Dict] = None):
...
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, AnswerStore):
+            return NotImplemented
         return self.answer_map == other.answer_map
 
     @staticmethod
-    def _build_map(answers: List[Dict]):
+    def _build_map(
+        answers: Iterable[Mapping],
+    ) -> dict[tuple[str, Optional[str]], Answer]:
```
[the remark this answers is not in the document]
petechd: I reused `SectionKeyType` from  progress store here.

app/views/handlers/feedback.py
petechd: I've changed it to `Mapping[str, Any]` and removed the ignores but it meant changing types in `questionnaire_form.py`, `submit_questionnaire.py`, `section_summary_context.py`, `globals.py` etc.

app/data_models/progress_store.py
```diff
@@ -22,19 +24,19 @@ class ProgressStore:
     that have been started.
     """
 
-    def __init__(self, in_progress_sections: List[Mapping] = None) -> None:
+    def __init__(self, in_progress_sections: Optional[list[Mapping]] = None) -> None:
```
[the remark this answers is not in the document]
petechd: Using typed dict (new `SectionsDict` typed dict class) now and changed to `Iterable`.

app/data_models/answer_store.py
```diff
@@ -19,7 +19,7 @@ class AnswerStore:
     }
     """
 
-    def __init__(self, existing_answers: List[Dict] = None):
+    def __init__(self, existing_answers: Optional[Iterable[dict]] = None):
```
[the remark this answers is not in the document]
petechd: Added answers typed dict class and used it here, thanks!

app/data_models/answer.py
```diff
@@ -28,7 +28,7 @@ class Answer:
     list_item_id: Optional[str] = field(default=None)
 
     @classmethod
-    def from_dict(cls, answer_dict: dict) -> Answer:
+    def from_dict(cls, answer_dict: Mapping) -> Answer:
```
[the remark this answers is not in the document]
petechd: Added here and reuse it in AnswerStore, thanks!

app/data_models/app_models.py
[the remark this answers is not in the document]
petechd: Added now, thanks and sorry!

app/data_models/progress_store.py
petechd: So I changed it from `None` to `IN_PROGRESS` when we first create it, it made `status` a `str`. That seems to work well now.

app/data_models/questionnaire_store.py
petechd: Ok, changed it to `Mapping[str, Any] = {}` now, thanks!

app/views/handlers/feedback.py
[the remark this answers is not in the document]
petechd: Done now!

app/data_models/answer_store.py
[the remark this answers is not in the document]
MebinAbraham: We shouldn't reuse that, as it isn't a section key.

app/authentication/authenticator.py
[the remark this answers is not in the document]
MebinAbraham: It should be `Mapping` as long as it doesn't need specific methods of `dict`.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -130,7 +130,7 @@ def add_relationships_unrelated_answers(
         relationship_store=relationship_store,
         section_id=section_id,
         list_name=list_name,
-        list_item_ids=list_store[list_name],
+        list_item_ids=list_store[list_name],  # type: ignore
```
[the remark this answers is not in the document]
petechd: because `list_item_ids` is `list[str]` and it is expected here to be `ListModel`.

app/data_models/progress.py
```diff
@@ -1,18 +1,25 @@
...
+    block_ids: list[str]
 
 
 @dataclass
 class Progress:
     section_id: str
-    block_ids: List[Optional[str]]
-    status: Optional[str] = None
+    block_ids: list[str]
+    status: str = "IN_PROGRESS"
```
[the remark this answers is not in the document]
petechd: Set on initialisation now using `CompletionStatus` data class for status.

app/data_models/progress_store.py
```diff
@@ -134,7 +140,7 @@ def get_section_status(
 
     def get_completed_block_ids(
         self, section_id: str, list_item_id: Optional[str] = None
-    ) -> List[Optional[str]]:
+    ) -> list:
```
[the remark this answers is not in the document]
petechd: Made it explicit but had to make `block_id` optional.

app/data_models/progress_store.py
```diff
@@ -194,7 +200,7 @@ def remove_progress_for_list_item_id(self, list_item_id: str) -> None:
 
             self._is_dirty = True
 
-    def serialize(self) -> List:
+    def serialize(self) -> list:
```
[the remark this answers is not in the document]
petechd: Explicit now, thanks!

app/data_models/progress_store.py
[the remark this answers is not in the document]
MebinAbraham: It should be `list[str]` as `list[Optional[str]]` means it can be `[None]` etc. Progress.block_ids type hints needs updating to match.

app/data_models/progress_store.py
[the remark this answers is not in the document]
MebinAbraham: I don't follow what you mean?

app/submitter/convert_payload_0_0_3.py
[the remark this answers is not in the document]
MebinAbraham: We should realistically update relation router to `list_model` as that what it actually gets. For now, just explicitly pass the IDs in `ist_item_ids=list_store[list_name].items`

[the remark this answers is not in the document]
petechd: Added, thanks!

app/data_models/progress_store.py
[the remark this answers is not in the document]
MebinAbraham: Yup

app/data_models/progress_store.py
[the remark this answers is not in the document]
petechd: Ok, made `block_ids` non optional str but that caused trouble in `location.py` where it defaults to `None`. Made it:
`block_id: Optional[str] = None  # type: ignore`

app/data_models/questionnaire_store.py
petechd: So yeah, it uses `.clear()` method of a dict so needs to stay as dict I think.

app/data_models/list_store.py
```diff
@@ -10,46 +12,45 @@
...
+    def __eq__(self, other: object) -> bool:
         if not isinstance(other, ListModel):
             return NotImplemented
         return self.items == other.items and self.primary_person == other.primary_person
 
-    def __iter__(self):
-        for list_item in self.items:
-            yield list_item
+    def __iter__(self) -> Iterator[str]:
+        yield from self.items
```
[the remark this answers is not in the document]
petechd: Just way shorter, I was not sure if you will like such refactor.

app/data_models/progress_store.py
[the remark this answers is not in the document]
MebinAbraham: Location itself doesn't be ignored. Just whether it is used should be. That is also something we could improve in future.

app/data_models/progress_store.py
petechd: Sorted in another thread now!

app/data_models/progress_store.py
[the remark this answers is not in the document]
petechd: Ok, type ignore removed.

app/data_models/answer.py
```diff
@@ -21,14 +21,20 @@
 ]
 
 
+class AnswersDictType(TypedDict, total=False):
```
MebinAbraham, line 24: Singular, as it represents 1 answer. I also don't think we need `Type` suffix for these one as it can be used as a normal dict, like list store does.

app/views/contexts/context.py
```diff
@@ -19,7 +17,7 @@ def __init__(
         answer_store: AnswerStore,
         list_store: ListStore,
         progress_store: ProgressStore,
-        metadata: ImmutableDict,
+        metadata: Mapping,
```
MebinAbraham, line 20: Worth doing the first level. Need to import `Any`.
```suggestion
        metadata: Mapping[str, Any],
```

app/data_models/answer.py
petechd: Fixed now, thanks!

app/helpers/template_helpers.py
```diff
@@ -200,7 +200,7 @@ def render_template(template: str, **kwargs: Union[str, Mapping]) -> str:
     template = f"{template.lower()}.html"
 
     session_expires_at = (
-        session_store.expiration_time.isoformat()
+        session_store.expiration_time.isoformat()  # type: ignore
         if (session_store := get_session_store())
         else None
     )
```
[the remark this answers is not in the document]
petechd: Added, thanks!

MebinAbraham (approved):
👍

MebinAbraham:
I think going forward, we should update the type hint guide, to suggest that any type ignore must a comment explaining why it has been done.

jamesnunn (approved):

PR #785, opened 2022-02-04
Conditional trad as

LJBabbage:
This PR adds the transform conditional_trad_as.

### How to review 
Using test_placeholder_transform.json
1. Provide a trad_as in launcher and confirm it appears in the question and summary correctly
2. Don't provide a trad_as and confirm it doesn't appear in the question and summary correctly
3. (optional) as kind of the same as 2 as we deal with empty strings in optional metadata the same as not passing them. Remove trad_as from the metadata and confirm it doesn't appear in the question and summary correctly

### Checklist

### DIFF
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -315,3 +315,9 @@ class PlaceholderTransforms:
         label may be simple string or a resolved placeholder string
         """
         return self.ops_helper.get_option_label_from_value(value, answer_id)
+
+    @staticmethod
+    def conditional_trad_as(trad_as: str) -> str:
+        if trad_as:
+            return f" ({trad_as})"
+        return ""
--- a/app/questionnaire/placeholder_transforms.py
+++ b/app/questionnaire/placeholder_transforms.py
@@ -317,7 +317,5 @@ class PlaceholderTransforms:
         return self.ops_helper.get_option_label_from_value(value, answer_id)
 
     @staticmethod
-    def conditional_trad_as(trad_as: str) -> str:
-        if trad_as:
-            return f" ({trad_as})"
-        return ""
+    def conditional_trad_as(trad_as: Optional[str]) -> str:
+        return f" ({trad_as})" if trad_as else ""
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=conditional-trad-as
+tag=latest
 TAG=${tag} docker-compose -f docker-compose-schema-validator.yml up -d
--- a/scripts/run_validator.sh
+++ b/scripts/run_validator.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=conditional-trad-as
 TAG=${tag} docker-compose -f docker-compose-schema-validator.yml up -d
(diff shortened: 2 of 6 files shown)


tests/integration/questionnaire/test_questionnaire_placeholders.py
```diff
@@ -50,11 +52,35 @@ def test_placeholders_rendered_in_pages(self):
...
         self.assertInUrl(SUBMIT_URL_PATH)
-        self.assertInBody("Please enter the total retail turnover")
+        self.assertInBody(
+            "For Integration Testing (Integration Tests), please enter the total retail turnover"
+        )
         self.assertInBody("Please enter the value of internet sales")
         self.assertInBody("Please enter the number of items")
         self.assertInBody("Do you want to add a 3rd item?")
 
+    def test_conditional_trad_as_without_trad_as_in_token(self):
```
LJBabbage, line 62: I don't believe this needs a functional tests and should only be integration as it is only adjusting a string and we don't need the expense of another functional test, when this is far more appropriate. It isn't pytest just because my thinking is the whole class needs doing for one test and it would also drag in integration_test_case.py. Naturally I can do that, but the refactor would be more lines than the rest put together 😆

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=conditional-trad-as
```
LJBabbage, line 2: To revert on merge

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -417,3 +417,15 @@ def test_option_label_from_value_with_placeholder_label(
     actual = placeholder_transform.option_label_from_value(value, answer_id)
 
     assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "trad_as, expected",
+    [
+        ("trad_as", " (trad_as)"),
```
jamesnunn, line 425: I know it sounds silly because it's arbitrary, but maybe make this something other than the variable name in the actual transform to avoid the test passing in case the f-string gets borked and it literally returns `" (trad_as)"`

jamesnunn (approved):

app/questionnaire/placeholder_transforms.py
```diff
@@ -315,3 +315,9 @@ def option_label_from_value(self, value: str, answer_id: str) -> str:
         label may be simple string or a resolved placeholder string
         """
         return self.ops_helper.get_option_label_from_value(value, answer_id)
+
+    @staticmethod
+    def conditional_trad_as(trad_as: str) -> str:
```
parimooa, line 320: `trad_as `can be optional
```suggestion
    def conditional_trad_as(trad_as: Optional[str]) -> str:
```

app/questionnaire/placeholder_transforms.py
```diff
@@ -315,3 +315,9 @@ def option_label_from_value(self, value: str, answer_id: str) -> str:
...
+        if trad_as:
+            return f" ({trad_as})"
+        return ""
```
parimooa, line 323: could be one liner if you like it 
```suggestion
       return  f" ({trad_as})" if trad_as else ""        
```

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -417,3 +417,15 @@ def test_option_label_from_value_with_placeholder_label(
     actual = placeholder_transform.option_label_from_value(value, answer_id)
 
     assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "trad_as, expected",
+    [
+        ("Apple", " (Apple)"),
+        (None, ""),
```
parimooa, line 426: Worth adding empty string `"" `as input

parimooa (commented):
Few minor comments, LGTM

parimooa (approved):

PR #786, opened 2022-02-04
Pytest helper framework module

parimooa:
Remove `framework` tests - not being used
Migrate helper tests to pytest

### How to review 

### Checklist


tests/app/helpers/conftest.py
```diff
@@ -1,12 +1,194 @@
...
+                is_post_submission=is_post_submission,
+                include_csrf_token=include_csrf_token,
+                survey_config=survey_config,
+            )
 
-    return _app
+    return _context_helper
+
+
+@fixture(name="footer_context")
```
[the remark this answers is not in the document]
jamesnunn: Would be useful to know where that object is coming from though? Seems odd for the linter to detect that from nowhere apparently.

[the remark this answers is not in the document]
parimooa: Linter doesn't understand that fixture can be re-used in another fixture. hence it flags it as that you are using variables (fixture) as global and local (within another fixture). See this https://stackoverflow.com/questions/46089480/pytest-fixtures-redefining-name-from-outer-scope-pylint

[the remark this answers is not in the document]
jamesnunn: Of course, within the same file it thinks function argument clobbers the fixture.

jamesnunn (approved):

tests/app/helpers/conftest.py
```diff
@@ -1,12 +1,194 @@
...
+        "newTabWarning": "The following links open in a new tab",
+        "copyrightDeclaration": {
+            "copyright": "Crown copyright and database rights 2020 OS 100019153.",
+            "text": "Use of address data is subject to the terms and conditions.",
+        },
+    }
+
+
+@fixture
+def expected_footer_census_theme(footer_context):
```
MebinAbraham, line 36: Personally I'm not a fan of these living here cause it not very clear from the test body, I would prefer if it was in the file it was tests as it's not common to all helpers. Would be nice if it was in its own conftest in its own folder. Not a blocker, don't have a perfect suggestion.

MebinAbraham (approved):

tests/app/helpers/conftest.py
parimooa: I prefer having data separate from tests. This keeps tests clean and readable. Having 100 lines of json in tests makes it quite noisy and imagine scrolling up and down these tests. Keep them in fixture reduced no of lines which are common across all json's. Happy to bring it back if anyone else agrees to it

MebinAbraham: I agree with that, I wasn't that it should be inline in the test body, it was more than it prob should be in a related file. i.e this conftest is for all helpers, but these snippets are only relevant to template helper tests.

parimooa: Keeping in fixture means it loads it automatically without needing any imports (if we keep it in separate file)

PR #791, opened 2022-02-14
Pytest app forms

jamesnunn:
Migrates `unittest` fixtures to `pytest` for `apps/forms`

### How to review 
Check tests pass, ensure coverage hasn't decreased

### Checklist


jamesnunn:
[Migrate app/forms (top level only) to pytest (2 SP)](https://trello.com/c/gmwmh7sN/5126-migrate-app-forms-top-level-only-to-pytest-2-sp)

tests/app/forms/test_duration_form.py
```diff
@@ -1,114 +1,112 @@
...
+        (False, None, "", True, None),
+        (True, None, "", False, "Enter a duration"),
+        (False, None, "word", False, "Enter a valid duration"),
+        (True, None, "word", False, "Enter a valid duration"),
+        (False, None, "-1", False, "Enter a valid duration"),
+        (True, None, "-1", False, "Enter a valid duration"),
+        (False, None, "12", True, None),
+        (True, None, "12", True, None),
+    ),
+)
```
petechd, line 85: Nice that you consolidated all this into one.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -16,1677 +16,1678 @@
...
-    AppContextTestCase
-):  # noqa: C901  pylint: disable=too-many-public-methods
-    @staticmethod
-    def _error_exists(answer_id, msg, mapped_errors):
-        error_id = f"{answer_id}-error"
-        return any(
-            e_id == error_id and str(msg) in ordered_errors
-            for e_id, ordered_errors in mapped_errors
-        )
+def _error_exists(answer_id, msg, mapped_errors):
```
petechd: Wasn't the underscore prefix meant as a hint that this method is intended for internal use? Since its class doesn't exist anymore we can rename it.

jamesnunn: I suppose it's still a private-ish function, it's still the same concept as having it in a class. It's harmless either way - it's a test module and unlikely to be imported ever

tests/app/forms/test_questionnaire_form.py
```diff
@@ -16,1677 +16,1678 @@
...
+        question_schema = schema.get_block("name-block").get("question")
 
-            for answer_id in schema.get_answer_ids_for_block("name-block"):
-                self.assertTrue(hasattr(form, answer_id))
+        form = generate_form(
+            schema,
+            question_schema,
+            answer_store,
+            list_store,
+            metadata=None,
```
parimooa: Since you are now refactoring, might as well fix this.  It is expecting `Mapping` 
```suggestion
            metadata={},
```

tests/app/forms/test_questionnaire_form.py
```diff
@@ -16,1677 +16,1678 @@
...
+            "csrf_token": None,
+            "date-range-from-answer": "2016-03-01",
+            "date-range-to-answer": "2016-03-31",
+        }
+        form = generate_form(
+            schema,
+            question_schema,
+            answer_store,
+            list_store,
+            metadata=None,
```
parimooa: There are few of them , you can simply find and replace
```suggestion
            metadata={},
```

tests/app/forms/test_questionnaire_form.py
```diff
@@ -16,1677 +16,1678 @@
...
 
-            expected_form_data = {
-                "csrf_token": None,
-                "date-range-from-answer": "2016-12-25",
-                "date-range-to-answer": "2016-12-24",
+        question_schema = schema.get_block("date-block").get("question")
+
+        form_data = MultiDict(
+            {
+                "date-range-from-answer-day": "25",
```
parimooa: Consider making them global/fixture/file as they seem to be repeated with same data in couple of tests. I'm ok if you keep it as is.

parimooa (commented):
LGTM, few minor comments. Few of test cases use same input data. May be worth reusing them to avoid duplication.

tests/app/forms/test_questionnaire_form.py
petechd: I was thinking we should do this here since you renamed `_test_validation` in `test_duration_form.py` but it's not crucially important.

tests/app/forms/test_field_factory.py
```diff
@@ -1,68 +1,66 @@
...
-            "ru_ref": "432423423423",
-            "ru_name": "Apple",
-            "return_by": "2016-07-07",
-            "case_id": "1234567890",
-            "case_ref": "1000000000000001",
         }
+    )
 
-        response_metadata = {}
+    metadata = {
```
petechd, line 20: We should probably make `metadata` an empty dict similar to `response metadata` as it's not needed for the test and passes without any key/value pairs in it (not required  for `ValueSourceResolver` and  `RuleEvaluator` to work).

parimooa (approved):

petechd (approved):

PR #793, opened 2022-02-16
Countdown thank you page

parimooa:

Add **countdown** on thank you page to view/download survey answers

Also:
- Fixes a few typos
- Updates to the latest version of the DS

#### With Javascript enable

1. Load design system changes ( new component `timeout-panel` )
2. Launch `test_view_submitted_response` or `test_thank_you survey`
3. Submit Answer , A countdown timer of 45 minutes should start counting down.
4. Respondent should be able to view/download answer during the countdown period
5. update `VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS=120`(restart runner) and repeat step 1 and 2 , 2 minutes countdown timer should start counting down
6. Wait for timer to run out and observe the page refreshes and user  doesn't have option of download or view answers

#### With Javascript disabled

1. Launch test_view_submitted_response or test_thank_you survey
2. Disable JS from chrome dev settings
3. submit answers
4. Observe there is no countdown timer, the message is static 
 
### Checklist


tests/app/questionnaire/test_questionnaire_submission.py
```diff
@@ -28,6 +30,7 @@ def _launch_and_submit_questionnaire(self):
         self.post()
         self.post()
 
+    @pytest.mark.xfail(reason="DS changes pending")
```
parimooa, line 33: Remove `xfail` markers when DS changes are merged

templates/thank-you.html
```diff
@@ -46,24 +47,31 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
...
+    {{ onsTimeoutPanel ({
+        "id": "view-submitted-response-countdown",
+        "countdown": {
+            "countdownInSeconds":content.view_submitted_response.countdown_seconds,
+            "minutesTextSingular":"minute",
+            "minutesTextPlural":"minutes",
+            "secondsTextSingular":"second",
+            "secondsTextPlural":"seconds",
+            "countdownText":"For security, your answers will only be available to view for another",
+            "nojsText": "For security, your answers will only be available to view for another " + content.view_submitted_response.countdown_minutes|string + " minutes",
```
parimooa, line 64: Added additional property of `countdown_minutes` instead of hard coded 45 minutes. This will be value set `VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS` converted to minutes

jamesnunn (commented):
LGTM but waiting for DS changes to test

templates/thank-you.html
```diff
@@ -46,24 +47,31 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
...
-    {% else %}
-      {% call onsPanel({
+      {{ onsTimeoutPanel ({
+          "id": "view-submitted-response-countdown",
+          "minutesTextSingular": _("minute"),
+          "minutesTextPlural": _("minutes"),
+          "secondsTextSingular": _("second"),
+          "secondsTextPlural": _("seconds"),
+          "countdownText": _("For security, your answers will only be available to view for another"),
+          "nojsText": _("For security, your answers will only be available to view for another ") + content.view_submitted_response.expiry_minutes|string + _(" minutes"),
```
LJBabbage, line 62: If these initial statements are the same, might be worth making a var.

templates/thank-you.html
```diff
@@ -46,24 +47,31 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
...
-        <p>{{ _("For security your answers will only be available to view for 45 minutes") }}</p>
-      {% endcall %}
-    {% else %}
-      {% call onsPanel({
+      {{ onsTimeoutPanel ({
+          "id": "view-submitted-response-countdown",
+          "minutesTextSingular": _("minute"),
+          "minutesTextPlural": _("minutes"),
+          "secondsTextSingular": _("second"),
+          "secondsTextPlural": _("seconds"),
```
LJBabbage, line 60: I find this odd, is it done purely for translation? As I can't think when minutesTextSingular would ever be anything other than minute

templates/thank-you.html
```diff
@@ -46,24 +47,31 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
...
+    {% endif %}
+  {% else %}
+    {% call onsPanel({
         "id": "view-submitted-response-guidance",
         "classes": "ons-u-mb-m"
       }) %}
-        {{ _("For security, you can no longer view your answers") }}
-      {% endcall %}
-    {% endif %}
+      {{ _("For security, you can no longer view or get a copy of your answers") }}
```
LJBabbage, line 73: This again looks like a duplication from above

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
```diff
@@ -21,6 +21,8 @@ describe("Feature: Repeating Sections with Hub and Spoke", () => {
   describe("Given the user has added some members to the household and is on the Hub", () => {
     before("Open survey and add household members", () => {
       browser.openQuestionnaire("test_repeating_sections_with_hub_and_spoke.json");
+      // Accept cookies, since they push submit button outside window
```
LJBabbage, line 24: Sorry, I don't quite understand this comment and how this relates to this story

tests/functional/spec/thank_you.spec.js
```diff
@@ -52,20 +52,19 @@ describe("Thank You Default View Response Enabled", () => {
       expect(browser.getUrl()).to.contain(ThankYouPage.pageName);
     });
 
-    it("When I navigate to the thank you page, and I have submitted less than 45 minutes ago, Then I should see the option to view my answers", () => {
+    it("When I navigate to the thank you page, and I have submitted less than 40 seconds ago, Then I should see the option to view my answers and countdown timer before it expires", () => {
```
LJBabbage, line 55: Do you need the "before it expires"? It's a set time in your when and you wouldn't put after in the other test. Also might need a "the" in before countdown

LJBabbage (changes requested):
Looks good, some minor stuff. You also said this updates chromedriver to 99, but I can't see that

scripts/load_print_styles_from_cdn.sh
```diff
@@ -8,7 +8,7 @@ if [ -z ${1+x} ]; then
   exit 1
 fi
 
-CDN_URL=${CDN_URL:-"https://cdn.eq.gcp.onsdigital.uk"}
+CDN_URL=${CDN_URL:-"https://cdn.ons.gov.uk"}
```
MebinAbraham, line 11: This highlights that we weren't passing a CDN_URL and in formal envs, we were building using dev cdn.

templates/thank-you.html
MebinAbraham, line 62: Not related to Lewis comments, but can't make new comments separately. This needs to be double-checked.

- This currently wouldn't work for when < 2 minutes are left, as it would say `... for another 1 minutes` or `... for another 0 minutes`. This can be solved by using `ngettext` which can support pluralisation. It still would work for < 60 seconds left without additional logic. Worth clarifying with Donna whether we need this, or for non-js it should just say `For security your answers will only be available to view for 45 minutes`
- If the decision is to show the remaining time, then the string concatenation should use `.format()` otherwise, the extracted string will not be friendly to the translator. It should be done in the same way as: https://github.com/ONSdigital/eq-questionnaire-runner/blob/0881539fa56052d07e06f5f4d44a8007661dfd68/templates/thank-you.html#L53

tests/functional/spec/thank_you.spec.js
```diff
@@ -52,20 +52,19 @@ describe("Thank You Default View Response Enabled", () => {
       expect(browser.getUrl()).to.contain(ThankYouPage.pageName);
```
MebinAbraham, line 52: So these will test with JS, I don't think we have any tests to ensure the content is as expected for non-js version?. Previously it didn't matter since the content was the same between the two.

templates/thank-you.html
MebinAbraham: Ya, it's for translations.

.functional-tests.env
```diff
@@ -27,4 +27,4 @@ CDN_ASSETS_PATH=/design-system
 ADDRESS_LOOKUP_API_URL=https://whitelodge-ai-api.census-gcp.onsdigital.uk
 COOKIE_SETTINGS_URL=#
 EQ_SUBMISSION_CONFIRMATION_BACKEND=log
-VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS=45
+VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS=35
```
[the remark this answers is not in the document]
parimooa: To minimise countdown and overall execution time

templates/thank-you.html
parimooa: As per contract with DS panel/modal, this needs to be there. we don't control this. probably best to raise with DS.

app/views/contexts/thank_you_context.py
```diff
@@ -45,11 +45,22 @@ def build_thank_you_context(
...
     view_submitted_response = {"enabled": schema.is_view_submitted_response_enabled}
 
     if schema.is_view_submitted_response_enabled:
+        view_submitted_expiry_seconds = current_app.config[
+            "VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS"
+        ]
+        response_expires_at = submitted_at + timedelta(
+            seconds=view_submitted_expiry_seconds
+        )
         expired = has_view_submitted_response_expired(submitted_at)
```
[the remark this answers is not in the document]
parimooa: We still need `VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS` for calculation of `nojsText` minutes , don't see reason to have another function if it can be done within the function

templates/thank-you.html
parimooa: You are correct, it wont reflect the exact expiry as it will round to nearest int. i think it is not worth to put extra effort on this one. best to use hard coded 45 minutes as ticket suggests. i was thinking it would have been nice to use what the env var (`VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS`) is set to which is more likely to be `2700`. This case is also an edge case . if you have JS disabled then you can't even proceed with survey.

templates/thank-you.html
parimooa: There was a bug which is fixed now. it is required when submitted response is enabled and session has expired. I have updated the condition

templates/thank-you.html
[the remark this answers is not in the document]
parimooa: updated to hardcoded 45 minutes , it is not worth to calculate minutes/seconds for `nojsText` which is an edge case

scripts/load_print_styles_from_cdn.sh
parimooa: Didn't get this , can you elaborate on this please

tests/functional/spec/thank_you.spec.js
parimooa: it appears that once cannot update the browser option during runtime. so can't test this option. If we disable before launching surveys then we can't do any survey.

app/views/contexts/thank_you_context.py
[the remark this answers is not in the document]
parimooa: i used your suggestion as the `nojsText` is a hardcoded 45 minutes text

tests/functional/spec/thank_you.spec.js
MebinAbraham: Not sure what you mean. The Non JS test would live as an integration test which has no JS. Runner doesn't require JS to work.

templates/thank-you.html
```diff
@@ -46,25 +47,36 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
...
-        {{ _("For security, you can no longer view your answers") }}
-      {% endcall %}
+      {{ onsTimeoutPanel ({
+          "id": "view-submitted-response-countdown",
+          "minutesTextSingular": _("minute"),
+          "minutesTextPlural": _("minutes"),
+          "secondsTextSingular": _("second"),
+          "secondsTextPlural": _("seconds"),
+          "countdownText": countdown_text,
+          "nojsText": countdown_text + _("45 minutes"),
```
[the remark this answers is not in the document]
MebinAbraham: The non JS text identical either, it should be `available to view for 45 minutes` vs timer `available to view for another x minutes`

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
parimooa: For some reason the submit button pushed the window down and it couldn't submit so we had to accept cookie banner to bring that submit button within window. it is weird only this test was failing because of that

templates/thank-you.html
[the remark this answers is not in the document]
parimooa: it is identical for prod , in tests we have lesser time (40 secs) so it won't be same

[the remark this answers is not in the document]
MebinAbraham: I wasn't referring to the time. The wording isn't exact, for the JS version it includes the word `another`.

tests/functional/spec/thank_you.spec.js
parimooa: added integration test

rmccar:
> Looks good, some minor stuff. You also said this updates chromedriver to 99, but I can't see that

We did update to 99 but I think another PR must've merged into master that did the same, so now it looks like our PR didn't. Ill remove it from the description

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
LJBabbage: It's the EQ_RUN_FUNCTIONAL_TESTS_HEADLESS issue again, Doing this to one test isn't the best. I would prefer we updated wdio.config window-size if it works, although still not ideal, because in time  the reason for doing this is lost. Even a comment saying this has been done down to a headless window-size issue would be better

tests/integration/routes/test_thank_you.py
```diff
@@ -103,3 +103,13 @@ def test_back_to_surveys_link_on_thank_you(self):
         self.assertInUrl("thank-you")
         self.assertInBody("Back to surveys")
         self.assertInBody(ACCOUNT_SERVICE_TODO_PATH)
+
+    def test_answer_availability_text(self):
```
MebinAbraham, line 107: I think something like `test_view_answers_after_submission_guidance` would be clearer.

templates/thank-you.html
```diff
@@ -46,25 +47,36 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
...
+      {{ onsTimeoutPanel ({
+          "id": "view-submitted-response-countdown",
+          "minutesTextSingular": _("minute"),
+          "minutesTextPlural": _("minutes"),
+          "secondsTextSingular": _("second"),
+          "secondsTextPlural": _("seconds"),
+          "countdownText": countdown_text,
+          "nojsText": _("For security, your answers will only be available to view for 45 minutes"),
+          "redirectUrl":  url_for("post_submission.get_thank_you"),
+          "countdownExpiredText": countdown_expired_text,
```
MebinAbraham, line 67: Just an observation. Slightly odd that this gets shown, then redirect happens. Instead of redirect on zero. This I would think should be configurable if `redirectUrl` wasn't provided. One for DS, but thought I'd highlight in case there was some reasoning behind it.

templates/thank-you.html
```diff
@@ -46,25 +47,36 @@ <h1 data-qa="title">{{ _("Thank you for completing the {survey_title}").format(
     <a href="https://www.ons.gov.uk/surveys">https://www.ons.gov.uk/surveys</a></p>
   {% endif %}
 
-  {% if content.view_submitted_response.enabled %}
+  {% set countdown_expired_text = _("For security, you can no longer view or get a copy of your answers") %}
+  {% set countdown_text = _("For security, your answers will only be available to view for another ") %}
```
MebinAbraham, line 51: This statement should be inside the if statement that uses it.

.design-system-version
```diff
@@ -1 +1 @@
-45.2.0
+46.1.3
```
MebinAbraham, line 1: For some reason, this isn't available via the AWS CDN, which means this can't go to preprod until it is fixed. Try to change the local CDN to AWS one, all the assets will be broken. Will progress via Slack.

templates/thank-you.html
parimooa: Yes i noticed that too, it shows it briefly. I removed `redirectUrl` , but it still shows up. One to raise with DS

.design-system-version
MebinAbraham: Will investigate separately why 46.1.3 is now available.

tests/functional/spec/features/repeating_sections/repeating_sections_with_hub_and_spoke.spec.js
MebinAbraham: If the comment is staying, it should mention the banner, otherwise it is a bid confusing as to why cookies has anything to do with it. The significance being the banner pushes the button outside the windows.

parimooa: updated comment now , it is headless window size issue again. I tried with few window size combination. it works on local but fails in actions

MebinAbraham (approved):

jamesnunn (approved):

LJBabbage (approved):

PR #797, opened 2022-02-21
Summary to display default when question is unanswered

petechd:
When onboarding QBS, it was noticed that when the respondent leaves a question unanswered, it doesn't display default answer value on summary page. This PR attempts to fix the problem.

### How to review 
There is one new unit test, one fixed functional test and one brand new functional test that checks exactly the behaviour described on the card. 

### Checklist


petechd:
[Summary to display default when question is unanswered (1 SP)](https://trello.com/c/yIS67qYQ/5137-summary-to-display-default-when-question-is-unanswered-1-sp)

parimooa (approved):
LGTM, working functionally

app/views/contexts/summary/question.py
```diff
@@ -39,7 +39,16 @@ def __init__(
         )
 
     def _get_answer(self, answer_store, answer_id):
-        return answer_store.get_escaped_answer_value(answer_id, self.list_item_id)
+        if answer_store.get_escaped_answer_value(
```
LJBabbage, line 42: This doesn't need to call answer_store.get_escaped_answer_value(answer_id, self.list_item_id) 3 times. Just assign it a var at the start

app/views/contexts/summary/question.py
```diff
@@ -39,7 +39,16 @@ def __init__(
...
+            answer_id, self.list_item_id
+        ) or isinstance(
+            answer_store.get_escaped_answer_value(answer_id, self.list_item_id),
+            int,  # protection against falsy answer value "0"
+        ):
+            return answer_store.get_escaped_answer_value(answer_id, self.list_item_id)
+        if self.schema.get_default_answer(answer_id):
+            return self.schema.get_default_answer(answer_id).value
+        return None
```
LJBabbage, line 51: What is coming back from the default answer? As you should be just be able to just return self.schema.get_default_answer(answer_id) and forget the rest

tests/app/views/contexts/summary/test_question.py
```diff
@@ -2,10 +2,88 @@
 
 from app.data_models import Answer, ListStore
 from app.data_models.answer_store import AnswerStore
+from app.questionnaire import QuestionnaireSchema
 from app.questionnaire.rules.rule_evaluator import RuleEvaluator
 from app.views.contexts.summary.question import Question
 from tests.app.app_context_test_case import AppContextTestCase
 
+ADDRESS_QUESTIONNAIRE_SCHEMA = QuestionnaireSchema(
```
LJBabbage, line 10: Would like this match the pattern in the rest the file (i.e def dynamic_answer_options_schema())

tests/app/views/contexts/summary/test_question.py
```diff
@@ -147,13 +225,17 @@ def test_create_question_with_answer_label_when_empty_title(self):
...
+        self.schema = ADDRESS_QUESTIONNAIRE_SCHEMA
 
         for concatenation_type, concatenation_character in answer_separators.items():
             with self.subTest(
                 concatenation_type=concatenation_type,
                 concatenation_character=concatenation_character,
             ):
                 # Given
+                self.answer_store.add_or_update(
+                    Answer(answer_id="building", value="Government Buildings")
```
LJBabbage, line 237: Would like to see this distinctly different to the default

tests/app/views/contexts/summary/test_question.py
```diff
@@ -216,14 +306,107 @@ def test_concatenate_textfield_answers(self):
...
+
+                # Then
+                self.assertEqual(
+                    question.answers[0]["value"],
+                    f"Government Buildings{concatenation_character}Cardiff Rd{concatenation_character}Newport{concatenation_character}NP10 8XG",
+                )
+                self.assertEqual(len(question.answers), 1)
+
+    def test_concatenate_textfield_answers_default(self):
+        answer_separators = {"Newline": "<br>", "Space": " "}
```
LJBabbage, line 321: A bit of duplication in this file now with the QuestionnaireSchema and the 2 tests having sharing all the same chunks 
```
{
                    "id": "building",
                    "label": "Building",
                    "mandatory": False,
                    "type": "TextField",
                    "default": "Government Buildings",
                }
```

LJBabbage (changes requested):

app/views/contexts/summary/question.py
petechd: It returns an Answer class instance with `answer_id`, `list_item_id` and `value` so we only need `value` key here.

LJBabbage: Ok, but you could also tidy this up a little, something like 
```    
def _get_answer(self, answer_store, answer_id):
        escaped_answer_value = answer_store.get_escaped_answer_value(answer_id, self.list_item_id)

        if escaped_answer_value or isinstance(escaped_answer_value, int):
            return escaped_answer_value

        default_answer = self.schema.get_default_answer(answer_id)
        return default_answer.value if default_answer else None
```

LJBabbage: Do you even need instance, isn't it just for 0?

app/views/contexts/summary/question.py
```diff
@@ -39,7 +39,16 @@ def __init__(
         )
 
     def _get_answer(self, answer_store, answer_id):
-        return answer_store.get_escaped_answer_value(answer_id, self.list_item_id)
+        if answer_store.get_escaped_answer_value(
+            answer_id, self.list_item_id
+        ) or isinstance(
+            answer_store.get_escaped_answer_value(answer_id, self.list_item_id),
+            int,  # protection against falsy answer value "0"
+        ):
```
MebinAbraham, line 47: This should be avoided as we should check for null values. I think we should scrap `get_escaped_answer_value` method, it is only used here. Makes sense if it was used in multiple places, but value source handles it explcity. I would therefore simplify this to something like:

```py
    def _get_answer(self, answer_store, answer_id):
        answer = answer_store.get_answer(
            answer_id, self.list_item_id
        ) or self.schema.get_default_answer(answer_id)

        return escape_answer_value(answer.value) if answer else None
```

If you use walrus, you could do `escape_answer_value` just do non default but it doesn't really matter.

tests/app/views/contexts/summary/test_question.py
```diff
@@ -141,6 +64,157 @@ def get_question_schema(answer_schema):
...
+                                ],
+                                "id": "address-group",
+                            }
+                        ],
+                    }
+                ]
+            }
+        )
+
+    def address_question(self, concatenation_type):
```
LJBabbage, line 146: Aren't these answers just a copy of address_questionnaire_schema?

app/views/contexts/summary/question.py
petechd: Ok, removed all repetition.

tests/app/views/contexts/summary/test_question.py
petechd: Changed, so it has different value than default now. Thanks!

tests/app/views/contexts/summary/test_question.py
petechd: Repetition removed now, no repeating chunks!

tests/app/views/contexts/summary/test_question.py
petechd: using the ones in `address_questionnaire_schema` now, no repetition. Thanks!

app/views/contexts/summary/question.py
petechd: So I had to chop it up a little bit otherwise it was never assigning answer the default value cause Answer class instance always exists but not always with value but this code didn't check that, now it is:
```py
def _get_answer(self, answer_store, answer_id):
        answer = answer_store.get_answer(answer_id, self.list_item_id)
        if answer and answer.value:
            return escape_answer_value(answer.value)
        if default_answer := self.schema.get_default_answer(answer_id):
            return default_answer.value
```
I scrapped `get_escaped_answer_value` from answer store as well.

MebinAbraham: `get_answer` will return `None` if the answer didn't exist?. So it will already be handled. Also, you don't want to check the value as value can be falsey. If you get an answer object back, we return the value.

petechd: I had to add zero check in `_get_answer` as we still need to check for `answer.value` and it evaluates `0` to `False`.

petechd: If we don't check if `answer.value` is empty string (which evaluates to `False` as you know) then it will never hit `default_value` as it will always pass empty string to the summary.

tests/app/views/contexts/summary/test_question.py
```diff
@@ -77,6 +78,126 @@ def get_question_schema(answer_schema):
...
+        question_schema = {
+            "id": "question_id",
+            "title": "question_title",
+            "type": "General",
+            "answers": answers,
+            "summary": {"concatenation_type": concatenation_type},
+        }
+
+        return Question(
+            question_schema,
```
MebinAbraham, line 178: You basically already have a full schema, so I would update `address_questionnaire_schema` to take `concatenation_type` as a param, then this can be simplied to:
```py
  question_schema = self.schema.get_questions("what-is-your-address-question")[0]
  return Question(
     question_schema,
```

You will have to move `self.schem` to inside the subtest loop and pass in the concat.
```
self.schema = self.address_questionnaire_schema(concatenation_type)
```

tests/app/views/contexts/summary/test_question.py
```diff
@@ -164,84 +285,45 @@ def test_create_question_with_answer_label_when_empty_title(self):
...
-                    "mandatory": False,
-                    "type": "TextField",
-                }
-                postcode = {
-                    "id": "postcode",
-                    "label": "Postcode",
-                    "mandatory": False,
-                    "type": "TextField",
-                }
+    def test_concatenate_textfield_answers_default(self):
```
MebinAbraham, line 308: The other issue with testing concat is that, non concat isn't tested. Personally, I would make `_get_answer` non-private and test that, otherwise you will need another test for non concat with default.

tests/app/views/contexts/summary/test_question.py
```diff
@@ -164,84 +285,45 @@ def test_create_question_with_answer_label_when_empty_title(self):
...
                 # Given
-                self.answer_store.add_or_update(
-                    Answer(answer_id="address-line-1", value="Cardiff Rd")
-                )
-                self.answer_store.add_or_update(
-                    Answer(answer_id="town-city", value="Newport")
-                )
-                self.answer_store.add_or_update(
-                    Answer(answer_id="postcode", value="NP10 8XG")
+                self.update_answer_store(False)
```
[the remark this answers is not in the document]
petechd: Removed the method and moved it inline, so clearer what it does, thanks.

tests/app/views/contexts/summary/test_question.py
petechd: `_get_answer` non private and tested in non concat now. Cheers!

tests/app/views/contexts/summary/test_question.py
petechd: Ok, added all your suggestions to the test, all simplified and working good. Thanks.

tests/app/views/contexts/summary/test_question.py
```diff
@@ -1098,7 +1087,7 @@ def test_get_answer_method(self):
                 question = Question(
                     question_schema,
                     answer_store=answer_store,
-                    schema=self.address_questionnaire_schema(),
+                    schema=self.address_questionnaire_schema("Newline"),
```
MebinAbraham, line 1090: Instead of this, I would put `self.schema = self.address_questionnaire_schema(concatenation_type)` inside this sub test.

petechd: Moved it up to top of sub test, I hope that's what you meant?

LJBabbage (approved):

tests/app/views/contexts/summary/test_question.py
```diff
@@ -969,3 +1047,58 @@ def test_dynamic_answer_options(self):
...
+                    rule_evaluator=self.get_rule_evaluator(),
+                    value_source_resolver=self.get_value_source_resolver(),
+                    location=None,
+                    block_id="address-group",
+                    return_to=None,
+                )
+
+                # Then
+                self.assertEqual(
+                    Question.get_answer(question, answer_store, "building"),
```
MebinAbraham, line 1102: How come this is not using the instance from above?.

tests/app/views/contexts/summary/test_question.py
```diff
@@ -969,3 +1047,58 @@ def test_dynamic_answer_options(self):
 
                 # Then
                 self.assertEqual(question.answers[0]["value"], expected_output)
+
+    def test_get_answer_method(self):
```
MebinAbraham, line 1051:
```suggestion
    def test_get_answer(self):
```

MebinAbraham, line 1051: We should have one to ensure the answer is also escaped.

tests/app/views/contexts/summary/test_question.py
```diff
@@ -181,67 +282,44 @@ def test_concatenate_textfield_answers(self):
...
+        answer_separators = {"Newline": "<br>", "Space": " "}
+
+        for concatenation_type, concatenation_character in answer_separators.items():
+            with self.subTest(
+                concatenation_type=concatenation_type,
+                concatenation_character=concatenation_character,
+            ):
+                # Given
+                self.schema = self.address_questionnaire_schema(concatenation_type)
+                self.answer_store.remove_answer("building")
```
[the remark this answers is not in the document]
petechd: Redundant, removed now, thanks.

tests/app/views/contexts/summary/test_question.py
petechd: Now it is using that instance, thanks!

tests/app/views/contexts/summary/test_question.py
petechd: Added one test for testing escaped result, thanks!

MebinAbraham (approved):

PR #807, opened 2022-03-03
Add guidance banner to introduction page

petechd:
Adds introduction page guidance banner as described on [this trello](https://trello.com/c/Dog3Cv1U). Tests and validator fix added.

### How to review 
New functional test added, do ` yarn test_functional --spec ./tests/functional/spec/introduction.spec.js`.

### Checklist


petechd:
[Banner on introduction page (3 SP)](https://trello.com/c/Dog3Cv1U/5169-banner-on-introduction-page-3-sp)

petechd:
Validator branch needs changing.

jamesnunn (approved):
Banner looks right to me per the card

templates/partials/contents.html
```diff
@@ -21,6 +22,18 @@ <h2>{{item.title}}</h2>
...
+  {%- if item.guidance -%}
+    {% set contents = item.guidance.contents %}
+      {% call onsPanel({
+      "id": "item-guidance-" ~ guidance_id,
+      "classes": "ons-u-mb-m"
+      }) %}
+        {% set contents = item.guidance.contents %}
+        {% include 'partials/contents.html' %}
+      {% endcall %}
+  {% endif %}
```
MebinAbraham, line 35: As per my comment on validator, we should handle this in `basic.html`. Introduction primary content is already an array so it gives them the flexibility they need which allowing guidance (panel) in generic content blocks.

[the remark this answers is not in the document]
rmccar: Ok I think all contents should be fixed because this doesn't currently fit the design for guidance panels. Also line 27 should removed because it is duplicated on line 32

schemas/test/en/test_introduction_with_guidance.json
```diff
@@ -0,0 +1,258 @@
+{
```
MebinAbraham, line 1: Since we are trying to reduce schema, I think we can scrap this and add guidance to the existing intro schema. It would be good to make it looks like the one attached to the card, with content above and below the guidance.

templates/partials/contents.html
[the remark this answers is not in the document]
petechd: Removed the duplication.

templates/partials/introduction/basic.html
```diff
@@ -1,7 +1,20 @@
+{% from "components/panel/_macro.njk" import onsPanel %}
+
 <div id="{{content_block.id}}" class="ons-u-mb-s">
   {% if content_block.title %}
     <h2>{{ content_block.title }}</h2>
   {% endif %}
   {% set contents = content_block.contents %}
+  {% set guidance_id = content_block.id %}
```
MebinAbraham: We should set this inside the guidance if check.

templates/partials/introduction/basic.html
```diff
@@ -1,7 +1,19 @@
...
+
+  {%- if content_block.guidance -%}
+    {% call onsPanel({
+      "id": "item-guidance-" ~ content_block.id,
+      "classes": "ons-u-mb-m"
+      }) %}
+      {% set contents = content_block.guidance.contents %}
+      {% include 'partials/contents.html' %}
+    {% endcall %}
+  {% endif %}
```
MebinAbraham, line 18: The order should be okay. Author will have to make use of multiple elements if they want guidance in between.

templates/partials/contents.html
MebinAbraham: Context for others. Moving back to contents, since we already allow definitions inside contents, which in itself is recursive. If it becomes a problen, we will revisit.

schemas/test/en/test_introduction_with_guidance.json
petechd: Moved to single schema now.

MebinAbraham (approved):

templates/partials/introduction/basic.html
```diff
@@ -1,3 +1,5 @@
+{% from "components/panel/_macro.njk" import onsPanel %}
```
rmccar, line 1: Think this line can be removed, other than that looks good

petechd: Removed, thanks!

rmccar (approved):

PR #809, opened 2022-03-04
Routing and skipping section dependencies

LJBabbage:
This PR allows when rules to references other sections in routing/skipping rules. Up to now routing and skipping was only possible inside the same section

### How to review 
Using test_new_routing_and_skipping_section_dependencies test with multiple different combinations, going back and forward changing things and see what happens. Your best bet is to look at the functional test combinations to see the possible routes. There is routing and skipping rules, nested skipping, dependencies which them have dependencies and repeat sections. Group skipping has also been fixed when referencing a value from a different section

N.B There is an outstanding issue with regards to completeness of a section when answers are changed.  Therefore it is possible to complete a section, change your answer which reveals a mandatory answer, but on clicking back into the that section you get the section summary with a mandatory answer say "No answer provided". 

### Checklist


app/questionnaire/path_finder.py
```diff
@@ -43,22 +44,69 @@ def routing_path(
...
+        return when_rule_block_id_dependencies
+
+    def _get_not_skipped_blocks_in_section(
+        self,
+        location: Location,
+        routing_path_block_ids: List[str],
+        section: ImmutableDict,
+        when_rule_block_id_dependencies: List[str],
+    ) -> List[Mapping]:
+        # :TODO: Fix group skipping in its own section. Routing path will be empty and therefore not checked
```
LJBabbage, line 93: This hasn't be introduced and a card has been raised

schemas/test/en/test_new_routing_and_skipping_section_dependencies.json
```diff
@@ -0,0 +1,549 @@
...
+                                },
+                                {
+                                    "==": [
+                                        {
+                                            "source": "answers",
+                                            "identifier": "skip-confirmation-answer"
+                                        },
+                                        "No"
+                                    ]
+                                }
```
LJBabbage, line 270: This might look a bit odd, but it is deliberate. I wanted to prove the answer was not on the path, even though it might have an answer in the store, we don't have an rule for that so this is the best I could do

app/questionnaire/questionnaire_schema.py
```diff
@@ -211,6 +211,16 @@ def get_section_ids_dependent_on_list(self, list_name: str) -> list[str]:
...
 
+    def get_section_ids_dependent_on_when_rules(
+        self, section: ImmutableDict
+    ) -> set[str]:
+        when_rules = self._get_values_for_key(section, "when")
+        rules: Union[dict, list] = next(when_rules, [])
+        section_ids_dependent_on_rules = self._get_section_ids_dependent_on_rules(
+            section["id"], rules
+        )
+        return section_ids_dependent_on_rules
```
petechd, line 222:
```suggestion
        return self._get_section_ids_dependent_on_rules(
            section["id"], rules
        )
```

tests/app/questionnaire/test_path_finder.py
```diff
@@ -535,3 +536,253 @@ def test_new_remove_answer_and_block_if_routing_backwards(list_store):
...
+    answer_store.add_or_update(
+        Answer(answer_id="first-name", value="John", list_item_id="lCIZsS")
+    )
+    answer_store.add_or_update(
+        Answer(answer_id="last-name", value="Smith", list_item_id="lCIZsS")
+    )
+
+    answer_store.add_or_update(Answer(answer_id="anyone-else", value="No"))
+
+    list_store = (ListStore([{"items": ["lCIZsS"], "name": "people"}]),)
```
petechd, line 748: PathFinder is expecting ListStore and here we get tuple, so maybe we should change it to `list_store = ListStore([{"items": ["lCIZsS"], "name": "people"}])`?

app/questionnaire/path_finder.py
```diff
@@ -43,22 +44,57 @@ def routing_path(
...
+
+            if blocks:
+                routing_path_block_ids = self._build_routing_path_block_ids(
+                    blocks, current_location, block_ids_dependent_on_when_rules
+                )
+
+        return RoutingPath(routing_path_block_ids, section_id, list_item_id, list_name)
+
+    def _get_block_ids_dependent_on_when_rules(
+        self, section: ImmutableDict
```
petechd, line 65: From [our type hints guide](https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/doc/python-type-hinting.md#generic-types): specify at least the top-level type parameters for all generic types?

app/questionnaire/path_finder.py
```diff
@@ -43,22 +44,57 @@ def routing_path(
...
+    def _get_not_skipped_blocks_in_section(
+        self,
+        location: Location,
+        routing_path_block_ids: List[str],
+        section: ImmutableDict,
+        block_ids_dependent_on_when_rule: List[str],
+    ) -> List[Mapping]:
+        # :TODO: Fix group skipping in its own section. Routing path will be empty and therefore not checked
+        if section:
+            not_skipped_blocks: List[Mapping] = []
```
petechd, line 83: Should use standard collection here?
```suggestion
            not_skipped_blocks: list[Mapping] = []
```

app/questionnaire/path_finder.py
LJBabbage: That is the type coming back from schema, what would you suggest?

app/questionnaire/questionnaire_schema.py
LJBabbage: Always a bit or personal choice, but I am not fussy so happy to change

app/questionnaire/path_finder.py
petechd: Comes from:
```python
self.schema.get_section(section_id)
```
and returns things like:
`ImmutableDict({'id': 'section', 'groups': (ImmutableDict({'id': 'group', 'title': 'List', 'blocks': (ImmutableDict({'type': 'Question', 'id': 'address-block-mandatory', 'question': ImmutableDict({'type': 'General', 'id': 'address-question-mandatory', 'title': 'What is your current address?', 'answers': (ImmutableDict({'id': 'address-mandatory', 'mandatory': True, 'type': 'Address'}),)})}), ImmutableDict({'type': 'Question', 'id': 'address-block-optional', 'question': ImmutableDict({'type': 'General', 'id': 'address-question-optional', 'title': 'What was your address one year ago?', 'answers': (ImmutableDict({'id': 'address-optional', 'mandatory': False, 'type': 'Address'}),)})}), ImmutableDict({'type': 'Question', 'id': 'address-confirmation', 'question': ImmutableDict({'type': 'General', 'id': 'address-confirmation-question', 'title': ImmutableDict({'placeholders': (ImmutableDict({'placeholder': 'household_address_line1', 'value': ImmutableDict({'identifier': 'address-mandatory', 'source': 'answers', 'selector': 'line1'})}),), 'text': 'Please confirm the first line of your address is {household_address_line1}'}), 'answers': (ImmutableDict({'type': 'Radio', 'id': 'address-confirmation-answer', 'mandatory': False, 'options': (ImmutableDict({'label': 'Yes', 'value': 'Yes'}), ImmutableDict({'label': 'No', 'value': 'No'}))}),)})}))}),)})`

so `ImmutableDict[str, Any]`?

tests/app/questionnaire/test_path_finder.py
```diff
@@ -535,3 +536,253 @@ def test_new_remove_answer_and_block_if_routing_backwards(list_store):
...
+            "test_routing_and_skipping_section_dependencies",
+            ["repeating-sex", "repeating-age"],
+        ),
+    ),
+)
+def test_routing_path_block_ids_dependent_on_when_rules_repeating(
+    skip_age_answer,
+    schema,
+    expected_route,
+):
```
MebinAbraham, line 732: Same deduplicate can be applied here:
```suggestion
@pytest.mark.parametrize(
    "skip_age_answer, expected_route",
    (
        (
            "Yes",
            ["repeating-sex"],
        ),
        (
            "No",
            ["repeating-sex", "repeating-age"],
        ),
    ),
)
@pytest.mark.parametrize(
    "schema",
    (
        [
            "test_new_routing_and_skipping_section_dependencies",
            "test_routing_and_skipping_section_dependencies",
        ]
    ),
)
def test_routing_path_block_ids_dependent_on_when_rules_repeating(
    schema,
    skip_age_answer,
    expected_route,
):
```

tests/app/questionnaire/test_path_finder.py
```diff
@@ -535,3 +536,253 @@ def test_new_remove_answer_and_block_if_routing_backwards(list_store):
...
+    progress_store = ProgressStore(progress)
+
+    # When
+    path_finder = PathFinder(
+        schema,
+        answer_store,
+        list_store,
+        progress_store,
+        {},
+        {},
```
MebinAbraham, line 691: Keyword args would help here.

tests/app/questionnaire/test_path_finder.py
```diff
@@ -535,3 +536,253 @@ def test_new_remove_answer_and_block_if_routing_backwards(list_store):
...
+    )
+
+    # When
+    path_finder = PathFinder(
+        schema,
+        answer_store,
+        list_store,
+        progress_store,
+        {},
+        {},
```
MebinAbraham, line 774: Keyword args would help here.

tests/app/questionnaire/test_path_finder.py
```diff
@@ -535,3 +536,253 @@ def test_new_remove_answer_and_block_if_routing_backwards(list_store):
...
+def test_routing_path_block_ids_dependent_on_when_rules(
+    list_store,
+    skip_age_answer,
+    skip_confirmation_answer,
+    schema,
+    section_id,
+    expected_route,
+):
+    # Given
+    schema = load_schema_from_name(schema)
```
MebinAbraham, line 654: This test is pretty hard to digest IMO as it is specific to the schema set up. Maybe some comments would help. I think when coming back to this in the future, will won't be very obvious on why this is doing the things it does.

tests/functional/spec/routing_and_skipping_section_dependencies.spec.js
```diff
@@ -0,0 +1,236 @@
...
+import SkipConfirmationPage from "../generated_pages/new_routing_and_skipping_section_dependencies/skip-confirmation.page";
+import SkipConfirmationSectionSummaryPage from "../generated_pages/new_routing_and_skipping_section_dependencies/skip-confirmation-section-summary.page";
+import SkipSectionSummaryPage from "../generated_pages/new_routing_and_skipping_section_dependencies/skip-section-summary.page";
+
+import HubPage from "../base_pages/hub.page";
+
+describe("Given the routing and skipping section dependencies questionnaire", () => {
+  beforeEach("Load the survey", () => {
+    browser.openQuestionnaire("test_new_routing_and_skipping_section_dependencies.json");
+  });
```
MebinAbraham, line 22: This could be tested for both schemas without duplicating the test case.

tests/functional/spec/routing_and_skipping_section_dependencies.spec.js
```diff
@@ -0,0 +1,236 @@
...
+    $(HouseHoldPersonalDetailsSectionSummaryPage.submit()).click();
+    $(HubPage.summaryRowLink("household-personal-details-section-2")).click();
+    $(RepeatingSexPage.male()).click();
+    $(RepeatingAgePage.submit()).click();
+
+    expect($(HouseHoldPersonalDetailsSectionSummaryPage.repeatingSexAnswer()).getText()).to.contain("Male");
+    expect($(HouseHoldPersonalDetailsSectionSummaryPage.repeatingAgeAnswer()).isExisting()).to.be.false;
+  });
+
+  function selectPrimaryPerson() {
```
MebinAbraham, line 149: Minor, but we should look to use arrow functions when specific features of full functions are not needed.

tests/functional/spec/routing_and_skipping_section_dependencies.spec.js
```diff
@@ -0,0 +1,236 @@
...
+    answerAndSubmitNameQuestion();
+    answerAndSubmitAgeQuestion();
+    answerAndSubmitReasonForNoConfirmationQuestion();
+
+    expectPersonalDetailsName();
+    expectPersonalDetailsAge();
+    expectReasonNoConfirmationAnswer();
+  });
+
+  it("When I answer 'Yes' to skipping the age question, Then in the Primary Person section I am asked my name and why I didn't confirm skipping", () => {
```
MebinAbraham, line 37: The `Then` should highlight the important parts which to me here is that you are not asked the age question. Even just adding `only`/`just` would help, i.e `Then in the Primary Person section I am only asked my name and why I didn't confirm skipping`. You use `just` in the next test case.

MebinAbraham (commented):
Functionally as expected.

app/questionnaire/path_finder.py
```diff
@@ -34,31 +36,65 @@ def routing_path(
...
         section = self.schema.get_section(section_id)
         list_name = self.schema.get_repeating_list_for_section(
             current_location.section_id
         )
 
         if section:
+            block_ids_dependent_on_when_rules = (
+                self._get_block_ids_dependent_on_when_rules(section)
+            )
+            blocks = self._get_not_skipped_blocks_in_section(
```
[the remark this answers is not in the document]
LJBabbage: Yeah just broken, I refactored to get it out the way, but it needs moving down

app/questionnaire/questionnaire_schema.py
```diff
@@ -659,3 +666,40 @@ def _get_error_messages(self) -> dict:
...
+                if block:
+                    section_id = self.get_section_id_for_block_id(block["id"])
+
+                    if section_id and section_id != current_section_id:
+                        section_ids_dependent_on_rules.add(section_id)
+
+            if any(operator in rule for operator in OPERATION_MAPPING):
+                section_ids_dependent_on_rules.update(
+                    self._get_section_ids_dependent_on_rules(current_section_id, rule)
+                )
```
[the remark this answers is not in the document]
LJBabbage: I think we pretty much agreed that we could use # type: ignore once we were transversing the schema so anything after answer to block

tests/functional/spec/routing_and_skipping_section_dependencies.spec.js
LJBabbage: I think we need to be consistent as we haven't done this for any other new rules. We also are looking to deprecate the other rules not add more code. Because we have unit test coverage for both and are looking to be pragmatic with functional tests with the other points mentioned above I think we should leave as is

.pylintrc
```diff
@@ -217,7 +217,7 @@ single-line-if-stmt=no
 no-space-check=trailing-comma,dict-separator
 
 # Maximum number of lines in a module
-max-module-lines=1150
+max-module-lines=1200
```
LJBabbage, line 220: Was left on a knife edge, one for tech

app/questionnaire/path_finder.py
```diff
@@ -34,31 +36,69 @@ def routing_path(
...
+
+            if blocks:
+                routing_path_block_ids = self._build_routing_path_block_ids(
+                    blocks, current_location, when_rules_block_ids_dependencies
+                )
+
+        return RoutingPath(routing_path_block_ids, section_id, list_item_id, list_name)
+
+    def _get_when_rules_block_ids_dependencies(self, section_id: str) -> list[str]:
+        """NB: Does not support repeating sections"""
```
MebinAbraham, line 65: I think a bit more would help. This makes me think repeating doesn't work at all, but it's to make it clear that, a repeating section can be a dependent but not depended upon.

MebinAbraham (commented):
Looking good.

app/questionnaire/path_finder.py
```diff
@@ -34,31 +36,69 @@ def routing_path(
...
+        when_rules_block_ids_dependencies: list[str] = []
+
+        if when_rules_section_ids_dependencies := self.schema.when_rules_section_ids_dependencies_map.get(
+            section_id
+        ):
+            for section_id_dependency in when_rules_section_ids_dependencies:
+                when_rules_block_ids_dependencies.extend(
+                    self.routing_path(section_id_dependency)
+                )
```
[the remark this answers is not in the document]
LJBabbage: Path generation and checking by validator has always been on the wish list of validator, as naturally the same kind of thing can happen with routing. Not got there yet.

app/questionnaire/questionnaire_schema.py
```diff
@@ -728,3 +733,50 @@ def _get_error_messages(self) -> dict:
...
+            rules = self.get_operands(rules)
+
+        for rule in rules:
+            if not isinstance(rule, dict):
+                continue
+
+            answer_id = None
+
+            if "id" in rule:
+                answer_id = rule["id"]
```
jamesnunn, line 766:
```suggestion
            answer_id = rule.get("id")
```

tests/app/questionnaire/test_path_finder.py
jamesnunn: Possibly (and I know I've not been doing it because the test params are fairly straightforward), you can add id's to `@pytest.mark.parametrize(..., ids=['skip age question results in skip confirmation question'], ...)`. I try and avoid where possible though because the list of id's should correspond in order to the test params which can get disordered easily. Solution would be to define a data structure which contains the params and the ids together then passed to the parametrize decorator but then there's a bit more stuff to grok when reviewing and maintaining.

tests/app/questionnaire/test_path_finder.py
```diff
@@ -535,3 +536,221 @@ def test_new_remove_answer_and_block_if_routing_backwards(list_store):
...
+    list_store,
+    skip_age_answer,
+    skip_confirmation_answer,
+    schema,
+    section_id,
+    expected_route,
+):
+    # Given a schema which has when rules in a section which has dependencies on other sections answers
+    schema = load_schema_from_name(schema)
+    answer_store = AnswerStore()
```
jamesnunn, line 620: you can use the `answer_store` fixture instead of defining this each time

tests/app/questionnaire/test_path_finder.py
```diff
@@ -535,3 +536,221 @@ def test_new_remove_answer_and_block_if_routing_backwards(list_store):
...
+        ]
+    ),
+)
+def test_routing_path_block_ids_dependent_on_other_sections_when_rules_repeating(
+    skip_age_answer,
+    schema,
+    expected_route,
+):
+    # Given a schema with repeating sections which has when rules dependent on another section
+    schema = load_schema_from_name(schema)
```
jamesnunn, line 702: This clobbers the arg to the test function

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
MebinAbraham: I'll raise some card, as we need to validate this, as previously, if it was done, it would just evaluate false whereas now it would end up in an infinite loop.

tests/app/questionnaire/test_path_finder.py
LJBabbage: Sorry, I don't have any idea what this comment means

app/questionnaire/questionnaire_schema.py
LJBabbage: Because of the elif, I am not convinced that would add anything, I would then need to something like  
```            
if not answer_id and rule.get("source") == "answers":
      answer_id = rule.get("identifier")
```

app/views/handlers/question.py
```diff
@@ -204,7 +204,7 @@ def evaluate_and_update_section_status_on_list_change(self, list_name):
         section_ids = self._schema.get_section_ids_dependent_on_list(list_name)
 
         section_keys_to_evaluate = (
-            self.questionnaire_store_updater.started_section_keys(
+            self._questionnaire_store.progress_store.started_section_keys(
```
MebinAbraham, line 207: We should avoid using questionnaire store progress here as that technically should be private. It would be better to keep the method on questionnaire store updater, which calls the progress store one. We already do that pattern for a couple of the progress store methods.

app/questionnaire/path_finder.py
```diff
@@ -34,31 +36,68 @@ def routing_path(
...
+        """NB: At present when rules block dependencies does not fully support repeating sections.
+        It is supported when the section is dependent i.e the current section is repeating and building the routing path for sections that are not,
+        It isn't supported if it needs to build the path for repeating sections"""
+        return [
+            block_id
+            for dependent_section in self.schema.when_rules_section_dependencies_map.get(
+                section_id, {}
+            )
+            for block_id in self.routing_path(dependent_section)
+            if dependent_section not in self.progress_store.started_section_keys()
```
MebinAbraham, line 74: Section keys are tuple, so this should be:
```suggestion
            if (dependent_section, None) in self.progress_store.started_section_keys()
```

However, the fact that no tests fails indicates a gap in our unit test, so we need to have one for when the section has not started.

app/questionnaire/questionnaire_schema.py
```diff
@@ -728,3 +733,50 @@ def _get_error_messages(self) -> dict:
...
+    ) -> set[str]:
+        rules_section_dependencies: set[str] = set()
+
+        if isinstance(rules, dict) and any(
+            operator in rules for operator in OPERATION_MAPPING
+        ):
+            rules = self.get_operands(rules)
+
+        for rule in rules:
+            if not isinstance(rule, dict):
```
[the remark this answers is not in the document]
MebinAbraham: Still outstanding.

jamesnunn (approved):

app/questionnaire/path_finder.py
MebinAbraham: There are tests for not started, but they don't test that the rule essentially evaluates false because dependent section block id are don't exist, hence the path should be different to if it was started.
> It was just all sections routing were still being evaluated

Correct, but that means that one test should have failed if we had that scenario covered. Coverage passes because it is a comprehension.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
MebinAbraham: The one on line 754 needs updating as well.

app/questionnaire/path_finder.py
MebinAbraham: As discussed, we can revisit this when we do iteration 2.

MebinAbraham (approved):
We can review performance once this goes in, unfortunately we are now at the mercy of author and how complex a schema is to how performant it will be. Would be good to progress the next phase soon, whereby we only evaluate the path for rules that need it.

PR #813, opened 2022-03-08
Upgrade to node 14

rmccar:
This PR is to update Runners version of node as v12 is about to go out of maintenance LTS next month. It also updates the documentation and .nvmrc to reflect this.

Ideally we will upgrade to v16 but there is a big breaking change with that, in that wdio/sync is no longer supported by v16. Because of this all the tests will need to be refactored to take advantage of js async/await. Because this is a much bigger job I will update runner to v14 for the time being to make sure we are on a supported version of node. Dealing with rewriting the tests and upgrading to v16 as another piece of work.

### How to review 
- Functional tests pass

### Checklist


rmccar:
[Upgrade to node v14](https://trello.com/c/SKWwJtdI/5189-upgrade-to-node-v14)

.nvmrc
```diff
@@ -1 +1 @@
-v8.16.0
+v14.19.0
```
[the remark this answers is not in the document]
rmccar: Ive explained my decision to initially go to 14 in the description at the top

README.md
```diff
@@ -151,12 +151,12 @@ Or set the `GOOGLE_CLOUD_PROJECT` environment variable to your gcp project id.
...
 
-The frontend tests use NodeJS to run. You will need to have node version 12.X to run these tests. To do this, do the following commands:
+The frontend tests use NodeJS to run. You will need to have node version 14.X to run these tests. To do this, do the following commands:
 
 ``` shell
 brew install nvm
-nvm install 12
-nvm use 12
+nvm install 14
+nvm use
```
petechd, line 159:
```suggestion
nvm use 14
```

petechd (changes requested):
Works as expected locally after installing 14, just that one typo above.

README.md
rmccar: Because I updated the `.nvmrc` file with the correct version of node, you no longer have to specify the version you want to use when running `nvm use`. It will now always get the correct version to be used for the repo. Because of this there is less chance of devs running the tests with the wrong version of node.

parimooa (approved):

README.md
rmccar: In fact, I've just realised we no longer have to specify it when we run `nvm install` so will remove it from there too

petechd (approved):

PR #820, opened 2022-03-18
Migrate tests/app/views/contexts to pytest

jamesnunn:
Migrates tests/app/views/contexts to pytest

### How to review 
Check the tests make sense

### Checklist

### DIFF
--- a/tests/app/views/contexts/conftest.py
+++ b/tests/app/views/contexts/conftest.py
@@ -231,7 +231,6 @@ def test_calculated_summary_answers():
         {"value": 10, "answer_id": "fifth-number-answer"},
         {"value": 11, "answer_id": "sixth-percent-answer"},
         {"value": 12, "answer_id": "sixth-number-answer"},
-        {"value": 12, "answer_id": "sixth-number-answer"},
     ]
     return AnswerStore(answers)
 
--- a/tests/app/views/contexts/conftest.py
+++ b/tests/app/views/contexts/conftest.py
@@ -7,6 +7,7 @@ from app.data_models.progress_store import ProgressStore
 from app.data_models.session_data import SessionData
 from app.forms.questionnaire_form import QuestionnaireForm
 from app.questionnaire.questionnaire_schema import QuestionnaireSchema
+from app.utilities.schema import load_schema_from_name
 
 
 @pytest.fixture
@@ -207,3 +208,34 @@ def fake_session_data():
         response_id="response_id",
         case_id="case_id",
     )
+
+
+@pytest.fixture
+def summary_context_schema():
+    return load_schema_from_name("test_calculated_summary")
+
+
+@pytest.fixture
+def summary_context_answers():
+    answers = [
+        {"value": 1, "answer_id": "first-number-answer"},
+        {"value": 2, "answer_id": "second-number-answer"},
+        {"value": 3, "answer_id": "second-number-answer-unit-total"},
+        {"value": 4, "answer_id": "second-number-answer-also-in-total"},
+        {"value": 5, "answer_id": "third-number-answer"},
+        {"value": 6, "answer_id": "third-and-a-half-number-answer-unit-total"},
+        {"value": "No", "answer_id": "skip-fourth-block-answer"},
+        {"value": 7, "answer_id": "fourth-number-answer"},
+        {"value": 8, "answer_id": "fourth-and-a-half-number-answer-also-in-total"},
+        {"value": 9, "answer_id": "fifth-percent-answer"},
+        {"value": 10, "answer_id": "fifth-number-answer"},
+        {"value": 11, "answer_id": "sixth-percent-answer"},
+        {"value": 12, "answer_id": "sixth-number-answer"},
+        {"value": 12, "answer_id": "sixth-number-answer"},
+    ]
+    return AnswerStore(answers)
+
+
+@pytest.fixture
+def test_section_summary_schema():
+    return load_schema_from_name("test_section_summary")
--- a/tests/app/views/contexts/conftest.py
+++ b/tests/app/views/contexts/conftest.py
@@ -216,7 +216,7 @@ def test_calculated_summary_schema():
 
 
 @pytest.fixture
-def summary_context_answers():
+def test_calculated_summary_answers():
     answers = [
         {"value": 1, "answer_id": "first-number-answer"},
         {"value": 2, "answer_id": "second-number-answer"},
--- a/tests/app/views/contexts/conftest.py
+++ b/tests/app/views/contexts/conftest.py
@@ -211,7 +211,7 @@ def fake_session_data():
 
 
 @pytest.fixture
-def summary_context_schema():
+def test_calculated_summary_schema():
     return load_schema_from_name("test_calculated_summary")
 
 
--- a/tests/app/views/contexts/test_calculated_summary_context.py
+++ b/tests/app/views/contexts/test_calculated_summary_context.py
@@ -15,7 +15,7 @@ from tests.app.views.contexts import assert_summary_context
 # pylint: disable=too-many-locals
 @pytest.mark.usefixtures("app")
 @pytest.mark.parametrize(
-    "block_id, locale, language, title, value, block_len, answers",
+    "block_id, locale, language, title, value, total_blocks, answers",
     (
         (
             "currency-total-playback-with-fourth",
@@ -64,15 +64,15 @@ from tests.app.views.contexts import assert_summary_context
         ),
     ),
 )
-def test_build_view_context_for_currency_calculated_summary_no_skip(
+def test_build_view_context_for_currency_calculated_summary(
     block_id,
     locale,
     language,
     title,
     value,
-    block_len,
+    total_blocks,
     answers,
-    summary_context_schema,
+    test_calculated_summary_schema,
     summary_context_answers,
     list_store,
     progress_store,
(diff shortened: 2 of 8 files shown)


tests/app/views/contexts/conftest.py
```diff
@@ -207,3 +208,34 @@ def fake_session_data():
         response_id="response_id",
         case_id="case_id",
     )
+
+
+@pytest.fixture
+def summary_context_schema():
```
MebinAbraham, line 214: I think it would be clearer if it was named after the schema.

tests/app/views/contexts/test_calculated_summary_context.py
```diff
@@ -9,237 +7,115 @@
...
-            context_summary["calculated_question"]["answers"][0]["value"], "22"
-        )
+    context = calculated_summary_context.build_view_context_for_calculated_summary(
+        current_location
+    )
+
+    for answer in answers:
+        summary_context_answers.add_or_update(answer)
+
+    assert "summary" in context, "Key value summary missing from context"
```
MebinAbraham, line 105: Confused as to why, some asserts have the exception text defined and some doesn't?. Don't think they really add much in these simple scenarios.

MebinAbraham, line 105: Confused as to why, some asserts have the exception text defined and some doesn't?. Don't think they really add much in these simple scenarios.

tests/app/views/contexts/test_calculated_summary_context.py
```diff
@@ -9,237 +7,115 @@
...
-
-        calculated_summary_context = CalculatedSummaryContext(
+from app.views.contexts.section_summary_context import SectionSummaryContext
+from tests.app.views.contexts import assert_summary_context
+
+
+# pylint: disable=too-many-locals
+@pytest.mark.usefixtures("app")
+@pytest.mark.parametrize(
+    "block_id, locale, language, title, value, block_len, answers",
```
MebinAbraham, line 18: `block_len` sounds like the length of the block than the number of blocks you expect. Might be clearer as `total_blocks` or `number_of_blocks`

MebinAbraham, line 18: `block_len` sounds like the length of the block than the number of blocks you expect. Might be clearer as `total_blocks` or `number_of_blocks`

tests/app/views/contexts/test_calculated_summary_context.py
```diff
@@ -9,237 +7,115 @@
...
+    current_location = Location(section_id="default-section", block_id=block_id)
+
+    calculated_summary_context = CalculatedSummaryContext(
+        language,
+        summary_context_schema,
+        summary_context_answers,
+        list_store,
+        progress_store,
+        {},
+        {},
```
MebinAbraham, line 95: Keyword argument would help here.

MebinAbraham, line 95: Keyword argument would help here.

tests/app/views/contexts/test_calculated_summary_context.py
```diff
@@ -9,237 +7,115 @@
...
-            context_summary["title"],
             "We calculate the total of number values entered to be 22. Is this correct?",
-        )
+            "22",
+            2,
+            [],
+        ),
+    ),
+)
+def test_build_view_context_for_currency_calculated_summary_no_skip(
```
MebinAbraham, line 67: Is the name right?. I don't see ones for `with_skip`?. Also, the answer to the skip block is `Yes`.

MebinAbraham, line 67: Is the name right?. I don't see ones for `with_skip`?.

tests/app/views/contexts/test_section_summary_context.py
```diff
@@ -2,149 +2,130 @@
...
+        answer_store,
+        list_store,
+        progress_store,
+        {},
+        {},
+        current_location=location,
+        routing_path=mocker.MagicMock(),
+    )
+    context = summary_context()
+    assert get_dict_address(context, dict_path) == expected_title
```
MebinAbraham, line 128: I think `get_dict_address` should be removed. It is just adding unnecessary ambiguity.
You are always asserting the same title key for the summary content, so this could be simplified to: `assert context['summary'][title_key] == expected_title`, where `title_key` or `key_to_assert` is the value such as `"page_title"` etc.

MebinAbraham, line 128: I think `get_dict_address` should be removed. It is just adding unnecessary ambiguity.
You are always asserting the same title key for the summary content, so this could be simplified to: `assert context['summary'][title_key] == expected_title`, where `title_key` or `key_to_assert` is the value such as `"page_title"` etc.

tests/app/views/contexts/test_section_summary_context.py
```diff
@@ -2,149 +2,130 @@
...
+        answer_store.add_or_update(answer)
+
+    summary_context = SectionSummaryContext(
+        "en",
+        test_section_summary_schema,
+        answer_store,
+        list_store,
+        progress_store,
+        {},
+        {},
```
MebinAbraham, line 123: Same here, keyword args would help readability. Couple more occurrences of this, I haven't marked them all.

MebinAbraham, line 123: Same here, keyword args would help readability.

tests/app/views/contexts/conftest.py
```diff
@@ -207,3 +208,34 @@ def fake_session_data():
...
+    return load_schema_from_name("test_calculated_summary")
+
+
+@pytest.fixture
+def summary_context_answers():
```
MebinAbraham, line 219: My preference would be these to be named after the schema itself. Not a blocker.

tests/app/views/contexts/test_calculated_summary_context.py
jamesnunn: I just migrated them over, I'll remove

tests/app/views/contexts/test_submit_context.py
```diff
@@ -1,105 +1,68 @@
...
+@pytest.mark.usefixtures("app")
+def test_context(answer_store, list_store, progress_store):
+    schema = load_schema_from_name("test_submit_with_summary")
+    submit_questionnaire_context = SubmitQuestionnaireContext(
+        "en",
+        schema,
+        answer_store,
+        list_store,
+        progress_store,
+        {},
```
parimooa, line 63: kwargs as suggested by Mebin

tests/app/views/contexts/test_submit_context.py
```diff
@@ -1,105 +1,68 @@
...
+    schema_name, expected, answer_store, list_store, progress_store
+):
+    schema = load_schema_from_name(schema_name)
+    submit_questionnaire_context = SubmitQuestionnaireContext(
+        "en",
+        schema,
+        answer_store,
+        list_store,
+        progress_store,
+        {},
```
parimooa, line 41: kwargs as suggested by Mebin

tests/app/views/contexts/test_submit_context.py
```diff
@@ -1,105 +1,68 @@
...
+
+    assert context["title"] == expected["title"]
+    assert context["guidance"] == expected["guidance"]
+    assert context["warning"] == expected["warning"]
+    assert context["submit_button"] == expected["submit_button"]
+    assert "summary" not in context
+
+
+@pytest.mark.usefixtures("app")
+def test_context(answer_store, list_store, progress_store):
```
parimooa, line 55: may be rename to `test_questionnaire_context`

tests/app/views/contexts/test_section_summary_context.py
```diff
@@ -2,149 +2,130 @@
...
-            self.response_metadata,
-            routing_path=MagicMock(),
-            current_location=Location(section_id="property-details-section"),
-        )
-        context = summary_context()
-        self.assertEqual(context["summary"]["title"], "Property Details Section")
+from tests.app.views.contexts import assert_summary_context
+
+
+def get_dict_address(d, args):
```
parimooa, line 16: don't see this getting used anywhere?

tests/app/views/contexts/test_calculated_summary_context.py
```diff
@@ -9,237 +7,115 @@
...
+        metadata={},
+        response_metadata={},
+    )
+
+    context = calculated_summary_context.build_view_context_for_calculated_summary(
+        current_location
+    )
+
+    for answer in answers:
+        test_calculated_summary_answers.add_or_update(answer)
```
MebinAbraham, line 103: This doesn't make sense here, this would have to come before initialisation `CalculatedSummaryContext`. We previously had a test when both skipping/not skipping, which we don't cover at the moment. This line at the moment, make no difference to the test.

jamesnunn: So you mean it needs to move before `CalculatedSummaryContext` or isn't needed at all now?

MebinAbraham: It will need to be before `CalculatedSummaryContext` if it needs to be used. Previous tests covered summary context is as expected when skipping vs not skipping, i.e 1 extra block, which it doesn't look like we cover at the moment.

jamesnunn: Yeah, the test passes regardless of where the it appears so I guess that single test that adds answers doesn't make a difference?

MebinAbraham: In this instance it doesn't make a difference, I guess I am saying it should?. We don't have explict tests that show skipping vs non skipping which we originally had, this is what is probably missing, as for that, you want answers in the store.

MebinAbraham (approved):

parimooa (approved):

PR #822, opened 2022-03-22
Add type hints to helpers

MebinAbraham:
Added type hints to the helpers module.

### How to review 
Ensure the type annotations are accurate.

Note: Decorator type hints return `Any` due to a lack of support by the language. It has been improved in Py 3.10, therefore, we can properly type check decorators once we are on Py 3.10 using [ParamSpec]( https://peps.python.org/pep-0612/).

### Checklist

### DIFF
--- a/app/helpers/header_helpers.py
+++ b/app/helpers/header_helpers.py
@@ -1,6 +1,13 @@
-def get_span_and_trace(headers):
+from typing import Union
+
+from werkzeug.datastructures import EnvironHeaders
+
+
+def get_span_and_trace(
+    headers: EnvironHeaders,
+) -> Union[tuple[None, None], tuple[str, str]]:
     try:
-        trace, span = headers.get("X-Cloud-Trace-Context").split("/")
+        trace, span = headers.get("X-Cloud-Trace-Context").split("/")  # type: ignore
     except (ValueError, AttributeError):
         return None, None
 
--- a/app/helpers/session_helpers.py
+++ b/app/helpers/session_helpers.py
@@ -1,18 +1,19 @@
 from functools import wraps
+from typing import Any, Callable, Mapping
 
 from flask_login import current_user
 
 from app.globals import get_questionnaire_store, get_session_store
 
 
-def with_questionnaire_store(function):
+def with_questionnaire_store(function: Callable) -> Any:
     """Adds the `questionnaire_store` as an argument, where the `current_user` is defined.
     Use on flask request handlers or methods called by flask request handlers.
 
     May error unless there is a `current_user`."""
 
     @wraps(function)
-    def wrapped_function(*args, **kwargs):
+    def wrapped_function(*args: tuple, **kwargs: Mapping) -> Any:
         questionnaire_store = get_questionnaire_store(
             current_user.user_id, current_user.user_ik
         )
@@ -21,13 +22,13 @@ def with_questionnaire_store(function):
     return wrapped_function
 
 
-def with_session_store(function):
+def with_session_store(function: Callable) -> Any:
     """Adds the `session_store` as an argument.
     Use on flask request handlers or methods called by flask request handlers.
     """
 
     @wraps(function)
-    def wrapped_function(*args, **kwargs):
+    def wrapped_function(*args: Any, **kwargs: Any) -> Any:
         session_store = get_session_store()
         return function(session_store, *args, **kwargs)
 
--- a/app/helpers/session_helpers.py
+++ b/app/helpers/session_helpers.py
@@ -1,5 +1,5 @@
 from functools import wraps
-from typing import Any, Callable, Mapping
+from typing import Any, Callable
 
 from flask_login import current_user
 
--- a/app/helpers/session_helpers.py
+++ b/app/helpers/session_helpers.py
@@ -13,7 +13,7 @@ def with_questionnaire_store(function: Callable) -> Any:
     May error unless there is a `current_user`."""
 
     @wraps(function)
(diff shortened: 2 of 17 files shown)


app/helpers/header_helpers.py
```diff
@@ -1,6 +1,13 @@
...
+
+from werkzeug.datastructures import EnvironHeaders
+
+
+def get_span_and_trace(
+    headers: EnvironHeaders,
+) -> Union[tuple[None, None], tuple[str, str]]:
     try:
-        trace, span = headers.get("X-Cloud-Trace-Context").split("/")
+        trace, span = headers.get("X-Cloud-Trace-Context").split("/")  # type: ignore
```
MebinAbraham, line 10: Ignore only needed because this is using `EAFP`, ignore can be removed if we did `LBYL`.  Reference: https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

app/helpers/session_helpers.py
```diff
@@ -21,13 +22,13 @@ def wrapped_function(*args, **kwargs):
...
 
-def with_session_store(function):
+def with_session_store(function: Callable) -> Any:
     """Adds the `session_store` as an argument.
     Use on flask request handlers or methods called by flask request handlers.
     """
 
     @wraps(function)
-    def wrapped_function(*args, **kwargs):
+    def wrapped_function(*args: Any, **kwargs: Any) -> Any:
```
parimooa, line 31: its looks identical as above why is it `Any`

tests/app/helpers/conftest.py
```diff
@@ -192,3 +198,28 @@ def expected_footer_nisra_theme():
             "alt": "NISRA - Northern Ireland Statistics and Research Agency",
         },
     }
+
+
+@pytest.fixture()
+def session_data():
```
parimooa, line 204: There is already one `session_data` at [test level](https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/tests/app/conftest.py#L78-L91) , may be re-use that

tests/app/helpers/conftest.py
```diff
@@ -192,3 +198,28 @@ def expected_footer_nisra_theme():
...
+        ru_name=None,
+        ru_ref=None,
+        submitted_at=datetime.now(timezone.utc).isoformat(),
+        response_id="321",
+        case_id="789",
+    )
+
+
+@pytest.fixture()
+def session_store(session_data):
```
parimooa, line 222: This is also existing [here](https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/tests/app/conftest.py#L73)

app/helpers/session_helpers.py
petechd: Since ** unpacks dictionaries and kwargs are keyword arguments shouldn't it have type `dict` or `Mapping` instead of `Any`?

tests/app/helpers/conftest.py
MebinAbraham: Done.

tests/app/helpers/conftest.py
MebinAbraham: Done.

app/helpers/uuid_helper.py
```diff
@@ -1,7 +1,7 @@
 from uuid import UUID
 
 
-def is_valid_uuid(uuid_string, version=4):
+def is_valid_uuid(uuid_string: str, version: int = 4) -> bool:
```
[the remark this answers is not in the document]
MebinAbraham: Given this is for typing, I don't think that should be done here.

app/helpers/session_helpers.py
MebinAbraham: The above existing one was wrong, I updated it to Any. For args and kwargs, they are tuple are dicts, but that is not how it is supposed to be typed. It should be typed with the actual types they can actually hold. https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

It is at the moment, any as this is a generic decorator that can be called by any function, and we can't infer the type of the callers args in 3.9. As per the PR description:
> Note: Decorator type hints return Any due to a lack of support by the language. It has been improved in Py 3.10, therefore, we can properly type check decorators once we are on Py 3.10 using [ParamSpec](https://peps.python.org/pep-0612/).

app/helpers/uuid_helper.py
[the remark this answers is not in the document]
MebinAbraham: I think I have misunderstood you. Did you need literally a constant, i.e like `DEFAULT_VERSION` etc?.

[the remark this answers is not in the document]
MebinAbraham: [`bc7b703` (#822)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/822/commits/bc7b70375fec18cc7291543383d598e1121a5d64)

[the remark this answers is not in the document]
parimooa: What i meant was , no need to pass args. I don't think there will be an instance where you will (call )validate uuid string with different versions.
```
def is_valid_uuid(uuid_string: str) -> bool:
    try:
        UUID(uuid_string, version=DEFAULT_UUID_VERSION)
    except ValueError:
```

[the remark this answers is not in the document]
MebinAbraham: Not really type hint specific, but changed while I'm here. Decided to remove the constant since it is only used one, and renamed the method to be specific.

MebinAbraham:
To retrigger actions.

PR #823, opened 2022-03-22
Answer dependencies for min and max validation

LJBabbage:
This PR adds answer dependencies for min and max validation

### How to review 
Using test_numbers.json, complete the survey then go back and change set min/max (first block). Make sure you have to confirm/validate any answers again that have dependencies before going back to summary, the best one to change is max as it has dependencies on the other 2 blocks in the questionnaire

### Checklist

### DIFF
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -249,9 +249,9 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
                             )
 
     def _update_answer_dependencies_for_calculated_summary(
-        self, answer_ids: list[str], block_id: str
+        self, calculated_summary_answer_ids: list[str], block_id: str
     ) -> None:
-        for answer_id in answer_ids:
+        for answer_id in calculated_summary_answer_ids:
             self._answer_dependencies_map[answer_id] |= {
                 self._get_answer_dependent_for_block_id(block_id)
             }
@@ -275,10 +275,9 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
         self, answer: Mapping, block_id: str
     ) -> None:
         for key in ["minimum", "maximum"]:
-            if answer.get(key) and isinstance(answer[key]["value"], dict):
-                self._update_answer_dependencies_for_value_source(
-                    answer[key]["value"], block_id
-                )
+            value = answer.get(key, {}).get("value")
+            if isinstance(value, dict):
+                self._update_answer_dependencies_for_value_source(value, block_id)
 
     def _update_answer_dependencies_for_value_source(
         self, value_source: dict, block_id: str
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -280,7 +280,7 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
                 self._update_answer_dependencies_for_value_source(value, block_id)
 
     def _update_answer_dependencies_for_value_source(
-        self, value_source: dict, block_id: str
+        self, value_source: Mapping, block_id: str
     ) -> None:
         if value_source["source"] == "answers":
             self._answer_dependencies_map[value_source["identifier"]] |= {
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -229,7 +229,7 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
     def _populate_answer_dependencies(self) -> None:
         for block in self.get_blocks():
             if block["type"] == "CalculatedSummary":
-                self._update_answer_dependencies_for_answers(
+                self._update_answer_dependencies_for_calculated_summary(
                     block["calculation"]["answers_to_calculate"], block["id"]
                 )
                 continue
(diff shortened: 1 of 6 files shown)


app/questionnaire/questionnaire_schema.py
```diff
@@ -229,7 +229,7 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
     def _populate_answer_dependencies(self) -> None:
         for block in self.get_blocks():
             if block["type"] == "CalculatedSummary":
-                self._update_answer_dependencies_for_answers(
+                self._update_answer_dependencies_for_calculated_summary(
```
LJBabbage, line 232: Updated the name as originally I thought it could be used for any answers dependencies, then realised it was a much more specialist situation and at present only viable for calculated_summary. Leaving the name as is could have caused confusion and be reuse by accident

schemas/test/en/test_numbers.json
```diff
@@ -290,6 +290,67 @@
                                 "title": "Please enter test values (none mandatory)",
                                 "type": "General"
                             }
+                        },
+                        {
+                            "type": "Question",
```
LJBabbage, line 295: Although this doesn't follow the pattern of the rest of the schema, I wasn't happy that I could exercise detail_answer correctly. The previous block already had a dependency on set-maximum, it was feasible that was the reason for block being removed and therefore details_answer wasn't doing it. Keeping it completely separate proved the code was working correctly for detail-answer

app/questionnaire/questionnaire_schema.py
```diff
@@ -264,14 +271,33 @@ def _update_answer_dependencies_for_calculations(
             }
             self._answer_dependencies_map[source_answer_id] |= dependents
 
-    def _update_answer_dependencies_for_answers(
-        self, answer_ids: list[str], block_id: str
+    def _update_answer_dependencies_for_min_max(
```
LJBabbage, line 274: Left as min_max, but can be more general when dynamic_options is implemented

tests/functional/spec/numbers.spec.js
```diff
@@ -1,62 +1,84 @@
...
+      $(TestMinMax.testPercent()).setValue("99");
+      $(TestMinMax.submit()).click();
+      $(DetailAnswer.other()).click();
+      $(DetailAnswer.otherDetail()).setValue("1020");
+      $(DetailAnswer.submit()).click();
+
+      expect(browser.getUrl()).to.contain(SubmitPage.pageName);
+    });
+
+    it("When I edit and change a question which has dependent answer, Then I must re-answer if required and re-submit before I can return to the summary", () => {
```
MebinAbraham, line 69: We should test both min and max, they are two separate constructs, so to verify both works, we should test both.

MebinAbraham, line 69: We should test both min and max, they are two separate constructs, so to verify both works, we should test both.

tests/functional/spec/numbers.spec.js
```diff
@@ -1,62 +1,84 @@
...
+      $(SetMinMax.submit()).click();
+      $(TestMinMax.submit()).click();
+      $(DetailAnswer.submit()).click();
+
+      expect($(DetailAnswer.errorNumber(1)).getText()).to.contain("Enter an answer less than or equal to 1,019");
+
+      $(DetailAnswer.otherDetail()).setValue("1019");
+      $(DetailAnswer.submit()).click();
 
-  it("Given values outside of the allowed decimal places then the correct error messages are displayed", () => {
```
MebinAbraham, line 52: Has this been removed?

app/questionnaire/questionnaire_schema.py
MebinAbraham: Not a blocker. I get your point, but I also find it confusing that it is named after calculated summary but code block doesn't seem to have anything relating to calculated summaries.

MebinAbraham (commented):
Looks good.

app/questionnaire/questionnaire_schema.py
```diff
@@ -264,14 +271,33 @@ def _update_answer_dependencies_for_calculations(
...
+        self, answer: Mapping, block_id: str
     ) -> None:
-        for answer_id in answer_ids:
-            self._answer_dependencies_map[answer_id] |= {
-                self._get_answer_dependent_for_block_id(block_id)
+        for key in ["minimum", "maximum"]:
+            if answer.get(key) and isinstance(answer[key]["value"], dict):
+                self._update_answer_dependencies_for_value_source(
+                    answer[key]["value"], block_id
+                )
```
jamesnunn, line 281: if value:= isinstance(answer.get(key, {}).get('value'), dict):
                    value, block_id
                )

app/questionnaire/questionnaire_schema.py
LJBabbage: I did have it inline for a while, maybe the best solution?

tests/functional/spec/numbers.spec.js
LJBabbage: At unit level makes sense, but as I am using the schema in that unit test the functional test will also get it

app/questionnaire/questionnaire_schema.py
```diff
@@ -264,14 +271,33 @@ def _update_answer_dependencies_for_calculations(
...
+
+    def _update_answer_dependencies_for_value_source(
+        self, value_source: dict, block_id: str
```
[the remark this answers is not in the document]
LJBabbage: Everywhere else in our code its dict

app/questionnaire/questionnaire_schema.py
MebinAbraham: I think what you have changed is fine for now. We can revisit later if needed.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
MebinAbraham: Ya, prob just mistyped. I don't think we do anything dict specific on value sources.

MebinAbraham (approved):

jamesnunn (approved):

PR #824, opened 2022-03-25
Additional logging and tests when loading schema from URL

MebinAbraham:
Simple PR to add some additional logging and cache usage info when loading schema via a URL.
**Note:** The additional logging is temporary and will be removed before loading from URL is used in production.

### How to review 
Ensure the changes satisfies the requirements on the card.

### Checklist

### DIFF
--- a/app/utilities/schema.py
+++ b/app/utilities/schema.py
@@ -1,4 +1,5 @@
 import os
+import time
 from functools import lru_cache
 from glob import glob
 from pathlib import Path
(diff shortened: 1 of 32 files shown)


app/utilities/schema.py
```diff
@@ -80,10 +81,26 @@ def get_allowed_languages(schema_name, launch_language):
...
+
+        start = time.time()
+        schema = load_schema_from_url(survey_url, metadata.get("language_code"))
+        duration = time.time() - start
+
+        cache_info = (
+            load_schema_from_url.cache_info()  # pylint: disable=no-value-for-parameter
+        )
+        logger.info(
+            f"load_schema_from_url took {duration} seconds",
```
jamesnunn, line 97: You might want to format the output here to milliseconds or something, it's a bit hard to grok scientific notation:

`load_schema_from_url took 9.5367431640625e-07 seconds`

jamesnunn (approved):

tests/app/utilities/test_schema.py
```diff
@@ -152,6 +156,32 @@ def test_load_schema_from_url_404():
     with pytest.raises(NotFound):
         load_schema_from_url(survey_url=TEST_SCHEMA_URL, language_code="en")
 
+    cache_info = load_schema_from_url.cache_info()
+    assert cache_info.currsize == 0
+    assert cache_info.hits == 0
```
pricem14pc, line 161: For completeness and visibility of the pattern is it worth asserting the misses too through each test?

pricem14pc:
We spoke about the cache being scoped to the gunicorn workers. I see the effects of that when running locally with 3 workers where the cache info `hits` shows one of the 3 workers caches. This has the effect of the cache info fluctuating (between the 3 caches). While this is understood, is there a way of including a worker ID of the cache being logged so we can correlate the cache info per worker when we begin testing?

tests/app/utilities/test_schema.py
MebinAbraham: Done

pricem14pc (approved):
retested with multiple gunicorn workers and a single python process. The process `pid` working as expected in both scenarios.

PR #826, opened 2022-03-28
Fix save and continue button spacing

rmccar:
Fixes issue with save and continue button that has too big a margin top on it.

Also removes a `ons-u-mb-xl` that is not needed too.

This PR also updates the wdio config to make the resolution 4k. This is because updating this bit by bit has caused issues in the tests. Updating in to a 4k resolution should stop anymore issues

### How to review 
- Check a few schemas in master and on this branch to see the button is now spaced correctly (`test_numbers`, `test_dates` are two good ones)
- Check `test_confirmation_email` see the gap underneath the button is reduced

### Checklist

### DIFF
--- a/tests/functional/spec/features/validation/date_validation/date-combined-yyyy.spec.js
+++ b/tests/functional/spec/features/validation/date_validation/date-combined-yyyy.spec.js
@@ -8,7 +8,7 @@ describe("Feature: Combined question level and single validation for MM-YYYY dat
 
   describe("Period Validation", () => {
     describe("Given I enter dates", () => {
-      it("When I enter a single dates that are too early/late, Then I should see single validation errors", () => {
+      it("When I enter dates that are too early and too late, Then I should see two validation errors", () => {
         $(DateRangePage.dateRangeFromYear()).setValue(2015);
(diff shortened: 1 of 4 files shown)


rmccar:
[Save and continue button spacing bug](https://trello.com/c/sYMj8Wdd/5236-save-and-continue-button-spacing-bug)

tests/functional/spec/features/validation/date_validation/date-combined-yyyy.spec.js
```diff
@@ -8,7 +8,7 @@ describe("Feature: Combined question level and single validation for MM-YYYY dat
 
   describe("Period Validation", () => {
     describe("Given I enter dates", () => {
-      it("When I enter a single dates that are too early/late, Then I should see a single validation errors", () => {
+      it("When I enter a single dates that are too early/late, Then I should see single validation errors", () => {
```
berroar, line 11: Think this description needs to be updated to something like:
```suggestion
      it("When I enter dates that are too early and too late, Then I should see two validation errors", () => {
```

berroar (approved):

jamesnunn (approved):
Button looks as described

PR #828, opened 2022-03-29
Update 'Save and sign out' button

petechd:

This changes the 'Save and sign out' button to 'Save and exit survey' and adds query string parameter `?todo=true`. Some handler logic was also replaced (check the card).

"There are currently two duplicate actions in the top right hand corner to sign out. During research 'Save and exit survey' was tested (to replace 'Save and sign out' and everyone understood this would take you back to RAS, whereas the 'Sign out' would take you to our sign out page. "

### How to review 
Extra tests added and some existing tests has been tweaked.

### Checklist


petechd:
[Update to 'Save and sign out' button (2 SP)](https://trello.com/c/8dS39BCp/5222-update-to-save-and-sign-out-button-2-sp)

app/routes/session.py
```diff
@@ -132,7 +132,16 @@ def get_sign_out():
     """
     Signs the user out of eQ and redirects to the log out url.
     """
-    log_out_url = get_survey_config().account_service_log_out_url
+    todo = request.args.get("todo", "").upper() == "TRUE"
```
jamesnunn, line 135: Maybe the default should be "FALSE" to keep the bool intention explicit rather than imply a tristate is possible (unless it is possible...)
```suggestion
    todo = request.args.get("todo", "FALSE").upper() == "TRUE"
```

jamesnunn: Also is it intentional that the string passed to the query param is case insensitive? It probably doesn't matter, just checking.

tests/app/test_session.py
```diff
@@ -0,0 +1,27 @@
...
+@pytest.mark.parametrize(
+    "todo, expected",
+    [
+        (
+            True,
+            "/sign-out?todo=True",
+        ),
+        (
+            False,
+            "/sign-out?todo=False",
```
jamesnunn, line 15: If the param arg is supposed to be case insensitive then maybe worth adding lower/upper/mixed cases

app/routes/session.py
petechd: We kept it this way because of possible discrepancies between python and jinja templates, I think, so it always evaluates to correct bool.

tests/app/test_session.py
petechd: I've changed it to case sensitive now as it always seems to be correct case in the template, even if we hardcode it to `todo=true` or `todo=false`.

petechd: I've changed it now so it's case sensitive as the boolean returned from the template is always correct case, so no need for extra tests.

parimooa (commented):
LGTM , only thing i notice is that `todo=true` in the card with all letters as lower case

petechd:
> LGTM , only thing i notice is that `todo=true` in the card with all letters as lower case

Added corrected lower case query string parameter, thanks Ash👍.

app/routes/session.py
```diff
@@ -132,7 +132,16 @@ def get_sign_out():
     """
     Signs the user out of eQ and redirects to the log out url.
     """
-    log_out_url = get_survey_config().account_service_log_out_url
+    todo = request.args.get("todo", "") == "True"
```
parimooa, line 135: This needs updating also after you updated template
` todo = request.args.get("todo", "") == "true"`

tests/app/test_session.py
```diff
@@ -0,0 +1,27 @@
...
+from flask import url_for
+
+
+@pytest.mark.usefixtures("app")
+@pytest.mark.parametrize(
+    "todo, expected",
+    [
+        (
+            True,
+            "/sign-out?todo=True",
```
parimooa, line 11: another scenario with all `todo=true`

app/routes/session.py
petechd: This might need to stay as `"True"` since that's how it is formatted when returned from the template.

tests/app/test_session.py
petechd: I would keep it as it is as it always returns as `True` or `False` from the template.

app/routes/session.py
MebinAbraham: I think we should either check where case don't matter (what most type enforced APIs does), or just check the existence of the key, which is what we do for resume, given we don't have an alternative path, it just if it has to do or not.

app/routes/session.py
```diff
@@ -132,7 +132,16 @@ def get_sign_out():
...
+    survey_config = get_survey_config()
+    if todo:
+        log_out_url = (
+            survey_config.account_service_todo_url
+            or survey_config.account_service_log_out_url
+        )
+    else:
+        # Future implementation when "todo" is not hardcoded on the template level
```
MebinAbraham, line 143: I don't understand this.

tests/app/test_session.py
```diff
@@ -0,0 +1,27 @@
...
+        ),
+    ],
+)
+def test_url(todo, expected):
+    url = url_for("session.get_sign_out", todo=todo)
+    assert url == expected
+
+
+@pytest.mark.usefixtures("app")
+def test_url_no_args():
```
MebinAbraham, line 25: I'm not convinced, we need this. This is testing flask's built in `url_for` method. This would work regardless of if the route handled query params. Since this is a route, the scenario should be tested under integrations/functional tests.

tests/integration/session/test_sign_out_and_exit.py
```diff
@@ -14,9 +14,18 @@
 
 
 class TestSaveAndSignOut(IntegrationTestCase):
```
MebinAbraham, line 16: We should update `TestCensusSignOut` to ensure the sign-out for that doesn't use the `todo` path.

MebinAbraham (commented):
Functionally as expected. Few things to consider.

tests/app/test_session.py
petechd: Yeah, I was not convinced but wanted to try to test, I was expecting you will want to scrap it as we don't do any similar test.  I'll remove it now.

app/routes/session.py
MebinAbraham: Ah I see what you mean. Ideally you would not always pass `true`, i.e for non-business, but I think for now, that is fine.

The logic just need's updating to something like:
```py
    log_out_url = survey_config.account_service_todo_url if todo else None

    if not log_out_url:
        log_out_url = survey_config.account_service_log_out_url
```

tests/integration/session/test_sign_out_and_exit.py
petechd: Thanks, added new test in `TestCensusSignOut` that checks for that.

app/routes/session.py
petechd: Re-introduced the case fix and default fallback `false`, thanks.

app/routes/session.py
petechd: Ok, added option 1 as I find it more readable, thanks.

app/routes/session.py
LJBabbage: Taking onboard what Mebin has said and doing a slight refactor, would this be better? 

```
    survey_config = get_survey_config()
    log_out_url = (
        survey_config.account_service_todo_url
        if "todo" in request.args
        else survey_config.account_service_log_out_url
    )

    # Check for GET as we don't want to log out for HEAD requests
    if request.method == "GET":
        logout_user()

    return redirect(log_out_url)

LJBabbage (commented):
Looks fine, bar a suggested refactor

tests/app/test_session.py
MebinAbraham: Resolving since it is not redundant.

MebinAbraham (approved):

LJBabbage (approved):

PR #832, opened 2022-04-04
Support for piping calculated values used by a calculated summary

MebinAbraham:
Support for https://github.com/ONSdigital/eq-questionnaire-validator/pull/142

**Note:** Due to runner's current implementation of calculated summary, value source depends on the author to ensure the calculated summary and its answers are on the path. A future iteration will look to refactor the calculated summary to take into account answers on the path, as it bring about other added complexity such as cross-section dependencies. Calculated support for repeats/loops will also be required in the future, which is currently not implemented.

### How to review 
Ensure the changes are appropriate. Test using, `test_calculated_summary.json`, the last page shows the piped values.

### Checklist


app/questionnaire/value_source_resolver.py
```diff
@@ -114,8 +117,37 @@ def _resolve_list_value_source(self, value_source: dict) -> Union[int, str, list
 
         return list(list_model)
 
+    def _resolve_calculated_summary_value_source(
+        self, value_source: Mapping
+    ) -> IntOrDecimal:
+        """Calculates the value for the 'calculation' used by the provided Calculated Summary.
```
petechd, line 123: Docstrings 👍

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -458,6 +458,45 @@ def test_response_metadata_source():
     )
 
 
+@pytest.mark.parametrize(
+    "list_item_id",
+    [None, "item-1"],
+)
+def test_calculated_summary_source(mocker, list_item_id):
```
LJBabbage, line 465: Source makes me think of the source of the calculation (i.e answers_to_calculate) rather than value_source, so  i would prefer test_calculated_summary_as_value_source/ test_calculated_summary_value_source

tests/functional/spec/features/calculated_summary.spec.js
```diff
@@ -104,25 +105,22 @@ describe("Feature: Calculated Summary", () => {
       expect($(CurrencyTotalPlaybackPageWithFourth.calculatedSummaryAnswer()).getText()).to.contain("£40.71");
     });
 
-    it("Given I leave an answer empty, When i get to the currency summary, Then I should see no answer provided and new total", () => {
-      $(CurrencyTotalPlaybackPageWithFourth.fourthNumberAnswerEdit()).click();
-      $(FourthNumberBlockPage.fourthNumber()).setValue("");
-      $(FourthNumberBlockPage.submit()).click();
+    it("Given I leave an answer empty, When I get to the currency summary, Then I should see no answer provided and new total", () => {
```
LJBabbage, line 108: Not sure this one is clear. Looks like you are editing a question and removing its value, and because of that the total changes. The description reads as if this was the first pass through the questionnaire

tests/functional/spec/features/calculated_summary.spec.js
```diff
@@ -104,25 +105,22 @@ describe("Feature: Calculated Summary", () => {
...
+        "We calculate the total of currency values entered to be £28.37. Is this correct?"
       );
-      expect($(CurrencyTotalPlaybackPageWithFourth.calculatedSummaryAnswer()).getText()).to.contain("£9.36");
-      expect($(CurrencyTotalPlaybackPageWithFourth.fourthNumberAnswer()).getText()).to.contain("No answer provided");
+      expect($(CurrencyTotalPlaybackPageWithFourth.calculatedSummaryAnswer()).getText()).to.contain("£28.37");
+      expect($(CurrencyTotalPlaybackPageWithFourth.fourthAndAHalfNumberAnswerAlsoInTotal()).getText()).to.contain("No answer provided");
     });
 
-    it("Given I skip the fourth page, When i get to the playback, Then I can should not see it in the total", () => {
+    it("Given I skip the fourth page, When I get to the playback, Then I can should not see it in the total", () => {
```
LJBabbage, line 123: Similar here it's not clear that this has been edited from a different value. Also why has it gone back to the third block just to step through it to get to the one it wants to change?

tests/functional/spec/features/calculated_summary.spec.js
```diff
@@ -192,8 +190,23 @@ describe("Feature: Calculated Summary", () => {
       expect($(NumberTotalPlaybackPage.sixthNumberAnswer()).getText()).to.contain("45.67");
     });
 
-    it("Given I confirm the total and am on the summary, When I edit and change an answer, Then I must re-confirm the calculated summary page which is dependent on the change before I can return to the summary", () => {
+    it("Given I complete every calculated summary, When I go to the next page, Then I should the see the piped calculated summary total for each summary", () => {
```
LJBabbage, line 193: Maybe not use next page as that could change in time, it doesn't necessarily have to be straight after the calculated summary

tests/functional/spec/features/calculated_summary.spec.js
MebinAbraham: > Similar here it's not clear that this has been edited from a different value.

Not sure I follow what you mean.

> Also why has it gone back to the third block just to step through it to get to the one it wants to change?

It's the first question from the calculated summary it can go to before the `skip yes / no` question. The blocks in between aren't on the first calculated summary, they belong to unit calculated summary, etc.

app/views/contexts/calculated_summary_context.py
```diff
@@ -42,19 +43,24 @@ def build_view_context_for_calculated_summary(self, current_location):
...
         )
+        calculation = block["calculation"]
 
         groups = self.build_groups_for_section(calculated_section)
 
         formatted_total = self._get_formatted_total(
-            groups or [], current_location=current_location
+            groups or [],
+            current_location=current_location,
+            calculation_type=ValueSourceResolver.get_calculation_operator(
```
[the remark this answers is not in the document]
MebinAbraham: Fixed, initially spotted it but forgot to change.

tests/functional/spec/features/calculated_summary.spec.js
MebinAbraham: Haven't really change the existing behaviour, just removed the fourth and a half block being edited so that we have 2 different total when skipping / no skipping 4th questions. It may need a general refactor regardless of my changes.

MebinAbraham:
> The only other issue is whether a dependent type being used as a source should have the block removed when the dependent is changed. Although other placeholder don't I think this is slightly different and needs confirming, I will message Donna when next in.

I think we do need to be consistent, I don't think piping "total" should get any special treatment, something like name/age piping could be just as important. However, for calculated summary we should be okay as the scenarios we were presented with were using calculated summary total for validation (min/max), for which the blocks would be cleared. But see what Donna/Sarah comes back with as it might lead to us wanting to do the same for others.

LJBabbage:
> I think we do need to be consistent, I don't think piping "total" should get any special treatment, something like name/age piping could be just as important

Calculated Summary in itself is deemed to be important enough to be re-visited, and this is based on Calculated Summary and exercises the same code, so why slightly different to normal piping

tests/functional/spec/features/calculated_summary.spec.js
LJBabbage: The first command is thirdNumberAnswerEdit, the description fails to mention it must have had a value already and is being edited. Can it not go to skip-fourth-block? But I will let it go because it wasn't the best to start with

tests/functional/spec/features/calculated_summary.spec.js
LJBabbage: As it's me and although you have touched it, I will let you off

app/questionnaire/value_source_resolver.py
```diff
@@ -114,8 +117,37 @@ def _resolve_list_value_source(self, value_source: dict) -> Union[int, str, list
...
+        answers_to_calculate = calculation["answers_to_calculate"]
+        operator = self.get_calculation_operator(calculation["calculation_type"])
+        list_item_id = self._resolve_list_item_id_for_value_source(value_source)
+
+        values = [
+            self._get_answer_value(answer_id=answer_id, list_item_id=list_item_id)
+            for answer_id in answers_to_calculate
+        ]
+        total: IntOrDecimal = operator([value for value in values if value])  # type: ignore
+        return total
```
petechd, line 138: I would shorten it even more but it's a question of personal preference.
```suggestion
        calculated_summary_block: Mapping[str, Any] = self.schema.get_block(value_source["identifier"])  # type: ignore
        calculation = calculated_summary_block["calculation"]
        operator = self.get_calculation_operator(calculation["calculation_type"])
        list_item_id = self._resolve_list_item_id_for_value_source(value_source)

        values = [
            self._get_answer_value(answer_id=answer_id, list_item_id=list_item_id)
            for answer_id in calculation["answers_to_calculate"]
        ]
        return operator([value for value in values if value])
```

petechd: Since we already have a docstring we don't need to assign a name to the returned value?

MebinAbraham: It was initially for type hint reasons i.e `total: IntOrDecimal` but have changed anyway, as the return is already `IntOrDecimal` so `type: ignore` for the return statement would be unaffected.

petechd (approved):

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag="calculated-summary-value-source"
```
MebinAbraham, line 2: Revert before merge
```suggestion
tag=latest
```

PR #834, opened 2022-04-05
Use default retry for GCS clients

MebinAbraham:
Uses Google's default retry strategy for GCS clients (Submission/Feedback).
Our current implementation will only retry if we specify an object generation version and if that matched, however we do not do this, nor is suitable for our use case. You can read more about object generation and the precondition at:
- https://cloud.google.com/storage/docs/metadata#generation-number
- https://cloud.google.com/storage/docs/request-preconditions

The decision was to use the default retry with the accepted caveat that, it is not idempotent, however, the impact of that is understood since it is negligible for our use case.

### How to review 
Ensure the changes are appropriate and that GCS now handles retries for transient errors.

Since this is relying on GCP, you will have to fake a failure using a code change. The easiest way to test this is to point your local runner to an actual GCS bucket. You can then add `timeout=0.1` to the `upload_from_string` function call. This will force it to retry, as it cannot manage to upload in 0.1 seconds. Since it is a multipart upload, it will eventually succeed. You should see the multiple requests to GCS in the logs. 

*Note: timeout may be hit-and-miss based on your connection speed, so you may need to try multiple times. If all fails, let me know, and I'll show you how to change the code to force a retriable response.*

**__Local runner with a real GCP bucket__**
- Create a GCP project and a bucket to store your submissions. 
- Locally set the following env vars:
   - `EQ_SUBMISSION_BACKEND="gcs"`
   - `EQ_GCS_SUBMISSION_BUCKET_ID=<your-gcp-bucket>"`
- Locally Auth with Google and set project in the following order:
   - `gcloud auth revoke --all` - This may error if you have not authed before (can ignore)
   - `gcloud auth login`
   - `gcloud auth application-default login` 
   - `gcloud config set project <your-project>`
- Run dev containers:
   - `make dev-compose-up`
- Start runner and submit a questionnaire. (If you get an error, you may need to enable Datastore)
   - `make run`
- Check to see if your submission landed in your bucket and more importantly, that the retries are working.

### Checklist


LJBabbage (approved):
Slightly hit and miss with the timeout=0.1 as it didn't always fail, however when it did and with logging in the library I could see it retrying

tests/app/submitter/test_submitter.py
```diff
@@ -212,7 +212,46 @@ def test_gcs_submitter_adds_metadata_when_sends_message(patch_client):
...
+    # When
+    bucket = patch_gcs_client.return_value.get_bucket.return_value
+    bucket.blob.return_value = gcs_blob_with_retry
+
+    function_to_call = getattr(gcs_submitter, entrypoint)
+    successful = function_to_call(**data_to_upload)
+
+    # Then
+    assert (
+        gcs_blob_with_retry._get_transport().request.call_count  # pylint: disable=protected-access
```
LJBabbage, line 248: You could make this slightly clearer with the comments, its not immediately obvious why it gets called twice, unless you go back to the fixture and see you have a list of side_effects

berroar (approved):
Followed the steps and retries were working as described 👍

pricem14pc (commented):
As discussed, Investigating a `403` during a specific test approach when testing with a retention policy.

pricem14pc (approved):
Retry behaving as expected

PR #838, opened 2022-04-08
Fix IE11 countdown panel bug

rmccar:
There is currently a bug with countdown panels not showing the countdown on IE11. This is because the time being passed to the panel from runner is not in ISO format which IE11 seems to be struggling with. This PR will convert the time being passed to the panel to ISO format.

### How to review 
- Tests pass
- Panel still works as it should when used in the `test_view_submitted_response` schema

### Checklist


rmccar:
[Fix IE11 bug with countdown panel](https://trello.com/c/xb86geZf/5252-fix-ie11-bug-with-countdown-panel)

MebinAbraham (approved):
Working as expected in IE 11 👍

PR #848, opened 2022-04-21
Rules require dependent section routing path re-evaluation

LJBabbage:
This PR adds answer id section dependents. When an answer_id is changed which would affect the status of another section the routing path of that section is re-evaluated to make sure it is still correct
 
### How to review 
Using test_new_routing_and_skipping_section_dependencies answer the first question as yes to skipping and add a couple of household members. Complete the primary person and one of the household members, now go back and change your answer to skipping to no, you should see the statuses change of the completed sections. Changing the skipping answer back will see them complete again

### Checklist


app/questionnaire/questionnaire_store_updater.py
```diff
@@ -298,16 +325,33 @@ def update_answers(
...
+
+        for section in self.dependent_sections:
+
+            if section.is_complete is None:
+                routing_path = self._router.routing_path(
+                    section.section_id, list_item_id=section.list_id
+                )
+                is_path_complete = self._router.is_path_complete(routing_path)
+            else:
+                is_path_complete = section.is_complete
```
LJBabbage, line 344: I realise this can be refactored, but couldn't seem to get it to work. I wIll work it out before I go, but I didn't want to hold up the review considering my imminent move

petechd (approved):
Can't find anything to dislike about this PR, great usage of docstrings!

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -298,16 +325,32 @@ def update_answers(
...
+        self._remove_dependent_blocks_from_progress_store()
+
+        for section in self.dependent_sections:
+            is_path_complete = section.is_complete
+
+            if section.is_complete is None:
+                routing_path = self._router.routing_path(
+                    section.section_id, list_item_id=section.list_id
+                )
+                is_path_complete = self._router.is_path_complete(routing_path)
```
MebinAbraham, line 343: This is checking for all dependent sections regardless of their current status. We do not need to do this for sections that have not been started.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -330,10 +373,8 @@ def update_progress_for_dependant_sections(self) -> None:
                 section_id != self._current_location.section_id
                 or list_item_id != self._current_location.list_item_id
             ):
-                self.update_section_status(
-                    is_complete=False,
-                    section_id=section_id,
-                    list_item_id=list_item_id,
+                self.dependent_sections.add(
+                    DependentSection(section_id, list_item_id, False)
                 )
```
MebinAbraham, line 378: Since this is setting the status to `False` we should handle duplicate section dependents. It won't currently handle duplicates as this is setting is_complete to False and the other is setting it to `None`.

I had a think about this and I don't believe, in this scenario, it can ever not be false. If you remove a block for a section, that section would be incomplete, however yes it is possible that another section may become complete etc, but that is not what is being handled here. This is handling status for the section the block was removed from. Therefore, due to the nature of the current call order, we would need to do something like this, which is not the best but avoid duplicates.

```py
				dependent = DependentSection(section_id, list_item_id, None)
                if dependent in self.dependent_sections:
                    self.dependent_sections.remove(dependent)

                self.dependent_sections.add(
                    DependentSection(section_id, list_item_id, False)
                )
```

The other option is to have 1 map in questionnaire store updater to handle both section and block level dependencies. But that would require a bit of refactor, which I don't think is necessary given where we are.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -298,16 +325,32 @@ def update_answers(
...
+                is_path_complete = self._router.is_path_complete(routing_path)
+
+            self.update_section_status(
+                is_complete=is_path_complete,
+                section_id=section.section_id,
+                list_item_id=section.list_id,
+            )
+
+    def _remove_dependent_blocks_from_progress_store(self) -> None:
+        """Removes dependent blocks from the progress store."""
```
MebinAbraham, line 352: This method does more than what the name and doc strings says. Based on how you have named `_capture_section_dependencies_for_answer`, this should be something like `_remove_dependent_blocks_and_capture_dependent_sections`

MebinAbraham (commented):
Might need to revisit how runner handles dependencies. It is getting fairly involved, and the complexity is piling upon questionnaire store updater.

app/questionnaire/questionnaire_schema.py
```diff
@@ -78,6 +79,10 @@ def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
     def when_rules_section_dependencies_map(self) -> ImmutableDict[str, set[str]]:
         return ImmutableDict(self._when_rules_section_dependencies_map)
 
+    @cached_property
+    def answer_id_section_dependents_map(self) -> ImmutableDict[str, set[str]]:
+        return ImmutableDict(self._answer_id_section_dependents_map)
```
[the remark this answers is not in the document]
petechd: Ok, renamed both accordingly, thanks.

app/questionnaire/questionnaire_schema.py
```diff
@@ -846,6 +851,9 @@ def _get_rules_section_dependencies(
                 section_id = self.get_section_id_for_block_id(block["id"])  # type: ignore
 
                 if section_id != current_section_id:
+                    self._answer_id_section_dependents_map[answer_id] |= {
+                        current_section_id
+                    }
```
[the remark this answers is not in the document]
petechd: Simplified using this suggestion, thanks!

app/questionnaire/questionnaire_store_updater.py
petechd: Ok, renamed now, thanks!

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -284,6 +290,27 @@ def _capture_dependencies_for_answer(self, answer_id: str) -> None:
...
+                ):
+
+                    for list_id in self._list_store[repeating_list].items:
+                        self.dependent_sections.add(
+                            DependentSection(section_id, list_id, None)
+                        )
+                else:
+                    self.dependent_sections.add(
+                        DependentSection(section_id, None, None)
+                    )
```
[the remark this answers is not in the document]
petechd: Nesting reduced, added your suggestion, thanks!

tests/app/questionnaire/test_questionnaire_store_updater.py
```diff
@@ -11,11 +11,13 @@
 from app.questionnaire.questionnaire_store_updater import QuestionnaireStoreUpdater
```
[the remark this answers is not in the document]
petechd: Flaky for me as well, trying to find a fix.

app/questionnaire/questionnaire_store_updater.py
petechd: I went for option 1 here and just added your suggested code, had to add some more tests.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -298,16 +325,32 @@ def update_answers(
...
             answer_updated = self._update_answer(answer_id, list_item_id, answer_value)
             if answer_updated:
-                self._capture_dependencies_for_answer(answer_id)
+                self._capture_section_dependencies_for_answer(answer_id)
+                self._capture_block_dependencies_for_answer(answer_id)
 
-    def update_progress_for_dependant_sections(self) -> None:
-        """Removes dependent blocks from the progress store and updates the progress to IN_PROGRESS.
+    def update_progress_for_dependent_sections(self) -> None:
+        """Section progress is not updated for the current location as it is handled by `handle_post` on block handlers. """
```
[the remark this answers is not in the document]
petechd: Kept, the original sentences and Lewis' one as the second line.

app/questionnaire/questionnaire_schema.py
```diff
@@ -78,6 +79,10 @@ def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
     def when_rules_section_dependencies_map(self) -> ImmutableDict[str, set[str]]:
         return ImmutableDict(self._when_rules_section_dependencies_map)
 
+    @cached_property
+    def answer_id_section_dependents_map(self) -> ImmutableDict[str, set[str]]:
```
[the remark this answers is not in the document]
petechd: Two new unit tests added for both methods, schema has to stay in this file as it would push `confest.py` over the limit.

rmccar (approved):

app/questionnaire/questionnaire_store_updater.py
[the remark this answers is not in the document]
MebinAbraham: Still outstanding.

tests/app/questionnaire/test_questionnaire_store_updater.py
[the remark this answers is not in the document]
MebinAbraham: Solved?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -298,16 +327,39 @@ def update_answers(
...
+        self._remove_dependent_blocks_and_capture_dependent_sections()
+
+        for section in self.dependent_sections:
+            is_path_complete = section.is_complete
+            is_started = (
+                self._progress_store.get_section_status(
+                    section.section_id, section.list_id
+                )
+                != "NOT_STARTED"
+            )
```
MebinAbraham, line 346: We already have a `self.started_section_keys()`, we should use that.

tests/app/questionnaire/test_questionnaire_schema.py
```diff
@@ -5,6 +5,63 @@
...
+                                },
+                                "No",
+                            ]
+                        },
+                    },
+                },
+            ],
+        },
+    ]
+}
```
MebinAbraham, line 63: Given we have gone down the road of already loading in existing schemas for the other dependency tests, we should do the same here and avoid putting schema in the code.

petechd: I tried but conftest.py would exceed the 1200 lines limit so had to put it here...

MebinAbraham: We should load in an existing schema not put json in that is basically invalid anyway. i.e you can't have exmpty questions etc. Have a look at the existing test for dependency which uses load schema method.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -298,16 +327,39 @@ def update_answers(
...
+
+            if section.is_complete is None and is_started:
```
MebinAbraham, line 348: I think these two should be separate. As something would be wrong if section.is_complete was not None and `is_started` was False. So just skip if section not started, and then reevaulate if is_complete is None/

MebinAbraham, line 348: We should do `is_started` logic inline, as you only need to do that is `section.is_complete` is None. If it was True/False, then the section had to have started.

tests/app/questionnaire/test_questionnaire_schema.py
petechd: Added reference to existing schema, thanks!

app/questionnaire/questionnaire_store_updater.py
petechd: Yeah I read your mind and put it inline, before even reading this comment!

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -298,16 +327,36 @@ def update_answers(
...
 
-        Section progress is not updated for the current location as it is handled by `handle_post` on block handlers.
+        self._remove_dependent_blocks_and_capture_dependent_sections()
+
+        for section in self.dependent_sections:
+            is_path_complete = section.is_complete
+
+            if (
+                section.is_complete is None
+                and section.section_id not in self.started_section_keys()
```
MebinAbraham, line 344: This would always be type. `started_section_keys` are tuples of sections keys, whereas section_id is a string. You need to check section key is not in `started_section_keys`.

tests/app/questionnaire/test_questionnaire_schema.py
```diff
@@ -5,6 +5,63 @@
 
 from app.questionnaire.questionnaire_schema import AnswerDependent, QuestionnaireSchema
 
+WHEN_RULES_SECTION_DEPENDENCIES_SCHEMA = {
```
MebinAbraham, line 8: Needs to be removed.

petechd: Schema removed thanks!

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -298,16 +327,35 @@ def update_answers(
...
+        Section progress is not updated for the current location as it is handled by `handle_post` on block handlers."""
 
-        Section progress is not updated for the current location as it is handled by `handle_post` on block handlers.
+        self._remove_dependent_blocks_and_capture_dependent_sections()
+
+        for section in self.dependent_sections:
+            is_path_complete = section.is_complete
+
+            if (section.section_id, section.list_id) in self.started_section_keys():
+                if section.is_complete is None:
```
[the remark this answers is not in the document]
petechd: Ok, added, thanks.

tests/app/questionnaire/test_questionnaire_store_updater.py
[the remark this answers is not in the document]
berroar: Added a fix for this

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -330,10 +378,12 @@ def update_progress_for_dependant_sections(self) -> None:
                 section_id != self._current_location.section_id
                 or list_item_id != self._current_location.list_item_id
             ):
-                self.update_section_status(
-                    is_complete=False,
-                    section_id=section_id,
-                    list_item_id=list_item_id,
+                dependent = DependentSection(section_id, list_item_id, None)
+                if dependent in self.dependent_sections:
+                    self.dependent_sections.remove(dependent)
```
MebinAbraham, line 383: I think this warrants a comment, as it is not obvious why we are removing.

tests/app/questionnaire/test_questionnaire_store_updater.py
```diff
@@ -627,13 +653,231 @@ def test_update_repeating_answers_with_answer_dependents(mock_schema):
...
+        current_location=location,
+        current_question=current_question,
+    )
+    questionnaire_store_updater.update_answers(form_data)
+
+    # This test case is dependent on the order that the dependent_sections set is iterated over,
+    # however as python sets are unordered we need to check that the first item is equal to our expected
+    # list_item_id so that we can set the correct side effect as per the test case
+    dependent_sections_set = questionnaire_store_updater.dependent_sections.copy()
+    first_item = dependent_sections_set.pop() if dependent_sections_set else None
```
MebinAbraham, line 851: Can be simplified.
```suggestion
    first_item = next(iter(questionnaire_store_updater.dependent_sections), None)
```

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -330,10 +378,14 @@ def update_progress_for_dependant_sections(self) -> None:
                 section_id != self._current_location.section_id
                 or list_item_id != self._current_location.list_item_id
             ):
-                self.update_section_status(
-                    is_complete=False,
-                    section_id=section_id,
-                    list_item_id=list_item_id,
+                # Remove any duplicate session dependents if they exist before setting
+                # the dependent section `is_complete` status to False
```
MebinAbraham, line 382: Sorry if I wasn't clear. I mean, why we need to do this instead of what it does. The code is self explanatory, we know what it is doing but not why.

app/questionnaire/questionnaire_store_updater.py
berroar: Added a comment

tests/app/questionnaire/test_questionnaire_store_updater.py
berroar: Added this 👍

MebinAbraham (approved):

MebinAbraham:
We'll need to revisit dependency handling at some point. It is a bit messy with all the different bits we have had to add.

PR #850, opened 2022-04-21
Make calculated summary return_to conditional

MebinAbraham:
- Made the `return_to` query string from the change link for Calculated summaries conditional on section summary/hub flow. We currently do not support jumping straight back to a calculated summary. Support for jumping straight back to a calculated summary will be implemented separately.
- Also removed tests in `test_calculated_summary_context.py` that were duplicated from `test_section_summary_context.py`

***Note: This is a temporary solution***

### How to review 
Ensure the change links on a calculated summary depend on the whether you have a section summary and or if hub is enabled.

- if section summary is enabled for the section you are on, then return to should be `section-summary`
- if section summary is disabled and hub is disabled, then return to should be `final-summary`
- otherwise, return to should be `None`. (This would go to the first block in the section since, hub without section summary would go to the first block anyway when you click (View answers))

Test the journeys do not result in 404s.
Test the following journey on master vs this branch.
- Launch construction
- Complete all sections
- From the hub, click View answer to the `Infrastructure` section. You should now be on the section summary.
- Click `Previous` to go to the Calculated summary.
- Click `Change` on one of the answers.
- Click `Save and continue` and you should see a 404 on master, but go to the section summary on this branch. 

### Checklist

### DIFF
--- a/app/views/contexts/calculated_summary_context.py
+++ b/app/views/contexts/calculated_summary_context.py
@@ -16,6 +16,17 @@ from app.views.contexts.summary.group import Group
 
 
 class CalculatedSummaryContext(Context):
+    def get_return_to(self, section_id: str):
+        # :TODO: Add support for returning to the calculated summary that was used
+        has_section_summary = bool(self._schema.get_summary_for_section(section_id))
+        if has_section_summary:
+            return "section-summary"
+
+        if not self._schema.is_flow_hub:
+            return "final-summary"
+
+        return None
+
     def build_groups_for_section(self, section):
         routing_path = self._router.routing_path(section["id"])
 
@@ -32,7 +43,7 @@ class CalculatedSummaryContext(Context):
                 self._schema,
                 location,
                 self._language,
-                return_to="final-summary",
+                return_to=self.get_return_to(section["id"]),
             ).serialize()
             for group in section["groups"]
         ]
--- a/app/views/contexts/summary/answer.py
+++ b/app/views/contexts/summary/answer.py
@@ -45,6 +45,6 @@ class Answer:
             block_id=block_id,
             list_item_id=list_item_id,
             return_to=return_to,
-            return_to_answer_id=self.id,
+            return_to_answer_id=self.id if return_to else None,
             _anchor=self.id,
         )
--- a/app/views/contexts/summary/question.py
+++ b/app/views/contexts/summary/question.py
@@ -60,7 +60,7 @@ class Question:
                 block_id=block_id,
                 list_item_id=self.list_item_id,
                 return_to=return_to,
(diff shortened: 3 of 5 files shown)


app/views/contexts/calculated_summary_context.py
```diff
@@ -16,6 +16,17 @@
 
 
 class CalculatedSummaryContext(Context):
+    def get_return_to(self, section_id: str):
+        # :TODO: Add support for returning to the calculated summary that was used
+        has_section_summary = bool(self._schema.get_summary_for_section(section_id))
+        if has_section_summary:
+            return "section-summary"
```
petechd, line 23:
```suggestion
        if has_section_summary := bool(self._schema.get_summary_for_section(section_id)):
            return "section-summary"
```

petechd: We can add a doctring explaining what we do instead of assigning a var?

app/views/contexts/calculated_summary_context.py
```diff
@@ -32,7 +43,7 @@ def build_groups_for_section(self, section):
                 self._schema,
                 location,
                 self._language,
-                return_to="final-summary",
+                return_to=self.get_return_to(section["id"]),
```
petechd, line 46:
```suggestion
                return_to=self.get_return_to(section.get("id")),
```

LJBabbage (approved):
It works, but it ain't pretty, to set a return to go to a location that isn't itself is not great, it might have been better to bin the return_to, but hey ho. With validator not enforcing calculated summary as the last block (at least I don't believe it does on a quick check) You open the door to some weird behaviour if it wasn't. One to revisit sooner rather than later, but for construction is should be ok

app/views/contexts/calculated_summary_context.py
MebinAbraham: I don't follow.

app/views/contexts/calculated_summary_context.py
MebinAbraham: Do we need one?.

MebinAbraham:
> it might have been better to bin the return_to

I did that initially and ran it by Donna, in that scenario you return to the hub straight from the question you are changing. Donna preferred going to the section summary.

app/views/contexts/calculated_summary_context.py
petechd: If it's for removal I am not precious about it and it can stay.

app/views/contexts/calculated_summary_context.py
petechd: I just meant, section needs to be a dict(?) so we should use dict specific method, but just a detail.

PR #854, opened 2022-04-27
Ability to return to calculated summary

berroar:

Adds the ability to return to a calculated summary block after using the change link. If an answer is changed via a calculated summary page, the `previous` and `save and continue` buttons will take the user back to the calculated summary page. Have added a new `return_to_block_id` request argument in order to support this behaviour. See the trello card for more info.

### How to review 

Use the `test_calculated_summary` schema to test this change. Test changing an answer on a calculated summary block and make sure that you are returned to the block when selecting `previous` or `save and continue`.

### Checklist


berroar:
[Ability to return to Calculated summary (3 SP)](https://trello.com/c/XVHmTBF4/5265-ability-to-return-to-calculated-summary-3-sp)

app/questionnaire/router.py
```diff
@@ -382,3 +401,12 @@ def _get_section_url(
...
         )
+
+    def get_calculated_summary_url(
+        self, location: Location, return_to: str, return_to_block_id: str
+    ) -> str:
+        return self._get_return_to_location_url(
+            location,
+            return_to,
+            return_to_block_id=return_to_block_id,
+        )
```
MebinAbraham, line 412: This could be inline once the above changes are done.

MebinAbraham, line 412: This could be inline once the above changes are done.

app/views/contexts/calculated_summary_context.py
```diff
@@ -43,20 +32,23 @@ def build_groups_for_section(self, section):
...
+                return_to="calculated-summary",
+                return_to_block_id=return_to_block_id,
             ).serialize()
             for group in section["groups"]
         ]
 
     def build_view_context_for_calculated_summary(self, current_location):
         block = self._schema.get_block(current_location.block_id)
 
+        return_to_block_id = block.get("id")
```
MebinAbraham, line 44: Can use bracket notation since `id` is mandatory.

MebinAbraham, line 44: Can use bracket notation since `id` is mandatory.

tests/functional/spec/features/calculated_summary.spec.js
```diff
@@ -92,25 +92,18 @@ describe("Feature: Calculated Summary", () => {
       $(CurrencyTotalPlaybackPageWithFourth.fourthNumberAnswerEdit()).click();
```
MebinAbraham, line 92: I think it would be good to show functionally, the behaviour of returning to a calculated summary. We should have the same test as we do for section summary. That I am return to the summary and that the anchor exists as expected.

berroar: Added functional tests 👍

app/questionnaire/router.py
```diff
@@ -205,19 +216,28 @@ def get_previous_location_url(
...
+            return_to == "final-summary"
+            and is_section_complete
+            and self.is_questionnaire_complete
+        ):
             return url_for(
                 "questionnaire.submit_questionnaire", _anchor=return_to_answer_id
             )
 
+        if return_to == "calculated-summary" and return_to_block_id:
+            return url_for("questionnaire.block", block_id=return_to_block_id)
```
MebinAbraham, line 239: These `if` statements will need to be nested. i.e if return to is `"section-summary"` but section isn't complete, we shouldn't evaluate the next if and so on. We should know at that point to return None.

Adding on to my previous comment, it could be something like:

```py
        if not return_to:
            return None

        if return_to == "calculated-summary":
            if return_to_block_id:
                return url_for("questionnaire.block", block_id=return_to_block_id)

            return None

        if is_section_complete is None:
            is_section_complete = self._progress_store.is_section_complete(
            )

        if not is_section_complete:
            return None

        if return_to == "section-summary":
            return self._get_section_url(
                location, return_to_answer_id=return_to_answer_id
            )
        if return_to == "final-summary" and self.is_questionnaire_complete:
            return url_for(
                "questionnaire.submit_questionnaire", _anchor=return_to_answer_id
            )
```
I agree it is a lot of `ifs` but unavoidable without a redesign.

app/questionnaire/router.py
```diff
@@ -205,19 +216,28 @@ def get_previous_location_url(
...
+        if (
+            return_to == "final-summary"
+            and is_section_complete
+            and self.is_questionnaire_complete
+        ):
             return url_for(
                 "questionnaire.submit_questionnaire", _anchor=return_to_answer_id
             )
 
+        if return_to == "calculated-summary" and return_to_block_id:
```
MebinAbraham, line 238: Reminder that `return_to_block_id` needs to be validated.

MebinAbraham, line 238: Reminder that `return_to_block_id` needs to be validated.

app/questionnaire/router.py
```diff
@@ -117,21 +117,23 @@ def get_next_location_url(
...
+            location,
+            return_to,
+            return_to_answer_id=return_to_answer_id,
+            return_to_block_id=return_to_block_id,
+        ):
+            return return_to_url
+
         if self._progress_store.is_section_complete(
             location.section_id, location.list_item_id
         ):
```
MebinAbraham, line 136: So in the suggestion, we were still passing `is_section_complete` as this section will need to know if section is complete anyway, therefore it avoided calling `is_section_complete` twice. i.e here if you have return to section summary, it would evaluate section completeness in get_return_to method and then if section is not completed, it would then revaluate completeness here.

app/questionnaire/router.py
```diff
@@ -207,12 +205,28 @@ def _get_return_to_location_url(
...
     ) -> Optional[str]:
+
+        if not return_to:
+            return None
+
+        if (
+            return_to == "calculated-summary"
+            and return_to_block_id
+            and self._schema.is_block_valid(return_to_block_id)
+        ):
```
MebinAbraham, line 218: The checks shouldn't be chained, it has to be nested. if you have calculated summary, but block wasn't valid, we do not need to progress down, we should return None. Same applies to the others, i.e returning to section summary, but section isn't complete, hence we can return None, currently it would then check if we are returning to final summary etc, then eventually return None hence why tests pass.

app/questionnaire/router.py
berroar: Addeed `is_section_complete` again

app/questionnaire/router.py
berroar: Nested these and had to add some tests to ensure code coverage.

tests/app/questionnaire/test_router.py
```diff
@@ -506,6 +506,81 @@ def test_section_summary_on_completion_false(self):
...
+    @pytest.mark.usefixtures("app")
+    def test_return_to_calculated_summary_invalid_return_to_block_id(self):
+        self.schema = load_schema_from_name("test_calculated_summary")
+
+        current_location = Location(
+            section_id="default-section", block_id="fifth-number-block"
+        )
+
+        routing_path = RoutingPath(
+            ["sixth-number-block", "fifth-number-block"],
```
MebinAbraham, line 543: Routing path would be linear.

tests/app/questionnaire/test_router.py
```diff
@@ -506,6 +506,81 @@ def test_section_summary_on_completion_false(self):
...
+        )
+        expected_location_url = Location(
+            section_id="default-section",
+            block_id="sixth-number-block",
+        ).url()
+
+        assert expected_location_url == next_location_url
+
+    @pytest.mark.usefixtures("app")
+    def test_return_to_calculated_summary_empty_return_to_block_id(self):
```
MebinAbraham, line 560: This and `test_return_to_calculated_summary_invalid_return_to_block_id` could be one test and parametrized.
Also, would it not be `None` instead of `""`.?

app/questionnaire/router.py
```diff
@@ -205,14 +208,36 @@ def get_previous_location_url(
...
+        return_to_block_id: Optional[str] = None,
     ) -> Optional[str]:
+        # pylint: disable=too-many-return-statements
+        if not return_to:
+            return None
+
+        if return_to == "calculated-summary" and return_to_block_id:
+            if (
+                self._schema.is_block_valid(return_to_block_id)
+                and return_to_block_id in routing_path.block_ids
```
MebinAbraham, line 224: Has been chained the wrong way around. It should be:
```py
if calculated_summary:
  if return_to_block and is_on_path():
    return url
  
  return None
```

rmccar (approved):

tests/app/questionnaire/test_router.py
```diff
@@ -506,6 +506,65 @@ def test_section_summary_on_completion_false(self):
```
MebinAbraham, line 506: We will need a test for when the block is valid but not on the path.

tests/app/views/contexts/summary/test_answer.py
```diff
@@ -5,17 +5,18 @@
 
 @pytest.mark.usefixtures("app")
 @pytest.mark.parametrize(
-    "return_to",
-    ["section-summary", None],
+    "return_to, return_to_block_id",
+    [("section-summary", None), (None, None)],
 )
-def test_create_answer(return_to):
+def test_create_answer(return_to, return_to_block_id):
```
MebinAbraham, line 11: The test asserts have not been updated, hence this is redundant since return_to_block_id is `None` in both cases. Need to test with a string value etc.

tests/app/questionnaire/test_router.py
```diff
@@ -506,6 +506,101 @@ def test_section_summary_on_completion_false(self):
...
+            current_location,
+            routing_path,
+            return_to="calculated-summary",
+            return_to_block_id="fourth-number-block",
+        )
+
+        assert (
+            "/questionnaire/sixth-number-block/?return_to=calculated-summary&return_to_block_id=fourth-number-block"
+            == next_location_url
+        )
```
MebinAbraham, line 602: This warrants a comment to explain why the same query string is passed along. i.e it may be on the path in future once incomplete questions are answered.

MebinAbraham (commented):
Looks good, just a couple of tweaks.

PR #855, opened 2022-04-28
Bug fix: Only include first 11 characters of ru_ref in help link

MebinAbraham:
- Temporary tactical solution to only send RAS the first 11 characters of the `ru_ref` when generating the help link. The first 11 characters are send because the reporting unit reference is always 11 characters.
- The `ru_ref` eQ gets from RAS actually includes `ru_ref` + `check_letter` but when using the help link, they do not expect the `check_letter` to be present. They will progress a more appropriate solution to their end in the future, but this is a temp fix to get the changes in preprod moving.

### How to review 
- Launch a schema and set the ru ref to your liking, and ensure that only the first 11 characters are used when generating the help link ru_ref query param.

### Checklist

### DIFF
--- a/tests/app/helpers/test_template_helpers.py
+++ b/tests/app/helpers/test_template_helpers.py
@@ -158,6 +158,23 @@ def test_get_page_header_context_census_nisra(app: Flask):
             None,
         ),
         (
+            BusinessSurveyConfig(),
+            False,
+            {
+                "toggleServicesButton": {
+                    "text": "Menu",
+                    "ariaLabel": "Toggle services menu",
+                },
+                "itemsList": [
+                    {
+                        "title": "Help",
+                        "url": "https://surveys.ons.gov.uk/help",
+                        "id": "header-link-help",
+                    }
+                ],
+            },
+        ),
+        (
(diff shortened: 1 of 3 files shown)


rmccar (approved):

petechd (approved):

tests/app/helpers/test_template_helpers.py
```diff
@@ -157,7 +158,7 @@ def test_get_page_header_context_census_nisra(app: Flask):
             None,
         ),
         (
-            BusinessSurveyConfig(),
+            BusinessSurveyConfig(schema=QuestionnaireSchema({"survey_id": "001"})),
```
petechd, line 161: Don't we want to keep a Business Survey Config test with no schema present?

MebinAbraham: Added a test for not authenticated. Current one wasn't quite right since it didn't have query param when authenticated. Didn't do no schema as if you are authed, you will have schema.

petechd: Ok, thanks. Just because there is such scenario in the code I preferred to have something in this tests.

berroar (approved):

PR #856, opened 2022-04-28
Update the anchor for concatenated answers on change link

berroar:
Updates the anchor for concatenated answers on summary page `change` links so that the input box for the first of the concatenated answers is focused on the question page being changed.

### How to review 
Use the `test_submit_with_summary_return_to_answer` schema to test this change. To test:
Complete the Name question and then go to the summary page. 
Click the change link for the Name question on the summary page
The `first-name` box should be highlighted when you are returned to the Name question.
Click `Save and Continue` to back to the summary - the anchor should be on the concatenated Name field

### Checklist


berroar:
[Anchor for concatenated answers Change link contains an unusable anchor](https://trello.com/c/Es3jw0u0/5113-anchor-for-concatenated-answers-change-link-contains-an-unusable-anchor)

app/views/contexts/summary/question.py
```diff
@@ -61,7 +60,7 @@ def _build_answers(
                 list_item_id=self.list_item_id,
                 return_to=return_to,
                 return_to_answer_id=answer_id if return_to else None,
-                _anchor=answer_id,
+                _anchor=question_schema.get("answers")[0]["id"],
```
MebinAbraham, line 63: "answers" is mandatory
```suggestion
                _anchor=question_schema["answers"][0]["id"],
```

MebinAbraham (approved):
There is an odd case whereby the anchor for the next screen should be set based on `return_to_answer_id`. However, even though we don't do that (unless return_to is used), because `return_to` does change the anchor to use `return_to_answer_id`, it looks like we should be okay.

PR #857, opened 2022-04-28
fix `mobilenumber` input type width

rmccar:
The `mobilenumber` type input field is only wide enough for 8 characters so will not fit a full mobile number especially if you suffix it with "+44". This will up the width to 15

### How to review 
- Open `test_multiple_answers` schema
- Enter "+44 0000 000000" into the mobile number field and see that the mobile number field is now wide enough for a full mobile number

### Checklist


rmccar:
[Mobilenumber input type not wide enough for full mobile number](https://trello.com/c/ixACb67U/5279-mobilenumber-input-type-not-wide-enough-for-full-mobile-number)

MebinAbraham (approved):
Was it like this in census 🤔

berroar (approved):

rmccar:
> Was it like this in census 🤔

I don't think so, I did some work on the DS a while ago to make the width value match up with the number of characters you expect in the input which is why this is now wrong but I think that was after census

PR #858, opened 2022-05-03
Add failed requests handling and retries for survey url

petechd:
Adds all the scenarios described [on this Trello card](https://trello.com/c/m8aROxY9) when obtaining schema from url.

### How to review 
Couple more tests added to check specific situations. Check the tests.

### Checklist


petechd:
[Runner to handle failed requests to survey_url (2 SP)](https://trello.com/c/m8aROxY9/5240-runner-to-handle-failed-requests-to-surveyurl-2-sp)

app/utilities/schema.py
```diff
@@ -177,7 +178,44 @@ def load_schema_from_url(schema_url, language_code):
...
+            501,
+            502,
+            503,
+            504,
+            505,
+            506,
+            507,
+            509,
+            510,
+            511,
```
MebinAbraham, line 200: Not all these are retriable. We should only need 500, 502, 503, 504.

app/utilities/schema.py
```diff
@@ -177,7 +178,44 @@ def load_schema_from_url(schema_url, language_code):
...
+        ],
+    )  # Codes to retry according to Google Docs
```
MebinAbraham, line 202: Worth linking to the docs.

app/utilities/schema.py
```diff
@@ -177,7 +178,44 @@ def load_schema_from_url(schema_url, language_code):
...
+
+    session.mount("https://", HTTPAdapter(max_retries=retries))
```
MebinAbraham, line 204: I think we should do this for both http and https. We should get https, but we shouldn't care since we are just fetching artifacts. We currently don't enforce that it must be https, but that is something we should do.

app/utilities/schema.py
```diff
@@ -177,7 +178,44 @@ def load_schema_from_url(schema_url, language_code):
 
     constructed_schema_url = f"{schema_url}?language={language_code}"
 
-    req = requests.get(constructed_schema_url)
+    session = requests.Session()
+
+    retries = Retry(
+        total=3,
+        allowed_methods=frozenset({"GET"}),
```
MebinAbraham, line 185: This doesn't need to be a frozenset?. Also does this have a default?.

app/utilities/schema.py
```diff
@@ -177,7 +178,44 @@ def load_schema_from_url(schema_url, language_code):
...
+        backoff_factor=0.1,
```
MebinAbraham, line 186: The card specifies no back off. Since the timeout is 3 seconds, with a backoff of 0.1, what is the longer time before a user may see a response?.

Edit:
Looks like with 0.1 it would be `[0.0s, 0.2s, 0.4s, …]`. 

Since we may change total retries at any point, we should set `backoff_max` to 0.5 seconds.

app/utilities/schema.py
```diff
@@ -177,7 +178,44 @@ def load_schema_from_url(schema_url, language_code):
...
+            511,
+        ],
+    )  # Codes to retry according to Google Docs
+
+    session.mount("https://", HTTPAdapter(max_retries=retries))
+
+    try:
+        req = session.get(constructed_schema_url, timeout=3)
+    except Exception as exc:
+        raise ConnectionError from exc
```
MebinAbraham, line 209: We should avoid catching a generic exception, it is usually bad practice. We should handle the errors that can we expect to occur. The rest should get raised.

All exceptions requests can raise will be caught under `RequestException`. Does the retry object retry ConnectionError implicitly since that doesn't return a status code?. Otherwise, we would need to catch and handle connection error first, as it won't strictly return a status code. Therefore, making the request with retry its own method would be beneficial as you need to manually retry.

app/utilities/schema.py
```diff
@@ -177,7 +178,44 @@ def load_schema_from_url(schema_url, language_code):
 
     constructed_schema_url = f"{schema_url}?language={language_code}"
 
-    req = requests.get(constructed_schema_url)
+    session = requests.Session()
+
+    retries = Retry(
```
MebinAbraham, line 183: I think the total retries, the status codes to retry, the max backoff and the timeout should be module level constants.

tests/app/utilities/test_schema.py
```diff
@@ -155,6 +155,20 @@ def test_load_schema_from_url_200():
     assert cache_info.hits == 0
 
 
+@responses.activate
+def test_load_schema_from_url_connection_error():
+    load_schema_from_url.cache_clear()
+
+    responses.add(responses.GET, "", json={"error": "not found"}, status=0)
```
MebinAbraham, line 162: Is this line needed?.

petechd: Yeah, works fine without it so I will remove it, thanks.

app/utilities/schema.py
petechd: Defaults are `DEFAULT_ALLOWED_METHODS = frozenset({'DELETE', 'GET', 'HEAD', 'OPTIONS', 'PUT', 'TRACE'})`. "By default, we only retry on methods which are considered to be idempotent (multiple requests with the same parameters end with the same state)". Do we want to only keep `GET`?

app/utilities/schema.py
petechd: Added `session.mount("http://", HTTPAdapter(max_retries=retries))`, thanks.

app/utilities/schema.py
petechd: Added all these, I am not sure do we want to keep ALLOWED_METHODS as a set with `GET` only?

app/utilities/schema.py
petechd: Overwriting default BACKOFF_MAX with value of 0.2, thanks!

app/utilities/schema.py
petechd: Link added, thanks!

app/utilities/schema.py
petechd: OK, so I only left the ones that are retriable, thanks.

app/utilities/schema.py
petechd: Changed `Exception` to `RequestException`. I went through some topics about `Retry` and `ConnectionError` indeed  is being implicitly retried despite no status code, so it will work fine here. Thanks!

app/utilities/schema.py
```diff
@@ -177,7 +178,44 @@ def load_schema_from_url(schema_url, language_code):
...
+    except Exception as exc:
+        raise ConnectionError from exc
+
+    if req.status_code == 404:
+        logger.error("no schema exists", schema_url=constructed_schema_url)
+        raise NotFound
+
+    if req.status_code != 200:
+        logger.error(f"status code {req.status_code}")
+        raise InternalServerError
```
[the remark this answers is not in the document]
petechd: Ok, this logic applied now, all non 200 bundled up together and resulting in log entry and `Exception`, thanks!

app/utilities/schema.py
petechd: Removed.

app/utilities/schema.py
petechd: Ok, removed allowed methods, thanks!

app/utilities/schema.py
```diff
@@ -177,20 +190,37 @@ def load_schema_from_url(schema_url, language_code):
...
 
-    logger.info(
-        f"schema request took {response_duration_in_milliseconds:.2f} milliseconds",
-        pid=pid,
-    )
+    retries = Retry(
+        total=RETRIES_TOTAL,
+        allowed_methods=ALLOWED_METHODS,
+        status_forcelist=STATUS_CODES,
+    )  # Codes to retry according to Google Docs https://cloud.google.com/storage/docs/retry-strategy#retryable
```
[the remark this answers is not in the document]
petechd: Added, thanks!

app/utilities/schema.py
```diff
@@ -22,6 +23,18 @@
...
+BACKOFF_MAX = 0.2
+RETRIES_TOTAL = 3
+STATUS_CODES = [
+    408,
+    429,
+    500,
+    502,
+    503,
+    504,
+]
```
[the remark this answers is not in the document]
petechd: Added, thanks!

tests/app/utilities/test_schema.py
```diff
@@ -155,14 +154,27 @@ def test_load_schema_from_url_200():
...
+        load_schema_from_url(schema_url=TEST_SCHEMA_URL, language_code="en")
+
+    cache_info = load_schema_from_url.cache_info()
+    assert cache_info.currsize == 0
+    assert cache_info.misses == 1
+    assert cache_info.hits == 0
+
+
 @responses.activate
 def test_load_schema_from_url_404():
```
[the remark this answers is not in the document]
petechd: Changed it to `from_url_non_200` and parametrized, couple of status codes, not sure which you want, thanks!

app/utilities/schema.py
```diff
@@ -177,20 +190,37 @@ def load_schema_from_url(schema_url, language_code):
...
+        )
 
-    if req.status_code == 404:
-        logger.error("no schema exists", schema_url=constructed_schema_url)
-        raise NotFound
+        return QuestionnaireSchema(json_loads(schema_response), language_code)
 
-    return QuestionnaireSchema(json_loads(schema_response), language_code)
+    logger.error(f"status code {req.status_code}")
+    raise Exception
```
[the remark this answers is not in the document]
petechd: Done, thanks!

app/utilities/schema.py
```diff
@@ -177,20 +190,37 @@ def load_schema_from_url(schema_url, language_code):
...
+
+    retries.BACKOFF_MAX = BACKOFF_MAX
+
+    session.mount("http://", HTTPAdapter(max_retries=retries))
+    session.mount("https://", HTTPAdapter(max_retries=retries))
+
+    try:
+        req = session.get(constructed_schema_url, timeout=3)
+    except RequestException as exc:
+        raise ConnectionError from exc
```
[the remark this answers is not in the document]
petechd: Added this as custom exception, in separate file.

rmccar (commented):
Looks good as long as the tests suggested are added 👍

app/utilities/schema.py
```diff
@@ -22,6 +24,18 @@
 
 LANGUAGES_MAP = {"test_language": [["en", "cy"]]}
 
+SCHEMA_REQUEST_MAX_BACKOFF = 0.2
+SCHEMA_REQUEST_MAX_RETRIES = 2  # Totals no. of request should be 3.
```
MebinAbraham, line 28: Worth expanding. `# Totals no. of request sshould be 3. The initial request + SCHEMA_REQUEST_MAX_RETRIES`

app/utilities/schema.py
```diff
@@ -177,20 +191,45 @@ def load_schema_from_url(schema_url, language_code):
 
     constructed_schema_url = f"{schema_url}?language={language_code}"
 
-    req = requests.get(constructed_schema_url)
-    schema_response = req.content.decode()
-    response_duration_in_milliseconds = req.elapsed.total_seconds() * 1000
+    session = requests.Session()
```
MebinAbraham, line 194: This will need revisiting in future, we most likely want to reuse a single session. Session already handles connection pooling for us.

app/utilities/schema_request_failed_exception.py
```diff
@@ -0,0 +1,2 @@
+class SchemaRequestFailed(Exception):
```
MebinAbraham, line 1: I don't think we need a separate file for this.

app/utilities/schema.py
```diff
@@ -177,20 +191,45 @@ def load_schema_from_url(schema_url, language_code):
...
+        return QuestionnaireSchema(json_loads(schema_response), language_code)
+
+    logger.error(
+        "got a non-200 response for schema url request",
+        status_code=req.status_code,
+        schema_url=constructed_schema_url,
+    )
 
-    return QuestionnaireSchema(json_loads(schema_response), language_code)
+    raise SchemaRequestFailed("failed to load schema from url")
```
MebinAbraham, line 232: Worth overriding `__str__` of this class and returning `"schema request failed"`, then it can just be `raise SchemaRequestFailed` and message would be consistent. i.e we would like to alert on this in future

tests/app/utilities/test_schema.py
```diff
@@ -155,21 +160,37 @@ def test_load_schema_from_url_200():
...
     load_schema_from_url.cache_clear()
-
     mock_schema = QuestionnaireSchema({})
-    responses.add(responses.GET, TEST_SCHEMA_URL, json=mock_schema.json, status=404)
+    responses.add(
+        responses.GET, TEST_SCHEMA_URL, json=mock_schema.json, status=status_code
+    )
 
-    with pytest.raises(NotFound):
+    with pytest.raises(SchemaRequestFailed) as excinfo:
```
MebinAbraham, line 175: Minor, but consistency, `exc`?

tests/app/utilities/test_schema.py
```diff
@@ -155,21 +160,37 @@ def test_load_schema_from_url_200():
...
     assert cache_info.misses == 1
     assert cache_info.hits == 0
 
+    assert str(excinfo.value) == "failed to load schema from url"
+
+
+@responses.activate
+def test_from_url_request_failed():
+    responses.add(responses.GET, TEST_SCHEMA_URL, body=RequestException())
+    with pytest.raises(SchemaRequestFailed) as excinfo:
```
MebinAbraham, line 189: Same as above.

MebinAbraham (commented):
Looks good. Just minor comments.

app/utilities/schema.py
petechd: Done, thanks!

app/utilities/schema_request_failed_exception.py
petechd: Removed separate file.

app/utilities/schema.py
petechd: Done 👍

tests/app/utilities/test_schema.py
petechd: OK, `exc` now thanks!

tests/app/utilities/test_schema.py
petechd: Done, thanks.

MebinAbraham (approved):

tests/app/utilities/test_schema.py
```diff
@@ -155,21 +160,37 @@ def test_load_schema_from_url_200():
     assert cache_info.hits == 0
 
 
+@pytest.mark.parametrize(
+    "status_code",
+    [401, 403, 404, 501, 511],
+)
 @responses.activate
-def test_load_schema_from_url_404():
+def test_from_url_non_200(status_code):
```
berroar, line 168:
```suggestion
def test_load_schema_from_url_non_200(status_code):
```

tests/app/utilities/test_schema.py
```diff
@@ -155,21 +160,37 @@ def test_load_schema_from_url_200():
...
     cache_info = load_schema_from_url.cache_info()
     assert cache_info.currsize == 0
     assert cache_info.misses == 1
     assert cache_info.hits == 0
 
+    assert str(exc.value) == "schema request failed"
+
+
+@responses.activate
+def test_from_url_request_failed():
```
berroar, line 187:
```suggestion
def test_load_schema_from_url_request_failed():
```

tests/app/utilities/test_schema.py
```diff
@@ -206,3 +227,95 @@ def test_load_schema_from_metadata_with_schema_url():
...
+):
+    connect_timeout_error = ConnectTimeoutError("connect timed out")
+    read_timeout_error = ReadTimeoutError(
+        pool=None, message="read timed out", url="test-url"
+    )
+
+    response_not_timed_out = HTTPResponse(status=200, headers={}, msg=HTTPMessage())
+    response_not_timed_out.drain_conn = Mock(return_value=None)
+
+    patched_make_request = mocker.patch.object(
```
berroar, line 273:
```suggestion
        return mocker.patch.object(
        HTTPConnectionPool,
        "_make_request",
        side_effect=[
            connect_timeout_error,
            read_timeout_error,
            response_not_timed_out,
        ],
    )
```

berroar (approved):
Looks good, minor comments

tests/app/utilities/test_schema.py
petechd: Ok, added thanks!

PR #862, opened 2022-05-10
Bug fix: Support for in-flight sessions that use survey_url

MebinAbraham:
Recently `schema_url` was introduced in favour of `survey_url`, however that breaks in-flight sessions for users that currently made use of `survey_url`.

- This PR adds supports so that in-flight session does not error when rolling out the `schema_url` change. 
- Although `survey_url` is now supported for in-flight sessions, the literal value for `survey_url` has been dropped. RAS does not use `survey_url` for any live surveys.
- With this change `SessionData` accepts a `schema_url` or `survey_url` on initialisation. This was simply to allows us to roll back this change in the event this errored in prod. Once `schema_url` has been rolled out successfully, `survey_url` will be removed in a subsequent PR.

### How to review 
1. Check out to v3.107.1 and launch a survey
2. Check out to this branch and try to continue the survey.
3.  Check out back to v3.107.1 and try complete and submit the survey
4. Ensure none of these steps result in an error.

When comparing these steps above between v3.107.1 and v4.0.0, it should result in errors.

Has any scenario been missed which would result in an error?.

### Checklist

### DIFF
--- a/app/data_models/session_data.py
+++ b/app/data_models/session_data.py
@@ -9,7 +9,6 @@ class SessionData:
         period_str: Optional[str],
         language_code: Optional[str],
         launch_language_code: Optional[str],
-        schema_url: Optional[str],
         ru_name: Optional[str],
         ru_ref: Optional[str],
         response_id: Optional[str],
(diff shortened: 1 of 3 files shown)


berroar (approved):
Tested, looks good 👍

petechd (approved):
Works fine, fixes the error.

PR #866, opened 2022-05-16
Add type hinting to secrets

rmccar:
This PR will add type hinting to secrets

Also removed an odd char on the end of a line in test_feedback.py

### How to review 
- Changes make sense
- Run `run_mypy.sh` see that theres no errors

### Checklist

### DIFF
--- a/app/secrets.py
+++ b/app/secrets.py
@@ -1,3 +1,5 @@
+from typing import Mapping, Optional, Any
+
 REQUIRED_SECRETS = [
     "EQ_SERVER_SIDE_STORAGE_USER_ID_SALT",
     "EQ_SERVER_SIDE_STORAGE_USER_IK_SALT",
@@ -7,8 +9,7 @@ REQUIRED_SECRETS = [
     "EQ_RABBITMQ_PASSWORD",
 ]
 
-
-def validate_required_secrets(secrets, additional_required_secrets=None):
+def validate_required_secrets(secrets: Mapping[str, str], additional_required_secrets: Optional[list[str]]=None) -> bool:
     all_required_secrets = (
         REQUIRED_SECRETS + additional_required_secrets
         if additional_required_secrets
@@ -20,8 +21,9 @@ def validate_required_secrets(secrets, additional_required_secrets=None):
 
 
 class SecretStore:
-    def __init__(self, secrets):
+    def __init__(self, secrets: dict) -> None:
+        print(secrets)
         self.secrets = secrets.get("secrets", {})
 
-    def get_secret_by_name(self, secret_name):
+    def get_secret_by_name(self, secret_name: str) -> Any:
         return self.secrets.get(secret_name)
--- a/app/secrets.py
+++ b/app/secrets.py
@@ -1,4 +1,4 @@
-from typing import Mapping, Optional, Any
+from typing import Any, Mapping, Optional
 
 REQUIRED_SECRETS = [
     "EQ_SERVER_SIDE_STORAGE_USER_ID_SALT",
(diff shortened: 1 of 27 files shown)


rmccar:
[Add type hinting for secrets (1 SP)](https://trello.com/c/gixvwj7Z/5306-add-type-hinting-for-secrets-1-sp)

app/secrets.py
```diff
@@ -20,8 +24,8 @@ def validate_required_secrets(secrets, additional_required_secrets=None):
 
 
 class SecretStore:
-    def __init__(self, secrets):
+    def __init__(self, secrets: dict) -> None:
```
berroar, line 27: Can `secrets` be more specific here like the `Mapping[str, str]` above?

app/secrets.py
```diff
@@ -20,8 +24,8 @@ def validate_required_secrets(secrets, additional_required_secrets=None):
...
         self.secrets = secrets.get("secrets", {})
 
-    def get_secret_by_name(self, secret_name):
+    def get_secret_by_name(self, secret_name: str) -> Any:
```
berroar, line 30: Isn't this returning a string?

app/secrets.py
rmccar: I did try this initially but it caused errors because str has no get function on line 28

app/secrets.py
rmccar: Same as above I did try this but it errored saying that it is expecting to return an any so I set it to any

app/secrets.py
```diff
@@ -8,7 +10,9 @@
 ]
 
 
-def validate_required_secrets(secrets, additional_required_secrets=None):
+def validate_required_secrets(
+    secrets: Mapping[str, str], additional_required_secrets: Optional[list[str]] = None
+) -> bool:
```
MebinAbraham, line 15: This function doesn't return anything.

app/secrets.py
```diff
@@ -8,7 +10,9 @@
 ]
 
 
-def validate_required_secrets(secrets, additional_required_secrets=None):
+def validate_required_secrets(
+    secrets: Mapping[str, str], additional_required_secrets: Optional[list[str]] = None
```
MebinAbraham, line 14: 1. `secrets` is a Mapping of Mapping which has string key values.

```js
{
   "secrets": {
 		"some-key": "some-value",
	}
}
```

2. `additional_required_secrets` is not optional from our current usage. Line 16 would throw an error if it was actually `None` as you can `+` a list and None. Also, we should use `Sequence` over `list` as per type hint style guide, abstract instead of concrete for arguments.

app/secrets.py
MebinAbraham: This should be match the type hint from my previous comment above. It's a mapping of mappings instead of mapping of strings.

app/secrets.py
MebinAbraham: Would be fixed once # is addressed.

app/secrets.py
```diff
@@ -7,8 +9,12 @@
     "EQ_RABBITMQ_PASSWORD",
 ]
 
+SecretsType = Mapping[str, Mapping[str, str]]
```
MebinAbraham, line 12: I think type alias should be able `REQUIRED_SECRETS`

MebinAbraham (approved):

app/secrets.py
rmccar: Don't think I quite know what you're getting at

MebinAbraham: I shouldn't have read before posting 🙈 . Should say `I think type alias should be above REQUIRED_SECRETS`

rmccar: Ooooooooh haha I'll sort that 👍

berroar (approved):

PR #867, opened 2022-05-23
Add percentage and unit placeholder transforms

petechd:
This adds the unit and percentage suffix when piping, new transforms based on jinja filters ones were added. It requires a validator definitions changes which are added in separate validator PR. 

### How to review 
Some new unit tests, open to add more scenarios. Functional test fixed.

### Checklist


petechd:
[Placeholder to support piping of unit and percentage answers (3 SP)](https://trello.com/c/OZVLqMQ3/5286-placeholder-to-support-piping-of-unit-and-percentage-answers-3-sp)

app/questionnaire/placeholder_transforms.py
```diff
@@ -132,6 +134,21 @@ def format_number(self, number: Union[int, Decimal, str]) -> str:
...
         return ""
 
+    @staticmethod
+    def format_percentage(value: Union[int, Decimal, str]) -> str:
+        return f"{value}%"
+
+    @staticmethod
+    def format_unit(
+        unit: str, value: Union[int, Decimal, str], length: str = "short"
+    ) -> Any:  # babel units module not typed hence no specific return value
```
MebinAbraham, line 144: We should explicitly type check.

schemas/test/en/test_calculated_summary.json
```diff
@@ -406,12 +406,14 @@
                                                         "placeholder": "total",
                                                         "transforms": [
                                                             {
-                                                                "transform": "format_number",
+                                                                "transform": "format_unit",
                                                                 "arguments": {
```
MebinAbraham, line 410: We should update one of the placeholder schemas with an example of each.

app/questionnaire/placeholder_transforms.py
```diff
@@ -132,6 +134,21 @@ def format_number(self, number: Union[int, Decimal, str]) -> str:
...
+        return units.format_unit(
+            value=value,
+            measurement_unit=unit,
+            length=length,
+            locale=flask_babel.get_locale(),
```
MebinAbraham, line 149: `PlaceholderTransforms` already has a `locale` attribute, so this call to `get_locale` should be avoided.

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -38,6 +38,34 @@ def test_format_number(number, expected, transformer):
...
+def test_format_percentage(value, expected, transformer):
+    assert transformer().format_percentage(value) == expected
+
+
+@pytest.mark.parametrize(
+    "unit, value, length, expected",
+    (
+        ("centimeter", "123", "short", "123 cm"),
+        ("kilometer", "123", "long", "123 kilometer"),
+        ("mile", "123", "short", "123 mi"),
```
MebinAbraham, line 61: Can we do an example for `narrow` length.

app/questionnaire/placeholder_transforms.py
petechd: Added explicit one thanks!

app/questionnaire/placeholder_transforms.py
petechd: Using `PlaceholderTransform`s own locale now, thanks!

schemas/test/en/test_calculated_summary.json
petechd: One of the schemas changed now to include all of these, tested in functional test, thanks!

app/questionnaire/placeholder_transforms.py
```diff
@@ -132,6 +133,23 @@ def format_number(self, number: Union[int, Decimal, str]) -> str:
...
+    @staticmethod
+    def format_percentage(value: Union[int, Decimal, str]) -> str:
+        return f"{value}%"
+
+    def format_unit(
+        self,
+        unit: str,
+        value: Union[int, Decimal, str],
+        unit_length: Optional[str] = "short",
+    ) -> units.format_unit:
```
MebinAbraham, line 145: This should be `str`

schemas/test/en/test_placeholder_default_value.json
```diff
@@ -158,6 +158,121 @@
                                 ]
                             },
                             "type": "Interstitial"
+                        },
+                        {
+                            "type": "Question",
```
MebinAbraham, line 163: I think it would be better suited to add test to `test_placeholder_transform`. This schema is just for the purposes of default values.

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -38,6 +38,34 @@ def test_format_number(number, expected, transformer):
     assert transformer().format_number(number) == expected
 
 
+@pytest.mark.parametrize(
+    "value, expected",
+    (
+        (123, "123%"),
```
MebinAbraham, line 44: A decimal object should be tested.

schemas/test/en/test_placeholder_default_value.json
petechd: Ok moved it to that schema, thanks.

tests/app/questionnaire/test_placeholder_transforms.py
petechd: We already are testing decimal here, I added `("123.4", "123.4%"),("123.40", "123.40%")`, I added `(0.00, "0%")` as well.

app/questionnaire/placeholder_transforms.py
MebinAbraham: No, but that is still correct. I mean the return type.

tests/app/questionnaire/test_placeholder_transforms.py
MebinAbraham: I mean, a `Decimal` object, `"123.4"` is just a string. We shouldn't get floats like `0.00` we should get `Decimal` object representing those.

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -38,6 +38,35 @@ def test_format_number(number, expected, transformer):
...
+    assert transformer().format_percentage(value) == expected
+
+
+@pytest.mark.parametrize(
+    "unit, value, unit_length, expected",
+    (
+        ("centimeter", "123", "short", "123 cm"),
+        ("kilometer", "123", "long", "123 kilometre"),
+        ("mile", "123", "short", "123 mi"),
+        ("mile", "123", "narrow", "123mi"),
```
MebinAbraham, line 63: Worth doing one without a unit length, to test the default.

petechd: Done

tests/app/questionnaire/test_placeholder_transforms.py
petechd: Done thanks!

petechd:
Reminder for myself to revert branch!

rmccar (approved):

tests/app/questionnaire/test_placeholder_transforms.py
MebinAbraham: Still outstanding. Passing `None` in wont use our default. It uses babel's default `long`.
To use our default the logic need to to like `unit_length = unit_length or "short"`

petechd:
Ok `Decimal` test for format unit added, that other test is using our default logic now, I changed it so we don't pass `unit_length` to our method in assertion when `unit_length` is `None`.

MebinAbraham:
> Ok, I was not able to reply to your last two comments so have to post it here. `Decimal` test for format unit added, that other test is using our default logic now, I changed it so we don't pass `unit_length` to our method in assertion when `unit_length` is `None`. I was not sure what logic should I choose, here's another way of doing it which I didn't use, don't know which one you prefer or maybe there is another way? 

I would keep the test as it was and update the logic of `format_unit` to explicitly default to `short`.
```py
    def format_unit(
        self,
        unit: str,
        value: Union[int, Decimal, str],
        unit_length: Optional[str] = None,
    ) -> str:
        length = unit_length or "short"
            value=value,
            measurement_unit=unit,
            length=length,
            locale=self.locale,
        )
        return formatted_unit
```

tests/app/questionnaire/test_placeholder_transforms.py
petechd: Ok, we have a Decimal test in test format unit now and here.

MebinAbraham (approved):

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -38,6 +38,38 @@ def test_format_number(number, expected, transformer):
...
+@pytest.mark.parametrize(
+    "value, expected",
+    (
+        (123, "123%"),
+        ("123.4", "123.4%"),
+        ("123.40", "123.40%"),
+        ("1000", "1000%"),
+        (0, "0%"),
+        (0.00, "0.0%"),
+        ("", "%"),
```
MebinAbraham, line 50: When would we get an empty answer?. We should remove this as we don't store empty value.

petechd: Removed the empty answer, thanks!

PR #868, opened 2022-05-23
Add calculated summary section dependencies

berroar:

Adds support so that `section enabled`, `routing_rules` and `skip_conditions` dependencies can be based on a `calculated_summary` total from another section.

### How to review 

Use the  new `test_new_routing_section_dependencies_calculated_summary.json` schema to test this change.

Follow the tests outlined in the newly added functional test. Has anything been missed?

### Checklist

### DIFF
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -843,14 +843,19 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
             if not isinstance(rule, Mapping):
                 continue
 
-            answer_id = None
+            answer_id_list: list = []
+            identifier: str = rule.get("identifier", "")
 
             if "id" in rule:
-                answer_id = rule["id"]
+                answer_id_list.append(rule["id"])
             elif rule.get("source") == "answers":
-                answer_id = rule.get("identifier")
+                answer_id_list.append(identifier)
+            elif rule.get("source") == "calculated_summary":
+                calculated_summary_block = self.get_block(identifier)
+                calculated_summary_answer_ids = calculated_summary_block["calculation"]["answers_to_calculate"]  # type: ignore
+                answer_id_list.extend(iter(calculated_summary_answer_ids))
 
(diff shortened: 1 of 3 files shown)


berroar:
[Section dependencies for when rules: Calculated summary (3 SP)](https://trello.com/c/ZAFYnQUV/5272-section-dependencies-for-when-rules-calculated-summary-3-sp)

app/questionnaire/questionnaire_schema.py
```diff
@@ -852,14 +852,19 @@ def _get_rules_section_dependencies(
             if not isinstance(rule, Mapping):
                 continue
 
-            answer_id = None
+            answer_id_list: list = []
+            identifier: str = rule.get("identifier", "")
```
petechd, line 856: Maybe we shouldn't use a fallback option if the specified key doesn't exist at this point (checked by validator)?
```suggestion
            identifier: str = rule.get("identifier")
```
otherwise we should probably check if identifier exists in the elif statements.

schemas/test/en/test_new_routing_section_dependencies_calculated_summary.json
```diff
@@ -0,0 +1,279 @@
...
+            "id": "calculated-summary-section",
+            "groups": [
+                {
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "first-question-block",
+                            "question": {
+                                "id": "first-question",
+                                "title": "How much do you spend on the following items?",
```
petechd, line 41: Maybe some explanation like:
```suggestion
                                "title": "How much do you spend on the following items?",
                                "description": [
                                    "<p>If the total is equal or over 100£ a new section will get enabled on the routing path</p>"
                                ],
```
The wording is not ideal but you will get the idea.

berroar: Added description 👍

rmccar (approved):

petechd (approved):

PR #873, opened 2022-05-26
Expose tx_id in dataLayer

petechd:
This adds the tx_id into the dataLayer for all survey types that currently support data layer. In an attempt to ensure we match what is in v2, we need to expose it.
Doesn't matter if it's part of the existing mapping as long as Ian is able to extract it.

### How to review 
Existing tests are refactored so up to date dataLayer is being compared. No need for new tests.

### Checklist


petechd:
[Expose tx_id in runner's dataLayer (2 SP)](https://trello.com/c/iPHXDYBS/5308-expose-txid-in-runners-datalayer-2-sp)

tests/integration/test_application_variables.py
```diff
@@ -23,8 +23,8 @@ def test_google_analytics_code_and_credentials_are_present(self):
         self.get("/questionnaire/feedback/")
         self.assertStatusOK()
         self.assertInHead("gtm.start")
-        self.assertInHead(
-            'dataLayer = [{"form_type": "H", "survey_id": "0", "title": "Feedback test schema"}]'
+        self.assertRegexPage(
```
berroar, line 26: Perhaps it would be clearer if loaded the actual tx_id from the response data json then did something like this here? 
```suggestion
        self.assertInHead(
            f'dataLayer = [{{"tx_id": "{actual["METADATA"]["tx_id"]}"}}, {{"form_type": "H", "survey_id": "0", "title": "Feedback test schema"}}]'
        )
```

petechd: Added to all three, thanks!

tests/integration/test_application_variables.py
```diff
@@ -16,22 +17,26 @@ def tearDown(self):
...
+        actual = json_loads(self.getResponseData())
         self._client.set_cookie(
             "localhost", key="ons_cookie_policy", value="'usage':true"
         )
         self.get("/questionnaire/feedback/")
         self.assertStatusOK()
         self.assertInHead("gtm.start")
         self.assertInHead(
-            'dataLayer = [{"form_type": "H", "survey_id": "0", "title": "Feedback test schema"}]'
+            f'dataLayer = [{{"tx_id": "{actual["METADATA"]["tx_id"]}"}}, {{"form_type": "H", "survey_id": "0", "title": "Feedback test schema"}}]'
```
MebinAbraham, line 30: I thought we said it should be part of the already existing mapping?. Happy to keep as is, if it makes no difference to Ian and might be better as one is schema data.

app/survey_config/survey_config.py
```diff
@@ -40,13 +42,18 @@ class SurveyConfig:
...
+    )
     sign_out_button_text: str = lazy_gettext("Save and exit survey")
     contact_us_url: str = field(init=False)
     cookie_settings_url: str = field(init=False)
     privacy_and_data_protection_url: str = field(init=False)
 
     def __post_init__(self):
+        if current_user and get_metadata(current_user):
+            if tx_id := get_metadata(current_user).get("tx_id"):
+                self.data_layer.append({"tx_id": tx_id})
```
MebinAbraham, line 56: App context should be avoided in these classes. Data layer will have to be done similarly to `get_service_links`. We should pass the tx_id in to avoid needed the app context. Same concept as the service links.

MebinAbraham (changes requested):

MebinAbraham (commented):
The data layer unit tests needs updating. They currently do not even pass the schema to test the existing keys.

tests/integration/test_application_variables.py
petechd: Ian said no difference and he is happy if it's this way, I prefer this approach as it's less "invasive" and we don't reassign existing var of the parent class,  just extend it.

petechd:
> The data layer unit tests needs updating. They currently do not even pass the schema to test the existing keys.

Bunch of new unit tests added and old refactored so schemas are taken into account every time now.

app/survey_config/survey_config.py
petechd: Done this as you suggested, a lot of refactoring involved. Thanks!

app/helpers/template_helpers.py
```diff
@@ -94,6 +94,18 @@ def service_links_context(
...
+    def data_layer_context(
+        self,
+    ) -> Union[list[dict], None, list]:
+        if metadata := get_metadata(current_user):
+            if tx_id := metadata.get("tx_id"):
+                return self._survey_config.get_data_layer(
+                    tx_id=tx_id,
+                )
+
+        return self._survey_config.get_data_layer()
```
MebinAbraham, line 107: I think call the function once and pass in a tx_id which may or may not be None. `tx_id` is also mandatory if we have metadata.

app/survey_config/business_config.py
```diff
@@ -102,6 +93,20 @@ def get_service_links(
 
         return links
 
+    def get_data_layer(  # pylint: disable=unused-argument, no-self-use
```
MebinAbraham, line 96: These disable statements are redundant?

app/survey_config/business_config.py
```diff
@@ -102,6 +93,20 @@ def get_service_links(
...
+        self, tx_id=None
```
MebinAbraham, line 97: Needs type annotation.

app/survey_config/census_config.py
```diff
@@ -57,12 +57,17 @@ class CensusSurveyConfig(
...
         compare=False,
     )
-    data_layer: Iterable[Mapping] = field(
-        default_factory=lambda: [{"nisra": False}], compare=False
-    )
+
     survey_title: LazyString = lazy_gettext("Census 2021")
     sign_out_button_text: str = lazy_gettext("Save and complete later")
 
+    def get_data_layer(  # pylint: disable=unused-argument, no-self-use
```
MebinAbraham, line 64: Could use `_` for unused arg.

app/survey_config/census_config.py
```diff
@@ -57,12 +57,17 @@ class CensusSurveyConfig(
...
+        self, tx_id=None
+    ) -> list[dict]:
+        if tx_id:
+            return [{"nisra": False}, {"tx_id": tx_id}]
+        return [{"nisra": False}]
```
MebinAbraham, line 69: If you assign an attribute call `self._is_nisra` and set the value in each class, then this methods doesn't need to be defined in each class.

app/helpers/template_helpers.py
petechd: Changed it to 
```python
tx_id = get_metadata(current_user).get("tx_id") if get_metadata(current_user) else None
```

app/survey_config/census_config.py
petechd: I've shortened it to list comprehension and only single call, cheers!

app/survey_config/census_config.py
petechd: Ok, unused argument error gone now but still needed `no-self-use` ignore in one place

app/survey_config/business_config.py
petechd: Added, thanks!

app/survey_config/business_config.py
petechd: Removed, thanks.

app/helpers/template_helpers.py
MebinAbraham: This would then call `get_metadata` twice, which should be avoided.

app/survey_config/census_config.py
```diff
@@ -57,11 +57,17 @@ class CensusSurveyConfig(
...
     survey_title: LazyString = lazy_gettext("Census 2021")
     sign_out_button_text: str = lazy_gettext("Save and complete later")
+    _is_nisra: bool = False
+
+    def get_data_layer(self, _tx_id: Optional[str] = None) -> list[dict]:
+        return (
+            [{"nisra": self._is_nisra}, {"tx_id": _tx_id}]
+            if _tx_id
+            else [{"nisra": self._is_nisra}]
+        )
```
MebinAbraham, line 70: - How come `tx_id` is prefxied with `_`?.
- I think having one data var would be better here. 
```py
        data_layer = [{"nisra": self._is_nisra}]
        if _tx_id:
            data_layer.append({"tx_id": _tx_id})

        return data_layer
```

petechd: Prefixed with `_` to avoid `W0237: Parameter '_tx_id' has been renamed to 'tx_id' in overridden 'CensusSurveyConfig.get_data_layer' method (arguments-renamed)`. Will use your suggestion code, thanks!

app/helpers/template_helpers.py
petechd: Ok removed duplication now.

app/survey_config/census_config.py
petechd: After refactor it seems that the var is always used so removing `_`, cheers.

app/survey_config/survey_config.py
```diff
@@ -61,3 +63,11 @@ def get_service_links(  # pylint: disable=unused-argument, no-self-use
         ru_ref: Optional[str],
     ) -> Optional[list[dict]]:
         return None
+
+    def get_data_layer(  # pylint: disable=no-self-use
+        self, tx_id: Optional[str] = None
+    ) -> Optional[list[dict]]:
```
MebinAbraham, line 69: This should always be returning a list unless you are explicitly checking for None, which it doesn't seem to do at the moment.

app/helpers/template_helpers.py
```diff
@@ -94,6 +94,14 @@ def service_links_context(
 
         return None
 
+    @property
+    def data_layer_context(
+        self,
+    ) -> Union[list[dict], None, list]:
```
MebinAbraham, line 100: We should always return a list that may contain dicts since we just pass this directly to the view.

app/survey_config/survey_config.py
petechd: Fixed this type hint now thanks!

app/helpers/template_helpers.py
petechd: As above, fixed now, thanks!

app/helpers/template_helpers.py
```diff
@@ -94,6 +94,14 @@ def service_links_context(
 
         return None
 
+    @property
+    def data_layer_context(
+        self,
+    ) -> Union[list[dict], list]:
```
MebinAbraham, line 100: Returning `list` and `list[dict]` makes `list[dict]` redundant as `list[Any]` would supersede. It should just be `list[dict]`.

petechd: Ok added.

MebinAbraham (approved):

berroar (approved):

PR #876, opened 2022-05-30
Add type hinting for jinja filters

rmccar:
This PR will add type hinting to jinja filters

### How to review 
- Changes make sense
- Run `run_mypy.sh` see that theres no errors

### Checklist


rmccar:
[Add type hinting for jinja filters (2 SP)](https://trello.com/c/mMcfE5TO/5309-add-type-hinting-for-jinja-filters-2-sp)

app/jinja_filters.py
```diff
@@ -1,41 +1,48 @@
...
 
 
-def get_formatted_address(address_fields):
+def get_formatted_address(address_fields: dict) -> str:
     address_fields.pop("uprn", None)
     return "<br>".join(address_field for address_field in address_fields.values())
 
 
-def get_formatted_currency(value, currency="GBP") -> str:
+def get_formatted_currency(value: Union[float, Decimal], currency: str = "GBP") -> Any:
```
berroar, line 45: The return here should be str, might need explicitly set the var as a str for this to work though.

app/jinja_filters.py
```diff
@@ -134,7 +149,7 @@ def get_format_date(value):
 
 @pass_eval_context  # type: ignore
 @blueprint.app_template_filter()
-def format_datetime(context, date_time):
+def format_datetime(context: str, date_time: str) -> str:
```
berroar, line 152: I think date_time is type `datetime` and not `str` here

app/jinja_filters.py
```diff
@@ -148,7 +163,7 @@ def format_datetime(context, date_time):
     return mark_safe(context, result)
 
 
-def get_format_date_range(start_date, end_date):
+def get_format_date_range(start_date: str, end_date: str) -> flask_babel:
```
berroar, line 166: These appear to be of type `Markup`

app/jinja_filters.py
```diff
@@ -157,35 +172,35 @@ def get_format_date_range(start_date, end_date):
...
-def setAttributes(dictionary, attributes):
+def setAttributes(dictionary: dict, attributes: dict) -> dict:
     for key in attributes:
         dictionary[key] = attributes[key]
     return dictionary
 
 
 @blueprint.app_template_filter()
-def should_wrap_with_fieldset(question):
+def should_wrap_with_fieldset(question: dict) -> Union[Any, bool]:
```
berroar, line 203: I think this is just returning `bool`

berroar (changes requested):

app/jinja_filters.py
```diff
@@ -1,41 +1,48 @@
...
 blueprint = flask.Blueprint("filters", __name__)
 
 
-def mark_safe(context, value):
+def mark_safe(context: Any, value: str) -> str:
     return Markup(value) if context.autoescape else value
 
 
-def strip_tags(value):
+def strip_tags(value: str) -> str:
```
petechd, line 24: This returns escaped Markup if you check escape() method return type what type does it return?

app/forms/validators.py
```diff
@@ -445,7 +454,7 @@ def _is_valid(
 
 def format_playback_value(
     value: Union[float, Decimal], currency: Optional[str] = None
-) -> str:
+) -> Any:
```
berroar, line 457: This should return a str, as we are enforcing the var being returned as a string.

app/jinja_filters.py
```diff
@@ -1,41 +1,48 @@
...
     return escape(Markup(value).striptags())
 
 
 @blueprint.app_template_filter()
-def format_number(value):
+def format_number(value: int) -> str:
```
petechd, line 29: This can accept both integers and floats as it uses `format_decimal()` further down which accepts floats as arguments.

app/jinja_filters.py
```diff
@@ -45,48 +52,56 @@ def get_formatted_currency(value, currency="GBP") -> str:
...
+def get_currency_symbol(currency: str = "GBP") -> str:
+    currency_symbol: str = numbers.get_currency_symbol(
+        currency, locale=flask_babel.get_locale()
+    )
+    return currency_symbol
 
 
 @blueprint.app_template_filter()
-def format_percentage(value):
+def format_percentage(value: str) -> str:
```
petechd, line 63: The value here can be an `int` or `Decimal`, then it is converted to `str`

app/jinja_filters.py
```diff
@@ -45,48 +52,56 @@ def get_formatted_currency(value, currency="GBP") -> str:
...
     return f"{value}%"
 
 
-def format_unit(unit, value, length="short"):
-    return units.format_unit(
+def format_unit(unit: str, value: str, length: str = "short") -> str:
```
petechd, line 67: `value` here can be `Decimal` or `int`.

app/jinja_filters.py
```diff
@@ -45,48 +52,56 @@ def get_formatted_currency(value, currency="GBP") -> str:
...
+            measurement_unit=unit,
+            length=unit_length,
+            locale=flask_babel.get_locale(),
+        ).strip()
+
+    return unit_label
 
 
-def format_duration(value):
+def format_duration(value: Mapping) -> str:
```
petechd, line 104: Needs at least the top-level type parameters.

petechd (changes requested):

app/forms/validators.py
rmccar: This needed to be any because the return on 459 returns `get_formatted_currency` which needs to return Any because it uses a function from babel that returns Any

app/jinja_filters.py
```diff
@@ -236,19 +252,25 @@ def get_width_for_number(answer):
...
         self.description = description
 
 
 class SelectConfig:
-    def __init__(self, option, index, answer, form=None):
+    def __init__(
+        self,
+        option: SelectFieldBase._Option,
+        index: int,
+        answer: Mapping,
```
petechd, line 271: These  mappings need top level parameters.

app/jinja_filters.py
```diff
@@ -295,7 +319,11 @@ def __init__(self, option, index, answer):
 
 
 class OtherConfig:
-    def __init__(self, detail_answer_field, detail_answer_schema):
+    def __init__(
+        self,
+        detail_answer_field: SelectFieldBase._Option,
+        detail_answer_schema: Mapping,
```
petechd, line 325: Mapping needs top level parameters (as described in type hinting doc in runner).

app/jinja_filters.py
```diff
@@ -320,7 +348,7 @@ def __init__(self, detail_answer_field, detail_answer_schema):
 
 
 @blueprint.app_template_filter()  # type: ignore
-def map_select_config(form, answer):
+def map_select_config(form: Mapping, answer: Mapping) -> list[SelectConfig]:
```
petechd, line 351: Same here, mappings need top level parameters.

app/jinja_filters.py
```diff
@@ -330,12 +358,14 @@ def map_select_config(form, answer):
...
 @blueprint.app_context_processor
-def map_select_config_processor():
+def map_select_config_processor() -> dict:
     return dict(map_select_config=map_select_config)
 
 
 @blueprint.app_template_filter()  # type: ignore
-def map_relationships_config(form, answer):
+def map_relationships_config(
+    form: Mapping, answer: Mapping
```
petechd, line 367: Mappings need top level parameters.

petechd (changes requested):

app/jinja_filters.py
```diff
@@ -534,27 +573,25 @@ def map_summary_item_config(
...
-    list_items,
-    icon,
-    edit_link_text=None,
-    edit_link_aria_label=None,
-    remove_link_text=None,
-    remove_link_aria_label=None,
-):
+    list_items: list,
+    icon: str,
+    edit_link_text: Optional[str] = "",
```
petechd, line 590: These two `Optional[str]` have default value of `""` so they shouldn't be optional. Also `list_items: list` needs top level parameter.

app/jinja_filters.py
```diff
@@ -534,27 +573,25 @@ def map_summary_item_config(
...
+    edit_link_aria_label: str = "",
+    remove_link_text: Optional[str] = "",
+    remove_link_aria_label: str = "",
+) -> list:
```
petechd, line 594: Returned `list` needs type of the top level parameter as well.

app/jinja_filters.py
```diff
@@ -603,5 +640,5 @@ def map_list_collector_config(
 
 
 @blueprint.app_context_processor
-def map_list_collector_config_processor():
+def map_list_collector_config_processor() -> dict:
```
petechd, line 643: Again, we need to know what this generic `dict` returns.

app/jinja_filters.py
```diff
@@ -534,27 +573,25 @@ def map_summary_item_config(
...
-            SummaryRow(calculated_question, summary_type, False, None, None, None)
-        )
+        rows.append(SummaryRow(calculated_question, summary_type, False, "", "", ""))
 
     return rows
 
 
 @blueprint.app_context_processor
-def map_summary_item_config_processor():
+def map_summary_item_config_processor() -> dict:
```
petechd, line 582: We need to know what generic type `dict` returns.

app/jinja_filters.py
petechd: This one is still outstanding.

app/jinja_filters.py
petechd: This one is still outstanding, it returns `Markup`.
Screenshot 2022-06-20 at 15 13 05

app/jinja_filters.py
```diff
@@ -1,92 +1,110 @@
...
 
 @blueprint.app_template_filter()
-def format_percentage(value):
+def format_percentage(value: Union[int, Decimal]) -> str:
     return f"{value}%"
 
 
-def format_unit(unit, value, length="short"):
-    return units.format_unit(
+def format_unit(unit: Union[int, Decimal], value: str, length: str = "short") -> str:
```
petechd, line 70: Here `value` can be integer but potentially also `Decimal`, not a string.

app/jinja_filters.py
```diff
@@ -104,14 +122,14 @@ def format_duration(value):
...
-def get_format_multilined_string(value):
+def get_format_multilined_string(value: str) -> str:
     escaped_value = escape(value)
     new_line_regex = r"(?:\r\n|\r|\n)+"
     value_with_line_break_tag = re.sub(new_line_regex, "<br>", escaped_value)
     return f"{value_with_line_break_tag}"
 
 
-def get_format_date(value):
+def get_format_date(value: str) -> str:
```
petechd, line 134: `value` is `Markup` here.

app/jinja_filters.py
```diff
@@ -134,7 +152,7 @@ def get_format_date(value):
 
 @pass_eval_context  # type: ignore
 @blueprint.app_template_filter()
-def format_datetime(context, date_time):
+def format_datetime(context: nodes.EvalContext, date_time: str) -> str:
```
petechd, line 157: Worth double checking but I think `date_time` parameter needs to be `datetime` type.

app/jinja_filters.py
```diff
@@ -134,7 +155,7 @@ def get_format_date(value):
 
 @pass_eval_context  # type: ignore
 @blueprint.app_template_filter()
-def format_datetime(context, date_time):
+def format_datetime(context: nodes.EvalContext, date_time: datetime) -> str:
```
petechd, line 158: This should have the same return value as the `markup_safe` method because that is what it returns.

app/jinja_filters.py
```diff
@@ -148,7 +169,7 @@ def format_datetime(context, date_time):
     return mark_safe(context, result)
 
 
-def get_format_date_range(start_date, end_date):
+def get_format_date_range(start_date: Markup, end_date: Markup) -> flask_babel:
```
petechd, line 172: It can't just have the whole `flask_babel` library as the return type.

app/jinja_filters.py
```diff
@@ -212,12 +237,12 @@ def should_wrap_with_fieldset(question):
...
 
 @blueprint.app_context_processor
-def should_wrap_with_fieldset_processor():
+def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
     return {"should_wrap_with_fieldset": should_wrap_with_fieldset}
 
 
 @blueprint.app_template_filter()
-def get_width_for_number(answer):
+def get_width_for_number(answer: AnswerType) -> int:
```
petechd, line 245: This method not always has a return (if width is > allowable_width) so needs to be `Optional`.

app/jinja_filters.py
petechd: Does this one not return `gettext` so `flask_babel.gettext`?

rmccar: I checked what was returned where this function was being called and it was some markup so I made this function return markup

petechd: In my check it returns a string so should be a string or Union of string and Markup?
Screenshot 2022-06-21 at 14 47 04

schemas/test/en/test_date_validation_range.json
```diff
@@ -47,7 +47,7 @@
                                 "guidance": {
                                     "contents": [
                                         {
-                                            "list": ["Enter a date range between 23 days and 1 month and 20 days apart"]
+                                            "description": "Enter a date range between 23 days and 1 month and 20 days apart"
```
petechd, line 50: Unnecessary change committed?

rmccar: No I noticed that in one of the test schemas the guidance was wrongly formatted so fixed it

app/jinja_filters.py
```diff
@@ -534,27 +589,25 @@ def map_summary_item_config(
...
-    list_items,
-    icon,
-    edit_link_text=None,
-    edit_link_aria_label=None,
-    remove_link_text=None,
-    remove_link_aria_label=None,
-):
+    list_items: list[dict[str, Union[str, int]]],
+    icon: str,
+    edit_link_text: str = "",
```
berroar, line 606: I think theses should probably be `Optional[str] = None` but will need to find a way to the `.format` method further down.

rmccar: Yeah thats why I didn't make them `Optional[str] = None`.
Ive come up with a solution though and pushed it up

petechd (approved):

berroar (approved):

PR #877, opened 2022-05-31
Add submitted date to pdf filename

rmccar:
Adds the submitted date to the end of the pdf download file

### How to review 
- Use `test_view_submitted_response` schema to download pdf answers
- See that the submitted date is added to the end of the filename

### Checklist


tests/app/views/handlers/test_view_submitted_response_pdf.py
```diff
@@ -33,4 +33,4 @@ def test_filename_uses_schema_name(storage, schema, language):
     questionnaire_store.set_metadata({"schema_name": "test_view_submitted_response"})
     pdf = ViewSubmittedResponsePDF(schema, questionnaire_store, language)
 
-    assert pdf.filename == "test_view_submitted_response.pdf"
+    assert pdf.filename == f"test_view_submitted_response-{str(date.today())}.pdf"
```
berroar, line 36: Just checking the trello card, and I think we are supposed to be using the `schema title` rather than the `schema name` for the first part?

app/views/handlers/view_submitted_response_pdf.py
```diff
@@ -54,7 +54,8 @@ def __init__(
     @property
     def filename(self) -> str:
         """The name to use for the PDF file"""
-        return f"{self._metadata['schema_name']}.pdf"
+        if self._questionnaire_store.submitted_at is not None:
+            return f"{self._metadata['schema_name']}-{self._questionnaire_store.submitted_at.date()}.pdf"
```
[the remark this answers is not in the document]
petechd: In here it should be probably something like 
```python
return f"{self._schema.json['title']}-{self._questionnaire_store.submitted_at.date()}.pdf
```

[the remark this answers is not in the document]
petechd: Spaces replaced by hyphens so something like:
```python
f"{self._schema.json['title']}-{self._questionnaire_store.submitted_at.date().strftime('%Y-%m-%d')}.pdf".replace(" ", "-")
```

app/views/handlers/view_submitted_response_pdf.py
```diff
@@ -54,7 +54,8 @@ def __init__(
     @property
     def filename(self) -> str:
         """The name to use for the PDF file"""
-        return f"{self._metadata['schema_name']}.pdf"
+        if self._questionnaire_store.submitted_at is not None:
```
petechd, line 57:
```suggestion
        if self._questionnaire_store.submitted_at:
```
but can it not have `submitted_at`?

app/views/handlers/view_submitted_response_pdf.py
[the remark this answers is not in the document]
rmccar: 👍 added all this except `strftime` wasn't needed `date` already outputs the correct format

tests/app/views/handlers/test_view_submitted_response_pdf.py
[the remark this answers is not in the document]
rmccar: I just removed `str` in the end because its in the correct format without it

tests/app/views/handlers/test_view_submitted_response_pdf.py
```diff
@@ -33,4 +33,4 @@ def test_filename_uses_schema_name(storage, schema, language):
     questionnaire_store.set_metadata({"schema_name": "test_view_submitted_response"})
     pdf = ViewSubmittedResponsePDF(schema, questionnaire_store, language)
 
-    assert pdf.filename == "test_view_submitted_response.pdf"
+    assert pdf.filename == f"test-view-submitted-response-{date.today()}.pdf"
```
berroar, line 36: We should use `submitted_at.date()` rather than `date.today()` for this

petechd:
Regarding the breaking test you need to add title to schema so:
```python
@pytest.fixture
def schema():
    return QuestionnaireSchema({"post_submission": {"view_response": True}, "title": "test"})
``` 
in `tests/app/views/handlers/conftest.py`, then the test assertion needs changing.

app/views/handlers/view_submitted_response_pdf.py
```diff
@@ -54,7 +54,10 @@ def __init__(
     @property
     def filename(self) -> str:
         """The name to use for the PDF file"""
-        return f"{self._metadata['schema_name']}.pdf"
+        if self._questionnaire_store.submitted_at:
+            return f"{self._schema.json['title']}-{self._questionnaire_store.submitted_at.date()}.pdf".replace(
+                " ", "-"
+            ).lower()
```
petechd, line 60: I think we want to add another replace, so chain them and avoid underscores in file name (if I understand the card correctly), so:
```suggestion
            ).replace("_", "-").lower()
```

app/views/handlers/view_submitted_response_pdf.py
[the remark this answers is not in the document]
petechd: I think we should use `strftime()` as it converts date explicitly, `date()` doesn't outputs a string but a date object, implicit type conversion is something we would like to avoid.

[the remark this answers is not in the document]
rmccar: Yeah but `.replace()` and `.lower()` will both return a string anyway. So adding `strftime()` wont really make a difference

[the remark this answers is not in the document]
petechd: As discussed, `.replace()` and `.lower()` are methods of string type so by the time you execute both of them the var is already converted implicitly to string.

tests/app/views/handlers/test_view_submitted_response_pdf.py
```diff
@@ -33,4 +33,4 @@ def test_filename_uses_schema_name(storage, schema, language):
     questionnaire_store.set_metadata({"schema_name": "test_view_submitted_response"})
     pdf = ViewSubmittedResponsePDF(schema, questionnaire_store, language)
 
-    assert pdf.filename == "test_view_submitted_response.pdf"
+    assert pdf.filename == f"test-view-submitted-response-{submitted_at.date()}.pdf"
```
MebinAbraham, line 36: We should check the string format explicitly so that we can verify the format is as expected.

app/views/handlers/view_submitted_response_pdf.py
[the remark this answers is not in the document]
MebinAbraham: I think either is fine but either way our test should ensure that we get the format that it expects, so that if the underlying `__str__` format of date changed, our test should break. (The format shouldn't change in the stdlib). But then again, explicit is better than implicit. If it were to change, I think we should change it to `date.isoformat()` instead of `strftime()` as ISO 8601 is a spec, so the format is explicit.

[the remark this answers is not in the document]
MebinAbraham: 1. I think it would be better for the call to `lower()` and `replace()` to be done just on the title instead of the entire string. 
2. Most schemas also use ` - ` separator in the title, which would result a title `MBS - Construction and Allied Trades` to be transformed into `mbs---construction-and-allied-trades`, therefore `--` needs to be replaced with `-` so that we end up with `mbs-construction-and-allied-trades`

app/views/handlers/view_submitted_response_pdf.py
```diff
@@ -54,7 +54,7 @@ def __init__(
     @property
     def filename(self) -> str:
         """The name to use for the PDF file"""
-        return f"{self._metadata['schema_name']}.pdf"
+        return f"{self._schema.json['title'].lower().replace(' ', '-').replace('--', '-')}-{self._questionnaire_store.submitted_at.date().isoformat()}.pdf"
```
MebinAbraham, line 57: A lot going on in 1 line. Would be more readable if it was split out.
```py
        formatted_title = (
            self._schema.json["title"].lower().replace(" ", "-").replace("--", "-")
        )
        formatted_date = self._questionnaire_store.submitted_at.date().isoformat()
        return f"{formatted_title}-{formatted_date}.pdf"
```

app/views/handlers/view_submitted_response_pdf.py
```diff
@@ -54,7 +54,12 @@ def __init__(
     @property
     def filename(self) -> str:
```
petechd, line 55:
```suggestion
    def filename(self) -> Optional[str]:
```

app/views/handlers/view_submitted_response_pdf.py
```diff
@@ -54,7 +54,11 @@ def __init__(
...
         """The name to use for the PDF file"""
-        return f"{self._metadata['schema_name']}.pdf"
+        formatted_title = (
+            self._schema.json["title"].lower().replace(" ", "-").replace("--", "-")
+        )
```
MebinAbraham, line 59: Currently this will still result in `--` in the title. Thinking about it, worth just replacing all alphanumeric chars would something like:
```py
formatted_title = re.sub('[^0-9a-zA-Z]+', '-', self._schema.json["title"].lower())
```
This fixes the issue above + handles other non alphanum chars. Used regex on purpose over built-in `isalnum` as it won't result in same behaviour. We want to replace non alpha num instead of not outputting it.

rmccar: Ok yeah good idea Ive added it in

MebinAbraham (approved):

petechd (approved):

tests/app/views/handlers/test_view_submitted_response_pdf.py
MebinAbraham: Superseded.

PR #882, opened 2022-06-20
Add value source for calculations

petechd:
Introduced value source to calculations so we no longer need to use hardcoded integer values. Two sources are supported: answers and calculated summary. Metadata source was dropped as it doesn't support integer data types atm. New unit tests added to improve coverage and functional tests to check updating answer dependencies. Backward compatible with older value for calculations. Another PR created in validator for new rules definitions. 

### How to review 
New functional tests and unit tests. New schema added that uses both answer and calculated summary source.

### Checklist


petechd:
[Validating a Calculated question against a calculated summary (3 SP)](https://trello.com/c/sz7f8zbQ/5129-validating-a-calculated-question-against-a-calculated-summary-3-sp)

berroar (commented):
Should we add metadata as a source for the schema/tests as I think the card suggests that this should be covered too?

Edit: Nevermind just re-read the PR description 😅

schemas/test/en/test_validation_sum_against_value_source.json
```diff
@@ -0,0 +1,180 @@
...
+                                    }
+                                ]
+                            }
+                        },
+                        {
+                            "type": "Question",
+                            "id": "breakdown-block",
+                            "question": {
+                                "id": "breakdown-question",
+                                "title": "Breakdown validated against answers value source",
```
berroar, line 68:
```suggestion
                                "title": "Breakdown validated against an answer value source",
```

petechd:
> Should we add metadata as a source for the schema/tests as I think the card suggests that this should be covered too?
> 
> Edit: Nevermind just re-read the PR description 😅

Yeah, the card should be updated, we had a brief discussion about this in the channel last week.

app/forms/questionnaire_form.py
```diff
@@ -176,11 +176,24 @@ def _get_target_total_and_currency(
...
     ) -> Optional[tuple[Union[Calculation, AnswerValueTypes], Optional[str]]]:
 
+        value_source_resolver = ValueSourceResolver(
+            answer_store=self.answer_store,
+            schema=self.schema,
+            metadata=self.metadata,
+            response_metadata=self.response_metadata,
+            list_store=self.list_store,
+            location=self.location,
+            list_item_id=None,
```
MebinAbraham, line 186: This needs to be set to work within repeats.

MebinAbraham, line 186: This needs to be set to work within repeats.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -1221,6 +1221,88 @@ def test_get_calculation_total_with_no_input(app, answer_store, list_store):
...
+            question_schema,
+            answer_store,
+            list_store,
+            metadata={},
+            response_metadata={},
+            form_data=form_data,
+        )
+
+        form.validate()
+        assert form.data == expected_form_data
```
MebinAbraham, line 1261: How does this assert test `sum_calculated_field_value_source_answer`. I was expecting to assert the value or validation errors?.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -1221,6 +1221,88 @@ def test_get_calculation_total_with_no_input(app, answer_store, list_store):
...
```
MebinAbraham, line 1303: Same as above.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -1221,6 +1221,88 @@ def test_get_calculation_total_with_no_input(app, answer_store, list_store):
...
+
+
+def test_sum_calculated_field_value_source_calculated_summary(
```
MebinAbraham, line 1264: We should have one that doesn't validate. Also, we should test within a repeat also.

app/forms/questionnaire_form.py
petechd: Ok added.

tests/app/forms/test_questionnaire_form.py
petechd: Added one with validation error.

tests/app/forms/test_questionnaire_form.py
petechd: It checks if the form passes the validation against the value source which is answers in this case.

tests/app/forms/test_questionnaire_form.py
petechd: Same, this time for calculated summary.

tests/app/forms/test_questionnaire_form.py
petechd: I've added repeat to existing schema that has a breakdown questions for a list collector, added another breakdown with value source. Adjusted a functional test accordingly.

tests/app/forms/test_questionnaire_form.py
MebinAbraham: If you are checking it passes validation, then it should assert that a validation error is not raised.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -1221,6 +1221,204 @@ def test_get_calculation_total_with_no_input(app, answer_store, list_store):
...
```
MebinAbraham, line 1264: Can these be parametrised?

tests/app/forms/test_questionnaire_form.py
```diff
@@ -1221,6 +1221,204 @@ def test_get_calculation_total_with_no_input(app, answer_store, list_store):
...
+            form_data=form_data,
+        )
+
+        form.validate()
+        assert form.question_errors["breakdown-question"] == schema.error_messages[
+            "TOTAL_SUM_NOT_EQUALS"
+        ] % {"total": "10"}
+
+
+def test_sum_calculated_field_value_source_calculated_summary_not_equal_validation_error(
```
MebinAbraham, line 1343: Same for these, can they be parametrised, a lot of duplication ath the moment.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -1221,6 +1221,204 @@ def test_get_calculation_total_with_no_input(app, answer_store, list_store):
...
+        ] == schema.error_messages["TOTAL_SUM_NOT_EQUALS"] % {"total": "10"}
+
+
+def test_sum_calculated_field_value_source_calculated_summary_repeat_not_equal_validation_error(
+    app, answer_store, list_store
+):
+
+    answer_store.add_or_update(
+        Answer(answer_id="entertainment-spending-answer", value=10)
+    )
```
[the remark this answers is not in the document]
petechd: Added list item id and all the other elements needed to make sure this is actual repeat.

tests/app/forms/test_questionnaire_form.py
petechd: Parametrised, thanks!

app/forms/questionnaire_form.py
```diff
@@ -176,11 +176,25 @@ def _get_target_total_and_currency(
...
+        list_item_id = self.location.list_item_id if self.location else None
+        value_source_resolver = ValueSourceResolver(
+            answer_store=self.answer_store,
+            schema=self.schema,
+            metadata=self.metadata,
+            response_metadata=self.response_metadata,
+            list_store=self.list_store,
+            location=self.location,
+            list_item_id=list_item_id,
+        )
```
MebinAbraham, line 188: I would make this a property of the form. You only need one instance of this per form instance.

tests/app/forms/test_questionnaire_form.py
MebinAbraham: Doesn't seem to have changed. Is it committed?

tests/app/forms/test_questionnaire_form.py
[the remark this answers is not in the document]
MebinAbraham: But what was the issue?. If it was passing without an id, is it still valid?. Also Line 1323 still doesn't have list item id, but that is a duplicate answer id?.

app/forms/questionnaire_form.py
petechd: Ok, top property of the form now, cheers!

tests/app/forms/test_questionnaire_form.py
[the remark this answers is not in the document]
petechd: I can't recreate it passing without list item id anymore, the test is fixed now had to mock value source resolve return value of one of the private methods.

tests/app/forms/test_questionnaire_form.py
petechd: This one is addressed now.

app/questionnaire/questionnaire_schema.py
```diff
@@ -268,19 +268,26 @@ def _update_answer_dependencies_for_calculated_summary(
...
+                    continue
+
             dependents = {
                 self._get_answer_dependent_for_block_id(
                     block_id=self.get_block_for_answer_id(answer_id)["id"]  # type: ignore
                 )
                 for answer_id in calculation["answers_to_calculate"]
             }
-            self._answer_dependencies_map[source_answer_id] |= dependents
+            self._answer_dependencies_map[source_answer_id] |= dependents  # type: ignore
```
MebinAbraham, line 290: This shouldn't be type ignored. The logic allows it to get this far without `source_answer_id`.

I would update to something like:
```py
            if source_answer_id := calculation.get("answer_id"):
                dependents = {
                    self._get_answer_dependent_for_block_id(
                        block_id=self.get_block_for_answer_id(answer_id)["id"]  # type: ignore
                    )
                    for answer_id in calculation["answers_to_calculate"]
                }
                self._answer_dependencies_map[source_answer_id] |= dependents

            elif isinstance(calculation.get("value"), dict):
                    calculation["value"],
                    block_id=block_id,
                )
```

*Note: not tested with suggested change.*

tests/app/forms/test_questionnaire_form.py
MebinAbraham: Not sure it is. https://github.com/ONSdigital/eq-questionnaire-runner/pull/882/files#r910129711

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,156 @@
...
+
+const answerAndSubmitSecondBreakdownQuestion = (breakdown1, breakdown2, breakdown3, breakdown4) => {
+  $(SecondBreakdownAnswerPage.secondBreakdown1()).setValue(breakdown1);
+  $(SecondBreakdownAnswerPage.secondBreakdown2()).setValue(breakdown2);
+  $(SecondBreakdownAnswerPage.secondBreakdown3()).setValue(breakdown3);
+  $(SecondBreakdownAnswerPage.secondBreakdown4()).setValue(breakdown4);
+  $(SecondBreakdownAnswerPage.submit()).click();
+};
+
+describe("Feature: Sum of grouped answers equal to validation against answer value source ", () => {
```
MebinAbraham, line 23:
```suggestion
describe("Feature: Sum of grouped answers equal to validation against value source ", () => {
```

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,156 @@
...
+      $(TotalAnswerPage.submit()).click();
+
+      answerAndSubmitBreakdownQuestion("3", "3", "3", "3");
+
+      expect(browser.getUrl()).to.contain(TotalPlaybackPage.pageName);
+    });
+  });
+
+  describe("Given I start a grouped answer validation survey and enter 12 into the total", () => {
+    it("When I continue to second breakdown and enter values equal to calculated summary total, Then I should be able to get to the summary page", () => {
```
MebinAbraham, line 40: This needs work, to me it is not very clear. It is saying `enter 12 into the total` then `enter values equal to calculated summary total` but the first case of entering 12 is totally redundant. I would mention Given you have a calculated summary value of 12, When ... etc.

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,156 @@
...
+
+      answerAndSubmitBreakdownQuestion("6", "3", "3", "3");
+
+      $(TotalPlaybackPage.submit()).click();
+
+      answerAndSubmitSecondBreakdownQuestion("3", "3", "2", "1");
+
+      expect(browser.getUrl()).to.contain(SubmitPage.pageName);
+    });
+  });
```
MebinAbraham, line 77: This test doesn't ensure you have to reconfirm with valid answers. It just shows that you walk back through it, which can change when we implement skipping completed questions. Like the existing test, I would show you submitting without a change which shows the error message meaning you have to reconfirm with correct answers. You have done it for the test case below.

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,156 @@
...
+  describe("Given I completed both grouped answer validation questions and I am on the summary", () => {
+    it("When I go back from the summary and change the total, Then I must reconfirm the breakdown question based on answer value source with valid answers before I can continue", () => {
+      $(TotalAnswerPage.total()).setValue("12");
+      $(TotalAnswerPage.submit()).click();
+
+      answerAndSubmitBreakdownQuestion("3", "3", "3", "3");
+
+      $(TotalPlaybackPage.submit()).click();
+
+      answerAndSubmitSecondBreakdownQuestion("2", "2", "1", "1");
```
MebinAbraham, line 88: You use these lines a couple of times, could be a `answer_all_questions()` method.

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,156 @@
...
+
+      $(SubmitPage.totalAnswerEdit()).click();
+      $(TotalAnswerPage.total()).setValue("15");
+      $(TotalAnswerPage.submit()).click();
+
+      answerAndSubmitBreakdownQuestion("0", "3", "3", "3");
+
+      expect(browser.getUrl()).to.contain(BreakdownAnswerPage.pageName);
+
+      $(BreakdownAnswerPage.submit()).click();
```
MebinAbraham, line 98: This is already done as part of `answerAndSubmitBreakdownQuestion`

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,156 @@
...
+      $(TotalAnswerPage.total()).setValue("5");
+      $(TotalAnswerPage.submit()).click();
+
+      answerAndSubmitBreakdownQuestion("3", "3", "3", "3");
+
+      expect($(BreakdownAnswerPage.errorNumber(1)).getText()).to.contain("Enter answers that add up to 5");
+    });
+  });
+
+  describe("Given I start a grouped answer validation survey and enter 5 into the total", () => {
```
MebinAbraham, line 142: Are these two not already covered?

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,156 @@
...
+      $(SubmitPage.totalAnswerEdit()).click();
+      $(TotalAnswerPage.total()).setValue("15");
+      $(TotalAnswerPage.submit()).click();
+
+      answerAndSubmitBreakdownQuestion("0", "3", "3", "3");
+
+      expect(browser.getUrl()).to.contain(BreakdownAnswerPage.pageName);
+
+      $(BreakdownAnswerPage.submit()).click();
+      expect($(BreakdownAnswerPage.errorNumber(1)).getText()).to.contain("Enter answers that add up to 15");
```
MebinAbraham, line 99: We should show you can actually continue by correcting it.

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,156 @@
...
+      answerAndSubmitBreakdownQuestion("6", "3", "3", "3");
+
+      $(TotalPlaybackPage.submit()).click();
+
+      answerAndSubmitSecondBreakdownQuestion("1", "1", "1", "1");
+
+      expect(browser.getUrl()).to.contain(SecondBreakdownAnswerPage.pageName);
+
+      $(SecondBreakdownAnswerPage.submit()).click();
+      expect($(BreakdownAnswerPage.errorNumber(1)).getText()).to.contain("Enter answers that add up to 9");
```
MebinAbraham, line 127: We should show you can actually continue by correcting it.

app/questionnaire/questionnaire_schema.py
petechd: Ok, changed it to your logic now, works fine.

tests/app/forms/test_questionnaire_form.py
petechd: We do:
```python 
if errors_text:
    for error_text in errors_text:
        assert (
            form.question_errors[question]
            == schema.error_messages[error_text] % value_dict
        )
else:
    assert len(form.question_errors) == 0
```
..so checking if validation errors are not raised?

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: do you mean in this spec? We do have validation passing when the value total is correct but not when it's less or more than total value entered before. Second one is for value source validation error, first for answer source.

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: Removed!

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: Added as `asnwerBothBreakdownQuestions` since these are not all the answers and there are only two breakdown ones which we both answer here.

app/questionnaire/questionnaire_schema.py
```diff
@@ -268,19 +268,32 @@ def _update_answer_dependencies_for_calculated_summary(
...
                 continue
+
             dependents = {
                 self._get_answer_dependent_for_block_id(
                     block_id=self.get_block_for_answer_id(answer_id)["id"]  # type: ignore
                 )
                 for answer_id in calculation["answers_to_calculate"]
             }
-            self._answer_dependencies_map[source_answer_id] |= dependents
+            self._answer_dependencies_map[source_answer_id] |= dependents  # type: ignore
```
MebinAbraham, line 296: This is a duplicate, and the `continue` above isn't needed.

tests/functional/spec/features/validation/sum/value_source.spec.js
MebinAbraham: > Like the existing test, I would show you submitting without a change which shows the error message meaning you have to reconfirm with correct answers. You have done it for the test case below.

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,142 @@
...
+  describe("Given I completed both grouped answer validation questions and I am on the summary", () => {
+    it("When I go back from the summary and change the total, Then I must reconfirm both breakdown questions with valid answers before I can get to the summary", () => {
+      $(TotalAnswerPage.total()).setValue("12");
+      $(TotalAnswerPage.submit()).click();
+
+      answerBothBreakdownQuestions(["3", "3", "3", "3"], ["2", "2", "1", "1"]);
+
+      $(SubmitPage.totalAnswerEdit()).click();
+      $(TotalAnswerPage.total()).setValue("15");
+      $(TotalAnswerPage.submit()).click();
```
MebinAbraham, line 67: After this it should be asserting that you can proceed, so I would assert the error message.

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,142 @@
...
+
+      answerAndSubmitBreakdownQuestion("0", "3", "3", "3");
+
+      expect(browser.getUrl()).to.contain(BreakdownAnswerPage.pageName);
```
MebinAbraham, line 88: Is this needed?. If you at this point the error should already show up as it wasn't equal 15 as previous values add up to 12.

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,142 @@
...
+      expect($(BreakdownAnswerPage.errorNumber(1)).getText()).to.contain("Enter answers that add up to 15");
+
+      answerBothBreakdownQuestions(["5", "4", "4", "2"], ["3", "3", "2", "1"]);
+
+      expect(browser.getUrl()).to.contain(SubmitPage.pageName);
+    });
+  });
+
+  describe("Given I completed both grouped answer validation questions and I am on the summary", () => {
+    it("When I go back from the summary and change the total, Then I must reconfirm the breakdown question based on answer value source with valid answers before I can continue", () => {
```
MebinAbraham, line 99: This is the same as one above, so it doesn't make sense.

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -0,0 +1,142 @@
...
+      $(SubmitPage.totalAnswerEdit()).click();
+      $(TotalAnswerPage.total()).setValue("15");
+      $(TotalAnswerPage.submit()).click();
+
+      answerBothBreakdownQuestions(["6", "3", "3", "3"], ["1", "1", "1", "1"]);
+
+      expect(browser.getUrl()).to.contain(SecondBreakdownAnswerPage.pageName);
+
+      expect($(SecondBreakdownAnswerPage.errorNumber(1)).getText()).to.contain("Enter answers that add up to 9");
+
```
MebinAbraham, line 114: Why are we changing the value source answer here?. This is just about calculated summary value and it's breakdown.

tests/functional/spec/features/validation/sum/value_source.spec.js
MebinAbraham: No I mean these are technically already covered by the other, but happy to keep.

tests/app/forms/test_questionnaire_form.py
petechd: It does, if validation errors are raised it checks its contents. In `"schema_name, block, answers, breakdowns, expected_form_data, question, errors_text, value_dict",`  `error_text` name might be misleading, it is `validation error text`.

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: Ok, added assertion that no errors are displayed on this page (after line 67), I hope that's what we wanted?

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: Yeah, wrong description the one above is `Then I must reconfirm the breakdown question based on answer value source`, this one `Then I must reconfirm the breakdown question based on calculated summary value source`.

app/questionnaire/questionnaire_schema.py
```diff
@@ -268,19 +268,23 @@ def _update_answer_dependencies_for_calculated_summary(
             }
 
     def _update_answer_dependencies_for_calculations(
-        self,
-        calculations: tuple[ImmutableDict[str, Any]],
+        self, calculations: tuple[ImmutableDict[str, Any]], block_id: str
```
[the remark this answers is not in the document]
petechd: Added, thanks!

tests/app/forms/test_questionnaire_form.py
MebinAbraham: If it raises a validator error, then that is what the first asset should be instead of `assert form.data == expected_form_data`, cause that doesn't really do anything.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
MebinAbraham: Outstanding. It isn't enforced. Need to use `*`.

tests/functional/spec/features/validation/sum/value_source.spec.js
MebinAbraham: Not quite, we need to assert you can't proceed until you change you answer. So maybe assert the error message or url before you answer the question.

tests/functional/spec/features/validation/sum/value_source.spec.js
MebinAbraham: Still outstanding.

tests/functional/spec/features/validation/sum/value_source.spec.js
MebinAbraham: But you are changing the first answer value which this test case shouldn't care about as that you test in the one above. This should just be when you edit the breakdown that make up the calculated summary total, then the second breakdown has validator errors?.

petechd: This test is also about changing the value of the total answer (first answer value), I just do what is in the description, if we don't want to change the total answer then this test should have different description or needs to be removed if we don't want to test that.

MebinAbraham: You already test changing answer value source in the previous spec. L76.

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: I changed the other description now so it is calculated summary value source. Did I missed something here?

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: Ok, assertion is added, we're checking that you cannot progress after trying to resubmit previous breakdown values.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
petechd:
```python 
def _update_answer_dependencies_for_calculations(
        self, calculations: tuple[ImmutableDict[str, Any]], *, block_id: str
    ) -> None:
```

tests/functional/spec/features/validation/sum/value_source.spec.js
MebinAbraham: The test is still edit the main answer instead of just the breakdown.

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: I've added error check when submitting without a change: `expect($(BreakdownAnswerPage.singleErrorLink()).isDisplayed()).to.be.true;` but will also add another one to check it's contents:
`expect($(BreakdownAnswerPage.errorNumber(1)).getText()).to.contain("Enter answers that add up to 15");`

MebinAbraham (approved):

berroar (approved):

PR #888, opened 2022-06-27
Add conditional links to error pages

petechd:
This PR adds conditional error pages based on survey_type/theme. Survey types are taken from new Enum class and can be easily changed/extended in one place. Error templates extend base error template now which allows us to set survey types in one template only. For details check [trello card](https://trello.com/c/HvFZVHba).

### How to review 
A new integration tests file added to only test error pages templates.

### Checklist


petechd:
[Conditional links: Error pages (3 SP)](https://trello.com/c/HvFZVHba/5338-conditional-links-error-pages-3-sp)

tests/integration/routes/test_error_page_templates.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+
+    def test_500_theme_default(self):
+        # Given
+        self.launchSurvey("test_introduction")
+
+        # When
+        with patch(
+            "app.routes.questionnaire.get_block_handler",
+            side_effect=Exception("You broked it"),
```
berroar, line 155:
```suggestion
            side_effect=Exception("You broke it"),
```

berroar (approved):
Tested, all looks ok to me 👍

app/routes/errors.py
```diff
@@ -49,10 +50,24 @@ def log_exception(exception, status_code):
 
 def _render_error_page(status_code, template=None, **kwargs):
     handle_language()
+    business_survey_config = get_survey_config(theme=SurveyTypes.business.value)
+    other_survey_config = get_survey_config(theme=SurveyTypes.social.value)
```
MebinAbraham, line 54: We should be passing around the survey type enum instead of the string value.

tests/integration/routes/test_error_page_templates.py
```diff
@@ -0,0 +1,233 @@
+from mock import Mock, patch
```
MebinAbraham, line 1: We already have a `test_errors.py` integration test file, these should live there.

MebinAbraham, line 1: We should have tests for the content which covers, the scenarios:
- authenticated
- not authenticated, but cookie exists
- not authenticated and no cookie. For this both the business and other link needs to be checked.

app/helpers/template_helpers.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         self._google_tag_manager_auth = current_app.config.get(
             "EQ_GOOGLE_TAG_MANAGER_AUTH"
         )
+        self._survey_type = cookie_session.get("theme")
```
MebinAbraham, line 52: The `cookie_has_theme` in `service_links_context` should be updated to use `self._survey_type` now.

tests/integration/routes/test_error_page_templates.py
petechd: I've added what I could and tried more descriptive test names but certain errors are impossible to recreate without being authenticated (eg. 500, submission failed) and certain occur only when unauthenticated (401)?

MebinAbraham: Apart from `submission failed`, the rest can be under those 3 categories.

app/survey_config/survey_types.py
```diff
@@ -0,0 +1,11 @@
+from enum import Enum
+
+
+class SurveyTypes(Enum):
+    business = "business"
+    social = "social"
+    default = "default"
+    health = "health"
+    northernireland = "northernireland"
```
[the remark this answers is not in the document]
petechd: We might have to keep it for the sake of this code:
```python
def get_survey_type() -> SurveyType:
    return SurveyType[cookie_session.get("theme", current_app.config["SURVEY_TYPE"])]
```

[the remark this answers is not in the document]
petechd: `"theme": "northernireland"` in some of ours schemas hence it won't get picked up in enum.

tests/integration/routes/test_errors.py
```diff
@@ -80,3 +80,265 @@ def test_errors_500_exception_during_error_handling(self):
...
+        self.launchSurvey("test_theme_social")
+        self.post()
+        self.post()
+        self.post()
+
+        # Then
+        self.assertStatusCode(500)
+        self.assertInBody(
+            '<p>If this problem keeps happening, please <a href="http://upstream.url/contact-us/">contact us</a> for help.</p>'
+        )
```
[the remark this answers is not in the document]
petechd: Updated, so was that technically a bug in social surveys? Tests updated as well.

templates/errors/403.html
```diff
@@ -1,10 +1,18 @@
...
+{% extends 'errors/_base.html' %}
 
 {% set page_title = _("Sorry, there is a problem") %}
 
 {% block main %}
-<h1>{{ _("Sorry, there is a problem") }}</h1>
-<p>{{ _("You may need to update your browser to a newer version.") }}</p>
-<p>{{ _("If the problem still occurs, try using a different browser or device.") }}</p>
-<p>{{ _("For further help, please <a href='{url}'>contact us</a>.").format(url=contact_us_url) }}</p>
+    <p>{{ survey_type }}</p>
```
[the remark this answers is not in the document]
petechd: Removed.

app/helpers/template_helpers.py
```diff
@@ -188,40 +191,39 @@ def survey_config_mapping(
...
     language = language or get_locale().language
-    theme = theme or get_survey_type()
+    survey_theme: str = theme.value if theme else get_survey_type()
     base_url = (
         cookie_session.get("account_service_base_url") or ACCOUNT_SERVICE_BASE_URL
     )
 
     return survey_config_mapping(
-        theme=theme,
+        theme=survey_theme,
```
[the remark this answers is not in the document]
petechd: Ok, passing `SurveyType` into `get_survey_type`.

app/survey_config/survey_types.py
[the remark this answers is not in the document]
MebinAbraham: I was referring to the attribute not the value. i.e `northern_ireland = "northernireland"`

tests/integration/routes/test_errors.py
```diff
@@ -80,3 +80,265 @@ def test_errors_500_exception_during_error_handling(self):
 
                     self.assertStatusCode(500)
                     self.assertInBody("Sorry, there is a problem with this service")
+
+    def test_401_theme_default_cookie_exists(self):
```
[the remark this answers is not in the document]
petechd: I removed some duplication/moved it to smaller methods but some of the tests do have differences in what looks like common parts.

app/survey_config/survey_type.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class SurveyType(Enum):
+    business = "business"
+    social = "social"
+    default = "default"
+    health = "health"
+    northernireland = "northernireland"
+    census = "census"
+    census_nisra = "census-nisra"
```
MebinAbraham, line 11: Normally constants are uppercase. No hard and fast rule but worth being consistent. They are uppercase in validator so worth doing the same here. Also, there's this: https://mail.python.org/pipermail/python-ideas/2016-September/042340.html

tests/integration/routes/test_errors.py
```diff
@@ -80,3 +80,250 @@ def test_errors_500_exception_during_error_handling(self):
...
+
+        # When
+        token = 123
+        self.get(url=f"/session?token={token}")
+
+        # Then
+        self.assertStatusForbidden()
+        self.assertInBody(
+            '<p>If you are completing a business survey and you need further help, please <a href="https://surveys.ons.gov.uk/contact-us/">contact us</a>.</p>'
+        )
```
MebinAbraham, line 176: Should assert both cases, so business and the "other". Same applies to the rest.

tests/integration/integration_test_case.py
```diff
@@ -452,6 +452,20 @@ def assertInRedirect(self, content):
...
+        self.post()
+        self.post()
+
+    def getUrlAndCookie(self, url):
+        self.get(url=url)
+        return self.getCookie()
+
+    def deleteCookieAndGetUrl(self, url):
+        self.deleteCookie()
+        self.get(url=url)
```
MebinAbraham, line 467: This file should only contain common methods. These seem specific, if so they should live in the test file.

tests/integration/routes/test_errors.py
```diff
@@ -80,3 +80,250 @@ def test_errors_500_exception_during_error_handling(self):
...
+        # When
+        current_url = self.last_url
+        self.exit()
+        self.deleteCookieAndGetUrl(current_url)
+
+        # Then
+        self.assertStatusUnauthorised()
+        self.assertInBody(
+            '<p>If you are completing a business survey, you need to sign back in to <a href="https://surveys.ons.gov.uk/sign-in/logout">your account</a>.</p>'
+        )
```
MebinAbraham, line 134: Should assert the "other" case also.

tests/integration/routes/test_errors.py
```diff
@@ -80,3 +80,250 @@ def test_errors_500_exception_during_error_handling(self):
...
+        self.launchSurvey("test_introduction")
+
+        # When
+        self.deleteCookieAndGetUrl("/abc123")
+
+        # Then
+        self.assertStatusNotFound()
+        self.assertInBody(
+            "<p>If the web address is correct or you selected a link or button, please see the following help links.</p>"
+        )
```
MebinAbraham, line 217: Should assert the cases i.e business/other text.

tests/integration/routes/test_errors.py
```diff
@@ -1,9 +1,9 @@
-from unittest.mock import patch
+from mock import Mock, patch
```
MebinAbraham, line 1: Worth updating the assert to cover for no cookie, the business content and the "other surveys" content.

app/survey_config/survey_types.py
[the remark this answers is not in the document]
MebinAbraham: Also, the code needs changing, it should be an instance not bracket access. That is why you had issue changing to `norther_ireland`.
`SurveyType["some_key"]` would be SurveyType by the key.
`SurveyType("some_key")` would be SurveyType by the value.

So the code should be updated to:
```py
def get_survey_type() -> SurveyType:
    survey_type = cookie_session.get("theme", current_app.config["SURVEY_TYPE"])
    return SurveyType(survey_type)
```

app/views/contexts/view_submitted_response_context.py
```diff
@@ -17,14 +18,14 @@ def build_view_submitted_response_context(
...
-    survey_type: str,
+    survey_type: SurveyType,
 ) -> dict[str, Union[str, datetime, dict]]:
 
     view_submitted_response_expired = has_view_submitted_response_expired(
         questionnaire_store.submitted_at  # type: ignore
     )
 
-    if survey_type == "social":
+    if survey_type.value == "social":
```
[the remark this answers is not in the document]
petechd: Removed all instances of this, there was one more in the code.

app/survey_config/survey_types.py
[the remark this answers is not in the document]
petechd: Ok, thanks for this great explanation, now it works like a charm!

app/survey_config/survey_type.py
petechd: All of these changed now, thanks!

tests/integration/routes/test_errors.py
petechd: Added  in templates where the "other" option exists, thanks!

tests/integration/integration_test_case.py
petechd: Moved them across but kept naming convention should we change that as well?

app/survey_config/social_survey_config.py
```diff
@@ -11,14 +11,20 @@
...
 
     def __post_init__(self):
         super().__post_init__()
 
+        self.base_url = "https://rh.ons.gov.uk"
+        self.contact_us_url = f"{self.base_url}/contact-us/"
+        self.cookie_settings_url = f"{self.base_url}/cookies/"
+        self.privacy_and_data_protection_url = (
+            f"{self.base_url}/privacy-and-data-protection/"
+        )
```
MebinAbraham, line 26: - This is already called in the `__post_init__` should be removed. 
- The base URL shouldn't be hardcoded here as it would override the account service URL set by the env var/cookie. When creating the social config for the non cookie pages, you will need to pass it the base url to use, otherwise it will get it from the cookie.

You will need to pass in a base url when you want the social survey config without the

petechd: This one is slightly confusing because of all the relations between the classes and conditions we need to meet, I left the hardcoded `base_url` but not in `post init`, only top level of social config class for the sake of correct `account_service_log_out_url` being generated for social. Otherwise `base_url` is initialised differently in template helpers module now:
```python
base_url = "https://rh.ons.gov.uk" if survey_theme == SurveyType.SOCIAL else (
        cookie_session.get("account_service_base_url") or ACCOUNT_SERVICE_BASE_URL
    )
```
that seems to do the work for us?

petechd: if we don't want hardcoded rh link in class init should we just?:
```python
if not self.account_service_log_out_url:
            self.account_service_log_out_url: str = "https://rh.ons.gov.uk/sign-in/logout"
```

petechd:
Ok, added what we discussed, the only difference: I was forced to assign the social base url in post init of social survey class.

MebinAbraham:
> I was forced to assign the social base url in post init of social survey class.

Why?

app/helpers/template_helpers.py
```diff
@@ -188,40 +193,41 @@ def survey_config_mapping(
...
     # The fallback to assigning SURVEY_TYPE to theme is only being added until
     # business feedback on the differentiation between theme and SURVEY_TYPE.
+    if session_store := get_session_store():
+        if session_data := session_store.session_data:
+            schema = load_schema_from_session_data(session_data)
+
     language = language or get_locale().language
-    theme = theme or get_survey_type()
-    base_url = (
+    survey_theme = theme if theme else get_survey_type()
```
MebinAbraham, line 208: Can this not stay as `theme or get_survey_type()`?

app/settings.py
```diff
@@ -149,6 +149,10 @@ def utcoffset_or_fail(date_value, key):
     "ACCOUNT_SERVICE_BASE_URL", "https://surveys.ons.gov.uk"
 )
 
+ACCOUNT_SERVICE_BASE_URL_SOCIAL = os.getenv(
+    "ACCOUNT_SERVICE_BASE_URL_SOCIAL", "https://rh.ons.gov.uk"
+)
```
MebinAbraham, line 154: Just needs to be a constant, as we would never set an env var for it given it is only used if you don't have a cookie.

petechd: Removed the env var, was not sure about it so added just in case, thanks.

petechd:
> > I was forced to assign the social base url in post init of social survey class.
> 
> Why?

it's because it uses Survey Config default base url otherwise, needs to be done post initialisation.

MebinAbraham:
> > > I was forced to assign the social base url in post init of social survey class.
> > 
> > 
> > Why?
> 
> I think it's because `contact us` is created "dynamically" in post init of Survey Config and we use Survey Config `post_init` in Social Survey Config which is based on its base_url. So we end up with base in contact us from the standard Survey Config.

The post init will inherit from the parent base_url. I am not seeing this issue?

petechd:
> > > > I was forced to assign the social base url in post init of social survey class.
> > > 
> > > 
> > > Why?
> > 
> > 
> > I think it's because `contact us` is created "dynamically" in post init of Survey Config and we use Survey Config `post_init` in Social Survey Config which is based on its base_url. So we end up with base in contact us from the standard Survey Config.
> 
> The post init will inherit from the parent base_url. I am not seeing this issueI

MebinAbraham:
> > > > > I was forced to assign the social base url in post init of social survey class.
> > > > 
> > > > 
> > > > Why?
> > > 
> > > 
> > > I think it's because `contact us` is created "dynamically" in post init of Survey Config and we use Survey Config `post_init` in Social Survey Config which is based on its base_url. So we end up with base in contact us from the standard Survey Config.
> > 
> > 
> > The post init will inherit from the parent base_url. I am not seeing this issue?
> 
> I think it's because we super parent class we somehow take the parent `base_url` in `contact us` with it. Breaks contact us tests if you try.

Can't recreate with the following:
```py
from app.survey_config import SocialSurveyConfig

social = SocialSurveyConfig()
print(social.base_url)
print(social.contact_us_url)


# https://rh.ons.gov.uk
# https://rh.ons.gov.uk/contact-us/

social_with_overridden_url = SocialSurveyConfig(base_url="https://example.com")
print(social_with_overridden_url.base_url)
print(social_with_overridden_url.contact_us_url)

# https://example.com
# https://example.com/contact-us/
```

app/survey_config/social_survey_config.py
```diff
@@ -11,12 +12,12 @@
 class SocialSurveyConfig(
     SurveyConfig,
 ):
-    base_url: str = "https://rh.ons.gov.uk"
     survey_title: str = "ONS Social Surveys"
     footer_links: Iterable[MutableMapping] = field(default_factory=list)
     footer_legal_links: Iterable[Mapping] = field(default_factory=list)
 
     def __post_init__(self):
+        self.base_url = ACCOUNT_SERVICE_BASE_URL_SOCIAL
```
MebinAbraham, line 20: This should be reverted. It should be on the instance attribute instead of `post_init`. This also mean you never use the base url you get passed in which is also not correct.

tests/integration/routes/test_errors.py
```diff
@@ -80,3 +80,287 @@ def test_errors_500_exception_during_error_handling(self):
...
+
+        # Then
+        self.assertStatusUnauthorised()
+        self.assertInBody(
+            [
+                (
+                    '<p>If you are completing a business survey, you need to sign back in to <a href="https://surveys.ons.gov.uk/sign-in/logout">your account</'
+                    "a>.</p>"
+                ),
+                '<p>If you started your survey using an access code, you need to <a href="https://rh.ons.gov.uk/sign-in/logout">re-enter your code</a>.</p>',
```
MebinAbraham, line 138: Worth referencing these links using the constants instead of  hardcoding, we will so have to change the RH URL so minimise the no. of changes. Same for the other tests also.

tests/app/helpers/test_template_helpers.py
```diff
@@ -48,7 +48,7 @@ def test_footer_context_business_theme(app: Flask, expected_footer_business_them
 
 def test_footer_context_social_theme(app: Flask, expected_footer_social_theme):
     with app.test_client():
-        survey_config = SocialSurveyConfig()
+        survey_config = SocialSurveyConfig(base_url="https://rh.ons.gov.uk")
```
MebinAbraham, line 51: Why did this change?

tests/app/helpers/test_template_helpers.py
```diff
@@ -291,7 +291,7 @@ def test_service_links_context(
             "https://surveys.ons.gov.uk/contact-us/",
         ),
         (
-            SocialSurveyConfig(),
+            SocialSurveyConfig(base_url="https://rh.ons.gov.uk"),
```
MebinAbraham, line 294: Same for these.

tests/integration/routes/test_errors.py
petechd: Added, cheers!

tests/integration/routes/test_errors.py
```diff
@@ -80,3 +80,287 @@ def test_errors_500_exception_during_error_handling(self):
...
+        self.saveAndSignOut()
+        self.get(current_url)
+
+        # Then
+        self.assertStatusUnauthorised()
+        cookie = self.getCookie()
+        self.assertEqual(cookie.get("theme"), "social")
+        self.assertInBody(
+            '<p>To access this page you need to <a href="https://rh.ons.gov.uk/sign-in/logout">re-enter your access code</a>.</p>'
+        )
```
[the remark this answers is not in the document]
petechd: Added, cheers!

berroar (approved):

MebinAbraham (approved):
As discussed, the issue with cookie banner will be addressed at a later stage.

PR #890, opened 2022-06-28
Update mutually exclusive options to use multiple

rmccar:
Updates Runner to use radio buttons when there is more than one mutually exclusive answer to a question.

Also updates some test text to make it make more sense

### How to review 
- Use the `test_mutually_exclusive_multiple` schema to test questions with multiple mutually exclusive options
- Use the `test_mutually_exclusive` schema to test questions with single mutually exclusive options still work as expected

### Checklist


scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=allow-radios-as-mutually-exclusive-answers
```
rmccar, line 2: Reminder to revert

rmccar:
> ed to add some additional functional tests for this. It is broken for me at the moment as whe

Ive just retested myself and it works ok. How can I recreate this error?

berroar:
> > We'll need to add some additional functional tests for this. It is broken for me at the moment as when I only answer with one of the new mutually radio options in the `test_mutually_exclusive_multiple` schema , I can't get to the next question as it fails validation with `Select at least one answer`
> 
> Ive just retested myself and it works ok for me. How can I recreate this error?

I think it's ok actually, it looks like it was because I was on older version of the DS. But we still will need some new tests to demonstrate this new feature.

rmccar:
> > > We'll need to add some additional functional tests for this. It is broken for me at the moment as when I only answer with one of the new mutually radio options in the `test_mutually_exclusive_multiple` schema , I can't get to the next question as it fails validation with `Select at least one answer`
> > 
> > 
> > Ive just retested myself and it works ok for me. How can I recreate this error?
> 
> I think it's ok actually, it looks like it was because I was on older version of the DS. But we still will need some new tests to demonstrate this new feature.

Yeah I'm just adding them now 👍

app/forms/questionnaire_form.py
```diff
@@ -155,8 +155,9 @@ def validate_mutually_exclusive_question(
             question["validation"].get("messages") if "validation" in question else None
         )
         answers = (getattr(self, answer["id"]).data for answer in question["answers"])
-        is_only_checkboxes = all(
-            answer["type"] == "Checkbox" for answer in question["answers"]
+        is_only_checkboxes_or_radios = all(
+            answer["type"] == "Checkbox" or answer["type"] == "Radio"
```
MebinAbraham, line 159: Did this need to change?. We can't have radios for the non-exclusive option but it only matters if it was checkbox anyway.

MebinAbraham (commented):
Needs tests, otherwise LGTM.

MebinAbraham:
> > > > We'll need to add some additional functional tests for this. It is broken for me at the moment as when I only answer with one of the new mutually radio options in the `test_mutually_exclusive_multiple` schema , I can't get to the next question as it fails validation with `Select at least one answer`
> > > 
> > > 
> > > Ive just retested myself and it works ok for me. How can I recreate this error?
> > 
> > 
> > I think it's ok actually, it looks like it was because I was on older version of the DS. But we still will need some new tests to demonstrate this new feature.
> 
> Yeah I'm just adding them now 👍

I would base them of the existing tests for mutually exclusive to avoid duplicate most of it

app/jinja_filters.py
```diff
@@ -364,11 +364,11 @@ def __init__(
 
 
 @blueprint.app_template_filter()  # type: ignore
-def map_select_config(form: FormType, answer: AnswerType) -> list[SelectConfig]:
-    options = form["fields"][answer["id"]]
+def map_select_config(form: FormType, answers: AnswerType) -> list[SelectConfig]:
+    options = form["fields"][answers["id"]]
```
MebinAbraham, line 368: How come this changed, is it still singular?.

rmccar: I changed it because it could now accept multiple so I just thought it made more sense

MebinAbraham: The answer type and code doesn't reflect that. You can't do key access on an array. It is still one answer?

MebinAbraham: Ya I think that make most sense, the options var is already plural.

tests/app/forms/validation/test_mutually_exclusive_validator.py
```diff
@@ -25,14 +25,14 @@
     ),
 )
 def test_mutually_exclusive_mandatory_checkbox_raises_ValidationError(
-    answer_permutations, is_mandatory, is_only_checkboxes, error_type
+    answer_permutations, is_mandatory, is_only_checkboxes_or_radios, error_type
```
MebinAbraham, line 28: - This test doesn't include an example with a radio answer, the current ones are for checkbox exclusive.
- Should rename the test to remove `checkbox`.
- `ValidationError` in the test name should be `validation_error` (I know it is existing but worth changing while you are here).

tests/app/forms/validation/test_mutually_exclusive_validator.py
```diff
@@ -57,5 +57,5 @@ def test_mutually_exclusive_mandatory_checkbox(answer_permutations, is_mandatory
     validator(
         answer_values=iter(answer_permutations),
         is_mandatory=is_mandatory,
-        is_only_checkboxes=True,
+        is_only_checkboxes_or_radios=True,
     )
```
MebinAbraham, line 61: This test also need to update the parametrize values to include an example that would match a radio. The current one are all a `list` which means it is testing checkbox.

MebinAbraham (commented):
Looks good, just a couple of issue with unit tests.

tests/app/forms/validation/test_mutually_exclusive_validator.py
petechd: Renamed the tests and added various radio scenarios (I hope I got it right), thanks.

tests/app/forms/validation/test_mutually_exclusive_validator.py
petechd: Renamed the test since it's not only checkboxes being tested now and added some radio parametrize values to include various radios.

MebinAbraham (approved):

berroar (approved):

PR #894, opened 2022-06-30
Fix detail answer value bug in convert_payload_0_0_1

berroar:

Identified a bug on submission in our `covert_payload_0_0_1.py` while testing the BICS survey. When a user selected a detail answer field but did not set an answer, we were getting a 500 error as `detail_answer.value` could not be resolved as the value the `detail_answer` was  a `None` type.

Added a check to ensure that the `detail_answer` was not a None type before getting the value.

### How to review 

Check all tests pass.

### Checklist

### DIFF
--- a/app/submitter/convert_payload_0_0_1.py
+++ b/app/submitter/convert_payload_0_0_1.py
@@ -127,6 +127,7 @@ def _get_checkbox_answer_data(
             if "detail_answer" in option:
                 detail_answer = answer_store.get_answer(option["detail_answer"]["id"])
                 if detail_answer:
+                    # Ignore mypy type because the answer type can be any non strings, but user_answer is expected to be a string.
                     user_answer = detail_answer.value  # type: ignore
 
             qcodes_and_values.append((option.get("q_code"), user_answer))
--- a/app/submitter/convert_payload_0_0_1.py
+++ b/app/submitter/convert_payload_0_0_1.py
@@ -126,10 +126,8 @@ def _get_checkbox_answer_data(
         if option:
             if "detail_answer" in option:
                 detail_answer = answer_store.get_answer(option["detail_answer"]["id"])
-                # if the user has selected an option with a detail answer we need to find the detail answer value it refers to.
-                # the detail answer maybe None and the value could be empty, in this case we just use the main value (e.g. other)
-                if detail_answer:
-                    user_answer = detail_answer.value or user_answer  # type: ignore
+                if detail_answer and detail_answer.value:
+                    user_answer = detail_answer.value  # type: ignore
 
             qcodes_and_values.append((option.get("q_code"), user_answer))
 
--- a/app/submitter/convert_payload_0_0_1.py
+++ b/app/submitter/convert_payload_0_0_1.py
@@ -126,7 +126,7 @@ def _get_checkbox_answer_data(
         if option:
             if "detail_answer" in option:
                 detail_answer = answer_store.get_answer(option["detail_answer"]["id"])
-                if detail_answer and detail_answer.value:
+                if detail_answer:
                     user_answer = detail_answer.value  # type: ignore
 
             qcodes_and_values.append((option.get("q_code"), user_answer))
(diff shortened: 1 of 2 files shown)


berroar:
[Fix detail answer value bug in convert payload](https://trello.com/c/ivnjFjvE/5357-fix-detail-answer-value-bug-in-convert-payload)

app/submitter/convert_payload_0_0_1.py
```diff
@@ -127,8 +127,9 @@ def _get_checkbox_answer_data(
             if "detail_answer" in option:
                 detail_answer = answer_store.get_answer(option["detail_answer"]["id"])
                 # if the user has selected an option with a detail answer we need to find the detail answer value it refers to.
-                # the detail answer value can be empty, in this case we just use the main value (e.g. other)
-                user_answer = detail_answer.value or user_answer  # type: ignore
+                # the detail answer maybe None and the value could be empty, in this case we just use the main value (e.g. other)
+                if detail_answer:
+                    user_answer = detail_answer.value or user_answer  # type: ignore
```
MebinAbraham, line 132: The `or` part isn't needed?

MebinAbraham (commented):
The change needs a unit test.

app/submitter/convert_payload_0_0_1.py
berroar: I added the or as there is an existing test which expect the `user_answer` to be `Other` rather than empty when detail_answer.value doesn't exist

berroar: The existing test is `test_converter_checkboxes_with_q_codes_and_empty_other_value`

MebinAbraham: We don't store empty answers?

berroar:
> The change needs a unit test.

Added a test for the scenario that caused the error.

app/submitter/convert_payload_0_0_1.py
berroar: I've updated this now but can't get past the mypy type hinting error  🤯

app/submitter/convert_payload_0_0_1.py
```diff
@@ -127,8 +127,9 @@ def _get_checkbox_answer_data(
             if "detail_answer" in option:
                 detail_answer = answer_store.get_answer(option["detail_answer"]["id"])
                 # if the user has selected an option with a detail answer we need to find the detail answer value it refers to.
-                # the detail answer value can be empty, in this case we just use the main value (e.g. other)
-                user_answer = detail_answer.value or user_answer  # type: ignore
+                # the detail answer maybe None and the value could be empty, in this case we just use the main value (e.g. other)
```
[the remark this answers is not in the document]
berroar: I have removed the comment as not sure it is really needed given the change

petechd (approved):

app/submitter/convert_payload_0_0_1.py
```diff
@@ -126,9 +126,8 @@ def _get_checkbox_answer_data(
         if option:
             if "detail_answer" in option:
                 detail_answer = answer_store.get_answer(option["detail_answer"]["id"])
-                # if the user has selected an option with a detail answer we need to find the detail answer value it refers to.
-                # the detail answer value can be empty, in this case we just use the main value (e.g. other)
-                user_answer = detail_answer.value or user_answer  # type: ignore
+                if detail_answer:
+                    user_answer = detail_answer.value  # type: ignore
```
MebinAbraham, line 130: I would add a comment that the ignore is because answer type can be any non strings, but this expects to be a string.

app/submitter/convert_payload_0_0_1.py
berroar: Removed invalid test that was causing confusion

tests/app/submitter/test_convert_payload_0_0_1.py
```diff
@@ -275,15 +275,14 @@ def test_converter_checkboxes_with_q_codes_and_other_value(fake_questionnaire_st
     assert answer_object["data"]["4"] == "Bacon"
 
 
-def test_converter_checkboxes_with_q_codes_and_empty_other_value(
+def test_converter_checkboxes_with_missing_detail_answer_value_in_answer_store(
```
MebinAbraham, line 278: Test name need updating.

app/submitter/convert_payload_0_0_1.py
berroar: Added a type hint comment

PR #900, opened 2022-07-13
Remove metadata from session

berroar:

Removes metadata from the session in order to reduce the amount of data stored as session data. This will help to reduce the overall datastore payload size, and will also prevent duplication as metadata is stored and available in our questionnaire state.

The following attributes will still be kept in our session data as they are updated as part of the session: `language_code`, `confirmation_email_count`, `feedback_count`,.

All other existing session_data values will be set to None to allow runner to be temporarily backwards compatible, and are to be removed in the future.

### How to review 

Check nothing is broken by this change.

Have any items been missed.

### Checklist


berroar:
[Remove metadata from session (5 SP)](https://trello.com/c/ptysFep6/4824-remove-metadata-from-session-5-sp)

app/utilities/schema.py
```diff
@@ -98,14 +98,15 @@ def get_allowed_languages(schema_name, launch_language):
     return [DEFAULT_LANGUAGE_CODE]
 
 
-def load_schema_from_metadata(metadata):
+def load_schema_from_metadata(metadata, language_code=None):
```
petechd, line 101: Enforcing keyword arguments would improve readability?
```suggestion
def load_schema_from_metadata(*, metadata, language_code=None):
```
This will break all the invocations of `load_schema_from_metadata` but easy to fix.

berroar: Added this 👍

app/routes/individual_response.py
```diff
@@ -60,11 +60,15 @@ def before_individual_response_request():
         "individual-response request", method=request.method, url_path=request.full_path
     )
 
-    handle_language()
+    handle_language(metadata=metadata)
```
petechd, line 63:
```suggestion
    handle_language(metadata)
```

app/routes/questionnaire.py
```diff
@@ -95,11 +95,13 @@ def before_questionnaire_request():
         "questionnaire request", method=request.method, url_path=request.full_path
     )
 
-    handle_language()
+    handle_language(metadata=metadata)
```
petechd, line 98:
```suggestion
    handle_language(metadata)
```

app/routes/questionnaire.py
```diff
@@ -112,21 +114,25 @@ def before_post_submission_request():
...
+    # pylint: disable=assigning-non-slot
     questionnaire_store = get_questionnaire_store(
         current_user.user_id, current_user.user_ik
     )
 
     if not questionnaire_store.submitted_at:
         raise NotFound
 
-    handle_language()
+    handle_language(metadata=metadata)
```
petechd, line 126:
```suggestion
    handle_language(metadata)
```

app/authentication/authenticator.py
```diff
@@ -103,8 +108,10 @@ def load_user(extend_session: bool = True) -> Optional[User]:
         user_ik = cookie_session.get(USER_IK)
         user = User(user_id, user_ik)
 
-        if session_store.session_data and session_store.session_data.tx_id:
-            logger.bind(tx_id=session_store.session_data.tx_id)
+        metadata = get_metadata(user)
+
+        if metadata and metadata.get("tx_id"):
+            logger.bind(tx_id=metadata.get("tx_id"))
```
petechd, line 114: I would do something similar to what you did with session data in langauge helpers, so we avoid getting it twice.
```suggestion
        if metadata and (tx_id := metadata.get("tx_id")):
            logger.bind(tx_id=tx_id)
```

app/data_models/session_data.py
```diff
@@ -1,13 +1,14 @@
...
 from typing import Any, Optional
 
 
 class SessionData:
     def __init__(
         self,
+        language_code: Optional[str],
         tx_id: Optional[str],
         schema_name: Optional[str],
         period_str: Optional[str],
```
MebinAbraham, line 11: Any reason for the order change?. Worth also enforcing keyword args. I think we always pass as keyword args already but just not enforced. If it was, this order change wouldn't have mattered.

app/helpers/language_helper.py
```diff
@@ -13,17 +14,18 @@
 }
 
 
-def handle_language() -> None:
+def handle_language(metadata: Optional[Mapping[str, Any]] = None) -> None:
+    metadata = metadata or get_metadata(current_user)
     session_store = get_session_store()
+    schema_name = metadata.get("schema_name") if metadata else None
```
MebinAbraham, line 20: You could set metadata to empty dict and avoid the `if`.
```py
metadata = metadata or get_metadata(current_user) or {}
schema_name = metadata["schema_name"]
```
Also used bracket access as `schema_name` is currently mandatory. If we had type hints, this would error at the moment. When we come to have schema_name/schema_url, this should break during dev which will help. Non blocker as this will change in the next card anyway as we will have a metadata proxy.

app/helpers/language_helper.py
```diff
@@ -13,17 +14,18 @@
 }
 
 
-def handle_language() -> None:
+def handle_language(metadata: Optional[Mapping[str, Any]] = None) -> None:
+    metadata = metadata or get_metadata(current_user)
```
MebinAbraham, line 18: This should be inside the `if session` check as it is only needed if that is true.

app/helpers/language_helper.py
```diff
@@ -13,17 +14,18 @@
...
+        launch_language = session_data.language_code or DEFAULT_LANGUAGE_CODE
         # pylint: disable=assigning-non-slot
-        g.allowed_languages = get_allowed_languages(
-            session_data.schema_name, launch_language
-        )
+        g.allowed_languages = get_allowed_languages(schema_name, launch_language)
         request_language = request.args.get("language_code")
         if request_language and request_language in g.allowed_languages:
-            session_store.session_data.language_code = request_language
+            session_data.language_code = request_language
```
MebinAbraham, line 28: We should keep this how it was `session_store.session_data.language_code`. If session data returned a copy, this would fail. It is usually not a good idea to updates vars that are technically referencing the same memory. It is clearer to update what we want at the source.

app/helpers/schema_helpers.py
```diff
@@ -27,8 +28,17 @@ def wrapped_function(*args: Any, **kwargs: Any) -> Any:
...
 
-        session_data = session_store.session_data
-        schema = load_schema_from_session_data(session_data)
+        metadata = get_metadata(current_user)
+
+        language_code = (
+            session_store.session_data.language_code
+            if session_store.session_data
+            else None
+        )
```
MebinAbraham, line 37: I would chain the if session data check with the if not session store check. If you didn't have session data, you wouldn't have metadata anyway. It would this make this clearer.

```py
        if not session_store or not session_store.session_data:
            raise Unauthorized

        language_code = session_store.session_data.language_code
```

We are missing a lot of issues because we haven't type checked the load schema methods. It's worth updating the methods the touches. i.e. here metadata won't be None but can be according to type hints and `load_schema_from_metadata` would fail if metadata was None.

app/routes/dump.py
```diff
@@ -17,9 +17,12 @@
 def requires_schema(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
-        session = get_session_store()
         # pylint: disable=assigning-non-slot
-        g.schema = load_schema_from_session_data(session.session_data)
+        session_store = get_session_store()
+        metadata = get_metadata(current_user)
+        g.schema = load_schema_from_metadata(
+            metadata=metadata, language_code=session_store.session_data.language_code
```
MebinAbraham, line 24: We should use `get_locale().language` instead of getting the session store here just for language.

Also worth updating the logic for `get_locale()` since we have started to call it from more places. It is currently doing:
```py
session = get_session_store()
return session.session_data.language_code if session else None
```

But should be updated to handle no session data.

```py
session = get_session_store()
if session and session_data := session.session_data:
	return session_data.language_code

return None
```

app/routes/individual_response.py
```diff
@@ -60,11 +60,15 @@ def before_individual_response_request():
...
+    handle_language(metadata)
 
     session_store = get_session_store()
+
     # pylint: disable=assigning-non-slot
-    g.schema = load_schema_from_session_data(session_store.session_data)
+    g.schema = load_schema_from_metadata(
+        metadata=questionnaire_store.metadata,
+        language_code=session_store.session_data.language_code,
+    )
```
MebinAbraham, line 71: I would remove the call to session store here and instead use `get_locale().language` which will return the language code from session.

app/utilities/schema.py
```diff
@@ -98,14 +98,15 @@ def get_allowed_languages(schema_name, launch_language):
     return [DEFAULT_LANGUAGE_CODE]
 
 
-def load_schema_from_metadata(metadata):
+def load_schema_from_metadata(*, metadata, language_code=None):
```
MebinAbraham, line 101: We should enforce the keyword argument just for `language_code`. Normally the keyword argument isn't enforced for the thing the method is named after, i.e `load_schema_from_metadata` implies you need to pass in metadata. Plus if you only had 2 args, enforcing on 1 covers all cases.

app/utilities/schema.py
```diff
@@ -98,14 +98,15 @@ def get_allowed_languages(schema_name, launch_language):
...
+    metadata = metadata or {}
     if schema_url := metadata.get("schema_url"):
         # :TODO: Remove before production uses schema_url
         # This is temporary and is only for development/integration purposes.
         # This should not be used in production.
 
         start = time.time()
-        schema = load_schema_from_url(schema_url, metadata.get("language_code"))
+        schema = load_schema_from_url(schema_url, language_code)
```
MebinAbraham, line 109: The logic for `language_code` should be handled in the method. Language code is optional, so it should fetch from metadata if it is not passed in.

app/utilities/schema.py
```diff
@@ -122,14 +123,10 @@ def load_schema_from_metadata(metadata):
         return schema
 
     return load_schema_from_name(
-        metadata.get("schema_name"), language_code=metadata.get("language_code")
+        metadata.get("schema_name"), language_code=language_code
     )
```
MebinAbraham, line 127: Same as above. Language should come from metadata if not passed in.

app/views/contexts/thank_you_context.py
```diff
@@ -19,23 +18,28 @@
 
 def build_thank_you_context(
     schema: QuestionnaireSchema,
-    session_data: SessionData,
+    questionnaire_store_metadata: dict,
```
MebinAbraham, line 21: Just call it `metadata` like we do elsewhere?

app/views/handlers/confirm_email.py
```diff
@@ -153,17 +148,17 @@ def _publish_fulfilment_request(self):
 @dataclass
 class ConfirmationEmailFulfilmentRequest(FulfilmentRequest):
     email_address: str
-    session_data: SessionData
+    questionnaire_store: QuestionnaireStore
```
MebinAbraham, line 151: We should only be passing info that is needed. Here, it should be passing in metadata, not the entire questionnaire store. Technically it should pass the 3 items it needs not the entire metadata either but happy to keep that given that is what we were doing with session data.

tests/app/helpers/test_schema_helper.py
```diff
@@ -6,7 +6,14 @@
...
+def test_questionnaire_schema_passed_into_function(
+    mocker, session_store, fake_questionnaire_store
+):
+    mocker.patch(
+        "app.helpers.schema_helpers.get_metadata",
+        return_value=fake_questionnaire_store.metadata,
+    )
+
     mocker.patch(
         "app.helpers.schema_helpers.get_session_store",
```
MebinAbraham, line 18: Is this still needed?

app/data_models/session_data.py
berroar: Enforced keyword args and removed the extra keys

app/helpers/schema_helpers.py
berroar: Updated this and added type hint

app/helpers/template_helpers.py
```diff
@@ -200,11 +200,14 @@ def get_survey_config(
 ) -> SurveyConfig:
     # The fallback to assigning SURVEY_TYPE to theme is only being added until
     # business feedback on the differentiation between theme and SURVEY_TYPE.
+
     if session_store := get_session_store():
         if session_data := session_store.session_data:
-            schema = load_schema_from_session_data(session_data)
+            language = session_data.language_code
```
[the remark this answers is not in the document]
berroar: Removed

app/routes/dump.py
berroar: Updated get_locale 👍

app/utilities/schema.py
berroar: Only enforcing language code now 👍

app/utilities/schema.py
berroar: Updated the method to get from metadata if the language code is not passed 👍

app/utilities/schema.py
berroar: Updated the method to get from metadata if the language code is not passed 👍

app/views/contexts/thank_you_context.py
berroar: Renamed it but there is another metadata var later in the method which I also had to rename

tests/app/helpers/test_schema_helper.py
berroar: Yeah the test still fails without it

app/authentication/authenticator.py
```diff
@@ -103,8 +108,10 @@ def load_user(extend_session: bool = True) -> Optional[User]:
         user_ik = cookie_session.get(USER_IK)
         user = User(user_id, user_ik)
 
-        if session_store.session_data and session_store.session_data.tx_id:
-            logger.bind(tx_id=session_store.session_data.tx_id)
+        metadata = get_metadata(user)
+
+        if metadata and (tx_id := metadata.get("tx_id")):
+            logger.bind(tx_id=tx_id)
```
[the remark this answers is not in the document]
rmccar: If we don't need it I think we should get rid of it

rmccar (approved):
Looks ok to me

app/views/handlers/confirm_email.py
```diff
@@ -153,17 +148,17 @@ def _publish_fulfilment_request(self):
...
-    session_data: SessionData
+    metadata: dict[str, Any]
     schema: QuestionnaireSchema
 
     def _payload(self) -> Mapping:
         return {
             "fulfilmentRequest": {
                 "email_address": self.email_address,
-                "display_address": self.session_data.display_address,
+                "display_address": self.metadata["display_address"],
```
MebinAbraham, line 158: I would use `.get()` as `display_address` would only exist in the census flow. If we make it graceful, we could reuse for others even if those didn't have `display_address`.

app/views/handlers/confirm_email.py
```diff
@@ -153,17 +148,17 @@ def _publish_fulfilment_request(self):
...
                 "form_type": self.schema.form_type,
-                "language_code": self.session_data.language_code,
+                "language_code": self.metadata["language_code"],
```
MebinAbraham, line 160: This language code should be the one from session as it should be the language the user is currently in.

tests/app/data_model/test_session_store.py
```diff
@@ -66,12 +65,12 @@ def test_add_data_to_session(app, app_session_store):
...
         ).save()
-        display_address = "68 Abingdon Road, Goathill"
-        app_session_store.session_store.session_data.display_address = display_address
+        feedback_count = 9
+        app_session_store.session_store.session_data.feedback_count = feedback_count
         app_session_store.session_store.save()
 
         session_store = SessionStore("user_ik", "pepper", "eq_session_id")
-        assert session_store.session_data.display_address == display_address
+        assert session_store.session_data.feedback_count == 9
```
MebinAbraham, line 73:
```suggestion
        assert session_store.session_data.feedback_count == feedback_count
```

tests/app/views/contexts/conftest.py
```diff
@@ -197,19 +197,26 @@ def response_metadata():
...
 
 
+@pytest.fixture
+def fake_questionnaire_store_metadata():
+    return {
+        "tx_id": "tx_id",
+        "schema_name": "some_schema_name",
+        "period_str": "period_str",
+        "language_code": None,
+        "launch_language_code": None,
```
MebinAbraham, line 211: `launch_language_cod` is not a valid metadata, that key only existed in session data.
This is the issue with just using an arbitrary dictionary to store this data. I was going to suggest we need a data model like session data, but given we need that for the new launch/reception pattern I am happy to leave the work for that.

tests/app/views/handlers/test_confirmation_email_fulfilment_request.py
```diff
@@ -8,22 +8,25 @@
 
 @freeze_time(time_to_freeze)
 def test_confirmation_email_fulfilment_request_message(
-    session_data, confirmation_email_fulfilment_schema
+    fake_questionnaire_store, confirmation_email_fulfilment_schema
 ):
```
MebinAbraham, line 12: This test doesn't seem to use `questionnaire_store` so just the metadata needs to be passed in as the fixture.

MebinAbraham (commented):
Minor issues.

app/helpers/language_helper.py
```diff
@@ -13,17 +14,21 @@
...
     if session_store and (session_data := session_store.session_data):
-        launch_language = session_data.launch_language_code or DEFAULT_LANGUAGE_CODE
-        # pylint: disable=assigning-non-slot
-        g.allowed_languages = get_allowed_languages(
-            session_data.schema_name, launch_language
+        metadata = metadata or get_metadata(current_user) or {}
+        schema_name = metadata.get("schema_name")
+
+        launch_language = (
+            session_store.session_data.language_code or DEFAULT_LANGUAGE_CODE
```
MebinAbraham, line 25: This needs to get `language_code` from metadata, as session data one is the current language code not necessarily the launch one.

app/data_models/session_data.py
```diff
@@ -1,46 +1,35 @@
...
-        self.case_ref = case_ref
-        self.trad_as = trad_as
-        self.account_service_base_url = account_service_base_url
-        self.account_service_log_out_url = account_service_log_out_url
-        self.display_address = display_address
         self.confirmation_email_count = confirmation_email_count
         self.feedback_count = feedback_count
-        self.schema_url = schema_url
+
+        # :TODO: To be removed when we migrate to get these vars from questionnaire_store.metadata.
```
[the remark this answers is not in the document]
berroar: Updated this 👍

app/routes/questionnaire.py
```diff
@@ -95,11 +95,12 @@ def before_questionnaire_request():
         "questionnaire request", method=request.method, url_path=request.full_path
     )
 
-    handle_language()
+    handle_language(metadata)
 
-    session_store = get_session_store()
     # pylint: disable=assigning-non-slot
```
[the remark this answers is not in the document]
berroar: Yeah this is still needed

app/routes/questionnaire.py
```diff
@@ -112,21 +113,24 @@ def before_post_submission_request():
     if not metadata:
         raise NoQuestionnaireStateException(401)
 
+    # pylint: disable=assigning-non-slot
```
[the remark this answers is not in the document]
berroar: This was on the wrong line, it is still required but have moved to the relevant line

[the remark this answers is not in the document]
MebinAbraham: But why, is it our code of the usage of `current_user`?

tests/app/views/contexts/conftest.py
MebinAbraham: Still outstanding.

app/routes/questionnaire.py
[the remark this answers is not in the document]
berroar: Ok have removed this now, it is not need since the last change

tests/app/views/contexts/conftest.py
berroar: Removed now

app/routes/questionnaire.py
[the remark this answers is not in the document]
berroar: Require for the method `g.schema = load_schema_from_metadata(
        metadata=metadata, language_code=get_locale().language
    )`

Otherwise get pylint error: `Assigning to attribute 'schema' not defined in class slots (assigning-non-slot)`

app/routes/questionnaire.py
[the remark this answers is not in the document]
berroar: Same as above

app/data_models/session_data.py
```diff
@@ -1,46 +1,35 @@
...
+        self.ru_name = None
+        self.ru_ref = None
+        self.response_id = None
+        self.case_id = None
+        self.case_ref = None
+        self.trad_as = None
+        self.account_service_base_url = None
+        self.account_service_log_out_url = None
+        self.display_address = None
+        self.schema_url = None
```
MebinAbraham, line 32: As discussed, we'll need to keep all these temporarily, just in case if we need to roll back. Once it has rolled out successfully, we can then remove these properties in a subsequent deployment. For now, they will be set but won't get used unless we rollback.

berroar: This has been reverted now 👍

MebinAbraham (approved):

tests/app/conftest.py
```diff
@@ -1,10 +1,12 @@
 # pylint: disable=redefined-outer-name
 
 from datetime import datetime, timedelta, timezone
+from unittest.mock import MagicMock
```
petechd, line 4:
```suggestion
from mock import MagicMock
```
we don't want to use unittest libraries with pytest

petechd (approved):

PR #903, opened 2022-07-20
Fix DAC issue summary table headers

rmccar:
Table heads in summaries are including change in summaries where the change link isnt used. This PR will fix this and only list the change header when the link is used in the table.

Also removes `test_questionnaire_content_variants.py` because it was an empty file

### How to review
Compare summaries with change answer links to summaries that don't have the change answer link. See that the table header in the html only has change listed when a change link is in the table. The `test-view-submitted-response` schema is good for this

### Checklist


rmccar:
[Third `th` is removed for non-editable summaries](https://trello.com/c/1IfsweRK/5379-third-th-is-removed-for-non-editable-summaries)

berroar (changes requested):
If this is a bug then it will need a functional test

tests/functional/spec/features/section_summary/section_summary.spec.js
```diff
@@ -164,4 +175,21 @@ describe("Section Summary", () => {
       expect($(HouseholdDetailsSummaryPage.heading()).getText()).to.contain("Household Summary - Semi-detached");
     });
   });
+  describe("Given I start a questionnaire that has a summary with no change links", () => {
+    it("When I then navigate to that summary, then the correct headers should be set", () => {
```
berroar, line 179: This could be more specific in the `then`, something like `change answer should not be displayed`?

templates/partials/summary/collapsible-summary.html
```diff
@@ -2,6 +2,10 @@
 {% from "components/summary/_macro.njk" import onsSummary %}
 
 {%- set itemList = [] -%}
+{%- set tableHeader = ["Question", "Answer given"] -%}
+{%- if content.summary.answers_are_editable -%}
+  {{ tableHeader.append( "Change answer" ) or "" }}
```
berroar, line 7: Why is the `or ""` needed, the functional tests appear to pass without it? Is there a test scenario where this is needed that we need to add?

templates/partials/summary/collapsible-summary.html
```diff
@@ -2,6 +2,10 @@
 {% from "components/summary/_macro.njk" import onsSummary %}
 
 {%- set itemList = [] -%}
+{%- set tableHeader = ["Question", "Answer given"] -%}
```
petechd, line 5: This is duplicated in the other template below, we could avoid adding the logic to the template (and repetition) by changing relevant context (eg. what `SummaryContext` class returns). `SectionSummaryContext` and `CalculatedSummaryContext` answers are always editable so we don't need to do the check.

templates/partials/summary/collapsible-summary.html
rmccar: Its needed because when there is a change link Jinja will print out "None" like this:

Screenshot 2022-08-08 at 08 49 29

This is because strings are truthy in Jinja, and `None` is falsy, adding the `or ""` will return the empty string and then output nothing on the page

berroar: I know this logic is in the backend now, but when I remove `else: headers.extend("")` from the code, I don't see this `None` when I get to a summary page

berroar: Which schema did you use to test this cause I can't get the `None` to appear when I remove the else?

berroar: So this means we can remove:
`else:
            headers.extend("")`?

rmccar: Yeah true, ill remove that 👍

app/views/contexts/summary_context.py
```diff
@@ -14,11 +14,16 @@ def __call__(
         groups = list(self._build_all_groups(return_to))
         summary_options = self._schema.get_summary_options()
         collapsible = summary_options.get("collapsible", False)
+        headers = ["Question", "Answer given"]
+        if answers_are_editable:
+            headers.extend("Change answer")
```
petechd, line 19: We could do some list comprehension here to reduce lines and I find it more readable but up to you.
```suggestion
        headers = ["Question", "Answer given", "Change answer"] if answers_are_editable else ["Question", "Answer given"]
```

petechd (approved):

tests/functional/spec/features/section_summary/section_summary.spec.js
```diff
@@ -24,6 +29,12 @@ describe("Section Summary", () => {
       expect($(PropertyDetailsSummaryPage.submit()).getText()).to.contain("Continue");
     });
 
+    it("When I get to the section summary page, Then it should have the correct table headers set'", () => {
+      expect($(PropertyDetailsSummaryPage.propertyDetailsSummaryTableHead()).getHTML()).to.contain("Question");
+      expect($(PropertyDetailsSummaryPage.propertyDetailsSummaryTableHead()).getHTML()).to.contain("Answer given");
+      expect($(PropertyDetailsSummaryPage.propertyDetailsSummaryTableHead()).getHTML()).to.contain("Change answer");
```
MebinAbraham, line 35: These shouldn't be functional tests, you are checking the source HTML. These should be in integration tests in which you can test the HTML.

tests/functional/spec/features/section_summary/section_summary.spec.js
```diff
@@ -164,4 +175,21 @@ describe("Section Summary", () => {
...
+      $(AddressBlockPage.answer()).setValue("NP10 8XG");
+      $(AddressBlockPage.submit()).click();
+      $(ViewSubmittedResponseSubmitPage.submit()).click();
+      expect(browser.getUrl()).to.contain(ThankYouPage.pageName);
+      expect($(ThankYouPage.title()).getHTML()).to.contain("Thank you for completing the Test");
+      $(ThankYouPage.savePrintAnswersLink()).click();
+      expect(browser.getUrl()).to.contain(ViewSubmittedResponsePage.pageName);
+      expect($(ViewSubmittedResponsePage.addressDetailsGroupSummaryTableHead()).getHTML()).to.contain("Question");
+      expect($(ViewSubmittedResponsePage.addressDetailsGroupSummaryTableHead()).getHTML()).to.contain("Answer given");
+      expect($(ViewSubmittedResponsePage.addressDetailsGroupSummaryTableHead()).getHTML()).to.not.contain("Change answer");
```
MebinAbraham, line 192: Same as above.

tests/functional/generate_pages.py
```diff
@@ -221,6 +221,14 @@
 """
 )
 
+SUMMARY_TABLE_HEAD = Template(
```
berroar, line 224: Is this still being used? Could remove it now we are only integration testing?

app/views/contexts/summary_context.py
```diff
@@ -14,11 +14,18 @@ def __call__(
         groups = list(self._build_all_groups(return_to))
         summary_options = self._schema.get_summary_options()
         collapsible = summary_options.get("collapsible", False)
+        headers = (
+            ["Question", "Answer given", "Change answer"]
+            if answers_are_editable
+            else ["Question", "Answer given"]
+        )
```
MebinAbraham, line 21: We should mark these strings for translation.

MebinAbraham: (Non blocker) I would avoid the duplication:
```suggestion
        headers = ["Question", "Answer given"]
        if answers_are_editable:
            headers.append("Change answer")
```

tests/integration/questionnaire/test_questionnaire_section_summary.py
```diff
@@ -0,0 +1,23 @@
...
+        self.launchSurvey("test_section_summary")
+        self.post()
+        self.post()
+        self.post()
+        self.post()
+        self.assertInBody("<th>Question</th>")
+        self.assertInBody("<th>Answer given</th>")
+        self.assertInBody("<th>Change answer</th>")
+
+    def test_calculated_summary_headers_without_change_link(self):
```
MebinAbraham, line 15: These shouldn't be in a new file, we already have tests for summary (TestQuestionnaireSubmitWithSummary) and view submitted response (TestViewSubmissionResponse), it should live there.

MebinAbraham: The name of the tests need changing, this is not doing anything from the calculated summary.

tests/functional/generate_pages.py
```diff
@@ -603,6 +611,7 @@ def write_summary_spec(page_spec, section, collapsible, answers_are_editable=Fal
                 "group_id": group["id"],
             }
             page_spec.write(SUMMARY_TITLE_GETTER.substitute(group_context))
+            page_spec.write(SUMMARY_TABLE_HEAD.substitute(group_context))
```
MebinAbraham, line 614: This also needs removing.

tests/app/views/contexts/test_calculated_summary_context.py
```diff
@@ -94,7 +94,7 @@ def test_build_view_context_for_currency_calculated_summary(
 
     assert "summary" in context
     assert_summary_context(context)
-    assert len(context["summary"]) == 6
+    assert len(context["summary"]) == 7
```
MebinAbraham, line 97: The content of `headers` should be unit tested. I don't see one where answers aren't editable.

tests/app/views/contexts/test_section_summary_context.py
```diff
@@ -53,7 +53,7 @@ def test_build_view_context_for_section_summary(
 
     assert "summary" in context
     assert_summary_context(context)
-    assert len(context["summary"]) == 6
+    assert len(context["summary"]) == 7
```
MebinAbraham, line 56: The content of `headers` should be unit tested. I don't see one where answers aren't editable.

tests/integration/questionnaire/test_questionnaire_submit.py
```diff
@@ -127,3 +127,23 @@ def test_is_displayed(self):
...
+        self.launchSurvey("test_section_summary")
+        self.post()
+        self.post()
+        self.post()
+        self.post()
+        self.assertInBody("<th>Question</th>")
+        self.assertInBody("<th>Answer given</th>")
+        self.assertInBody("<th>Change answer</th>")
+
+    def test_summary_headers_without_change_link(self):
```
MebinAbraham, line 141: This should be in the view submitted response test file.

MebinAbraham (approved):

tests/integration/routes/test_view_submitted_response.py
```diff
@@ -58,6 +58,9 @@ def test_enabled(self):
         self.assertInBody("NP10 8XG")
         self.assertIsNotNone(self.get_print_button())
         self.assertIsNotNone(self.get_download_button())
+        self.assertInBody("<th>Question</th>")
+        self.assertInBody("<th>Answer given</th>")
+        self.assertNotInBody("<th>Change answer</th>")
```
MebinAbraham, line 63: How come these moved?. These shouldn't be in this test case.

rmccar: The test was failing where it was originally because the timeout was being expired so the table wasn't displayed so I moved it into this test because I thought it was a good place to put it as it was testing all other parts of the page

MebinAbraham: Alright, happy for it to stay.

berroar (approved):

PR #909, opened 2022-07-28
Change jinja config to mimic nunjucks more closely

rmccar:
This PR will update to v55.2.0 of the DS (the version where parent param checks are starting to be removed) and change the jinja config to use `ChainableUndefined` for undefined params. This should mean that there should be much less problems integrating DS changes into Runner in future as parameters will no longer needed to be tested to see if they exist before using them in jinja.

### How to review 
- Tests pass with this change implemented
- Manually run through some surveys to make sure everything looks as expected

### Checklist


rmccar:
[Update jinja config to mimic nunjucks more closely](https://trello.com/c/wTADCvVG/5396-update-jinja-config-to-mimic-nunjucks-more-closely)

MebinAbraham (approved):
Not my preference but being pragmatic.

berroar (approved):

PR #910, opened 2022-07-29
Add conditional title in survey header

petechd:
This adds conditional title in survey header based on the presence of cookie (theme).

### How to review 
Added extra tests for  survey configs with no cookie and updated the ones were cookie theme key is supposed to be present now. Check the tests. 

### Checklist


petechd:
[Conditional text: Survey header (1 SP)](https://trello.com/c/X2TNCqym/5376-conditional-text-survey-header-1-sp)

app/helpers/template_helpers.py
```diff
@@ -112,10 +112,11 @@ def page_header_context(self) -> dict[str, Union[bool, str, LazyString]]:
         context: dict[str, Union[bool, str, LazyString]] = {
             "orgLogo": f"{self._survey_config.page_header_logo}",
             "orgLogoAlt": f"{self._survey_config.page_header_logo_alt}",
+            "title": self._survey_title
+            if self._survey_title and cookie_session.get("theme")
+            else "ONS Surveys",
```
MebinAbraham, line 117: The title is being set in two different ways now. You have this "title" inside the header context but we also set the title in the page config in the `_base.html`. Need to find out which way it should be done. Currently they are both being set. It currently works because at least one or the other is set, but we need to understand when they should be used and whether both should be used.

tests/app/helpers/test_template_helpers.py
```diff
@@ -103,86 +103,115 @@ def test_footer_context_census_nisra_theme(app: Flask, expected_footer_nisra_the
...
-        "titleLogoAlt": "Census 2021",
-        "customHeaderLogo": True,
-        "orgMobileLogo": "nisra-logo-mobile",
-    }
-
+@pytest.mark.parametrize(
+    "theme, survey_config, expected",
+    (
+        (
+            "business",
```
MebinAbraham, line 110: I would replace these string with the enum class reference.

app/helpers/template_helpers.py
```diff
@@ -112,10 +112,11 @@ def page_header_context(self) -> dict[str, Union[bool, str, LazyString]]:
         context: dict[str, Union[bool, str, LazyString]] = {
             "orgLogo": f"{self._survey_config.page_header_logo}",
             "orgLogoAlt": f"{self._survey_config.page_header_logo_alt}",
+            "title": self._survey_title
+            if self._survey_title and cookie_session.get("theme")
```
MebinAbraham, line 116: `cookie_session.get("theme")` is available as `self._survey_type`.

tests/app/helpers/test_template_helpers.py
petechd: Replaced with enum.

app/helpers/template_helpers.py
petechd: Using `self._survey_type` now.

app/helpers/template_helpers.py
MebinAbraham: They do overlap because you introduced the `title` in the header context in this PR but the title for the banner already works.

petechd: But I didn't introduce `title` in header context as it was already:
```python
if self._survey_title:
            context["title"] = self._survey_title
```
Do you mean I introduced default for the title in header context?

petechd: From what I can see if we don't define it template helpers it is taken from env var `SURVEY_TYPE` and if that is not defined it defaults in settings.py to `BusinessSurveyConfig` survey_title since `SURVEY_TYPE = os.getenv("SURVEY_TYPE", "business")` in `setttings.py`

petechd: If we don't have it defined in context (as we did before) it uses value from the template which is different as it uses both page title and survey title instead of just survey title in the header element.

MebinAbraham: > But I didn't introduce `title` in header context as it was already:
> 
> ```python
> if self._survey_title:
>             context["title"] = self._survey_title
> ```
> 
> Do you mean I introduced default for the title in header context (and the default was taken from somewhere else before)?

Ah okay. So the duplicate title setting is an existing issue, therefore, happy to leave it here.

tests/app/helpers/test_template_helpers.py
```diff
@@ -103,86 +103,115 @@ def test_footer_context_census_nisra_theme(app: Flask, expected_footer_nisra_the
...
+    (
+        (
+            SurveyType.BUSINESS,
+            BusinessSurveyConfig(),
+            {
+                "orgLogo": "ons-logo-en",
+                "orgLogoAlt": "Office for National Statistics logo",
+                "title": "ONS Business Surveys",
+            },
+        ),
```
MebinAbraham, line 117: We need a test that show a non default title working. All these tests asserts the defaults only. If the cookie session has a `survey_title`, then that is what should be rendered.

petechd: I've added couple variants with `'survey_title` in cookie present, should cover the scenarios.

MebinAbraham (approved):

tests/app/helpers/test_template_helpers.py
```diff
@@ -103,86 +103,162 @@ def test_footer_context_census_nisra_theme(app: Flask, expected_footer_nisra_the
...
+):
     with app.app_context():
-        survey_config = CensusNISRASurveyConfig()
+        for cookie_name, cookie_value in {
+            "theme": theme,
+            "survey_title": survey_title,
+        }.items():
+            if cookie_value:
+                cookie_session[cookie_name] = cookie_value
+        config = survey_config
```
berroar, line 255: Don't think this is needed, can just use `survey_config`

berroar (approved):
Looks good, just the one minor comment

tests/app/helpers/test_template_helpers.py
petechd: Done, thanks!

PR #911, opened 2022-08-01
Add conditional links in footer

petechd:
This adds conditional links in footer based on the presence of cookie (theme).

### How to review 
Added extra tests for  survey configs with no cookie and updated the ones were cookie theme key is supposed to be present now. Check the tests. 

### Checklist


petechd:
[Conditional links: Footer (1 SP)](https://trello.com/c/aTyrw8zh/5377-conditional-links-footer-1-sp)

app/helpers/template_helpers.py
```diff
@@ -144,9 +144,21 @@ def footer_context(self) -> dict[str, Any]:
...
+            if self._survey_type:
+                context["rows"] = [{"itemsList": self._survey_config.footer_links}]
+
+            else:
+                context["rows"] = [
+                    {
+                        "itemsList": [
+                            footer_link
+                            for footer_link in self._survey_config.footer_links
+                            if "Contact us" not in footer_link.values()
```
berroar, line 156: Could potentially do this the other way around so it is more explicit, i.e check for the values we want rather than the not in?

petechd: So I inverted the logic and now we see the fields that we want to keep:
```python
for footer_link in self._survey_config.footer_links:
      items_list.extend(
          footer_link
          for item in [
              "What we do",
              "Accessibility",
              "Help",
              "Languages",
              "BSL and audio videos",
          ]
          if item in footer_link.values()
      )

      context["rows"] = [{"itemsList": items_list}]
```
Thanks!

app/helpers/template_helpers.py
```diff
@@ -144,9 +144,28 @@ def footer_context(self) -> dict[str, Any]:
...
-            context["rows"] = [{"itemsList": self._survey_config.footer_links}]
-
-        if self._survey_config.footer_legal_links:
+            if self._survey_type:
+                context["rows"] = [{"itemsList": self._survey_config.footer_links}]
+
+            else:
+                items_list: list[Mapping] = []
+
+                for footer_link in self._survey_config.footer_links:
```
MebinAbraham, line 153: The survey config class should be responsible for returning the links it needs instead of it being handled externally by template helpers. Instead of filtering out stuff we don't need, we should populate with what we need from the get go, which you can, if the logic was handled by the survey config object. I would convert the attributes to methods which gets passed in a `bool` to whether you have a cookie theme, similar to how `get_service_links` is done.

petechd: Ok, moved all the links generation similar to `get_service_links`, added extra test for Welsh census for coverage, have a look!

app/survey_config/business_config.py
```diff
@@ -101,6 +85,38 @@ def get_service_links(
...
+                ).__dict__,
+            ]
+
+        return [
+            Link(lazy_gettext("What we do"), self.what_we_do_url).__dict__,
+            Link(
+                lazy_gettext("Accessibility"),
+                self.accessibility_url,
+            ).__dict__,
+        ]
```
MebinAbraham, line 106: Good to avoid the duplication. Applies to the other suvery configs also.
```suggestion
    def get_footer_links(self, cookie_has_theme: bool) -> list[dict]:
        links = [Link(lazy_gettext("What we do"), self.what_we_do_url).__dict__]

        if cookie_has_theme:
            links.append(Link(lazy_gettext("Contact us"), self.contact_us_url).__dict__)

        links.append(
            Link(
                lazy_gettext("Accessibility"),
                self.accessibility_url,
            ).__dict__
        )

        return links
```

petechd: Removed duplication from all configs.

app/survey_config/census_config.py
```diff
@@ -71,6 +35,54 @@ def get_data_layer(self, tx_id: Optional[str] = None) -> list[dict]:
...
+                    lazy_gettext("Privacy and data protection"),
+                    f"{EN_BASE_URL}/privacy-and-data-protection/",
+                ).__dict__,
+                Link(
+                    lazy_gettext("Terms and conditions"),
+                    f"{EN_BASE_URL}/terms-and-conditions/",
+                ).__dict__,
+            ]
+
+        return []
```
MebinAbraham, line 84: Minor, but the types for these methods among the classes should be consistent. Either all should be Optional list of dict, or just list of dict. If you are returning an empty list, then you don't need the optional bit. Alternately and my preference is unless you do a list operation on the values that get returned (it doesn't look like it), I would return None.

tests/app/helpers/test_template_helpers.py
```diff
@@ -17,49 +17,54 @@
...
+    expected_footer_social_theme,
+    expected_footer_social_theme_no_cookie,
+)
 
 DEFAULT_URL = "http://localhost"
 
 
-def test_footer_context_census_theme(app: Flask, expected_footer_census_theme):
+@pytest.mark.parametrize(
+    "expected_footer, theme, survey_config",
```
MebinAbraham, line 35: (Non-blocker) Minor, it would be better for `expected_footer` to be the last arg, like in most other cases. So it is more like `GIVEN a survey WHEN I have theme and suvery_config THEN I get the expected footer.

MebinAbraham (approved):

app/survey_config/census_config.py
petechd: My understanding was empty list is evaluated as `None` but yeah, I will return `None` explicitly and make it Optional where it applies.

MebinAbraham: Ya, it does, but if we don't do any list operation on it, then we should just return None.

tests/app/helpers/test_template_helpers.py
petechd: Ok, added, thanks!

berroar (approved):

PR #917, opened 2022-08-16
Remove metadata from session

petechd:
This removes metadata fields in Session Data (and authenticator) that were previously kept for compatibility between versions. This will reduce datastore payload size and is no longer need. Some tests had to removed and some where kept with slightly different assertions/descriptions.

### How to review 
Use existing tests, there might be some references to metadata in session that where overlooked.

### Checklist


petechd:
[Remove metadata from session (2 SP)](https://trello.com/c/tQIr6cgs/5398-remove-metadata-from-session-2-sp)

berroar (approved):

MebinAbraham (commented):
There is still a bit of code that sets display address on session data that should be removed. https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/views/handlers/submission.py#L62

app/data_models/session_data.py
```diff
@@ -4,47 +4,11 @@
...
-        account_service_base_url: Optional[str] = None,
-        account_service_log_out_url: Optional[str] = None,
-        trad_as: Optional[str] = None,
-        display_address: Optional[str] = None,
         confirmation_email_count: int = 0,
         feedback_count: int = 0,
-        schema_url: Optional[str] = None,
-        survey_url: Optional[str] = None,  # pylint: disable=unused-argument
         **_: Any,
     ):  # pylint: disable=too-many-locals
```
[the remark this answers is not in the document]
petechd: It was causing a lot of trouble because it was needed by that https://github.com/ONSdigital/eq-questionnaire-runner/blob/master/app/views/handlers/submission.py#L62 method but now possible to remove, thanks!

tests/app/data_model/test_session_store.py
```diff
@@ -197,7 +168,8 @@ def test_legacy_load(app_session_store_encoded):
     )
 
     assert (
-        session_store.session_data.tx_id == app_session_store_encoded.session_data.tx_id
+        session_store.session_data.confirmation_email_count
+        == app_session_store_encoded.session_data.confirmation_email_count
     )
```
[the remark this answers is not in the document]
petechd: Ok, we discussed it and next best thing is comparing both session data objects using `__dict__` or `vars()`. Added, thanks!

tests/app/data_model/test_session_store.py
```diff
@@ -215,7 +187,8 @@ def test_load(app_session_store_encoded):
         app_session_store_encoded.session_id,
     )
     assert (
-        session_store.session_data.tx_id == app_session_store_encoded.session_data.tx_id
```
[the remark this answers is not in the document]
petechd: Same as above, thanks!

app/data_models/session_data.py
```diff
@@ -1,50 +1,13 @@
...
-        case_ref: Optional[str] = None,
-        account_service_base_url: Optional[str] = None,
-        account_service_log_out_url: Optional[str] = None,
-        trad_as: Optional[str] = None,
-        display_address: Optional[str] = None,
         confirmation_email_count: int = 0,
         feedback_count: int = 0,
-        schema_url: Optional[str] = None,
-        survey_url: Optional[str] = None,  # pylint: disable=unused-argument
-        **_: Any,
```
MebinAbraham, line 25: This needs to stay. This is what makes it forwards compatible.

petechd: Ok, tests removal reverted as well. Cheers!

tests/app/data_model/test_session_store.py
```diff
@@ -133,13 +131,13 @@ def test_session_store_stores_trading_as_value_if_present(
 
         session_store = SessionStore("user_ik", "pepper", "eq_session_id")
 
-        assert hasattr(session_store.session_data, "trad_as") is True
+        assert hasattr(session_store.session_data, "language_code") is True
```
MebinAbraham, line 134: Existing but check the value?. This would be true even if value didn't exist as it defaults to None.

MebinAbraham (approved):

tests/app/data_model/test_session_store.py
petechd: I've reworded them so it makes sense now, this test and the one below.

MebinAbraham: My point is that the test on its own isnt very useful without checking the value. It will have the attr because it always does, but it doesn't check the attr get set properly.

petechd: Ok, now it tests if language code is being created from `session_data` and added as `session_store.session_data` instead of just the existence of it. Thanks!

PR #920, opened 2022-08-18
Update DS version and resolve DAC summary row title issue

rmccar:
Updates to the latest version of the DS (v58.0.0) to fix a DAC issue where summaries aren't marked up correctly which is causing issues for screenreaders.

Also removed the tests for table headers because we are no longer using tables for summaries so there will be no headers set.

### How to review 
Test the `test_confirmation_question` schema and check row titles are now styled correctly and not bold and still look and work as expected

### Checklist


rmccar:
[Should be using `groups` for summary of multiple answers to single question](https://trello.com/c/SBGYAYRE/5380-should-be-using-groups-for-summary-of-multiple-answers-to-single-question)

petechd:
I can see rowTitles are being nested correctly now but it still doesn't seem like they are being styled correctly, so they are still marked up with `<div>` tags instead of a `<h3>`.

rmccar:
> I can see rowTitles are being nested correctly now but it still doesn't seem like they are being styled correctly, so they are still marked up with `<div>` tags instead of a `<h3>`.

Have you managed to work out what your issue was? Because it is still working correctly for me

petechd:
Ok, I got it working now, disregard my previous comment!

petechd (approved):

app/jinja_filters.py
```diff
@@ -541,7 +541,6 @@ def __init__(
         edit_link_text: str,
         edit_link_aria_label: str,
     ) -> None:
-        self.rowTitle = strip_tags(question["title"])
```
MebinAbraham, line 544: Why is the question title being removed...?. If you have a question with multiple answers then you will never see the question title with this change.

MebinAbraham (changes requested):
Changes should have tests. Also highlight that there is no functional test for the title at the moment when using multiple answers.

rmccar:
[Refactor summaries to use definition lists (3 SP)](https://trello.com/c/SBGYAYRE/5380-refactor-summaries-to-use-definition-lists-3-sp)

templates/layouts/configs/_save-sign-out-button.html
```diff
@@ -24,11 +21,8 @@
                 "data-ga": "click",
                 "data-ga-category": "Navigation",
                 "data-ga-action": "Exit click"
-            },
-            "iconType": "exit",
-            "iconPosition": 'after'
```
MebinAbraham, line 29: How is this handled now that it is removed?.

app/views/contexts/calculated_summary_context.py
```diff
@@ -70,11 +68,6 @@ def build_view_context_for_calculated_summary(self, current_location):
                 "title": block.get("title") % dict(total=formatted_total),
                 "collapsible": block.get("collapsible", False),
                 "summary_type": "CalculatedSummary",
-                "headers": [
```
MebinAbraham, line 73: How comes headers are removed throughout?. How does the DS handle this now?. How will it have context?.

templates/layouts/configs/_save-sign-out-button.html
rmccar: These params have now been removed from the header macro in the DS. The save and sign out button is no longer customisable so setting them no longer have any affect. This is the PR that the change was made in: https://github.com/ONSdigital/design-system/pull/2438

PR #922, opened 2022-08-24
Check that operators are of type Mapping

berroar:

In some areas in runner we were checking whether a routing rule has a valid operator, but not checking that it is of type Mapping. As a result, routing on backward journeys for confirmation questions sometimes was resulting in an error.

Adds a method to check that the operator is a Mapping type when evaluating routing rules.

### How to review 

Check routing in the `test_new_confirmation_question` schema is not broken (routing backward from the confirmation question).

Check that nowhere else needs to be updated.

### Checklist


berroar:
[New routing rules: Missing check that operators are Mappings](https://trello.com/c/rGuP2STh/5406-new-routing-rules-missing-check-that-operators-are-mappings)

tests/app/questionnaire/test_questionnaire_schema.py
```diff
@@ -771,3 +775,22 @@ def test_when_rules_section_dependencies_calculated_summary(
...
+    (
+        ([], False),
+        ("This is a string", False),
+        ({"key": "value"}, False),
+        (
+            {"invalid-operator": ({"source": "answers", "identifier": "answer"}, 123)},
+            False,
+        ),
+        ({"==": ({"source": "answers", "identifier": "answer"}, 123)}, True),
+        ({">": ({"source": "answers", "identifier": "answer"}, 123)}, True),
```
petechd, line 791: We could add some kind of similar scenario to what actually happened, but not essential change.
```suggestion
        ({">": ({"source": "answers", "identifier": "answer"}, 123)}, True),
        ({"or": ({"source": "answers", "identifier": "answer"}, "No I need to correct this")}, True),
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -400,11 +398,11 @@ def _is_list_name_in_rule(
                 )
 
             # Nested rules
-            if any(operator in rule for operator in OPERATION_MAPPING):
+            if has_operator(rule):
                 return self._is_list_name_in_rule(rule, list_name)
 
     @staticmethod
-    def get_operands(rules: Mapping) -> list:
+    def get_operands(rules: Union[Mapping, Sequence]) -> list:
```
MebinAbraham, line 405: How come this changed?. When would this be a Sequence, as the code below would fail if this was?

MebinAbraham (commented):
Just some minor comments, otherwise LGTM.

app/questionnaire/questionnaire_schema.py
berroar: I can't seem to get around this one, not sure why it thinks there is a Sequence being passed to this method, but I get `Argument 1 to "get_operands" of "QuestionnaireSchema" has incompatible type "Union[Mapping[Any, Any], Sequence[Any]]"; expected "Mapping[Any, Any]"` in mypy

app/questionnaire/questionnaire_schema.py
```diff
@@ -885,9 +881,15 @@ def _get_rules_section_dependencies(
...
-            if any(operator in rule for operator in OPERATION_MAPPING):
+            if has_operator(rule):
                 rules_section_dependencies.update(
                     self._get_rules_section_dependencies(current_section_id, rule)
                 )
 
         return rules_section_dependencies
+
+
+def has_operator(rule: Any) -> bool:
```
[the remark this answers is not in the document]
MebinAbraham: I would make it a static method of questionnaire schema. The `get_operands` lives there already. Alternatively they both live in rules.helper as it isn't anything to do with schema really.

app/questionnaire/questionnaire_schema.py
MebinAbraham: Mypy can't work out this would only get called with a Mapping. But this method shouldn't change because it technically couldn't be a sequence anyway as it would fall over. 

I think for this reason, we should put the `isinstance` check inline as well which would fix the issue, i.e: 
```py
if isinstance(rules, Mapping) and has_operator(rules):
   do_stuff()
```

berroar: Updated this 👍

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
berroar: Moved this now 👍

app/questionnaire/rules/operations.py
```diff
@@ -191,12 +192,10 @@ def _resolve_self_reference(
...
         for operand in operands:
-            if isinstance(operand, dict) and any(
-                operator in operand for operator in OPERATION_MAPPING
-            ):
-                operator_name = next(iter(operand))
+            if has_operator(operand):
+                operator_name = next(iter(operand))  # type: ignore
                 resolved_nested_operands = self._resolve_self_reference(
-                    self_reference_value, operand[operator_name]
+                    self_reference_value, operand[operator_name]  # type: ignore
```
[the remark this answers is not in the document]
berroar: Resolved the type hint issue now 👍

app/questionnaire/rules/rule_evaluator.py
```diff
@@ -81,10 +81,8 @@ def _resolve_operand(
         if isinstance(operand, dict) and "source" in operand:
             return self.value_source_resolver.resolve(operand)
 
-        if isinstance(operand, dict) and any(
-            operator in operand for operator in OPERATION_MAPPING
-        ):
-            return self._evaluate(operand)
+        if has_operator(operand):
+            return self._evaluate(operand)  # type: ignore
```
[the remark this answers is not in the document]
berroar: Resolved the type hint issue now 👍

app/questionnaire/path_finder.py
```diff
@@ -303,21 +302,22 @@ def _remove_current_blocks_answers_for_backwards_routing(
...
         self, rules: dict, answer_ids_for_current_block: list[str]
     ) -> None:
-        operands = self.schema.get_operands(rules)
-        for rule in operands:
-            if isinstance(rule, dict) and (
-                "identifier" in rule
-                and rule["identifier"] in answer_ids_for_current_block
-            ):
-                if (
+        if isinstance(rules, Mapping) and QuestionnaireSchema.has_operator(rules):
```
MebinAbraham, line 305: How come this is at this level?. It doesn't match what we had before.

berroar: Removed this now

app/questionnaire/rules/operations.py
```diff
@@ -191,17 +191,15 @@ def _resolve_self_reference(
...
-                resolved_nested_operands = self._resolve_self_reference(
-                    self_reference_value, operand[operator_name]
-                )
-                resolved_value = getattr(self, OPERATION_MAPPING[operator_name])(
-                    *resolved_nested_operands
-                )
-
+            if QuestionnaireSchema.has_operator(operand):
+                operator_name = next(iter(operand))  # type: ignore
+                if isinstance(operand, dict):
```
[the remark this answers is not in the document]
berroar: This was only added to get around the type hinting issue

app/questionnaire/path_finder.py
```diff
@@ -303,21 +302,22 @@ def _remove_current_blocks_answers_for_backwards_routing(
...
         self, rules: dict, answer_ids_for_current_block: list[str]
     ) -> None:
-        operands = self.schema.get_operands(rules)
-        for rule in operands:
-            if isinstance(rule, dict) and (
-                "identifier" in rule
-                and rule["identifier"] in answer_ids_for_current_block
-            ):
-                if (
+        if QuestionnaireSchema.has_operator(rules):
```
MebinAbraham, line 305: This still doesn't make sense. Rules is typed as a dict, this check would break nested rules. The previous code was not doing this at this level, so how come it has been introduced now?.

app/questionnaire/rules/operations.py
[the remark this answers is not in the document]
MebinAbraham: This check should be on the first if. This line isn't needed, the check should be chained like it was previously. That should still keep mypy happy.

app/questionnaire/path_finder.py
berroar: Ok removed this 👍

app/questionnaire/rules/operations.py
```diff
@@ -191,17 +191,14 @@ def _resolve_self_reference(
     ) -> list[Union[ValueSourceTypes, date]]:
         resolved_operands = []
         for operand in operands:
-            if isinstance(operand, dict) and any(
-                operator in operand for operator in OPERATION_MAPPING
-            ):
-                operator_name = next(iter(operand))
+            if isinstance(operand, dict) and QuestionnaireSchema.has_operator(operand):
+                operator_name = next(iter(operand))  # type: ignore
```
MebinAbraham, line 195: Is there ignore needed anymore?

MebinAbraham (approved):

rmccar (approved):

app/questionnaire/questionnaire_schema.py
```diff
@@ -380,9 +380,7 @@ def get_post_submission(self) -> ImmutableDict:
     def _is_list_name_in_rule(
         self, rules: Union[Mapping, Sequence], list_name: str
     ) -> bool:
-        if isinstance(rules, Mapping) and any(
-            operator in rules for operator in OPERATION_MAPPING
-        ):
+        if isinstance(rules, Mapping) and QuestionnaireSchema.has_operator(rules):
```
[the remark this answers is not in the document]
berroar: As discussed will leave this for now 👍

rmccar:
[Release: 4.9.2](https://trello.com/c/M2yyXshE/5421-release-492)

PR #927, opened 2022-09-06
Update id name and remove unneeded button param

rmccar:
The `name` param for buttons is only used for buttons it sets the `name` attribute in the html for the button. The signout link is a link that is styled to look like a button and in that case the `name` param isn't used because there is no `name` attribute for links, therefore this param doesn't need to be set and can be removed.

The other change is just a fix to the format of an id in a schema.

### How to review 
- Tests pass
- Signout button still works as expected

### Checklist


rmccar:
[Remove unneeded param from signout button and fix id format](https://trello.com/c/JTR0TWuR/5418-remove-unneeded-param-from-signout-button-and-fix-id-format)

MebinAbraham (approved):

berroar (approved):

PR #930, opened 2022-09-16
Fix percentage symbol missing on input

rmccar:
Percentage symbol is missing from the suffix on `test_percentage` schema

Screenshot 2022-09-16 at 15 04 47

### How to review 
Open the `test_percentage` schema and see that the percentage symbol suffix is displayed

### Checklist

### DIFF
--- a/templates/partials/answers/percentage.html
+++ b/templates/partials/answers/percentage.html
@@ -18,7 +18,8 @@
   },
   "suffix": {
     "id": answer.id ~ "-type",
-    "title": "%"
+    "title": "Percent",
+    "text": "%"
   },
   "dontWrap": true if mutuallyExclusive else false,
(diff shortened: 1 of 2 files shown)


rmccar:
[Percentage suffix missing](https://trello.com/c/hJhAB5Ie/5422-percentage-suffix-missing)

berroar (commented):
Since this is a bug prob worth adding a test to make sure that this is always displayed.

rmccar:
> Since this is a bug prob worth adding a test to make sure that this is always displayed.

There was a test but it was falsely passing because one of the attributes was set to "%", Ive fixed it now so that it will only pass if the contents of the element is "%"

berroar (approved):

petechd (approved):

PR #931, opened 2022-09-21
Enable GA tracking on timeout modal

rmccar:
To track users being timed out we needed to supply the survey id and form type. The modal open and closing is now being tracked GA side using show and hide triggers. Because this metadata is only available while the session is still active and once the user is signed out this data is then lost, this PR adds that data to the cookies and then that data is read from the cookie into the data layer for GA to pick up. That way this data is always available to be analysed.

- This has been tested with the service design using GA and they are happy that they can get the data they want in GA in order to track this

### How to review 
- Tests pass
- Schemas launch as expected and new metadata is in the HTML head in dataLayer

### Checklist


rmccar:
[Enable GA tracking on modals](https://trello.com/c/qieSbSj9/5424-enable-ga-tracking-on-modals)

berroar (dismissed):

app/survey_config/survey_config.py
```diff
@@ -71,14 +71,7 @@ def get_footer_legal_links(  # pylint: disable=unused-argument, no-self-use
...
-        if self.schema:
-            data_layer.append(
-                {
-                    key: self.schema.json[key]
-                    for key in ["form_type", "survey_id", "title"]
-                    if key in self.schema.json
-                }
-            )
-        return data_layer
+    def get_additional_data_layer_context(  # pylint: disable=no-self-use
```
petechd, line 74: Why `self` in here when not used? Why create a method that always returns empty list (also wrong return type)?

app/helpers/language_helper.py
```diff
@@ -35,8 +35,8 @@ def handle_language(metadata: Optional[MetadataProxy] = None) -> None:
                 schema = load_schema_from_metadata(
                     metadata=metadata, language_code=request_language
                 )
-                if schema.json["title"] != cookie_session.get("survey_title"):
-                    cookie_session["survey_title"] = schema.json["title"]
+                if schema.json["title"] != cookie_session.get("title"):
```
petechd, line 38: Can we not keep the 'survey_title' instead of  changing  it to 'title'? What is the motivation for this change?

app/survey_config/survey_config.py
rmccar: No I think this is correct, it will return the empty list because it is only used to return something when used in census but here we need to return nothing

app/helpers/language_helper.py
rmccar: This is the title in the cookie and has a different name to differentiate from the `survey_title` in the survey config

app/survey_config/survey_config.py
petechd: Yeah it is complicated type hint since we overwrite it in census config, we probably need to keep it as it is now.

petechd (approved):

app/survey_config/survey_config.py
```diff
@@ -71,14 +71,7 @@ def get_footer_legal_links(  # pylint: disable=unused-argument, no-self-use
...
+        self,
+    ) -> list[dict]:
```
berroar, line 76: I would just have this as list since only the child class method adds a dict?
```suggestion
    ) -> list:
```

rmccar: It does here but when the census equivalent is used it returns a list of dicts so I thought it would be confusing if they had different return types?

rmccar: I understand your argument though because the actual function does just return an empty list

berroar: Not sure, I don't think it's confusing so long as the type hint for the census config is correct, as anything that uses this parent class would only ever return an empty list. Won't block it on this anyway

rmccar: Ok I'm happy to make the change

berroar (approved):

PR #933, opened 2022-09-23
Fix session expired page links

petechd:
This fixes the way we render `session-expired` page that uses `401.html` template. The correct way is to use the `_render_error_page()` method for all error pages (that creates all the required vars) instead of ordinary `render_template()`.

### How to review 
Two new assertions added that check if required urls exists.

### Checklist


petechd:
[Fix session expired page links](https://trello.com/c/YrdTchaU/5425-fix-session-expired-page-links)

MebinAbraham:
This can go into the main branch. Don't need a bug fix branch.

rmccar (approved):

berroar (approved):

PR #936, opened 2022-09-27
Fix h1 DAC accessibility issue

rmccar:
On survey introductions there is currently no h1 being set and this is an accessibility issue as there should be a h1 on every page. This PR fixes that, and also makes the h3s on the page h2s which is also recommended by DAC.

### How to review 
- Check the `test_introductions` schema and check that the top title is now a h1.
- Check the panel title and the bottom title are now a h2
- Compare this with master where they are a h2 and h3 respectively

### Checklist


rmccar:
[H1 missing from multiple pages](https://trello.com/c/14r0W9g8/5387-h1-missing-from-multiple-pages)

templates/introduction.html
```diff
@@ -12,31 +12,33 @@
...
-      {% endif %}
-
-      {% if content.block.secondary_content %}
-          {% for content_block in content.block.secondary_content %}
-              {% include 'partials/introduction/basic.html' %}
-          {% endfor %}
-      {% endif %}
+        {% if content.block.primary_content %}
+            {% for content_block in content.block.primary_content %}
+                {% set titleTag = 'h1' %}
```
petechd, line 17: Since we use snake case throughout these files I wouldn't start introducing camel case.

rmccar: Done

petechd (approved):

PR #937, opened 2022-09-28
Fix double h1 on thank you page DAC issue

rmccar:
This was flagged up in DAC testing, there shouldn't be more than one h1 on a page. This PR removes the second h1 on the thank you page and makes it a h2 instead.

### How to review 
Check thank you page using the `test-view-submitted-response` schema and see that now there is only one h1 on the page

### Checklist


rmccar:
[“Get a copy of your answers” on confirmation page should be ‘h2’ not ‘h1’. (1 SP)](https://trello.com/c/gmueUsOz/5383-get-a-copy-of-your-answers-on-confirmation-page-should-be-h2-not-h1-1-sp)

petechd (approved):

PR #939, opened 2022-09-28
Fix test interstitial page title schema

rmccar:
Fixes a bug with the `test_interstitial_page_title` schema which errors when loaded because it currently needs a `case_id` set and then that `case_id` is used in the title. This removes that requirement and also updates the content to make more sense and use `ru_ref` instead.

### How to review 
`test_interstitial_page_title` schema launches without error

### Checklist


rmccar:
[Fix test_interstitial_page_title schema error](https://trello.com/c/8PBqv40T/5428-fix-testinterstitialpagetitle-schema-error)

Yuyuutsu (approved):

petechd (approved):

PR #944, opened 2022-10-21
Djlint intergration

Yuyuutsu:
Investigate the appropriateness of DJLint ([HTML Template Linter and Formatter | djLint](https://www.djlint.com/)) for runner HTML/Jinja templates.

Djlint has problems reformatting long json inside an element so design system files have flattened code. The creator has said he is looking into fixing this. 

### How to review 

### Checklist


Yuyuutsu:
We will not be proceeding with Djlint. We have a big formatting error in Design System templates where it will flatten any json formatted code. This is an issue that has been raised with Djlint and is being looked into by the creator. As of right now It does more harm then good, but there will be big feature updates coming in the near future for Djlint (version 2.0) we shall look into it on a later date.

PR #948, opened 2022-11-01
Support for 'health' theme within runner 

Yuyuutsu:
For PHM, we will be using the `health` theme given it is a health surveys. 
- Changed health theme from using default `SurveyConfig` class to `SocialSurveyConfig`
- Added a new test schema called `test_theme_health`
- Added unit tests which asserts that health theme instantiates a `SocialSurveyConfig` instance

### How to review 
- Test the `test_theme_health` schema in runner 
- `make test` and `make test-unit` pass

### Checklist


Yuyuutsu:
[Support for 'health' theme within runner (1 SP)](https://trello.com/c/rUb0gVsx/5447-support-for-health-theme-within-runner-1-sp)

tests/app/helpers/test_template_helpers.py
```diff
@@ -127,6 +127,26 @@ def test_footer_warning_not_in_context_census_theme(app: Flask):
...
+        (
+            SurveyType.HEALTH,
+            "Test",
+            SocialSurveyConfig(),
+            {
+                "orgLogo": "ons-logo-en",
+                "orgLogoAlt": "Office for National Statistics logo",
+                "title": "Test",
+            },
+        ),
```
MebinAbraham, line 149: I wouldn't bother with this. For now given health uses the `SocialSurveyConfig`, just have a single test that ensures when initialising a health theme, its instance is a `SocialSurveyConfig` instance (use the `isinstance` check).

Yuyuutsu: Would `test_get_survey_config` (line 590) check if it is an instance of `SocialSurveyConfig` if I change the value of the parameter to `(SurveyType.HEALTH, "en", SocialSurveyConfig)`

MebinAbraham: Ya, HEALTH is already defined for that test, which should currently fail. That would just need updating.

petechd: OK, added what you've wanted now! Thanks

schemas/test/en/test_theme_health.json
```diff
@@ -0,0 +1,75 @@
...
+            "type": "string"
+        },
+        {
+            "name": "period_id",
+            "type": "string"
+        },
+        {
+            "name": "ru_name",
+            "type": "string"
+        }
```
berroar, line 22: Minor but we should prob use social claims for this schema, like `test_theme_social`  (i know questionnaire_id will be renamed soon) e.g. 
```suggestion
        {
            "name": "case_ref",
            "type": "string"
        },
        {
            "name": "questionnaire_id",
            "type": "string"
        },
```

berroar (approved):

tests/app/helpers/test_template_helpers.py
```diff
@@ -813,3 +813,14 @@ def test_include_csrf_token(app: Flask, include_csrf_token: bool):
...
+
+def test_health_survey_config(app: Flask):
+    with app.app_context():
+        survey_config = get_survey_config(
+            theme=SurveyType.HEALTH,
+            schema=QuestionnaireSchema({}),
+            language="en",
+        )
+
+    assert isinstance(survey_config, SocialSurveyConfig)
```
MebinAbraham, line 826: This isn't needed. A test to check in instance already existing. You just need to update it https://github.com/ONSdigital/eq-questionnaire-runner/blob/support-health-theme/tests/app/helpers/test_template_helpers.py#L579-L580

petechd: Oops, I didn't realise we check instance, updated tests now and removed the new one, thanks!

MebinAbraham (approved):

MebinAbraham (changes requested):
Actually we need a change

MebinAbraham (approved):
Fine for now actually as our runner default is BUSINESS.

PR #954, opened 2022-11-07
Evaluate calculated summary is complete when returning to a calculated summary

berroar:
Fixes a bug where users were seeing a 404 page when attempting to return to a calculated summary page. This was because we were attempting to return to the calculated summary page before visiting dependent questions.

Updated the logic so that we only return to the calculated summary page when all dependent questions are complete.

Added a test with a new `test_calculated_summary_dependent_questions` schema which emulates the journey that resulted in the bug.

Updated bug in journeys in the existing calculated summary tests.

### How to review 

Check that you see any dependent questions before returning to a calculated summary page, and that a `404` page is no longer displayed.

Use the `test_calculated_summary_dependent_questions`:
Complete all questions -> Use the change link on the second question on the calculated summary -> Update the value -> You should be taken to the dependent question before routing back to the calculated summary (previously saw a 404 error page)

### Checklist

### DIFF
--- a/app/questionnaire/router.py
+++ b/app/questionnaire/router.py
@@ -219,15 +219,20 @@ class Router:
         if not return_to:
             return None
 
-        if return_to == "calculated-summary":
-            if return_to_block_id in routing_path:
-                return url_for(
-                    "questionnaire.block",
-                    list_item_id=location.list_item_id,
-                    block_id=return_to_block_id,
-                    _anchor=return_to_answer_id,
-                )
-            return None
+        if (
+            return_to == "calculated-summary"
+            and return_to_block_id in routing_path
+            and self.can_access_location(
+                Location(block_id=return_to_block_id, section_id=location.section_id),
+                routing_path,
+            )
+        ):
+            return url_for(
+                "questionnaire.block",
+                list_item_id=location.list_item_id,
+                block_id=return_to_block_id,
+                _anchor=return_to_answer_id,
+            )
 
         if is_section_complete is None:
             is_section_complete = self._progress_store.is_section_complete(
--- a/app/questionnaire/router.py
+++ b/app/questionnaire/router.py
@@ -219,13 +219,13 @@ class Router:
         if not return_to:
             return None
 
-        if (
-            return_to == "calculated-summary"
-            and return_to_block_id in routing_path
-            and self.can_access_location(
-                Location(block_id=return_to_block_id, section_id=location.section_id),
-                routing_path,
-            )
+        if return_to == "calculated-summary" and self.can_access_location(
+            Location(
+                block_id=return_to_block_id,
+                section_id=location.section_id,
+                list_item_id=location.list_item_id,
+            ),
+            routing_path,
         ):
             return url_for(
                 "questionnaire.block",
--- a/schemas/test/en/test_calculated_summary_dependent_questions.json
+++ b/schemas/test/en/test_calculated_summary_dependent_questions.json
@@ -8,9 +8,7 @@
     "description": "A schema to showcase Calculated Summary with dependent questions.",
     "questionnaire_flow": {
         "type": "Linear",
-        "options": {
-
-        }
+        "options": {}
     },
     "sections": [
         {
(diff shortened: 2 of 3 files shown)


berroar:
[404 error when routing back to calculated summary](https://trello.com/c/cRBeXHUi/5468-404-error-when-routing-back-to-calculated-summary)

Yuyuutsu (approved):
Test works and could not recreate 404 error when returning back to a calculated summary page

app/questionnaire/router.py
```diff
@@ -219,15 +219,20 @@ def _get_return_to_location_url(
...
-                return url_for(
-                    "questionnaire.block",
-                    list_item_id=location.list_item_id,
-                    block_id=return_to_block_id,
-                    _anchor=return_to_answer_id,
-                )
-            return None
+        if (
+            return_to == "calculated-summary"
+            and return_to_block_id in routing_path
```
MebinAbraham, line 224: This check isn't needed as it will be covered by the `can_access_location` check.

app/questionnaire/router.py
```diff
@@ -219,15 +219,20 @@ def _get_return_to_location_url(
...
+            and self.can_access_location(
+                Location(block_id=return_to_block_id, section_id=location.section_id),
```
MebinAbraham, line 226: Should add the list item id.

schemas/test/en/test_calculated_summary_dependent_questions.json
```diff
@@ -0,0 +1,193 @@
...
+                            "question": {
+                                "id": "question-4",
+                                "title": "Of the money spent on clothes, how much did you spend on shoes?",
+                                "type": "General",
+                                "answers": [
+                                    {
+                                        "id": "answer-4",
+                                        "mandatory": true,
+                                        "type": "Currency",
+                                        "label": "Capital expenditure",
```
MebinAbraham, line 117: These all need to be updated, the schema doesn't quite make sense. Even test schema should be easy to digest. They all have the same labels, so Calculated summary looks odd.

schemas/test/en/test_calculated_summary_dependent_questions.json
```diff
@@ -0,0 +1,193 @@
...
+        },
+        {
+            "name": "ref_p_start_date",
+            "type": "date"
+        },
+        {
+            "name": "ref_p_end_date",
+            "type": "date"
+        }
+    ],
```
MebinAbraham, line 181: Removed things you don't need.

tests/app/questionnaire/test_router.py
```diff
@@ -539,7 +538,50 @@ def test_return_to_calculated_summary(self):
...
-            _anchor="first-number-answer",
+            return_to="calculated-summary",
+            return_to_answer_id="first-number-answer",
+            return_to_block_id="currency-total-playback-skipped-fourth",
+        )
+
+        assert expected_location_url == next_location_url
+
+    @pytest.mark.usefixtures("app")
+    def test_return_to_calculated_summary_with_dependent_questions(self):
```
MebinAbraham, line 549: A comment or renaming this would help such as `test_return_to_calculated_summary_not_on_allowable_path`.

MebinAbraham (commented):
Looks good, just a few minor things.

app/questionnaire/router.py
berroar: Removed now 👍

app/questionnaire/router.py
berroar: Added 👍

schemas/test/en/test_calculated_summary_dependent_questions.json
berroar: Updated the labels now 👍

schemas/test/en/test_calculated_summary_dependent_questions.json
berroar: Removed some parts of the schema 👍

tests/app/questionnaire/test_router.py
berroar: Renamed now 👍

MebinAbraham (approved):

PR #957, opened 2022-11-10
Handling double submission error

Yuyuutsu:
 GCS upload is a complete upload and doesn't allow partially uploaded data (corrupted). Since no corrupted data is being uploaded we can handle the double submission error on the runner. 

### How to review 
- Unit Test Pass
- Check if the unit test works as expecteds
### Checklist


Yuyuutsu:
[Update runner to handle submission error to allow duplicates to be ignored (2 SP)](https://trello.com/c/V7qoRYML/5463-update-runner-to-handle-submission-error-to-allow-duplicates-to-be-ignored-2-sp)

app/submitter/submitter.py
```diff
@@ -39,7 +40,14 @@ def send_message(self, message: str, tx_id: str, case_id: str) -> bool:
...
         # However, this behaviour was deemed acceptable for our use case.
-        blob.upload_from_string(str(message).encode("utf8"), retry=DEFAULT_RETRY)
+        # Treats doubble submission error as successful because GCS doesn't allow partial data to be uploaded.
+        try:
+            blob.upload_from_string(str(message).encode("utf8"), retry=DEFAULT_RETRY)
+        except Forbidden as e:
+            if "storage.objects.delete" in e.message:
+                logger.info("Double Submission found")
+            else:
+                logger.debug("Forbidden Error")
```
MebinAbraham, line 50: If it was Forbidden and wasn't a delete operation, then we want to raise the error as at that point it is an unexpected error.
```suggestion
            
            raise
```

app/submitter/submitter.py
```diff
@@ -39,7 +40,14 @@ def send_message(self, message: str, tx_id: str, case_id: str) -> bool:
 
         # DEFAULT_RETRY is not idempotent.
         # However, this behaviour was deemed acceptable for our use case.
-        blob.upload_from_string(str(message).encode("utf8"), retry=DEFAULT_RETRY)
+        # Treats doubble submission error as successful because GCS doesn't allow partial data to be uploaded.
+        try:
+            blob.upload_from_string(str(message).encode("utf8"), retry=DEFAULT_RETRY)
+        except Forbidden as e:
+            if "storage.objects.delete" in e.message:
+                logger.info("Double Submission found")
```
MebinAbraham, line 48: Maybe something like `Questionnaire submission exists, ignoring delete operation error` will be clearer for the log message as double submission is still possible if SDX deletes the object in time, i.e. retry strategy kicks in etc.
Also, worth adding a comment why were are doing this as it is not strictly obvious

app/submitter/submitter.py
```diff
@@ -39,7 +40,14 @@ def send_message(self, message: str, tx_id: str, case_id: str) -> bool:
 
         # DEFAULT_RETRY is not idempotent.
         # However, this behaviour was deemed acceptable for our use case.
-        blob.upload_from_string(str(message).encode("utf8"), retry=DEFAULT_RETRY)
+        # Treats doubble submission error as successful because GCS doesn't allow partial data to be uploaded.
```
MebinAbraham, line 43: This isn't strictly true. Maybe something like:
```
If an object exists then the GCS Client will attempt to delete the existing object before reuploading.
However, runner does not delete permission in an attempt to reduce duplicate receipts.
The first version of the object is acceptable as it is an extreme edge case for two submissions to contain different response data.
```

Also, I suggest adding this comment above the if statement instead of the entire block.

MebinAbraham (commented):
As discussed, need a unit test for this.

app/submitter/submitter.py
```diff
@@ -39,8 +40,18 @@ def send_message(self, message: str, tx_id: str, case_id: str) -> bool:
 
         # DEFAULT_RETRY is not idempotent.
         # However, this behaviour was deemed acceptable for our use case.
-        blob.upload_from_string(str(message).encode("utf8"), retry=DEFAULT_RETRY)
+        try:
+            blob.upload_from_string(str(message).encode("utf8"), retry=DEFAULT_RETRY)
+        except Forbidden as e:
+            # If an object exists then the GCS Client will attempt to delete the existing object before reuploading.
+            # However, runner does not delete permission in an attempt to reduce duplicate receipts.
```
petechd, line 47: "runner does not _**have**_ delete permission"?
"in an attempt to reduce duplicate receipts we didn't grant runner delete permission"?
You will know better anyway, just word missing.

tests/app/submitter/test_submitter.py
```diff
@@ -336,3 +338,58 @@ def test_gcs_feedback_submitter_uploads_feedback(patch_gcs_client):
...
+    bucket.blob.return_value = gcs_blob_storage_delete_error_message
+
+    published = gcs_submitter.send_message(
+        message={"test_data"}, tx_id="123", case_id="456"
+    )
+    # Then
+    assert published
+
+
+def test_forbidden_error_raised(
```
berroar, line 360: I feel like the test names here could be more descriptive so that it is easier to understand the scenarios being covered, i.e something like `test_double_submission_is_forbidden_when_not_delete_operation_error`?

tests/app/submitter/test_submitter.py
```diff
@@ -336,3 +338,57 @@ def test_gcs_feedback_submitter_uploads_feedback(patch_gcs_client):
...
+    with pytest.raises(Forbidden):
+        gcs_submitter.send_message(message={"test_data"}, tx_id="123", case_id="456")
+
+
+@pytest.fixture
+def gcs_blob_storage_error_message(mocker):
+    blob = Blob(name="test-blob", bucket=mocker.Mock())
+
+    blob.upload_from_string = mocker.Mock(  # pylint: disable=protected-access
+        side_effect=Forbidden("wrong-error")
```
MebinAbraham, line 380: For clarity, this should be something meaning ful. So maybe use `storage.objects.create`.

tests/app/submitter/test_submitter.py
```diff
@@ -336,3 +338,57 @@ def test_gcs_feedback_submitter_uploads_feedback(patch_gcs_client):
...
+    )
+
+    return blob
+
+
+@pytest.fixture
+def gcs_blob_storage_delete_error_message(mocker):
```
MebinAbraham, line 387: These return the blob, so fixure might be clearer as

```suggestion
def gcs_blob_delete_forbidden(mocker):
```

tests/app/submitter/test_submitter.py
```diff
@@ -336,3 +338,57 @@ def test_gcs_feedback_submitter_uploads_feedback(patch_gcs_client):
...
+    bucket = patch_gcs_client.return_value.get_bucket.return_value
+    bucket.blob.return_value = gcs_blob_storage_error_message
+
+    # Then
+    with pytest.raises(Forbidden):
+        gcs_submitter.send_message(message={"test_data"}, tx_id="123", case_id="456")
+
+
+@pytest.fixture
+def gcs_blob_storage_error_message(mocker):
```
MebinAbraham, line 376: These return the blob, so fixure might be clearer as
```suggestion
def gcs_blob_create_forbidden(mocker):
```

MebinAbraham (approved):

berroar (approved):

PR #965, opened 2022-11-25
Refactor calculated summary

berroar:

Adds support for a new way to format `CalculatedSummary` blocks in our schema json. This is required to in order to support future features such as GrandSummary. The new format supports the use of ValueSources in CalculatedSummary blocks.

Updated format:
```
"calculation": {
   "operation": {
       "+": [
           {
               "source": "answers",
               "identifier": "first-number-answer"
           },
           {
               "source": "answers",
               "identifier": "second-number-answer"
           },
           {
               "source": "answers",
               "identifier": "third-number-answer"
           }
       ]
   },
   "title": "Grand total of previous values"
},
```

Also adds the ability to use calculated summary in a repeating section.

### How to review 
Needs to be tested alongside the following validator [PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/158).

Check that pre existing calculated summary format has not been broken by this change and is still supported.

Ensure the new calculated summary format supports all the existing calculated summary scenarios. Use the `test_new_calculated_summary` schema.

Check that the new calculated summary format works in a repeating section, and all scenarios for a normal calculated summary are supported in a repeating section. Use the `test_new_calculated_summary_repeating_sections` schema.

### Checklist


berroar:
[Refactor calculated summary (5 SP)](https://trello.com/c/qVCAwIZQ/5350-refactor-calculated-summary-5-sp)

app/views/contexts/calculated_summary_context.py
```diff
@@ -162,18 +179,62 @@ def _get_formatted_total(self, groups, current_location, calculation_operator):
...
+            self._response_metadata,
+            location=current_location,
+        )
+
+        calculated_summary_total = evaluate_calculated_summary.evaluate(calculation)
+
+        return self._format_total(answer_format, calculated_summary_total)
+
+    @staticmethod
+    def _format_total(answer_format, total):
```
petechd, line 223: Since you are adding all new methods we should type hint them? So type hint the whole file since there isn't much here?

rmccar (approved):

app/views/contexts/calculated_summary_context.py
berroar: Added type hints to this module 👍

app/questionnaire/rules/operations.py
```diff
@@ -242,3 +242,11 @@ def evaluate_option_label_from_value(self, value: str, answer_id: str) -> str:
...
 
         return label
+
+    @staticmethod
+    def evaluate_sum(*args: tuple) -> Union[int, Decimal]:
+        values_list: list = []
+        for arg in args:
+            if arg and isinstance(arg, (int, Decimal)):
+                values_list.append(arg)
+        return sum(values_list)
```
MebinAbraham, line 252: We are a bit inconsistent in runner, most of the operations don't validate and expect the correct types.
However, happy to keep as it does guard if and when we pull from metadata etc.
```suggestion
        sum(value for value in args if isinstance(value, (int, float, Decimal)))
```

app/questionnaire/rules/rule_evaluator.py
```diff
@@ -15,7 +16,7 @@
     ValueSourceTypes,
 )
 
-RuleEvaluatorTypes = Union[bool, Optional[date], list[str], list[date]]
+RuleEvaluatorTypes = Union[bool, Optional[date], list[str], list[date], int, Decimal]
```
MebinAbraham, line 19: I would add floats as that could be possible from static value or metadata etc

app/questionnaire/value_source_resolver.py
```diff
@@ -127,13 +128,25 @@ def _resolve_calculated_summary_value_source(
...
+        evaluate_calculated_summary = rule_evaluator.RuleEvaluator(
+            self.schema,
+            self.answer_store,
+            self.list_store,
+            self.metadata,
+            self.response_metadata,
+            location=self.location,
+        )
+
+        return evaluate_calculated_summary.evaluate(calculation["operation"])  # type: ignore
```
MebinAbraham, line 149: I understand why this is done but it highlights a problem in design. We now have a RE > VSR > RE circular dependency. I suspect that why it has even been imported entirely?.

In a way it does give us the flexility to just pass to rule evaluator, but a bit of a code smell. Happy to be pragmatic and look at this when we have some time. In future or when it becomes a problem. It might become one when you do the answer only on path fix.

app/views/contexts/calculated_summary_context.py
```diff
@@ -30,78 +35,112 @@ def build_groups_for_section(self, section, return_to_block_id):
...
         ]
 
-    def build_view_context_for_calculated_summary(self, current_location):
-        block = self._schema.get_block(current_location.block_id)
+    def build_view_context_for_calculated_summary(
+        self, current_location: Location
+    ) -> Mapping:
+        if (block_id := current_location.block_id) and (
+            block := self._schema.get_block(block_id)
+        ):
```
MebinAbraham, line 51: If this was done just for type hinting I suggest removing it, we are making the code less readable for type hints. You can override it as it shouldn't get this far if the block was invalid.

app/views/contexts/calculated_summary_context.py
```diff
@@ -162,21 +203,70 @@ def _get_formatted_total(self, groups, current_location, calculation_operator):
...
+                    current_location=current_location,
+                )
+                for answer in question["answers"]:
+                    if not answer_format["type"]:
+                        answer_format = {
+                            "type": answer["type"],
+                            "unit": answer.get("unit"),
+                            "unit_length": answer.get("unit_length"),
+                            "currency": answer.get("currency"),
+                        }
```
MebinAbraham, line 231: Avoid duplicating this, this is the main bit, so ideally 1 source. I suggest a single method and you can just check if it is legacy or new using `if "calculation_type" in calculation:` I suggest keep the existing method, but the end just either call RE or the current `sum()`

app/views/contexts/calculated_summary_context.py
```diff
@@ -162,21 +203,70 @@ def _get_formatted_total(self, groups, current_location, calculation_operator):
...
+            self._list_store,
+            self._metadata,
+            self._response_metadata,
+            location=current_location,
+        )
+
+        # type ignore as rule evaluator will only return an int or decimal in this instance
+        calculated_summary_total: Union[int, Decimal] = evaluate_calculated_summary.evaluate(calculation)  # type: ignore
+
+        return self._format_total(answer_format, calculated_summary_total)
```
MebinAbraham, line 245: This can stay separate. Helps with readibility.

tests/app/questionnaire/test_router.py
```diff
@@ -646,6 +647,145 @@ def test_return_to_calculated_summary_return_to_block_id_not_on_path(self):
             == next_location_url
         )
 
+    @pytest.mark.usefixtures("app")
```
MebinAbraham, line 650: Do you need all these test?. These all exist for the old format, so you can just parametrise the schema?. I think you made the IDs between the two the same, so apart from the schema that is loaded, it should be identical?.

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -534,6 +534,57 @@ def test_calculated_summary_value_source(mocker, list_item_id):
...
+    schema.get_block = Mock(
+        return_value={
+            "id": "number-total",
+            "type": "CalculatedSummary",
+            "calculation": {
+                "operation": {
+                    "+": [
+                        {"source": "answers", "identifier": "number-answer-1"},
+                        {"source": "answers", "identifier": "number-answer-2"},
+                    ]
```
MebinAbraham, line 552: The new operator should be tested as its own thing also. Same as if we were to add any other new operators. Need to test a recursive one also.

tests/functional/generate_pages.py
```diff
@@ -813,9 +813,16 @@ def process_block(
...
-                block["calculation"]["answers_to_calculate"], page_spec
-            )
+            if block["calculation"].get("answers_to_calculate"):
+                process_calculated_summary(
+                    block["calculation"]["answers_to_calculate"], page_spec
+                )
+            else:
+                answer_ids_for_block = [
+                    value["identifier"]
+                    for value in block["calculation"]["operation"]["+"]
```
MebinAbraham, line 823: Same here about, not relying on first level of operator.

tests/functional/spec/features/new_calculated_summary.spec.js
```diff
@@ -0,0 +1,331 @@
...
+import PercentageTotalPlaybackPage from "../../generated_pages/calculated_summary/percentage-total-playback.page.js";
+import NumberTotalPlaybackPage from "../../generated_pages/calculated_summary/number-total-playback.page.js";
+import CalculatedSummaryTotalConfirmation from "../../generated_pages/calculated_summary/calculated-summary-total-confirmation.page.js";
+import SubmitPage from "../../generated_pages/calculated_summary/submit.page";
+import ThankYouPage from "../../base_pages/thank-you.page.js";
+
+describe("Feature: Calculated Summary", () => {
+  describe("Given I have a Calculated Summary", () => {
+    before("Get to Calculated Summary", () => {
+      browser.openQuestionnaire("test_new_calculated_summary.json");
```
MebinAbraham, line 23: Make this a test case like the time-out modal ones to avoid the duplicate test?. This and the legacy format tests are identical apart from the schema loaded right?.

MebinAbraham (commented):
We may see a few changes when we come to support the "only using answers on the path" work. For the most part, it looks good.

app/questionnaire/rules/operations.py
berroar: Updated this 👍

app/questionnaire/rules/rule_evaluator.py
berroar: Added float 👍

tests/functional/spec/features/new_calculated_summary.spec.js
berroar: Updated this 👍

tests/app/questionnaire/test_value_source_resolver.py
berroar: Add tests for the operator and a recursive test 👍

tests/functional/generate_pages.py
berroar: Updated this 👍

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -534,6 +534,108 @@ def test_calculated_summary_value_source(mocker, list_item_id):
...
+    schema = mocker.MagicMock()
+    schema.get_block = Mock(
+        return_value={
+            "id": "number-total",
+            "type": "CalculatedSummary",
+            "calculation": {
+                "operation": {
+                    "+": [
+                        {"+": [{"source": "answers", "identifier": "number-answer-1"}]},
+                        {"+": [{"source": "answers", "identifier": "number-answer-2"}]},
```
MebinAbraham, line 602: This isn't valid. "+" should receive at least 2 items. So could be:
```suggestion
                        {"+": [{"source": "answers", "identifier": "number-answer-1"}, {"source": "answers", "identifier": "number-answer-2"}]},
                        {"source": "answers", "identifier": "number-answer-3"},
```

Which translates to `(1 + 2) + 3`.

app/questionnaire/questionnaire_schema.py
```diff
@@ -888,3 +894,18 @@ def _get_rules_section_dependencies(
                 )
 
         return rules_section_dependencies
+
+
+def get_calculated_summary_answer_ids(calculated_summary_block: Mapping) -> list:
```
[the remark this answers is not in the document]

app/views/contexts/calculated_summary_context.py
berroar: Updated now 👍

app/views/contexts/calculated_summary_context.py
```diff
@@ -30,78 +35,101 @@ def build_groups_for_section(self, section, return_to_block_id):
...
                 "groups": groups,
                 "answers_are_editable": True,
                 "calculated_question": self._get_calculated_question(
                     calculation, formatted_total
                 ),
-                "title": block.get("title") % dict(total=formatted_total),
-                "collapsible": block.get("collapsible", False),
+                "title": block_title % dict(total=formatted_total)
+                if block_title
+                else None,
```
MebinAbraham, line 84: Title is mandatory for calculated summary.

app/views/contexts/calculated_summary_context.py
```diff
@@ -30,78 +35,101 @@ def build_groups_for_section(self, section, return_to_block_id):
...
-        ]
-        unique_blocks = list(
-            {block["id"]: block for block in blocks_to_calculate}.values()
-        )
-
-        for block in unique_blocks:
-            if QuestionnaireSchema.is_question_block_type(block["type"]):
-                transformed_block = self._remove_unwanted_questions_answers(
-                    block, answers_to_calculate, current_location=current_location
+        if (block_id := current_location.block_id) and (
```
MebinAbraham, line 94: Same here, this indentation seems to be for the sake of type hints. Should be removed.

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -534,6 +534,116 @@ def test_calculated_summary_value_source(mocker, list_item_id):
...
+            },
+        },
+    )
+
+    if list_item_id:
+        location = Location(
+            section_id="test-section", block_id="test-block", list_item_id="item-1"
+        )
+    else:
+        location = None
```
MebinAbraham, line 563: You should still have a location?. Just the `list_item_id` should be `None`.

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -534,6 +534,116 @@ def test_calculated_summary_value_source(mocker, list_item_id):
...
```
MebinAbraham, line 619: Same as above.

app/views/contexts/calculated_summary_context.py
MebinAbraham: Still outstanding.

app/views/contexts/calculated_summary_context.py
berroar: Updated 👍

app/views/contexts/calculated_summary_context.py
berroar: Updated 👍

tests/app/questionnaire/test_value_source_resolver.py
berroar: Updated 👍

tests/app/questionnaire/test_value_source_resolver.py
berroar: Updated 👍

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -534,6 +534,124 @@ def test_calculated_summary_value_source(mocker, list_item_id):
...
+
+    location = (
+        Location(
+            section_id="test-section", block_id="test-block", list_item_id="item-1"
+        )
+        if list_item_id
+        else Location(
+            section_id="test-section", block_id="test-block", list_item_id=None
+        )
+    )
```
MebinAbraham, line 566: You have the list item already so don't need the `if`?
```suggestion
    location = Location(
    	section_id="test-section", block_id="test-block", list_item_id=list_item_id
    )
```

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -534,6 +534,124 @@ def test_calculated_summary_value_source(mocker, list_item_id):
...
```
MebinAbraham, line 627: Same as above.

tests/app/questionnaire/test_value_source_resolver.py
berroar: Fixed 👍

tests/app/questionnaire/test_value_source_resolver.py
berroar: Fixed

app/views/contexts/calculated_summary_context.py
```diff
@@ -30,65 +35,85 @@ def build_groups_for_section(self, section, return_to_block_id):
...
+            )
+
         blocks_to_calculate = [
             self._schema.get_block_for_answer_id(answer_id)
             for answer_id in answers_to_calculate
         ]
+
         unique_blocks = list(
-            {block["id"]: block for block in blocks_to_calculate}.values()
+            {block["id"]: block for block in blocks_to_calculate if block}.values()
```
MebinAbraham, line 112: For what reason was the `if block` introduced?. Would a block not exist?

app/views/contexts/calculated_summary_context.py
```diff
@@ -30,65 +35,85 @@ def build_groups_for_section(self, section, return_to_block_id):
...
         )
 
         for block in unique_blocks:
-            if QuestionnaireSchema.is_question_block_type(block["type"]):
+            if block and QuestionnaireSchema.is_question_block_type(block["type"]):
```
MebinAbraham, line 116: Same, here why is the `if block` needed.

app/views/contexts/calculated_summary_context.py
berroar: Bad mypy fix - sorted it now 😅

app/views/contexts/calculated_summary_context.py
berroar: As above 👍

schemas/test/en/test_new_calculated_summary_repeating_section.json
```diff
@@ -0,0 +1,852 @@
...
+                                        ],
+                                        "type": "Radio"
+                                    }
+                                ],
+                                "id": "proxy-question",
+                                "title": "Are you answering the questions on behalf of someone else?",
+                                "type": "General"
+                            },
+                            "type": "Question"
+                        },
```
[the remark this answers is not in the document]
berroar: Removed now 👍

MebinAbraham (commented):
As discussed, dependent section completeness is not working with a repeating calculated summary.

berroar:
> As discussed, dependent section completeness is not working with a repeating calculated summary.

Updated the code with your suggestion as discussed. Fixed existing and updated the unit tests with additional parameters to test inside a repeat. Also added a new functional test for a repeating calculated summary.

app/views/contexts/calculated_summary_context.py
MebinAbraham: Happy this is resolved.

MebinAbraham (approved):

PR #966, opened 2022-11-28
Add language toggle support for survey title

petechd:
Survey title in header (semantic html `header` element) was not translated when switching the language as it was not updated in the session cookie. This PR fixes that issue.

### How to review 
Extra checks added in functional test so just double check if that is correct and satisfying.

### Checklist

### DIFF
--- a/app/helpers/language_helper.py
+++ b/app/helpers/language_helper.py
@@ -31,9 +31,9 @@ def handle_language(metadata: Optional[MetadataProxy] = None) -> None:
         g.allowed_languages = get_allowed_languages(schema_name, launch_language)
         request_language = request.args.get("language_code")
         if request_language and request_language in g.allowed_languages:
-            if cookie_session.get("survey_title"):
+            if metadata:
                 schema = load_schema_from_metadata(
-                    metadata=metadata, language_code=request_language  # type: ignore
+                    metadata=metadata, language_code=request_language
                 )
                 cookie_session["survey_title"] = schema.json["title"]
             session_store.session_data.language_code = request_language
--- a/app/helpers/language_helper.py
+++ b/app/helpers/language_helper.py
@@ -2,12 +2,13 @@ from typing import Optional, Union
 from urllib.parse import urlencode
 
 from flask import g, request
+from flask import session as cookie_session
 from flask_login import current_user
 
 from app.data_models.metadata_proxy import MetadataProxy
 from app.globals import get_metadata, get_session_store
 from app.questionnaire.questionnaire_schema import DEFAULT_LANGUAGE_CODE
-from app.utilities.schema import get_allowed_languages
+from app.utilities.schema import get_allowed_languages, load_schema_from_metadata
 
 LANGUAGE_TEXT = {
     "en": "English",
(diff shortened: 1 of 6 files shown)


petechd:
[Bug: Survey title does not support language toggle](https://trello.com/c/emHMCugI/5483-bug-survey-title-does-not-support-language-toggle)

Yuyuutsu (approved):

app/helpers/language_helper.py
```diff
@@ -30,6 +31,11 @@ def handle_language(metadata: Optional[MetadataProxy] = None) -> None:
         g.allowed_languages = get_allowed_languages(schema_name, launch_language)
         request_language = request.args.get("language_code")
         if request_language and request_language in g.allowed_languages:
+            if cookie_session.get("survey_title"):
```
MebinAbraham, line 34: When would we not have a survey title at this point?. I would just check that we have metadata then proceed.

app/helpers/language_helper.py
```diff
@@ -30,6 +31,11 @@ def handle_language(metadata: Optional[MetadataProxy] = None) -> None:
...
+                schema = load_schema_from_metadata(
+                    metadata=metadata, language_code=request_language  # type: ignore
```
MebinAbraham, line 36: What is the type ignore for?.

petechd: I was checking few different solutions here and possibly forgot to remove it after changing the code. Removed now, thanks!

app/helpers/language_helper.py
```diff
@@ -30,6 +31,11 @@ def handle_language(metadata: Optional[MetadataProxy] = None) -> None:
         g.allowed_languages = get_allowed_languages(schema_name, launch_language)
         request_language = request.args.get("language_code")
         if request_language and request_language in g.allowed_languages:
+            if metadata:
+                schema = load_schema_from_metadata(
+                    metadata=metadata, language_code=request_language
+                )
+                cookie_session["survey_title"] = schema.json["title"]
```
MebinAbraham, line 38: The latest flask login changes mean the cookie is only sent when the cookie session is modified, therefore to avoid sending the cookie when it hasn't changed, i.e. page refresh etc, the title in the cookie session should only be set if it doesn't match what is set in there. A cookie is sent even if you set the value to the same existing value.

```suggestion
                if schema.json["title"] != cookie_session.get("survey_title"):
                    cookie_session["survey_title"] = schema.json["title"]
```

app/helpers/language_helper.py
```diff
@@ -30,6 +31,12 @@ def handle_language(metadata: Optional[MetadataProxy] = None) -> None:
         g.allowed_languages = get_allowed_languages(schema_name, launch_language)
         request_language = request.args.get("language_code")
         if request_language and request_language in g.allowed_languages:
+            if metadata:
+                schema = load_schema_from_metadata(
+                    metadata=metadata, language_code=request_language
+                )
+                if schema.json["title"] != cookie_session.get("survey_title"):
+                    cookie_session["survey_title"] = schema.json["title"]
```
MebinAbraham, line 39: Newline would help readability.

MebinAbraham (approved):

PR #970, opened 2022-12-01
Upgrade python 3.10.8

Yuyuutsu:
- Upgraded all `python versions`to`3.10.8` as well as upgraded all other packages. 
- `Flake8 `is pinned to` 4.0.1 `because Flake8 versions `5.x.x `is not compatible with `flake8-mock `
- Updated `unit-tests` in `test-errors` due to new flask-login update where session is only sent to the client when there is a change.
- Added some type hinting
- Benchmark results are on [Add python 3.10.8 benchmark results](https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/120)
- Updated pipenv and ubuntu in YAML files
- Removed thread.local from application, due to deprecation 
- Changed FLASK_ENV to FLASK_DEBUG because ENV is deprecated
### How to review
Run runner locally and check if it is working as it should
Check changes to type hinting if they are relevant, as well as comments regarding` mypy ignore`
Check `test-errors` unit test changes
Check benchmark results on the 
Check if `FLASK_DEBUG` is configured correctly

### Checklist


Yuyuutsu:
Have to revert changes to an older branch due to merge conflicts and keyboard acting up

Yuyuutsu:
[Upgrade runner to Python 3.10 (2 SP)](https://trello.com/c/33ZeJcSt/5228-upgrade-runner-to-python-310-2-sp)

.pylintrc
```diff
@@ -1,421 +1,614 @@
-[MASTER]
+[MAIN]
```
berroar, line 1: Is this supposed to be Master?

Yuyuutsu: Changed to Master 👍

Pipfile
```diff
@@ -72,9 +72,10 @@ google-cloud-tasks = "*"
 simplejson = "*"
 markupsafe = "*"
 pdfkit = "*"
+typing-extensions = "*"
```
MebinAbraham, line 75: This is a dev dependency. Is it needed?.

tests/integration/session/test_multiple_login.py
```diff
@@ -13,6 +13,7 @@ def setUp(self):
 
         self.cache = {}
 
+    # pylint: disable=arguments-renamed
```
MebinAbraham, line 16: I suggest fixing the issue.

app/forms/validators.py
```diff
@@ -417,8 +417,8 @@ def __call__(
             condition = f"{conditions[0]} or equals"
         else:
             condition = conditions[0]
-
-        is_valid, message = self._is_valid(condition, total, target_total)
+        # is_valid never returns a "None" object. Therefore a none object being iterated will never occur
+        is_valid, message = self._is_valid(condition, total, target_total)  # type: ignore[misc]
```
MebinAbraham, line 421: What is the use of `[misc]`

application.py
```diff
@@ -14,7 +14,7 @@
 
 def configure_logging():
     log_level = logging.INFO
-    debug = os.getenv("FLASK_ENV") == "development"
+    debug = os.getenv("FLASK_DEBUG")
```
MebinAbraham, line 17: Need to ensure this doesn't evaluate `"0"` as debug, currently it would.

tests/app/forms/field_handlers/test_address_handler.py
```diff
@@ -6,8 +6,11 @@
...
 
 
-def get_test_form_class(
-    answer_schema, value_source_resolver, rule_evaluator, messages=error_messages.copy()
+def get_test_form_class(  # pylint: disable=dangerous-default-value
+    answer_schema,
+    value_source_resolver,
+    rule_evaluator,
+    messages=error_messages.copy(),
 ):
```
MebinAbraham, line 14: I would resolve this.

Should just be:

```py
def get_test_form_class(
    answer_schema,
    value_source_resolver,
    rule_evaluator,
    messages=None,
):

  if not messages:
    messages = error_messages.copy()
```

tests/app/forms/field_handlers/test_number_handler.py
```diff
@@ -13,7 +13,7 @@
 # wtforms Form parents are not discoverable in the 2.3.3 implementation
 
 
-def get_test_form_class(
+def get_test_form_class(  # pylint: disable=dangerous-default-value
```
MebinAbraham, line 16: Same as above.

tests/app/questionnaire/rules/test_rule_evaluator.py
```diff
@@ -854,7 +854,9 @@ def test_answer_source_default_answer_used_when_no_answer(
 def test_raises_exception_when_bad_operand_type():
     with pytest.raises(TypeError):
         rule_evaluator = get_rule_evaluator()
-        rule_evaluator.evaluate(rule={Operator.EQUAL: {1, 1}})
+        rule_evaluator.evaluate(
+            rule={Operator.EQUAL: {1, 1}}  # pylint: disable=duplicate-value
```
MebinAbraham, line 858: The issue should be fixed.

tests/app/forms/field_handlers/test_address_handler.py
Yuyuutsu: Made the suggested changes 👍

tests/app/forms/field_handlers/test_number_handler.py
Yuyuutsu: Made the suggested changes 👍

tests/app/questionnaire/rules/test_rule_evaluator.py
Yuyuutsu: Updated, please check if the change I made was what is expected with this test.

application.py
Yuyuutsu: Updated `debug = os.getenv("FLASK_DEBUG")== "1"

app/forms/validators.py
Yuyuutsu: Mypy performs numerous other, less commonly failing checks that don’t have specific error codes. They are labeled as miscs. This one suggested` is_valid` fails to iterate through a `None object`, but I believe `is_valid `should never be None.

Pipfile
Yuyuutsu: GIthub actions was failing without me adding this. I will re-try without it and see if this is needed or not.

tests/integration/session/test_multiple_login.py
Yuyuutsu: Does this need to be changed? T because `schema_name="test_textfield"` is a default value, which also means that `client` has to be before it, which pylint is thinking that `schema_name` has been renamed to` client`

Yuyuutsu: Removed the `schema_name` default value and swapped it with the `client`. This has removed the pylint error.

app/forms/validators.py
Yuyuutsu: It only occurred when I was adding Optional type hinting due to it being also "potentially" a None I can have a look through to see if it can be fixed.

Yuyuutsu:
Getting new type hinitng errors because this repo had to merge with the eq-runner. So i am currently updating them

berroar:
When I run locally I get this warning:  `DeprecationWarning: structlog.threadlocal is deprecated, please use structlog.contextvars instead`, referring to `context_class=wrap_dict(dict)` in application.py - so may need updating?

tests/integration/session/test_multiple_login.py
Yuyuutsu: Oops forgot to update the tests.

tests/app/questionnaire/rules/test_rule_evaluator.py
Yuyuutsu: Updated to `rule={Operator.EQUAL: {1, 2}}`. Thanks for the suggestion 👍

tests/integration/session/test_multiple_login.py
berroar: Can `# pylint: disable=arguments-renamed` be removed now?

app/jinja_filters.py
```diff
@@ -603,7 +603,9 @@ def map_summary_item_config(
                 item_label=block.get("item_label"),
                 item_anchor=block.get("item_anchor"),
             )
-            rows.extend(list_collector_rows)
+
+            # map_list_collector_config returns List[Dict[str, List[Any]]] but mypy expects Iterable[SummaryRow]
```
berroar, line 607: Doesn't this suggest that the return type for this function is incorrect i.e it won't only be a list of SummaryRows?

Yuyuutsu: Yeah, I tried `-> Union[SummaryRows,  list[Dict[str, List[Any]]]]` but I still get that error. I'll have another look 👍

tests/integration/session/test_multiple_login.py
Yuyuutsu: Removed 👍

berroar:
Ran a couple of benchmark tests using this branch and although no big performance hit it looks like response times for are slightly higher (5-10ms) on average for the 50th to 95th percentiles 👍

Pipfile
Yuyuutsu: Removed `typing-extensions` and GA are passing 👍

app/forms/validators.py
Yuyuutsu: I cannot seem to find a solution so I believe in keeping `type ignore[misc]`. We should leave`[misc]` specifically because mypy can still check for other typing errors on that line, and with new updates to mypy certain `misc` errors will get their own specific error message.

Yuyuutsu:
Going to do Baseline benchmarks and then creating a PR on EQ-Runner-Benchmark

app/jinja_filters.py
Yuyuutsu: Removed the type ignore. Pete and I seemed to find that if you specifically type that `rows : list = []`, resolves the issue.

tests/integration/integration_test_case.py
```diff
@@ -167,7 +167,8 @@ def get(self, url, follow_redirects=True, **kwargs):
         :param url: the URL to GET
         """
         response = self._client.get(url, follow_redirects=follow_redirects, **kwargs)
-
+        if last_cookie is not None:
+            response.headers["Set-Cookie"] = last_cookie
```
MebinAbraham, line 171: Can we not store the last cookie on the instance?. That way we don't have to pass it into the "get" method.

app/jinja_filters.py
```diff
@@ -577,7 +577,7 @@ def map_summary_item_config(
     icon: Optional[str] = None,
 ) -> list[SummaryRow]:
 
-    rows = []
+    rows: list = []
```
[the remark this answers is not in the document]
Yuyuutsu: Changing that results into mypy raising an error `Argument 1 to "extend" of "list" has incompatible type "List[Dict[str, List[Any]]]"; expected "Iterable[SummaryRow]"`. I remember Pete and I tried different ways, like this but it just failing when mypy ran. Should I revert it back to ` rows: list = []`

app/questionnaire/path_finder.py
```diff
@@ -111,7 +111,7 @@ def _build_routing_path_block_ids(
         blocks: list[Mapping],
         current_location: Location,
         when_rules_block_dependencies: list[str],
-    ) -> list[str]:
+    ) -> Optional[list[str]]:
```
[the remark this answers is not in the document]
Yuyuutsu: `_def_build_routing_path_block_ids` is now explicitly returning `routing_path_block_ids` and reverted back to `-> list[str]`

app/questionnaire/placeholder_transforms.py
```diff
@@ -261,7 +261,7 @@ def format_ordinal(
 
         return f"{number_to_format}{indicator}"
 
-    def get_ordinal_indicator(self, number_to_format: int) -> str:
+    def get_ordinal_indicator(self, number_to_format: int) -> Optional[str]:
```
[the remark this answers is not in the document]
Yuyuutsu: Typing has been reverted 👍

app/questionnaire/router.py
```diff
@@ -144,7 +145,9 @@ def get_next_location_url(
         # Due to backwards routing you can be on the last block of the path but with an in_progress section
         is_last_block_on_path = routing_path[-1] == location.block_id
         if is_last_block_on_path:
-            return self._get_first_incomplete_location_in_section(routing_path).url()
+            # This is never called if next_location = None so we have to type ignore
+            next_location: Location = self._get_first_incomplete_location_in_section(routing_path)  # type: ignore
```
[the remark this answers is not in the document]
Yuyuutsu: Comment Updated 👍

app/questionnaire/router.py
```diff
@@ -66,7 +66,8 @@ def get_first_incomplete_location_in_questionnaire_url(self) -> str:
         return self.get_next_location_url_for_end_of_section()
 
     def get_last_location_in_questionnaire_url(self) -> str:
-        routing_path = self.routing_path(*self._get_last_complete_section_key())
+        # List or Tuple is never going to be None value so will always be expected as a varadic arugment
+        routing_path = self.routing_path(*self._get_last_complete_section_key())  # type: ignore[misc]
```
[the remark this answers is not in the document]
Yuyuutsu: Changed to `Optional[str]`.

tests/integration/integration_test_case.py
Yuyuutsu: Its been a while since I made this change so I will have to have a look 👍

Yuyuutsu: I had a look, I don't think I can cause the header doesn't exist for the `Set-Cookie` unless you save it prior to the `getUrlCookie`. I am not entirely sure if thats what you meant , sorry 😅

app/forms/validators.py
MebinAbraham: This is a genuine warning. `_is_valid` could return None and this would result in an error. It never does at the moment because we don't have another other condition. 

This can be fixed by updating that method to raise a `NotImplementedError` and should always return the tuple.

app/jinja_filters.py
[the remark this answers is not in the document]
MebinAbraham: That's because `map_list_collector_config` return a a dict instead of summary row, so this and the return type needs to be `list[Union[dict[str, list], SummaryRow]]`

app/questionnaire/placeholder_transforms.py
```diff
@@ -287,6 +287,7 @@ def get_ordinal_indicator(self, number_to_format: int) -> str:
                 17: "eg",
                 19: "eg",
             }.get(number_to_format, "fed")
+        raise NotImplementedError
```
MebinAbraham, line 290: New line for readibility and text for context if the error were to happen.

```suggestion
        
        raise NotImplementedError(f"Language code '{self.language}' not implemented.")
```

app/questionnaire/router.py
```diff
@@ -66,7 +66,8 @@ def get_first_incomplete_location_in_questionnaire_url(self) -> str:
         return self.get_next_location_url_for_end_of_section()
 
     def get_last_location_in_questionnaire_url(self) -> str:
-        routing_path = self.routing_path(*self._get_last_complete_section_key())
+        # Expects a vardic argument but can be None
+        routing_path = self.routing_path(*self._get_last_complete_section_key())  # type: ignore[misc]
         return self.get_last_location_in_section(routing_path).url()
```
MebinAbraham, line 71: The comment should be about why we need to ignore it. The actual warning is genuine. This shouldn't error given it is currently only called from the submit handler (very end). But it would be better to fix this, if this was called by anything else, that wouldn't be true. Also this is used for `previous_location_url` which in the html handles not existing.

```py
    def get_last_location_in_questionnaire_url(self) -> Optional[str]:
        # Expects a vardic argument but can be None

        section_key = self._get_last_complete_section_key()
        if section_key:        
            routing_path = self.routing_path(*section_key)
            return self.get_last_location_in_section(routing_path).url()
```

Once this is changed, `get_previous_location_url` of submit questionnaires needs to change to `Optional[str]`.

app/questionnaire/router.py
```diff
@@ -144,7 +145,9 @@ def get_next_location_url(
         # Due to backwards routing you can be on the last block of the path but with an in_progress section
         is_last_block_on_path = routing_path[-1] == location.block_id
         if is_last_block_on_path:
-            return self._get_first_incomplete_location_in_section(routing_path).url()
+            # The section is not complete therefore we must have a location
```
MebinAbraham, line 148: Maybe we should prefix comment for type ignore.

i.e
```py
# Type ignore: The section is not complete therefore we must have a location
```

app/questionnaire/router.py
```diff
@@ -420,6 +423,7 @@ def get_next_block_url(
             block_id=next_block_id,
             list_name=routing_path.list_name,
             list_item_id=routing_path.list_item_id,
+            _external=False,
```
MebinAbraham, line 426: Was this needed from some breaking change?. Is there a link to the doc?.

tests/integration/integration_test_case.py
MebinAbraham: I mean store it on the instance like we do with `last_csrf_token` but when you do get a cookie, update it.

```py
class IntegrationTestCase(unittest.TestCase):  # pylint: disable=too-many-public-methods
    def setUp(self):
        # Cache for requests
        self.last_url = None
        self.last_response = None
        self.last_csrf_token = None
        self.redirect_url = None
        self.last_response_headers = None
```

So another attribute here. Then in `get(...)`, if the response header has `Set-Cookie` set `self.cookie` to the header cookie value.

app/questionnaire/router.py
Yuyuutsu: Yes, I believe we discussed this in a tech session. 
[Flask Version 2.2.x](https://flask.palletsprojects.com/en/2.2.x/api/#flask.Flask.url_for)
[Flask Version 2.0.x](https://flask.palletsprojects.com/en/2.0.x/api/#:~:text=flask.-,url_for,-(endpoint%2C)
The parameters have been reorganised with Flask 2.2.x .
I had to specifically set `_external=False ` otherwise any vardic argument  would be passed as through the `_external` param rather then `**kwargs`.

app/forms/validators.py
Yuyuutsu: Changes added, thanks for the help on this one. It makes a lot more sense now

app/questionnaire/questionnaire_schema.py
```diff
@@ -379,7 +379,7 @@ def get_post_submission(self) -> ImmutableDict:
 
     def _is_list_name_in_rule(
         self, rules: Union[Mapping, Sequence], list_name: str
-    ) -> bool:
+    ) -> Optional[bool]:
```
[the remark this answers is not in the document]
Yuyuutsu: Function is now explicitly returning False a the end 👍

app/questionnaire/router.py
Yuyuutsu: Sounds like a good idea, eaiser to understand what we are commenting. 👍  Added

app/questionnaire/router.py
Yuyuutsu: Changes added, also had to change `render_template` type hinting from `**kwargs: Union[str, Mapping]` to `**kwargs: Union[Optional[str], Mapping]`

tests/integration/integration_test_case.py
Yuyuutsu: Updated and comment added 👍

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
Yuyuutsu: Adding Unit test for the change.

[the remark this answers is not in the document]
Yuyuutsu: Struggling to add unit test for the explicit return. What conditions would have to be in the Schema for it to reach the explicit return r`outing_path_block_ids`

app/forms/validators.py
```diff
@@ -445,6 +445,7 @@ def _is_valid(
             return total >= target_total, "TOTAL_SUM_NOT_GREATER_THAN_OR_EQUALS"
         if condition == "less than or equals":
             return total <= target_total, "TOTAL_SUM_NOT_LESS_THAN_OR_EQUALS"
+        raise NotImplementedError("Condition is not implemented")
```
MebinAbraham, line 448:
```suggestion

        raise NotImplementedError(f"Condition '{condition}' is not implemented")
```

app/helpers/template_helpers.py
```diff
@@ -216,7 +216,7 @@ def get_survey_config(
     )
 
 
-def render_template(template: str, **kwargs: Union[str, Mapping]) -> str:
+def render_template(template: str, **kwargs: Union[Optional[str], Mapping]) -> str:
```
MebinAbraham, line 219: As per style guide, for more than 2 types it should be `Union[None, str, Mapping]`

tests/integration/session/test_multiple_login.py
```diff
@@ -13,7 +13,7 @@ def setUp(self):
 
         self.cache = {}
 
-    def launchSurvey(self, client, schema_name="test_textfield", **payload_kwargs):
+    def launchSurvey(self, schema_name, client, **payload_kwargs):
```
MebinAbraham, line 16: Any reason this was changed?. This is inconsistent with post where it passes a client as the first arg. My preference is to keep as it was.

MebinAbraham (approved):

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
MebinAbraham: It should ever happen, therefore happy for it to be ignored via `# pragma: no cover`.

app/helpers/template_helpers.py
Yuyuutsu: Change added to fit style guide 👍

tests/integration/session/test_multiple_login.py
Yuyuutsu: I forgot why, but changes have been reverted now 👍

ONS-Guilhem-Forey (approved):

PR #973, opened 2022-12-05
Fix mutually exclusive validation for zero as unanswered

Hajara-I:
This adds a fix to the mutually exclusive validation when one of the answer values is zero. Checks to make sure that answer value is not in the list of empty-answer-values. Previously '0' was evaluated as false hence total-answer was less than zero and raising a validation error. 

### How to review 
New test scenario added to verify that the the fix works

### Checklist


Hajara-I:
[Bug: Mutually exclusive validation treating zero as unanswered](https://trello.com/c/QItQMPYH/5486-bug-mutually-exclusive-validation-treating-zero-as-unanswered)

app/forms/validators.py
```diff
@@ -468,7 +468,13 @@ def __call__(
         is_mandatory: bool,
         is_only_checkboxes_or_radios: bool,
     ) -> None:
-        total_answered = sum(1 for value in answer_values if value)
+
+        empty_answer_values: tuple = (None, [], "", {})
```
berroar, line 472: We should reuse the `EMPTY_ANSWER_VALUES` from the QuestionnaireStoreUpdater class

tests/app/forms/validation/test_mutually_exclusive_validator.py
```diff
@@ -72,6 +72,7 @@ def test_mutually_exclusive_mandatory_answers_raise_validation_error(
         (["British, Irish", None], True),
         ([None, "I prefer not to say"], True),
         (["", "I prefer not to say"], True),
+        ([0, []], True),
```
MebinAbraham, line 75: I would demo using `Decimal` class also.

Hajara-I: Added Decimal class to unit test.

MebinAbraham (approved):

berroar (approved):

PR #983, opened 2022-12-19
Add BEIS logos

rmccar:
Add BEIS logos and add themes and tests for them

- Also removes a couple of Welsh tests for Northern Ireland which aren't needed

### How to review 
Test new schemas look as expected with new logos

### Checklist


rmccar:
[Importing UKIS logos to eQ (2 SP)](https://trello.com/c/XH2sOlhL/5517-importing-ukis-logos-to-eq-2-sp)

tests/app/helpers/test_template_helpers.py
```diff
@@ -216,6 +217,16 @@ def test_footer_warning_not_in_context_census_theme(app: Flask):
                 read_file("./templates/assets/images/ni-finance-mobile-logo.svg"),
             ],
         ),
+        (
+            None,
```
berroar, line 221: Should we use the type here? 
```suggestion
            SurveyType.BEIS,
```

rmccar: I was following the NI suff when doing this, does this mean you think we should add the equivalent for NI?

berroar: Just not clear to me why either are `None`, the normal business config seems to have a test for both

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=add-beis-orr-themes
```
berroar, line 2: Reminder to revert 👍

berroar (approved):
Looks good to me 👍

Yuyuutsu (approved):
LGTM!

PR #985, opened 2022-12-19
Make survey title generic [Bug]

Yuyuutsu:
The survey title is currently set to ONS Business Survey even when no cookie exists. This PR changes it so the generic title, which is "ONS Surveys" which no cookie exist

### How to review 
- Locally run runner
- Start any survey, and remove the session using developer tools
- Refresh the page and it should return an error page with the title "ONS Surveys"
- Check if unit tests pass 

### Checklist


Yuyuutsu:
[Bug: Survey title is not generic when no cookie exists](https://trello.com/c/o6tYkmvJ/5505-bug-survey-title-is-not-generic-when-no-cookie-exists)

rmccar (approved):

petechd (approved):

app/helpers/template_helpers.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         self._include_csrf_token = include_csrf_token
         self._survey_config = survey_config
         self._survey_title = cookie_session.get(
-            "title", self._survey_config.survey_title
+            "survey_title", lazy_gettext("ONS Surveys")
```
MebinAbraham, line 46: The title should only use the default when we have no cookie, "survey_title" shouldn't be overridden, this should still be "title". At the moment, all titles are suffixed with "ONS Surveys", but that should only be the case when we don't have a cookie.

MebinAbraham (changes requested):

app/helpers/template_helpers.py
Yuyuutsu: Ahh I think me changing `tests` to `survey_title` is why I have so many errors with the tests. The tests are working as they should now, just need to revert changes 👍

Yuyuutsu: Changes Reverted. Once this is resolved would it be okay to merge, since the only change made after the two approvals I previously got is changing `survey_title` to `title`.

petechd (approved):

rmccar (approved):

tests/app/helpers/test_template_helpers.py
```diff
@@ -820,17 +820,17 @@ def test_correct_theme_in_context(app: Flask, theme: str, language: str, expecte
...
+        (SurveyType.SOCIAL, "cy", "ONS Surveys"),
+        (SurveyType.NORTHERN_IRELAND, "en", "ONS Surveys"),
+        (SurveyType.BEIS, "en", "ONS Surveys"),
+        (SurveyType.ORR, "en", "ONS Surveys"),
+        (SurveyType.CENSUS, "en", "ONS Surveys"),
+        (SurveyType.CENSUS, "cy", "ONS Surveys"),
+        (SurveyType.CENSUS_NISRA, "en", "ONS Surveys"),
     ],
 )
 def test_correct_survey_title_in_context(
```
MebinAbraham, line 836: Worth renaming.

```suggestion
def test_use_default_survey_title_in_context_when_no_cookie(
```

MebinAbraham (approved):

PR #990, opened 2023-01-03
Update year in test to fix new year bug

rmccar:
Bug fix for functional tests Confirmation Question Within A Repeating Section now that the year is 2023. If the year in the test is longer than 15 years from the current date then the dob question will not route to the confirmation question once answered which breaks the test. The discussion on whether this should be a dynamic number has not come up with another solution and any solution to create the date dynamically could cause the test to be flakey.

### How to review 
Test now passes

### Checklist

### DIFF
--- a/app/translations/messages.pot
+++ b/app/translations/messages.pot
@@ -1,14 +1,14 @@
 # Translations template for PROJECT.
-# Copyright (C) 2022 ORGANIZATION
+# Copyright (C) 2023 ORGANIZATION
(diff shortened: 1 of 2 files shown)


rmccar:
[Bug: Functional test failing on year change](https://trello.com/c/wBVY9u79/5524-bug-functional-test-failing-on-year-change)

petechd (approved):

berroar (approved):

PR #998, opened 2023-01-12
Add BEIS-NI theme

rmccar:
Add BEIS-NI logos and add themes and tests for them

Also:
- Updates `mastheadLogo` and `mastheadLogoMobile` var names to `masthead_logo` and `masthead_logo_mobile` to conform with other var naming
- Updates DS version which fixes a bug where we need to provide a mobile version of the logo (in this case we don't need to because we cant to also use the regular logo for mobile)
- Update NI finance logo name to conform with BEIS name

### How to review 
Test new schema look as expected with new logos

### Checklist


rmccar:
[Create runner theme for NI BEIS](https://trello.com/c/ODxARf77/5533-create-runner-theme-for-ni-beis)

app/survey_config/__init__.py
```diff
@@ -22,6 +23,7 @@
     "BusinessSurveyConfig",
     "NorthernIrelandBusinessSurveyConfig",
     "BEISBusinessSurveyConfig",
+    "NIBEISBusinessSurveyConfig",
```
berroar, line 26: I know this is just copying the others, but now that there are more of these does it make sense to append `BusinessSurveyConfig` to all of these?  Might be clearer without it i.e just `NIBEIS`? Not sure what others think though. Also we now have a mismatch between `NI` and `NorthernIreland`

rmccar: With the NI bit do you think `NIBEISBusinessSurveyConfig` should be called `NorthernIrelandBEISBusinessSurveyConfig`? I just thought that was a bit long. Could change `NorthernIrelandBusinessSurveyConfig` to `NIBusinessSurveyConfig` though?

As for the `BusinessSurveyConfig` bit Im happy to reduce it to `SurveyConfig` or even `config`? I think removing it completely would make it a bit less clear what was being used here

berroar: Personally I'm fine with NI for these

MebinAbraham: I think you should flip them so it is `BEISNI` to match `CensusNisra`

rmccar: Done the NI stuff 👍

MebinAbraham (commented):
According to the Git viewer, it says the SVG is not valid https://raw.githubusercontent.com/ONSdigital/eq-questionnaire-runner/add-nibeis-theme/templates/assets/images/beis-ni-logo.svg

I ran it through a [SVG Viewer](https://www.svgviewer.dev/) which also says it has error.

MebinAbraham:
> > According to the Git viewer, it says the SVG is not valid [`add-nibeis-theme`/templates/assets/images/beis-ni-logo.svg (raw)](https://raw.githubusercontent.com/ONSdigital/eq-questionnaire-runner/add-nibeis-theme/templates/assets/images/beis-ni-logo.svg)
> > I ran it through a [SVG Viewer](https://www.svgviewer.dev/) which also says it has error.
> 
> Yeah its because its 2 svgs in one file. I can look into sorting that out but it does work like this when used

It essentially comes out as an error. Seems like we misusing it?. Could this possibly cause issues for screen reader or will the alt tags work as expected for both logos. It seems to render fine in runner but multiple other things fails to do so. How do we know it will work in all the browsers we need to support.

rmccar:
> 

Its just 2 svgs in one svg file, the way we use the file is just to read the html out of it and pass that to the component not use the file itself so it doesnt really matter. But I could possibly see if I can find a way of either combining the svgs in one file, reading 2 files in runner or making a change to the DS to allow for 2 logos

rmccar:
> > > According to the Git viewer, it says the SVG is not valid [`add-nibeis-theme`/templates/assets/images/beis-ni-logo.svg (raw)](https://raw.githubusercontent.com/ONSdigital/eq-questionnaire-runner/add-nibeis-theme/templates/assets/images/beis-ni-logo.svg)
> > > I ran it through a [SVG Viewer](https://www.svgviewer.dev/) which also says it has error.
> > 
> > 
> > Yeah its because its 2 svgs in one file. I can look into sorting that out but it does work like this when used
> 
> It essentially comes out as an error. Seems like we misusing it?. Could this possibly cause issues for screen reader or will the alt tags work as expected for both logos. It seems to render fine in runner but multiple other things fails to do so. How do we know it will work in all the browsers we need to support. Does the DS potentially need to support multiple SVGs?

Its just 2 svgs in one svg file, the way we use the file is just to read the html out of it and pass that to the component not use the file itself so it doesnt really matter. But I could possibly see if I can find a way of either combining the svgs in one file, reading 2 files in runner or making a change to the DS to allow for 2 logos

rmccar:
> > > > > According to the Git viewer, it says the SVG is not valid [`add-nibeis-theme`/templates/assets/images/beis-ni-logo.svg (raw)](https://raw.githubusercontent.com/ONSdigital/eq-questionnaire-runner/add-nibeis-theme/templates/assets/images/beis-ni-logo.svg)
> > > > > I ran it through a [SVG Viewer](https://www.svgviewer.dev/) which also says it has error.
> > > > 
> > > > 
> > > > Yeah its because its 2 svgs in one file. I can look into sorting that out but it does work like this when used
> > > 
> > > 
> > > It essentially comes out as an error. Seems like we misusing it?. Could this possibly cause issues for screen reader or will the alt tags work as expected for both logos. It seems to render fine in runner but multiple other things fails to do so. How do we know it will work in all the browsers we need to support. Does the DS potentially need to support multiple SVGs?
> > 
> > 
> > Its just 2 svgs in one svg file, the way we use the file is just to read the html out of it and pass that to the component not use the file itself so it doesnt really matter. But I could possibly see if I can find a way of either combining the svgs in one file, reading 2 files in runner or making a change to the DS to allow for 2 logos
> 
> In that case given we have both logos already, just concatenate it in code and remove the new file added?.

Done

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=add-nibeis-theme
```

MebinAbraham (approved):

Yuyuutsu (approved):

berroar (approved):

PR #1001, opened 2023-01-18
Calculated Summary only show answers on the path

berroar:

Updates Calculated Summaries so that only answers on the path are displayed. 

This means we will no longer need skip conditions routing to multiple calculated summaries as can be seen in the original `test_calculated_summary` schema. The change applies to both the [new calculated summary format](https://github.com/ONSdigital/eq-questionnaire-runner/pull/965) and the original calculated summary format.

Adds support for cross section dependencies i.e. piping a calculated summary value from one section into another.

### How to review 

Check that the calculated summary schemas that have been updated still behave in the same way:

`test_calculated_summary`, `test_new_calcualted_summary`, `test_new_calcualted_summary_repeating_section`, `test_new_routing_and_skipping_section_dependencies_calculated_summary.json`

Test cross section dependencies with: `test_new_calculated_summary_cross_section_dependencies` and `test_new_calculated_summary_cross_section_dependencies_repeating`

Additional tests have been added to cover scenarios where editing answers leads to new answers on the path that effect the total of a calculated summary.

### Checklist


berroar:
[Refactor Calculated Summary: Only show answers on the path](https://trello.com/c/abxCKQeE/5497-refactor-calculated-summary-only-show-answers-on-the-path)

app/views/handlers/content.py
```diff
@@ -23,7 +23,9 @@ def rendered_block(self):
         ) or self._get_content_title(transformed_block)
         self._set_page_title(content_page_title)
         return self.placeholder_renderer.render(
-            transformed_block, self._current_location.list_item_id
+            transformed_block,
```
petechd, line 26: Since you use named attributes when you invoke this method in `list_collector_block.py` I would suggest using named attributes here as well.

app/views/handlers/question.py
```diff
@@ -73,7 +75,9 @@ def rendered_block(self):
 
         self._set_page_title(page_title)
         rendered_question = self.placeholder_renderer.render(
-            transformed_block["question"], self._current_location.list_item_id
+            transformed_block["question"],
```
petechd, line 78: Same as my comment above, do what you did in list collector block, keyword attributes?

app/questionnaire/placeholder_renderer.py
```diff
@@ -99,7 +104,10 @@ def render_placeholder(
         return formatted_placeholder_data
 
     def render(
-        self, dict_to_render: Mapping[str, Any], list_item_id: Optional[str]
+        self,
+        dict_to_render: Mapping[str, Any],
```
petechd, line 108:
```suggestion
        self,
        *,
        dict_to_render: Mapping[str, Any],
```
Enforce named attributes?

app/questionnaire/placeholder_renderer.py
```diff
@@ -108,7 +116,9 @@ def render(
         pointers = find_pointers_containing(dict_to_render, "placeholders")
 
         for pointer in pointers:
-            rendered_text = self.render_pointer(dict_to_render, pointer, list_item_id)
+            rendered_text = self.render_pointer(
```
petechd, line 119: Worth adding and enforcing named attributes in `render_pointer` as there is four of them now, but not a blocker.

tests/integration/questionnaire/test_questionnaire_calculated_summary.py
```diff
@@ -48,31 +48,27 @@ def _complete_calculated_summary_path_no_skip(self):
...
         self._complete_calculated_summary_path_with_skip()
-        self.assertInBody("Skipped Fourth")
         self.assertInBody(
             "We calculate the total of currency values entered to be £80.00"
         )
 
     def test_calculated_summary_no_skip(self):
         self.launchSurvey("test_calculated_summary")
         self._complete_calculated_summary_path_no_skip()
-        self.assertNotInBody("Skipped Fourth")
```
petechd, line 59: Instead of removing this shouldn't we check that `"Fourth answer label"` is in body now?

app/questionnaire/value_source_resolver.py
```diff
@@ -144,6 +145,7 @@ def _resolve_calculated_summary_value_source(
             self.metadata,
             self.response_metadata,
             location=self.location,
+            routing_path_block_ids=self.routing_path_block_ids,
         )
 
         return evaluate_calculated_summary.evaluate(calculation["operation"])  # type: ignore
```
petechd, line 151: We probably need a comment here explaining type ignore.

app/views/handlers/content.py
berroar: Updated this 👍

app/views/handlers/question.py
berroar: Updated 👍

app/questionnaire/placeholder_renderer.py
berroar: Enforcing now and updated in relevant calls 👍

tests/integration/questionnaire/test_questionnaire_calculated_summary.py
berroar: Not sure we necessarily need this as it is really just testing the total which is only correct if the fourth is skipped?

app/questionnaire/placeholder_renderer.py
berroar: Enforcing this one as well now 👍

ONS-Guilhem-Forey (approved):
Had a good look and it makes sense to me! Also tested a few schemas locally and LGTM 👍

app/forms/questionnaire_form.py
```diff
@@ -449,6 +449,7 @@ def get_answer_fields(
     metadata: Optional[MetadataProxy],
     response_metadata: Mapping[str, Any],
     location: Union[Location, RelationshipLocation, None],
+    routing_path_block_ids: Optional[tuple] = None,
```
MebinAbraham, line 452: Optional Iterable/Sequence of strings?.

app/questionnaire/placeholder_renderer.py
```diff
@@ -39,13 +39,17 @@ def __init__(
 
     def render_pointer(
         self,
+        *,
         dict_to_render: Mapping[str, Any],
         pointer_to_render: str,
         list_item_id: Optional[str],
+        routing_path_block_ids: Optional[tuple] = None,
```
MebinAbraham, line 46: This shouldn't be passed to pointer, this should be on the object itself.

app/questionnaire/placeholder_renderer.py
```diff
@@ -39,13 +39,17 @@ def __init__(
...
     ) -> str:
         pointer_data = resolve_pointer(dict_to_render, pointer_to_render)
 
-        return self.render_placeholder(pointer_data, list_item_id)
+        return self.render_placeholder(
+            pointer_data, list_item_id, routing_path_block_ids
```
MebinAbraham, line 51: Same as above.

app/questionnaire/placeholder_renderer.py
```diff
@@ -65,6 +69,7 @@ def render_placeholder(
         self,
         placeholder_data: MutableMapping[str, Any],
         list_item_id: Optional[str],
+        routing_path_block_ids: Optional[tuple] = None,
```
MebinAbraham, line 72: Same as above.

app/questionnaire/placeholder_renderer.py
```diff
@@ -99,7 +105,11 @@ def render_placeholder(
         return formatted_placeholder_data
 
     def render(
-        self, dict_to_render: Mapping[str, Any], list_item_id: Optional[str]
+        self,
+        *,
+        dict_to_render: Mapping[str, Any],
+        list_item_id: Optional[str],
+        routing_path_block_ids: Optional[tuple] = None,
```
MebinAbraham, line 112: Same as above.

app/questionnaire/rules/rule_evaluator.py
```diff
@@ -29,7 +29,7 @@ class RuleEvaluator:
     metadata: Optional[MetadataProxy]
     response_metadata: Mapping
     location: Union[None, Location, RelationshipLocation]
-    routing_path_block_ids: Optional[list] = None
+    routing_path_block_ids: Union[Optional[tuple], list[str]] = None
```
MebinAbraham, line 32: Always an optional list of strings?. If we don't need it to be list specific, it can be `Iterable/Sequence`

app/questionnaire/value_source_resolver.py
```diff
@@ -162,6 +164,9 @@ def resolve(
     ) -> Union[ValueSourceEscapedTypes, ValueSourceTypes]:
         source = value_source["source"]
 
+        if not self.assess_routing_path:
+            self.assess_routing_path = source == "calculated_summary"
```
MebinAbraham, line 168: It would be useful to add a comment as to which this is special.

app/views/contexts/calculated_summary_context.py
```diff
@@ -24,7 +25,9 @@ class CalculatedSummaryContext(Context):
     def build_groups_for_section(
         self, section: Mapping, return_to_block_id: str, current_location: Location
     ) -> List[Group]:
-        routing_path = self._router.routing_path(section["id"])
+        routing_path = self._router.routing_path(
+            section["id"], list_item_id=current_location.list_item_id
+        )
```
MebinAbraham, line 30: How did repeats work without this?. What is this fixing and is there a test?.

app/views/contexts/calculated_summary_context.py
```diff
@@ -45,7 +48,7 @@ def build_groups_for_section(
         ]
 
     def build_view_context_for_calculated_summary(
-        self, current_location: Location
+        self, current_location: Location, routing_path: RoutingPath
```
MebinAbraham, line 51: You don't need to pass the path as you can assess it from the instance already?.

app/views/contexts/calculated_summary_context.py
```diff
@@ -167,6 +171,7 @@ def _get_formatted_total(
         groups: list,
         current_location: Location,
         calculation: Union[Callable, ImmutableDict],
+        routing_path: RoutingPath,
```
MebinAbraham, line 174: Same as above.

app/views/contexts/calculated_summary_context.py
MebinAbraham, line 30: I know this is existing but is this evaluation even needed?. Although the method is build groups for section, this is only every called for the current locations section, so we don't need this re-evaluation?. You should just be able to use `self._routing_path`?

app/views/contexts/calculated_summary_context.py
MebinAbraham: The current location doesn't need to be passed either, it is already accessible.

app/views/contexts/calculated_summary_context.py
MebinAbraham: The only thing this method needs should be groups and calculation.

app/questionnaire/placeholder_renderer.py
berroar: Moved to the object now 👍

app/questionnaire/placeholder_renderer.py
berroar: As above 👍

app/questionnaire/placeholder_renderer.py
berroar: As above 👍

app/questionnaire/placeholder_renderer.py
berroar: As above 👍

app/forms/questionnaire_form.py
berroar: Updated 👍

app/questionnaire/rules/rule_evaluator.py
berroar: Updated now 👍

app/views/contexts/calculated_summary_context.py
berroar: Removed these now 👍

app/views/contexts/calculated_summary_context.py
berroar: Updated this 👍

app/views/contexts/calculated_summary_context.py
berroar: I have replaced this with self._routing_path now 👍

berroar: Removed it now as per the comment below 👍

app/questionnaire/value_source_resolver.py
berroar: Added a comment now

app/questionnaire/placeholder_parser.py
```diff
@@ -58,6 +63,9 @@ def __init__(
         self._placeholder_map: MutableMapping[
             str, Union[ValueSourceEscapedTypes, ValueSourceTypes, None]
         ] = {}
+        self._routing_path_block_ids = routing_path_block_ids
```
petechd, line 66: Do we need to create all these private vars when in reality they are not being used except as attributes for the resolver? We only really need transformer and placeholder_map:
Screenshot 2023-02-21 at 10 22 04

berroar: Updated this

app/questionnaire/value_source_resolver.py
```diff
@@ -183,4 +196,33 @@ def resolve(
...
+                block_id
+                for block_id in completed_block_ids
+                if block_id not in self.routing_path_block_ids
+            ]:
+                self.routing_path_block_ids = [
+                    *list(self.routing_path_block_ids),
+                    *routing_path_block_ids,
+                ]
+        else:
+            self.routing_path_block_ids = completed_block_ids
```
ONS-Guilhem-Forey, line 228: Optional, but I reckon this if/else could be simplified with:

```python
            self.routing_path_block_ids = set([
                *list(self.routing_path_block_ids or []),
                *routing_path_block_ids
            ])
```

app/questionnaire/value_source_resolver.py
```diff
@@ -144,6 +151,7 @@ def _resolve_calculated_summary_value_source(
             self.metadata,
             self.response_metadata,
             location=self.location,
+            routing_path_block_ids=self.routing_path_block_ids,
         )
 
         return evaluate_calculated_summary.evaluate(calculation["operation"])  # type: ignore
```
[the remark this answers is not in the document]
berroar: Updated now 👍

app/questionnaire/value_source_resolver.py
```diff
@@ -183,4 +196,33 @@ def resolve(
...
+                if not self.list_item_id:
+                    routing_path_for_section = self.path_finder.routing_path(
+                        section_id, list_item_id
+                    )
+                    completed_block_ids.extend(routing_path_for_section.block_ids)
+                elif list_item_id == self.list_item_id or list_item_id is None:
+                    routing_path_for_section = self.path_finder.routing_path(
+                        section_id, list_item_id
+                    )
+                    completed_block_ids.extend(routing_path_for_section.block_ids)
```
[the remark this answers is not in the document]
berroar: Refactored the logic here 👍

app/questionnaire/value_source_resolver.py
```diff
@@ -183,4 +196,23 @@ def resolve(
             return self.response_metadata.get(value_source.get("identifier"))
 
         if source == "calculated_summary":
+            self.evaluate_completed_blocks_for_calculated_summary()
             return self._resolve_calculated_summary_value_source(value_source)
+
+    def evaluate_completed_blocks_for_calculated_summary(self) -> None:
```
MebinAbraham, line 202: We can't do it like this as it would be very inefficient and would bottleneck eQ. This is evaluating the path for every section in the progress store. If calculated summary was the 10th section, this would evaluate the path for at least 9 + no. of repeating sections. If you have 4 repeats, that is 13 routing paths even if only 1 section is all you need.

Value source evaluator should be passed the block ids it needs to consider. It shouldn't have to work it out. 

If VSR is used by placeholder, then get placeholder should pass the block_ids, if used by RuleEvaluator, then RE should pass it. Or if directly using VSR then, it should pass it directly. You need to create a map of sections that the calculated summary depends on  instead of evaluating every section.

berroar: Refactored now 👍

app/questionnaire/placeholder_parser.py
```diff
@@ -134,3 +156,81 @@ def _resolve_value_source_list(
             else:
                 values.append(value)
         return values
+
+    def _get_routing_path_block_ids(self) -> list[str]:
+        return self._get_block_ids_for_calculated_summary_dependencies()
+
+    def _get_block_ids_for_calculated_summary_dependencies(self) -> list:
```
MebinAbraham, line 163: This is almost identical to the one in questionnaire form. I suggest having one, returning something generic that both can make use of. If you give a list of section to ignore, you can make use of cache and then set the cache based on the return value. It should return a dict keyed by section id and then blocks (same as the cache itself).

In placeholder, it can then set the cache, in questionnaire form, it just needs to flatten the block ids.

Something like (not tested FYI):
```py
    def _get_block_ids_for_calculated_summary_dependencies(
        self, sections_to_ignore: Sequence | None = None
    ) -> dict[str, list[str]]:
        # Type ignore: Added to this method as the block will exist at this point
        blocks_id_by_section = {}

        sections_to_ignore = sections_to_ignore or []
        dependent_sections = self._schema.calculated_summary_section_dependencies_by_block.get(  # type: ignore
            self._location.section_id
        )
        if block_id := self._location.block_id:
            dependents = dependent_sections[block_id]

        else:
            dependents = {
                section
                for dependents in dependent_sections.values()
                if dependents
                for section in dependents
            }

        for section in dependents:
            if section in sections_to_ignore:
                continue

            key = (section, self._location.list_item_id)
            if key in self._path_finder.progress_store.started_section_keys():
                path = self._path_finder.routing_path(*key)
                blocks_id_by_section[section] = path.block_ids

        return blocks_id_by_section
```

then in placeholder, you can update the cache. The blocks ids you pass can just be a flattened list of the cached routing path ids. You can have a common method to do this, as you need to flatten in both form and parser.

app/questionnaire/questionnaire_schema.py
```diff
@@ -898,6 +907,60 @@ def _get_rules_section_dependencies(
...
+        for section in self.get_sections():
+            for block in self.get_blocks_for_section(section):
+                sources = self._get_dictionaries_with_key(
+                    "source", block, ignore_keys=["when"]
+                )
+
+                calculated_summary_sources = [
+                    source
+                    for source in sources
+                    if source.get("source") == "calculated_summary"
```
MebinAbraham, line 920: Minor, use bracket access instead of `.get()` when the key is mandatory

app/forms/questionnaire_form.py
```diff
@@ -597,3 +617,54 @@ class DynamicForm(QuestionnaireForm):
...
+            block_dependencies.extend(path)
+
+        if (
+            dependent_section,
+            location.list_item_id,
+        ) in path_finder.progress_store.started_section_keys():
+            path = path_finder.routing_path(
+                section_id=dependent_section, list_item_id=location.list_item_id
+            )
+            block_dependencies.extend(path)
```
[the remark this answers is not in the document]
berroar: If I remove this it breaks when i'm piping from a non repeating section into a repeating section for my `test_new_calculated_summary_cross_section_dependencies_repeating` as it will only set the dependency with the list_item_id

app/forms/questionnaire_form.py
```diff
@@ -597,3 +617,54 @@ class DynamicForm(QuestionnaireForm):
...
+
+    dependent_sections: set = set()
+    if block_id := location.block_id:
+        if schema.get_block(block_id).get("type") not in [  # type: ignore
+            "ListEditQuestion",
+            "ListAddQuestion",
+            "ListRemoveQuestion",
+            "UnrelatedQuestion",
+            "PrimaryPersonListAddOrEditQuestion",
+        ]:
```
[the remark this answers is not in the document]
berroar: Without these I get a key error when getting the dependency

app/questionnaire/placeholder_parser.py
berroar: I've refactored these methods in-line with the above 👍

app/forms/questionnaire_form.py
```diff
@@ -597,3 +617,54 @@ class DynamicForm(QuestionnaireForm):
...
+
+
+def _get_block_ids_for_calculated_summary_dependencies(
+    path_finder: PathFinder | None,
+    location: Location | RelationshipLocation | None,
+    schema: QuestionnaireSchema,
+) -> list:
+    # Type ignore: Added to this method as the block will exist at this point
+    if not path_finder:
+        raise ValueError("PathFinder not set")
```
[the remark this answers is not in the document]
berroar: Removed now 👍

tests/app/questionnaire/test_placeholder_renderer.py
```diff
@@ -67,8 +74,13 @@ def test_renders_json(placholder_transform_question_json, mocker):
         ]
     )
 
-    renderer = get_placeholder_render(mocker=mocker, answer_store=answer_store)
-    rendered_schema = renderer.render(json_to_render, list_item_id=None)
+    mocker.patch(
+        "app.questionnaire.placeholder_parser.PlaceholderParser._get_block_ids_for_calculated_summary_dependencies",
+        return_value=[],
+    )
```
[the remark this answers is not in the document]
berroar: These have been removed now 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -788,22 +792,27 @@ def _get_values_for_key(
...
+        if key in dictionary and key not in ignore_keys:
             yield dictionary
 
-        for value in dictionary.values():
-            if isinstance(value, Sequence):
-                for element in value:
+        for k, v in dictionary.items():
+            if k in ignore_keys:
+                continue
+            if isinstance(v, dict):
```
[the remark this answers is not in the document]

app/questionnaire/questionnaire_schema.py

app/questionnaire/questionnaire_schema.py
```diff
@@ -898,6 +907,60 @@ def _get_rules_section_dependencies(
...
+            identifier: Optional[str] = value.get("identifier")
+            source: Optional[str] = value.get("source")
+
+            if source == "calculated_summary" and identifier:
+                calculated_summary_block = self.get_block(identifier)
+                calculated_summary_answer_ids = self.get_calculated_summary_answer_ids(
+                    calculated_summary_block  # type: ignore
+                )
+                answer_id_list.extend(iter(calculated_summary_answer_ids))
+            for answer_id in answer_id_list:
```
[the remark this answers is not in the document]
berroar: Updated 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -898,6 +907,60 @@ def _get_rules_section_dependencies(
...
+                block = self.get_block_for_answer_id(answer_id)  # type: ignore
+                section_id = self.get_section_id_for_block_id(block["id"])  # type: ignore
+
+                if section_id:
```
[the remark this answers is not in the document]

app/questionnaire/value_source_resolver.py
```diff
@@ -42,9 +49,18 @@ def _is_answer_on_path(self, answer_id: str) -> bool:
...
-    ) -> Optional[AnswerValueTypes]:
-        if not self._is_answer_on_path(answer_id):
+        self,
+        answer_id: str,
+        list_item_id: str | None,
+        assess_routing_path: bool | None = None,
+    ) -> AnswerValueTypes | None:
+        assess_routing_path = (
+            assess_routing_path
+            if assess_routing_path is True
```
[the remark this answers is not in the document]
berroar: Updated 👍

app/views/contexts/summary/group.py
```diff
@@ -20,43 +21,57 @@ def __init__(
...
-            language=language,
-            answer_store=answer_store,
-            list_store=list_store,
-            location=location,
-            metadata=metadata,
-            response_metadata=response_metadata,
-            schema=schema,
-        )
+
+        if self.location:
```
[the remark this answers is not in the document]
berroar: This is mainly for the type hint issues, location is initialised in the context class and it's optional there

app/views/contexts/context.py
```diff
@@ -38,11 +42,30 @@ def __init__(
...
+        )
+
+        if self._location:
+            self._routing_path = self._path_finder.routing_path(
+                self._location.section_id, self._location.list_item_id
+            )
+            self._routing_path_block_ids = self._routing_path.block_ids
+        else:
+            self._routing_path_block_ids = None
+
```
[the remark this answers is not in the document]
berroar: Needed to do this as we don't have location for all context classes

app/forms/questionnaire_form.py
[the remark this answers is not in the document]
MebinAbraham: Handle that or use a default dict?

app/forms/questionnaire_form.py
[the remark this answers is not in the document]
MebinAbraham: I still don't quite follow. I guess you need to use the list item id if you are currently in a repeat and none if it is not a repeat. A repeating section can't be a dependent unless it is the current repeat.

app/views/contexts/context.py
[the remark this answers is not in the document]
MebinAbraham: But did we always need a routing path?

app/views/handlers/block.py
```diff
@@ -43,6 +44,14 @@ def __init__(
         self._return_to_answer_id = request_args.get("return_to_answer_id")
         self._return_to_block_id = request_args.get("return_to_block_id")
         self.resume = "resume" in request_args
+        self._path_finder = PathFinder(
```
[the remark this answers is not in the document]
berroar: Refactored to use Router now 👍

app/forms/questionnaire_form.py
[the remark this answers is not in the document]
berroar: Handling the error now

app/forms/questionnaire_form.py
```diff
@@ -597,3 +617,54 @@ class DynamicForm(QuestionnaireForm):
...
+    dependent_sections: set = set()
+    if block_id := location.block_id:
+        if schema.get_block(block_id).get("type") not in [  # type: ignore
+            "ListEditQuestion",
+            "ListAddQuestion",
+            "ListRemoveQuestion",
+            "UnrelatedQuestion",
+            "PrimaryPersonListAddOrEditQuestion",
+        ]:
+            dependent_sections = schema.calculated_summary_section_dependencies_by_block.get(  # type: ignore
```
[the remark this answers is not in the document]
berroar: Removed this 👍

app/questionnaire/placeholder_renderer.py
```diff
@@ -24,32 +28,55 @@ def __init__(
...
+            list_store=self._list_store,
+            metadata=self._metadata,
+            response_metadata=self._response_metadata,
+            schema=self._schema,
+            list_item_id=list_item_id,
+            location=self._location,
+            renderer=self,
+            progress_store=self._progress_store,
+            path_finder=self._path_finder,
+        )
```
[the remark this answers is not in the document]
berroar: When I try to move this a level up, it breaks tests for the multiple piping schema 🤔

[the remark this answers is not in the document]
MebinAbraham: How come?

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +163,62 @@ def _resolve_value_source_list(
                 values.append(value)
         return values
 
+    def _get_routing_path_block_ids(
+        self, sections_to_ignore: list | None = None
+    ) -> dict[str, list[str]]:
+        if self._location and self._router:
+            return get_block_ids_for_calculated_summary_dependencies(
```
MebinAbraham, line 170: This doesn't cover that we might not pass then even when we need it. Can we not always have these?. If not, raise an error in the methods that uses it. At the moment, it would silently pass.

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +163,62 @@ def _resolve_value_source_list(
...
+    sections_to_ignore = sections_to_ignore or []
+    dependent_sections = schema.calculated_summary_section_dependencies_by_block.get(
+        location.section_id
+    )
+
+    if block_id := location.block_id:
+        try:
+            dependents = dependent_sections[block_id]  # type: ignore
+        except KeyError:
+            dependents = set()
```
MebinAbraham, line 201: Make use of default dicts instead?.

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +163,62 @@ def _resolve_value_source_list(
...
+            if dependents
+            for section in dependents
+        }
+
+    for section in dependents:
+        if section in sections_to_ignore:
+            continue
+
+        keys = [(section, location.list_item_id), (section, None)]
+        for key in keys:
```
MebinAbraham, line 215: This still seems wrong, why would we be evaluating 2 paths for every dependent section?.

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +163,62 @@ def _resolve_value_source_list(
...
+            if key in router.progress_store.started_section_keys():
+                path = router.routing_path(*key)
+                blocks_id_by_section[section] = path.block_ids
+
+    return blocks_id_by_section
+
+
+def flatten_block_ids_map(routing_path_block_ids_map: dict[str, list[str]]) -> set[str]:
```
MebinAbraham, line 223: Name this generically. Maybe. `get_flattened_mapping_values`?

app/questionnaire/questionnaire_schema.py
```diff
@@ -57,6 +57,9 @@ def __init__(
             set
         )
         self._when_rules_section_dependencies_by_section: dict[str, set[str]] = {}
+        self.calculated_summary_section_dependencies_by_block: OrderedDict[
+            str, dict[str, set[str]]
+        ] = OrderedDict()
```
MebinAbraham, line 62: Why does it need ordered dict?

app/questionnaire/questionnaire_schema.py
```diff
@@ -896,6 +905,58 @@ def _get_rules_section_dependencies(
...
+        self,
+        current_section_id: str,
+        current_block_id: str,
+        values: list[Mapping],
+    ) -> None:
+        section_dependencies: set[str] = set()
+
+        for value in values:
+            answer_id_list: list = []
+            identifier: Optional[str] = value["identifier"]
```
MebinAbraham, line 937: Not optional?

app/questionnaire/questionnaire_schema.py
```diff
@@ -801,14 +805,19 @@ def get_values_for_key(
                 continue
 
     def _get_dictionaries_with_key(
-        self, key: str, dictionary: Mapping
+        self, key: str, dictionary: Mapping, ignore_keys: list[str] | None = None
```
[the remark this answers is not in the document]
berroar: Renamed 👍

app/questionnaire/placeholder_parser.py
berroar: Renamed 👍

app/questionnaire/questionnaire_schema.py

app/questionnaire/placeholder_parser.py
```diff
@@ -44,46 +45,68 @@ def __init__(
...
+
+        if routing_path_block_ids_map := self._get_routing_path_block_ids(
+            self._sections_to_ignore
+        ):
+            self._sections_to_ignore.extend(iter(routing_path_block_ids_map.keys()))
+
+            routing_path_block_ids = flatten_block_ids_map(routing_path_block_ids_map)
+            self._value_source_resolver = self._get_value_source_resolver(
+                routing_path_block_ids
+            )
```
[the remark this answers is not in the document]
berroar: Updated this 👍

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +163,62 @@ def _resolve_value_source_list(
...
+        }
+
+    for section in dependents:
+        if section in sections_to_ignore:
+            continue
+
+        keys = [(section, location.list_item_id), (section, None)]
+        for key in keys:
+            if key in router.progress_store.started_section_keys():
+                path = router.routing_path(*key)
```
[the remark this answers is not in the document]
berroar: Updated this so am passing progress_store, need to look into circular import issue for the path finder which is now being initialised in parser

app/questionnaire/placeholder_parser.py
berroar: Updated the logic here to check if section is repeating 👍

app/questionnaire/placeholder_parser.py
berroar: Using default dict now 👍

app/questionnaire/placeholder_renderer.py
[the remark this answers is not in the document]
berroar: I have updated this, but it broke tests in places where there are multiple placeholders with the same name in the same block, so had to update a few schemas. May need to add validation to cover this

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +169,66 @@ def _resolve_value_source_list(
...
+            key = (section, None)
+
+        if key in progress_store.started_section_keys():
+            routing_path = path.routing_path(*key)
+            blocks_id_by_section[section] = routing_path.block_ids
+
+    return blocks_id_by_section
+
+
+def get_flattened_mapping_value(
```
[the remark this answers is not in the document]
berroar: I have moved this to a utils file 👍

app/views/contexts/context.py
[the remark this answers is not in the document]
berroar: Routing path is needed for the CalculatedSummary Context class, no longer need to see routing path block ids here though

app/views/contexts/context.py
```diff
@@ -39,12 +43,28 @@ def __init__(
...
+            self._schema,
+            self._answer_store,
+            self._list_store,
+            self._progress_store,
+            self._metadata,
+            self._response_metadata,
+        )
+
+        if self._location:
+            self._routing_path = self._path_finder.routing_path(
```
MebinAbraham, line 56: Since you got router here already, you don't need path finder as well, it can just be `self._router.routing_path()`

MebinAbraham, line 56: We should set the routing path to `None` then create it in `CalculatedSummaryContext`, otherwise you are creating it for all context classes when only Cal summary needs it.

Actually have a look at `SectionSummaryContext` that passes in routing path and location as it needs it, then you also won't get type hint issues.

app/views/contexts/summary/group.py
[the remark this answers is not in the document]
MebinAbraham: This should raise an error or make it, so location isn't optional see comment above. Currently, this would silently fail.

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
MebinAbraham: Has the circular import issue been resolved?

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +171,60 @@ def _resolve_value_source_list(
...
+
+    if block_id := location.block_id:
+        dependents = dependent_sections[block_id]
+    else:
+        dependents = {
+            section
+            for dependents in dependent_sections.values()
+            if dependents
+            for section in dependents
+        }
```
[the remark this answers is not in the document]
berroar: Using this method now 👍

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +171,60 @@ def _resolve_value_source_list(
...
     def _all_value_sources_metadata(self, placeholder: Mapping) -> bool:
         sources = self._schema.get_values_for_key(placeholder, key="source")
         return all(source == "metadata" for source in sources)
+
+
+def get_block_ids_for_calculated_summary_dependencies(
+    schema: QuestionnaireSchema,
+    location: Location | RelationshipLocation,
+    progress_store: ProgressStore,
+    path: "PathFinder",
```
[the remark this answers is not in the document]
berroar: I had to name it something other than path_finder as otherwise it fails pylint for redefined-outer-name

app/questionnaire/questionnaire_schema.py
```diff
@@ -896,13 +903,65 @@ def _get_rules_section_dependencies(
...
+                    current_section_id=section["id"],
+                    current_block_id=block["id"],
+                    values=calculated_summary_sources,
+                )
+
+    def _get_calculated_summary_section_dependencies(
+        self,
+        current_section_id: str,
+        current_block_id: str,
+        values: list[Mapping],
```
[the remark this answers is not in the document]
berroar: Renamed this 👍

app/questionnaire/value_source_resolver.py
```diff
@@ -25,14 +25,15 @@
...
-    metadata: Optional[MetadataProxy]
+    metadata: MetadataProxy | None
     response_metadata: Mapping
     schema: QuestionnaireSchema
-    location: Union[None, Location, RelationshipLocation]
-    list_item_id: Optional[str]
-    routing_path_block_ids: Optional[list] = None
+    location: Location | RelationshipLocation | None
+    list_item_id: str | None
+    routing_path_block_ids: Iterable | None = None
```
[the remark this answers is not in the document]
berroar: Added 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -896,13 +903,65 @@ def _get_rules_section_dependencies(
...
+    ) -> None:
+        section_dependencies: set[str] = set()
```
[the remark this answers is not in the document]
berroar: Added 👍

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +171,60 @@ def _resolve_value_source_list(
...
+    sections_to_ignore: list | None = None,
+) -> dict[str, list[str]]:
+    blocks_id_by_section = {}
+
+    sections_to_ignore = sections_to_ignore or []
+    dependent_sections = defaultdict(
+        set,
+        schema.calculated_summary_section_dependencies_by_block[location.section_id],
```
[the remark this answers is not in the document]
berroar: Updated this 👍

app/views/contexts/context.py
berroar: Updated so that it is similar to Section Summary 👍

schemas/test/en/test_routing_and_skipping_section_dependencies_new_calculated_summary.json
```diff
@@ -79,6 +79,62 @@
                                 ]
```
MebinAbraham, line 79: I think the existing guidance is wrong, it says:
```
If the total is equal to £100 a new section will appear on the hub and if it is greater than or equal to £100 a dependent question will appear in the dependent question section
```

According to the schema it should be `10` not 100 and `if it is less than 10 then it will appear`?

app/questionnaire/placeholder_parser.py
```diff
@@ -140,6 +170,53 @@ def _resolve_value_source_list(
...
+                sections_to_ignore=sections_to_ignore,
+            )
+        return {}
+
     def _all_value_sources_metadata(self, placeholder: Mapping) -> bool:
         sources = self._schema.get_values_for_key(placeholder, key="source")
         return all(source == "metadata" for source in sources)
+
+
+def get_block_ids_for_calculated_summary_dependencies(
```
MebinAbraham, line 191: There is an issue with this in that, every time placeholder render is called, we will evaluate the path, although we cache the path, if the renderer is different, it will reevaluate. If you look at section summary, you can see that the path is evaluated multiple times, this is because a renderer is used to resolve the title for the block which is different to the used for the summary blocks, but since we say the block has dependencies, it will evaluate the path even if no placeholder contain any calc summaries sources.

You will want to add a check to see if there is any calc summary sources. I would check if after finding out if you have any dependencies as it would be more performant as dependencies are pre resolved.

I would make a generic method like this that you can reuse.
```py
    def get_sources_for_type_from_data(self, *, source_type: str, data, ignore_keys):
        sources = self._get_mappings_with_key(
            "source", data, ignore_keys=ignore_keys
        )

        return [
            source
            for source in sources
            if source["source"] == source_type
        ]
```

Then it would be:

```py
if dependents and not calc_sources := get_sources_for_type_from_data("calculated_summary", data=data, ignore_keys=["when"]):
	return blocks_id_by_section
```

- `data` would be the placeholders that you need to pass down.

app/utilities/mappings.py
```diff
@@ -0,0 +1,8 @@
+from collections import defaultdict
+from typing import Iterable, Mapping
+
+
+def get_flattened_mapping_value(
```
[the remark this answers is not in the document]
berroar: Updated this 👍

tests/functional/spec/features/calculated_summary_test_case.js
```diff
@@ -327,11 +338,101 @@ class TestCase {
...
+    });
+
     it("Given I am on the summary, When I submit the questionnaire, Then I should see the thank you page", async () => {
       await $(SubmitPage.submit()).click();
       await expect(await browser.getUrl()).to.contain(ThankYouPage.pageName);
     });
   }
+
+  testCrossSectionDependencies(schema) {
+    before("Get to the question containing calcualted summary values with cross section dependcies", async () => {
```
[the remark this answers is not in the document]
berroar: Fixed typos 👍

app/questionnaire/placeholder_parser.py
berroar: Have updated this method and added the logic above 👍

schemas/test/en/test_routing_and_skipping_section_dependencies_new_calculated_summary.json
berroar: Updated this 👍

app/views/handlers/question.py
```diff
@@ -83,13 +87,14 @@ def rendered_block(self):
...
-            self._questionnaire_store.metadata,
-            self._questionnaire_store.response_metadata,
+            language=self._language,
+            schema=self._schema,
+            answer_store=self._questionnaire_store.answer_store,
+            list_store=self._questionnaire_store.list_store,
+            progress_store=self._questionnaire_store.progress_store,
+            metadata=self._questionnaire_store.metadata,
+            response_metadata=self._questionnaire_store.response_metadata,
+            location=self.current_location,
```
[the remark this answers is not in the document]
berroar: Updated this and the above now 👍

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
berroar: Updated as discussed 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -896,13 +903,66 @@ def _get_rules_section_dependencies(
...
+            for answer_id in answer_id_list:
+                block = self.get_block_for_answer_id(answer_id)  # type: ignore
+                section_id = self.get_section_id_for_block_id(block["id"])  # type: ignore
+
+                section_dependencies.add(section_id)  # type: ignore
+
+        if section_dependency := self.calculated_summary_section_dependencies_by_block.get(
+            current_section_id
+        ):
+            section_dependency[current_block_id] = section_dependencies
```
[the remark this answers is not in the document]
berroar: I have updated this 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -896,13 +903,66 @@ def _get_rules_section_dependencies(
...
+        else:
+            self.calculated_summary_section_dependencies_by_block[
+                current_section_id
+            ] = defaultdict(set, {current_block_id: section_dependencies})
+
```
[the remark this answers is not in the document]
berroar: Returning from the get method now 👍

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
petechd: Maybe we should add something like `similar issues below` for all the type ignores in this method?

app/forms/questionnaire_form.py
```diff
@@ -440,15 +445,39 @@ def _option_value_in_data(
...
-    metadata: Optional[MetadataProxy],
+    metadata: MetadataProxy | None,
     response_metadata: Mapping[str, Any],
-    location: Union[Location, RelationshipLocation, None],
+    location: Location | RelationshipLocation | None,
+    progress_store: ProgressStore | None,
 ) -> dict[str, FieldHandler]:
     list_item_id = location.list_item_id if location else None
+
+    routing_path_block_ids: Optional[dict[tuple, list[str]]] = {}
```
MebinAbraham, line 459: This isn't optional, is always a dict, is also initialised as empty dict.

app/questionnaire/placeholder_parser.py
```diff
@@ -40,50 +38,85 @@ def __init__(
...
-            schema=self._schema,
-            location=self._location,
-            list_item_id=self._list_item_id,
-            escape_answer_values=True,
             response_metadata=self._response_metadata,
-            use_default_answer=True,
         )
 
+        self._value_source_resolver = self._get_value_source_resolver()
+        self._routing_paths: dict = {}
```
MebinAbraham, line 74: Minor, for clarity, I would rename to `self._routing_path_block_ids_by_section_key`

app/questionnaire/placeholder_renderer.py
```diff
@@ -38,20 +40,24 @@ def __init__(
         self._response_metadata = response_metadata
         self._schema = schema
         self._location = location
+        self._progress_store = progress_store
+        self._block_ids_calculated_summary: list = []
```
MebinAbraham, line 44: This is unused.

app/questionnaire/questionnaire_schema.py
```diff
@@ -896,13 +911,68 @@ def _get_rules_section_dependencies(
...
+                if self.calculated_summary_section_dependencies_by_block.get(
+                    section["id"]
+                ):
+                    self.calculated_summary_section_dependencies_by_block[
+                        section["id"]
+                    ][block["id"]] = section_dependencies
+                else:
+                    self.calculated_summary_section_dependencies_by_block[
+                        section["id"]
+                    ] = defaultdict(set, {block["id"]: section_dependencies})
```
MebinAbraham, line 942: Can this now not be simplified to:
```py
self.calculated_summary_section_dependencies_by_block[section["id"]][
    block["id"]
].update(section_dependencies)
```

app/views/contexts/context.py
```diff
@@ -28,6 +30,7 @@ def __init__(
         self._progress_store = progress_store
         self._metadata = metadata
         self._response_metadata = response_metadata
+        self._location = location
```
MebinAbraham, line 33: Can this now be reverted?

app/forms/questionnaire_form.py
berroar: Updated 👍

app/questionnaire/placeholder_parser.py
berroar: Updated 👍

app/questionnaire/placeholder_renderer.py
berroar: Removed this 👍

app/questionnaire/questionnaire_schema.py

app/questionnaire/placeholder_parser.py
MebinAbraham: This is still an issue but we are okay for now.

app/questionnaire/questionnaire_schema.py
```diff
@@ -896,13 +911,61 @@ def _get_rules_section_dependencies(
...
+
+        for source in sources:
+            answer_id_list: list = []
+            identifier: str = source["identifier"]
+
+            calculated_summary_block = self.get_block(identifier)  # type: ignore
+            calculated_summary_answer_ids = self.get_calculated_summary_answer_ids(
+                calculated_summary_block  # type: ignore
+            )
+            answer_id_list.extend(iter(calculated_summary_answer_ids))
```
MebinAbraham, line 952: Minor, you don't need iter.

app/questionnaire/rules/rule_evaluator.py
```diff
@@ -54,6 +56,7 @@ def __post_init__(self) -> None:
             response_metadata=self.response_metadata,
             schema=self.schema,
             location=self.location,
+            progress_store=self.progress_store,  # type: ignore
```
MebinAbraham, line 59: Why is this needed?. I see no comment about it.

MebinAbraham (commented):
LGTM, outstanding comments needs resolving, that's all.

app/views/contexts/summary/group.py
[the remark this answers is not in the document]
MebinAbraham: Okay for now, being pragmatic. We'll deal with it if it becomes a problem.

app/questionnaire/placeholder_parser.py
berroar: Moved this now 👍

app/questionnaire/questionnaire_schema.py

ONS-Guilhem-Forey (approved):
LGTM

app/questionnaire/questionnaire_schema.py
```diff
@@ -935,3 +975,70 @@ def get_item_anchor(self, section_id: str, list_name: str) -> Optional[str]:
             for item in summary.get("items", []):
                 if item["for_list"] == list_name and item.get("item_anchor_answer_id"):
                     return f"#{str(item['item_anchor_answer_id'])}"
+
+    def get_block_ids_for_calculated_summary_dependencies(
```
MebinAbraham, line 979: This doesn't seem right with 2 circular dependencies, let's move it to a new file called `dependencies.py` in `app/questionnaire`?

MebinAbraham (approved):
👍

app/questionnaire/questionnaire_schema.py
berroar: Moved this now 👍

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
berroar: Type ignore comment was updated to say "Added to this method" - will resolve this comment now 👍

PR #1003, opened 2023-01-19
Fix bug: add to list is returning straight back to summary

ONS-Guilhem-Forey:
In a schema with a list collector and a summary, when the user completes the list collector block and lands on the summary page, but then adds another item to the list, they are redirected to the summary.

Instead, we want the user to be redirected to the list collector parent page, asking them if they want to add another item to the list collector.

This change also impacts the `change` and `remove` actions.

Thus here are the specs:
* If the user adds from the summary, return to the parent location
* If the user changes an item from the summary, return to the summary
* If the user removes an item from the summary:
    * if the change of list length introduces a new incomplete block, redirect to this block
    * Otherwise, go back to the summary 

### How to review

Smoke tests on the list collector schemas.
Changes in the way list collector schemas behave is expected. See integration and functional tests.

### DIFF
--- a/app/views/handlers/list_action.py
+++ b/app/views/handlers/list_action.py
@@ -34,11 +34,21 @@ class ListAction(Question):
         return True
 
     def get_previous_location_url(self):
-        if self._return_to == "section-summary":
+        if (
+            self._return_to == "section-summary"
+            and self.router.can_display_section_summary(
+                self.parent_location.section_id, self.parent_location.list_item_id
+            )
+        ):
             return self.get_section_summary_url()
 
         block_id = self._request_args.get("previous")
-        return self._get_location_url(block_id)
+        return self._get_location_url(
+            block_id=block_id,
+            return_to=self._return_to,
+            return_to_answer_id=self._return_to_answer_id,
+            return_to_block_id=self._return_to_block_id,
+        )
 
     def get_section_summary_url(self):
         return url_for(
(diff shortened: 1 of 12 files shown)


ONS-Guilhem-Forey:
[Bug: Add to list is returning straight back to summary](https://trello.com/c/QWMgvND7/5510-bug-add-to-list-is-returning-straight-back-to-summary)

berroar (approved):
Looks good, seems to have fixed the bug 👍

app/views/handlers/list_action.py
```diff
@@ -52,21 +52,31 @@ def get_next_location_url(self):
...
-        )
-        self.questionnaire_store_updater.remove_answers(answer_ids_to_remove)
-        self.evaluate_and_update_section_status_on_list_change(
-            self.parent_block["for_list"]
+
+        self._routing_path = self.router.routing_path(
+            self.current_location.section_id, self.current_location.list_item_id
         )
+
         self.questionnaire_store_updater.save()
```
MebinAbraham, line 80: We should only need to do this if any store has been updated.

```suggestion
        if self.questionnaire_store_updater.is_dirty():
            self._routing_path = self.router.routing_path(
                self.current_location.section_id, self.current_location.list_item_id
            )

            self.questionnaire_store_updater.save()
```

app/views/handlers/list_action.py
```diff
@@ -52,21 +52,31 @@ def get_next_location_url(self):
...
+        ):
+            return self.router.get_next_location_url(
+                self.parent_location,
+                self._routing_path,
+                self._return_to,
+                self._return_to_answer_id,
+                self._return_to_block_id,
+            )
+
         return self.parent_location.url()
```
MebinAbraham, line 68: This will need to persist the query params. A test will need to be added to ensure this is working.

```py
        return self.parent_location.url(
            return_to=self._return_to,
        )
```

app/views/handlers/list_add_question.py
```diff
@@ -9,13 +9,29 @@ def is_location_valid(self):
             return False
         return True
 
+    def get_next_location_url(self):
+        return self.parent_location.url()
```
MebinAbraham, line 13: This will need to persist the query params. A test will need to be added to ensure this is working. You will most likely need to add another block after the list collector to be able to test this.

```py
        return self.parent_location.url(
            return_to=self._return_to,
        )
```

app/views/handlers/list_action.py
ONS-Guilhem-Forey: Query params added

app/views/handlers/list_add_question.py
ONS-Guilhem-Forey: Query params added

petechd:
Maybe I missed something here but what is the "remove" action expected behaviour? From your PR description: 
`If the user adds or **removes** from the summary, **return to the parent location**`
 but then: `If the user **removes** an item from the summary:
if the change of list length introduces a new incomplete block, **redirect to this block**
Otherwise, **go back to the summary**`

ONS-Guilhem-Forey:
> * If the user adds from the summary, return to the parent location

Good shout, it's a mistake in the description.
Replaced
```
If the user adds or removes from the summary, return to the parent location
```

with

```
If the user adds from the summary, return to the parent location
```

petechd (approved):

MebinAbraham (commented):
The list collector page currently doesn't use the query param in the Change and Remove links. If you add from the summary, the url will have return to, but if then click Change or Remove from the parent page, then the query params is removed as those links haven't been rendered with the query params.

app/views/handlers/list_collector.py
```diff
@@ -39,6 +39,7 @@ def get_context(self):
                 for_list=self.rendered_block["for_list"],
                 edit_block_id=self.rendered_block["edit_block"]["id"],
                 remove_block_id=self.rendered_block["remove_block"]["id"],
+                return_to=self._return_to,
```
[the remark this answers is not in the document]
MebinAbraham: The return journey after using Change/Remove link will. need tests.

tests/integration/questionnaire/test_questionnaire_list_collector.py
```diff
@@ -320,3 +320,170 @@ def test_adding_person_using_second_list_collector_when_no_people(
...
+
+        self.assertInBody("List Collector Summary")
+
+        change_link = self.get_link("change", 1)
+
+        self.get(change_link)
+
+        self.assertInUrl("?return_to=section-summary")
+
+    def test_adding_from_the_summary_page_and_then_removing_from_parent_page_keeps_return_to_url_param(
```
ONS-Guilhem-Forey, line 409: @MebinAbraham This test and the one below ensure the URL param remains when changing and removing from the list collector, after adding from the summary.

MebinAbraham (commented):
The query param still isn’t persisted for some scenarios.
- If you say Add from Summary, then go to parent page, then say “Add” again, then query param isn’t persisted.
- I see you fixed the issue with the Change/Remove link not persisting, but you will need to update the Previous link also. If you click Change/Remove then click Previous, the query param isn’t persisted.

ONS-Guilhem-Forey:
> The query param still isn’t persisted for some scenarios.
> 
> * If you say Add from Summary, then go to parent page, then say “Add” again, then query param isn’t persisted.
> * I see you fixed the issue with the Change/Remove link not persisting, but you will need to update the Previous link also. If you click Change/Remove then click Previous, the query param isn’t persisted.

@MebinAbraham Addressed in the last 3 commits

MebinAbraham (approved):
👍

Yuyuutsu:
Tested the functionality and works as expected. Also had a look through the changes but out of my scope of knowledge to approve 👍

PR #1009, opened 2023-01-27
Placeholders should take into account answers on the path

Yuyuutsu:
Currently, placeholders do not take into account whether answers are on the path. Historically, this wasn't needed as it is not the placeholders' job to work that out, however, in v3 we added a placeholder transform `first_non_empty_item` which requires us to check that the answers are on the path. 

### How to review 
See if the error is still occurs : 
  - Launch `mbs_0106`
  - Click Start Survey
  - Submit `No, I need to report for a different period`
  - Enter a valid date range i.e 1 May 2016 to 20 May 2016
  - You should see the question has piped in your date range.
  - Go back to the first question and change your answer to `Yes, I can report for this period`
  - The next question will still pipe in your custom date range, but it should have piped in the date range from the metadata. 

- Check unit tests pass
- Check type hinting comments 

### Checklist

### DIFF
--- a/app/questionnaire/placeholder_renderer.py
+++ b/app/questionnaire/placeholder_renderer.py
@@ -99,16 +99,20 @@ class PlaceholderRenderer:
         return formatted_placeholder_data
 
     def render(
-        self, dict_to_render: Mapping[str, Any], list_item_id: Optional[str]
-    ) -> Mapping[str, Any]:
+        self, data_to_render: Mapping[str, Any], list_item_id: Optional[str]
+    ) -> dict[str, Any]:
         """
         Transform the current schema json to a fully rendered dictionary
         """
-        dict_to_render = QuestionnaireSchema.get_mutable_deepcopy(dict_to_render)
-        pointers = find_pointers_containing(dict_to_render, "placeholders")
+        data_to_render_mutable: dict[
+            str, Any
+        ] = QuestionnaireSchema.get_mutable_deepcopy(data_to_render)
+        pointers = find_pointers_containing(data_to_render_mutable, "placeholders")
 
         for pointer in pointers:
-            rendered_text = self.render_pointer(dict_to_render, pointer, list_item_id)
-            set_pointer(dict_to_render, pointer, rendered_text)
+            rendered_text = self.render_pointer(
+                data_to_render_mutable, pointer, list_item_id
+            )
+            set_pointer(data_to_render_mutable, pointer, rendered_text)
 
-        return dict_to_render
+        return data_to_render_mutable
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -494,7 +494,8 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
         self, section_id: str
     ) -> Optional[ImmutableDict]:
         if repeat := self.get_repeat_for_section(section_id):
-            return repeat.get("title")
+            title: ImmutableDict = repeat["title"]
+            return title
 
     def get_repeating_page_title_for_section(self, section_id: str) -> Optional[str]:
         if repeat := self.get_repeat_for_section(section_id):
@@ -929,7 +930,6 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
         if summary := self.get_summary_for_section(section_id):
             for item in summary.get("items", []):
                 if item["for_list"] == list_name and item.get("item_label"):
-
                     return str(item["item_label"])
 
     def get_item_anchor(self, section_id: str, list_name: str) -> Optional[str]:
--- a/app/views/contexts/calculated_summary_context.py
+++ b/app/views/contexts/calculated_summary_context.py
@@ -1,6 +1,6 @@
 from copy import deepcopy
 from decimal import Decimal
-from typing import Callable, List, Mapping, Tuple, Union
+from typing import Any, Callable, Iterable, Mapping, Tuple, Union
 
 from werkzeug.datastructures import ImmutableDict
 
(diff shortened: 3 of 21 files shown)


Yuyuutsu:
[Answer dependencies: Placeholders should take into account answers on the path (3 SP)](https://trello.com/c/QB5n5BGv/5407-answer-dependencies-placeholders-should-take-into-account-answers-on-the-path-3-sp)

app/questionnaire/placeholder_renderer.py
```diff
@@ -36,6 +37,8 @@ def __init__(
         self._response_metadata = response_metadata
         self._schema = schema
         self._location = location
+        self._block_on_routing_path = False
```
ONS-Guilhem-Forey, line 40: Could we rename it to `self._needs_block_on_routing_path` (or similar) to enforce its boolean type?
Even if it's initialised to False, I find it can get confusing when the variable is used elsewhere.
i.e "is it a block or a flag"?

app/questionnaire/placeholder_renderer.py
```diff
@@ -112,3 +126,36 @@ def render(
...
+    def check_transform(self, transform: MutableMapping[str, Any]) -> bool:
+        if transform.get("transform") in self._placeholder_transforms:
+            if items := transform["arguments"].get("items"):
+                for item in items:
+                    if self.check_item(item):
+                        return True
+
+        return False
+
+    def check_item(self, item: MutableMapping[str, Any]) -> bool:
```
ONS-Guilhem-Forey, line 139: Could we rename this specifically,  i.e
`is_item_on_routing_path()`

Also method could be private if we don't intend to use it outside of the placeholder renderer

app/questionnaire/placeholder_renderer.py
```diff
@@ -112,3 +126,36 @@ def render(
             set_pointer(dict_to_render, pointer, rendered_text)
 
         return dict_to_render
+
+    def check_transform(self, transform: MutableMapping[str, Any]) -> bool:
```
ONS-Guilhem-Forey, line 130: Could we rename this specifically, i.e
`check_transform_needs_routing_path()` or something similar

And make it private

I find it makes the code easier to understand

app/questionnaire/placeholder_renderer.py
```diff
@@ -66,6 +69,16 @@ def render_placeholder(
...
+
+        for placeholder in placeholder_data.get("placeholders", []):
+            if transforms := placeholder.get("transforms"):
+                for transform in transforms:
+                    if check_transform := self.check_transform(transform):
+                        self._block_on_routing_path = check_transform
+                        break
+            if self._block_on_routing_path:
+                break
+
```
ONS-Guilhem-Forey, line 81: We can probably extract this to its own method to keep `render_placeholder` segmented:
i.e
```python
    def _check_needs_block_on_routing_path(self, placeholder_data):
        for placeholder in placeholder_data.get("placeholders", []):
            if transforms := placeholder.get("transforms"):
                for transform in transforms:
                    if self.check_transform(transform):
                        return True

        return False
```

And in `render_placeholder()`:
```python
        self._needs_block_on_routing_path = self._needs_block_on_routing_path(
            placeholder_data
        )
```

app/questionnaire/placeholder_transforms.py
```diff
@@ -32,12 +32,14 @@ def __init__(
...
         schema: QuestionnaireSchema,
         renderer: "PlaceholderRenderer",
+        block_on_routing_path: bool = False,
     ):
         self.language = language
         self.schema = schema
         self.renderer = renderer
         self.locale = DEFAULT_LOCALE if language in {"en", "eo"} else language
         self.ops_helper = OperationHelper(self.locale, self.schema, self.renderer)
+        self.block_on_routing_path = block_on_routing_path
```
ONS-Guilhem-Forey, line 42: Could we rename this var in a boolean way too?

app/questionnaire/placeholder_transforms.py
```diff
@@ -297,8 +299,13 @@ def first_non_empty_item(self, items: Sequence[str]) -> str:
...
-            return item
+        MIN_LIST_LENGTH = 1
+        filtered_list = self.remove_empty_from_list(items)
+
+        for item in filtered_list:
+            if len(filtered_list) == MIN_LIST_LENGTH:
+                return item
+            return filtered_list[0] if self.block_on_routing_path else filtered_list[1]
 
         return ""
```
ONS-Guilhem-Forey, line 310: I don't really understand why the for loop is necessary here, and in what case we return "" below, since the iterator `item` isn't really used inside the for loop?

app/questionnaire/placeholder_renderer.py
Yuyuutsu: Yeah I agree sounds clearer . Added 👍

app/questionnaire/placeholder_renderer.py
Yuyuutsu: Updated and made private 👍

app/questionnaire/placeholder_renderer.py
Yuyuutsu: Updated and made private 👍

app/questionnaire/placeholder_transforms.py
Yuyuutsu: Updated the logic. So it first checks if the ` filtered list` is empty. If it is then it `returns ""`.
I have tested it, but could you also test it locally on runner to see if the solution still works. Thanks 👍

app/questionnaire/placeholder_transforms.py
Yuyuutsu: updated to `is_block_on_routing_path`

app/questionnaire/placeholder_renderer.py
Yuyuutsu: Extracted to own method with suggested solution. Changed the names of the functions.

ONS-Guilhem-Forey (approved):
Lgtm

MebinAbraham:
This will need some rework but it will be better to discuss it instead of leaving comments here. However, some points to consider are:
- Cross-sections dependency doesn't work.
- The routing path is reevaluated multiple times which is very inefficient, it should only be evaluated once per section. Right now every answer id revaluates the path.
- It doesn't take into account repeating sections. This should work within a repeat also.
- `first_non_empty_item` should be unchanged. With the latest change, it won't work when you have more than 2 items.

PR #1013, opened 2023-01-31
Split tests into matrix suites

petechd:
To speed up the Unit and Functional test suites that are running on Pull Request we split up Github Action matrix suites further.  Mainly Unit/Integration tests are affected, timeout modal is split into separate suites for each of its test. The longest running suite is features in functional tests now but it can be divided more if needed. Some tasks like linting and translations checks are moved to separate jobs and more dependancies are cached now (pipenv).

### How to review 
Run build to see how it works, `test-functional(timeout-modal)` and `test-unit` jobs are still present because they are required by other PRs. These need to be removed from required in Github settings.

### Checklist


MebinAbraham:
It looks like the virtual environment is being recreated for each of the matrix, I would test to see if you can use the cache from the initial python dependencies and node dependences install. That would save over a minute for each matrix.

petechd:
[Speed up runner GitHub Actions (3 SP)](https://trello.com/c/gWy87RqI/5511-speed-up-runner-github-actions-3-sp)

.github/workflows/pull_request.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ env.PYTHON_VERSION }}
+          cache: 'pipenv'
```
MebinAbraham, line 21: Do we need our own `Cache virtualenv` set if we can use built in caching?.

.github/workflows/pull_request.yml
```diff
@@ -66,6 +67,7 @@ jobs:
...
       - name: Install apt dependencies
```
MebinAbraham, line 71: Why do we install the apt dependencies multiple times?. Only needed for the unit tests.

.github/workflows/pull_request.yml
```diff
@@ -107,10 +148,50 @@ jobs:
         run: ./scripts/run_validator.sh
       - name: Running schema tests
         run: ./scripts/validate_test_schemas.sh
+  lint-javascript:
```
MebinAbraham, line 151: Any reason to not use existing `python-dependencies` and `node-dependencies`

MebinAbraham: This also isn't saving us much time, as you need python and node deps for the JS one also, since generate pages is python. I think the lint can just be a separate job, to unit test, but I would do python and js lint in same job as 2 steps (can be run in parallel).

.github/workflows/pull_request.yml
petechd: We can probably use built in caching but I didn't test it as it was cached the other way already.

.github/workflows/pull_request.yml
petechd: Removed, only installing for unit test now.

.github/workflows/pull_request.yml
petechd: UPDATE: Removed all virtualenv install instances.

.github/workflows/pull_request.yml
petechd: > Any reason to not use existing `python-dependencies` and `node-dependencies`

I deleted virtualenv install steps everywhere and cache pipenv using the newer syntax. I don't think there's much more we can improve here.

rmccar (approved):
Looks good, compared to previous runs the tests seem to have been sped up quite a bit

tests/functional/wdio.conf.js
```diff
@@ -18,7 +18,12 @@ exports.config = {
   //
   specs: ["./tests/functional/spec/**/*.js"],
   suites: {
-    timeout_modal: ["./tests/functional/spec/timeout/*.spec.js"],
+    timeout_modal_expired: ["./tests/functional/spec/timeout/timeout_modal_expired.spec.js"],
+    timeout_modal_extended: ["./tests/functional/spec/timeout/timeout_modal_extended.spec.js"],
+    timeout_modal_extended_new_window: ["./tests/functional/spec/timeout/timeout_modal_extended_new_window.spec.js"],
+    timeout_modal_post_submission_expired: ["./tests/functional/spec/timeout/timeout_modal_post_submission_expired.spec.js"],
+    timeout_modal_post_submission_extended: ["./tests/functional/spec/timeout/timeout_modal_post_submission_extended.spec.js"],
+    timeout_modal_post_submission_extended_new_window: ["./tests/functional/spec/timeout/timeout_modal_post_submission_extended_new_window.spec.js"],
```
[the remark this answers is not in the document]
petechd: The only problem is these will take longer, it will be consecutive 2 minutes waiting time times 3 plus test time so more than longest test we have now. I've tested these as 3 per spec before and there was no improvement compared to all of them in one spec if I recall correctly.

ONS-Guilhem-Forey (approved):
CI tested, changes look good to me
Aware of Mebin's last pending comment

tests/functional/wdio.conf.js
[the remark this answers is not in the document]
MebinAbraham: All of it together takes ~10 minutes, so how does splitting them into 3 also take ~10 minutes?.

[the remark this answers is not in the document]
petechd: it takes 6 when you split them into 3, each of them around 6 minutes.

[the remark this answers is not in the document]
MebinAbraham: Ya, which is good, given each on their own also takes 6 minutes. Split into 3, takes ~ 6 because it uses both cores. So all good then?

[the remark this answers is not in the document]
petechd: Ok, got it now, thanks.

.github/workflows/pull_request.yml
```diff
@@ -56,6 +46,49 @@ jobs:
...
+        with:
+          node-version: "14.19.0"
+      - name: Write app version
+        run: printf "${{ github.event.pull_request.head.sha }}" > .application-version
+      - name: Install pipenv
+        run: pip install pipenv==2022.11.30
+      - name: Install virtual environment
+        run: pipenv install --dev
+      - name: Load templates
+        run: make load-design-system-templates
```
MebinAbraham, line 70: Is this needed here?

.github/workflows/pull_request.yml
```diff
@@ -56,6 +46,49 @@ jobs:
...
+      - name: Compile translations
+        run: make translate
+      - name: Running translation tests
+        run: pipenv run python -m scripts.extract_translation_templates --test
+      - name: Running lint tests
```
MebinAbraham, line 75: Worth updating this and the JS to be similar, maybe just `Python liniting` and `Javascript linting`?

.github/workflows/pull_request.yml
```diff
@@ -56,6 +46,49 @@ jobs:
       - name: Install yarn deps
         if: steps.cache-yarn.outputs.cache-hit != 'true'
         run: yarn
+  lint:
+    needs: python-dependencies
```
MebinAbraham, line 50: Does this not need both py and js deps?

.github/workflows/pull_request.yml
```diff
@@ -110,7 +133,7 @@ jobs:
   test-functional:
```
MebinAbraham, line 131: How come functional doesn't need js/py deps?

.github/workflows/pull_request.yml
petechd: Not needed, removed, thanks.

.github/workflows/pull_request.yml
petechd: Changed them now, thanks.

.github/workflows/pull_request.yml
petechd: As above, added `needs`.

.github/workflows/pull_request.yml
```diff
@@ -56,6 +46,47 @@ jobs:
...
+        run: printf "${{ github.event.pull_request.head.sha }}" > .application-version
+      - name: Install pipenv
+        run: pip install pipenv==2022.11.30
+      - name: Install virtual environment
+        run: pipenv install --dev
+      - name: Compile translations
+        run: make translate
+      - name: Running translation tests
+        run: pipenv run python -m scripts.extract_translation_templates --test
+      - name: Python liniting
```
MebinAbraham, line 73: Spelling

PR #1015, opened 2023-01-31
Add support for answer codes in runner

berroar:

Adds support for user defined answer codes set in the schema to be passed downstream in the payload (applies only to data version `0.0.3`).

### How to review 

Use the new `test_answer_codes` schema to test the changes.

Check that the payload generated matches the spec outlined in [the example downstream survey response in this document](https://github.com/ONSdigital/ons-schema-definitions/blob/main/examples/eq_runner_to_downstream/payload_v1/surveyresponse_0_0_3_with_answer_codes.json)

### Checklist


berroar:
[Implement answer_codes in runner and validator (5 SP)](https://trello.com/c/5k8NixXD/5515-implement-answercodes-in-runner-and-validator-5-sp)

app/submitter/converter_v2.py
```diff
@@ -134,4 +136,22 @@ def get_payload_data(
             "lists": list_store.serialize(),
         }
 
+        if answer_codes := schema.json.get("answer_codes"):
+            get_filtered_answer_codes(answer_codes, data)
```
MebinAbraham, line 140: - The function `get_filtered_answer_codes` implies you get sometime back but it is mutating which is always good to avoid. I would update this to return the answer codes then assign `data["answer_codes"] = filtered_answer_codes` here. 

- It's input should just be, `answer_codes` and `answer_ids_to_keep`. It should only be passed what it needs.

- Answer codes should only be assigned if we have at least one. So empty answer codes (because we have no answers) shouldn't be assigned. We can look at changing "answers" and "lists" to match in future as no need to add keys that are empty. But currently validation for answer_codes enforce, it must have at least one item.

schemas/test/en/test_answer_codes.json
```diff
@@ -0,0 +1,169 @@
...
+                {
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "mandatory-checkbox",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "mandatory-checkbox-answer",
+                                        "mandatory": true,
```
MebinAbraham, line 89: I would make this voluntary so we can test, no answer codes are sent if no answer.

app/submitter/converter_v2.py
```diff
@@ -113,7 +113,7 @@ def get_payload_data(
     routing_path: RoutingPath,
     metadata: MetadataProxy,
     response_metadata: Mapping,
-) -> Union[OrderedDict[str, Any], dict[str, Union[list[Any]]]]:
+) -> Union[OrderedDict[str, Any], dict[str, Union[list[Any], object]]]:
```
MebinAbraham, line 116: Why is it returning `object`?

app/submitter/converter_v2.py
```diff
@@ -134,4 +136,22 @@ def get_payload_data(
...
+        return data
+
     raise DataVersionError(schema.json["data_version"])
+
+
+def get_filtered_answer_codes(
+    answer_codes: Iterable[OrderedDict[str, Any]], data: dict
+) -> None:
+    answer_ids_to_filter = {answer.answer_id for answer in data["answers"]}
+    filtered_answer_codes: list = []
```
MebinAbraham, line 151:
```suggestion
    filtered_answer_codes: list[dict] = []
```

app/submitter/converter_v2.py
```diff
@@ -134,4 +136,22 @@ def get_payload_data(
...
+        if answer_codes := schema.json.get("answer_codes"):
+            get_filtered_answer_codes(answer_codes, data)
+
+        return data
+
     raise DataVersionError(schema.json["data_version"])
+
+
+def get_filtered_answer_codes(
+    answer_codes: Iterable[OrderedDict[str, Any]], data: dict
```
MebinAbraham, line 148: answer_codes: Iterable Mapping?. Ordered dict isn't important for this?.

schemas/test/en/test_answer_codes.json
```diff
@@ -0,0 +1,169 @@
...
+    "data_version": "0.0.3",
+    "survey_id": "0",
+    "title": "Other input fields",
+    "theme": "default",
+    "description": "A questionnaire to demo answer codes.",
+    "messages": {
+        "NUMBER_TOO_LARGE": "Number is too large",
+        "NUMBER_TOO_SMALL": "Number cannot be less than zero",
+        "INVALID_NUMBER": "Please enter an integer"
+    },
```
MebinAbraham, line 14: Can be removed.

MebinAbraham (commented):
Just some minor issues.

schemas/test/en/test_answer_codes.json
berroar: Removed now

app/submitter/converter_v2.py
berroar: Updated this 👍

app/submitter/converter_v2.py
berroar: Updated the method now 👍

app/submitter/converter_v2.py
berroar: I had to make it an iterable dict as otherwise mypy was complaining about the .extend method

app/submitter/converter_v2.py
berroar: Can't seem to fix the issue here at the moment, mypy seems to think the generation of answers/lists in the data payload can be returned as an object 😐

schemas/test/en/test_answer_codes.json
berroar: Updated this 👍

app/submitter/converter_v2.py
berroar: OK, removed this now 👍

app/submitter/converter_v2.py
```diff
@@ -123,15 +123,39 @@ def get_payload_data(
...
-                full_routing_path=routing_path,
-            ),
+        answers = convert_answers_to_payload_0_0_3(
+            answer_store=answer_store,
+            list_store=list_store,
+            schema=schema,
+            full_routing_path=routing_path,
+        )
+
+        data: dict[str, Union[list[Any]]] = {
```
MebinAbraham, line 135:
```suggestion
        data: dict[str, list[dict]] = {
```

app/submitter/converter_v2.py
```diff
@@ -123,15 +123,39 @@ def get_payload_data(
...
+            ):
+                data["answer_codes"] = filtered_answer_codes
+
+        return data
+
     raise DataVersionError(schema.json["data_version"])
+
+
+def get_filtered_answer_codes(
+    answer_codes: Iterable[dict], answer_ids_to_filter: set[str]
```
MebinAbraham, line 153: I suggest enforcing keyword args so these can't be passed wrong way around.

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=support-answer-codes
```

tests/app/submitter/test_convert_payload_0_0_3.py
```diff
@@ -1351,3 +1351,79 @@ def test_relationship_answers_not_on_path_in_payload(version):
     assert ("related-to-anyone-else-answer", "person1") in answers
     relationships_answer = answers[("relationship-answer", None)]
     assert expected_relationships_answer == relationships_answer["value"]
+
+
+@pytest.mark.parametrize(
```
MebinAbraham, line 1356: I would add a test to ensure the `answer_codes` property isn't in the payload when no answers

schemas/test/en/test_answer_codes.json
```diff
@@ -0,0 +1,164 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
+    "schema_version": "0.0.1",
+    "data_version": "0.0.3",
+    "survey_id": "0",
+    "title": "Other input fields",
```
MebinAbraham, line 7: Needs updating.

berroar: Updated this 👍

app/submitter/converter_v2.py
berroar: Enforcing now 👍

MebinAbraham (approved):

ONS-Guilhem-Forey (approved):
LGTM

tests/app/submitter/test_convert_payload_0_0_3.py
```diff
@@ -1351,3 +1352,189 @@ def test_relationship_answers_not_on_path_in_payload(version):
...
+    "version",
+    (
+        None,
+        AuthPayloadVersion.V2,
+    ),
+)
+def test_all_answers_codes_for_list_collector_questions(version):
```
MebinAbraham, line 1440: I'm not convinced we need this. There is no special behaviour in runner for list collector and answer codes. Testing of list collectors needs to be in validator (if not already). The schema can say, to prove it passes validation.

berroar: Removed the test 👍

MebinAbraham (approved):

Hajara-I:
Not yet ready to approve the PR but have read through and works as expected.

PR #1018, opened 2023-02-01
Bug Fix logger values not being bound to logs

petechd:
Since vars are not binding to logger, Structlog needs to be pinned to older version, along with flask-babel, black and pylint which cause some issues when unpinned after dependencies update. This pr only temporarily pins them so the logs issue is fixed.

### How to review 
Check if runner is behaving as expected after dependencies update.

### Checklist


petechd:
[Bug: Runner logger binds are not being bound](https://trello.com/c/cKQnfkrq/5555-bug-runner-logger-binds-are-not-being-bound)

berroar (approved):
Looks like the missing logs are back 👍

MebinAbraham:
Can you update the PR title to what the change actually does.

ONS-Guilhem-Forey (approved):
Tested locally with updated py version and packages, and seems to be working for me too 👍

PR #1020, opened 2023-02-07
Bug: Flush always using v1 converter

creativeappdevamit:
[Trello](https://trello.com/c/mqAs98GC/5556-phm-flush-using-v1-instead-of-v2-which-is-failing-over-validating-in-sdx)

**Comment From Developer**
I was not sure how to test this change, the unit tests covers the actual convertors already. Hence used mock to just insure that the appropriate converter is called based on the launch version selected.

### DIFF
--- a/app/routes/flush.py
+++ b/app/routes/flush.py
@@ -5,11 +5,13 @@ from sdc.crypto.decrypter import decrypt
 from sdc.crypto.encrypter import encrypt
 from structlog import get_logger
 
+from app.authentication.auth_payload_version import AuthPayloadVersion
 from app.authentication.user import User
 from app.globals import get_answer_store, get_metadata, get_questionnaire_store
 from app.keys import KEY_PURPOSE_AUTHENTICATION, KEY_PURPOSE_SUBMISSION
 from app.questionnaire.router import Router
 from app.submitter.converter import convert_answers
+from app.submitter.converter_v2 import convert_answers_v2
 from app.submitter.submission_failed import SubmissionFailedException
 from app.utilities.json import json_dumps
 from app.utilities.schema import load_schema_from_metadata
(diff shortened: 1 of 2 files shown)


creativeappdevamit:
[PHM Flush using V1 instead of V2 which is failing over validating in SDX](https://trello.com/c/mqAs98GC/5556-phm-flush-using-v1-instead-of-v2-which-is-failing-over-validating-in-sdx)

app/routes/flush.py
```diff
@@ -76,7 +78,15 @@ def _submit_data(user):
...
-            convert_answers(
+            convert_answers_v2(
+                schema,
+                questionnaire_store,
+                full_routing_path,
+                submitted_at,
+                flushed=True,
+            )
+            if metadata.version is AuthPayloadVersion.V2
+            else convert_answers(
```
MebinAbraham, line 89: Minor, instead of defining the function argument twice, work out the method to call upfront. i.e

```py
        converter = (
            convert_answers_v2
            if metadata.version is AuthPayloadVersion.V2
            else convert_answers
        )

        message = json_dumps(
            converter(
                schema,
                questionnaire_store,
                full_routing_path,
                submitted_at,
                flushed=True,
            )
        )
```

MebinAbraham (commented):
Could you update the title to be something like `Bug: Flush always using v1 converter`

app/routes/flush.py
creativeappdevamit: Hi Mebin - I agree but I was not able create another local as the lint fails as the  methods seems to have too may locals. Allowed is fifteen, me adding this will make it 16 and will have to apply an exception. I also like this style :-) makes it easier to understand whats happening :-)

MebinAbraham: Hey Amit, I think we address it otherwise, we are leaving the problem to the next person to deal with. Parts of it could be refactored into method or in this case I am happy to ignore it. Will leave it to you on whether you refactor or ignore the warning.

creativeappdevamit: 👍  refactored!

ONS-Guilhem-Forey (approved):
Lgtm

PR #1024, opened 2023-02-08
Update Runner logger binds to use contextvars and unpin dependencies 

Yuyuutsu:
Runner is currently using `structlog.threadlocal` class which is deprecated. We used `threadlocal` as a quick fix because runner logger binds was not binding hence we got no `tx_ids`. We also had to pin some dependencies for runner to function.
This PR adds the up to date  `structlog.contextvars` class . Also unpins the pinned dependencies from the previous [PR:](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1018)
  - Flask-Babel
  - Pylint
  - Black
 
We have added a new test to test for logger outputs. `test_get_request_logs_output`
### How to review 
Run , runner locally and launch any schema. Go through the schema and check in the Run Tab to see if all the variables are being logged. IE: `tx_id`, `request_ids`.
Run tests and see if they pass
Check updated `raise errors` and see if they are correctly named and used
Have a look at the reformatted `def format_list`  using str.join (suggested by pylint). See if the readability is affected by change

### Checklist


Yuyuutsu:
[Bug: Runner logger binds are not being bound](https://trello.com/c/cKQnfkrq/5555-bug-runner-logger-binds-are-not-being-bound)

app/authentication/authenticator.py
```diff
@@ -109,7 +109,7 @@ def load_user(extend_session: bool = True) -> Optional[User]:
         user = User(user_id, user_ik)
 
         if metadata := get_metadata(user):
-            logger.bind(tx_id=metadata.tx_id)
+            contextvars.bind_contextvars(tx_id=metadata.tx_id)
```
berroar, line 112: Is tx_id the only var that was missing? When I compare the new output with master I notice that `request_id` is no longer in the output either

app/setup.py
```diff
@@ -326,7 +340,9 @@ def setup_submitter(application):
         bucket_name = application.config.get("EQ_GCS_SUBMISSION_BUCKET_ID")
 
         if not bucket_name:
```
berroar, line 342: Could be simplified with a conditional e.g `if bucket_name :=` and then `else`. so we can remove `if not`

app/setup.py
```diff
@@ -376,15 +394,17 @@ def setup_publisher(application):
...
     else:
-        raise Exception("Unknown EQ_PUBLISHER_BACKEND")
+        raise UnknownEnviormentVariable("Unknown EQ_PUBLISHER_BACKEND")
 
 
 def setup_feedback(application):
     if application.config["EQ_FEEDBACK_BACKEND"] == "gcs":
         bucket_name = application.config.get("EQ_GCS_FEEDBACK_BUCKET_ID")
 
         if not bucket_name:
```
berroar, line 404: Same as above, this could be simplified

app/authentication/authenticator.py
petechd: There were few missing, all that was attached using logger.bind() in `get_questionnaire` method, we test the missing ones now in here:
```
        for output in logs.output:
            self.assertIn("questionnaire request", output)
            self.assertIn("tx_id", output)
            self.assertIn("ce_id", output)
            self.assertIn("schema_name", output)
            self.assertIn("method", output)
            self.assertIn("url_path", output)
            self.assertIn("request_id", output)
```
`type="questionnare request" is the new "questionnaire request" as contextvars's bind only accepts key-value pairs. `request_id` now added in `before_request` method and present in logs. Thanks!

app/questionnaire/placeholder_transforms.py
```diff
@@ -60,14 +60,12 @@ def format_date(self, date_to_format: str, date_format: str) -> str:
     @staticmethod
     def format_list(list_to_format: Sequence[str]) -> str:
         formatted_list = "<ul>"
-        for item in list_to_format:
-            formatted_list += f"<li>{item}</li>"
+        formatted_list += "".join([f"<li>{item}</li>" for item in list_to_format])
```
berroar, line 63:
```
        formatted_list = "<ul>" + "".join(
            [f"<li>{item}</li>" for item in list_to_format]
        )
```

Yuyuutsu: Updated the `format_list` to your suggested code👍

app/setup.py
Yuyuutsu: Updated and  if not is removed 👍

app/questionnaire/placeholder_transforms.py
```diff
@@ -59,15 +59,14 @@ def format_date(self, date_to_format: str, date_format: str) -> str:
 
     @staticmethod
     def format_list(list_to_format: Sequence[str]) -> str:
-        formatted_list = "<ul>"
-        for item in list_to_format:
-            formatted_list += f"<li>{item}</li>"
+        formatted_list = "<ul>" + "".join(
+            [f"<li>{item}</li>" for item in list_to_format]
```
MebinAbraham, line 63: This list comp isn't needed, can just be:
```py
"".join(f"<li>{item}</li>" for item in list_to_format)
```

app/routes/questionnaire.py
```diff
@@ -85,19 +85,21 @@ def before_questionnaire_request():
...
-        logger.bind(schema_name=schema_name)
+        contextvars.bind_contextvars(schema_name=schema_name)
 
     if schema_url := metadata.schema_url:
-        logger.bind(schema_url=schema_url)
+        contextvars.bind_contextvars(schema_url=schema_url)
 
-    logger.info(
-        "questionnaire request", method=request.method, url_path=request.full_path
+    contextvars.bind_contextvars(
```
MebinAbraham, line 99: How come this has changed from `logger.info`?

app/routes/questionnaire.py
```diff
@@ -132,16 +134,18 @@ def before_post_submission_request():
...
```
MebinAbraham, line 145: Same as above.

app/routes/questionnaire.py
```diff
@@ -363,6 +367,7 @@ def relationships(
 @with_session_store
 @with_schema
 def get_thank_you(schema, session_store, questionnaire_store):
+    contextvars.clear_contextvars()
```
MebinAbraham, line 370: How come this is now needed here?.

app/setup.py
```diff
@@ -132,11 +140,15 @@ def create_app(  # noqa: C901  pylint: disable=too-complex, too-many-statements
     @application.before_request
     def before_request():  # pylint: disable=unused-variable
         request_id = str(uuid4())
-        logger.new(request_id=request_id)
+
+        if flask_request.method == "OPTIONS":
+            contextvars.clear_contextvars()
```
MebinAbraham, line 145: What is the reasoning behind this?. Why clear for OPTIONS?

app/setup.py
```diff
@@ -225,6 +237,8 @@ def after_request(response):  # pylint: disable=unused-variable
         # We're using the stringified version of the Flask session to get a rough
         # length for the cookie. The real length won't be known yet as Flask
         # serializes and adds the cookie header after this method is called.
+        if flask_request.method == "OPTIONS":
+            contextvars.clear_contextvars()
```
MebinAbraham, line 241: Same as above, why is this now needed?. Is there a link to the doc that explains this?.

app/setup.py
```diff
@@ -323,21 +337,23 @@ def setup_redis(application):
...
-        if not bucket_name:
-            raise Exception("Setting EQ_GCS_SUBMISSION_BUCKET_ID Missing")
-
-        application.eq["submitter"] = GCSSubmitter(bucket_name=bucket_name)
+        if bucket_name := application.config.get("EQ_GCS_SUBMISSION_BUCKET_ID"):
+            application.eq["submitter"] = GCSSubmitter(bucket_name=bucket_name)
+        else:
+            raise MissingEnvironmentVariable(
+                "Setting EQ_GCS_SUBMISSION_BUCKET_ID Missing"
+            )
```
MebinAbraham, line 345: Minor, but I think the previous way is clearer. We should escape early to reduce indentation, this changes increases the indentation. You can still exist early and make use of walrus.

app/setup.py
```diff
@@ -376,24 +392,24 @@ def setup_publisher(application):
...
     if application.config["EQ_FEEDBACK_BACKEND"] == "gcs":
-        bucket_name = application.config.get("EQ_GCS_FEEDBACK_BUCKET_ID")
-
-        if not bucket_name:
-            raise Exception("Setting EQ_GCS_FEEDBACK_BUCKET_ID Missing")
-
-        application.eq["feedback_submitter"] = GCSFeedbackSubmitter(
-            bucket_name=bucket_name
-        )
+        if bucket_name := application.config.get("EQ_GCS_FEEDBACK_BUCKET_ID"):
```
MebinAbraham, line 400: Same s above.

app/setup.py
```diff
@@ -448,11 +463,14 @@ def get_locale():  # pylint: disable=unused-variable
 
         return None
 
-    @application.babel.timezoneselector
     def get_timezone():  # pylint: disable=unused-variable
         # For now regardless of locale we will show times in GMT/BST
         return "Europe/London"
 
+    application.babel.init_app(
```
MebinAbraham, line 470: Can you provide link to the doc, that shows this update?.

Yuyuutsu: https://pyup.io/packages/pypi/flask-babel/changelog#:~:text=%2D%20Babel.locale_selector%20and%20Babel.timezone_selector%20no%20longer%20exist.%20They%20must%20be%20provided%0Aeither%20when%20the%20Babel()%20object%20is%20created%20or%20when%20init_app()%20is%20called.%20This%20is%20to%20support%0Ahaving%20a%20single%20Babel%20object%20for%20multiple%20Flask%20apps%20(107)%20as%20well%20as%20to%20simplify%20settings%0Aand%20multi%2Dthreaded%20state.

app/questionnaire/placeholder_transforms.py
Yuyuutsu: Would we not need the `<ul> </ul>` anymore?

app/setup.py
Yuyuutsu: Before every request we need to clear the contextvars bind. If we did not clear for the request `OPTIONS ` the contextvars wont bind.

app/setup.py
Yuyuutsu: Here is the documentation highlighting the [part](https://www.structlog.org/en/stable/contextvars.html#:~:text=Call%20structlog.contextvars.clear_contextvars()%20at%20the%20beginning%20of%20your%20request%20handler%20(or%20whenever%20you%20want%20to%20reset%20the%20context%2Dlocal%20context).).

app/routes/questionnaire.py
petechd: I got carried away here and started adding some unnecessary logger changes, I compared it with master and the logs are working as expected without this change so reverted it.

app/routes/questionnaire.py
petechd: Removed, thanks!

app/setup.py
petechd: This was for some unit tests that were not staged and committed, reverted now.

app/setup.py
petechd: Same as above, not needed here, removed now.

app/setup.py
Yuyuutsu: Changed back to the old version but refactored to use walrus operator 👍

app/setup.py
Yuyuutsu: Changed back to the old version but refactored to use walrus operator 👍

app/setup.py
Yuyuutsu: UPDATE: changed back to old way suggested by Mebin, but implemented walrus operator 👍

app/setup.py
Yuyuutsu: Updated : Mebin suggested to keep the if not. But i have added walrus operator to the if not 👍

app/questionnaire/placeholder_transforms.py
Yuyuutsu: Changed 👍

app/setup.py
```diff
@@ -439,7 +450,6 @@ def setup_babel(application):
     application.babel = Babel(application)
     application.jinja_env.add_extension("jinja2.ext.i18n")
 
-    @application.babel.localeselector
     def get_locale():  # pylint: disable=unused-variable
```
MebinAbraham, line 453: This no longer needs to be a method inside the `create_app`

app/setup.py
```diff
@@ -448,11 +458,14 @@ def get_locale():  # pylint: disable=unused-variable
 
         return None
 
-    @application.babel.timezoneselector
     def get_timezone():  # pylint: disable=unused-variable
```
MebinAbraham, line 461: This no longer needs to be a method inside the `create_app`

app/routes/questionnaire.py
MebinAbraham: But is this not needed, in the before request of each endpoint?. As per the doc, Yuyu linked in https://github.com/ONSdigital/eq-questionnaire-runner/pull/1024#discussion_r1104147878?

petechd: From looking at the logs and comparing them to what we have on master it seems like it is not needed. structlog's `BoundLogger` has [unbind()](https://www.structlog.org/en/stable/api.html#structlog.BoundLogger.unbind) that we never use, so I didn't add [contextvars. unbind_contextvars()](https://www.structlog.org/en/stable/api.html#structlog.contextvars.unbind_contextvars). Seems like [clear_contextvars()](https://www.structlog.org/en/stable/api.html#structlog.contextvars.clear_contextvars) is only optional, we can add it in the before request handler if we want to "reset the context-local context" [docs](https://www.structlog.org/en/stable/contextvars.html) but do we want to?

app/setup.py
```diff
@@ -491,3 +479,16 @@ def get_minimized_asset(filename):
...
+
+def get_locale():  # pylint: disable=unused-variable
+    session = get_session_store()
+
+    if session and (session_data := session.session_data):
+        return session_data.language_code
+
+    return None
+
+def get_timezone():  # pylint: disable=unused-variable
```
MebinAbraham, line 493: The pylint disable statement still needed?

app/routes/questionnaire.py
MebinAbraham: Need to read up on the usage of `clear_contextvar`, it could be new and might be required. On the surface it may see to work but may have mixed context if we don't implement it. So have a look through the docs/online resource on the usage of it and when it is required.

app/setup.py
Yuyuutsu: Removed the pylint disable comments, Pylint test passed 👍

app/setup.py
Yuyuutsu: `get_locale()` is removed from the create app object

app/routes/questionnaire.py
petechd: Ok, so I've checked the docs, looked at our implementation and it turns out assertions in our tests were badly written and `clear_contextvars()` was indeed needed at the very initial before_request. Now it does behave as the main branch logs and clears keys/values like `tx_d` before the new request. `clear_contextvars()` is added as described in the docs `The typical use-case for this function is to invoke it early in request- handling code` so now no previous request context is used. When compared to `main` it produces the same outputs.

MebinAbraham: Nice one. If not already, can you ensure this is added to every before request hander. We have multiple blueprints handlers.

app/setup.py
```diff
@@ -356,7 +369,7 @@ def setup_submitter(application):
         application.eq["submitter"] = LogSubmitter()
 
     else:
-        raise Exception("Unknown EQ_SUBMISSION_BACKEND")
+        raise UnknownEnviormentVariable("Unknown EQ_SUBMISSION_BACKEND")
```
MebinAbraham, line 372: These don't need a custom one, this can use the builtin `NotImplementedError`

MebinAbraham (commented):
The pylint changes muddies this PR which makes it hard to see just the logger changes, is it possible to split or will the dependency update conflict?.

tests/integration/test_app_create.py
```diff
@@ -89,8 +89,8 @@ def test_adds_logging_of_span_and_trace(self):
             }
             application.test_client().get("/", headers=x_cloud_headers)
 
-            self.assertEqual(1, logger.bind.call_count)
-            _, kwargs = logger.bind.call_args
+            self.assertEqual(2, bind_contextvars.call_count)
```
MebinAbraham, line 92: How come the count increased?.

Yuyuutsu:
> The pylint changes muddies this PR which makes it hard to see just the logger changes, is it possible to split or will the dependency update conflict?.

We should be able to put pylint chnages in another PR and pin pylint here. It shouldn't affect the context vars changes

tests/integration/test_app_create.py
petechd: We replaced `logger.new(request_id=request_id)` we do `contextvars.bind_contextvars(request_id=request_id)` now.

MebinAbraham: How come that increases the count?.

petechd: it calls `bind_contextvars()` twice now in setup and then when making a request?  before it was `bind()` once and `new()` once?

petechd: and we only counted calls of `logger.bind()`?

app/routes/questionnaire.py
petechd: I am not sure if we can add this in other before request handlers as it clears  context vars,  like `request_id` for `questionnaire request` logger.info for example.

tests/integration/test_app_create.py
MebinAbraham: Okay, so there is no replacement for `logger.new`?. Is that deprecated?. Is there some doc that mentions this?.

petechd: [doc](https://www.structlog.org/en/stable/api.html#structlog.BoundLoggerBase.new), it clears context and binds initial_values using [bind](https://www.structlog.org/en/stable/api.html#structlog.BoundLoggerBase.bind) so that's why we had to remove it, cause w don't use the old `.bind()` anymore.

app/routes/questionnaire.py
MebinAbraham: Okay ya, that's good, that before request is app level which is call for every endpoint. 👍

MebinAbraham:
> > The pylint changes muddies this PR which makes it hard to see just the logger changes, is it possible to split or will the dependency update conflict?.
> 
> We should be able to put pylint chnages in another PR and pin pylint here. It shouldn't affect the context vars changes. Should we do that ?

Leave it as is now.

tests/integration/routes/test_questionnaire.py
```diff
@@ -53,3 +53,21 @@ def test_options_request_before_request(self):
...
+    def test_get_request_logs_output(self):
+        self.launchSurvey("test_hub_and_spoke")
+        with self.assertLogs() as logs:
+            self.get("/questionnaire/")
+            self.assertStatusOK()
+
+            request_log = logs.output[0]
+            questionnaire_request_log = logs.output[1]
+
+            self.assertNotIn("tx_id", request_log)
```
MebinAbraham, line 66: This should also assert things that should be in request log.

app/questionnaire/when_rules.py
```diff
@@ -228,7 +231,7 @@ def _get_when_rule_value(
     elif "list" in when_rule:
         value = get_list_count(list_store, when_rule["list"])
     else:
-        raise Exception("The when rule is invalid")
+        raise InvalidWhenRule("The when rule is invalid")
```
MebinAbraham, line 234: This can be `NotImplementedError`

MebinAbraham (approved):
👍

tests/integration/routes/test_questionnaire.py
petechd: Added, thanks!

app/questionnaire/when_rules.py
```diff
@@ -12,6 +12,10 @@
 logger = logging.getLogger(__name__)
 
 
+class InvalidWhenRule(Exception):
```
MebinAbraham, line 15: Can be removed

berroar (approved):

app/questionnaire/when_rules.py
petechd: Removed.

PR #1026, opened 2023-02-09
Update python type hints style guide

Hajara-I:
Update our Python Type hints style guide.

### How to review 
Read through style guide markdown file to verify updates.

### Checklist


Hajara-I:
[Update our Python Type hints style guide (1 SP)](https://trello.com/c/cUz06U4y/5547-update-our-python-type-hints-style-guide-1-sp)

doc/python-type-hinting.md
```diff
@@ -80,8 +78,41 @@ This import is not necessary in Python 3.10.
...
+
+To mark portions of the program that should not be covered by type hinting, use the following on a particular line:
+
+```python
+# type: ignore
+```
+
+A `# type: ignore` comment on a line by itself at the top of a file silences all errors in the file.
+
+`# type: ignore` should only be used when unavoidable. Ensure that a comment is added to explain why it has been used.
```
Yuyuutsu, line 91: May want to specifically state that the comment used to explain the type ignore needs a prefix of `Type ignore:` so that we can tell what the comment is for and allow us to easily find it future.
For example:
"Ensure that a comment is added to explain why it has been used and have a prefix of `Type ignore:`"

doc/python-type-hinting.md
```diff
@@ -80,8 +78,41 @@ This import is not necessary in Python 3.10.
...
+
+## ParamSpec
+
+Used to forward the parameter types of one callable to another callable e.g. to add basic logging to a function, create a decorator add_logging to log function calls:
+
+```python
+T = TypeVar('T')
```
petechd, line 98: More of a question to all the reviewers. As I mentioned in tech session, do we want to restrict types in TypeVar? Some explanation:

Screenshot 2023-02-13 at 09 59 31

petechd: Also do we want to enforce [NoReturn](https://docs.python.org/3/library/typing.html#typing.NoReturn) instead of `None`? It will be replaced by [Never](https://docs.python.org/3/library/typing.html#typing.Never) in python 3.11.

doc/python-type-hinting.md
Hajara-I: Agreed.

doc/python-type-hinting.md
```diff
@@ -40,18 +40,16 @@ def get_objects_matching(ids: Sequence[str]) -> dict[int, dict]
...
-def test(self, var: Optional[int]) -> None:
+def test(self, var: [None | int]) -> None:
 ```
 
-For arguments that can be one of multiple types or `None`, use `Union` with `None`:
+For arguments that can be one of multiple types or None, use the shorthand `|`. This is recommended instead of using older `Union` keyword:
 
 ```python
-def test(self, var: Union[None, int, str]) -> None:
+def test(self, var: [None | int | str]) -> None:
```
MebinAbraham, line 52: I'm of the opinion that `Optional` is clearer when there is only one type.

This:
```py
var: Optional[int]
```

instead of: 

```py
var: int | None
```

---

But, this: 

```py
var: int | str | None
``` 

instead of: 

```py
var: Union[Optional[int], Optional[str]]
```

See what the others think. Happy to go with the consensus.

doc/python-type-hinting.md
```diff
@@ -80,8 +78,41 @@ This import is not necessary in Python 3.10.
...
+# type: ignore
+```
+
+A `# type: ignore` comment on a line by itself at the top of a file silences all errors in the file.
+
+`# type: ignore` should only be used when unavoidable. Ensure that a comment is added to explain why it has been used and have a prefix of Type ignore:.
+
+## ParamSpec
+
+Used to forward the parameter types of one callable to another callable e.g. to add basic logging to a function, create a decorator add_logging to log function calls:
```
MebinAbraham, line 95:
```suggestion
Used to forward the parameter types of one callable to another callable. This is useful when type annotating decorators.

 For example, a basic logging function decorator can be typed as:
```

doc/python-type-hinting.md
MebinAbraham: As discussed, `NoReturn` isn't a replacement for `None`.

doc/python-type-hinting.md
MebinAbraham: I don't think this has been done?

MebinAbraham (commented):
The doc around `Forward declarations` are now redundant in 3.10. We should replace it with the `Self` type. https://peps.python.org/pep-0673/

doc/python-type-hinting.md
ONS-Guilhem-Forey: Don't have a preference at first glance although:

It would seem Python's core development team [recommends](https://stackoverflow.com/a/69440627) using `Type | None` over `Optional` for Python 10+.
Thus I suspect the piping symbol will be widely adopted moving forward and `Optional` will become outdated.

Using the piping symbol will make it easier for new joiners to get familiar with the code base.

doc/python-type-hinting.md
ONS-Guilhem-Forey: I am in favour of encouraging TypeVar restricting where possible, as it improves the code documentation and error prevention.
However I think there might be legitimate cases for an unrestricted TypeVar, typically a very generic util function.

doc/python-type-hinting.md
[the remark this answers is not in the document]
MebinAbraham: For this one, this comes from the option which we disabled called `implict_optional`. As recommended and following zen of Python "explict is better than implict" is why this is enforced. The original PEP discourages this. https://peps.python.org/pep-0484/#union-types

From the PEP:

A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:
```py
def handle_employee(e: Employee = None): ...
```

This would have been treated as equivalent to:
```py
def handle_employee(e: Optional[Employee] = None) -> None: ...
```
This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

MebinAbraham: My view is that, I think the repeated `None` makes it less readable. The official doc doesn't mention the deprecation of it or preference. I'm only in favour of using `Optional` if it is a single type. If you have 2 types + None, then the new operator would be best.

To me, this looks isn't very clear:

```py
def test(self, var: int | None = None) -> None:
```

compared to:

```py
def test(self, var: Optional[int] = None) -> None:
```

Again, happy to go with consensus if others think it is readable.

doc/python-type-hinting.md
Hajara-I: I have now updated this section as suggested above

doc/python-type-hinting.md
Hajara-I: New section now added for `TypeVar` with examples on how to implement it.

doc/python-type-hinting.md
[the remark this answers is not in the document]
Hajara-I: As this is no longer the recommended behaviour, I understand that I do not need to add a section on this to the documentation?

petechd: I prefer if we only use `|`, I find it more readable, even in `int | None` situation, but up to you guys.

Yuyuutsu: Looking at the examples I like the idea of using the `|` :
`int | str | None` 

But when there default types
`Optional[t] = None`
is easier to read to me then:
`t | None = None`

MebinAbraham: Okay, in that case let's go with the single use of optional using `pipe` operator i.e `Type | None = None`

MebinAbraham (commented):
I think we should add a section about `TypeAlias`. It was introduced in 3.10. We use variables for types that are reused. This should now be explicitly typed as type aliases. https://peps.python.org/pep-0613/

doc/python-type-hinting.md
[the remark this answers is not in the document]
ONS-Guilhem-Forey: Thanks @MebinAbraham , wasn't aware of this change, happy to ignore this comment @Hajara-I  👍

ONS-Guilhem-Forey (approved):

Yuyuutsu (approved):

petechd (approved):

doc/python-type-hinting.md
```diff
@@ -40,18 +40,16 @@ def get_objects_matching(ids: Sequence[str]) -> dict[int, dict]
```
MebinAbraham, line 40: There are references to "Optional" still

doc/python-type-hinting.md
MebinAbraham, line 52: The `[]` are not needed for union. Should just be `var: None | int`?

doc/python-type-hinting.md
```diff
@@ -64,24 +62,100 @@ def test(self, var: Union[None, int, str]) -> None:
...
+To mark portions of the program that should not be covered by type hinting, use the following on a particular line:
+
+```python
+# type: ignore
+```
+
+A `# type: ignore` comment on a line by itself at the top of a file silences all errors in the file.
+
+`# type: ignore` should only be used when unavoidable. Ensure that a comment is added to explain why it has been used and have a prefix of `Type ignore:`.
+
```
berroar, line 107: I would add an example here to make this clearer, maybe the comment and a line below with the #type: ignore added?

doc/python-type-hinting.md
```diff
@@ -64,24 +62,100 @@ def test(self, var: Union[None, int, str]) -> None:
...
+```python
+T = TypeVar('T', int, float)
+
+def increment_value(self, value: T) -> T:
+    return value + 1
 ```
 
-This import is not necessary in Python 3.10.
+This tells the typechecker that values other than `int` and `float` are not allowed, resulting in an error if a string is passed for example:
```
berroar, line 151: Think the examples are ok, maybe some of the sentences here don't add much over what is already in the docs and not really team specific?

ONS-Guilhem-Forey:
One last comment, is worth enforcing `Tuple` rather than `list` on the return type when the length is known?
Typically:

```python
from typing import Tuple

def foo() -> Tuple[str, list[int]]
   a = "hello"
   b = [0, 1, 2]

   return a, b

# IntelliSense will be more helpful at providing insight on the return type
v1, v2 = foo()
```


Is much better at documenting the code for the caller than:

```python
def foo() -> list[str | int]
   a = "hello"
   b = [0, 1, 2]

   return a, b

v1, v2 = foo()
```


As it documents the length of the fixed Sequence, as well as the type of its items, when known.

doc/python-type-hinting.md
Hajara-I: Reference to optional refactored with pipe operator

doc/python-type-hinting.md
Hajara-I: Refactored to remove `[]`

doc/python-type-hinting.md
Hajara-I: Example added to better illustrate the use of `Type ignore`

doc/python-type-hinting.md
Hajara-I: Sections refactored to be more team specific

rmccar (approved):

doc/python-type-hinting.md
```diff
@@ -40,18 +38,16 @@ def get_objects_matching(ids: Sequence[str]) -> dict[int, dict]
...
 ## Optional arguments
 
-When `Optional` is used for function arguments, it means that the argument can have the value `None`, not that the argument is optional. 
-
-For arguments that can be a single type or `None`, use `Optional`:
+For arguments that can be a single type or None, use the shorthand `|` instead of using the older `Optional` keyword:
 
 ```python
-def test(self, var: Optional[int]) -> None:
+def test(self, var: [None | int]) -> None:
```
MebinAbraham, line 44: The square brackets shouldn't exist.

doc/python-type-hinting.md
```diff
@@ -64,24 +60,98 @@ def test(self, var: Union[None, int, str]) -> None:
...
+```python
+# type: ignore
+```
+
+`# type: ignore` should only be used when unavoidable. Ensure that a comment is added to explain why it has been used and have a prefix of `Type ignore:`
+
+```python
+# Type ignore: Explain why type ignore has been used in this context
+x: int
+x: str  # type: ignore
```
MebinAbraham, line 107: We should add a more meaningful example. This to me isn't valid. Maybe one where it returns any then you explicitly type it.

MebinAbraham (approved):

berroar (approved):

PR #1031, opened 2023-02-10
BUG: Fixing schema validation failure on large errors

creativeappdevamit:
[Trello](https://trello.com/c/8GgAVQ00/5563-schema-validation-script-is-falling-over-when-the-validation-errors-are-large)


berroar (commented):
Attempted to test this with a large schema with lots of errors and can't see too much difference between this and master - unless I am doing something wrong 😅 Did you see a larger hang on master? It hangs for a few seconds on both for me

creativeappdevamit:
> Attempted to test this with a large schema with lots of errors and can't see too much difference between this and master - unless I am doing something wrong 😅 Did you see a larger hang on master? It hangs for a few seconds on both for me

Hi @berroar - Yes it does hang for me on master. Do you mind if I am able to show you this please

PR #1029, opened 2023-02-09
Update to DS version 62

rmccar:
### What is the context of this PR?
Update to DS version 62.

### How to review
Tests pass

### Checklist
* [ ] New static content marked up for translation
* [ ] Newly defined schema content included in eq-translations repo

PR #1037, opened 2023-02-15
Bug: Runner page width set to 12 columns

MebinAbraham:
DS V62 sets the default width to 12 columns, this PR sets it to 8 to as that is what runner requires. This was regressed in https://github.com/ONSdigital/eq-questionnaire-runner/pull/1029

Although this is a UI change which we don't normally test, I have added an assertion which checks the width using the CSS class. Although this couples the test with the DS classes, this change is too important to regress.

### How to review
Ensure the width for pages in runner look as expected.

### Checklist

### DIFF
--- a/templates/layouts/_base.html
+++ b/templates/layouts/_base.html
@@ -36,6 +36,7 @@
 
 {% set pageConfig = {
   "title": full_page_title,
+  "pageColNumber": 8,
   "footer": footer,
   "cdn": {
     "url": cdn_url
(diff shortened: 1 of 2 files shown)


Yuyuutsu (approved):
Launched local runner on this branch and compared with the master as well as the DS62. There is no more stretch compared to runner on the dsv62 branch.

ONS-Guilhem-Forey (approved):
Tested locally like Yuyu, LGTM

PR #1070, opened 2023-03-24
Confirmation page after signout

rmccar:
### What is the context of this PR?
Adds a sign out confirmation page for when a user is in the middle of a survey and sign out to confirm their answers have been saved.

Also:
- adds the health survey type
- fixes an accessibility issue in the error pages that were using a `<h3>` tag out of sequence
- Also fixes a typo in the error tests ("broked")
- Changes the `test_session` file to remove hardcoded urls in test

### How to review
Describe the steps required to test the changes (include screenshots if appropriate).

### Checklist
* [ ] New static content marked up for translation
* [ ] Newly defined schema content included in eq-translations repo

PR #1072, opened 2023-03-27
Session timeout RH link defect

rmccar:
Fixes session timeout page link to take the base url from the cookie if it exists.

*This PR has been branched off my other PR (#1070) because some changes needed to be made to that PR too for this*

### How to review
Open a schema with a `Account Service URL` set. Check that the session expired page uses this as a base url for the page link.

### Checklist


rmccar:
[Defect: Link to re-enter AUC following a session timeout in pre-prod and Int are pointing to Prod RH-UI](https://trello.com/c/Mtpxe4HM/5638-defect-link-to-re-enter-auc-following-a-session-timeout-in-pre-prod-and-int-are-pointing-to-prod-rh-ui)

kylelawsonAND:
Link appears to redirect correctly locally (no page to load of course but url is correct I think). Can't test in a different env but otherwise all looks sensible, unofficial ✅  from me

petechd (changes requested):
You need to change the base branch for this PR to main.

rmccar:
> You need to change the base branch for this PR to main.

Yeah I know I will change it when my other PR goes in. Because I branched off my other PR this allows me to see the changes without the ones from that branch at the moment

app/routes/errors.py
```diff
@@ -52,7 +53,9 @@ def _render_error_page(status_code, template=None, **kwargs):
     handle_language()
     business_survey_config = get_survey_config(theme=SurveyType.BUSINESS)
     other_survey_config = get_survey_config(
-        theme=SurveyType.SOCIAL, base_url=ACCOUNT_SERVICE_BASE_URL_SOCIAL
+        theme=SurveyType.SOCIAL,
+        base_url=cookie_session.get("account_service_base_url")
+        or ACCOUNT_SERVICE_BASE_URL_SOCIAL,
     )
```
MebinAbraham, line 59: Actually this won't work as the base url will be used by both types in this case. We actually don't need any changes in runner. You need to update the runner terraform and eq pipelines to add in `ACCOUNT_SERVICE_BASE_URL_SOCIAL`, at the moment, it doesn't set it, therefore it defaults to prod.

PR #1076, opened 2023-03-30
Fix language changing

rmccar:
Adds PHM and CRIS to language toggle mapping

### How to review
- Launch PHM and CRIS schemas in Cymraeg and see that they work as expected and the correct ONS logo is shown and you can change between the two languages.

### Checklist

### DIFF
--- a/app/utilities/schema.py
+++ b/app/utilities/schema.py
@@ -93,8 +93,10 @@ def _schema_exists(language_code, schema_name):
 
 
 def get_allowed_languages(launch_language):
-    for language in LANGUAGES:
-        return LANGUAGES if launch_language in language else [DEFAULT_LANGUAGE_CODE]
+    for language_combination in LANGUAGES:
+        if launch_language in language_combination:
+            return language_combination
+    return [DEFAULT_LANGUAGE_CODE]
 
 
 def load_schema_from_metadata(
--- a/app/utilities/schema.py
+++ b/app/utilities/schema.py
@@ -22,7 +22,7 @@ logger = get_logger()
 SCHEMA_DIR = "schemas"
 LANGUAGE_CODES = ("en", "cy")
 
-LANGUAGES_MAP = {"test_language": [["en", "cy"]]}
+LANGUAGES_MAP = [["en", "cy"]]
 
 SCHEMA_REQUEST_MAX_BACKOFF = 0.2
 SCHEMA_REQUEST_MAX_RETRIES = 2  # Totals no. of request should be 3. The initial request + SCHEMA_REQUEST_MAX_RETRIES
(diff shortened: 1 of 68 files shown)


app/utilities/schema.py
```diff
@@ -92,8 +92,8 @@ def _schema_exists(language_code, schema_name):
     )
 
 
-def get_allowed_languages(schema_name, launch_language):
```
petechd, line 95: No need for nested list and then just:
```python
def get_allowed_languages(launch_language):
    if launch_language in LANGUAGES_MAP:
        return language_combination
    return [DEFAULT_LANGUAGE_CODE]
```

app/utilities/schema.py
```diff
@@ -22,7 +22,7 @@
 SCHEMA_DIR = "schemas"
 LANGUAGE_CODES = ("en", "cy")
 
-LANGUAGES_MAP = {"test_language": [["en", "cy"]]}
+LANGUAGES_MAP = [["en", "cy"]]
```
berroar, line 25: This is no longer a language map, has the requirement to have a map by schema name been removed?

During census the map looked something like this:
LANGUAGES_MAP = {
    "test_language": [["en", "cy"], ["en", "ga"]],
    "ccs_household_gb_wls": [["en", "cy"]],
    "census_household_gb_wls": [["en", "cy"]],
    "census_individual_gb_wls": [["en", "cy"]],
    "census_household_gb_nir": [["en"], ["en", "ga"], ["en", "eo"]],
    "census_individual_gb_nir": [["en"], ["en", "ga"], ["en", "eo"]],
    "census_communal_establishment_gb_wls": [["en", "cy"]],
}
Do we not want to do something similar for CRIS/PHM?

rmccar: The requirement hasn't changed but because at the moment we only have the `test_language` line in this. So this means it only sets the correct language for the page in the `test_language` schema. For everything else, every page is "en" even if you launch setting the language to "cy". So this isn't just broken for CRIS and PHM. Considering we only seem to have "en" or "cy" languages at the moment is there the need for a map like this? We could get rid of this and simplify it.

app/utilities/schema.py
rmccar: Tried this but needed to revert it because removing the for loop broke it because its being called with multiple languages

app/utilities/schema.py
rmccar: Ive had a word with Mebin and realise that we have some tech debt to sort this out so ill just add the broken schemas for now

berroar (approved):

app/utilities/schema.py
petechd: Yeah that was meant for flattened list, then it would work, but since you removed this change now it doesn't matter anymore.

petechd (approved):

PR #1079, opened 2023-04-03
Add dynamic answers list source

petechd:

### How to review

### Checklist

### DIFF
--- a/app/questionnaire/placeholder_renderer.py
+++ b/app/questionnaire/placeholder_renderer.py
@@ -117,7 +117,11 @@ class PlaceholderRenderer:
         resolved_dynamic_answers = []
         list_items = []
         if dynamic_answers := data_to_render_mutable.get("dynamic_answers", {}):
-            list_items = self._list_store.get(dynamic_answers["values"].get("identifier")).serialize().get("items")
+            list_items = (
+                self._list_store.get(dynamic_answers["values"].get("identifier"))
+                .serialize()
+                .get("items")
+            )
             for dynamic_answer in dynamic_answers["answers"]:
                 for item in list_items:
                     resolved_dynamic_answer = self._schema.get_mutable_deepcopy(
@@ -130,12 +134,16 @@ class PlaceholderRenderer:
                     # self._schema._parent_id_map[
                     #     dynamic_answer['id']
                     # ] = data_to_render_mutable.get("id")
-                    self._schema._parent_id_map[dynamic_answer['id']] = data_to_render_mutable.get("id")
+                    self._schema._parent_id_map[
+                        dynamic_answer["id"]
+                    ] = data_to_render_mutable.get("id")
 
                     resolved_dynamic_answers.append(resolved_dynamic_answer)
 
             data_to_render_mutable["answers"] += resolved_dynamic_answers
-            data_to_render_mutable["dynamic_answers_list_id"] = data_to_render_mutable["dynamic_answers"]["values"]["identifier"]
+            data_to_render_mutable["dynamic_answers_list_id"] = data_to_render_mutable[
+                "dynamic_answers"
+            ]["values"]["identifier"]
             del data_to_render_mutable["dynamic_answers"]
 
         pointers = find_pointers_containing(data_to_render_mutable, "placeholders")
--- a/app/questionnaire/placeholder_renderer.py
+++ b/app/questionnaire/placeholder_renderer.py
@@ -114,12 +114,40 @@ class PlaceholderRenderer:
         data_to_render_mutable: dict[
             str, Any
         ] = QuestionnaireSchema.get_mutable_deepcopy(data_to_render)
+        resolved_dynamic_answers = []
+        list_items = []
+        if dynamic_answers := data_to_render_mutable.get("dynamic_answers", {}):
+            list_items = self._list_store.get(dynamic_answers["values"].get("identifier")).serialize().get("items")
+            for dynamic_answer in dynamic_answers["answers"]:
+                for item in list_items:
+                    resolved_dynamic_answer = self._schema.get_mutable_deepcopy(
+                        dynamic_answer
+                    )
+                    resolved_id = f"{dynamic_answer['id']}-{item}"
+                    resolved_dynamic_answer["id"] = resolved_id
+
+                    # Updating parent id map for dependencies purposes (used later by questionnaire store updater on post)
+                    # self._schema._parent_id_map[
+                    #     dynamic_answer['id']
+                    # ] = data_to_render_mutable.get("id")
+                    self._schema._parent_id_map[dynamic_answer['id']] = data_to_render_mutable.get("id")
+
+                    resolved_dynamic_answers.append(resolved_dynamic_answer)
+
+            data_to_render_mutable["answers"] += resolved_dynamic_answers
+            data_to_render_mutable["dynamic_answers_list_id"] = data_to_render_mutable["dynamic_answers"]["values"]["identifier"]
+            del data_to_render_mutable["dynamic_answers"]
+
         pointers = find_pointers_containing(data_to_render_mutable, "placeholders")
 
-        for pointer in pointers:
-            rendered_text = self.render_pointer(
-                data_to_render_mutable, pointer, list_item_id
-            )
+        for index, pointer in enumerate(pointers):
+            if list_items:
+                rendered_text = self.render_pointer(
+                    data_to_render_mutable, pointer, list_items[index]
+                )
+            else:
+                rendered_text = self.render_pointer(
+                    data_to_render_mutable, pointer, list_item_id
+                )
             set_pointer(data_to_render_mutable, pointer, rendered_text)
-
         return data_to_render_mutable
--- a/app/questionnaire/placeholder_renderer.py
+++ b/app/questionnaire/placeholder_renderer.py
@@ -115,10 +115,10 @@ class PlaceholderRenderer:
             str, Any
         ] = QuestionnaireSchema.get_mutable_deepcopy(data_to_render)
         resolved_dynamic_answers = []
-        list_items = []
+        list_items: list = []
         if dynamic_answers := data_to_render_mutable.get("dynamic_answers", {}):
             list_items = (
-                self._list_store.get(dynamic_answers["values"].get("identifier"))
+                self._list_store.get(dynamic_answers["values"].get("identifier"))  # type: ignore
                 .serialize()
                 .get("items")
             )
@@ -130,13 +130,11 @@ class PlaceholderRenderer:
                     resolved_id = f"{dynamic_answer['id']}-{item}"
                     resolved_dynamic_answer["id"] = resolved_id
 
-                    # Updating parent id map for dependencies purposes (used later by questionnaire store updater on post)
-                    # self._schema._parent_id_map[
-                    #     dynamic_answer['id']
-                    # ] = data_to_render_mutable.get("id")
-                    self._schema._parent_id_map[
+                    self._schema._parent_id_map[  # pylint: disable=protected-access
                         dynamic_answer["id"]
-                    ] = data_to_render_mutable.get("id")
+                    ] = data_to_render_mutable.get(
+                        "id"
+                    )  # type: ignore
 
                     resolved_dynamic_answers.append(resolved_dynamic_answer)
 
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -342,16 +342,25 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
                     }
         if value_source["source"] == "list":
             section = self.get_section_for_block_id(block_id)
-            list_collector = self.get_list_collector_for_list(section=section, for_list=value_source["identifier"])
-            add_block_question = self.get_add_block_for_list_collector(list_collector["id"])["question"]
-            answer_id_for_block = list(self.get_answers_for_question_by_id(add_block_question).keys())[0]
+            list_collector = self.get_list_collector_for_list(
+                section=section, for_list=value_source["identifier"]
+            )
+            add_block_question = self.get_add_block_for_list_collector(
+                list_collector["id"]
+            )["question"]
+            answer_id_for_block = list(
+                self.get_answers_for_question_by_id(add_block_question).keys()
+            )[0]
             self._answer_dependencies_map[answer_id_for_block] |= {
                 self._get_answer_dependent_for_block_id(block_id=block_id, for_list=value_source["identifier"])  # type: ignore
             }
 
-
     def _get_answer_dependent_for_block_id(
-        self, *, block_id: str, answer_id: Optional[str] = None, for_list: Optional[str] = None
+        self,
+        *,
+        block_id: str,
+        answer_id: Optional[str] = None,
+        for_list: Optional[str] = None,
     ) -> AnswerDependent:
         section_id: str = self.get_section_id_for_block_id(block_id)  # type: ignore
         if not for_list:
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -340,12 +340,22 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
                     self._answer_dependencies_map[answer_id_for_block] |= {
                         self._get_answer_dependent_for_block_id(block_id=block_id, answer_id=answer_id)  # type: ignore
                     }
+        if value_source["source"] == "list":
+            section = self.get_section_for_block_id(block_id)
+            list_collector = self.get_list_collector_for_list(section=section, for_list=value_source["identifier"])
+            add_block_question = self.get_add_block_for_list_collector(list_collector["id"])["question"]
+            answer_id_for_block = list(self.get_answers_for_question_by_id(add_block_question).keys())[0]
+            self._answer_dependencies_map[answer_id_for_block] |= {
+                self._get_answer_dependent_for_block_id(block_id=block_id, for_list=value_source["identifier"])  # type: ignore
+            }
+
 
     def _get_answer_dependent_for_block_id(
-        self, *, block_id: str, answer_id: Optional[str] = None
+        self, *, block_id: str, answer_id: Optional[str] = None, for_list: Optional[str] = None
     ) -> AnswerDependent:
         section_id: str = self.get_section_id_for_block_id(block_id)  # type: ignore
-        for_list = self.get_repeating_list_for_section(section_id)
+        if not for_list:
+            for_list = self.get_repeating_list_for_section(section_id)
 
         return AnswerDependent(
             block_id=block_id,
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -343,10 +343,10 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
         if value_source["source"] == "list":
             section = self.get_section_for_block_id(block_id)
             list_collector = self.get_list_collector_for_list(
-                section=section, for_list=value_source["identifier"]
+                section=section, for_list=value_source["identifier"]  # type: ignore
             )
-            add_block_question = self.get_add_block_for_list_collector(
-                list_collector["id"]
+            add_block_question = self.get_add_block_for_list_collector(  # type: ignore
+                list_collector["id"]  # type: ignore
             )["question"]
             answer_id_for_block = list(
                 self.get_answers_for_question_by_id(add_block_question).keys()
--- a/app/questionnaire/questionnaire_schema.py
+++ b/app/questionnaire/questionnaire_schema.py
@@ -251,6 +251,13 @@ class QuestionnaireSchema:  # pylint: disable=too-many-public-methods
                     )
                     continue
 
+                if dynamic_answers := question.get("dynamic_answers"):
+                    for answer in dynamic_answers.get("answers"):
+                        value_source = dynamic_answers["values"]
+                        self._update_answer_dependencies_for_value_source(
+                            value_source, block_id=block["id"], answer_id=answer["id"]
+                        )
+
                 for answer in question.get("answers", []):
                     self._update_answer_dependencies_for_answer(
                         answer, block_id=block["id"]
--- a/app/questionnaire/questionnaire_store_updater.py
+++ b/app/questionnaire/questionnaire_store_updater.py
@@ -281,12 +281,10 @@ class QuestionnaireStoreUpdater:
                     self._answer_store.remove_answer(
                         dependency.answer_id, list_item_id=list_item_id
                     )
-
-                if question := self._schema.get_block(dependency.block_id).get("question"):
-                    if question.get("dynamic_answers"):
-                        self.dependent_block_id_by_section_key[
-                            (dependency.section_id, None)
-                        ].add(dependency.block_id)
+                if self.dependency_has_dynamic_answers(dependency.block_id):
+                    self.dependent_block_id_by_section_key[
+                        (dependency.section_id, None)
+                    ].add(dependency.block_id)
                 else:
                     self.dependent_block_id_by_section_key[
                         (dependency.section_id, list_item_id)
@@ -413,3 +411,9 @@ class QuestionnaireStoreUpdater:
 
     def started_section_keys(self, section_ids: Optional[Iterable[str]] = None):
         return self._progress_store.started_section_keys(section_ids)
+
+    def dependency_has_dynamic_answers(self, block_id):
+        if question := self._schema.get_block(block_id).get("question"):
+            if question.get("dynamic_answers"):
+                return True
+        return False
--- a/app/questionnaire/questionnaire_store_updater.py
+++ b/app/questionnaire/questionnaire_store_updater.py
@@ -333,7 +333,10 @@ class QuestionnaireStoreUpdater:
             self._current_question
         )
         for item in answer_ids_for_question:
-            if item.split("-")[-1] in self._list_store._list_item_ids():
+            if (
+                item.split("-")[-1]
+                in self._list_store._list_item_ids()  # pylint: disable=protected-access
+            ):
                 answer_ids_for_question.append(list(form_data.keys())[0])
                 answer_ids_for_question.remove(item)
 
--- a/app/questionnaire/questionnaire_store_updater.py
+++ b/app/questionnaire/questionnaire_store_updater.py
@@ -328,6 +328,10 @@ class QuestionnaireStoreUpdater:
         answer_ids_for_question = self._schema.get_answer_ids_for_question(
             self._current_question
         )
+        for item in answer_ids_for_question:
+            if item.split("-")[-1] in self._list_store._list_item_ids():
+                answer_ids_for_question.append(list(form_data.keys())[0])
+                answer_ids_for_question.remove(item)
 
         for answer_id, answer_value in form_data.items():
             if answer_id not in answer_ids_for_question:
--- a/app/questionnaire/questionnaire_store_updater.py
+++ b/app/questionnaire/questionnaire_store_updater.py
@@ -282,9 +282,15 @@ class QuestionnaireStoreUpdater:
                         dependency.answer_id, list_item_id=list_item_id
                     )
 
-                self.dependent_block_id_by_section_key[
-                    (dependency.section_id, list_item_id)
-                ].add(dependency.block_id)
+                if question := self._schema.get_block(dependency.block_id).get("question"):
+                    if question.get("dynamic_answers"):
+                        self.dependent_block_id_by_section_key[
+                            (dependency.section_id, None)
+                        ].add(dependency.block_id)
+                else:
+                    self.dependent_block_id_by_section_key[
+                        (dependency.section_id, list_item_id)
+                    ].add(dependency.block_id)
 
     def _get_list_item_ids_for_dependency(
         self, dependency: AnswerDependent, is_repeating_answer: Optional[bool] = False
--- a/app/submitter/convert_payload_0_0_3.py
+++ b/app/submitter/convert_payload_0_0_3.py
@@ -152,12 +152,16 @@ def add_relationships_unrelated_answers(
         ):
             answers_payload.add_or_update(unrelated_answer)
 
+
 def resolve_dynamic_answers(
-    block: dict, answer_store: AnswerStore, answers_payload: AnswerStore, list_store: ListStore
+    block: dict,
+    answer_store: AnswerStore,
+    answers_payload: AnswerStore,
+    list_store: ListStore,
 ) -> None:
     for answer in block["dynamic_answers"].get("answers"):
-        for list_item_id in list_store.get(block["dynamic_answers"]["values"].get("identifier")):
-            answer = answer_store.get_answer(
-                answer["id"], list_item_id=list_item_id
-            )
+        for list_item_id in list_store.get(
+            block["dynamic_answers"]["values"].get("identifier")
+        ):
+            answer = answer_store.get_answer(answer["id"], list_item_id=list_item_id)
             answers_payload.add_or_update(answer)
--- a/app/submitter/convert_payload_0_0_3.py
+++ b/app/submitter/convert_payload_0_0_3.py
@@ -162,6 +162,8 @@ def resolve_dynamic_answers(
     for answer in block["dynamic_answers"].get("answers"):
         for list_item_id in list_store.get(
             block["dynamic_answers"]["values"].get("identifier")
-        ):
-            answer = answer_store.get_answer(answer["id"], list_item_id=list_item_id)
-            answers_payload.add_or_update(answer)
+        ).items:
+            extracted_answer = answer_store.get_answer(
+                answer["id"], list_item_id=list_item_id
+            )
+            answers_payload.add_or_update(extracted_answer)
--- a/app/submitter/convert_payload_0_0_3.py
+++ b/app/submitter/convert_payload_0_0_3.py
@@ -166,4 +166,4 @@ def resolve_dynamic_answers(
             extracted_answer = answer_store.get_answer(
                 answer["id"], list_item_id=list_item_id
             )
-            answers_payload.add_or_update(extracted_answer)
+            answers_payload.add_or_update(extracted_answer)  # type: ignore
--- a/app/submitter/convert_payload_0_0_3.py
+++ b/app/submitter/convert_payload_0_0_3.py
@@ -68,6 +68,12 @@ def convert_answers_to_payload_0_0_3(
                         list_collector_block=block,
                         answers_payload=answers_payload,
                     )
+                if (question := block.get("question")) and (
+                    question.get("dynamic_answers")
+                ):
+                    resolve_dynamic_answers(
+                        question, answer_store, answers_payload, list_store
+                    )
 
                 answer_ids = schema.get_answer_ids_for_block(block_id)
                 answers_in_block = answer_store.get_answers_by_answer_id(
@@ -145,3 +151,13 @@ def add_relationships_unrelated_answers(
             )
         ):
             answers_payload.add_or_update(unrelated_answer)
+
+def resolve_dynamic_answers(
+    block: dict, answer_store: AnswerStore, answers_payload: AnswerStore, list_store: ListStore
+) -> None:
+    for answer in block["dynamic_answers"].get("answers"):
+        for list_item_id in list_store.get(block["dynamic_answers"]["values"].get("identifier")):
+            answer = answer_store.get_answer(
+                answer["id"], list_item_id=list_item_id
+            )
+            answers_payload.add_or_update(answer)
--- a/app/views/contexts/summary/question.py
+++ b/app/views/contexts/summary/question.py
@@ -5,7 +5,6 @@ from markupsafe import Markup, escape
 
 from app.data_models import AnswerStore, ListStore
 from app.data_models.answer import AnswerValueEscapedTypes, escape_answer_value
-from app.data_models.metadata_proxy import MetadataProxy
 from app.forms.field_handlers.select_handlers import DynamicAnswerOptions
 from app.questionnaire import Location, QuestionnaireSchema, QuestionSchemaType
 from app.questionnaire.placeholder_renderer import PlaceholderRenderer
@@ -73,6 +72,7 @@ class Question:
 
         return escape_answer_value(answer.value) if answer else None
 
+    # pylint: disable=too-many-locals
     def _build_answers(
         self,
         *,
@@ -117,15 +117,18 @@ class Question:
                         list_store=self.list_store,
                         schema=self.schema,
                         language="en",
-                        metadata={},
+                        metadata=None,
                         response_metadata={},
                     )
                     self.list_item_id = list_item_id
-                    answer_value: Optional[AnswerValueEscapedTypes] = self.get_answer(
-                        answer_store, answer_schema["id"]
-                    )
+                    dynamic_answer_value: Optional[
+                        AnswerValueEscapedTypes
+                    ] = self.get_answer(answer_store, answer_schema["id"])
                     answer = self._build_answer(
-                        answer_store, question_schema, answer_schema, answer_value
+                        answer_store,
+                        question_schema,
+                        answer_schema,
+                        dynamic_answer_value,
                     )
 
                     resolved_answer_schema = self.schema.get_mutable_deepcopy(
--- a/app/views/contexts/summary/question.py
+++ b/app/views/contexts/summary/question.py
@@ -3,10 +3,12 @@ from typing import Any, Mapping, Optional
 from flask import url_for
 from markupsafe import Markup, escape
 
-from app.data_models import AnswerStore
+from app.data_models import AnswerStore, ListStore
 from app.data_models.answer import AnswerValueEscapedTypes, escape_answer_value
+from app.data_models.metadata_proxy import MetadataProxy
 from app.forms.field_handlers.select_handlers import DynamicAnswerOptions
 from app.questionnaire import Location, QuestionnaireSchema, QuestionSchemaType
+from app.questionnaire.placeholder_renderer import PlaceholderRenderer
 from app.questionnaire.rules.rule_evaluator import RuleEvaluator
 from app.questionnaire.value_source_resolver import ValueSourceResolver
 from app.views.contexts.summary.answer import (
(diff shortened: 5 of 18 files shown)


petechd:
[Iteration 3 - Dynamically creating answers based on a list collector (8 SP)](https://trello.com/c/eO99WHvl/5323-iteration-3-dynamically-creating-answers-based-on-a-list-collector-8-sp)

app/views/handlers/question.py
```diff
@@ -194,7 +226,27 @@ def get_list_summary_context(self):
     def handle_post(self):
         # pylint: disable=no-member
         # wtforms Form parents are not discoverable in the 2.3.3 implementation
-        self.questionnaire_store_updater.update_answers(self.form.data)
+        dynamic_answers = False
+        for item in self.form.data:
+            if list_item_id := item.split("-")[-1]:
```
berroar, line 231: Is there a cleaner way of identifying a dynamic answer? We should already know the answer id/prefix for all dynamic answers - rather than checking the part after the `-` for every item and checking if it's in the list store? We would have the dynamic_answers key in the block at this point

app/questionnaire/placeholder_renderer.py
```diff
@@ -125,6 +125,12 @@ def render(
         data_to_render_mutable: dict[
             str, Any
         ] = QuestionnaireSchema.get_mutable_deepcopy(data_to_render)
+
+        list_items: list = []
```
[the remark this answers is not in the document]
petechd: I moved it to that method and tried to make it more readable/improve the flow, method returns a tuple now. List items are still needed for placeholder renderer so I initialise it empty slightly later.

app/questionnaire/placeholder_renderer.py
```diff
@@ -141,13 +147,67 @@ def render(
...
+        )
+        for dynamic_answer in dynamic_answers["answers"]:
+            for item in list_items:  # type: ignore
+                resolved_dynamic_answer = self._schema.get_mutable_deepcopy(
+                    dynamic_answer
+                )
+                resolved_id = f"{dynamic_answer['id']}-{item}"
+                resolved_dynamic_answer["id"] = resolved_id
+
+                self._schema._parent_id_map[  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
petechd: `parent_id_map` is immutable dict unlike `_parent_id_map`.

app/views/contexts/summary/question.py
```diff
@@ -95,24 +108,91 @@ def _build_answers(
...
-                answer_schema=answer_schema,
-                answer_value=answer,
-                block_id=block_id,
-                list_name=list_name,
-                list_item_id=self.list_item_id,
-                return_to=return_to,
-                return_to_block_id=return_to_block_id,
-            ).serialize()
-            summary_answers.append(summary_answer)
+        if self.dynamic_answer_schemas:
```
MebinAbraham, line 112: This should be simplified, instead of doing:
```py
if dynamic:
	process_normal()
	process_dynamic()
else:
	process_normal()
```

it should just be:

```py
if dynamic and list source:
	process_dynamic()

process_normal_if_any()
```

app/views/contexts/summary/question.py
```diff
@@ -95,24 +108,91 @@ def _build_answers(
...
+
+                    summary_answer = Answer(
+                        answer_schema=rendered_schema,
+                        answer_value=answer,
+                        block_id=block_id,
+                        list_name=list_name,
+                        list_item_id=self.list_item_id,
+                        return_to=return_to,
+                        return_to_block_id=return_to_block_id,
+                    ).serialize()
```
MebinAbraham, line 176: The change links do not work correct for dynamic answers. They need to be updated and tested, i.e when you click change, the right item is focused.

app/views/handlers/list_remove_question.py
```diff
@@ -21,6 +21,14 @@ def is_location_valid(self):
         return True
 
     def handle_post(self):
+        answer_block = self._schema.get_add_block_for_list_collector(  # pylint: disable=protected-access
+            self.parent_block["id"]
+        )
+        answer_id = self._schema.get_first_answer_id_for_block(answer_block["id"])
```
MebinAbraham, line 27: Why the first answer?.

app/views/handlers/list_remove_question.py
```diff
@@ -21,6 +21,14 @@ def is_location_valid(self):
...
+        self.questionnaire_store_updater._capture_block_dependencies_for_answer(  # pylint: disable=protected-access
+            answer_id
+        )
```
MebinAbraham, line 30: This shouldn't be called outside questionnaire store updated. It's private for a reason, we need to be careful before just disabling these alerts, it signifies something is not right. Make this call inside `remove_list_item_and_answers` if any answer were removed, similar to `update_answers.

app/views/handlers/list_remove_question.py
```diff
@@ -32,6 +40,8 @@ def handle_post(self):
                 self.parent_block["for_list"]
             )
 
+        self.questionnaire_store_updater.update_progress_for_dependent_sections()
```
MebinAbraham, line 43: Is this needed on top of the existing `evaluate_and_update_section_status_on_list_change`?. If so why?. In add and remove the order of them is also different. Are we not reevaluating the same sections in the two methods, or at least a sub set of the are duplicates?.

app/views/handlers/question.py
```diff
@@ -113,8 +113,40 @@ def get_next_location_url(self):
             self._return_to_block_id,
         )
 
+    # pylint: disable=protected-access
     def _get_answers_for_question(self, question_json) -> dict[str, Any]:
         answer_ids = self._schema.get_answer_ids_for_question(question_json)
+        if question_json.get("dynamic_answers_list_id"):
```
MebinAbraham, line 119: This should check the dynamic answers key, also the change below shouldn't repeat for normal answers, it should be:

```py
if dynamic:
	handle_dynamic()

handle_static_if_any()
```

schemas/test/en/test_dynamic_answers_list.json
```diff
@@ -0,0 +1,306 @@
+{
```
MebinAbraham, line 1: You will need 1 with static answers, 1 without. Currently, answers are expected in certain parts of runner but they haven't been updated to handle that it might no longer exist. So testing both will highlight those issues. Tests should be updated accordingly.

MebinAbraham (commented):
Initial pass, functionally mostly works fine. However, we need to think about where the logic should be done and need to keep it extensible to new sources, we shouldn't assume the source is list.

app/questionnaire/placeholder_renderer.py
```diff
@@ -141,13 +149,70 @@ def render(
...
```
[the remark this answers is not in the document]
petechd: This has to do with summary display of questions, necessary for it.

app/questionnaire/placeholder_renderer.py
```diff
@@ -141,13 +149,70 @@ def render(
...
+                    dynamic_answer["id"]
+                ] = data_to_render_mutable.get(
+                    "id"
+                )  # type: ignore
+
+                resolved_dynamic_answers.append(resolved_dynamic_answer)
+
+        data_to_render_mutable["answers"] += resolved_dynamic_answers
+        data_to_render_mutable["dynamic_answers_list_id"] = data_to_render_mutable[
```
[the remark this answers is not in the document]
petechd: Ok, structure stays more or less the same now as a result of changes in placeholder renderer, see https://github.com/ONSdigital/eq-questionnaire-runner/pull/1079#discussion_r1163391717

app/questionnaire/placeholder_renderer.py
```diff
@@ -141,13 +149,70 @@ def render(
...
+            "dynamic_answers"
+        ]["values"]["identifier"]
+        del data_to_render_mutable["dynamic_answers"]
+
+        return data_to_render_mutable, list_items
```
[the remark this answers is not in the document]
petechd: Only doing data to render now as a result of revamp of placeholder renderer requested in https://github.com/ONSdigital/eq-questionnaire-runner/pull/1079#discussion_r1163391717

app/views/contexts/summary/question.py
petechd: From what I can see in schema the Change button click focuses on the right answer?

app/questionnaire/questionnaire_schema.py
```diff
@@ -336,12 +343,32 @@ def _update_answer_dependencies_for_value_source(
...
+            list_collector = self.get_list_collector_for_list(
+                section=section, for_list=value_source["identifier"]  # type: ignore
+            )
+            # Returns optional, but it always returns at this point, same with optional returns below
+            add_block_question = self.get_add_block_for_list_collector(  # type: ignore
+                list_collector["id"]  # type: ignore
+            )["question"]
+            answer_id_for_block = list(
+                self.get_answers_for_question_by_id(add_block_question).keys()
+            )[0]
```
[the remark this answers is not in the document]
petechd: We fetch first here because that method (`get_answers_for_question_by_id`) returns a list even when there is only one element on it. Answer dependencies for list source were not populated before hence this change. I had to add the whole list source logic from scratch.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -281,10 +281,14 @@ def _capture_block_dependencies_for_answer(self, answer_id: str) -> None:
...
-                    (dependency.section_id, list_item_id)
-                ].add(dependency.block_id)
+                if self.dependency_has_dynamic_answers(dependency.block_id):
+                    self.dependent_block_id_by_section_key[
+                        (dependency.section_id, None)
+                    ].add(dependency.block_id)
+                else:
+                    self.dependent_block_id_by_section_key[
+                        (dependency.section_id, list_item_id)
+                    ].add(dependency.block_id)
```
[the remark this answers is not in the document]
petechd: This resets dynamic answers when list collector changes.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -328,6 +332,13 @@ def update_answers(
         answer_ids_for_question = self._schema.get_answer_ids_for_question(
             self._current_question
         )
+        for item in answer_ids_for_question:
+            if (
+                item.split("-")[-1]
+                in self._list_store._list_item_ids()  # pylint: disable=protected-access
+            ):
+                answer_ids_for_question.append(list(form_data.keys())[0])
```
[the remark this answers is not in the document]
petechd: Are dynamic shouldn't be first? https://github.com/ONSdigital/eq-questionnaire-runner/pull/1079#discussion_r1163404235

app/views/handlers/question.py
```diff
@@ -194,7 +226,27 @@ def get_list_summary_context(self):
...
+                    )
+                    dynamic_answers = True
+                elif item in self._schema.get_answer_ids_for_question(
+                    self.rendered_block["question"]
+                ):
+                    form_data = {item: self.form.data[item]}
+                    self.questionnaire_store_updater.update_answers(form_data)
+
+        if not dynamic_answers:
+            self.questionnaire_store_updater.update_answers(self.form.data)
```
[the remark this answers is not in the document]
petechd: Regarding your last sentence: how do we know answer in form is dynamic? Do we look it up in dynamic answers key in in that resolved block? Check if we get a match?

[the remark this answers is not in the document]
petechd: I don't do this in `update_answers` cause it updates for each list item id and here in this `form.data` we (can) have several.

schemas/test/en/test_dynamic_answers_list.json
petechd: I've added one without static to test schema, it seems to be working fine when running functional tests against it.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -328,6 +332,13 @@ def update_answers(
...
+                answer_ids_for_question.remove(item)
```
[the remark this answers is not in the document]
petechd: But in here we are only passing in one list_item_id so how would that work for two dynamic answers that require two list item ids?

app/views/handlers/list_remove_question.py
petechd: There is no method in schema class to extract the only answer id in a question block so I found this one useful.

app/views/handlers/list_remove_question.py
petechd: I am not sure we can move to `remove_list_item_and_answers` of questionnaire store updater as we need ListAction class context to invoke all the methods, we need answer id, parent block/list collector block to extract add block id.

app/views/handlers/list_remove_question.py
petechd: Moved it to `remove_list_item_and_answers` but still using this method.

app/views/handlers/list_remove_question.py
petechd: Disregard the comment above, moved it to `remove_list_item_and_answers` now.

schemas/test/en/test_dynamic_answers_list.json
berroar, line 1: I would call this `test_dynamic_answers_list` source. When functionally testing this, is it expected behaviour to see  the list collector driving question on the summary? I can see there is a test for it but It's not on the prototype  (but there is no driving question there anyway 😕)

app/questionnaire/placeholder_renderer.py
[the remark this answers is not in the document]
petechd: `parent_id_map` is not picking up id from dynamic answers object answer list when building `parent_id_map` while running the method initially for a schema and needs to be added manually here as we resolve dynamic answers and add them to `answers` list.

app/questionnaire/placeholder_renderer.py
```diff
@@ -149,5 +166,56 @@ def render(
...
             )
             set_pointer(data_to_render_mutable, pointer, rendered_text)
-
         return data_to_render_mutable
+
+    def resolve_dynamic_answers(
+        self, data_to_render_mutable: dict, list_items: list[str]
+    ) -> None:
+        dynamic_answers = data_to_render_mutable["dynamic_answers"]
+        if dynamic_answers["values"]["source"] == "list":
```
kylelawsonAND, line 175: Do we need to handle placeholder resolution for dynamic answers sourced from "answers" instead of "list"? I can see they're piped in as answer options for dynamic answer questions but could there be a scenario in which these "answers" sourced dynamic answer options are used for question text and thus need placeholder resolution like is being handled here?

app/questionnaire/questionnaire_schema.py
```diff
@@ -336,12 +343,32 @@ def _update_answer_dependencies_for_value_source(
                     self._answer_dependencies_map[answer_id_for_block] |= {
                         self._get_answer_dependent_for_block_id(block_id=block_id, answer_id=answer_id)  # type: ignore
```
kylelawsonAND, line 344: Add #type: ignore reason above usage (also on unchanged line 333)

app/questionnaire/questionnaire_schema.py
```diff
@@ -336,12 +343,32 @@ def _update_answer_dependencies_for_value_source(
...
                     }
+        if value_source["source"] == "list":
+            section = self.get_section_for_block_id(block_id)
+            list_collector = self.get_list_collector_for_list(
+                section=section, for_list=value_source["identifier"]  # type: ignore
```
kylelawsonAND, line 349: as above

app/questionnaire/questionnaire_schema.py
```diff
@@ -336,12 +343,32 @@ def _update_answer_dependencies_for_value_source(
...
+            self._answer_dependencies_map[answer_id_for_block] |= {
+                self._get_answer_dependent_for_block_id(block_id=block_id, for_list=value_source["identifier"])  # type: ignore
+            }
 
     def _get_answer_dependent_for_block_id(
-        self, *, block_id: str, answer_id: Optional[str] = None
+        self,
+        *,
+        block_id: str,
+        answer_id: Optional[str] = None,
```
kylelawsonAND, line 366: Make use of new pipe syntax `str | None`

app/questionnaire/questionnaire_schema.py
kylelawsonAND: Could also add type hinting and pipe syntax to this to method's declaration while here

app/questionnaire/placeholder_renderer.py
kylelawsonAND: + function driven dynamic answers but I think they count as "answers" sourced dynamic answer options

petechd: Kyle not at the moment, we abandoned the idea of resolving based on "answers" (for now). There is a comment on the card: "A decision was made to only support dynamic answers based on a list collector". Same with function driven.

kylelawsonAND: Ah yep I see that comment now, no worries at all my bad

app/questionnaire/questionnaire_schema.py
petechd: That's not something I've added so I didn't really check what is the reason but can look into it when working on the one below.

tests/functional/spec/dynamic_answers_list_value_source.spec.js
```diff
@@ -0,0 +1,167 @@
...
+    await $(SectionSummaryPage.supermarketsListAddLink()).click();
+    await $(ListCollectorAddPage.supermarketName()).setValue("Aldi");
+    await $(ListCollectorAddPage.submit()).click();
+    await $(ListCollectorPage.no()).click();
+    await $(ListCollectorPage.submit()).click();
+    await expect(await $$(labels)[0].getText()).to.equal("Percentage of shopping at Tesco");
+    await expect(await $$(labels)[1].getText()).to.equal("Percentage of shopping at Aldi");
+    await expect(await $$(labels).length).to.equal(2);
+  });
+  it("Given list items have been added and the dynamic answers are submitted, When the summary is displayed, Then the correct answers should be visible and have correct values", async () => {
```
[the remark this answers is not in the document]
petechd: I don't think we cover such validation in this iteration, that would require changing question json based on answer/list store on top of what we try to add here.

app/questionnaire/questionnaire_schema.py
petechd: This has a comment below it, I will move it above.

kylelawsonAND: ah my bad sorry!

rmccar (approved):

rmccar (commented):
Once tests are passing looks ok to me

kylelawsonAND:
Once all other people's comments are resolved, approval from me 👍

petechd:
> Once tests are passing looks ok to me

I had to rename schemas in tests by hand since PyCharm didn't refactor for me very well.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -145,3 +151,21 @@ def add_relationships_unrelated_answers(
...
+    dynamic_answers = block["dynamic_answers"]
+    for answer in dynamic_answers["answers"]:
+        values = dynamic_answers["values"]
+        if values["source"] == "list":
+            for list_item_id in list_store[values["identifier"]].items:
+                extracted_answer = answer_store.get_answer(
+                    answer["id"], list_item_id=list_item_id
+                )
+                answers_payload.add_or_update(extracted_answer)  # type: ignore
+                # answer store's get_answer return is optional but at this point it always returns an Answer
```
Yuyuutsu, line 171:
```suggestion
                # Type Ignore: answer store's get_answer return is optional but at this point it always returns an Answer
```

Yuyuutsu (commented):
Tested against scenarios and everything checks out. Also ran added unit tests through debugger works as expected 👍

app/views/contexts/summary/question.py
```diff
@@ -95,10 +108,58 @@ def _build_answers(
...
+                    summary_answers.append(summary_answer)
+
         for answer_schema in self.answer_schemas:
-            answer_value: Optional[AnswerValueEscapedTypes] = self.get_answer(
-                answer_store, answer_schema["id"]
-            )
+            if self.dynamic_answer_schemas:
+                self.list_item_id = None
```
MebinAbraham, line 161: This isn't right?. You might have a repeating answer?

app/views/contexts/summary/question.py
```diff
@@ -95,10 +108,58 @@ def _build_answers(
...
+                    self.dynamic_answer_identifier
+                ).items:
+                    placeholder_renderer = PlaceholderRenderer(
+                        answer_store=self.answer_store,
+                        list_store=self.list_store,
+                        progress_store=self.progress_store,
+                        schema=self.schema,
+                        language="en",
+                        metadata=None,
+                        response_metadata={},
```
MebinAbraham, line 124: These need to be passed in, not empty/none.

app/views/contexts/summary/question.py
```diff
@@ -95,10 +108,58 @@ def _build_answers(
             ]
 
         summary_answers = []
+
+        if self.dynamic_answer_schemas:
+            for answer_schema in self.dynamic_answer_schemas:
+                for list_item_id in self.list_store.get(
+                    self.dynamic_answer_identifier
+                ).items:
+                    placeholder_renderer = PlaceholderRenderer(
```
MebinAbraham, line 117: Could we not resolve the schema upfront then iterate over it, then we can handle dynamic/static with 1 loop?.

MebinAbraham (commented):
Have made an update to the branch based on our discussions. Dynamic is resolved and added to answers in placeholders now. Gives us code safety.

- The form still needs to be looked at
- We need to see if we can simplify summary. I noticed if placeholder is called on the entire block it doesn't resolve dynamic answers, if we can make it work then in summary we could just resolve it before doing the whole "build_answers" which allows us to not "resolve" it in multiple places, in summary you are manually resolving the looping items.
- `_get_answers_for_question` needs looking at as it still uses privates properties.

Can discuss these Fri/Mon.

app/views/contexts/summary/question.py
petechd: Removed now.

app/submitter/convert_payload_0_0_3.py
```diff
@@ -68,6 +68,12 @@ def convert_answers_to_payload_0_0_3(
                         list_collector_block=block,
                         answers_payload=answers_payload,
                     )
+                if (question := block.get("question")) and (
```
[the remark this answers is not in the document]
petechd: Not necessarily, could be list collector type of block?

app/views/contexts/summary/question.py
petechd: So this one is refactored now, it is resolved before `for schema in schemas` loop.

app/views/contexts/summary/question.py
petechd: Do we need metadata/response metadata for dynamic answers placeholders, I thought we only need list/answer sources for labels here?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -158,6 +158,17 @@ def remove_list_item_and_answers(self, list_name: str, list_item_id: str):
...
+        for list_collector in self._schema.get_list_collectors_for_list(
+            for_list=list_name,
+            section=self._schema.get_section(self._current_location.section_id),  # type: ignore
+            # Section and answer_id below must exist at this point
+        ):
+            answer_block = self._schema.get_add_block_for_list_collector(
+                list_collector["id"]
+            )
+            answer_id = self._schema.get_first_answer_id_for_block(answer_block["id"])  # type: ignore
+            self._capture_block_dependencies_for_answer(answer_id)
```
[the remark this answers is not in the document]
petechd: Ok, doing it for all items now.

app/views/handlers/list_remove_question.py
petechd: This one is removed now.

app/views/contexts/summary/question.py
```diff
@@ -71,6 +92,9 @@ def _build_answers(
         list_name: Optional[str],
```
kylelawsonAND, line 92: While modifying this method, may as well change optionals to |

app/views/contexts/summary/question.py
```diff
@@ -51,17 +68,21 @@ def __init__(
...
+        self, answer_store: AnswerStore, answer_id: str, list_item_id: str | None = None
     ) -> Optional[AnswerValueEscapedTypes]:
         answer = answer_store.get_answer(
-            answer_id, self.list_item_id
+            answer_id, list_item_id or self.list_item_id
         ) or self.schema.get_default_answer(answer_id)
 
         return escape_answer_value(answer.value) if answer else None
 
+    # pylint: disable=too-many-locals
```
kylelawsonAND, line 85: I'd be inclined to listen to the pylint and split this method up into some smaller methods.

app/views/handlers/question.py
```diff
@@ -113,12 +113,42 @@ def get_next_location_url(self):
             self._return_to_block_id,
         )
 
+    # pylint: disable=protected-access
```
kylelawsonAND, line 116: Similar to prev comment, if this is now being used non-privately, consider making public or creating a public alias to safely call it.

kylelawsonAND (changes requested):
All tests working as expected - some non-func comments

app/views/handlers/question.py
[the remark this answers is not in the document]
MebinAbraham: We still shouldn't be mutating the form. Runner should be able to work out if it is dynamic or not and handle it accordingly.

This can be reverted and `update_answers` can be updated to handle this in a much simpler way due to the changes we made to how placeholder handles this.

```py
def update_answers(
    self, form_data: Mapping[str, Any], list_item_id: Optional[str] = None
) -> None:
    list_item_id = list_item_id or self._current_location.list_item_id
    answers_by_answer_id = self._schema.get_answers_for_question_by_id(self._current_question)

    for answer_id, answer_value in form_data.items():
        if answer_id not in answers_by_answer_id:
            continue

        resolved_answer = answers_by_answer_id[answer_id]
        answer_id_to_use = resolved_answer.get("original_answer_id") or answer_id
        list_item_id_to_use = resolved_answer.get("list_item_id") or list_item_id

        answer_updated = self._update_answer(answer_id_to_use, list_item_id_to_use, answer_value)
        if answer_updated:
            self._capture_section_dependencies_for_answer(answer_id_to_use)
            self._capture_block_dependencies_for_answer(answer_id_to_use)
```

With this change you can remove the form manipulation and let update answer handles answer updation.

app/views/contexts/summary/question.py
```diff
@@ -16,25 +18,40 @@
...
+        metadata: MetadataProxy | None,
+        response_metadata: Mapping,
+        language: str,
     ) -> None:
         self.list_item_id = location.list_item_id if location else None
         self.id = question_schema["id"]
         self.type = question_schema["type"]
         self.schema = schema
-        self.answer_schemas = iter(question_schema["answers"])
+        self.answer_schemas = iter(question_schema.get("answers", []))
```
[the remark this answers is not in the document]
petechd: Needed in:
```python
def _build_date_range_answer(
        self, answer_store: AnswerStore, answer: Optional[AnswerValueEscapedTypes]
    ) -> dict[str, Optional[AnswerValueEscapedTypes]]:
       next_answer = next(self.answer_schemas)
```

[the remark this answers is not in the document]
petechd: For `next`, sorry I didn't clarify, breaks on `next()`.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -438,8 +438,9 @@ def started_section_keys(
     ) -> list[SectionKeyType]:
         return self._progress_store.started_section_keys(section_ids)
 
-    def dependency_has_dynamic_answers(self, block_id):
-        if question := self._schema.get_block(block_id).get("question"):
+    def dependency_has_dynamic_answers(self, block_id: str) -> bool:
```
MebinAbraham, line 441: This would error for variant, but I still don't get why this is needed?. There is an outstanding comment for it.

app/questionnaire/placeholder_renderer.py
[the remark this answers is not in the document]
MebinAbraham: Redundant now.

app/questionnaire/placeholder_renderer.py
```diff
@@ -141,13 +149,70 @@ def render(
...
+                .serialize().get("items")
+            )
+        )
+        for dynamic_answer in dynamic_answers["answers"]:
+            for item in list_items:  # type: ignore
+                resolved_dynamic_answer = self._schema.get_mutable_deepcopy(
+                    dynamic_answer
+                )
+                resolved_id = f"{dynamic_answer['id']}-{item}"
+                resolved_dynamic_answer["id"] = resolved_id
```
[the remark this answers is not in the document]
MebinAbraham: Redundant now.

app/views/contexts/summary/question.py
```diff
@@ -95,9 +119,29 @@ def _build_answers(
...
+                language=language,
+                metadata=metadata,
+                response_metadata=response_metadata,
+            )
+
+            resolved_question = placeholder_renderer.render(
+                data_to_render=question_schema, list_item_id=self.list_item_id
+            )
+
+        for answer_schema in resolved_question["answers"]:
```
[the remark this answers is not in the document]
petechd: I've used the second approach but happy to update `get_mutable_deepcopy` if needed.

app/views/handlers/question.py
```diff
@@ -254,9 +254,8 @@ def evaluate_and_update_section_status_on_list_change(self, list_name):
         )
 
         for section_id, list_item_id in section_keys_to_evaluate:
-            path = self.router.routing_path(section_id, list_item_id)
             self.questionnaire_store_updater.update_section_status(
-                is_complete=self.router.is_path_complete(path),
+                is_complete=self.router.is_path_complete(self._routing_path),
```
MebinAbraham, line 258: This shouldn't have changed?

petechd: We don't want to re-evaluate the path so where is path supposed to be from if not self?

app/questionnaire/questionnaire_store_updater.py
petechd: I think we don't need this now since we refactored dynamic answers resolution (and ids resolution till the very last step before form is displayed). I've tested it and works fine so will remove it now.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -403,3 +424,9 @@ def _remove_dependent_blocks_and_capture_dependent_sections(self) -> None:
 
     def started_section_keys(self, section_ids: Optional[Iterable[str]] = None):
         return self._progress_store.started_section_keys(section_ids)
+
+    def dependency_has_dynamic_answers(self, block_id):
+        if question := self._schema.get_block(block_id).get("question"):
```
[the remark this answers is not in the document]
petechd: See https://github.com/ONSdigital/eq-questionnaire-runner/pull/1079#discussion_r1179005877

app/views/handlers/question.py
MebinAbraham: The path of the current section in the loop. Right now this is evaluating the path for the current location for every section to evaluate.

We don't want to reevaluate the same path multiple times. Prior to your change, it would have evaluated the path for these "sections" and then when `update_progress_for_dependent_sections` was called any sections that are dependents based on any of our other dependencies would have been reevaluated. However, the sections here and the dependencies ones could overlap hence you would be evaluating the path for the same section multiple times.

Instead of evaluating the path, add it to `self._dependent_sections` which will reevaluate it when `update_progress_for_dependent_sections` is called.

app/views/handlers/question.py
petechd: This code now reverted to its original state.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -328,6 +342,13 @@ def update_answers(
         answer_ids_for_question = self._schema.get_answer_ids_for_question(
             self._current_question
         )
+        for item in answer_ids_for_question:
+            if (
+                item.split("-")[-1]
+                in self._list_store._list_item_ids()  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
petechd: Code refactored and protected access issue resolved.

app/views/handlers/question.py
petechd: Adding this change breaks some of our existing behaviour, for example `return_to_url` parameter doesn't persist after clicking previous on edit or remove pages for list collector. So I am not sure if we should carry out this refactor? I've tried to only update dependencies instead of re-evaluating like this:
```python
        for section_id, list_item_id in section_keys_to_evaluate:
            path = self.router.routing_path(section_id, list_item_id)
            self.questionnaire_store_updater.dependent_sections.add(
                DependentSection(
                    section_id=section_id,
                    list_item_id=list_item_id,
                    is_complete=self.router.is_path_complete(path),
                )
            )
```

app/views/handlers/list_remove_question.py
petechd: Related comment https://github.com/ONSdigital/eq-questionnaire-runner/pull/1079#discussion_r1180014934

petechd: OK, updated it all as suggested, can't find method name suggestion so I tried to be creative.

kylelawsonAND (approved):
Re-ran testing scenarios ✅ 
Nothing else to raise in code changes ✅

app/views/handlers/question.py
```diff
@@ -210,22 +225,20 @@ def get_return_to_hub_url(self):
         ):
             return url_for(".get_questionnaire")
 
-    def evaluate_and_update_section_status_on_list_change(self, list_name):
+    def add_dependent_sections_for_list(self, list_name):
```
MebinAbraham, line 228: Minor, for consistency with other method names, maybe call it `capture` instead of `add?.

app/views/handlers/question.py
MebinAbraham: What was the issue around `return_to`?

app/submitter/convert_payload_0_0_3.py
```diff
@@ -145,3 +151,20 @@ def add_relationships_unrelated_answers(
             )
         ):
             answers_payload.add_or_update(unrelated_answer)
+
+
+def resolve_dynamic_answers(
+    block: dict,
```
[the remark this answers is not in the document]
MebinAbraham: Outstanding.

schemas/test/en/test_dynamic_answers_list_source.json
```diff
@@ -0,0 +1,462 @@
...
+                                                    }
+                                                ]
+                                            },
+                                            "id": "amount-of-shopping",
+                                            "mandatory": false,
+                                            "type": "Currency",
+                                            "currency": "GBP",
+                                            "decimal_places": 2,
+                                            "description": {
+                                                "text": "The answers must be less than {max_value} and greater than {min_value}",
```
MebinAbraham, line 419: `less than or equal {max_value} and greater than or equal {min_value}`. Current equals would work. Either change this or the validation.

app/views/handlers/list_collector.py
```diff
@@ -54,6 +54,8 @@ def handle_post(self):
             # pylint: disable=no-member
             # wtforms Form parents are not discoverable in the 2.3.3 implementation
             self.questionnaire_store_updater.update_answers(self.form.data)
+            self.questionnaire_store_updater.update_progress_for_dependent_sections()
             self.questionnaire_store_updater.save()
         else:
+            self.questionnaire_store_updater.update_progress_for_dependent_sections()
```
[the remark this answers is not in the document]
petechd: Both this one and the one in primary removed.

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=add-dynamic-answers-validation
```
```suggestion
tag=latest
```

app/submitter/convert_payload_0_0_3.py
```diff
@@ -145,3 +160,20 @@ def add_relationships_unrelated_answers(
             )
         ):
             answers_payload.add_or_update(unrelated_answer)
+
+
+def resolve_dynamic_answers(
+    block: Mapping,
```
MebinAbraham, line 166: This is a question not block.

app/views/handlers/list_collector.py
```diff
@@ -54,6 +54,7 @@ def handle_post(self):
             # pylint: disable=no-member
             # wtforms Form parents are not discoverable in the 2.3.3 implementation
             self.questionnaire_store_updater.update_answers(self.form.data)
+            self.questionnaire_store_updater.update_progress_for_dependent_sections()
```
MebinAbraham, line 57: Is this needed?. As before, length won't change on a collector page?

app/views/handlers/primary_person_list_collector.py
```diff
@@ -47,5 +48,6 @@ def handle_post(self):
             # `super().handle_post()`. The section status will eventually get
             # determined correctly when the parent class' `update_section_status`
             # method is called.
-            self.evaluate_and_update_section_status_on_list_change(list_name)
+            self.add_dependent_sections_for_list(list_name)
+            self.questionnaire_store_updater.update_progress_for_dependent_sections()
```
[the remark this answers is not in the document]
MebinAbraham: The one in the main `if` isn't needed either is it?

app/views/handlers/question.py
```diff
@@ -195,12 +210,12 @@ def handle_post(self):
         # pylint: disable=no-member
         # wtforms Form parents are not discoverable in the 2.3.3 implementation
         self.questionnaire_store_updater.update_answers(self.form.data)
-        self.questionnaire_store_updater.update_progress_for_dependent_sections()
         if self.questionnaire_store_updater.is_dirty():
             self._routing_path = self.router.routing_path(
                 section_id=self._current_location.section_id,
                 list_item_id=self._current_location.list_item_id,
             )
+        self.questionnaire_store_updater.update_progress_for_dependent_sections()
```
MebinAbraham, line 218: This would result in a duplicate call as the super handle post also calls this.

tests/app/submitter/test_convert_payload_0_0_3.py
```diff
@@ -1483,3 +1483,55 @@ def test_no_answers_codes_in_payload_when_no_questions_answered(version):
 
     # Then
     assert "answer_codes" not in data_payload
+
+
+@pytest.mark.parametrize(
+    "version",
+    (AuthPayloadVersion.V2,),
+)
```
MebinAbraham, line 1491: How come v1 (None) is not tested?

tests/functional/spec/features/dynamic_answers_list_value_source.spec.js
```diff
@@ -0,0 +1,192 @@
...
+import SetMinimumPage from "../../generated_pages/dynamic_answers_list_source/minimum-spending.page";
+import SectionSummaryPage from "../../generated_pages/dynamic_answers_list_source/section-summary.page";
+
+describe("Dynamic answers list value source", () => {
+  const labels = 'label[class="ons-label"]';
+  const inputs = 'input[class="ons-input ons-input--text ons-input-type__input ons-input-number--w-3"]';
+  const group = 'div[id="group-2"]';
+  const summaryTitles = 'dt[class="ons-summary__item-title"]';
+  const summaryValues = 'dd[class="ons-summary__values"]';
+  const summaryActions = 'dd[class="ons-summary__actions"]';
```
MebinAbraham, line 16: This is not ideal and prone to error but being pragmatic so we can get this in. We can revisit generate pages/selectors as a whole when we have a bit more time.

app/views/handlers/list_collector.py
petechd: Not needed, thanks.

app/views/handlers/primary_person_list_collector.py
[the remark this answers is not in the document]
petechd: We don't `super` `handle_post` in `if` so it is still needed?

app/views/handlers/question.py
petechd: It doesn't update progress for dependent sections in some scenarios when we rely on original `handle_post` `update_progress_for_dependent_sections` (some edit answers).  Dependant sections are not re-enabled if we don't run `update_progress_for_dependent_sections` before calling `super().handle_post()`.

petechd: It seems like we don't do the capture step on edit answers as answers are not updated so this `if` doesn't apply:
```python
answer_updated = self._update_answer(
                answer_id_to_use, list_item_id_to_use, answer_value
            )
            if answer_updated:
                self._capture_section_dependencies_for_answer(answer_id_to_use)
                self._capture_block_dependencies_for_answer(answer_id_to_use)
```

app/views/handlers/primary_person_list_collector.py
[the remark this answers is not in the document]
MebinAbraham: But it is the collector, which doesn't change the length right?.

tests/functional/spec/features/dynamic_answers_list_value_source.spec.js
```diff
@@ -171,8 +171,8 @@ describe("Dynamic answers list value source", () => {
     await expect(await $(group).$$(summaryValues)[0].getText()).to.equal("12%");
     await expect(await $(group).$$(summaryTitles)[1].getText()).to.equal("Percentage of shopping at Aldi");
     await expect(await $(group).$$(summaryValues)[1].getText()).to.equal("21%");
-    await expect(await $(group).$$(summaryTitles).length).to.equal(6);
-    await expect(await $(group).$$(summaryValues).length).to.equal(6);
+    await expect(await $(group).$$(summaryTitles).length).to.equal(8);
```
MebinAbraham, line 174: We should assert the values also like you have done for some.

app/views/handlers/question.py
petechd: Hard to explain, it's just that if you don't add that line, which looks like duplication, it is not called at all. Probably best if we have a chat.

petechd: This one is done now, fixed. It was order of method calls in block's handle_post that was incorrect. Needs to be:
```python
    def handle_post(self):
        self._set_started_at_metadata()
        self.questionnaire_store_updater.update_progress_for_dependent_sections()
        self._update_section_completeness()
```
Instead of:
```python
    def handle_post(self):
        self._set_started_at_metadata()
        self._update_section_completeness()
        self.questionnaire_store_updater.update_progress_for_dependent_sections()
```

MebinAbraham: This would break it for @berroar's progress to change. As dependent section will be using the old status.

app/views/handlers/question.py
petechd: Addressed now.

app/views/handlers/question.py
petechd: Addressed.

MebinAbraham (approved):
👍

berroar (approved):

PR #1085, opened 2023-04-06
Add type hints to questionnaire module

katie-gardner:
This PR is adding type hints to the questionnaire module (and altering the types in a few other places if affected by the changes to the questionnaire module) as part of the ongoing goal to type hint the code base

### How to review
Check that mypy is correctly enforcing the hints, and that any type: ignores were correct to be ignored
To test myself I also temporarily used the typeguard library on the questionnaire module and ran through various surveys.

### Checklist
* [ ] Mypy is enforcing types in the questionnaire module
* [ ] Types in the questionnaire module are using the new syntax


katie-gardner:
[Type hints: Add type hints to the Questionnaire module (2 SP)](https://trello.com/c/r5qGtJ4W/5657-type-hints-add-type-hints-to-the-questionnaire-module-2-sp)

Yuyuutsu (commented):
My review is just adding the prefix `Type ignore: ` before comments. Other than that looks good to me 👍

app/questionnaire/path_finder.py
```diff
@@ -251,13 +254,15 @@ def evaluate_skip_conditions(
...
-            return self.schema.get_first_block_id_for_group(rule["group"])
-        return rule["block"]
+            # by this point the block for the rule will exist
+            return self.schema.get_first_block_id_for_group(rule["group"])  # type: ignore
+        # the rules block will be a string
+        return rule["block"]  # type: ignore
 
     def _remove_current_blocks_answers_for_backwards_routing(
-        self, rule: dict, this_location: Location
+        self, rule: Mapping, this_location: Location
```
petechd, line 265: From our type hints doc: `Specify at least the top-level type parameters for all generic types`.

app/questionnaire/path_finder.py
```diff
@@ -251,13 +254,15 @@ def evaluate_skip_conditions(
 
         return when_rule_evaluator.evaluate(skip_conditions["when"])
 
-    def _get_next_block_id(self, rule):
+    def _get_next_block_id(self, rule: Mapping) -> str:
```
petechd, line 257: Same as the comment below.

app/questionnaire/placeholder_parser.py
```diff
@@ -178,7 +178,7 @@ def _resolve_value_source_list(
 
     def _get_routing_path_block_ids(
         self, data: Mapping | Sequence, sections_to_ignore: list | None = None
```
petechd, line 180: Since we are here we could add some top level arguments to these generic types as well.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -106,42 +110,46 @@ def update_relationship_question_completeness(self, list_name: str) -> None:
...
                     self.add_completed_location(location)
 
-    def _get_relationship_collectors_by_list_name(self, list_name: str):
+    def _get_relationship_collectors_by_list_name(
+        self, list_name: str
+    ) -> list[ImmutableDict] | None:
         return self._schema.get_relationship_collectors_by_list_name(list_name)
 
-    def _get_relationships_in_answer_store(self, relationship_answer_id: str):
+    def _get_relationships_in_answer_store(self, relationship_answer_id: str) -> list[RelationshipDict]:  # type: ignore
```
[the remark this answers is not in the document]
katie-gardner: good point - removed

app/questionnaire/location.py
```diff
@@ -24,15 +24,15 @@ class Location:
...
+    list_name: str | None = None
+    list_item_id: str | None = None
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(frozenset(self.__dict__.values()))
 
     @classmethod
-    def from_dict(cls, location_dict: Mapping):
+    def from_dict(cls, location_dict: Mapping[str, str]) -> Location:
```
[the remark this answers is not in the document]
katie-gardner: I'm happy to add it, but what is the benefit in this case? It seems to be inferred without adding it

app/questionnaire/path_finder.py
katie-gardner: added

app/questionnaire/path_finder.py
katie-gardner: added

app/questionnaire/placeholder_parser.py
katie-gardner: Done

katie-gardner: I've added the top level args to some more generic types within the module as well, in most cases, looking at validator, it was instances of dictionaries with a possible combination of string, list or more dictionaries as values, so it seemed suitable to use Mapping[str, Any] as other places have done, but can revisit if not

petechd: Should be fine, since we do this in other places.

app/questionnaire/location.py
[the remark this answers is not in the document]
petechd: Yeah, since it is a class method of this class we can leave it as it is, maybe it's too much detail.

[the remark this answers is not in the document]
katie-gardner: Sounds good, will leave it be for this instance

app/questionnaire/questionnaire_schema.py
```diff
@@ -407,9 +422,9 @@ def _is_list_name_in_rule(self, when_rule: Mapping, list_name: str) -> bool:
         return False
 
     @staticmethod
-    def get_operands(rules: Mapping) -> list:
+    def get_operands(rules: Mapping[str, Any]) -> Sequence[Any]:
         operator = next(iter(rules))
-        operands: list = rules[operator]
+        operands: Sequence[Any] = rules[operator]
```
petechd, line 427: `Any` doesn't have a purpose here, it's the same as you'd give this sequence no attribute.

app/questionnaire/questionnaire_schema.py
```diff
@@ -459,56 +472,54 @@ def get_individual_response_show_on_hub(self) -> bool:
...
         if section := self.get_section(section_id):
             return section.get("title")
 
-    def get_show_on_hub_for_section(self, section_id: str) -> Optional[bool]:
+    def get_show_on_hub_for_section(self, section_id: str) -> bool | None:
         if section := self.get_section(section_id):
             return section.get("show_on_hub", True)
 
-    def get_summary_for_section(self, section_id: str) -> Optional[ImmutableDict]:
+    def get_summary_for_section(self, section_id: str) -> ImmutableDict | None:
```
petechd, line 489: Can we give returned `immutable dicts` top level arguments? There is few more below.

app/questionnaire/questionnaire_schema.py
```diff
@@ -714,7 +723,7 @@ def get_unrelated_block_no_answer_values(
             ]
 
     @staticmethod
-    def get_single_string_value(schema_object: Union[Mapping, str]) -> str:
+    def get_single_string_value(schema_object: Mapping | str) -> str:
```
petechd, line 726: And these mappings here and line 858 need top level arguments, same with any sequences.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -16,15 +19,15 @@
 class QuestionnaireStoreUpdater:
     """Component responsible for any actions that need to happen as a result of updating the questionnaire_store"""
 
-    EMPTY_ANSWER_VALUES: Tuple = (None, [], "", {})
+    EMPTY_ANSWER_VALUES: tuple = (None, [], "", {})
```
petechd, line 22: Tuple of?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -195,12 +205,12 @@ def update_same_name_items(
...
                 same_name_items |= {current_list_item_id, matching_list_item_id}
             else:
-                people_names[current_list_item_name] = current_list_item_id  # type: ignore
+                people_names[current_list_item_name] = current_list_item_id
 
         list_model.same_name_items = list(same_name_items)  # type: ignore
 
     def remove_relationship_answers_for_list_item_id(
-        self, list_item_id: str, answers: List
+        self, list_item_id: str, answers: list[Any]
```
petechd, line 213: Can we avoid using `Any` here?

app/submitter/convert_payload_0_0_1.py
```diff
@@ -50,7 +53,7 @@ def convert_answers_to_payload_0_0_1(
             for answer_in_block in answers_in_block:
                 answer_schema = None
 
-                block = schema.get_block_for_answer_id(answer_in_block.answer_id)
+                block: ImmutableDict = schema.get_block_for_answer_id(answer_in_block.answer_id)  # type: ignore
```
petechd, line 56: Some arguments for this `immutable dict`?

app/questionnaire/questionnaire_store_updater.py
MebinAbraham: `Any` is not needed at the top level. `list` is implicitly `list[Any]`

app/questionnaire/relationship_router.py
```diff
@@ -87,12 +89,12 @@ def _individual_relationships_routing_path(
...
+                    # Type ignore: unrelated_answer_id will exist if the unrelated_block_id does
+                    self.unrelated_answer_id,  # type: ignore
+                    from_list_item_id,
                 )
-                if (
-                    unrelated_answer
-                    and unrelated_answer.value in self.unrelated_no_answer_values
-                ):
+                # Type ignore: unrelated_no_answer_values exists when the unrelated answer does
+                if unrelated_answer and unrelated_answer.value in self.unrelated_no_answer_values:  # type: ignore
```
MebinAbraham, line 97: This type ignore can be removed if we just initialise as empty list.
```py
self.unrelated_no_answer_values = unrelated_no_answer_values or []
```

app/questionnaire/variants.py
```diff
@@ -1,22 +1,33 @@
...
+    schema: QuestionnaireSchema,
+    metadata: MetadataProxy | None,
+    response_metadata: Mapping[str, Any],
+    answer_store: AnswerStore,
+    list_store: ListStore,
+    variants_key: str,
+    single_key: str,
+    current_location: Location | RelationshipLocation,
+    progress_store: ProgressStore,
+) -> Mapping[str, Any]:
```
MebinAbraham, line 27: Return types should be concrete types like `dict`.

app/questionnaire/variants.py
```diff
@@ -31,19 +42,20 @@ def choose_variant(
...
-):
+    block: ImmutableDict[str, Any],
+    schema: QuestionnaireSchema,
+    metadata: MetadataProxy | None,
+    response_metadata: Mapping[str, Any],
+    answer_store: AnswerStore,
+    list_store: ListStore,
+    current_location: Location | RelationshipLocation,
+    progress_store: ProgressStore,
+) -> Mapping[str, Any]:
```
MebinAbraham, line 58: Return types should be concrete types like `dict`.

app/questionnaire/variants.py
```diff
@@ -59,15 +71,15 @@ def choose_question_to_display(
...
```
MebinAbraham, line 82: Return types should be concrete types like `dict`.

MebinAbraham (commented):
Looks good, nothing major.

app/questionnaire/questionnaire_store_updater.py
katie-gardner: Since the type will be inferred I've dropped this hint

app/questionnaire/placeholder_parser.py
```diff
@@ -24,7 +26,7 @@
         PlaceholderRenderer,  # pragma: no cover
     )
 
-TransformedValueTypes = Union[None, str, int, Decimal, bool]
+TransformedValueTypes = None | str | int | Decimal | bool
```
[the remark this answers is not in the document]
katie-gardner: Fixed

app/questionnaire/location.py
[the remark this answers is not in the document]
katie-gardner: All for the self type, added that instead and removed the import

app/questionnaire/relationship_location.py
```diff
@@ -12,13 +12,13 @@ class RelationshipLocation:
...
-    to_list_item_id: Optional[str] = None
+    to_list_item_id: str | None = None
 
-    def for_json(self) -> Mapping:
+    def for_json(self) -> Mapping[str, Any]:
         attributes = vars(self)
         return {k: v for k, v in attributes.items() if v is not None}
 
-    def url(self, **kwargs) -> str:
+    def url(self, **kwargs: Any) -> str:
```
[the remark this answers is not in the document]
katie-gardner: it could include _external which is a bool, I wasn't able to type it because if I specify str | None | bool mypy complains not knowing which is which and I can't edit the method itself, open to any other ideas though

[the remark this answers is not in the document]
MebinAbraham: Ah okay, happy to keep as is then 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -166,8 +166,11 @@ def is_view_submitted_response_enabled(self) -> bool:
         is_enabled: bool = schema.get("view_response", False)
         return is_enabled
 
-    def _get_sections_by_id(self) -> dict[str, ImmutableDict]:
-        return {section["id"]: section for section in self.json.get("sections", [])}
+    def _get_sections_by_id(self) -> dict[str, ImmutableDict[str, Any]]:
+        return {
+            section["id"]: section
+            for section in self.json.get("sections", ImmutableDict({}))
```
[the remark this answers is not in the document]
katie-gardner: Yep, hundreds of test failures, looks like we have a lot of tests that use QuestionnaireSchema({}) or initialise it with a single key specific to that test, but since its always mandatory I can go through and update these with an empty sections dict so its closer to an actual run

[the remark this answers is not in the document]
MebinAbraham: I think leave it as it is for now, when we come to add actual data models (data classes for the schemas), we will have to handle it then.

app/questionnaire/path_finder.py
```diff
@@ -77,12 +77,12 @@ def _get_not_skipped_blocks_in_section(
         self,
         location: Location,
         routing_path_block_ids: list[str],
-        section: ImmutableDict,
+        section: ImmutableDict[str, Any],
         when_rules_block_dependencies: list[str],
-    ) -> Optional[list[Mapping]]:
+    ) -> list[Mapping[str, Any]] | None:
```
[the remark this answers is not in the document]
katie-gardner: Fixed - it forced me to change the _build_routing_path_block_ids function to accept blocks as list[dict], as mypy isn't liking the more generic list[Mapping]. I could ignore it but its only used here

app/questionnaire/placeholder_parser.py
```diff
@@ -74,8 +76,8 @@ def __init__(
         self._routing_path_block_ids_by_section_key: dict = {}
 
     def __call__(
-        self, placeholder_list: Sequence[Mapping]
-    ) -> MutableMapping[str, Union[ValueSourceEscapedTypes, ValueSourceTypes]]:
+        self, placeholder_list: Sequence[Mapping[str, Any]]
```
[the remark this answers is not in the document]
MebinAbraham: My opinion is if it is deterministic, this assign the types. So `dict[str, str]` but avoid `dict[str, str | list]` because that mean any key could be a list or str which isn't helpful. I would say:
1. Explicitly type if the values have the same type, i.e `dict[str, str]`
2. If the value can be multiple types, but the keys are deterministic and is not a schema partial, we should use `TypedDict`.
3. All else, just generics like `dict`.

Thoughts?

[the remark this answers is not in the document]
katie-gardner: I'm happy with that approach, I think thats a sensible choice so will go with that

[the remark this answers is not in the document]
kylelawsonAND: I think that approach sounds good, I'll update my PR's Mapping[str, Any] instances to Mapping (same with MutableMappings).

app/submitter/convert_payload_0_0_1.py
katie-gardner: ommitted due to being [str, Any]

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
katie-gardner: I've done the same, removed any [str, Any]s I've added alongside a couple of others in the questionnaire module just for consistency with my changes but I've left any outside of that module

app/questionnaire/questionnaire_schema.py
petechd: This still need changing.

app/questionnaire/questionnaire_store_updater.py
petechd: Still outstanding?

app/questionnaire/questionnaire_store_updater.py
petechd: Still outstanding?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -16,15 +19,15 @@
 class QuestionnaireStoreUpdater:
     """Component responsible for any actions that need to happen as a result of updating the questionnaire_store"""
 
-    EMPTY_ANSWER_VALUES: Tuple = (None, [], "", {})
+    EMPTY_ANSWER_VALUES: tuple[None, list, str, dict] = (None, [], "", {})
```
katie-gardner, line 22: @petechd - I think the comments are against a prior commit, should all be ok here?

app/questionnaire/questionnaire_schema.py
katie-gardner: I did miss this one though, thanks, fixed

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
rmccar: I haven't got a huge preference either way to be honest. I do see your point but I can also see that to me `Mapping[str, Any]` is a bit more explicit and adds a bit more explanation even if it is only a small amount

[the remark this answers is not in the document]
berroar: Happy with this approach 👍

app/questionnaire/questionnaire_store_updater.py
petechd: @katie-gardner thanks somehow I didn't see them, all good.

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
Yuyuutsu: Had a read through, sounds good 👍 . Should we update our type hinting gudie to reference this ?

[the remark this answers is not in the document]
petechd: `[str, Any]` doesn't bring much value imho, but we do Mappings with sets as keys etc. so it is not obvious it will always be `str` or `int`.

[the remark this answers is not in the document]
MebinAbraham: If the value is the same for all keys then we would still type it, regardless of it if is set etc.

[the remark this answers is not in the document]
MebinAbraham: @Yuyuutsu Good shout. Looks like @kylelawsonAND has done it in his PR 👍

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
MebinAbraham: That is due to invariance vs covariance. Have a read of this: https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance
To fix it, you can change it to `Sequence[Mapping]` then mypy will be happy.

app/questionnaire/questionnaire_store_updater.py
MebinAbraham: There are quite a few where `Any` can be dropped, i.e `Mapping[str, Any]`. However, I've approved it, will leave it to others to check that.

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
katie-gardner: Ah thank you! fixed

app/questionnaire/questionnaire_store_updater.py
katie-gardner: outdated, should be ok to resolve now

PR #1089, opened 2023-04-14
Update MAX_NUMBERS to 15 digits

Yuyuutsu:

Credit Granters and Insurance surveys require larger numbers to be collected. To be applied to numbers, units and currency.
The scenarios are on this [trello card](https://trello.com/c/0yJIjinU/5658-enable-up-to-15-digits-in-relevant-fields-1-sp)
~~Updated `MAX_NUMBERS` to 15 digits (13 whole numbers and 2 decimal points)~~
Updated  `MAX_NUMBERS` to 15 digits (15 whole numbers) and up to (6 decimal points)
Updated `test_numbers`, `test_currency` and `test_unit_pattern`

#### Functional 
1. Open a schema for example stocks, and see if you can enter 15 digits.
2. It should let you enter 13 whole numbers and 2 decimal points
3. Open `test_numbers` and test the limitations. It should allow for negative numbers in some of the questions. Test, the highest and lowest ranges (via looking at the schema) to see if correct messages are sent
4. Open `test_currency` and test it and see if it works fine
5. Open `test_unit_pattern` and test if the new min-max test works with the boundaries expected
#### Other

- Should the `test_numbers` be separated with two different pages (like in `test_currency`) so you can test negative in one page and positive in another page. Or is the way I have set it up okay
- Check if tests checks out all the [scenarios](https://trello.com/c/0yJIjinU/5658-enable-up-to-15-digits-in-relevant-fields-1-sp)
- Should I add more testing in the negative numbers for `test_unit_pattern` . Not sure which units would be the best to be tested when negative
### Checklist


Yuyuutsu:
[Enable up to 15 digits in relevant fields (1 SP)](https://trello.com/c/0yJIjinU/5658-enable-up-to-15-digits-in-relevant-fields-1-sp)

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=update-max-lenght-to-15-digits
```
Yuyuutsu, line 2: REMINDER
```suggestion
tag=latest
```

app/settings.py
```diff
@@ -126,7 +126,7 @@ def utcoffset_or_fail(date_value, key):
 EQ_SESSION_ID = "eq-session-id"
 
 EQ_LIST_ITEM_ID_LENGTH = 6
-MAX_NUMBER = 9999999999
+MAX_NUMBER = 9999999999999
```
MebinAbraham, line 129: Makes it easier to read with the `_` separator (it is only syntactic sugar). 
```suggestion
MAX_NUMBER = 9_999_999_999_999
```

schemas/test/en/test_currency.json
```diff
@@ -93,10 +93,93 @@
                                         "maximum": {
                                             "value": 1000000
                                         }
+                                    },
+                                    {
+                                        "id": "answer-jpy",
```
kylelawsonAND, line 98: This has the same id as the Yen input above it, so the value is tied to that input in the store is incorrect, meaning the value entered for this input is discarded and the test for default max value isn't working/enforced. 

https://user-images.githubusercontent.com/108286538/232751997-c6a16124-7a93-440c-baef-cfbe4862a4fc.mov

schemas/test/en/test_currency.json
```diff
@@ -93,10 +93,93 @@
...
+                                        }
+                                    },
+                                    {
+                                        "id": "negative-answer-min-max-range",
+                                        "label": "How much did you spend? (Max range)",
+                                        "mandatory": false,
+                                        "type": "Currency",
+                                        "currency": "GBP",
+                                        "minimum": {
+                                            "value": -999999999999
```
kylelawsonAND, line 177: Error message displayed for this doesn't seem right
Screenshot 2023-04-18 at 11 43 43

schemas/test/en/test_currency.json
```diff
@@ -93,10 +93,93 @@
...
+                                        "maximum": {
+                                            "value": 1000000
+                                        },
+                                        "minimum": {
+                                            "value": -1000000
+                                        }
+                                    },
+                                    {
+                                        "id": "negative-answer-min-max-range",
+                                        "label": "How much did you spend? (Max range)",
```
kylelawsonAND, line 172: Max range -> Min range?

schemas/test/en/test_currency.json
Yuyuutsu: These videos are pretty neat 👍

schemas/test/en/test_currency.json
Yuyuutsu: Changed to Min Range. Should I put min max range since it is both.

schemas/test/en/test_currency.json
```diff
@@ -9,7 +9,6 @@
     "description": "A questionnaire to test currency input type",
     "messages": {
```
Yuyuutsu, line 10: More of a question to other devs. Should I remove the NUMBER_TOO_LARGE message since it is generic and we have more dynamic error handling.

schemas/test/en/test_currency.json
Yuyuutsu: Updated should show the correct error message for all the times you exceed the minimum boundary now in any box. 👍

schemas/test/en/test_currency.json
```diff
@@ -93,10 +92,93 @@
                                         "maximum": {
                                             "value": 1000000
                                         }
+                                    },
+                                    {
+                                        "id": "answer-gbp-max-range",
+                                        "label": "How much did you spend? (Max range)",
```
kylelawsonAND, line 98: The error message generated when going above the max value is less descriptive ("Number is too large") than the one for going below the minimum ("Enter an answer more than or equal to X"). 
The latter is generated dynamically but I think a generic one specified in this schema is being used instead for the former.

Yuyuutsu: I was thinking about removing the generic error message, or making both messages generic. Any preferences you have or other devs ?

Yuyuutsu: Actually i shall make it generic since the previous was generic 👍

kylelawsonAND: How much bother is it to make it dynamic? The other test cases show dynamic errors, that feels more appropriate for the tester to know they've hit the ranges without having to go back and forth between the schema and tested survey

Yuyuutsu: No bother just removing the error messages at the top of the schema. I shall update it so its dynamic

kylelawsonAND: Ah nice one, so do error messages defined in the schema take precedence over dynamic ones created in runner?

Yuyuutsu: Yeah it should be, I think

app/settings.py
```diff
@@ -126,7 +126,7 @@ def utcoffset_or_fail(date_value, key):
 EQ_SESSION_ID = "eq-session-id"
 
 EQ_LIST_ITEM_ID_LENGTH = 6
-MAX_NUMBER = 9999999999
+MAX_NUMBER = 9_999_999_999_999
```
MebinAbraham, line 129: This is only max 13 digits. Card say max 15 for positive, 14 for negative. The decimal shouldn't matter?

Do we not enforce a MIN?

kylelawsonAND: I counted the 2 decimal places as part of the 15/14. The card says 15/14 **digits**, and it seems online consensus is digits == integer for cases like this i.e. decimals don't count, so I agree now it should be 15/14 whole numbers and decimals are don't count towards that. 
Revoking approval 👍

kylelawsonAND (changes requested):
Awaiting consensus on 15/14 whole numbers vs whole + decimals

app/settings.py
MebinAbraham: 👍 Ya, I think decimal should only matter if we were dealing with finite memory/characters.

@Yuyuutsu Check with Donna, what downstream can handle, should we be limiting the number of characters, or the total value to 15 integers. Ask if 15 digits with decimal is supported downstream. We currently support up to 6 decimals.

Yuyuutsu: Oh okay. It was my assumption, since the card said we already support 12 digits with only 10 9's in the max_number variable. I shall message Donna. 
I was going to add Min_number variable but in some functions we default it to 0. Which certain schemas rely on, I think like stocks schemas. They don't have a minimum set in the json.

Yuyuutsu: Spoke to donna, she says that she believes it should be total value is 15 integers. Also down stream is making sure they can acoomdate 👍

app/settings.py
```diff
@@ -126,7 +126,7 @@ def utcoffset_or_fail(date_value, key):
 EQ_SESSION_ID = "eq-session-id"
 
 EQ_LIST_ITEM_ID_LENGTH = 6
-MAX_NUMBER = 9999999999
+MAX_NUMBER = 900_999_999_999_999
```
kylelawsonAND, line 129: Should this not be **999**_999_999_999_999?

Yuyuutsu: Oops, updated 👍

app/settings.py
```diff
@@ -126,7 +126,7 @@ def utcoffset_or_fail(date_value, key):
 EQ_SESSION_ID = "eq-session-id"
 
 EQ_LIST_ITEM_ID_LENGTH = 6
-MAX_NUMBER = 9999999999
+MAX_NUMBER = 999_999_999_999_999
```
kylelawsonAND, line 129: Just commenting this here as I'm not sure where it should be exactly:

The negative limit is -999999999999999, 15 digits instead of 14 as shown in screenshot:

Screenshot 2023-04-20 at 16 40 06

Yuyuutsu: Yeah that is a schema problem . I just need to adjust the minimum values in the `test_currency `schema to 14 now instead of 12

kylelawsonAND: It's not a ticket AC but it is in the tech details that the min should be negative 14 9s, which I think should be enforced in runner, not in the schema. Maybe someone could clarify this?

Yuyuutsu: Right now default the minimum to 0, on line 41. If u change that value to take the `MIN_NUMBER`. It will be used in schemas that may not want negative answers, like in stocks schemas since they don't speicfy the min number. Not sure if we can change this, any other devs have a input on this?
image

MebinAbraham: No changes should be needed in runner. Validator currently checks schema min is greater than the min limit. However, it looks like it only checks static values and any reference values using answer sources are not checked, so we could update that. It is done in `NumberAnswerValidator.validate_value_in_limits`. There is a method on questionnaire schema to get a referred value from an answer source: `_get_numeric_value_for_value_source`

schemas/test/en/test_numbers.json
```diff
@@ -187,11 +187,11 @@
                                     },
                                     {
                                         "id": "test-min",
-                                        "label": "Min Test (123 to 9,999,999,999)",
+                                        "label": "Min Test (-123 to 9,999,999,999,999)",
```
kylelawsonAND, line 190: the enforced limit in this test is actually -123 to 999,999,999,999,999, not -123 to 9,999,999,999,999

schemas/test/en/test_numbers.json
```diff
@@ -205,7 +205,7 @@
                                     },
                                     {
                                         "id": "test-min-exclusive",
-                                        "label": "Min Exclusive Test (124 to 9,999,999,999 - 123 Exclusive)",
+                                        "label": "Min Exclusive Test (124 to 9,999,999,999,999 - 123 Exclusive)",
```
kylelawsonAND, line 208: Same as prev comment - enforced upper limit here is 999,999,999,999,999, not 9,999,999,999,999 as stated in test description.

kylelawsonAND (changes requested):

schemas/test/en/test_currency.json
```diff
@@ -93,10 +91,93 @@
                                         "maximum": {
                                             "value": 1000000
                                         }
+                                    },
```
katie-gardner, line 94: The inputs allowing decimal places are all size bound, so is it worth having an additional input like this one for decimal places at min/max range? ( I can't see one in any of the other updated tests )

schemas/test/en/test_numbers.json
Yuyuutsu: Updated 👍

schemas/test/en/test_numbers.json
Yuyuutsu: Updated 👍

schemas/test/en/test_currency.json
```diff
@@ -93,10 +91,93 @@
...
+                                "type": "General"
+                            }
+                        },
+                        {
+                            "type": "Question",
+                            "id": "negative-block",
+                            "question": {
+                                "answers": [
+                                    {
+                                        "id": "negtaive-answer-gbp",
```
katie-gardner, line 114: typo - probably doesn't matter here though

tests/app/test_jinja_filters.py
```diff
@@ -237,7 +237,7 @@ def test_format_duration(duration, formatted_duration, app):
 @pytest.mark.parametrize(
     "answer, width",
     (
-        ({}, 10),
+        ({}, 15),
```
katie-gardner, line 240: at the moment, if I allow decimal places, I can enter a 15 digit number _and_ two decimal places as the limit, should that be tested if thats correct?

katie-gardner (changes requested):
For whole numbers all looks good although if allowing decimals on the max-range fields I can enter 15 digits as well as two decimal points rather than the 13 + 2 stated on the PR, due to the limit just being 999,999,999,999,999
One to fix or if that was intended should description be updated?

schemas/test/en/test_currency.json
Yuyuutsu: Updated 👍

tests/app/test_jinja_filters.py
Yuyuutsu: Sorry forgot to update the PR. It should say we should allow 15 whole intergers + up to 6 decimals. Will update PR

schemas/test/en/test_currency.json
Yuyuutsu: Yeah I don't see why not, I shall update `test_numbers` to test decimal boundaries 👍

Yuyuutsu: Update: so we can implement the 6 decimals in test_numbers when setting boundaries up to 6 however, there is currently a bug where it would not display past the fist 3 decimals. So if you entered the boundires as 1`0.00315` and `1000.1234` you would get in the range collum. 10 and 1000.123. 
image
image

rmccar:
Screenshot 2023-05-02 at 12 27 12

On the test numbers schema it is adding more errors to the error list than there are inputs when I enter a number that is too small. Numbers seem to be limited to 2 decimal places whereas I thought we would support 6?

Yuyuutsu:
> Screenshot 2023-05-02 at 12 27 12
> 
> On the test numbers schema it is adding more errors to the error list than there are inputs when I enter a number that is too small. Numbers seem to be limited to 2 decimal places whereas I thought we would support 6?

I dont seem to be getting that issue. The test currently limits the decimals to 2 but runner has already supported 6 decimals.  Same with Katies comments above I can update it to use 6 but it wont show 6 decimals on the screen it only displays up to 3 decimal points. 
image

MebinAbraham:
> > On the test numbers schema it is adding more errors to the error list than there are inputs when I enter a number that is too small. Numbers seem to be limited to 2 decimal places whereas I thought we would support 6?
> 
> I dont seem to be getting that issue. The test currently limits the decimals to 2 but runner has already supported 6 decimals. Same with Katies comments above, I can update it to use 6 but it wont show 6 decimals on the screen it only displays up to 3 decimal points. Up to you guys if I should change it to 6, happy to change it 👍 

I suggest keeping it as is, validator enforces max 6 decimal, but schema limit might be 2 etc. There is a known issue where decimals don't format to more than 2 or 3. At the moment, live surveys only use like 2, so not an issue yet but we will need to support more for Complex 2.

rmccar:
> > On the test numbers schema it is adding more errors to the error list than there are inputs when I enter a number that is too small. Numbers seem to be limited to 2 decimal places whereas I thought we would support 6?
> 
> I dont seem to be getting that issue. The test currently limits the decimals to 2 but runner has already supported 6 decimals. Same with Katies comments above, I can update it to use 6 but it wont show 6 decimals on the screen it only displays up to 3 decimal points. Up to you guys if I should change it to 6, happy to change it 👍 image

In that case leave it, It could make the functional tests confusing.
As for the error I just tried it again and same thing happened, this is how I created it:
- I spun up your branch
- loaded the `test_numbers` schema
- entered "-99999999999999.999999" in the first input
- left the second input blank
- clicked the submit button

MebinAbraham:
> > > On the test numbers schema it is adding more errors to the error list than there are inputs when I enter a number that is too small. Numbers seem to be limited to 2 decimal places whereas I thought we would support 6?
> > 
> > 
> > I dont seem to be getting that issue. The test currently limits the decimals to 2 but runner has already supported 6 decimals. Same with Katies comments above, I can update it to use 6 but it wont show 6 decimals on the screen it only displays up to 3 decimal points. Up to you guys if I should change it to 6, happy to change it 👍 image
> 
> In that case leave it, It could make the functional tests confusing. As for the error I just tried it again and same thing happened, this is how I created it:
> 
> * I spun up your branch
> * loaded the `test_numbers` schema
> * entered "-99999999999999.999999" in the first input
> * left the second input blank
> * clicked the submit button

Is the errors not correct?. There are 3 errors, first input must be >= -1000 and must be 2 decimals max. Second input is mandatory, so 3 errors in total is correct?. Unless you meant something different.

Yuyuutsu:
> > > On the test numbers schema it is adding more errors to the error list than there are inputs when I enter a number that is too small. Numbers seem to be limited to 2 decimal places whereas I thought we would support 6?
> > 
> > 
> > I dont seem to be getting that issue. The test currently limits the decimals to 2 but runner has already supported 6 decimals. Same with Katies comments above, I can update it to use 6 but it wont show 6 decimals on the screen it only displays up to 3 decimal points. Up to you guys if I should change it to 6, happy to change it 👍 
> 
> In that case leave it, It could make the functional tests confusing. As for the error I just tried it again and same thing happened, this is how I created it:
> 
> * I spun up your branch
> * loaded the `test_numbers` schema
> * entered "-99999999999999.999999" in the first input
> * left the second input blank
> * clicked the submit button

Just had a look, I think that is what is meant to happen, same thing occurs on the Main Branch 👍 
image

schemas/test/en/test_currency.json
katie-gardner: Ahhh I see - so the range still works correctly but doesn't display properly. I wonder if that'd cause confusion for someone later if they don't remember this, might be best to stick to 3 for the range boundaries? though I'd still be in favour of having a field somewhere on the test that allows 6dp

rmccar:
> > > > On the test numbers schema it is adding more errors to the error list than there are inputs when I enter a number that is too small. Numbers seem to be limited to 2 decimal places whereas I thought we would support 6?
> > > 
> > > 
> > > I dont seem to be getting that issue. The test currently limits the decimals to 2 but runner has already supported 6 decimals. Same with Katies comments above, I can update it to use 6 but it wont show 6 decimals on the screen it only displays up to 3 decimal points. Up to you guys if I should change it to 6, happy to change it 👍 
> > 
> > 
> > In that case leave it, It could make the functional tests confusing. As for the error I just tried it again and same thing happened, this is how I created it:
> > 
> > * I spun up your branch
> > * loaded the `test_numbers` schema
> > * entered "-99999999999999.999999" in the first input
> > * left the second input blank
> > * clicked the submit button
> 
> Is the errors not correct?. There are 3 errors, first input must be >= -1000 and must be 2 decimals max. Second input is mandatory, so 3 errors in total is correct?. Unless you meant something different.

They are correct, I was just under the impression that the design was to have one error per output. I could be wrong though

tests/app/test_jinja_filters.py
katie-gardner: thanks Yuyu, just looking at the tests here, I think it'd be good to have the extremes added to `test_get_width_for_number`

rmccar (approved):

kylelawsonAND (approved):
Working as expected 👍

tests/app/test_jinja_filters.py
katie-gardner: Sorry for making a meal of the above comment, moving it here - but looking closer at this test, I'm not sure if its working exactly as intended

in the test_numbers json the "value" for the minimum and maximum is provided as an "identifier" and "source" rather than a number, but the function `get_width_for_number` is doing

`min_value = answer.get("minimum", {}).get("value", 0)`
`min_value_width=len(str(min_value))` 

which is going to use the length of the object as a string rather than the value, does this need updating? and should allowable_widths also be updated given the new limits. That said I'm not entirely sure how this is being used

MebinAbraham: Re: `get_width_for_number`, good spot. But given that is an existing issue, I'll raise a card to look at that separately. Must have been an oversight when value sources for that were added.

Yuyuutsu: Added extremities 👍

tests/app/test_jinja_filters.py
```diff
@@ -237,12 +237,18 @@ def test_format_duration(duration, formatted_duration, app):
...
         ({"maximum": {"value": 1}}, 1),
         ({"maximum": {"value": 123456}}, 6),
         ({"maximum": {"value": 12345678901}}, 15),
         ({"minimum": {"value": -123456}, "maximum": {"value": 1234}}, 7),
         ({"decimal_places": 2, "maximum": {"value": 123456}}, 8),
+        ({"maximum": {"value": 999_999_999_999_999}}, 15),
+        ({"decimal_places": 5, "maximum": {"value": 999_999_999_999_999}}, 20),
+        ({"decimal_places": 6, "maximum": {"value": 999_999_999_999_999}}, 30),
+        ({"minimum": {"value": -99_999_999_999_999}}, 15),
+        ({"decimal_places": 5, "minimum": {"value": 999_999_999_999_999}}, 20),
```
katie-gardner, line 250: I think these last two were meant to be negative, currently they duplicate the two on lines 247 & 248, otherwise looks good

katie-gardner (changes requested):

schemas/test/en/test_currency.json
Yuyuutsu: Yeah after looking it is out of scope for this card, but there is no ticket yet. You can make a card if you would like on the trello board using the `bug template` on the far left and put it in `Tech Analysis` section. Otherwise I do not mind popping it on the trello board.

katie-gardner: done!

katie-gardner (approved):
Happy with the test updated for high precision numbers and rendering of those deferred to follow up. All working as expected otherwise 👍

PR #1090, opened 2023-04-14
Update cookie banner to be multilingual

rmccar:
Sets the language code on cookie banners and adds the babel extension to the babel config to prevent the cookie banner strings from being ignored by translations.
Also removes an import that isn't being used.

### How to review
- Clear your cookies, load test_language schema, switch to Welsh and see that the cookie banner buttons will change to Welsh.
- Test this in comparison to the `main` branch

### Checklist

### DIFF
--- a/templates/layouts/_base.html
+++ b/templates/layouts/_base.html
@@ -72,6 +72,7 @@
         "primaryButtonText": _('Accept all cookies'),
         "secondaryButtonText": _('Set cookie preferences'),
         "confirmationButtonText": _('Hide'),
+        'lang': language_code
       })
     }}
   {% endif %}
--- a/app/translations/messages.pot
+++ b/app/translations/messages.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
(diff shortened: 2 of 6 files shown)


rmccar:
[Update cookie banner config to be multilingual](https://trello.com/c/Styo1ls0/5669-update-cookie-banner-config-to-be-multilingual)

templates/layouts/_base.html
```diff
@@ -72,6 +72,7 @@
         "primaryButtonText": _('Accept all cookies'),
         "secondaryButtonText": _('Set cookie preferences'),
         "confirmationButtonText": _('Hide'),
+        'lang': language_code
```
berroar, line 75: Can we add a test to show that the cookie banner is now being translated?

Yuyuutsu (commented):
Buttons change to welsh as expected. Compared to Main. Just to double check, it is just the buttons that change?

rmccar:
> Buttons change to welsh as expected. Compared to Main. Just to double check, it is just the buttons that change?

Yeah because the translations haven't been done yet

Yuyuutsu (approved):

kylelawsonAND:
Would have asked same question as Yuyu so you've answered that 👍

kylelawsonAND (approved):

PR #1091, opened 2023-04-19
Update summaries to allow support for repeating sections

berroar:

Previously summary pages did not support list collectors/repeating sections, this issue was highlighted via a bug when users were attempting to view submitted responses for the `berd` schema, which contains repeating sections. The view submitted response page was only showing the first two sections prior to the list collector. I have updated the summary context so that summary pages are able to display answers repeating sections.

### How to review

Check if anything has been missed. Use the `test_view_submitted_response_repeating_sections` and `test_list_collector_section_summary` schema to check the change. 

Complete the questionnaire -> Go to the Thank You page -> Click the `save or print your answers` link -> Check the summary on the page is working.

Check the view submittted response page on the berd schema on this branch and compare with the main branch.

### Checklist


berroar:
[Bug: Issue Printing BERD Submission](https://trello.com/c/kyj3H3Fj/5653-bug-issue-printing-berd-submission)

templates/partials/summary/summary.html
```diff
@@ -4,6 +4,8 @@
   {% from "components/summary/_macro.njk" import onsSummary %}
 
   {% set summaryGroups = [] %}
+  {% set summaryType = content.summary.summary_type %}
```
petechd, line 7: I don't like that we set vars in camel case here when they are not DS elements, I mean all three. In other files we have things like:
- `{% set save_on_signout = true %}` in calculated summary html
- `{% set sign_out_url = content.sign_out_url %}`  and `{% set hide_sign_out_button = content.hide_sign_out_button %}` on thank you page
- `{% set page_title = _("Confirmation email sent") %}` on confirmation page

We should follow suit.

app/views/contexts/summary_context.py
```diff
@@ -4,48 +4,119 @@
...
+                    view_submitted_response=self.view_submitted_response,
+                ).get("groups", [])
+
+
+def set_unique_group_ids(groups: list[ImmutableDict]) -> list[ImmutableDict]:
+    checked_ids = set()
+    id_value = 0
+
+    for group in groups:
+        if "id" in group:
```
petechd, line 115: Groups always have id? from validator: `"required": ["id", "blocks"]`

app/views/contexts/summary_context.py
```diff
@@ -4,48 +4,119 @@
...
+            group_id = group["id"]
+            if group_id in checked_ids:
+                id_value += 1
+            checked_ids.add(group_id)
+            group["id"] = f"{group_id}-{id_value}"
+
+    return groups
```
petechd, line 122:
```suggestion

    for index, group in enumerate(groups):
        group_id = group["id"]
        group["id"] = f"{group_id}-{index}"

    return groups
```

app/views/contexts/summary/group.py
```diff
@@ -28,9 +28,15 @@ def __init__(
...
+        summary_type: str | None = None,
+        view_submitted_response: bool | None = False,
     ) -> None:
         self.id = group_schema["id"]
-        self.title = group_schema.get("title")
+
+        if summary_type == "Summary" and not group_schema.get("title"):
+            self.title = schema.get_title_for_section(location.section_id)
+        else:
+            self.title = group_schema.get("title")
```
petechd, line 39: What if it's not summary_type "Summary" and no group title (`summary_type != "Summary" and not group_schema.get("title")`? if it's calculated summary as `schemas/test/en/test_new_calculated_summary_repeating_section.json`?

app/views/contexts/summary_context.py
```diff
@@ -4,48 +4,119 @@
...
+                            current_location=location,
+                            routing_path=self._router.routing_path(
+                                section_id=section_id, list_item_id=item
+                            ),
+                        )
+
+                        yield from section_summary_context.build_summary(
+                            return_to=return_to,
+                            get_refactored_groups=False,
+                            summary_type="Summary",
```
petechd, line 85: This part is done twice so make it a method? `def build_item(list_item_id: str = None, list_name: str, section_id: str, return_to: str) -> None`? Then call it every time?

templates/partials/summary/summary.html
berroar: Updated this 👍

app/views/contexts/summary_context.py
berroar: Removed the if 👍

app/views/contexts/summary/group.py
berroar: I only added this as the Summaries looked slightly odd when displaying list collectors when there was no group title, as per the `berd` schema.  Calc summaries should stay the same

app/views/contexts/summary_context.py
berroar: Updated now 👍

Yuyuutsu (approved):
Functionally tested and compared with a staging branch of Runner. Was able to see all the questions I answered . Changes look good to me 👍

MebinAbraham (commented):
We should display each group in the context of the section title, we there is a section title.

katie-gardner:
Functionally tested and for test_list_collector_section_summary, I'm happy that all my answers now show up on the summary and printed answers, whilst they did not before. Likewise for the new test

Although having made the same change to test_list_collector (adding view_response = True to post_submission) the PDF of answers is empty, is this expected due to additional config that would get this schema to work? They do show when changing the show non item answers option but other surveys have a printed summary working without that so wasn't sure if it was necessary

app/views/contexts/summary_context.py
```diff
@@ -4,48 +4,112 @@
...
+            progress_store=self._progress_store,
+            metadata=self._metadata,
+            response_metadata=self._response_metadata,
+            current_location=location,
+            routing_path=self._router.routing_path(section_id),
+        )
+
+        return section_summary_context.build_summary(
+            return_to=return_to,
+            get_refactored_groups=False,
```
katie-gardner, line 98: not sure I totally follow the difference having this false is making

app/views/contexts/summary_context.py
```diff
@@ -4,48 +4,112 @@
...
+    id_value = 0
+
+    for group in groups:
+        group_id = group["id"]
+        if group_id in checked_ids:
+            id_value += 1
+        checked_ids.add(group_id)
+        group["id"] = f"{group_id}-{id_value}"
+
+    return groups
```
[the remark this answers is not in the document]
berroar: I needed to add something so the group ids were unique, adding a number rather than a list-item-id followed the same pattern we had in the section summary context and made it easy for the functional tests, as they use the same methods as section summary.

app/views/contexts/section_summary_context.py
```diff
@@ -91,7 +91,13 @@ def get_page_title(self, title_for_location: Union[Mapping, str]) -> str:
             page_title = page_title.format(list_item_position=list_item_position)
         return page_title
 
-    def _build_summary(self, return_to: Optional[str]) -> dict[str, Any]:
+    def build_summary(
+        self,
+        return_to: Optional[str],
+        get_refactored_groups: bool = True,
```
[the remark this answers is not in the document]
berroar: I add this so the Summary wouldn't call the `get_refactored_groups` method in the section summary context, as the `non refactored` groups appeared to work. Have reverted this now though.

app/views/contexts/summary_context.py
berroar: See response to Mebin above, I have reverted this now 👍

app/views/contexts/summary_context.py
```diff
@@ -4,48 +4,112 @@
...
+                    section_id=section_id, return_to=return_to
+                )
+
+    def build_summary_item(
+        self,
+        section_id: str,
+        return_to: str | None,
+        list_name: str | None = None,
+        list_item_id: str | None = None,
+    ) -> Generator[ImmutableDict[str, Any], None, None]:
```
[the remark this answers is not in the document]
berroar: I have updated the type hints here 👍

app/views/contexts/summary/group.py
katie-gardner: same_title
Having entered "Person a", "Person b", and "Person c" in the household section, it'd be nice to see those names on the summary, rather than the group title for each one, as currently I lose which item corresponds to each person in the list. I don't know if thats possible to pipe those in (or if in typical use it doesn't matter) but I think it looks slightly odd in this particular test

app/views/contexts/summary_context.py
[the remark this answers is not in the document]
katie-gardner: Looks good, based off the recent conversations the [str, Any]s could be removed as well but there are others - so if better done in another debt ticket feel free to resolve this

katie-gardner (commented):
Functionally reviewed again and all is still working - my comments addressed 👍
I'm just not sure about the title on the PDF (see image in above comment) as the question titles are the names, but the PDF titles are all the group title, is that intended?

app/views/contexts/summary/group.py
berroar: It would be possible, but we'd need to allow group titles to support placeholders which they do not at the moment. This is what I was doing initially but I realised that the schema was failing validation, and it prompted me to raise this question is slack: https://surveydatacollection.slack.com/archives/CLR17KQKU/p1681906866743749

app/views/contexts/summary_context.py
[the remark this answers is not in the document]
MebinAbraham: Right okay makes sense, but given we have a method that changes the group ids, I suggest that is the single place we set group ids. It's confusing to me that we "refactor" the group and set unique group id (in the context of a single section) then we refactor again to set unique id (in the context of an entire survey). If we don't want to set the list item id to make testing easier, you could set the index of the list item id?. You have both the list store and location when calling `_get_refactored_groups`. This would mean that single methods handles refactors of groups including unique ids.

app/views/contexts/section_summary_context.py
```diff
@@ -91,7 +91,11 @@ def get_page_title(self, title_for_location: Union[Mapping, str]) -> str:
             page_title = page_title.format(list_item_position=list_item_position)
         return page_title
 
-    def _build_summary(self, return_to: Optional[str]) -> dict[str, Any]:
+    def build_summary(
+        self,
+        return_to: Optional[str],
```
kylelawsonAND, line 96: -> str | None

app/views/contexts/section_summary_context.py
```diff
@@ -91,7 +91,11 @@ def get_page_title(self, title_for_location: Union[Mapping, str]) -> str:
...
+        view_submitted_response: bool = False,
+    ) -> dict[str, Any]:
```
kylelawsonAND, line 98: dict[str, Any] -> dict

New type hinting rule

app/views/contexts/summary/group.py
```diff
@@ -77,6 +80,7 @@ def _build_blocks_and_links(
         progress_store: ProgressStore,
         language: str,
         return_to_block_id: Optional[str],
+        view_submitted_response: Optional[bool] = False,
```
kylelawsonAND, line 83: Optionals to | None

app/views/contexts/summary_context.py
```diff
@@ -1,51 +1,111 @@
...
 
 class SummaryContext(Context):
+    def __init__(
+        self,
+        language: str,
+        schema: QuestionnaireSchema,
+        answer_store: AnswerStore,
+        list_store: ListStore,
+        progress_store: ProgressStore,
+        metadata: Optional[MetadataProxy],
```
kylelawsonAND, line 20: Optional -> | None

app/views/contexts/summary_context.py
```diff
@@ -1,51 +1,111 @@
...
             "summary_type": "Summary",
+            "view_submitted_response": self.view_submitted_response,
         }
 
     def _build_all_groups(
         self, return_to: Optional[str]
-    ) -> Generator[ImmutableDict[str, Any], None, None]:
-        """NB: Does not support repeating sections"""
-
+    ) -> Generator[dict[str, Any], None, None]:
```
kylelawsonAND, line 54: dict[str, Any] -> dict

app/views/contexts/summary_context.py
```diff
@@ -1,51 +1,111 @@
...
+                    section_id=section_id, return_to=return_to
+                )
+
+    def build_summary_item(
+        self,
+        section_id: str,
+        return_to: str | None,
+        list_name: str | None = None,
+        list_item_id: str | None = None,
+    ) -> list[dict[str, Any]]:
```
kylelawsonAND, line 78: dict[str, Any] -> dict

kylelawsonAND (changes requested):

app/views/contexts/section_summary_context.py
berroar: Updated 👍

app/views/contexts/section_summary_context.py
berroar: Updated 👍

app/views/contexts/summary/group.py
berroar: Updated 👍

app/views/contexts/summary_context.py
berroar: Updated 👍

app/views/contexts/summary_context.py
berroar: Updated 👍

app/views/contexts/summary_context.py
berroar: Updated 👍

app/views/contexts/summary_context.py
```diff
@@ -91,10 +91,23 @@ def build_summary_item(
...
+            title_for_location = section_summary_context.title_for_location()
+            title = (
+                self._placeholder_renderer.render_placeholder(
+                    title_for_location, list_item_id
+                )
+                if isinstance(title_for_location, dict)
+                else title_for_location
+            )
+            for group in groups:
+                group["title"] = title
```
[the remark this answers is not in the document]
berroar: I've updated now so that we are using section titles 👍

kylelawsonAND (approved):
All re-tested, looks and feels right 👍

katie-gardner:
Re-tested and happy with how the answers are now looking when the sections are created with placeholders, all looking good and rendering as expected

only part that now looks a bit odd to me is for lists inside a section with a static name due to the duplicate titles - this comparison of answering questions vs the summary looks a bit off
Questions
Summary

 I think thats what Mebin was referring to above as well. Is it possible to only have section title for the first of these?

berroar:
> Re-tested and happy with how the answers are now looking when the sections are created with placeholders, all looking good and rendering as expected
> 
> only part that now looks a bit odd to me is for lists inside a section with a static name due to the duplicate titles - this comparison of answering questions vs the summary looks a bit off Questions Summary
> 
> I think thats what Mebin was referring to above as well. Is it possible to only have section title for the first of these?

Yeah this should be updated now 👍

katie-gardner (approved):
Re-reviewed with the most recent changes to section titles and looks much better, all working well from my functional tests, with both placeholder sections and static ones 👍

app/views/contexts/summary_context.py
```diff
@@ -1,51 +1,133 @@
...
+                item_title = (
+                    self._placeholder_renderer.render_placeholder(
+                        title_for_location, list_item_id
+                    )
+                    if isinstance(title_for_location, dict)
+                    else title_for_location
+                )
+                group["section_title"] = item_title
+
+        return groups
```
MebinAbraham, line 112: This is odd as we are setting the section title, then unsetting it. You don't need to set the title on the group, you already have the title calling `section_summary_context()`. 

We could clean this up if we refactored to use sections of groups instead of a flattened list of group.

For now, instead of setting the title then unsetting it, just set it for the first one.


```py
        summary = section_summary_context(
            view_submitted_response=self.view_submitted_response, return_to=return_to
        )["summary"]
        groups: list = summary.get("groups", [])

        for group in groups:
            group["section_title"] = summary["title"]
            break

        return groups
```

app/views/contexts/summary_context.py
```diff
@@ -1,51 +1,133 @@
...
+
+    for group in groups:
+        group_id = group["id"]
+        section_title = group["section_title"]
+        if group_id in checked_ids:
+            id_value += 1
+            group["id"] = f"{group_id}-{id_value}"
+
+        if section_title in checked_titles:
+            group["section_title"] = ""
```
MebinAbraham, line 128: The section title in this will need reverting when the above is addressed.

app/views/contexts/summary_context.py
berroar: Updatted thos 👍

app/views/contexts/summary_context.py
```diff
@@ -76,6 +76,8 @@ def build_summary_item(
         list_name: str | None = None,
         list_item_id: str | None = None,
     ) -> list[dict]:
+        final_groups = []
```
MebinAbraham: This is unused.

templates/partials/summary/summary.html
```diff
@@ -3,23 +3,28 @@
...
+  {% set summary_groups = [] %}
+  {% set summary_types = content.summary.summary_type %}
+  {% set view_submitted_response = content.summary.view_submitted_response %}
 
   {%- for group in content.summary.groups if group.blocks -%}
-    {% do summaryGroups.append
+    {% do summary_groups.append
       (
         {
+          "summaryTitle": group.section_title,
```
MebinAbraham, line 14: At the moment the rendering is not correct when there is no group title. On the submitted response, when you don't have a section title, the group title are h2, otherwise it is h3, which is odd. Check with Rich is there is a way around it. Setting to an empty string won't work either.

The padding is also off as it is expecting a title after each group. We might have to make these into sections of groups instead of a flatten list of group.
image

app/views/contexts/summary_context.py
berroar: Removed this 👍

tests/functional/spec/features/view_submitted_response/view_submitted_response.spec.js
```diff
@@ -43,3 +59,88 @@ describe("View Submitted Response", () => {
...
+    await expect(await $(ViewSubmittedResponseRepeatingPage.heading()).getText()).to.equal("Answers submitted for Apple");
+    await expect(await $(ViewSubmittedResponseRepeatingPage.metadataTerm(1)).getText()).to.equal("Submitted on:");
+    await expect(await $(ViewSubmittedResponseRepeatingPage.metadataTerm(2)).getText()).to.equal("Submission reference:");
+    await expect(await $(ViewSubmittedResponseRepeatingPage.personalDetailsGroupTitle()).getText()).to.equal("Personal Details");
+    await expect(await $(ViewSubmittedResponseRepeatingPage.nameQuestion()).getText()).to.equal("What is your name?");
+    await expect(await $(ViewSubmittedResponseRepeatingPage.nameAnswer()).getText()).to.equal("John Smith");
+    await expect(await $(ViewSubmittedResponseRepeatingPage.addressDetailsGroupTitle()).getText()).to.equal("Address Details");
+    await expect(await $(ViewSubmittedResponseRepeatingPage.addressQuestion()).getText()).to.equal("What is your address?");
+    await expect(await $(ViewSubmittedResponseRepeatingPage.addressAnswer()).getText()).to.equal("NP10 8XG");
+    await expect(await $("body").getHTML()).to.contain("Marcus Twin");
```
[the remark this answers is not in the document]
berroar: Probably me but couldn't get a selector like the others working 😐

MebinAbraham (approved):

MebinAbraham:
We need to get this in so they can progress testing. The rendering issue can be fixed in a separate PR.

templates/partials/summary/summary.html
berroar: As discussed this is an issue in our implementation which will be resolved in another PR

katie-gardner (commented):
Happy with the changes, simpler way of managing the group titles 👍 
Re-tested and functionally working as expected

Only thing that came up, was I tried the summary with test_list_collector.json and this has a block which adds items to an existing list, so on the summary page it takes both blocks and just shows the same list twice without anything to distinguish between them, I'm not sure when this would get used, but it looks like something that might need addressing in a follow up? I don't necessarily see it as a blocker to this though, so happy to approve if deemed a non-issue

berroar:
> Happy with the changes, simpler way of managing the group titles 👍 Re-tested and functionally working as expected
> 
> Only thing that came up, was I tried the summary with test_list_collector.json and this has a block which adds items to an existing list, so on the summary page it takes both blocks and just shows the same list twice without anything to distinguish between them, I'm not sure when this would get used, but it looks like something that might need addressing in a follow up? I don't necessarily see it as a blocker to this though, so happy to approve if deemed a non-issue

As discussed I will investigate this issue in the follow up PR 👍

app/views/contexts/summary/group.py
berroar: This issue has been resolved but the rendering issue will be fixed in a follow up PR.

PR #1094, opened 2023-04-25
Make placeholders to take into account answers on a path

Yuyuutsu:
Currently, placeholders do not take into account whether answers are on the path. In v3 we added a placeholder transform `first_non_empty_item` which requires us to check that the answers are on the path. 

This PR adds the ability to pipe answers from any previous section that are on the path.

##### Functional
To test this change you need to follow the steps on the [card](https://trello.com/c/QB5n5BGv/5407-answer-dependencies-placeholders-should-take-into-account-answers-on-the-path-3-sp)
- Test the `test_first_non_empty_item`, `test_first_non_empty_item_cross_section_dependencies`, `test_first_non_empty_item_repeating_sections` schema 
- Test on any MBS schema
- Check `test_answer_source_with_routing_path` and  `first_non_empty_spec.js` run as expected.
Have a look at the repeating sections test and schema to see if that was expected. I was not 100% sure on that one

##### Other

- Check if the changes make sense
- Can `get_placeholder_dependencies` be simplified ?
- There are two new type ignores for `block` and `section_id` for ` def _get_placeholder_section_dependencies`. Check if  it makes sense. I could have also added two if statements instead of the type ignore.
- Check if unit test and functional tests are good
### Checklist


Yuyuutsu:
[Answer dependencies: Placeholders should take into account answers on the path (3 SP)](https://trello.com/c/QB5n5BGv/5407-answer-dependencies-placeholders-should-take-into-account-answers-on-the-path-3-sp)

kylelawsonAND (changes requested):

app/questionnaire/questionnaire_schema.py
```diff
@@ -961,6 +961,20 @@ def get_item_anchor(self, section_id: str, list_name: str) -> Optional[str]:
...
+        for transform in mapping:
+            if transform["transform"] == transform_name:
+                for item in transform.get("arguments", []).get("items", []):
+                    if item.get("source") == "answers":
+                        block = self.get_block_for_answer_id(item["identifier"])
+                        section_id = self.get_section_id_for_block_id(block["id"])  # type: ignore
+                        if section_id not in section_ids:
+                            section_ids.append(section_id)
+
+        return section_ids
```
MebinAbraham, line 976: Couple of things:
1. You need to evaluate the dependencies up front and not at run time. This can be pre-determined at load time. It is wasteful to fetch this every time, for a given schema the dependency won't change.
2. This is fetching the dependency for the entire survey, it should be by "block_id" of the current location and if you haven't got a block_id (i.e summary), then by section. Have a look at how the calculated summary one is done in placeholder, it needs to be similar to that. At the moment, if section 2 depends on section 1, and section 4 depends on section 3. Section 2 needs section 1's path, section 4 needs section 3's path, but the current implementation would mean section 2 and 4 will have section 1 and section 3's path which is wasteful and inefficient.

app/questionnaire/value_source_resolver.py
```diff
@@ -104,7 +104,7 @@ def _resolve_answer_value_source(
         answer_id = value_source["identifier"]
 
         answer_value = self._get_answer_value(
-            answer_id=answer_id, list_item_id=list_item_id
+            answer_id=answer_id, list_item_id=list_item_id, assess_routing_path=True
```
MebinAbraham, line 107: This can't now just be True, should only be True for the cases it needs the path, i.e for placeholder, just the placeholders we care about.

app/questionnaire/placeholder_parser.py
```diff
@@ -193,3 +206,22 @@ def _get_routing_path_block_ids(
...
+        ):
+            return self._value_source_resolver
+
+        complete_routing_path_block_ids = []
+
+        for section_id in section_ids:
+            routing_path_block_ids = self._path_finder.routing_path(
+                section_id=section_id  # type: ignore
+                # section_id always exists
+            ).block_ids
```
MebinAbraham, line 224: 1. You need to use the `self._routing_path_block_ids_by_section_key` cache, as the routing path existing alredy.
2. This doesn't take into account repeating sections, so you need to handle that.

MebinAbraham (changes requested):

app/questionnaire/placeholder_parser.py
```diff
@@ -138,21 +140,31 @@ def _parse_transforms(
 
         for transform in transform_list:
             transform_args: MutableMapping[str, Any] = {}
```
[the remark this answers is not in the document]
Yuyuutsu: Updated 👍

app/questionnaire/questionnaire_schema.py
Yuyuutsu: Updated to evaluate dependencies on load time 👍

Yuyuutsu: Using block_id of current location now 👍

app/questionnaire/value_source_resolver.py
Yuyuutsu: Updated should only be true when its a placeholder transform we want

app/questionnaire/placeholder_parser.py
```diff
@@ -107,7 +110,9 @@ def __call__(
         return self._placeholder_map
 
     def _get_value_source_resolver(
-        self, routing_path_block_ids: OrderedSet[str] | None = None
+        self,
```
petechd, line 113: Since we touch this, can we force named arguments? so:
```
def _get_value_source_resolver(
        self,
        *,
        routing_path_block_ids: OrderedSet[str] | None = None,
        assess_routing_path: bool = False,
    ) -> ValueSourceResolver:
```
which will force you to
```
return self._get_value_source_resolver(
                routing_path_block_ids=OrderedSet(routing_path_block_ids), assess_routing_path=True
            )
```

app/questionnaire/placeholder_parser.py
```diff
@@ -142,19 +147,24 @@ def _parse_transforms(
 
         for transform in transform_list:
             transform_args: MutableMapping = {}
+            value_source_resolver = self._get_value_source_resolver_for_transform(
```
petechd, line 150: What is the purpose of TRANSFORMS_REQUIRING_ROUTING_PATH? We only want to initialise new VSR when certain transform is being processed?

tests/functional/spec/first_non_empty.spec.js
berroar, line 1: I'd consider renaming this file cause we're not really only testing first non empty placeholder here, testing that placeholders are on the path.

schemas/test/en/test_first_non_empty_item_cross_section_dependencies.json
```diff
@@ -0,0 +1,376 @@
+{
```
petechd, line 1: This would be better if it's hub type schema with first section required to see the hub, atm it is not clear when you enter second section until you are on the summary.

berroar (changes requested):

schemas/test/en/test_first_non_empty_item_cross_section_dependencies.json
```diff
@@ -0,0 +1,376 @@
...
+    "mime_type": "application/json/ons/eq",
```
petechd, line 2: And some descriptions/guidance on date range question since I don't know what range is allowed etc.

app/questionnaire/placeholder_parser.py
Yuyuutsu: Removed, was meant to be removed before just forgot . Should be evaluated at questionnaire schema 👍

app/questionnaire/dependencies.py
```diff
@@ -55,3 +55,35 @@ def get_block_ids_for_calculated_summary_dependencies(
             block_ids_by_section[key] = routing_path.block_ids
 
     return block_ids_by_section
+
+
+def get_block_ids_for_placeholder_dependencies(
```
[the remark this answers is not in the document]
Yuyuutsu: Method is now  generic and called `get_block_ids_for_dependencies` 👍

tests/functional/spec/first_non_empty.spec.js
Yuyuutsu: Renamed 👍

schemas/test/en/test_first_non_empty_item_cross_section_dependencies.json
Yuyuutsu: Changed to a Hub 👍

app/questionnaire/placeholder_parser.py
Yuyuutsu: Tried it got a lot of unit test errors will have another look tomorrow morning

app/questionnaire/questionnaire_schema.py
kylelawsonAND: Resolving on behalf of Mebin

app/questionnaire/value_source_resolver.py
kylelawsonAND: Resolving on behalf of Mebin.

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
Yuyuutsu: Updated 👍

schemas/test/en/test_first_non_empty_item_cross_section_dependencies.json
Yuyuutsu: I have also changed the schema a bit so there shouldnt be a total turnover question in the first section. Just asks for the date

app/questionnaire/placeholder_parser.py
```diff
@@ -81,11 +81,19 @@ def __call__(
         self, placeholder_list: Sequence[Mapping]
     ) -> MutableMapping[str, ValueSourceEscapedTypes | ValueSourceTypes]:
         placeholder_list = QuestionnaireSchema.get_mutable_deepcopy(placeholder_list)
```
petechd, line 83: @Yuyuutsu why mutable deepcopy? it doesn't look like your placeholder list is ever modified after?

tests/app/questionnaire/test_placeholder_parser.py
```diff
@@ -1137,3 +1137,99 @@ def test_placeholder_parser_calculated_summary_dependencies_cache(
     placeholder_2 = placeholder_parser(placeholder_list=placeholder_list_2)
     assert placeholder_2["unit-total-playback"] == 11
     assert path_finder.called == 1
+
+
+def test_placeholder_dependencies_cache(mocker, mock_renderer):
+    schema = load_schema_from_name("test_first_non_empty_item")
+    path_finder = mocker.patch("app.questionnaire.path_finder.PathFinder.routing_path")
```
petechd, line 1144: I think it is easy enough to initialise path finder here and just mock, or make up some of its inputs, like we do elsewhere:
```python
schema = load_schema_from_name("test_skip_condition_group")
    answer_store.add_or_update(
        Answer(answer_id="do-you-want-to-skip-answer", value="Yes")
    )

    # When
    path_finder = PathFinder(schema, answer_store, list_store, progress_store, None, {})
```
remember schema and answer store, list store can be only minimal objects.

app/questionnaire/dependencies.py
```diff
@@ -15,35 +15,42 @@
     from app.questionnaire.path_finder import PathFinder  # pragma: no cover
 
 
-def get_block_ids_for_calculated_summary_dependencies(
+def get_block_ids_for_dependencies(
```
petechd, line 18: Can we force named arguments here? It's too many of them to follow.

app/questionnaire/placeholder_parser.py
```diff
@@ -179,16 +189,20 @@ def _resolve_value_source_list(
         return values
 
     def _get_routing_path_block_ids(
-        self, data: Sequence[Mapping], sections_to_ignore: list | None = None
+        self,
```
petechd, line 192: Same here, force'em.

app/questionnaire/questionnaire_schema.py
```diff
@@ -1248,6 +1262,29 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
                 if item["for_list"] == list_name and item.get("item_anchor_answer_id"):
                     return f"#{str(item['item_anchor_answer_id'])}"
 
+    def _populate_placeholder_section_dependencies(self) -> None:
+        for section in self.get_sections():
```
petechd, line 1266: Can we not do one for loop and something like we did before:
```python
```
and whenever you need section (id) just call
```python
self.get_section_for_block_id(block_id)
```

app/questionnaire/placeholder_parser.py
Yuyuutsu: Good question. Ill have a look into this

Yuyuutsu: Had a look into it, not sure why it was here change was three years ago. Removed and working as expected 👍

app/questionnaire/dependencies.py
Yuyuutsu: Forced 👍

app/questionnaire/dependencies.py
```diff
@@ -15,35 +15,42 @@
...
+    ignore_keys: list[str] | None = None,
 ) -> dict[tuple, tuple[str, ...]]:
     block_ids_by_section: dict[tuple, tuple[str, ...]] = {}
-
     sections_to_ignore = sections_to_ignore or []
-    dependent_sections = schema.calculated_summary_section_dependencies_by_block[
-        location.section_id
-    ]
+    dependent_sections: dict[str, set[str]] | dict[str, OrderedSet[str]]
+    if source_type == "calculated_summary":
```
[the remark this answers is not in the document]
Yuyuutsu: Yeah that sounds way better. Updated 👍

app/questionnaire/placeholder_parser.py
```diff
@@ -30,6 +28,8 @@
 
 TransformedValueTypes: TypeAlias = None | str | int | Decimal | bool
 
+TRANSFORMS_REQUIRING_ROUTING_PATH = ["first_non_empty_item"]
```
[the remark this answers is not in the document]
Yuyuutsu: Import added 👍

app/questionnaire/placeholder_parser.py
Yuyuutsu: Forced 👍

app/questionnaire/placeholder_parser.py
Yuyuutsu: Update: Forced 👍

app/questionnaire/questionnaire_schema.py
Yuyuutsu: Updated using `self.get_blocks() `🖖

tests/app/questionnaire/test_placeholder_parser.py
Yuyuutsu: Added 👍  The path finder check is also different now

app/questionnaire/placeholder_parser.py
```diff
@@ -80,12 +79,19 @@ def __init__(
     def __call__(
         self, placeholder_list: Sequence[Mapping]
     ) -> MutableMapping[str, ValueSourceEscapedTypes | ValueSourceTypes]:
-        placeholder_list = QuestionnaireSchema.get_mutable_deepcopy(placeholder_list)
-
         sections_to_ignore = list(self._routing_path_block_ids_by_section_key)
+        source_type = "calculated_summary"
```
berroar, line 83: I think setting source_type as calculated_summary for everything is a bit confusing, rather than using source_type, could potentially use the assess_routing_path bool? As if that is False we know to get the placeholder dependencies in the `get_routing_path_block_ids` method?

Yuyuutsu: Updated to use `assess_routing_path` instead of `source_type` here.

app/questionnaire/placeholder_parser.py
Yuyuutsu: Forgot to reply to this comment. Way we get the value source resolver is now different and should follow how we did it with calculated summaries. Also now supports repeating sections.

schemas/test/en/test_first_non_empty_item_cross_section_dependencies.json
kylelawsonAND, line 1: I wonder if it's worth marking the dependant section incomplete when a placeholder dependency changes? Not sure if this is out of scope/wouldn't be needed in a real survey.

In this Schema for example if I complete section 2 where the placeholder is the default one, I can then complete section 1, changing the dates to something else, but the second section is still marked completed.

Yuyuutsu: Ooo good point i am not too sure. I wonder what the other devs think about this?

Yuyuutsu: Update after Tech session. Decided not to make it incomplete, since in a bigger schema we would be asking the user to re input all information again. Which is inconvenient for them.

app/questionnaire/placeholder_parser.py
```diff
@@ -179,18 +187,39 @@ def _resolve_value_source_list(
...
+            ]
+            source_type = "answers"
+        else:
+            dependent_sections = (
+                self._schema.calculated_summary_section_dependencies_by_block[
+                    self._location.section_id
+                ]
             )
-        return {}
+            source_type = "calculated_summary"
```
MebinAbraham, line 211: Why do we have `assess_routing_path` here?. Currently it could be true if any of the placeholders in the list has a transform that needs routing path. But that list could also contain a calculated summary placeholder which would then break?. As it would be True, for a calc summary,

app/questionnaire/placeholder_parser.py
```diff
@@ -95,7 +99,8 @@ def __call__(
                 routing_path_block_ids_map
             )
             self._value_source_resolver = self._get_value_source_resolver(
-                routing_path_block_ids
+                routing_path_block_ids=routing_path_block_ids,
+                assess_routing_path=assess_routing_path,
```
MebinAbraham, line 103: This is setting assess true for the main resolver that is used by multiple transforms even the ones that dont use the transform we care about and the ones that shouldn't use the path. You need to get a value source resolver based on the transform at the point of resolution. It works for calc summary because, calc summary here sets assess false and in Value source resolver, it will set to True when needed.

For yours, you need to create a new instance of VSE with assess True when it is a transform that require path. You need to do it at the `for transform in transform_list:` level.

app/questionnaire/placeholder_parser.py
Yuyuutsu: Removed 👍

app/questionnaire/placeholder_parser.py
Yuyuutsu: Made some changes. Is this what you were meaning?

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +171,70 @@ def _parse_transforms(
...
-                data=data,
                 path_finder=self._path_finder,
+                data=transform,
+                source_type="answers",
+                dependent_sections=dependent_sections,
+            )
+            self._routing_path_block_ids_by_section_key.update(block_ids)
+            routing_path_block_ids: list = [
+                value for values in block_ids.values() for value in values
+            ]
```
MebinAbraham, line 234: All of this is redundant if we don't have dependent sections, so I suggest making it conditional.

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +171,70 @@ def _parse_transforms(
...
+            block_ids = get_block_ids_for_dependencies(
                 location=self._location,
                 progress_store=self._progress_store,
-                sections_to_ignore=sections_to_ignore,
-                data=data,
                 path_finder=self._path_finder,
+                data=transform,
+                source_type="answers",
+                dependent_sections=dependent_sections,
+            )
```
MebinAbraham, line 230: This isn't passing the `section_to_ignore` therefore cache isn't being used.

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +171,70 @@ def _parse_transforms(
...
+        )
+
+        return get_block_ids_for_dependencies(
+            location=self._location,
+            progress_store=self._progress_store,
+            sections_to_ignore=sections_to_ignore,
+            data=data,
+            path_finder=self._path_finder,
+            source_type="calculated_summary",
+            ignore_keys=None,
```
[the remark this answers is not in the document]
Yuyuutsu: Added 👍

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +171,70 @@ def _parse_transforms(
...
-                sections_to_ignore=sections_to_ignore,
-                data=data,
                 path_finder=self._path_finder,
+                data=transform,
+                source_type="answers",
+                dependent_sections=dependent_sections,
+            )
+            self._routing_path_block_ids_by_section_key.update(block_ids)
+            routing_path_block_ids: list = [
+                value for values in block_ids.values() for value in values
```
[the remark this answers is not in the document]
Yuyuutsu: Updated 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -1248,6 +1262,25 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
...
+                for transform in transforms
+                if transform["transform"] in TRANSFORMS_REQUIRING_ROUTING_PATH
+                for item in transform["arguments"]["items"]
+                if item.get("source") == "answers"
+            ]
+            placeholder_dependencies = self._get_placeholder_section_dependencies(
+                placeholder_answer_ids
+            )
+
+            if section := self.get_section_for_block_id(block["id"]):
```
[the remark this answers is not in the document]
Yuyuutsu: Added

schemas/test/en/test_first_non_empty_item.json
```diff
@@ -0,0 +1,283 @@
+{
```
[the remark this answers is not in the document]
Yuyuutsu: Name updated for all three

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +171,70 @@ def _parse_transforms(
...
+        return get_block_ids_for_dependencies(
+            location=self._location,
+            progress_store=self._progress_store,
+            sections_to_ignore=sections_to_ignore,
+            data=data,
+            path_finder=self._path_finder,
+            source_type="calculated_summary",
+            ignore_keys=None,
+            dependent_sections=dependent_sections,
+        )
```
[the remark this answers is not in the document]
Yuyuutsu: Added a function in dependencies.py

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +171,70 @@ def _parse_transforms(
...
+                data=transform,
+                source_type="answers",
+                dependent_sections=dependent_sections,
+            )
+            self._routing_path_block_ids_by_section_key.update(block_ids)
+            routing_path_block_ids: list = [
+                value for values in block_ids.values() for value in values
+            ]
+            return self._get_value_source_resolver(
+                routing_path_block_ids=OrderedSet(routing_path_block_ids),
```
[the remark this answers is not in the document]
Yuyuutsu: Removed Ordered set call and now it should use the routing block ids form the main VSR. Not too sure if I did it right

tests/app/questionnaire/test_placeholder_parser.py
```diff
@@ -1137,3 +1139,103 @@ def test_placeholder_parser_calculated_summary_dependencies_cache(
...
+    expected_routing_path = RoutingPath(
+        ["date-question-block", "date-entry-block", "total-turnover-block"],
+        section_id="default-section",
+    )
+
+    placeholder = placeholder_parser(placeholder_list=placeholder_list)
+
+    assert placeholder["date_entry_answer_from"] == "16 April 2016"
+    assert placeholder["date_entry_answer_to"] == "28 April 2016"
+    assert routing_path == expected_routing_path
```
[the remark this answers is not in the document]
Yuyuutsu: Test reverted

app/questionnaire/questionnaire_schema.py
```diff
@@ -31,6 +31,8 @@
 
 DependencyDictType: TypeAlias = dict[str, OrderedSet[str]]
 
+TRANSFORMS_REQUIRING_ROUTING_PATH = ["first_non_empty_item"]
```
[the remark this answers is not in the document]
Yuyuutsu: Changed to a set 👍

app/questionnaire/dependencies.py
```diff
@@ -55,3 +52,28 @@ def get_block_ids_for_calculated_summary_dependencies(
             block_ids_by_section[key] = routing_path.block_ids
 
     return block_ids_by_section
+
+
+def get_calculated_summary_block_ids_of_dependent_section(
```
katie-gardner, line 57: similar comment I think this doesn't indicate that its returning a map and a name adjustment would be helpful

katie-gardner (changes requested):
Had a read through and functionally tested and all works as expected! 👍 just a few minor thoughts

app/questionnaire/dependencies.py
```diff
@@ -15,35 +15,32 @@
...
+    dependents = (
+        OrderedSet(dependent_sections.get(location.block_id, []))
+        if location.block_id
+        else get_flattened_mapping_values(dependent_sections)
+    )
 
     if dependents and not get_sources_for_type_from_data(
-        source_type="calculated_summary",
-        data=data,
-        ignore_keys=["when"],
```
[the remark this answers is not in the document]
Yuyuutsu: It should pass in ["when"] when the function is called now for calculated summary. Or should we always call ["when"]

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +174,70 @@ def _parse_transforms(
...
         for value_source in value_source_list:
-            value = self._value_source_resolver.resolve(value_source)
+            value = value_source_resolver.resolve(value_source)
             if isinstance(value, list):
                 values.extend(value)
             else:
                 values.append(value)
         return values
 
     def _get_routing_path_block_ids(
```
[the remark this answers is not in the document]
Yuyuutsu: Updated 👍

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +174,70 @@ def _parse_transforms(
 
         return transformed_value
 
+    @staticmethod
     def _resolve_value_source_list(
```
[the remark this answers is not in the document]
Yuyuutsu: Happy to refactor. What do other devs think about this?

app/questionnaire/dependencies.py
```diff
@@ -15,35 +15,32 @@
     from app.questionnaire.path_finder import PathFinder  # pragma: no cover
 
 
-def get_block_ids_for_calculated_summary_dependencies(
-    schema: QuestionnaireSchema,
+def get_block_ids_for_dependencies(
```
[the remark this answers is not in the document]
Yuyuutsu: Updated to `map_calculated_summary_block_ids_to_routing_path`.

app/questionnaire/dependencies.py
Yuyuutsu: Updated to `map_section_ids_to_routing_path`.

app/questionnaire/questionnaire_schema.py
```diff
@@ -1248,6 +1268,28 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
                 if item["for_list"] == list_name and item.get("item_anchor_answer_id"):
                     return f"#{str(item['item_anchor_answer_id'])}"
 
+    def _populate_placeholder_section_dependencies(self) -> None:
+        for block in self.get_blocks():
+            transforms = get_mappings_with_key("transform", block)
+            placeholder_answer_ids = [
```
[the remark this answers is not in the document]
Yuyuutsu: Updated to a set 👍

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
Yuyuutsu: Sorry I am a bit confused 😅, which variable/function you would like to rename?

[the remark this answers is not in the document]
Yuyuutsu: Updated to _get_routing_path_block_ids_and_section_key_map

app/questionnaire/placeholder_parser.py
Yuyuutsu: If Condition has been updated

app/questionnaire/questionnaire_schema.py
```diff
@@ -1220,6 +1225,15 @@ def _get_calculated_summary_section_dependencies(
...
         return section_dependencies
 
+    def _get_placeholder_section_dependencies(self, answer_ids: list) -> set[str]:
+        section_dependencies: set[str] = set()
+        for answer_id in answer_ids:
+            block = self.get_block_for_answer_id(answer_id)
+            # Type ignore: block_id and section_id is never None
+            section_id = self.get_section_id_for_block_id(block["id"])  # type: ignore
+            section_dependencies.add(section_id)  # type: ignore
+        return section_dependencies
```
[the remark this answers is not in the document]
Yuyuutsu: updated 👍

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
Yuyuutsu: Updated: OrderedSet is used again because VSR routing path block ids is an Iterable[STR] and routing path ids in get_value_source_resolver is typed as an Ordered Set

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
MebinAbraham: I think it was just where it was initially introduced as only placeholders would be given a list, but that is not true longer term and VSR seems the appropriate place to resolve values so 👍

app/questionnaire/dependencies.py
```diff
@@ -15,35 +15,32 @@
     from app.questionnaire.path_finder import PathFinder  # pragma: no cover
 
 
-def get_block_ids_for_calculated_summary_dependencies(
-    schema: QuestionnaireSchema,
+def map_section_ids_to_routing_path(
```
MebinAbraham, line 18: The naming implies it either doesn't return anything and that it mutates something but it doesn't mutate but also returns. The name some reflect the body, so dependencies are an important part. I suggest something like `get_routing_path_block_ids_by_section_for_dependent_sections`. Open to other suggestions.

app/questionnaire/dependencies.py
```diff
@@ -55,3 +52,28 @@ def get_block_ids_for_calculated_summary_dependencies(
             block_ids_by_section[key] = routing_path.block_ids
 
     return block_ids_by_section
+
+
+def map_calculated_summary_block_ids_to_routing_path(
```
MebinAbraham, line 57: Similar to the above, this should be something like `get_routing_path_block_ids_by_section_for_calculated_summary_dependencies`. I don't think `block_ids` are necessary as if you iterate over a routing path, you would iterate over the block ids, so could also be `get_routing_paths_by_section_for_calculated_summary_dependencies`. Happy to keep block_ids bit for clarity. See what others think.

app/questionnaire/placeholder_parser.py
```diff
@@ -107,7 +107,10 @@ def __call__(
         return self._placeholder_map
 
     def _get_value_source_resolver(
-        self, routing_path_block_ids: OrderedSet[str] | None = None
+        self,
+        *,
+        routing_path_block_ids: OrderedSet[str] | None = None,
+        assess_routing_path: bool = False,
```
MebinAbraham, line 113: Existing issue, but why is the default False?. Should it not be `None`?. False has a different meaning to None when used in VSR.

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
MebinAbraham: That type should just be `Iterable[str]`

app/questionnaire/questionnaire_schema.py
```diff
@@ -1248,6 +1268,28 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
...
+                if item.get("source") == "answers"
+            }
+            placeholder_dependencies = self._get_section_ids_for_answer_ids(
+                answer_ids=placeholder_answer_ids
+            )
+            if placeholder_dependencies:
+                # Type Ignore: At this point we section id cannot be None
+                section_id = self.get_section_id_for_block_id(block["id"])
+                self._placeholder_section_dependencies_by_block[section_id][  # type: ignore
+                    block["id"]
```
MebinAbraham, line 1288: Insted of type ignore this, type ignore the source of the issue, which is the line above.

tests/app/questionnaire/test_placeholder_parser.py
```diff
@@ -1137,3 +1137,99 @@ def test_placeholder_parser_calculated_summary_dependencies_cache(
     placeholder_2 = placeholder_parser(placeholder_list=placeholder_list_2)
     assert placeholder_2["unit-total-playback"] == 11
     assert path_finder.called == 1
+
+
+def test_placeholder_dependencies_cache(mocker, mock_renderer):
```
MebinAbraham, line 1142: Worth adding some doc strings to explain what exactly this is testing. Would be good to add to the existing one also.

MebinAbraham (approved):
Looking good

app/questionnaire/dependencies.py
rmccar: Happy to go with `get_routing_path_block_ids_by_section_for_calculated_summary_dependencies` personally I like the clarity

rmccar (approved):
Other than the outstanding comments looks good

app/questionnaire/dependencies.py
katie-gardner: likewise 👍

app/questionnaire/dependencies.py
Yuyuutsu: New name suggestion sounds clearer to me - Updated to  `get_routing_path_block_ids_by_section_for_dependent_sections`

app/questionnaire/dependencies.py
Yuyuutsu: Updated to `get_routing_path_block_ids_by_section_for_calculated_summary_dependencies`

app/questionnaire/placeholder_parser.py
Yuyuutsu: Updated to default to be None 👍 and updated type hinting

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
Yuyuutsu: Updated 👍

app/questionnaire/questionnaire_schema.py
Yuyuutsu: Does not seem to work when type ignoring section_id at the line above

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
Yuyuutsu: Moved to Value Source Resolver 🎉

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +174,72 @@ def _parse_transforms(
...
+            value = value_source_resolver.resolve(value_source)
             if isinstance(value, list):
                 values.extend(value)
             else:
                 values.append(value)
         return values
 
-    def _get_routing_path_block_ids(
-        self, data: Sequence[Mapping], sections_to_ignore: list | None = None
+    def _get_routing_path_block_ids_and_section_key_map(
```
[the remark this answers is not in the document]
Yuyuutsu: Updated to `_get_routing_path_block_ids_by_section_for_calculated_summary_dependencies`. But happy to do the latter if other people have strong views 👍

app/forms/questionnaire_form.py
```diff
@@ -465,21 +465,23 @@ def get_answer_fields(
...
     if location and progress_store:
-        routing_path_block_ids = get_block_ids_for_calculated_summary_dependencies(
-            schema=schema,
-            location=location,
-            progress_store=progress_store,
-            path_finder=PathFinder(
-                schema=schema,
-                answer_store=answer_store,
-                list_store=list_store,
+        routing_path_block_ids = (
```
MebinAbraham, line 468: This will store a map, so maybe name it `block_ids_by_section`. I don't necessary think it needs to be prefixed here with `routing_path` is there is enough context in the method.

app/questionnaire/dependencies.py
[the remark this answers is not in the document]
MebinAbraham: Only if needed, calling from placeholder would need ignore keys but this method is generic so the current change is appropriate.

app/questionnaire/placeholder_parser.py
MebinAbraham: This default was correct actually. By default, placeholder don't need to assess the path, so False is valid.

tests/app/questionnaire/test_placeholder_parser.py
MebinAbraham: I was expecting a comment around the technical details as I don't think what is there currently adds much more than the method name and asserts?. I was expecting an explanation of why this test should work and how. Such as the first placeholder list uses transform x which is one of the one that require the path. The second list also uses the same transform and answers from the same section as the first call. Therefore, we are checking, on the second call the cache is used which was created when the first call was made. Something like this, need cleaning up but you get the idea. It will make it easier to understand when coming back to this, as a lot of going on and it isn't obvious why the test is doing the things it does.

app/questionnaire/dependencies.py
```diff
@@ -15,35 +15,32 @@
...
     if dependents and not get_sources_for_type_from_data(
-        source_type="calculated_summary",
-        data=data,
-        ignore_keys=["when"],
+        source_type=source_type, data=data, ignore_keys=ignore_keys
     ):
         return block_ids_by_section
 
     for section in dependents:
-        # Dependent sections other than the current section cannot be a repeating section
```
MebinAbraham, line 46: How come this was removed?.

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,6 +98,13 @@ def __init__(
         self._populate_answer_dependencies()
         self._populate_when_rules_section_dependencies()
         self._populate_calculated_summary_section_dependencies()
+        self._populate_placeholder_section_dependencies()
+
+    @property
+    def placeholder_section_dependencies_by_block(
+        self,
+    ) -> dict[str, dict[str, set[str]]]:
+        return self._placeholder_section_dependencies_by_block
```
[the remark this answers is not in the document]
Yuyuutsu: Changed to `placeholder_transform_section_dependencies_by_block`

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +179,61 @@ def _parse_transforms(
...
-            else:
-                values.append(value)
-        return values
-
-    def _get_routing_path_block_ids(
-        self, data: Sequence[Mapping], sections_to_ignore: list | None = None
+    def _get_routing_path_block_ids_by_section_for_calculated_summary_dependencies(
+        self,
+        data: Sequence[Mapping],
+        sections_to_ignore: list | None = None,
```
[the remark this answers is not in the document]
Yuyuutsu: Added 👍

app/questionnaire/dependencies.py
Yuyuutsu: Reverted 👍

app/questionnaire/value_source_resolver.py
```diff
@@ -33,7 +33,7 @@ class ValueSourceResolver:
     routing_path_block_ids: Iterable[str] | None = None
     use_default_answer: bool = False
     escape_answer_values: bool = False
-    assess_routing_path: bool = True
+    assess_routing_path: bool | None = True
```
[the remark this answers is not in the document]
Yuyuutsu: Reverted 👍

app/forms/questionnaire_form.py
Yuyuutsu: Updated 👍

Yuyuutsu: Sorry a bit confused, where is the routing_path prefix?

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +179,61 @@ def _parse_transforms(
...
     ) -> dict[tuple, tuple[str, ...]] | None:
```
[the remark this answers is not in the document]
Yuyuutsu: We have a `SectionKeyType = tuple[str, Optional[str]]` already in progress store. Should I update it to `SectionKey` and put that in `types.py`?

app/questionnaire/questionnaire_schema.py
```diff
@@ -1248,6 +1268,28 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
                 if item["for_list"] == list_name and item.get("item_anchor_answer_id"):
                     return f"#{str(item['item_anchor_answer_id'])}"
 
+    def _populate_placeholder_section_dependencies(self) -> None:
+        for block in self.get_blocks():
+            transforms = get_mappings_with_key("transform", block)
+            placeholder_answer_ids = {
+                item.get("identifier")
```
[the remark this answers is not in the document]
Yuyuutsu: Updated to `item["identifier"]`

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
MebinAbraham: See if people like the idea of a named tuple to give it context. I don't mind either.

app/questionnaire/value_source_resolver.py
```diff
@@ -217,6 +218,16 @@ def _resolve_location_source(self, value_source: Mapping) -> str | None:
     def _resolve_response_metadata_source(self, value_source: Mapping) -> str | None:
         return self.response_metadata.get(value_source.get("identifier"))
 
+    def resolve_list(self, value_source_list: list[dict]) -> list[ValueSourceTypes]:
```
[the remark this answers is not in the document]
katie-gardner: ah so recursive, if list, call itself for each item?
the only issue I could imagine is if there were lots of value sources in the transform arguments list all the isinstance checks might not be as fast but I suspect its fairly negligible 

I don't really mind, though since its a different arg type and different outcome, and easy to leave separate, I quite like it as it is

app/questionnaire/placeholder_parser.py
```diff
@@ -166,32 +179,61 @@ def _parse_transforms(
...
         return all(source == "metadata" for source in sources)
+
+    def _get_value_source_resolver_for_transform(
+        self, transform: Mapping
+    ) -> ValueSourceResolver:
+        if self._location and (
+            dependent_sections := self._schema.placeholder_section_dependencies_by_block[
+                self._location.section_id
+            ]
+        ):
```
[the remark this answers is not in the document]
Yuyuutsu: Updated, to check the transform is one that requires a routing path. Then if it does, it will get the dependent sections.

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
Yuyuutsu: Updated it so SectionKeyType is renamed to SectionKey and updated to = `tuple[str, str | None]` and is stored in the types.py file. Happy to revert if others think differently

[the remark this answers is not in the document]
MebinAbraham: Go with namedtuple Yuyu, which won't be a type alias, but you can use it as a type. I don't think it is strictly wrong to add that into `types.py` even if it isn't still an alias. So you would have:

```py
SectionKey = namedtuple("SectionKey", "section_id list_item_id")
```

Then you can use it like a tuple or used the named keys:
```py
section_key: SectionKey = SectionKey("section-1", "abEW")

section_id = section_key[0]

# or the clearer option

section_id = section_key.section_id
```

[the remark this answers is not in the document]
Yuyuutsu: Updated to namedtuple in `types.py`. Made a few changes, hope they all make sense 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -1220,6 +1231,15 @@ def _get_calculated_summary_section_dependencies(
...
         return section_dependencies
 
+    def _get_section_ids_for_answer_ids(self, answer_ids: set[str]) -> set[str]:
+        section_dependencies: set[str] = set()
+        for answer_id in answer_ids:
+            block = self.get_block_for_answer_id(answer_id)
+            # Type ignore: block_id and section_id is never None
+            section_id = self.get_section_id_for_block_id(block["id"])  # type: ignore
+            section_dependencies.add(section_id)  # type: ignore
+        return section_dependencies
```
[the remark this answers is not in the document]
Yuyuutsu: Changed to OrderedSet

app/utilities/types.py
```diff
@@ -1,6 +1,9 @@
+from collections import namedtuple
 from typing import TypeAlias
 
 from app.questionnaire.location import Location
 from app.questionnaire.relationship_location import RelationshipLocation
 
 LocationType: TypeAlias = Location | RelationshipLocation
+
+SectionKey = namedtuple("SectionKey", "section_id list_item_id")
```
katie-gardner, line 9: I think the typed version of this would be even better (importing from typing)
```
class SectionKey(NamedTuple):
    section_id: str
    list_item_id: str | None
```

Yuyuutsu: Added 👍

tests/app/views/contexts/test_summary_context.py
```diff
@@ -438,5 +438,7 @@ def test_context_for_summary():
         "summary_type": "Summary",
         "view_submitted_response": False,
     }
-
+    print(context)
+    print("---------------------")
```
katie-gardner, line 442: why the print statements?

app/questionnaire/router.py
```diff
@@ -507,16 +508,16 @@ def _get_allowable_path(self, routing_path: RoutingPath) -> list[str]:
...
         self,
-    ) -> Generator[SectionKeyType, None, None]:
+    ) -> Generator[SectionKey, None, None]:
         for section_id in self.enabled_section_ids:
             repeating_list = self._schema.get_repeating_list_for_section(section_id)
 
             if repeating_list:
                 for list_item_id in self._list_store[repeating_list]:
-                    section_key: SectionKeyType = (section_id, list_item_id)
+                    section_key: SectionKey = SectionKey(section_id, list_item_id)
```
katie-gardner, line 517: inline type hint not needed

app/questionnaire/router.py
```diff
@@ -415,7 +414,9 @@ def get_next_location_url_for_end_of_section(self) -> str:
         return self.get_first_incomplete_location_in_questionnaire_url()
 
     def get_section_resume_url(self, routing_path: RoutingPath) -> str:
-        section_key = (routing_path.section_id, routing_path.list_item_id)
+        section_key: SectionKey = SectionKey(
```
katie-gardner, line 417: inline type hint not needed

app/data_models/progress_store.py
```diff
@@ -209,8 +207,8 @@ def remove_progress_for_list_item_id(self, list_item_id: str) -> None:
         *Not efficient.*
         """
 
-        section_keys_to_delete = [
-            (section_id, progress_list_item_id)
+        section_keys_to_delete: list[SectionKey] = [
```
katie-gardner, line 210: inline type hint not needed

app/data_models/progress_store.py
```diff
@@ -187,7 +185,7 @@ def add_completed_location(self, location: LocationType) -> None:
             self._is_dirty = True
 
     def remove_completed_location(self, location: LocationType) -> bool:
-        section_key = (location.section_id, location.list_item_id)
+        section_key: SectionKey = SectionKey(location.section_id, location.list_item_id)
```
katie-gardner, line 188: inline type hint not needed

app/data_models/progress_store.py
```diff
@@ -172,7 +170,7 @@ def add_completed_location(self, location: LocationType) -> None:
         if location.block_id not in completed_block_ids:
             completed_block_ids.append(location.block_id)  # type: ignore
 
-            section_key = (section_id, list_item_id)
+            section_key: SectionKey = SectionKey(section_id, list_item_id)
```
katie-gardner, line 173: inline type hint not needed

app/data_models/progress_store.py
```diff
@@ -155,7 +153,7 @@ def get_block_status(
     def get_completed_block_ids(
         self, *, section_id: str, list_item_id: str | None = None
     ) -> list[str]:
-        section_key = (section_id, list_item_id)
+        section_key: SectionKey = SectionKey(section_id, list_item_id)
```
katie-gardner, line 156: inline type hint not needed

app/data_models/progress_store.py
```diff
@@ -135,7 +133,7 @@ def update_section_status(
     def get_section_status(
         self, section_id: str, list_item_id: Optional[str] = None
     ) -> str:
-        section_key = (section_id, list_item_id)
+        section_key: SectionKey = SectionKey(section_id, list_item_id)
```
katie-gardner, line 136: inline type hint not needed

app/data_models/progress_store.py
```diff
@@ -114,7 +112,7 @@ def update_section_status(
         self, section_status: str, section_id: str, list_item_id: Optional[str] = None
     ) -> bool:
         updated = False
-        section_key = (section_id, list_item_id)
+        section_key: SectionKey = SectionKey(section_id, list_item_id)
```
katie-gardner, line 115: Now that this is initialised with the named tuple, you won't need to also type hint it

katie-gardner (changes requested):
Other than a couple of now redundant type hints, looks good 👍

tests/app/views/contexts/test_summary_context.py
Yuyuutsu: Oh was just using to see context when my debugger was not working 👍

app/data_models/progress_store.py
Yuyuutsu: Updated 👍

app/data_models/progress_store.py
Yuyuutsu: Updated 👍

app/data_models/progress_store.py
Yuyuutsu: Updated 👍

app/data_models/progress_store.py
Yuyuutsu: Updated 👍

app/data_models/progress_store.py
Yuyuutsu: Updated 👍

app/questionnaire/router.py
Yuyuutsu: Updated 👍

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
MebinAbraham: Still an issue, `section_dependencies` is still a normal set, so return it in an order set at the end wouldn't make it ordered. You need to set `section_dependencies` to an ordered set.

app/questionnaire/questionnaire_schema.py
```diff
@@ -1220,6 +1231,15 @@ def _get_calculated_summary_section_dependencies(
...
         return section_dependencies
 
+    def _get_section_ids_for_answer_ids(self, answer_ids: set[str]) -> OrderedSet[str]:
+        section_dependencies: set[str] = set()
+        for answer_id in answer_ids:
+            block = self.get_block_for_answer_id(answer_id)
+            # Type ignore: block_id and section_id is never None
+            section_id = self.get_section_id_for_block_id(block["id"])  # type: ignore
+            section_dependencies.add(section_id)  # type: ignore
+        return OrderedSet(section_dependencies)
```
MebinAbraham, line 1241: Needs to be an order set on initialisation or you lose the order even if you return `OrderedSet` at the end.
```suggestion
        section_dependencies: set[str] = OrderedSet()
        for answer_id in answer_ids:
            block = self.get_block_for_answer_id(answer_id)
            # Type ignore: block_id and section_id is never None
            section_id = self.get_section_id_for_block_id(block["id"])  # type: ignore
            section_dependencies.add(section_id)  # type: ignore
        return section_dependencies
```

katie-gardner (approved):
Working as expected and everything I've raised is resolved, nice 👍

tests/app/questionnaire/test_placeholder_parser.py
MebinAbraham: Still outstanding?

tests/functional/spec/placeholder_answers_on_the_path.spec.js
```diff
@@ -0,0 +1,111 @@
...
+    await expect(await $(FoodQuestionBlockPage.questionTitle()).getText()).to.contain("1 January 2017 to 1 February 2017");
+  });
+});
+
+describe("First Non Empty Item Transform Repeating Sections", () => {
+  before("Launch survey", async () => {
+    await browser.openQuestionnaire("test_placeholder_first_non_empty_item_repeating_sections.json");
+    await $(HubPage.submit()).click();
+  });
+  it("When the custom date range is entered and the answer is changed back to metadata range, then metadata should be displayed for the repeating section title", async () => {
```
MebinAbraham, line 74:
```suggestion
  it("When the custom date range is entered and the answer is changed back to metadata range, Then metadata should be displayed for the repeating section title", async () => {
```

tests/functional/spec/placeholder_answers_on_the_path.spec.js
```diff
@@ -0,0 +1,111 @@
...
+  });
+});
+
+describe("First Non Empty Item Transform Cross Section", () => {
+  before("Launch survey", async () => {
+    await browser.openQuestionnaire("test_placeholder_first_non_empty_item_cross_section_dependencies.json");
+    await $(HubPage.submit()).click();
+  });
+
+  it("When the custom date range is entered and the answer is changed back to metadata range, then metadata should be displayed for both sections", async () => {
```
MebinAbraham, line 43:
```suggestion
  it("When the custom date range is entered and the answer is changed back to metadata range, Then metadata should be displayed for both sections", async () => {
```

MebinAbraham, line 43: Should be Given When Then
```suggestion
  it("Given a custom date range is entered, When the answer is changed back to metadata range, Then the metadata date should be displayed for both sections", async () => {
```

tests/functional/spec/placeholder_answers_on_the_path.spec.js
MebinAbraham, line 74: As above.
```suggestion
  it("Given a custom date range is entered, When the answer is changed back to metadata range, Then the metadata date should be displayed for the repeating section title", async () => {
```

schemas/test/en/test_placeholder_first_non_empty_item.json
```diff
@@ -0,0 +1,283 @@
...
+                        {
+                            "id": "date-entry-block",
+                            "type": "Question",
+                            "question": {
+                                "id": "date-entry-question",
+                                "title": "What are the dates of the period that you will be reporting for?",
+                                "guidance": {
+                                    "contents": [
+                                        {
+                                            "description": "Enter a date between 24 June 2023 and 2 August 2023"
```
MebinAbraham, line 142: This doesn't match with the allowed values.

schemas/test/en/test_placeholder_first_non_empty_item_cross_section_dependencies.json
```diff
@@ -0,0 +1,291 @@
...
+                                    }
+                                ],
+                                "period_limits": {
+                                    "minimum": {
+                                        "days": 10
+                                    },
+                                    "maximum": {
+                                        "days": 50
+                                    }
+                                }
```
MebinAbraham, line 183: We should remove these from the test schema as it make it harder to test.

schemas/test/en/test_placeholder_first_non_empty_item_cross_section_dependencies.json
```diff
@@ -0,0 +1,291 @@
...
+                        {
+                            "id": "date-entry-block",
+                            "type": "Question",
+                            "question": {
+                                "id": "date-entry-question",
+                                "title": "What are the dates of the period that you will be reporting for?",
+                                "guidance": {
+                                    "contents": [
+                                        {
+                                            "description": "Enter a date between 24 June 2023 and 2 August 2023"
```
MebinAbraham, line 139: This doesn't match with the allowed values.

MebinAbraham (commented):
Other than the remaining comments, LGTM. Functionally all good also 👍

schemas/test/en/test_placeholder_first_non_empty_item.json
Yuyuutsu: Do they change this every month now? Changed it earlier this month too. Updated 🖖

schemas/test/en/test_placeholder_first_non_empty_item_cross_section_dependencies.json
Yuyuutsu: Removed 👍

schemas/test/en/test_placeholder_first_non_empty_item_cross_section_dependencies.json
Yuyuutsu: Changed 👍

app/questionnaire/router.py
Yuyuutsu: Removed 👍

schemas/test/en/test_placeholder_first_non_empty_item.json
MebinAbraham: The launcher dates changed sometime recently to use current date, we should change back.

schemas/test/en/test_placeholder_first_non_empty_item_cross_section_dependencies.json
MebinAbraham: Can you change these to be the original date check test launcher. As date were reverted in the launcher PR.

tests/app/questionnaire/test_placeholder_parser.py
Yuyuutsu: Oh i had updated it but I stashed it when going to another branch. Pushed 👍

schemas/test/en/test_placeholder_first_non_empty_item_repeating_sections.json
```diff
@@ -0,0 +1,398 @@
+{
```
MebinAbraham, line 1: This schema doesn't test repeating sections. It only adds items. The list collector page isn't a repeating section. To test repeating section, you need to add a section that repeats over the list of items.

MebinAbraham (commented):
Issue with test schema otherwise LGTM.

kylelawsonAND (approved):
Still some outstanding open comments from Mebin but looks good mate, nice job 👍

app/utilities/types.py
```diff
@@ -9,3 +9,8 @@
 class SectionKey(NamedTuple):
     section_id: str
     list_item_id: str | None
+
+
+class ProgressKeyType(NamedTuple):
```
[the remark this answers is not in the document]
Yuyuutsu: Removed ProgressKeyType and replaced where seen with SectionKey 👍

PR #1099, opened 2023-05-03
Update question definition

rmccar:
Update question definition structure to be the same as array type definitions. Also updates the array type schema name.

### How to review
- Tests pass
- Schemas with definitions launch correctly and definitions are displayed correctly

### Checklist


kylelawsonAND (approved):
Tests all good + func test definitions all working 👍

PR #1102, opened 2023-05-09
Fix rendering on Summary pages

berroar:
Fixes the rendering for Summary and View Submitted response pages so that rendering on the page is correct. 

As we were previously using the summary component incorrectly, group titles within sections with multiple groups on the summary page were not being rendered properly.

### How to review
Use the `test_view_submitted_response_repeating_sections` schema and compare the rendering on the view submitted response page between this branch and master. 

Check that Section Summaries, Calculated Summaries and list collectors on summary pages are all still being rendered as expected.

### Checklist


berroar:
[Bug: Issue Printing BERD Submission](https://trello.com/c/kyj3H3Fj/5653-bug-issue-printing-berd-submission)

kylelawsonAND (approved):
Group titles look good - rest of summary looks correct too 👍

tests/app/views/contexts/test_section_summary_context.py
```diff
@@ -156,6 +156,7 @@ def test_context_for_section_list_summary(people_answer_store):
         "summary": {
             "answers_are_editable": True,
             "collapsible": False,
+            "sections": {"groups": None},
```
MebinAbraham, line 159: Why would this happen?. Add this if we have one?

app/views/contexts/summary_context.py
```diff
@@ -96,25 +92,21 @@ def build_summary_item(
...
-        return groups
-
+        title: str = summary["title"]
 
-def set_unique_group_ids_and_titles(groups: list[dict]) -> list[dict]:
-    checked_ids = set()
-    id_value = 0
+        if groups := summary.get("sections", {}).get("groups", []):
+            if final_groups := [group for group in groups if len(group["blocks"]) > 0]:
+                self.summaries[title] = final_groups
```
katie-gardner, line 99: in `section_summary_context.py` the `title_for_location` might be null (for instance in `test_calculated_summary.json`) which passed through to this and we get None as the key for the dictionary entry - this seems to be as it doesn't have a section title/repeating section title etc, so in addition to above comment I think theres another issue with which title is being picked up

templates/partials/summary/collapsible-summary.html
```diff
@@ -3,44 +3,46 @@
...
 
-{%- for group in content.summary.groups -%}
-  {%- if group["blocks"] | length -%}
-    {%- if group.title -%}
-      {%- set item = {
-        "title": group.title,
-        "id": group.id,
-        "content": ""
-      } -%}
+{%- for section, groups in content.summary.sections.items() -%}
```
katie-gardner, line 6: section isn't being used so this could be an iteration over the dictionary values, (or over the list as per point 1. on Mebins comment)

katie-gardner (changes requested):
All still looks OK when using the test schema, just some small comments

app/views/contexts/calculated_summary_context.py
```diff
@@ -110,9 +110,11 @@ def build_view_context_for_calculated_summary(self) -> dict[str, dict[str, Any]]
         collapsible = block.get("collapsible") or False
         block_title = block["title"]
 
+        sections = {self.current_location.section_id: groups}
```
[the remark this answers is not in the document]
berroar: Updated this

app/views/contexts/section_summary_context.py
```diff
@@ -64,7 +64,10 @@ def __call__(
                 "summary_type": "SectionSummary",
                 "answers_are_editable": True,
                 "sections": {
-                    "groups": summary.get("groups"),
+                    self.current_location.section_id: {
+                        "title": title,
+                        "groups": summary.get("groups"),
+                    }
                 },
```
MebinAbraham: Should be a list to match our schema, if you need `id`, add it into the object.

app/views/contexts/summary_context.py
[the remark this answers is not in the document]
berroar: I've updated this so that each section has a title and groups as per the suggestion.

tests/app/views/contexts/__init__.py
```diff
@@ -1,21 +1,22 @@
...
-        for block in group["blocks"]:
-            assert "question" in block
-            assert "title" in block["question"]
-            assert "answers" in block["question"]
-            for answer in block["question"]["answers"]:
-                assert "id" in answer
-                assert "value" in answer
-                assert "type" in answer
+    for section in summary_context["sections"]:
+        for group in summary_context["sections"][section]:
```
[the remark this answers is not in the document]
berroar: This has been updated now

tests/app/views/contexts/test_section_summary_context.py
berroar: This happens at the moment when we have a section with just a list collector and a custom summary.

app/views/contexts/summary_context.py
```diff
@@ -31,51 +31,47 @@ def __init__(
             response_metadata,
         )
         self.view_submitted_response = view_submitted_response
+        self.summaries: dict = {}
```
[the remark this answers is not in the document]
berroar: Updated this 👍

app/views/contexts/summary_context.py
```diff
@@ -96,25 +92,21 @@ def build_summary_item(
...
-        group_id = group["id"]
-        if group_id in checked_ids:
-            id_value += 1
-            group["id"] = f"{group_id}-{id_value}"
+    def set_unique_group_ids(self) -> None:
+        checked_ids = set()
+        id_value = 0
 
-        checked_ids.add(group_id)
+        for _, groups in self.summaries.items():
```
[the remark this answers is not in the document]
berroar: Updated this 👍

app/views/contexts/summary_context.py
berroar: Changed the logic here as per Mebin's comment above 👍

app/views/contexts/section_summary_context.py
berroar: Updated now 👍

templates/partials/summary/collapsible-summary.html
berroar: Using a list now instead 👍

app/views/contexts/section_summary_context.py
```diff
@@ -63,8 +63,12 @@ def __call__(
                 "page_title": page_title,
                 "summary_type": "SectionSummary",
                 "answers_are_editable": True,
-                **summary,
-            }
+                "sections": {
+                    "groups": summary.get("groups"),
```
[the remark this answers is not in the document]
berroar: For sections with a list collector that have a custom summary we may not have groups.

rmccar:
Ive opened [this PR](https://github.com/ONSdigital/design-system/pull/2651) on the DS to fix the issue

katie-gardner (changes requested):
Everything still shows up on the summary as expected and rendering looking the same, some more small comments

tests/app/views/contexts/test_section_summary_context.py
```diff
@@ -156,7 +156,12 @@ def test_context_for_section_list_summary(people_answer_store):
         "summary": {
             "answers_are_editable": True,
             "collapsible": False,
-            "sections": {"groups": None},
+            "sections": [
+                {
+                    "groups": None,
+                    "title": "People who live here and overnight " "visitors",
```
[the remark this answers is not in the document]
berroar: Tried to fix the instances of these in this file 👍

tests/app/views/contexts/test_summary_context.py
```diff
@@ -110,379 +110,411 @@ def test_context_for_summary():
...
+                                "question": {
+                                    "answers": [
+                                        {
+                                            "currency": "GBP",
+                                            "id": "second-number-answer-also-in-total",
+                                            "label": "Second "
+                                            "answer "
+                                            "label "
+                                            "also "
+                                            "in "
```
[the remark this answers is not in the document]
berroar: Same as above, tried to fix these 👍

app/views/contexts/summary_context.py
```diff
@@ -96,25 +92,18 @@ def build_summary_item(
...
-def set_unique_group_ids_and_titles(groups: list[dict]) -> list[dict]:
-    checked_ids = set()
-    id_value = 0
 
-    for group in groups:
-        group_id = group["id"]
-        if group_id in checked_ids:
-            id_value += 1
-            group["id"] = f"{group_id}-{id_value}"
+        self.summaries.append(*summary["sections"])
```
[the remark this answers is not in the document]
berroar: Updated to use extended here, can't use update as it is a list?

app/views/contexts/summary_context.py
```diff
@@ -31,51 +31,47 @@ def __init__(
             response_metadata,
         )
         self.view_submitted_response = view_submitted_response
+        self.summaries: list = []
```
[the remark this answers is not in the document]
berroar: Updated this 👍

app/views/contexts/summary_context.py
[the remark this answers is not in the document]
katie-gardner: oops - thats what I meant, thanks

tests/app/views/contexts/test_section_summary_context.py
```diff
@@ -253,147 +256,149 @@ def test_context_for_section_summary_with_list_summary(companies_answer_store):
...
+                            "title": None,
+                        },
+                        {
+                            "blocks": [
+                                {
+                                    "add_link": "/questionnaire/companies/add-company/?return_to=section-summary",
+                                    "add_link_text": "Add another UK company or branch",
+                                    "empty_list_text": "No UK company or branch added",
+                                    "item_anchor": "#company-or-branch-name",
+                                    "item_label": "Name of UK " "company or " "branch",
```
katie-gardner, line 277: still a few places where this has gone wrong - I think you could fix it with a manual black command using the flag `--preview` which includes the experimental string processing option

katie-gardner (changes requested):

app/views/contexts/section_summary_context.py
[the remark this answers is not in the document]
MebinAbraham: Then `sections` should be added to the context. They are mutually exclusive. So add sections if you have one etc.

tests/app/views/contexts/test_section_summary_context.py
MebinAbraham: As above. Add this if we have sections.

templates/partials/summary/summary.html
```diff
@@ -3,44 +3,55 @@
...
-                calculated_question=content.summary.calculated_question,
-                icon=None
-              ),
-              "classes": "ons-u-mt-m" if loop.index > 1 else "",
-              "placeholderText": group.placeholder_text,
-              "summaryLink": group.links.add_link,
-            }
-          ]
+          "summaryTitle": section.title if summary_type == "Summary" and summary_groups else None,
+          "groups": summary_groups
```
MebinAbraham, line 46: I don't get the purpose of this?. Ends up with empty sections?.

image

tests/app/views/contexts/test_section_summary_context.py
berroar: Used the --preview option to update these but it has converted some of them so that they span over multiple lines 😕

berroar:
> > Still an issue with multiple summaries. Think it will need a DS update.
> > Screenshot 2023-05-10 at 14 53 33
> 
> This can now be fixed by updating to v62.2.1 of the DS

Updated, thanks

tests/app/views/contexts/test_section_summary_context.py
MebinAbraham: I suggest use a regex to find and replace these.

katie-gardner: you might also be able to get away with something like
`black bad_file --preview --line-length 1000`
just to get rid of the breaks like this
then run the normal command

templates/partials/summary/summary.html
berroar: The if is currently guarding against the sections with empty groups and blocks that are being generated when there is only a list collector.

tests/app/views/contexts/test_section_summary_context.py
berroar: Ok that worked, thanks 😀

app/views/contexts/section_summary_context.py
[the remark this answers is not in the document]
berroar: Updated now

tests/app/views/contexts/test_section_summary_context.py
berroar: Updated 👍

templates/partials/summary/summary.html
berroar: I've add logic to filter these empty ones now 👍

app/views/contexts/section_summary_context.py
```diff
@@ -57,23 +57,30 @@ def __call__(
...
-                ],
                 "collapsible": summary.get("collapsible"),
                 "custom_summary": summary.get("custom_summary"),
             },
         }
 
+        if summary.get("custom_summary"):
+            summary_context["summary"]["custom_summary"] = summary["custom_summary"]
+
+        if summary.get("groups") and not summary_context.get("custom_summary"):
```
MebinAbraham, line 74: This can just be `elif groups` the not summary check if implicit at that point.

berroar: Updated 👍

app/views/contexts/section_summary_context.py
```diff
@@ -57,16 +57,29 @@ def __call__(
...
+        summary_context = {
             "summary": {
                 "title": title,
                 "page_title": page_title,
                 "summary_type": "SectionSummary",
                 "answers_are_editable": True,
-                **summary,
-            }
+                "collapsible": summary.get("collapsible"),
+                "custom_summary": summary.get("custom_summary"),
```
MebinAbraham, line 67: This can be removed?. You are setting it below?

app/views/contexts/section_summary_context.py
```diff
@@ -57,16 +57,29 @@ def __call__(
...
         }
 
+        if summary.get("custom_summary"):
+            summary_context["summary"]["custom_summary"] = summary["custom_summary"]
+        elif summary.get("groups"):
+            summary_context["summary"]["sections"] = [
+                {
+                    "title": title,
+                    "groups": summary.get("groups"),
+                }
```
MebinAbraham, line 78: Minor, could use walrus.

app/views/contexts/summary_context.py
```diff
@@ -94,7 +94,9 @@ def build_summary_item(
         )["summary"]
 
         if summary.get("sections"):
-            self.summaries.extend(summary["sections"])
+            for section in summary["sections"]:
+                if any(group["blocks"] for group in section["groups"]):
+                    self.summaries.extend(summary["sections"])
```
MebinAbraham: The indentation can be removed by using `for section in summary.get("sections", []):`

Is it possible, some groups might not have block?

templates/partials/summary/summary.html
```diff
@@ -42,7 +42,7 @@
     {% do summary_sections.append
       (
         {
-          "summaryTitle": section.title if summary_type == "Summary" and summary_groups else None,
+          "summaryTitle": section.title if summary_type == "Summary",
```
MebinAbraham, line 45: Can you clarify why we do this?. Why is it set for `Summary` and not others?

app/views/contexts/summary_context.py
berroar: Yeah that's why we were getting some of the empty sections, for example in my test schema the household section was generating groups with empty blocks

app/views/contexts/section_summary_context.py
MebinAbraham: `custom_summary` can also be a walrus.

app/views/contexts/summary_context.py
berroar: Updated the above 👍

templates/partials/summary/summary.html
berroar: Without this on a section summary the section title gets repeated e.g.
Screenshot 2023-05-12 at 12 49 48

app/views/contexts/section_summary_context.py
berroar: Updated 👍

templates/partials/summary/summary.html
MebinAbraham: Ah I see, because we set the page title differently for section summaries?

berroar: Yeah, i've accidentally pushed it back up with it removed, should I add it back?

MebinAbraham: Is it the first or the section one that is this one?. What happens if you remove the separate title on section summary?

berroar: Could do that but then on the section summary I think the title would be a h2 rather than h1

MebinAbraham: Okay, revert it and keep it as h1

berroar: Reverted now 👍

rmccar (approved):

app/views/contexts/summary_context.py
MebinAbraham: The check if does any groups have a block, is it possibe, some group will have block and some won't. In that case, you would still get empty blocks?. Or is that check in templates?

katie-gardner (approved):
Most recent changes all look OK to me and on re-functionally testing the rendering is still all working as expected 👍

app/views/contexts/summary_context.py
berroar: Yeah that is guarded against in the template

app/views/contexts/summary_context.py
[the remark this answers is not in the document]
MebinAbraham: Resolving, calc summary title is done seperately.

PR #1105, opened 2023-05-16
Feat/looping 2.5 list item progress

kylelawsonAND:
This PR builds upon the [simple repeating blocks iteration ](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1100) of the [Looping 2.5 Repeating Blocks feature](https://trello.com/c/FYXB2EYJ). 
This PR augments the `Progress Store` to capture the completion progress of list items. This allows runner to have context about whether list items have been fully completed between block routings, enabling the ability for the list collector summary to show the correct icon based on completion of items (not yet done), as well as route to incomplete ones (done).

### Changes

**progress_store.py**:Updated some type hints. 
List items are added to this store keyed by id of the section in which a list item's list collector appears, and the list item id. Their completed locations are the add block and repeating blocks if they exist. 

**questionnaire_schema**: 
Used walrus operator to simplify condition check of repeating blocks.
Returning false if answer id does not belong to a block when calling `is_answer_in_repeating_blocks `

**questionnaire_store_updater.py**: Updated the method description for `remove_list_item_and_answers` to reflect the changes it has on the progress store for list items and renamed the method to `remove_list_item_data `. 

**list_remove_question.py**: Makes use of newly named `remove_list_item_data `

**list_repeating_block.py**: Added cached property for `repeating_block_id`.
`get_next_location_url` now makes use of `get_first_incomplete_repeating_block_location_for_list_item` to navigate to the next repeating block.
`handle_post(self)` now calls `add_completed_location` on the `QuestionnaireStoreUpdater` and marks the list item complete if all of the repeating blocks have a completed location on the list items progress. 

**list_collector.py**:  Added cached properties for `repeating_block_ids` and `list_name` for the list that it represents. 
`get_next_location_url(self)` now navigates to the first incomplete block of the first incomplete list item, if any exist, else navigate to the next block on the path as normal. 
Added method to evaluate list collector completeness.

**handlers/question.py**: `capture_dependent_sections_for_list(self, list_name)` ensures that only repeating sections and the section within which the list collector for the list item occurs, are captured as dependant sections - filters out the list items progresses as they should not be dependants.
Added public methods for use by list collector handlers to get the first incomplete block from a list collector or from a list item, the latter is used by the former, but may be used individually. 

### How to review
The current implementation of list item progress makes sense and the happy path of completing the Looping 2.5 journey still works while tracking progress between routes - without the list collector summary showing list item completeness of list items, however it should enforcing returning to list items repeating blocks if they were skipped.

Make use of the test schema `list_collector_repeating_blocks_section_summary.json` when testing this.
You can visit the dump/debug route to see the list item progress as you complete the list collector and see how it changes as blocks are completed as well as when list items are removed. Once all repeating blocks of a list item are complete, you should see that the list item itself is marked complete in the dump. 

Most pre-existing tests should be passing but if they fail (particularly test_with_primary_person and test_without_primary_person) they will be fixed in a future PR.

### Future PRs
Displaying list item completeness on list collector summary
Unit testing
Functional testing
Additional test schemas
Type hinting files/methods touched by this feature impl


kylelawsonAND:

app/views/handlers/question.py
```diff
@@ -228,9 +228,11 @@ def capture_dependent_sections_for_list(self, list_name):
...
 
-        section_keys_to_add = self.questionnaire_store_updater.started_section_keys(
-            section_ids=section_ids
-        )
+        # Only add sections which are repeated sections for this list, or the section in which this list is collected
+        # Prevents list item progresses being added as dependants
+        section_keys_to_add = [section_key
+                               for section_key in self.questionnaire_store_updater.started_section_keys(section_ids=section_ids)
+                               if not (section_key[0] == self.current_location.section_id and section_key[1])]
         for section_id, list_item_id in section_keys_to_add:
```
katie-gardner, line 236: I think it'd be slightly more readable to iterate over `self.questionnaire_store_updater.started_section_keys..` and do your check in the for loop, rather than creating a list and only using it once

katie-gardner (changes requested):
Functionally progress updating as expected, adding from the summary, removing, and then going back into the list collector if adding more, working well

PR description/changes is a bit outdated but rest is good - might be good to run the formatter on it so it doesn't auto fail the linter on merge

app/views/handlers/list_repeating_block.py
```diff
@@ -1,27 +1,37 @@
+from typing import Any
+
 from flask import url_for
 
 from app.views.handlers.list_edit_question import ListEditQuestion
 
 
 class ListRepeatingBlock(ListEditQuestion):
+    def __init__(self, *args: Any) -> None:
+        super().__init__(*args)
```
katie-gardner, line 10: Add kwargs for safety?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -152,7 +152,7 @@ def remove_primary_person(self, list_name: str) -> None:
             self.remove_list_item_and_answers(list_name, list_item_id)
 
     def remove_list_item_and_answers(self, list_name: str, list_item_id: str) -> None:
-        """Remove answers from the answer store and update the list store to remove it.
+        """Remove answers from the answer store, remove list item progress form the progress store and update the list store to remove it.
```
MebinAbraham, line 155: I don't see any changes that removes the progress from this method?

app/views/handlers/list_add_question.py
```diff
@@ -41,6 +42,12 @@ def handle_post(self):
         self._list_item_id = self.questionnaire_store_updater.add_list_item(
             self.parent_block["for_list"]
         )
+        completed_location = Location(
+            section_id=self.current_location.section_id,
+            block_id=self.current_location.block_id,
+            list_item_id=self._list_item_id,
+            list_name=self.parent_block["for_list"])
+        self.questionnaire_store_updater.add_completed_location(completed_location)
```
MebinAbraham, line 50: This is the add block, we don't need to track the add block's progress?

app/views/handlers/list_collector.py
```diff
@@ -21,6 +23,18 @@ def get_next_location_url(self):
             )
             return add_url
 
+        if not self._is_list_collector_complete() and (incomplete_block := next(self._get_incomplete_repeating_block_ids(), None)):
```
MebinAbraham, line 26: We don't need `_is_list_collector_complete`, if `_get_incomplete_repeating_block_ids` returns an id then we know it is not complete. Then it is consistent with how we currently do it.

So logic should just be `if incomplete_block: return url`

app/views/handlers/list_collector.py
```diff
@@ -55,5 +69,22 @@ def handle_post(self):
             # wtforms Form parents are not discoverable in the 2.3.3 implementation
             self.questionnaire_store_updater.update_answers(self.form.data)
             self.questionnaire_store_updater.save()
-        else:
+        elif not self.rendered_block["repeating_blocks"] or (self.rendered_block["repeating_blocks"] and self._is_list_collector_complete()):
             return super().handle_post()
+
+    def _is_list_collector_complete(self):
```
MebinAbraham, line 75: This would return false if the schema didn't have repeating blocks?

app/views/handlers/list_collector.py
```diff
@@ -55,5 +69,22 @@ def handle_post(self):
...
+        list_model = self._questionnaire_store.list_store.get(self.rendered_block["for_list"])
+        return all(
+            self.questionnaire_store_updater.is_section_complete(section_id=self.current_location.section_id, list_item_id=list_item_id)
+            for list_item_id in list_model.items)
+
+    def _get_incomplete_repeating_block_ids(self) -> Generator:
+        list_model = self._questionnaire_store.list_store.get(self.rendered_block["for_list"])
+        repeating_blocks = self.rendered_block.get("repeating_blocks")
```
MebinAbraham, line 83: Do this first and return if we don't have a repeating block?. Doing so makes this method reusable in a lot of places.

app/views/handlers/list_repeating_block.py
MebinAbraham: Instead of overriding the `__init__` just for an extra attr, I suggest just return what you need as a property or method.

```py
    @property
    def repeating_block_ids(self):
        return [block["id"] for block in self.parent_block["repeating_blocks"]]
```

app/views/handlers/list_repeating_block.py
```diff
@@ -1,27 +1,37 @@
...
                 return_to_block_id=self._return_to_block_id,
             )
             return repeating_block_url
 
         return super().get_next_location_url()
+
+    def handle_post(self):
+        self.questionnaire_store_updater.add_completed_location(self.current_location)
+        completed_block_ids = self.questionnaire_store_updater.get_completed_block_ids(self.current_location.section_id, self.current_location.list_item_id)
+        if all(repeating_block_id in completed_block_ids for repeating_block_id in self._repeating_block_ids):
```
MebinAbraham, line 35: If you share the `_get_incomplete_repeating_block_ids` method, this could be simplified to `if not incomplete_block: mark_as_complete()`

app/views/handlers/question.py
```diff
@@ -228,9 +228,11 @@ def capture_dependent_sections_for_list(self, list_name):
...
         section_ids.append(self.current_location.section_id)
 
-        section_keys_to_add = self.questionnaire_store_updater.started_section_keys(
-            section_ids=section_ids
-        )
+        # Only add sections which are repeated sections for this list, or the section in which this list is collected
+        # Prevents list item progresses being added as dependants
+        section_keys_to_add = [section_key
+                               for section_key in self.questionnaire_store_updater.started_section_keys(section_ids=section_ids)
+                               if not (section_key[0] == self.current_location.section_id and section_key[1])]
```
MebinAbraham, line 235: I'm a bit confused by this change. Why are we doing this?. The comment didn't clear up much.

app/questionnaire/questionnaire_store_updater.py
kylelawsonAND: There isn't a code change, I updated the comment to reflect what the function was already doing -  on line 169 it removes all progresses associated with the given list item (repeating sections and list item progresses)

app/views/handlers/list_collector.py
kylelawsonAND: So this is why I added the add block as part of list item progress, so list item completeness and then list collector completeness could all be evaluated in the same way via `is_section_complete` for list items, and then this method for the list collector itself, and there's no additional handle case for whether or not there is repeating blocks.

app/views/handlers/list_add_question.py
kylelawsonAND: See [other comment](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1105#discussion_r1197924625) for reason why this is done

app/views/handlers/question.py
kylelawsonAND: `started_section_keys(section_ids=section_ids)` with the current section id will return the list items' progresses as they are keyed by this section and their list item id.

So this filter is to ensure they aren't captured as dependant sections (unless they are supposed to be? In this impl I don't think they are supposed to be). I can reword to try and be more clear + I've refactored the function to be more elegant anyway so might make more sense.

app/views/handlers/list_collector.py
```diff
@@ -55,5 +69,22 @@ def handle_post(self):
             # wtforms Form parents are not discoverable in the 2.3.3 implementation
             self.questionnaire_store_updater.update_answers(self.form.data)
             self.questionnaire_store_updater.save()
-        else:
+        elif not self.rendered_block["repeating_blocks"] or (self.rendered_block["repeating_blocks"] and self._is_list_collector_complete()):
```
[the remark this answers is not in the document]
kylelawsonAND: Ahhh you're so right - updated

app/views/handlers/list_repeating_block.py
kylelawsonAND: Added cached property good suggestion 👍

app/views/handlers/list_collector.py
kylelawsonAND: Yep good catch! Have removed.

app/views/handlers/list_collector.py
kylelawsonAND: moved it, good call 👍

app/views/handlers/list_add_question.py
kylelawsonAND: In addition to the above reason, it also aids the debug dump as list item completeness shows by default

app/views/handlers/list_collector.py
```diff
@@ -79,12 +79,16 @@ def _is_list_collector_complete(self):
...
-                complete_block_ids = self.questionnaire_store_updater.get_completed_block_ids(self.current_location.section_id, list_item_id)
-                for repeating_block in repeating_blocks:
-                    if repeating_block["id"] not in complete_block_ids:
-                        yield list_item_id, repeating_block["id"]
+            if self.questionnaire_store_updater.is_section_complete(self.current_location.section_id, list_item_id):
+                continue
+
+            complete_block_ids = self.questionnaire_store_updater.get_completed_block_ids(self.current_location.section_id, list_item_id)
+            for repeating_block in repeating_blocks:
+                if repeating_block["id"] not in complete_block_ids:
```
katie-gardner, line 93: walrus the id?

app/views/handlers/list_repeating_block.py
```diff
@@ -1,26 +1,27 @@
...
 
 
 class ListRepeatingBlock(ListEditQuestion):
-    def __init__(self, *args: Any) -> None:
-        super().__init__(*args)
-        self._repeating_block_ids: list[str] = [block["id"] for block in self.parent_block["repeating_blocks"]]
+
+    @cached_property
+    def repeating_block_ids(self) -> list[str]:
+        return [block["id"] for block in self.parent_block["repeating_blocks"]]
```
katie-gardner, line 12: returning a mutable object from a cached property will return a reference rather than a value - you'd need to make this a tuple

katie-gardner (changes requested):

app/questionnaire/questionnaire_store_updater.py
MebinAbraham: Ah right. While we are here, update the method name to reflect that?.

app/views/handlers/list_repeating_block.py
MebinAbraham: > On post of A, _get_incomplete_repeating_block_ids would return no incomplete blocks

Why would it return no incomplete blocks?

app/views/handlers/list_collector.py
```diff
@@ -55,5 +69,26 @@ def handle_post(self):
...
+
+        list_model = self._questionnaire_store.list_store.get(self.rendered_block["for_list"])
+        for list_item_id in list_model.items:
+            if self.questionnaire_store_updater.is_section_complete(self.current_location.section_id, list_item_id):
+                continue
+
+            complete_block_ids = self.questionnaire_store_updater.get_completed_block_ids(self.current_location.section_id, list_item_id)
+            for repeating_block in repeating_blocks:
+                if repeating_block["id"] not in complete_block_ids:
+                    yield list_item_id, repeating_block["id"]
```
MebinAbraham, line 94: Instead of returning a tuple, this should return a `Location` object to be consistent with the rest of the codebase.

MebinAbraham: Secondly, I don't see any place where you are using more than the first incomplete location, therefore to be consistent this should return the first incomplete location, which is consistent with our other router methods called `_get_first_incomplete_location...` etc. This would then just be `get_first_incomplete_repeating_location`.

app/views/handlers/list_collector.py
```diff
@@ -55,5 +69,26 @@ def handle_post(self):
...
             return super().handle_post()
+
+    def _is_list_collector_complete(self):
+        list_model = self._questionnaire_store.list_store.get(self.rendered_block["for_list"])
+        return all(
+            self.questionnaire_store_updater.is_section_complete(section_id=self.current_location.section_id, list_item_id=list_item_id)
+            for list_item_id in list_model.items)
+
+    def _get_incomplete_repeating_block_ids(self) -> Generator:
+        repeating_blocks = self.rendered_block.get("repeating_blocks")
```
MebinAbraham, line 82: Minor making this a property?.

app/views/handlers/list_collector.py
MebinAbraham: I'm not sure if I follow but it seems like we are doing things we don't need to do. For example, we are adding the add block to the completed blocks just so this check passes when we don't have repeating blocks, even though the check is redundant in the scenario. We should avoid checking the store if we don't need to.

We could simplify this to not check the store if not needed. Plus in turn, you can simply your elif logic in handle_post.
```py
	@cached_property
    def repeating_blocks(self) -> list[dict] | None:
        return self.rendered_block.get("repeating_blocks")

    def _is_list_collector_complete(self) -> bool:
        if not self.repeating_blocks:
            return True

        return bool(self._get_first_incomplete_repeating_location())
```

Then handle post elif would become:

```py
        elif self._is_list_collector_complete():
            return super().handle_post()
```

app/views/handlers/question.py
MebinAbraham: I'm not sure I still follow here, would be good to have a chat about it.

app/questionnaire/questionnaire_store_updater.py
kylelawsonAND: Updated 👍

app/views/handlers/list_add_question.py
kylelawsonAND: Addressed in another comment

app/views/handlers/list_collector.py
kylelawsonAND: Don't think it's walrus-able after latest changes - but maybe I can't figure out the syntax

app/views/handlers/list_collector.py
kylelawsonAND: good call 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -475,16 +475,16 @@ def get_operands(rules: Mapping) -> Sequence:
         operands: Sequence = rules[operator]
         return operands
 
-    def _section_ids_associated_to_list_name(self, list_name: str) -> list[str]:
-        section_ids: list[str] = []
+    def _section_ids_associated_to_list_name(self, list_name: str) -> set[str]:
+        section_ids: set[str] = set()
```
MebinAbraham, line 479: The order can matter for chained relationships, so either use list or ordered set.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -149,10 +149,10 @@ def remove_primary_person(self, list_name: str) -> None:
         """
         list_item_id = self._list_store[list_name].primary_person
         if list_item_id:
-            self.remove_list_item_and_answers(list_name, list_item_id)
+            self.remove_list_item_from_all_stores(list_name, list_item_id)
 
-    def remove_list_item_and_answers(self, list_name: str, list_item_id: str) -> None:
-        """Remove answers from the answer store and update the list store to remove it.
+    def remove_list_item_from_all_stores(self, list_name: str, list_item_id: str) -> None:
```
MebinAbraham, line 154: Minor, all store isn't true?. We will soon has sds store. Maybe just `remove_list_item_data`?

app/views/handlers/question.py
kylelawsonAND: No action needed following tech session convo 24/05/23

app/questionnaire/questionnaire_schema.py
kylelawsonAND: Yep I thought as much, I have a stash which has the orderedset but couldn't safely resolve all dependency breakages - I think i may revert to list for now until Progress Value Source merges to main then I can make use of orderedset

app/views/handlers/question.py
```diff
@@ -226,12 +226,13 @@ def get_return_to_hub_url(self):
...
+        section_ids.add(self.current_location.section_id)
 
-        section_keys_to_add = self.questionnaire_store_updater.started_section_keys(
-            section_ids=section_ids
-        )
-        for section_id, list_item_id in section_keys_to_add:
+        for section_id, list_item_id in self.questionnaire_store_updater.started_section_keys(section_ids=section_ids):
+            # Only add sections which are repeated sections for this list, or the section in which this list is collected
+            # Prevents list item progresses being added as dependants as these are captured by started_section_keys(section_ids=section_ids)
+            if section_id == self.current_location.section_id and list_item_id or section_id != self.current_location.section_id and not list_item_id:
```
MebinAbraham, line 234: I think variables and parentheses would make this clearer.

Is `section_id != self.current_location.section_id and not list_item_id:` correct?. This is could be another section which dependent the list?. Section id doesn't match and there is no list id which mean it is a normal section that depends on the section, for example it uses the list length for a when rule.

app/views/handlers/list_collector.py
```diff
@@ -55,5 +77,14 @@ def handle_post(self):
...
             return super().handle_post()
+
+    def _is_list_collector_complete(self):
+        if not self.repeating_block_ids:
+            return True
+
+        list_model = self._questionnaire_store.list_store.get(self.list_name)
+        return all(
+            self.questionnaire_store_updater.is_section_complete(section_id=self.current_location.section_id, list_item_id=list_item_id)
+            for list_item_id in list_model.items)
```
MebinAbraham, line 90: Can this not use `get_first_incomplete_repeating_block_location`?

app/views/handlers/question.py
kylelawsonAND: yep good catch, not sure where I was going with that

app/views/handlers/list_collector.py
kylelawsonAND: Another nice catch - the method will essentially just be calling that method and returning the inverted bool interpretation of the result so arguably could be removed, but I still think it's worth keeping so the usage of `get_first_incomplete_repeating_block_location` for the purpose of list collector completeness is clear.

MebinAbraham: I think the usage of `get_first_incomplete_repeating_block_location` doesn't become unclear if we did this?.
I think it is beneficial because:
1. Avoid having another bit of logic to get the same end result
2. Keeps it consistent with how we already do it. This is what happen in the `is_path_complete` method, which inverts `get_first_incomplete_...` method.
3. Personally, I don't think it is unclear either, currently it is, all sections for the list item are complete, it would just become, all sections don't have an incomplete or not any section have incomplete blocks.

If I had to choose one reason it would simply be for consistency. See what others think and go with consensus.

kylelawsonAND: Sorry I wasn't meaning that I was keeping this method as is, I agree with you that it should be changed to use the inverted result of `get_first_incomplete_repeating_block_location`, which has now been implemented

app/questionnaire/questionnaire_store_updater.py
kylelawsonAND: Done 👌

app/views/handlers/list_collector.py
MebinAbraham: Ah cool, I probably misunderstood. 👍

MebinAbraham (approved):
Looking good, almost there

app/views/handlers/list_collector.py
kylelawsonAND: Bad phrasing from me, is my bad

app/views/handlers/question.py
```diff
@@ -251,3 +252,29 @@ def clear_radio_answers(self):
...
+                return incomplete_location
+
+    def get_first_incomplete_repeating_block_location_for_list_item(self, repeating_block_ids: Sequence[str], section_id: str, list_item_id: str,
+                                                                    list_name: str) -> Location | None:
+        if self.questionnaire_store_updater.is_section_complete(section_id, list_item_id):
+            return None
+
+        complete_block_ids = self.questionnaire_store_updater.get_completed_block_ids(section_id, list_item_id)
+        for repeating_block_id in repeating_block_ids:
+            if repeating_block_id not in complete_block_ids:
```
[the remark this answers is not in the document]
kylelawsonAND: Ah didn't catch that method was kicking around in router, have removed mine and used that 👍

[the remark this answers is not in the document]
MebinAbraham: I clearly was testing you by putting `if complete`  instead of `if not complete` 😄 
Looks good 👍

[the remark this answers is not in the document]
kylelawsonAND: Hahah github comment box is the best python IDE i heard ;)

Yuyuutsu (commented):
Functionally works as expected, I can see the progress for each list item when dump/debug. 
I just have a question just for my own curiosity,  in the future are we planning to make it clearer which list items we are completing. For example in this photo `car two` and `car four `are incomplete. When I click No and then Save and continue , it shows the next part for one of the car companies (car two) but it does not clarify which one.
image

image

Yuyuutsu (approved):

kylelawsonAND:
> Functionally works as expected, I can see the progress for each list item when dump/debug. Also happy with the changes from my level of understanding. I just have a question just for my own curiosity, in the future are we planning to make it clearer which list items we are completing. For example if I enter car companies `one`, `two`, `three` and `four`. I complete `one` and `three` but do not fully complete `two` and `four`. Then I press save and continue, it displays the next sections for `car two` since its in-complete but it is a generic page that does not have any indication that we are filling it out for car two. Sorry if that did not make any sense. image

Yeah this was a concern of mine, but it will be handled via placeholders which will say the item's Add block answer values which will be displayed ✨ somewhere ✨ 

I'll update my test schema to use placeholders in the next mini PR

PR #1106, opened 2023-05-16
Looping 2.5 | Repeating Blocks for List Items

kylelawsonAND:
This PR implements the feature [ Looping 2.5](https://trello.com/c/FYXB2EYJ).
It is required for future surveys that items added via a list collector can have additional information provided for them as they are added on following pages after addition to the list. 
To support this, the list collector has been augmented to support repeating blocks where are a list of blocks shown after every add block, the answers from which are stored in the same way answers from repeating sections would be stored. 
Repeating blocks are mandatory and thus will be redirected to from the list collector pages if the suer attempts to proceed from the list collector block without having completed all repeating blocks for the added list items. 

### Changes
**app/data_models/progress_store.py**: Misc type hinting

**app/jinja_filters.py**: `map_summary_item_config` now no longer takes an icon string to pass to `map_list_collector_config`, which also now does not take this parameter. Instead this method takes a bool flag indicating whether list item icons should render. The icon that renders is the green checkmark icon if the a list item has repeating blocks and is complete, else no icon should render. 
Additional type hinting. 

**app/questionnaire/questionnaire_schema.py**: Added `ListRepeatingBlock` to `LIST_COLLECTOR_CHILDREN`.
Added a new private class attribute to track repeating blocks by their block id in a dictionary. 
Repeating blocks are added as entries to the the `self._blocks_by_id` attribute as well as the new attribute when `self._blocks_by_id` is populated at construction. Parent id maps are populated as a normal block.
Redundant type ignore removed from `_update_answer_dependencies_for_value_source`.
`_block_for_answer` returns the repeating block id for an answer id from a repeating block instead of the list collector block which is the case for the other other list collector child blocks.
Removal of redundant type ignore.
Added public class method `is_block_in_repeating_blocks` to return true if the given block id corresponds to a repeating block, else false. 

**app/questionnaire/questionnaire_store_updater.py**: Public method `remove_list_item_and_answers` has been renamed to `remove_list_item` as it now removes list item progress from the progress store. No code was needed to add this, as the list item progress is keyed in the same way as the repeating section progress and is thus removed the same way.
Public methods `get_relationship_answers_for_list_name` and `get_chronological_section_dependents` are now private as they are not used publicly and prevents the mypy/pylint error of too many public methods. 
Kwargs enforced on public method `update_section_status`.
Added a public method `is_section_complete` which simply calls and returns the result of `ProgressStore.is_section_complete`, allowing services which do not have access to the ProgressStore to check section progress. 

**app/views/contexts/list_context.py**:  Misc type hinting throughout.
Private method `_build_list_items_context` now takes `section_id` and a bool flag `has_repeating_blocks` which is used to build the list item context.

**app/views/contexts/summary/list_collector_block.py**: Public method `list_summary_element` passes the repeating_blocks (if any) the private method `_get_related_answers` which now expects this. This allows the repeating blocks to have renderabel Blocks generated. This method also supplies the new params to `ListContext._build_list_items_context`.
Public method `get_blocks_for_related_answers` has been made private. 

**app/views/handlers/block_factory.py**: New block type `ListRepeatingBlock` has been added. 

**app/views/handlers/list_action.py**: New private method `_is_returning_to_section_summary` has been added to remove duplication from other methods. 

**app/views/handlers/list_add_question.py**: Added class attribute `_list_item_id` so that computing the next location url after posting, has context of the list item. 
Public method `get_next_location_url` now navigates to the first repeating block if repeating blocks exist, else goes to the parent location as normal. 
Public method `handle_post` now assigns the class attribute `_list_item_id` once the list item has been added. 

**app/views/handlers/list_collector.py**: Added a new cached property `repeating_block_ids` which returns repeating block ids if repeating blocks exist. 
Added a new cached property `list_name` which returns the list name.
Public method `get_next_location_url` now navigates to any incomplete repeating blocks. 
Public method `get_context` pass section id and `has_repeating_blocks` bool flag to the list context constructor. 
Private method `_is_list_collector_complete` which makes use of the new parent method `get_first_incomplete_repeating_block_location` to return if the list collector this instance represents, is complete. 

**app/views/handlers/list_remove_question.py**: Makes use of newly renamed `QuestionnaireStoreUpdater. remove_list_item_data`.

**app/views/handlers/list_repeating_block.py**: New block handler for Repeating Blocks.
This class extends ListAction as it can make use of the `get_next_location_url` method in the parent. 
`get_next_location_url` is overridden to return to section summary if the url specifies that it should. It will also navigate to the next repeating block by reusing the `get_first_incomplete_repeating_block_location_for_list_item` method from the parent. If it is the last incomplete repeating block, it navigates via the parent's overridden method. 

**app/views/handlers/question.py**: Public method `get_list_summary_context` now populates the new LisContext constructor parameters. 
Public method `capture_dependent_sections_for_list` will not add list item progresses as dependants (which they should not be). This is a workaround to allow us to reuse the ProgressStore for tracking list item progresses. 
New public method `get_first_incomplete_repeating_block_location_for_list_item` added which will check all repeating blocks for a given list item and return the location of the first incomplete one. returning None before iterating over the repeating blocks if the list item is already marked complete in the progress store. 
New public method `get_first_incomplete_repeating_block_location` added which will return the location of the first incomplete repeating block from a list model specified by section, list name and repeating blocks. This method iterates through all list items in the model and calls the former method on each, returning early if an incomplete location is found. 

**app/views/handlers/relationships/unrelated_question.py**: Public method `get_list_summary_context` supplies the new ListContext constructor parameters. 

**schemas/test/en/test_list_collector_repeating_blocks.json**: A new test schema which contains only a list collector with repeating blocks and no driving question, section summary or placeholders. 

**schemas/test/en/test_list_collector_repeating_blocks_section_summary.json**: A new test schema which contains a list collector with repeating blocks that does have a driving question, section summary and placeholders in the repeating blocks.

**tests/app/conftest.py**: Moved the location fixture here to be used by the questionnaire tests as well as contexts tests.

**templates/listrepeatingblock.html**: A new jinja template which is used to render the repeating block which simply extends the list-action.html template. 

**templates/partials/summary/collapsible-summary.html**: Removed passing icon to `map_summary_item_config`.

**templates/partials/summary/list-summary.html**: Using kwargs when calling `map_list_collector_config` and passing the new `render_icon` bool flag. 

**templates/partials/summary/summary.html**: Removed passing icon to `map_summary_item_config`.

**tests/app/conftest.py**: Moved the location fixture to here from the tests/app/questionnaire level to be used by the questionnaire tests as well as contexts tests.

**tests/app/questionnaire/conftest.py**: Moved the location fixture from here to the tests/app level to be used by the questionnaire tests as well as contexts tests.

**tests/app/questionnaire/test_questionnaire_schema.py**: Test coverage for getting repeating blocks and answers by id and adding them to the recorded blocks in QuestionnaireSchema. 

**tests/app/questionnaire/test_questionnaire_store_updater.py**: Ammending to call a public, now private method, in existing tests. 
Additional formatting. 

**test_jinja_filters**: Amending tests which expected an icon string to now expect the render icon bool. 

**tests/app/views/contexts/conftest.py**: Adding fixtures for an AnswerStore and ListStore with repeating blocks list collector. 

**test_list_context**: Many tests now have a `ProgressStore` fixture passed to them to be used to construct the tested `ListContext`, additionally they have the location fixture passed in as well to support the new constructor param. 
Some assertions now also assert that the `"is_complete"` and `"repeating_blocks" `bool flags are also present in the built list summaries/contexts. 
Also added test coverage to check ListContexts build correctly with complete and incomplete list collectors with and without repeating_blocks. 

**test_section_summary**: 
Many tests now have a `ProgressStore` fixture passed to them to be used to construct the tested `ListContext`.
Some assertions now also assert that the `"is_complete"` and `"repeating_blocks" `bool flags are also present in the built list summaries/contexts. 

**tests/functional/generate_pages.py**: `process_block` now calls itself on repeating blocks if they exist.

**tests/functional/helpers.js**: Renamed and modified method `checkPeopleInList` to `checkItemsInList` to be more generic and use with repeating blocks func tests. 

**tests/functional/spec/features/repeating_blocks/list_collector_repeating_blocks.spec.js**: Added functional tests to cover the [feature scenarios](https://docs.google.com/spreadsheets/d/1UjC6yX5tmmTcXuZ0Y7BZuF3uS07GmCWvDtB__JQoOWs/edit#gid=2059582577).

**tests/integration/integration_test_case.py**: Added `getLinksByAttribute` helper method to find links by a given dict of attributes.

**tests/integration/questionnaire/test_questionnaire_list_collector_repeating_blocks.py**: Added integration tests to cover the [feature scenarios](https://docs.google.com/spreadsheets/d/1UjC6yX5tmmTcXuZ0Y7BZuF3uS07GmCWvDtB__JQoOWs/edit#gid=2059582577) and supplement coverage. 

### How to review
Familiarise yourself with the Looping 2.5 card and scenarios linked above. Make use of the test schema `test_list_collector_repeating_blocks_section_summary.json` to manually verify and attempt to break the scenarios and requirements.
Try out other test schemas that make use of List Collectors and check that other features work as expected.
All new and existing unit, integration and functional tests should be passing locally as well as on PR.


kylelawsonAND:

app/views/contexts/summary/list_collector_block.py
```diff
@@ -235,3 +242,22 @@ def get_blocks_for_related_answers(
...
+            list_answers = [
+                (answer_id, answer_list_id)
+                for (answer_id, answer_list_id) in self._answer_store.answer_map
+                if answer_list_id == list_id
+                and self._schema.is_answer_in_repeating_blocks(answer_id)
+            ]
+            for answer_id, answer_list_id in list_answers:
+                block = self._schema.get_block_for_answer_id(answer_id)
+                if block:
+                    blocks[block].append(answer_id)
```
MebinAbraham, line 258: We shouldn't do this like this. If 2 list collector have the same list name, this would break. It seems like we are doing all this, just to get a list of blocks we need. It should be all blocks inside `repeating_blocks` for this list name. Any blocks without an answer should display as `No answers provided`.

We care about all blocks, not just the ones that have at least 1 answer.

Therefore, this method can be removed and you can have something like:

Pass in `repeating_blocks` which is in scope.

```PY
    def _get_related_answers(
        self, list_model: ListModel, repeating_blocks
    ) -> dict[str, list[dict]] | None:
        section_id = self._section["id"]

        related_answers = self._schema.get_related_answers_for_list_for_section(
            section_id=section_id, list_name=list_model.name
        )

        blocks: list[dict] = []

        if related_answers:
            blocks += self._get_blocks_for_related_answers(related_answers)

        if repeating_blocks:
            blocks += repeating_blocks
```

app/views/handlers/list_repeating_block.py
```diff
@@ -0,0 +1,49 @@
...
+        if not self.get_first_incomplete_repeating_block_location_for_list_item(
+            repeating_block_ids=self.repeating_block_ids,
+            section_id=self.current_location.section_id,
+            list_item_id=self.current_location.list_item_id,
+            list_name=self.current_location.list_name,
+        ):
+            self.questionnaire_store_updater.update_section_status(
+                True,
+                self.current_location.section_id,
+                self.current_location.list_item_id,
```
MebinAbraham, line 47: Minor, for clarity update this existing method to enforce keyword args.

app/views/contexts/summary/list_collector_block.py
kylelawsonAND: Great catch, not sure where Iw as going with this looking at it now, change made 👍

app/views/handlers/list_repeating_block.py
kylelawsonAND: Enforced 👍

schemas/test/en/test_list_collector_repeating_blocks.json
```diff
@@ -0,0 +1,227 @@
...
+                    "title": "List",
+                    "blocks": [
+                        {
+                            "id": "list-collector-block",
+                            "type": "ListCollector",
+                            "for_list": "repeating_blocks_list_collector",
+                            "question": {
+                                "id": "list-collector-next-question",
+                                "type": "General",
+                                "title": "Do you need to add any more to this list?",
```
katie-gardner, line 73: As per new rule - I think context of the schema could be improved (subject to not creating too much overhaul in tests)

schemas/test/en/test_list_collector_repeating_blocks.json
```diff
@@ -0,0 +1,227 @@
...
+        }
+    },
+    "sections": [
+        {
+            "id": "section",
+            "groups": [
+                {
+                    "id": "group",
+                    "title": "List",
+                    "blocks": [
```
katie-gardner, line 65: Should this have a driving question?

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
```diff
@@ -0,0 +1,361 @@
...
+                                                "type": "Number",
+                                                "maximum": {
+                                                    "value": 999,
+                                                    "exclusive": false
+                                                },
+                                                "decimal_places": 0
+                                            },
+                                            {
+                                                "id": "registration-date",
+                                                "label": "Date of Registration (Mandatory)",
```
katie-gardner, line 215: If you enter a date in the future you'll then be asked for details about how your company has been trading over the last -124 years. 😁 Not an issue with the feature but might be nice to cap the date

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
```diff
@@ -0,0 +1,361 @@
...
+                                                    },
+                                                    {
+                                                        "label": "No",
+                                                        "value": "No"
+                                                    }
+                                                ]
+                                            },
+                                            {
+                                                "type": "Radio",
+                                                "label": "Has this company been trading in the EU? (Mandatory)",
```
katie-gardner, line 276: You've written (Mandatory) but specified `mandatory: false`

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
```diff
@@ -0,0 +1,361 @@
...
+        {
+            "answer_id": "authorised-trader-eu-radio",
+            "code": "1e"
+        },
+        {
+            "answer_id": "any-other-companies-or-branches-answer",
+            "code": "2"
+        }
+    ],
+    "questionnaire_flow": {
```
katie-gardner, line 54: change one of the tests to use `Hub`? probably this one

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
```diff
@@ -0,0 +1,361 @@
...
+                                    "placeholders": [
+                                        {
+                                            "placeholder": "company_name",
+                                            "value": {
+                                                "source": "answers",
+                                                "identifier": "company-or-branch-name"
+                                            }
+                                        }
+                                    ]
+                                }
```
katie-gardner, line 353: I'd probably sandwich the list collector with a question or two either side, just to check it shows up nicely on the summary page with non list collector related stuff (Spoiler alert, I tried it and it does, and it also works with dynamic answers)

katie-gardner (changes requested):
Haven't quite finished but heres my initial comments from reading through as a whole - functionally this is sound, not been able to break anything so far, really nice 👍

app/views/handlers/list_repeating_block.py
```diff
@@ -0,0 +1,49 @@
...
+        return super().get_next_location_url()
+
+    def handle_post(self):
+        self.questionnaire_store_updater.add_completed_location(self.current_location)
+        if not self.get_first_incomplete_repeating_block_location_for_list_item(
+            repeating_block_ids=self.repeating_block_ids,
+            section_id=self.current_location.section_id,
+            list_item_id=self.current_location.list_item_id,
+            list_name=self.current_location.list_name,
+        ):
```
kylelawsonAND, line 43: Add comment to explain update_section_status is updating list item progress - maybe edit method name?

app/views/handlers/question.py
```diff
@@ -254,3 +262,44 @@ def clear_radio_answers(self):
...
+                return incomplete_location
+
+    def get_first_incomplete_repeating_block_location_for_list_item(
+        self,
+        repeating_block_ids: Sequence[str],
+        section_id: str,
+        list_item_id: str,
+        list_name: str,
+    ) -> Location | None:
+        if self.questionnaire_store_updater.is_section_complete(
```
berroar, line 289: I know we spoke about this briefly in the session yesterday, but any reason why we can't do this here: `self._questionnaire_store.progress_store.is_section_complete(section_id=section_id, list_item_id=list_item_id)`? Then we wouldn't need the added method in updater

app/views/handlers/question.py
```diff
@@ -254,3 +262,44 @@ def clear_radio_answers(self):
                 answer_ids_to_remove, self.current_location.list_item_id
             )
             self.questionnaire_store_updater.save()
+
+    def get_first_incomplete_repeating_block_location(
```
petechd, line 266: Just having a quick look before real review, this one and especially the one below begs for forced named arguments.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -247,6 +247,11 @@ def update_section_status(
             status, section_id, list_item_id
         )
 
+    def is_section_complete(
```
petechd, line 250: No need for this duplicate method.

app/questionnaire/questionnaire_schema.py
```diff
@@ -1249,6 +1269,9 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
                 if item["for_list"] == list_name and item.get("item_anchor_answer_id"):
                     return f"#{str(item['item_anchor_answer_id'])}"
 
+    def is_block_in_repeating_blocks(self, block_id: str) -> bool:
```
petechd, line 1272: Is this ever being used outside tests?

app/views/handlers/list_repeating_block.py
```diff
@@ -0,0 +1,49 @@
...
+
+from flask import url_for
+
+from app.views.handlers.list_edit_question import ListEditQuestion
+
+
+class ListRepeatingBlock(ListEditQuestion):
+    @cached_property
+    def repeating_block_ids(self) -> list[str]:
+        return [block["id"] for block in self.parent_block["repeating_blocks"]]
```
berroar, line 11: Any reason why you're not reusing the map you populated upfront in questionnaire schema here -`repeating_blocks_by_id`?

app/views/handlers/list_add_question.py
```diff
@@ -1,15 +1,35 @@
...
     def is_location_valid(self):
         if not super().is_location_valid() or self._current_location.list_item_id:
             return False
         return True
 
     def get_next_location_url(self):
+        if self._list_item_id and (
+            repeating_blocks := self.parent_block.get("repeating_blocks")
+        ):
+            repeating_block_url = url_for(
```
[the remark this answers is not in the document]
kylelawsonAND: Yeah I do agree, I was keeping consistent with the pattern used elsewhere in the handlers. Will change and start the movement

tests/integration/questionnaire/test_questionnaire_list_collector_repeating_blocks.py
```diff
@@ -0,0 +1,478 @@
...
+    ):
+        self.add_company_from_list_collector(
+            company_name=company_name, is_driving=is_driving
+        )
+        self.post_repeating_block_1(
+            registration_number=registration_number, registration_date=registration_date
+        )
+        self.post_repeating_block_2(trader_uk=trader_uk, trader_eu=trader_eu)
+
+    def assert_company_completed(self, company_name: str, selector: str):
```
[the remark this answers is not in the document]
kylelawsonAND: I'd need to passing label number as well as company number will remain the same for items, but the label number is sequential, so removed items with say company number 3 will have label number 2 is company 2 is removed. Will do this anyway as it does reduce duplication.

tests/integration/questionnaire/test_questionnaire_list_collector_repeating_blocks.py
```diff
@@ -0,0 +1,478 @@
...
+        self.assertStatusNotFound()
+
+    def test_happy_path(self):
+        self.launchSurvey("test_list_collector_repeating_blocks_section_summary")
+
+        # Add some items to the list
+        self.add_three_companies()
+
+        # Remove item 2
+
```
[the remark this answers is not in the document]
kylelawsonAND: Will group all statements by comment to remove empty lines

app/views/contexts/list_context.py
```diff
@@ -41,16 +45,20 @@ def __call__(
...
-    ) -> Generator[dict[str, Any], Any, None]:
+        section_id: str,
+        has_repeating_blocks: bool,
+        return_to: str | None,
+        summary_definition: Mapping,
+        edit_block_id: str | None,
+        remove_block_id: str | None,
+        primary_person_edit_block_id: str | None,
+        for_list_item_ids: Sequence[str] | None,
+    ) -> Generator[dict, Any, None]:
```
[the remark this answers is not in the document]
kylelawsonAND: 🧠

app/views/handlers/list_collector.py
```diff
@@ -9,6 +11,16 @@ def __init__(self, *args):
...
 
+    @cached_property
+    def repeating_block_ids(self) -> list[str]:
+        return [
+            block["id"] for block in self.rendered_block.get("repeating_blocks", [])
+        ]
+
+    @cached_property
+    def list_name(self) -> str:
+        return self.rendered_block.get("for_list")
```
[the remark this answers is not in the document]
kylelawsonAND: Great catch, surprised this wasn't caught by linters 🤔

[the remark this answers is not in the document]
katie-gardner: I think its because mypy is turned off for handlers at the moment and Micks ticket will turn it on

schemas/test/en/test_list_collector_repeating_blocks.json
kylelawsonAND: Will augment title on list collector and repeating blocks

schemas/test/en/test_list_collector_repeating_blocks.json
kylelawsonAND: Nope, this is a really barebones list collector only with repeating blocks.
`test_list_collector_repeating_blocks_section_summary` has one to cover testing that.

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
kylelawsonAND: Hahah yeah noticed this in the PR run through, will add validation for this (assumign it works 🤞 ).
Regarding this causing a dependancy on a repeating block on a prev repeating block, the requirements do not specify that must be handled at this moment in time.

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
kylelawsonAND: Woops

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
kylelawsonAND: Not sure if there's much value in this, Hub and repeating blocks wouldn't interact?

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
kylelawsonAND: I actually had this before and removed because of my manual test fatigue, will add it back in and update any affected tests to expect it.

app/views/handlers/list_repeating_block.py
kylelawsonAND: Have reworded the ProgressStore comments, methods and variable names to reflect what it stores and tracks now, which is section progress and list item progress.

app/views/handlers/question.py
kylelawsonAND: Progress store is a private class attribute so could either make it public or keep the forwarding method

kylelawsonAND: Exposing as public implies it should be accessed by those who have access to the QuestionnaireStoreUpdater, whereas having the forwarding method limits accessing it to only the correct circumstances

app/views/handlers/question.py
kylelawsonAND: Yeah good shout

app/questionnaire/questionnaire_store_updater.py
kylelawsonAND: I [explained above](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1106#discussion_r1231922361) in Rhys' comment the rationale for creating the forwarding method. 
Let me know what you think given that justification.

app/questionnaire/questionnaire_schema.py
kylelawsonAND: Great point, I made this in an earlier iteration and guess we don't need it now!

app/views/handlers/list_repeating_block.py
kylelawsonAND: Great shout, will do that now

app/views/handlers/list_add_question.py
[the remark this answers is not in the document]
kylelawsonAND: Done 👍

tests/integration/questionnaire/test_questionnaire_list_collector_repeating_blocks.py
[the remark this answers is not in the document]
kylelawsonAND: Done 👍

tests/integration/questionnaire/test_questionnaire_list_collector_repeating_blocks.py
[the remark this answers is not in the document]
kylelawsonAND: Done 👍

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
kylelawsonAND: Done 👍

app/views/handlers/list_repeating_block.py
kylelawsonAND: I'll leave this to another dev to mark as resolved

app/views/handlers/question.py
kylelawsonAND: Done 👍

app/questionnaire/questionnaire_schema.py
kylelawsonAND: Done 👍

app/views/handlers/question.py
berroar: Not quite sure what you mean by it being private? We already access progress store via the questionnaire store in this class (for example in the form method), the same way we do with the answer_store and list_store etc. I do mean `questionnaire_store` rather than `questionnaire_store_updater` if that's causing any confusion

app/jinja_filters.py
```diff
@@ -695,6 +693,14 @@ def map_list_collector_config(
                 }
             )
 
+        icon = (
```
petechd, line 696: Were we supposed to remove the person icon from list collector? Primary person and old lc will be affected, I can't find any mention on your card.

app/views/handlers/list_repeating_block.py
```diff
@@ -0,0 +1,58 @@
+from functools import cached_property
+
+from flask import url_for
+
+from app.views.handlers.list_action import ListAction
+
+
+class ListRepeatingBlock(ListAction):
```
petechd, line 8: I've just compared with the card and we are asked to add `New each block type inside  repeating_blocks will be of type ListRepeatingQuestion`. So the new block should be called `ListRepeatingQuestion`?

app/data_models/progress_store.py
```diff
@@ -76,87 +77,93 @@ def is_dirty(self) -> bool:
     def is_routing_backwards(self) -> bool:
         return self._is_routing_backwards
 
-    def is_section_complete(
+    def is_section_or_list_item_complete(
```
berroar, line 80: Not sure about some of the renaming being done in this class e.g. `is_section_or_list_item_complete`, can a list item be in progress/complete? We are really only checking completeness of sections and not list_items just that some sections have attached list_item_ids when in repeats?

kylelawsonAND: After re-reveiwing, let me know if methods should be renamed `is_section_or_list_item_complete` to `is_progress_complete` - same for other similar method names.

katie-gardner: I think I'd be in favour of the method name sticking with what its checking
Maybe it could be something like `is_section_or_list_item_repeating_blocks_complete`? but im not sure if thats just more convoluted... in which case perhaps progress is the name to go for

app/views/handlers/question.py
kylelawsonAND: Ahhh I see! I didn't realise that QuestionnaireStore was available via parent BlockHandler and that ProgressStore was public via that, thanks for spotting will change that

app/questionnaire/questionnaire_store_updater.py
kylelawsonAND: Will fix in the way Rhys identified.

app/jinja_filters.py
kylelawsonAND: Yeah fair point other List Collector without repeating Blocks will lose the Person icon. Will modify to default to Person if not repeating blocks, the check or None if complete or not with repeating blocks.

app/views/handlers/list_repeating_block.py
kylelawsonAND: i could have sworn I was following the pattern of calling list actions [Action]Block but they're all Question, will change thanks!

app/questionnaire/questionnaire_schema.py
```diff
@@ -290,6 +292,14 @@ def _get_blocks_by_id(self) -> dict[str, ImmutableDict]:
                             nested_block_id = nested_block["id"]
                             blocks[nested_block_id] = nested_block
                             self._parent_id_map[nested_block_id] = block_id
+                    if repeating_blocks := block.get("repeating_blocks"):
+                        for repeating_block in repeating_blocks:
+                            repeating_block_id = repeating_block["id"]
+                            blocks[repeating_block_id] = repeating_block
+                            self._parent_id_map[repeating_block_id] = block_id
+                            self._repeating_blocks_by_id[
```
katie-gardner, line 300: I think this can be a list of ids, the dict values are never used?

app/data_models/progress_store.py
kylelawsonAND: I'll go with `is_progress-complete` and rely on the new docstrings to fully explain what is being elements are having their progresses tracked.

app/questionnaire/questionnaire_schema.py
```diff
@@ -707,6 +717,9 @@ def get_first_block_id_for_section(self, section_id: str) -> str | None:
     def get_blocks(self) -> Iterable[ImmutableDict]:
         return self._blocks_by_id.values()
 
+    def get_repeating_blocks(self) -> dict[str, ImmutableDict]:
+        return self._repeating_blocks_by_id
```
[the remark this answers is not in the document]
kylelawsonAND: Yeah I did think this when adding, I opted for returning the blocks by id so that anything else in the future (routing/path finding etc etc) can make use of this, the result can be transformed in whichever way the caller requires.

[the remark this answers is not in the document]
kylelawsonAND: Will change to only returning IDs tho

app/questionnaire/questionnaire_schema.py
kylelawsonAND: I think these should be tracked as blocks by id like the other x_by_ids, as a means of collecting of them up front and making searching for the blocks by id more efficient

kylelawsonAND: Ah but then you can just use blocks_by_id for the block so double recording the repeating blocks is redundant -  yeah fair will change to only ids

app/views/handlers/list_add_question.py
```diff
@@ -1,15 +1,34 @@
...
+        self._list_item_id: str | None = None
+        super().__init__(*args)
+
     def is_location_valid(self):
         if not super().is_location_valid() or self._current_location.list_item_id:
             return False
         return True
 
     def get_next_location_url(self):
+        if self._list_item_id and (
```
katie-gardner, line 19: Just had a thought - when would `self._list_item_id` actually be none?

katie-gardner (approved):
Aside from the minor comments, this looks great, there are some unit test gaps in `get_next_location_url` or getting first incomplete repeating block - however this is comprehensively covered in the integration test so I'm gonna give this a thumbs up 👍

kylelawsonAND:
> Aside from the minor comments, this looks great, there are some unit test gaps in `get_next_location_url` or getting first incomplete repeating block - however this is comprehensively covered in the integration test so I'm gonna give this a thumbs up (subject to resolving last couple things) 👍

Thanks, none of the list collector handlers had been specifically targeted by unit tests, they've only ever been covered by integration tests, so just followed that approach.

schemas/test/en/test_list_collector_repeating_blocks_section_summary.json
berroar: Just noticed this comment - I think it may be worth changing this to a hub in order to be able to functionally test the completeness of the section when adding new items etc?

app/views/handlers/question.py
kylelawsonAND: Now making use of `ProgressStore` directly via `QuestionnaireStore` 👍

app/questionnaire/questionnaire_store_updater.py
kylelawsonAND: Removed 👍

app/views/handlers/list_add_question.py
katie-gardner: ahhh I see, so you'd end up being routed back to the 'any other items to add' block if it was misused. I imagine we'd have a lot of issues with progress in that case, as blocks wouldn't be marked complete before finding next incomplete url, but not against doing this

app/views/handlers/list_repeating_block.py
kylelawsonAND: Renamed to ListCollectorQuestion 👍

app/data_models/progress_store.py
kylelawsonAND: Renamed to use `progress` as the term

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
kylelawsonAND: Now returns just the ids 👍

app/questionnaire/questionnaire_schema.py
kylelawsonAND: Done 👍

tests/app/data_model/test_progress_store.py
```diff
@@ -413,9 +433,18 @@ def test_is_section_complete():
 
     store = ProgressStore(completed)
 
-    assert store.is_section_complete(section_id="s1", list_item_id=None) is True
-    assert store.is_section_complete(section_id="s4", list_item_id="123abc") is True
-    assert store.is_section_complete(section_id="s5", list_item_id="456def") is True
+    assert (
+        store.is_section_or_list_item_complete(section_id="s1", list_item_id=None)
```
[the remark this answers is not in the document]
kylelawsonAND: Removed 👍

app/jinja_filters.py
kylelawsonAND: Actually I think I did go over this with Mebin, we were to remove the icons because the person is appropriate for all lists. I'll document this in the trello card, but I'm sure that was the decision, my bad for not documenting on the card. There is mention of it on my [mini PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1112) which was approved by Mebin.

tests/functional/spec/features/repeating_blocks/list_collector_repeating_blocks.spec.js
berroar, line 1: May be worth converting this schema to hub in order to functionally test the completion status that is displayed on the hub when adding/editing items?

Also minor but perhaps worth adding a helper method to answer some of the questions to reduce some of the repetition in this file?

schemas/test/en/test_list_collector_repeating_blocks.json
```diff
@@ -0,0 +1,227 @@
...
+                        {
+                            "id": "list-collector-block",
+                            "type": "ListCollector",
+                            "for_list": "repeating_blocks_list_collector",
```
berroar, line 69: Really minor but generally we use `-` rather than `_` for fields like this

schemas/test/en/test_list_collector_repeating_blocks.json
```diff
@@ -0,0 +1,227 @@
...
+        {
+            "name": "period_id",
+            "type": "string"
+        },
+        {
+            "name": "ru_name",
+            "type": "string"
+        }
+    ],
+    "answer_codes": [
```
berroar, line 24: I'm not sure the answer_codes are behaving as expected. When I test functionally even though I have answered the questions for repeating blocks they don't appear to be in the submission payload.

Looks like this will need a fix and some tests

berroar (changes requested):

schemas/test/en/test_list_collector_repeating_blocks.json
kylelawsonAND: Great catch thanks Rhys! I had only been checking the debug dump, didn't realise those values weren't being added to the submission

schemas/test/en/test_list_collector_repeating_blocks.json
kylelawsonAND: Fixed 👍

tests/functional/spec/features/repeating_blocks/list_collector_repeating_blocks.spec.js
kylelawsonAND: Added a hub schema and added a func test to verify incomplete repeating blocks prevent section completing and exitting to hub.

app/data_models/progress_store.py
```diff
@@ -62,10 +70,10 @@ def _build_map(section_progress_list: Iterable[ProgressDictType]) -> MutableMapp
 
         return {
             (
-                section_progress["section_id"],
-                section_progress.get("list_item_id"),
-            ): Progress.from_dict(section_progress)
-            for section_progress in section_progress_list
+                progress["section_id"],
```
petechd, line 73: Is this not less readable now? I mean, we still track section progress (and this will be less obvious). I still need to check this code some more but are list items (their progress) part of section and its progress? This name sort of suggests it's an instance of Progress class? Do other people have an opinion?

kylelawsonAND: We'd had a discussion about naming [here](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1106#discussion_r1234155990). I'm happy with it being very verbose e.g. `section_or_list_item_progress`, or brief and generic `progress`, with the docs throughout the ProgressStore explaining it is not for only sections, it is for sections, repeating sections and list items.

kylelawsonAND: Regarding "are list items (their progress) part of section and its progress?":

List item progress is to track if a list item with repeating blocks has been fully completed or partially completed. 
A list item with list item id "item1" from a list collector within section "section1", will be keyed in the ProgressStore as `("section1", "item1")`. The `Progress` object for the `_progress` entry keyed by `("section1", "item1")`, will have the `block_ids` attribute populated by the completed repeating block ids for this list item. Its `status` attribute will be `CompletionStatus.IN_PROGRESS` if the list item has been added via the list collector's add block, but not all repeating blocks have been completed and added to the `block_ids` attribute of the relevant Progress object. The `status` attribute will be `CompletionStatus.COMPLETED` when all repeating blocks have been completed and their ids added to the `block_ids` attribute of the relevant Progress object.

Section and Repeating Section progress works as it always has - there is no conflict in using progress store to track the list item progress. 

All of the renaming of methods and variables has been made to make it more inclusive of the elements for which it tracks progress.

petechd: Regarding `section_or_list_item_progress` the problem here is list item is an item from a list in a list store, it's not meant to have progress. Maybe it will be better if we use `repeating blocks`?

katie-gardner: I like the clarity of it being exactly whats stored in it.. `section_or_repeating_blocks_progress`?

kylelawsonAND: I think we need to introduce a concept of List Items being complete or incomplete now, because routing out of List Collectors and rendering now depend on know if the List Items are complete (have their repeating blocks been completed) So the repeating blocks are the metric by which a List Item is deemed complete, just like how sections are already, they track completeness by which blocks within them have been completed.

katie-gardner: Screenshot 2023-06-21 at 13 06 14
I get the dilemma, if you for some reason had repeating sections that used the same list, that list item id appears in a key for both the repeating blocks, and that latter section, so I wouldn't say the list item is done as such, but the repeating blocks are

katie-gardner: so it could have the additional context of `section_or_list_item_repeating_blocks_progress`

katie-gardner: we could always roll with it as is and change it later

kylelawsonAND: renamed to section_or_repeating_blocks as result of vote 👍

app/data_models/progress_store.py
```diff
@@ -76,91 +94,130 @@ def is_dirty(self) -> bool:
...
+            return self._section_and_repeating_blocks_progress[progress_key].status
 
         return CompletionStatus.NOT_STARTED
 
     def get_block_status(
         self, *, block_id: str, section_id: str, list_item_id: str | None = None
     ) -> str:
-        section_blocks = self.get_completed_block_ids(
+        """
+        Return the completion status of the block specified by th given block_id,
```
berroar, line 191:
```suggestion
        Return the completion status of the block specified by the given block_id,
```

app/data_models/progress_store.py
```diff
@@ -20,34 +20,52 @@ def __iter__(self) -> Iterator[tuple[str]]:
...
+        self,
+        in_progress_sections_and_repeating_blocks: Iterable[ProgressDictType]
+        | None = None,
     ) -> None:
         """
-        Instantiate a ProgressStore object that tracks the status of sections and its completed blocks
+        Instantiate a ProgressStore object that tracks the progress status of Sections & Repeating Sections,
+        and their completed blocks, as well as Repeating Blocks for List Items.
+            - Standard Sections are keyed by Section ID, and a None List Item ID
+            - Repeating Sections (dynamic Sections created for List Items that have been added to a List Collector)
```
berroar, line 36: Is this the wrong way around, it is the list item id that is randomly generated 😅?

app/jinja_filters.py
petechd: Worth confirming with Mebin.

berroar (approved):
Retested functionally - LGTM! 🚀

app/data_models/progress_store.py
kylelawsonAND: Have reworded, mistakingly thought section id would be randomly generated for each list item but it's predetermined in schema and same for each list item

app/data_models/progress_store.py
kylelawsonAND: fixed

app/data_models/progress_store.py
```diff
@@ -20,34 +20,52 @@ def __iter__(self) -> Iterator[tuple[str]]:
...
+            self,
+            in_progress_sections_and_repeating_blocks: Iterable[ProgressDictType]
+                                                       | None = None,
     ) -> None:
         """
-        Instantiate a ProgressStore object that tracks the status of sections and its completed blocks
+        Instantiate a ProgressStore object that tracks the progress status of Sections & Repeating Sections,
+        and their completed blocks, as well as Repeating Blocks for List Items.
+            - Standard Sections are keyed by Section ID, and a None List Item ID
+            - Repeating Sections (dynamic Sections created for List Items that have been added to a List Collector)
```
petechd, line 36: Non blocker, I don't want to be annoying here, but we don't add list items to list collector.
```suggestion
            - Repeating Sections (dynamic Sections created for List Items that have been added using a List Collector)
```

schemas/test/en/test_list_collector_repeating_blocks.json
```diff
@@ -0,0 +1,227 @@
...
+                                    "id": "repeating-block-1",
+                                    "type": "ListRepeatingQuestion",
+                                    "question": {
+                                        "id": "repeating-block-1-question",
+                                        "type": "General",
+                                        "title": "This is the question for repeating block 1, which relates to an item that was added to the list collector.",
+                                        "answers": [
+                                            {
+                                                "id": "repeating-block-1-question-answer-1",
+                                                "label": "Answer 1",
```
petechd, line 123: Should have some description otherwise you hit validation errors cause you don't know what answer type is expected.

schemas/test/en/test_list_collector_repeating_blocks.json
```diff
@@ -0,0 +1,227 @@
+{
```
petechd, line 1: I am not sure if these schemas should be so abstract if we modelled them after prototypes that were based on real life surveys in all previous iterations of looping.

kylelawsonAND: Yeah I agree, I regretted making it the way I did afterwards. I had been using `test_list_collector_repeating_blcoks_section_summary` as a the functional and manual testing schema, and made this one for unit test purposes. 
Won't do it again!

app/data_models/progress_store.py
kylelawsonAND: Sorted 👍

schemas/test/en/test_list_collector_repeating_blocks.json
```diff
@@ -0,0 +1,227 @@
...
+                    "title": "List",
+                    "blocks": [
+                        {
+                            "id": "list-collector-block",
+                            "type": "ListCollector",
+                            "for_list": "repeating-blocks-list-collector",
+                            "question": {
+                                "id": "list-collector-next-question",
+                                "type": "General",
+                                "title": "This List Collector demonstrates a collector with Repeating Blocks which are asked after adding an item, and that must be completed before navigated past the list collector.\nDo you need to add any more to this list?",
```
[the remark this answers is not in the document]
kylelawsonAND: Sorted 👍

schemas/test/en/test_list_collector_repeating_blocks.json
```diff
@@ -0,0 +1,227 @@
...
+                                    }
+                                ]
+                            },
+                            "add_block": {
+                                "id": "list-collector-add-block",
+                                "type": "ListAddQuestion",
+                                "question": {
+                                    "id": "list-collector-add-block-question",
+                                    "type": "General",
+                                    "title": "What is the name of the list item? This item will be followed up by 2 pages of related questions (repeating blocks)",
```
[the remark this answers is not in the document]
kylelawsonAND: Sorted 👍

schemas/test/en/test_list_collector_repeating_blocks.json
kylelawsonAND: Sorted 👍

app/data_models/progress_store.py
```diff
@@ -76,91 +94,130 @@ def is_dirty(self) -> bool:
...
+        """
+        Return the completion status of the block specified by the given block_id,
+        if it is part of the progress of the given Section or Repeating Blocks for list item
+        specified by the given section_id or list_item_id
+        """
+        section_or_list_item_blocks = self.get_completed_block_ids(
             section_id=section_id, list_item_id=list_item_id
         )
-        if block_id in section_blocks:
+        if block_id in section_or_list_item_blocks:
```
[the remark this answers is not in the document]
kylelawsonAND: Sorted 👍

schemas/test/en/test_list_collector_repeating_blocks.json
petechd: I see, do we need both schemas? Can we not do all the testing with only one and one for the hub?

petechd: We had to investigate and delete duplicated/bloated schemas before, so just want to avoid that in future.

app/views/contexts/summary/list_collector_block.py
```diff
@@ -160,20 +163,27 @@ def _add_link(
...
             section_id=section_id, list_name=list_model.name
         )
-        if not related_answers:
+
+        blocks: list[dict] = []
+
+        if related_answers:
+            blocks += self._get_blocks_for_related_answers(related_answers)
+
+        if list_model:
```
petechd, line 179: Does list_model not always exist if you always pass a list_model with type ListModel into this method?

app/views/contexts/summary/list_collector_block.py
```diff
@@ -160,20 +163,27 @@ def _add_link(
...
+        self, list_model: ListModel, repeating_blocks: Sequence[ImmutableDict]
+    ) -> dict[str, list[dict]] | None:
         section_id = self._section["id"]
 
         related_answers = self._schema.get_related_answers_for_list_for_section(
             section_id=section_id, list_name=list_model.name
         )
-        if not related_answers:
+
+        blocks: list[dict] = []
```
petechd, line 174: Wrong type here as it is empty list?

app/views/handlers/list_add_question.py
```diff
@@ -18,7 +37,7 @@ def get_next_location_url(self):
 
     def handle_post(self):
         # Ensure the section is in progress when user adds an item
-        list_item_id = self.questionnaire_store_updater.add_list_item(
+        self._list_item_id = self.questionnaire_store_updater.add_list_item(
```
petechd, line 40: I am not sure about reassigning the value of self._list_item_id, can we keep it as it was?

app/views/contexts/summary/list_collector_block.py
katie-gardner: its initialised empty but then populated with `_get_blocks_for_related_answers` or `repeating_blocks` so specifying a type is correct, though if repeating blocks are indeed immutable dictionaries then this would be a list of `dict | ImmutableDict`? but I think the type for `repeating_blocks` might be the incorrect one, rather than this

app/views/contexts/summary/list_collector_block.py
kylelawsonAND: Default behaviour when `__len__` is implemented, but `__bool__` is not, is to evaluate as `len(list_model) != 0`, which what we're checking, if there are any list items. 
https://blog.finxter.com/python-__len__-magic-method/ - can't find the PEP standard for it though. 

I'll change to `len(list_model) != 0` to be more clear

app/views/contexts/summary/list_collector_block.py
kylelawsonAND: Yep will change to list[ImmutableDict], its only initialised empty.

app/views/handlers/list_add_question.py
kylelawsonAND: We need an attribute to capture the generated list item id as it is needed in the `get_next_location_url method`

schemas/test/en/test_list_collector_repeating_blocks.json
kylelawsonAND: Understand, will change the schema to be less abstract instead of removing as I think it's still beneficial to have a schema of only the list collector with RB + means not having to alter tests as much and getting this merged quicker to unblock

kylelawsonAND: Added context to schema to make it less abstract 👍

app/views/contexts/summary/list_collector_block.py
kylelawsonAND: Done 👍

app/views/contexts/summary/list_collector_block.py
kylelawsonAND: done 👍

schemas/test/en/test_list_collector_repeating_blocks.json
```diff
@@ -0,0 +1,258 @@
...
+                    "answers": [
+                      {
+                        "id": "repeating-block-1-question-answer-1",
+                        "label": "How many grams of this item do you need?",
+                        "mandatory": true,
+                        "type": "Number"
+                      },
+                      {
+                        "id": "repeating-block-1-question-answer-2",
+                        "label": "How much  are you willing to spend on this item?",
```
katie-gardner, line 153: Changes to this are great, only (extremely minor) suggestion would be to change type of this one to currency

schemas/test/en/test_list_collector_repeating_blocks.json
petechd: What is the difference between these two schemas except extra question block after list collector? If it's duplicated schema we shouldn't keep both if they are essentially the same.

schemas/test/en/test_list_collector_repeating_blocks.json
kylelawsonAND: Changed one to grams and one to currency 👍

schemas/test/en/test_list_collector_repeating_blocks.json
kylelawsonAND: The difference is one has a section summary and one doesn't

katie-gardner: I'd be in favour of keeping it - seems like a good idea to have a really simple one that isolates the new feature

petechd (approved):

Yuyuutsu (approved):

PR #1107, opened 2023-05-16
Feat/Grand calculated summary

katie-gardner:
Implements the ability to sum up multiple calculated summaries as part of a grand calculated summary. This is done by adding an additional GrandCalculatedSummary block type, which expects a calculation using only calculated summary sources.

Additionally this makes a minor modification to routing allowing for `return_to` and `return_to_block_id` to be lists so that it is possible to use the change link on a calculated summary, and then on an answer, and still be taken back to the grand calculated summary.

The other change to routing is that for the next/submit button when going between calculated summaries / their answers, complete blocks are skipped. The expected behaviour is that if you have a return to location in the url that you cannot reach yet, because there is an incomplete block prior to that location, you should be routed to the next incomplete block of the section, until you can reach the return location, at which point you go there. This is only true when pressing next, if you click previous, you will either go to the return to location, or if that is inaccessible, the previous block in the section and the params for the url are preserved.

### Points to note

- Currently if the answers on the path for a calculated summary change, but the answer values themselves do not, then the progress of the calculated summary doesn't update. To see this:
  1. use `test_new_calculated_summary_cross_section_dependencies.json`
  2. complete up to the calculated summary
  3. then go back and edit your response to 'skip first question' without changing an answer. 
  4. If you go back to the section Hub the calculated summary will still show as completed. 

  Grand calculated summaries will behave the same, and it has been confirmed that changing this is out of scope and will be done in a follow up if needed in future. The current behaviour has been covered in a new functional test.

### How to review
Check that the approach is consistent with the implementation of calculated summaries, and that validation is also consistent with calculated summaries. Read through the proposal and check all scenarios have been covered (including those relying on progress), and are working as expected. Test schemas added:

- `test_grand_calculated_summary`
- `test_grand_calculated_summary_multiple_sections`
- `test_grand_calculated_summary_cross_section_dependencies`
- `test_grand_calculated_summary_overlapping_answers`

### How to review functionally
the `test_grand_calculated_summary_cross_section_dependencies` has a skippable question, skippable calculated summary and a question which is dependant on the answer of another. Check the unhappy path of making changes to the inclusion of the first question and calculated summary, and changing the answer to the third question to add in an additional question. Routing should be as outlined in the scenarios on the ticket and the grand calculated summary section should revert to 'partially completed' when changes occur.

`test_grand_calculated_summary` sums up calculated summaries in multiple different sections. Check that the change links on the grand calculated summary and calculated summary take you back to the correct block and that you are able to edit multiple answers and be taken back to the grand calculated summary upon confirming the calculated summary.

`test_grand_calculated_summary_overlapping_answers` tests summing up calculated summaries which reuse answers. Check that changing which calculated summary is included updates the progress of the grand calculated summary and that if there are two calculated summaries sharing answers, one of which is on the path and one that isn't, that the grand calculated summary includes the correct values. It should allow for multiple calculated summaries to use the same answer. The grand calculated summary section is enabled on completion of section-1. Check this works as expected

For the overlapping schema, it is also important to test the functionality of the change links for calculated summaries with shared answers. In particular, check that if you select 'extra bread and cheese' that when you change these answers from either of the calculated summary change links on the grand calculated summary page, that you are taken to the first calculated summary depending on the answer, then the second, and then the grand calculated summary without any other completed blocks being visited

### Checklist
* [ ] suitable unit tests, integration tests and functional tests have been added
* [ ] 100% test coverage of the new feature
* [ ] tests all passing
* [ ] all scenarios in the google sheet have been covered
* [ ] no missing validation 


berroar:
Also a query about the expected behaviour for the overlapping schema - If i complete the first section, then from the Grand Summary section, I click to change the first calculated summary total, then from the calculated summary page I click to change the first answer, I am shown the dependent questions and asked to reconfirm the grand calculated summary total, but section one (Charity donations) then shows as being only partially complete on the hub page - Should the dependent questions/summaries for that section have been completed before returning to the grand calculated summary?

tests/functional/spec/features/grand_calculated_summary/grand_calculated_summary_cross_section_dependencies.spec.js
```diff
@@ -0,0 +1,123 @@
...
+      await $(CurrencyAllPage.submit()).click();
+    });
+    it("Given I go back and skip the second calculated summary, it is not included in the grand calculated summary", async () => {
+      await $(HubPage.summaryRowLink("calculated-summary-section")).click();
+      await $(CalculatedSummarySectionSummaryPage.skipAnswer2Edit()).click();
+      await $(SkipCalculatedSummaryPage.yes()).click();
+      await $(SkipCalculatedSummaryPage.submit()).click();
+      await $(CalculatedSummarySectionSummaryPage.submit()).click();
+      // Currently the grand calculated summary remains 'Completed' because none of the answers have changed
+      await expect(await $(HubPage.summaryRowState("grand-calculated-summary-section")).getText()).to.equal("Completed");
```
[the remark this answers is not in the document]
katie-gardner: I'd asked Mebin this too and he's confirming with Donna but currently calculated summaries don't do this (e.g. on the `test_calculated_summary_cross_section_dependencies.json` if you change that 'skip first block' without changing any values, the calculated summary remains complete)

We concluded that if it is needed then we'll have a separate task as a follow up to handle updating progress in the case of no answers changing but the path changing. I've just put this in the test so we're aware of how it works currently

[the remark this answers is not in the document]
katie-gardner: Got confirmation that this is out of scope - PR description updated

app/jinja_filters.py
```diff
@@ -606,7 +621,7 @@ def map_summary_item_config(
 
             rows.extend(list_collector_rows)
 
-    if summary_type == "CalculatedSummary":
+    if summary_type in ("CalculatedSummary", "GrandCalculatedSummary"):
```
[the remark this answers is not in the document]
katie-gardner: Changed

petechd:
To fix the complexity error you could do something like:
```python
            if block["type"] in ["CalculatedSummary", "GrandCalculatedSummary"]:
                self._update_answer_dependencies_for_summary(block)
                continue

            for question in self.get_all_questions_for_block(block):
                if question["type"] == "Calculated":
                    self._update_answer_dependencies_for_calculations(
                        question["calculations"], block_id=block["id"]
                    )
                    continue

                if dynamic_answers := question.get("dynamic_answers"):
                    for answer in dynamic_answers["answers"]:
                        value_source = dynamic_answers["values"]
                            value_source, block_id=block["id"], answer_id=answer["id"]
                        )

                        self._dynamic_answer_ids.add(answer["id"])

                        self._update_answer_dependencies_for_answer(
                            answer, block_id=block["id"]
                        )

                for answer in question.get("answers", []):
                    self._update_answer_dependencies_for_answer(
                        answer, block_id=block["id"]
                    )
                        if "detail_answer" in option:
                            self._update_answer_dependencies_for_answer(
                                option["detail_answer"], block_id=block["id"]
                            )
    def _update_answer_dependencies_for_summary(self, block: ImmutableDict) -> None:
        if block["type"] == "CalculatedSummary":
            answer_ids_for_block = get_calculated_summary_answer_ids(block)
            self._update_answer_dependencies_for_calculated_summary(
                answer_ids_for_block, block["id"]
            )
        else:
            calculated_summary_ids = get_grand_calculated_summary_block_ids(block)
            self._update_answer_dependencies_for_grand_calculated_summary(
                calculated_summary_ids, block["id"]
            )
```

app/questionnaire/router.py
```diff
@@ -224,22 +224,26 @@ def _get_return_to_location_url(
...
-            return url_for(
-                "questionnaire.block",
-                block_id=return_to_block_id,
-                list_name=location.list_name,
-                list_item_id=location.list_item_id,
-                return_to=return_to,
-                _anchor=return_to_answer_id,
-            )
+        if return_to == "grand-calculated-summary":
+            if url := self._get_return_to_for_grand_calculated_summary(
```
MebinAbraham, line 228: Once the progress source merges, worth updating a schema to use that. We need to check what happens if grand summary become inaccessible.
On another note, the nested if can be both single line statements?. Applies to the one below also.

```py
if return_to == "something" and url := get_url():
  return url
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -1018,19 +1063,33 @@ def get_sources_for_type_from_data(
...
+        source_type=source_type, data=calculation_block["calculation"]["operation"]
+    )
+
+    return [value["identifier"] for value in values]
+
+
 def get_calculated_summary_answer_ids(calculated_summary_block: Mapping) -> list[str]:
     if calculated_summary_block["calculation"].get("answers_to_calculate"):
-        return calculated_summary_block["calculation"]["answers_to_calculate"]  # type: ignore
+        return list(calculated_summary_block["calculation"]["answers_to_calculate"])
```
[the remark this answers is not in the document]
katie-gardner: It's actually coming out as a tuple at runtime, the sections come from the json property which returns the serialized immutable version of questionnaire json. It seemed slightly nicer than changing the type hints to cover that case, but could always remove the cast and make the other function return a tuple as well

app/questionnaire/questionnaire_schema.py
```diff
@@ -1018,19 +1063,33 @@ def get_sources_for_type_from_data(
...
 
-    values = get_mappings_with_key(
-        "source", calculated_summary_block["calculation"]["operation"]
-    )
+    return get_identifiers_from_calculation_block(calculated_summary_block, "answers")
 
-    return [value["identifier"] for value in values if value["source"] == "answers"]
+
+def get_grand_calculated_summary_block_ids(
```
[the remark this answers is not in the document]
katie-gardner: I've gone for ultimate clarity with `get_calculated_summary_ids_for_grand_calculated_summary`

app/questionnaire/router.py
katie-gardner: Its currently possible to test what happens when it becomes inaccessible in the `cross_section_dependencies` schema, the grand calculated summary depends on providing an answer to part b of question 3 - that said, I haven't made that clear, so I'll update that schema to make it better. I'll add a functional test that checks routing for this case and once progress is in it can just be updated with a slightly different when rule

fixed if statements

app/questionnaire/router.py
```diff
@@ -258,6 +262,86 @@ def _get_return_to_location_url(
...
+        return_to_answer_id: str | None = None,
+    ) -> str | None:
+        """
+        Builds the return url for a grand calculated summary,
+        and accounts for it possibly being in a different section to the calculated summaries it references
+        """
+        # Type ignore: the section for the grand calculated summary must exist
+        grand_calculated_summary_section = location.section_id
+        if return_to_block_id:
+            grand_calculated_summary_section: str = self._schema.get_section_id_for_block_id(return_to_block_id)  # type: ignore
```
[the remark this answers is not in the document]
katie-gardner: good catch 👍

app/questionnaire/router.py
```diff
@@ -258,6 +262,86 @@ def _get_return_to_location_url(
...
+        so that both the calculated summary id and the grand calculated summary ids are stored.
+        """
+        # the calculated summary to go back to first
+        block_id = None
+        # the grand calculated summary to go back to second (if applicable)
+        return_to_block_id = None
+
+        if return_to_block_ids:
+            # the first block is the block id to route to, and whatever is left (if anything) forms where to go next
+            block_id = return_to_block_ids.pop(0)
```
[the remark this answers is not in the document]
katie-gardner: Fixed

app/questionnaire/value_source_resolver.py
```diff
@@ -128,12 +136,17 @@ def _resolve_list_value_source(self, value_source: Mapping) -> int | str | list:
...
 
-        The caller is responsible for ensuring the provided Calculated Summary and its answers are on the path.
+        The caller is responsible for ensuring the provided Calculated Summary and its answers are on the path,
+        or providing routing_path_block_ids when initialising the value source resolver.
         """
-        calculated_summary_block: Mapping = self.schema.get_block(value_source["identifier"])  # type: ignore
+        calculated_summary_block: ImmutableDict = self.schema.get_block(value_source["identifier"])  # type: ignore
+
+        if not self._is_block_on_path(calculated_summary_block):
+            return None
```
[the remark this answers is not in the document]
katie-gardner: in the context class, `_build_grand_calculated_summary_section`, the blocks to calculate might not all be on the path, and I was getting the wrong answer when evaluating the calculation. As it stands the value source resolver ignores answers not on the path, but if I pass it a calculated summary that isn't on the path, with answers that are, it returns the total. 

The filtering could happen in the context class instead but I think it more suitable here - if that makes sense?

app/views/contexts/grand_calculated_summary_context.py
```diff
@@ -0,0 +1,119 @@
...
+    def _build_grand_calculated_summary_section(
+        self, rendered_block: ImmutableDict
+    ) -> dict:
+        """
+        Build list of calculated summary blocks that the grand calculated summary will be adding up
+        """
+        # Type ignore: the block, group and section will all exist at this point
+        block_id: str = self.current_location.block_id  # type: ignore
+        calculated_summary_group: ImmutableDict = self._schema.get_group_for_block_id(block_id)  # type: ignore
+        section_id: str = self._schema.get_section_id_for_block_id(block_id)  # type: ignore
```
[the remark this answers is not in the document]
katie-gardner: You are absolutely right - removed

app/questionnaire/router.py
```diff
@@ -258,6 +262,86 @@ def _get_return_to_location_url(
...
+            return_to_block_id = (
+                ",".join(return_to_block_ids) if return_to_block_ids else None
+            )
```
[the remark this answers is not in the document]
katie-gardner: in the case that `return_to_block_ids` is a list with a single item, the pop(0) makes it falsy so the check wasn't redundant but I've changed it to a starred assignment that I think is clearer

tests/integration/questionnaire/test_questionnaire_grand_calculated_summary.py
```diff
@@ -0,0 +1,140 @@
...
+        self.post()
+        self.post({"radio-funding": radio_answer})
+        if radio_answer != "No":
+            # in the no overlap case, the calculated summary is skipped entirely
+            self.post()
+        self.post()
+        self.post()
+
+    def test_grand_calculated_summary_overlapping_answers_full_overlap(self):
+        self.launchSurvey("test_grand_calculated_summary_overlapping_answers")
```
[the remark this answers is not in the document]
katie-gardner: Was caused by the way I was getting the format in the context class - now that I've moved this to a shared function in questionnaire_schema the issue this should be resolved 👍 
as for being able to access the grand calculated summary before any answers, investigating

app/views/handlers/grand_calculated_summary.py
```diff
@@ -0,0 +1,29 @@
...
+            return_to_block_id=self.return_to_block_id,
+        )
+        context = (
+            grand_calculated_summary_context.build_view_context_for_grand_calculated_summary()
+        )
+
+        if not self.page_title:
+            self.page_title = context["summary"]["calculated_question"]["title"]
+
+        return context
```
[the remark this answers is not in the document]
katie-gardner: great shout 🎉 done

app/questionnaire/questionnaire_schema.py
```diff
@@ -288,11 +297,29 @@ def _populate_answer_dependencies(self) -> None:
     def _update_answer_dependencies_for_calculated_summary(
         self, calculated_summary_answer_ids: Iterable[str], block_id: str
     ) -> None:
+        self._calculated_summary_answer_dependencies[block_id].update(
+            calculated_summary_answer_ids
+        )
```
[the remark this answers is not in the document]
MebinAbraham: Technically it is already cached. It gets it from `self._blocks_by_id` which is just a lookup in a prebuilt dict.

👍

app/questionnaire/router.py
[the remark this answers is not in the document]
MebinAbraham: The stared assigned is cool 👍

app/questionnaire/value_source_resolver.py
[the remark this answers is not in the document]
MebinAbraham: Yup, makes sense. 👍

app/jinja_filters.py
```diff
@@ -463,6 +463,7 @@ def __init__(  # noqa: C901, R0912 pylint: disable=too-complex, too-many-branche
                 multiple_answers
                 or answer_type == "relationship"
                 or summary_type == "CalculatedSummary"
+                or summary_type == "GrandCalculatedSummary"
```
MebinAbraham, line 466: Minor and non-blocking, could use a constant or Enum given we have got the literal in a lot.

app/questionnaire/questionnaire_schema.py
```diff
@@ -293,6 +304,23 @@ def _update_answer_dependencies_for_calculated_summary(
                 self._get_answer_dependent_for_block_id(block_id=block_id)
             }
 
+    def _update_answer_dependencies_for_grand_calculated_summary(
+        self,
+        grand_calculated_summary_calculated_summary_ids: Iterable[str],
+        block_id: str,
+    ) -> None:
+        # validator ensures that any grand calculated summary will come after the calculated summaries it depends on
+        # so the calculated summary dependencies will already be done, and this can just take those
```
MebinAbraham, line 313: This comment no longer applies since you are not creating your own cache?.

app/questionnaire/questionnaire_schema.py
```diff
@@ -293,6 +304,23 @@ def _update_answer_dependencies_for_calculated_summary(
...
+        for calculated_summary_id in grand_calculated_summary_calculated_summary_ids:
+            # Type ignore: safe to assume block exists
+            calculated_summary_block: ImmutableDict = self.get_block(calculated_summary_id)  # type: ignore
+            for answer_id in get_calculated_summary_answer_ids(
+                calculated_summary_block
+            ):
+                self._answer_dependencies_map[answer_id] |= {
+                    self._get_answer_dependent_for_block_id(block_id=block_id)
+                }
```
MebinAbraham, line 322: Instead of duplicating this, I suggest reusing `_update_answer_dependencies_for_calculated_summary` here.

Could even move getting the answer ids for the block into that method.

app/questionnaire/questionnaire_schema.py
```diff
@@ -725,6 +753,26 @@ def get_first_answer_id_for_block(self, block_id: str) -> str:
...
         return answer_ids[0]
 
+    def get_answer_format_for_calculated_summary(
+        self, calculated_summary_block_id: str
+    ) -> dict:
+        """
+        Given a calculated summary block id, find the format of the total by using the first answer
+        """
+        answer_format = {"type": None}
+        if calculated_summary_block := self.get_block(calculated_summary_block_id):
```
MebinAbraham, line 763: Is it possible for this to not return a block?. If not, remove the if and type ignore?

app/questionnaire/router.py
```diff
@@ -224,22 +224,30 @@ def _get_return_to_location_url(
...
+                location=location,
+                routing_path=routing_path,
+                return_to_answer_id=return_to_answer_id,
+            )
+        ):
+            return url
+
+        if return_to.startswith("calculated-summary") and (
+            url := self._get_return_to_for_calculated_summary(
+                # if there are multiple return to block types, the last one is where to go next
```
MebinAbraham, line 240: This is based on the assumption we only support 2 locations right?. Otherwise, it is next in line?.

app/questionnaire/router.py
```diff
@@ -258,6 +266,86 @@ def _get_return_to_location_url(
...
+        grand_calculated_summary_section = location.section_id
+        if return_to_block_id:
+            grand_calculated_summary_section: str = self._schema.get_section_id_for_block_id(return_to_block_id)  # type: ignore
+        if grand_calculated_summary_section != location.section_id:
+            # the grand calculated summary is in a different section which will have a different routing path
+            routing_path = self._path_finder.routing_path(
+                section_id=grand_calculated_summary_section
+            )
+        if self.can_access_location(
+            # grand calculated summaries don't support repeating sections so no need for list item here
```
MebinAbraham, line 292: Is this true?. It is true that a grand summary won't reference calc summary that is in a repeating section unless the grand summary is in a repeating section itself.

app/views/contexts/grand_calculated_summary_context.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+        routing_path_block_ids = self._blocks_on_routing_path(calculated_summary_ids)
+
+        calculated_section = self._build_grand_calculated_summary_section(block)
+
+        groups = self.build_groups_for_section(
+            calculated_section, block_id, routing_path_block_ids
+        )
+        total = self._get_evaluated_total(
+            calculation["operation"], routing_path_block_ids
```
MebinAbraham, line 107: Would benefit from keyword args.

app/questionnaire/router.py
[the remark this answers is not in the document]
katie-gardner: Added a test to `test_router`, should now behave similarly to the calculated summary behaviour when making url invalid

app/questionnaire/questionnaire_schema.py
katie-gardner: Updated - it would only not return a block if the function was called with an invalid calculated summary id, but that should never happen because its only used in the context builder where the ids are definitely valid.

katie-gardner:
> Also a query about the expected behaviour for the overlapping schema - If i complete the first section, then from the Grand Summary section, I click to change the first calculated summary total, then from the calculated summary page I click to change the first answer, I am shown the dependent questions and asked to reconfirm the grand calculated summary total, but section one (Charity donations) then shows as being only partially complete on the hub page - Should the dependent questions/summaries for that section have been completed before returning to the grand calculated summary?

This was a good spot - basically the problem is, lets say you have:

- q1
- q2
- calculated summary A = q1 + q2
- q3
- calculated summary B = q2 + q3
- grand calculated summary

From the grand calculated summary

- If you edit calculated summary A and change q2
  - router will check if you can return to calculated summary A, which you can, so you go back there. 
  - Then it will take you back to the grand calculated summary as I had not accounted for this
- However if you edit calculated summary B and change q2
  - router will check if you can return to calculated summary B, which you can't, because calculated summary A is in the same section, and isn't complete, so calculated summary B is not yet an accessible location.
  - As a result you'll be take to every single block between q2 and calculated summary B in turn, then once confirmed you'll be taken back to the grand calculated summary

What should have happened in both of these cases, is that regardless of which link you edit q2 from, you are taken to calculated summary A, then calculated summary B and then the grand calculated summary

I've added a functional test for this scenario and redone the overlapping answers schema as it was getting confusing and I'm hoping this is now much clearer for testing.

app/views/contexts/grand_calculated_summary_context.py
katie-gardner: added 👍

app/questionnaire/questionnaire_schema.py
katie-gardner: added the dependent block to that function so it can be reused 👍

app/jinja_filters.py
katie-gardner: gone with a function instead since the summary type check is always against both - arguably I could have just made the type of the block a CalculatedSummary for both but its probably clearer to keep it separate

app/questionnaire/router.py
katie-gardner: Updated to make it more generic, ie next in line as you say, leaves scope to use it for more items later on

app/questionnaire/router.py
katie-gardner: So I could add list_item_id as the section and list item will be consistent in the case of a repeating section, but as theres a follow up ticket to handle grand calculated summaries with repeating calculated summaries I think it's best I leave as is? I've updated the comment

katie-gardner: ahhh I see - I misunderstood, I think a new card would be good as in that case there should be a test schema for a grand calculated summary inside a repeat and new sets of tests to cover this scenario

katie-gardner: cheers Mebin 👍

MebinAbraham: Will be done in https://trello.com/c/gP4302Vl/5733-grand-calculated-summary-support-within-repeating-sections

tests/integration/questionnaire/test_questionnaire_grand_calculated_summary.py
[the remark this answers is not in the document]
katie-gardner: Once progress value source is in this schema can be updated to only show the GCS once the calculated summaries are done, until then, (unless I'm mistaken) I think theres little that can be done about the GCS being available with no answers, its not like calculated summaries where its in the same section and you can be made to do the incomplete ones first. 
500 fixed though

app/questionnaire/questionnaire_schema.py
```diff
@@ -248,11 +248,8 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
 
     def _populate_answer_dependencies(self) -> None:
         for block in self.get_blocks():
-            if block["type"] == "CalculatedSummary":
-                answer_ids_for_block = get_calculated_summary_answer_ids(block)
-                self._update_answer_dependencies_for_calculated_summary(
-                    answer_ids_for_block, block["id"]
-                )
+            if block["type"] in {"CalculatedSummary", "GrandCalculatedSummary"}:
```
petechd, line 251: Here we should reuse already existing `is_summary_with_calculation()`  passing block["type"] as an argument.

app/views/contexts/calculated_summary_context.py
```diff
@@ -253,9 +288,7 @@ def _get_answer_format(self, groups: list) -> Tuple[dict[str, Any], list]:
         return answer_format, values_to_calculate
 
     @staticmethod
-    def _format_total(
-        answer_format: Mapping[str, str], total: Union[int, float, Decimal]
-    ) -> str:
+    def _format_total(answer_format: Mapping[str, str], total: NumericType) -> str:
```
petechd, line 291: Since we are refactoring these we should probably force positional arguments. Same below.

app/views/contexts/grand_calculated_summary_context.py
```diff
@@ -0,0 +1,126 @@
...
+    get_calculation_block_ids_for_grand_calculated_summary,
+)
+from app.views.contexts.calculated_summary_context import CalculatedSummaryContext
+from app.views.contexts.summary.group import Group
+
+
+class GrandCalculatedSummaryContext(CalculatedSummaryContext):
+    def _build_grand_calculated_summary_section(
+        self, rendered_block: ImmutableDict
+    ) -> dict:
```
petechd, line 15: Here we need top level for the dict.

app/views/contexts/grand_calculated_summary_context.py
```diff
@@ -0,0 +1,126 @@
...
+
+        groups = self.build_groups_for_section(
+            section=calculated_section,
+            return_to_block_id=block_id,
+            routing_path_block_ids=routing_path_block_ids,
+        )
+        total = self._get_evaluated_total(
+            calculation=calculation["operation"],
+            routing_path_block_ids=routing_path_block_ids,
+        )
```
petechd, line 112: This way you get one type ignore less, which is probably better, then extract id from created `block` (may fail linting):
```suggestion
        block: ImmutableDict = self._schema.get_block(self.current_location.block_id)  # type: ignore

        calculation = block["calculation"]
        calculated_summary_ids = get_calculation_block_ids_for_grand_calculated_summary(
            block
        )
        routing_path_block_ids = self._blocks_on_routing_path(calculated_summary_ids)

        calculated_section = self._build_grand_calculated_summary_section(block)

        groups = self.build_groups_for_section(
            section=calculated_section,
            return_to_block_id=block["id"],
            routing_path_block_ids=routing_path_block_ids,
        )
        total = self._get_evaluated_total(
            calculation=calculation["operation"],
            routing_path_block_ids=routing_path_block_ids,
        )
```

app/views/contexts/grand_calculated_summary_context.py
```diff
@@ -0,0 +1,126 @@
...
+class GrandCalculatedSummaryContext(CalculatedSummaryContext):
+    def _build_grand_calculated_summary_section(
+        self, rendered_block: ImmutableDict
+    ) -> dict:
+        """
+        Build list of calculated summary blocks that the grand calculated summary will be adding up
+        """
+        # Type ignore: the block, group and section will all exist at this point
+        block_id: str = self.current_location.block_id  # type: ignore
+        calculated_summary_group: ImmutableDict = self._schema.get_group_for_block_id(block_id)  # type: ignore
```
petechd, line 21: Better to have only one type ignore here?
```suggestion
        calculated_summary_group: ImmutableDict = self._schema.get_group_for_block_id(self.current_location.block_id)  # type: ignore
```

app/views/contexts/grand_calculated_summary_context.py
katie-gardner: I need the block id for line 105 as well so I've left this one be, but changed your other suggestion 👍

app/views/contexts/grand_calculated_summary_context.py
katie-gardner: done 👍

app/views/contexts/grand_calculated_summary_context.py
katie-gardner: added for return type but rendered_block would be str | Any so left that as is

app/views/contexts/calculated_summary_context.py
katie-gardner: good shout 👍

app/questionnaire/questionnaire_schema.py
katie-gardner: moved the method into questionnaire_schema and imported it in the filters so it can be used here too

app/views/contexts/grand_calculated_summary_context.py
katie-gardner: if I use that though I'll need to type ignore twice instead of once?

templates/calculatedsummary.html
```diff
@@ -1,34 +1,5 @@
...
-      "attributes": {
-        "data-qa": "btn-submit",
-        "data-ga-category": "Submit button",
-        "data-ga-action": "Confirm",
-        "data-ga-label": "Confirm button click"
-      }
-    })
-  }}
+{% block form_title %}
+  <h1 data-qa="calculated-summary-title">{{ content.summary.title }}</h1>
```
petechd, line 4: We could possibly use one template and use existing `content.summary.summary_type` for the logic but then, is using one template better?

app/views/contexts/grand_calculated_summary_context.py
katie-gardner: updated 👍

templates/calculatedsummary.html
katie-gardner: the summary type would be CalculatedSummary or GrandCalculatedSummary so it might be a bit ugly to get the data-qa that way, unless theres another one?

app/questionnaire/questionnaire_schema.py
```diff
@@ -285,14 +282,45 @@ def _populate_answer_dependencies(self) -> None:
...
                             )
 
-    def _update_answer_dependencies_for_calculated_summary(
-        self, calculated_summary_answer_ids: Iterable[str], block_id: str
+    def _update_answer_dependencies_for_summary(self, block: ImmutableDict) -> None:
+        if block["type"] == "CalculatedSummary":
+            self._update_answer_dependencies_for_calculated_summary_dependency(
+                calculated_summary_block=block, dependent_block=block
+            )
+        else:  # must be GrandCalculatedSummary
```
kylelawsonAND, line 290: I think it'd be worth guarding against misuse of this function by 
`if block["type"] == "GrandCalculatedSummary":`
and having nothing else happen if it's something else, or even throw an exception but do nothing seems more in line with rest of the QuestionnaireSchema.

app/questionnaire/questionnaire_schema.py
```diff
@@ -285,14 +282,45 @@ def _populate_answer_dependencies(self) -> None:
...
+            self._update_answer_dependencies_for_grand_calculated_summary(block)
+
+    def _update_answer_dependencies_for_calculated_summary_dependency(
+        self, *, calculated_summary_block: ImmutableDict, dependent_block: ImmutableDict
     ) -> None:
+        """
+        Given a calculated summary block and a block which depends on that calculated summary (may just be the calculated summary itself)
+        update the answer to be a dependency for the dependent block
```
kylelawsonAND, line 298: If I'm understanding the function correctly this should be:

"update all calculated summary answers to be dependencies of the dependent block"

app/questionnaire/questionnaire_schema.py
```diff
@@ -725,6 +753,23 @@ def get_first_answer_id_for_block(self, block_id: str) -> str:
...
+        self, calculated_summary_block_id: str
+    ) -> dict:
+        """
+        Given a calculated summary block id, find the format of the total by using the first answer
+        """
+        # Type ignore: the block will exist for any valid calculated summary id
+        calculated_summary_block: ImmutableDict = self.get_block(calculated_summary_block_id)  # type: ignore
+        first_answer_id = get_calculated_summary_answer_ids(calculated_summary_block)[0]
+        first_answer = self.get_answers_by_answer_id(first_answer_id)[0]
+        return {
```
kylelawsonAND, line 766: Is it valid for an integer to be summed with a decimal for a grand calculation? Or will the calculated summaries always be same type and thus this is safe every time?

kylelawsonAND (changes requested):

app/questionnaire/questionnaire_schema.py
katie-gardner: the calculated summaries will always be the same type, currency unit, etc, and the GCS re-uses the validation for answer types, so it would never sum up a unit and a currency, but it does support an integer with decimal places, just like calculated summaries does

app/questionnaire/questionnaire_schema.py
katie-gardner: You are indeed understanding and thanks, thats a better more concise comment - updated

app/questionnaire/questionnaire_schema.py
katie-gardner: happy to make that change, safe and difference is negligible

berroar (commented):
Couldn't find any issues testing functionally 👍

app/questionnaire/router.py
```diff
@@ -258,6 +268,121 @@ def _get_return_to_location_url(
...
+        is_for_previous: bool,
+        return_to_answer_id: str | None = None,
+    ) -> str | None:
+        """
+        Builds the return url for a grand calculated summary,
+        and accounts for it possibly being in a different section to the calculated summaries it references
+        """
+        grand_calculated_summary_section = location.section_id
+        # if return_to_block_id is invalid and no block is found, the section doesn't update but can_access_location will be false, so it still returns None
+        if return_to_block_id and self._schema.get_block(return_to_block_id):
```
MebinAbraham, line 287: For clarity use `self._schema.is_block_valid`

app/questionnaire/router.py
```diff
@@ -258,6 +268,121 @@ def _get_return_to_location_url(
...
+            # Type ignore: if the block is valid, then we'll be able to find a section for it
+            grand_calculated_summary_section: str = self._schema.get_section_id_for_block_id(return_to_block_id)  # type: ignore
+        if grand_calculated_summary_section != location.section_id:
```
MebinAbraham, line 290: If the first if fails, then you should return None. At the moment, it can go into this if then evaluate the path again for no reason. If you guarded this second if, then it would also work as it would return false for can_access_location but I would return soon as you can.

app/questionnaire/router.py
```diff
@@ -258,6 +268,121 @@ def _get_return_to_location_url(
...
+            # the grand calculated summary is in a different section which will have a different routing path
+            # BUT - we should not jump back to the grand calculated summary if the current section is not complete
+            # instead go to the next incomplete block in the section
+            if next_incomplete_location := self._get_first_incomplete_location_in_section(
```
MebinAbraham, line 294: This is inconsistent between same section vs different section. If diff section, it could return first incomplete, but not if it was same section, then it could return None as location isn't accessible?.

app/questionnaire/router.py
```diff
@@ -258,6 +268,121 @@ def _get_return_to_location_url(
...
+                    # but only in the case of going forwards, not for previous
+                    return None
+                return next_incomplete_location.url(
+                    return_to=return_to,
+                    return_to_block_id=return_to_block_id,
+                )
+
+            routing_path = self._path_finder.routing_path(
+                section_id=grand_calculated_summary_section
+            )
```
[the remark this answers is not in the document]
katie-gardner: Thanks! ended up doing something simillar but a bit different as the incomplete location was being used in 3 places after I addressed another comment

app/questionnaire/router.py
katie-gardner: updated 👍

app/questionnaire/router.py
katie-gardner: Fixed - thanks 👍

app/questionnaire/router.py
katie-gardner: Yep, good spot, I've updated the router test to have a test-case for an inaccessible GCS in the same section as well as one in another section, and fixed the issue

kylelawsonAND:
Functionally tested and can't break it so thumbs up there, will await walkthrough with you til I give the green tick

schemas/test/en/test_grand_calculated_summary_cross_section_dependencies.json
```diff
@@ -0,0 +1,341 @@
...
+                                ]
+                            }
+                        },
+                        {
+                            "type": "Question",
+                            "id": "skip-calculated-summary",
+                            "question": {
+                                "type": "General",
+                                "id": "skip-question-2",
+                                "title": "Skip the calculated summary of other outgoing costs so it isn`t included in the grand calculated summary?",
```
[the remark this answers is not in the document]
katie-gardner: Yeah the backtick did look a bit odd - validator disallows normal apostrophe's so I've swapped it out for the character the other surveys use - fixed!

schemas/test/en/test_grand_calculated_summary_overlapping_answers.json
```diff
@@ -0,0 +1,329 @@
...
+                                },
+                                "answers": [
+                                    {
+                                        "type": "Radio",
+                                        "id": "radio-extra",
+                                        "mandatory": true,
+                                        "options": [
+                                            {
+                                                "label": "Yes, I am going to buy two of everything",
+                                                "value": "AB"
```
berroar, line 194: Minor, but clearer when the labels and values match in our schemas. This rule used to be enforced by validator but was removed so we could support dynamic answers, but still best practice to match where possible

app/questionnaire/value_source_resolver.py
```diff
@@ -163,12 +164,19 @@ def _resolve_list_value_source(self, value_source: Mapping) -> int | str | list:
 
     def _resolve_calculated_summary_value_source(
         self, value_source: Mapping, *, assess_routing_path: bool
-    ) -> IntOrDecimal:
+    ) -> IntOrDecimal | None:
         """Calculates the value for the 'calculation' used by the provided Calculated Summary.
 
-        The caller is responsible for ensuring the provided Calculated Summary and its answers are on the path.
+        The caller is responsible for ensuring the provided Calculated Summary and its answers are on the path,
```
MebinAbraham, line 170: This comment is no longer true?

app/questionnaire/router.py
```diff
@@ -258,6 +268,140 @@ def _get_return_to_location_url(
...
+                block_id=return_to_block_id,
+                return_to=return_to,
+                _anchor=return_to_answer_id,
+            )
+        return self._get_return_url_for_inaccessible_location(
+            is_for_previous=is_for_previous,
+            return_to_block_id=return_to_block_id,
+            return_to=return_to,
+            routing_path=routing_path,
+        )
```
MebinAbraham, line 323: Im not sure that we got this quite right, for the overlapping answers scenario, if you change the answer from the second CS then click Continue on first CS, then if you click Previous, you go to the Hub which doesn't seem right. We should stay in the same section if we can't route out?. If prev is GCS but we can't access it, go the block before it?

MebinAbraham (approved):
👍

app/questionnaire/router.py
```diff
@@ -258,6 +268,140 @@ def _get_return_to_location_url(
                 "questionnaire.submit_questionnaire", _anchor=return_to_answer_id
             )
 
+    def _get_return_to_for_grand_calculated_summary(
```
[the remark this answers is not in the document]
katie-gardner: I noticed that this was something calculated summaries was doing as well, where the return_to is the calculated_summary even on the calculated summary page and the return_to_block_id is None, and I didn't change it at the time as it was a number of test updates - but should the return_to just be none in this case? If so shall I go ahead and fix for both?

app/questionnaire/value_source_resolver.py
katie-gardner: is `self.routing_path_block_ids` is None, (as in quite a few tests) then it will return the sum of the calculated summary regardless of whether it is on the path

app/questionnaire/router.py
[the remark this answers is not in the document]
katie-gardner: Going to leave it as is for this one, changing it causes a lot of tests failures so think its best to stick with the existing functionality

berroar (approved):

kylelawsonAND (approved):
Had a final look and can't pick any more holes, nice one 🌟

kylelawsonAND (approved):
Had a final look and can't pick any more holes, nice one 🌟

templates/calculatedsummary.html
katie-gardner: @petechd  happy for this to stay as is?

petechd (approved):

PR #1111, opened 2023-05-23
Fix invalid values in numerical form inputs 

petechd:
This adds some extra form data clean up to strip down all supported number symbols but don't cause further errors as a result of bad string to integer/decimal conversion in transforms. It should prevent the issues with invalid number inputs we experienced in the past. Some logic changes were needed for handling all the expected behaviours in runner (eg. percentage field with decimal value as input need to get through `process_formdata` method and raise `ValueError` exception in order to fail form validation).

### How to review
New tests added, make sure all possible inputs are covered. Are there any other form field classes that require this change?

### Checklist

### DIFF
--- a/app/forms/fields/decimal_field_with_separator.py
+++ b/app/forms/fields/decimal_field_with_separator.py
@@ -23,8 +23,16 @@ class DecimalFieldWithSeparator(DecimalField):
     def process_formdata(self, valuelist):
         if valuelist:
             try:
-                self.data = Decimal(
-                    valuelist[0].replace(numbers.get_group_symbol(DEFAULT_LOCALE), "")
-                )
+                data = valuelist[0]
+                if numbers.get_group_symbol(DEFAULT_LOCALE) in data:
+                    data = data.replace(numbers.get_group_symbol(DEFAULT_LOCALE), "")
+                try:
+                    self.data = Decimal(
+                        numbers.format_decimal(
+                            data, locale=DEFAULT_LOCALE, group_separator=False
+                        )
+                    )
+                except InvalidOperation:
+                    self.data = Decimal(data)
             except (ValueError, TypeError, InvalidOperation):
                 pass
--- a/app/forms/fields/integer_field_with_separator.py
+++ b/app/forms/fields/integer_field_with_separator.py
@@ -1,3 +1,5 @@
+from decimal import InvalidOperation
+
 from babel import numbers
 from wtforms import IntegerField
 
@@ -21,8 +23,16 @@ class IntegerFieldWithSeparator(IntegerField):
     def process_formdata(self, valuelist):
         if valuelist:
             try:
-                self.data = int(
-                    valuelist[0].replace(numbers.get_group_symbol(DEFAULT_LOCALE), "")
-                )
+                data = valuelist[0]
+                if numbers.get_group_symbol(DEFAULT_LOCALE) in data:
+                    data = data.replace(numbers.get_group_symbol(DEFAULT_LOCALE), "")
+                try:
+                    self.data = int(
+                        numbers.format_decimal(
+                            data, locale=DEFAULT_LOCALE, group_separator=False
+                        )
+                    )
+                except InvalidOperation:
+                    self.data = int(data)
             except ValueError:
                 pass
--- a/tests/app/forms/test_custom_fields.py
+++ b/tests/app/forms/test_custom_fields.py
@@ -1,3 +1,5 @@
+from decimal import Decimal
+
 import pytest
 from wtforms.fields import Field
 
@@ -29,6 +31,45 @@ def test_integer_field(mock_form):
         integer_field.process_formdata(["NonInteger"])
     except IndexError:
         pytest.fail("Exceptions should not thrown by CustomIntegerField")
+
+
+@pytest.mark.parametrize(
+    "number_input, result",
+    [
+        ("_110", 110),
+        ("1_10", 110),
+        ("1__10", 110),
+        ("1_1,0", 110),
+        ("_1_1,0,0", 1100),
+        ("1.10", None),
+    ],
+)
+def test_integer_field_formats(mock_form, number_input, result):
+    integer_field = IntegerFieldWithSeparator(_form=mock_form, name="aName")
+    assert isinstance(integer_field, Field)
+
+    integer_field.process_formdata([number_input])
+
+    assert integer_field.data == result
+
+
+@pytest.mark.parametrize(
+    "number_input, result",
+    [
+        ("1_1,0", Decimal("110")),
+        ("1.10", Decimal("1.1")),
+        ("_1.1_0", Decimal("1.1")),
+        ("_1.1_0,0", Decimal("1.1")),
+        ("_1._1,0,0", Decimal("1.1")),
+    ],
+)
+def test_decimal_field_formats(mock_form, number_input, result):
+    integer_field = DecimalFieldWithSeparator(_form=mock_form, name="aName")
+    assert isinstance(integer_field, Field)
+
+    integer_field.process_formdata([number_input])
+
+    assert integer_field.data == result
 
 
 def test_decimal_field(mock_form):


app/forms/fields/integer_field_with_separator.py
```diff
@@ -21,8 +23,16 @@ def __init__(self, **kwargs):
...
+                if numbers.get_group_symbol(DEFAULT_LOCALE) in data:
+                    data = data.replace(numbers.get_group_symbol(DEFAULT_LOCALE), "")
+                try:
+                    self.data = int(
+                        numbers.format_decimal(
+                            data, locale=DEFAULT_LOCALE, group_separator=False
+                        )
+                    )
+                except InvalidOperation:
+                    self.data = int(data)
```
MebinAbraham, line 36: I don't follow why we are formatting to decimal then converting back. Don't you simply need to replace underscores?. i.e `.replace("_", "")`?

Applies to the other one also.

```py
    def process_formdata(self, valuelist):
        if valuelist:
            try:
                self.data = int(
                    valuelist[0].replace(numbers.get_group_symbol(DEFAULT_LOCALE), "").replace("_", "")
                )
            except ValueError:
                pass
```

Yuyuutsu (commented):
My runner is breaking when inputting `NaN`.

app/forms/fields/integer_field_with_separator.py
petechd: My idea here was to format and get rid off all the other possible characters except "," and "_" that would get to this step in code, but can just chain `replace`.

MebinAbraham: I think the `get_group_symbol` and `_` is fine. We don't need to support every character is they are not valid, `_` is commonly used as separators. However, any other character should raise the appropriate validation errors.

petechd:
> My runner is breaking when inputting `NaN`.

This should result in "INVALID NUMBER" and "INVALID DECIMAL" form validation errors now, thanks.

app/forms/fields/integer_field_with_separator.py
petechd: Ok, added `replace` and some extra tests. Thanks!

app/forms/fields/decimal_field_with_separator.py
```diff
@@ -22,9 +23,13 @@ def __init__(self, **kwargs):
 
     def process_formdata(self, valuelist):
         if valuelist:
+            if valuelist[0] == "NaN":
+                raise ValidationError(error_messages["INVALID_NUMBER"])
```
MebinAbraham, line 27: This shouldn't be done here, this is a job for the validators. This should be done in our `NumberRange` validator.

You need to update it to:
```py
value: Union[int, Decimal] = field.data
if value is not None and not math.isnan(value):
```

This matches how wtform official handles it. https://github.com/wtforms/wtforms/pull/548/files

petechd: The problem here is, by the time we use `NumberRange` validator the form field.data is already transformed to `None` from the original `NaN`, so I am not sure what is the solution here.

MebinAbraham: But that is not an issue is it because they are asked to enter a valid number...?

Yuyuutsu (commented):
Functionally works as expected 👍  Will have a relook after changes relating to mebins comments

app/forms/fields/decimal_field_with_separator.py
petechd: Ok, had to slightly modify your suggestion to make it work, thanks.

app/forms/fields/integer_field_with_separator.py
```diff
@@ -20,9 +21,13 @@ def __init__(self, **kwargs):
...
     def process_formdata(self, valuelist):
         if valuelist:
+            if valuelist[0] == "NaN":
+                raise ValidationError(error_messages["INVALID_NUMBER"])
             try:
                 self.data = int(
-                    valuelist[0].replace(numbers.get_group_symbol(DEFAULT_LOCALE), "")
+                    valuelist[0]
+                    .replace(numbers.get_group_symbol(DEFAULT_LOCALE), "")
+                    .replace("_", "")
```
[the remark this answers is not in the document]
petechd: Added now, thanks!

app/forms/validators.py
```diff
@@ -48,11 +49,17 @@ def __call__(
         form: FlaskForm,
         field: Union[DecimalFieldWithSeparator, IntegerFieldWithSeparator],
     ) -> None:
+        try:
+            if math.isnan(float(field.raw_data[0])):
+                raise validators.StopValidation(error_messages["INVALID_NUMBER"])
```
kylelawsonAND, line 54: Not familiar with wtforms validation - would be good to have a quick chat about this raise-pass pattern so I can understand it.

app/forms/validators.py
```diff
@@ -48,11 +49,17 @@ def __call__(
         form: FlaskForm,
         field: Union[DecimalFieldWithSeparator, IntegerFieldWithSeparator],
     ) -> None:
+        try:
+            if math.isnan(float(field.raw_data[0])):
```
MebinAbraham, line 53: Why are we casting to a float?. Also error message should be `self.message` as this is in the context of a validator. Same for the other one.

app/forms/validators.py
petechd: I mean, it is more general logic than wtforms, here we only want to raise an error (stop validation and raise error) when `isnan()` can be executed, some of the data we try to perform `isnan()` on will be invalid input and will break runner.  But I don't want to break runner on inputs that I don't want to check. That's why I had to add `except` type and value errors, for all the edge cases. But when inputs are not valid I `pass` cause I want to continue to the next lines. I only want to make this nan check on numeric inputs otherwise do the rest of the validation.

app/forms/validators.py
petechd: Just cause `nan` is a string here and it wouldn't evaluate as `True` if you try `math.isnan()` on a string. Added the self messages now. Thanks.

MebinAbraham: You can remove this `try catch` approach and do this at the end:

```py
        if "e" in field.raw_data[0].lower() or math.isnan(field.data):
            raise validators.StopValidation(self.message)
```

If data is None this can error but this should never get called when data is None as validation like this only kicks in if we have a value. You can also use `field.data` will be the processed value.

petechd: Also `math.isnan()` will error if input is not numerical value so we need this `try`?

app/forms/validators.py
```diff
@@ -126,6 +133,14 @@ def __call__(
         field: Union[DecimalFieldWithSeparator, IntegerFieldWithSeparator],
     ) -> None:
         value: Union[int, Decimal] = field.data
+        try:
+            raw_value = float(field.raw_data[0])
+        except (TypeError, ValueError):
+            raw_value = False
+
+        if math.isnan(raw_value):
+            raise validators.ValidationError(self.messages["INVALID_NUMBER"])
```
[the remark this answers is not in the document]
petechd: Added, thanks!

app/forms/validators.py
MebinAbraham: `field.data` isn't None when it is `NaN`, it will be `Decimal('NaN')`, same as you casting to float. So you should be able to use it as is.

petechd: Correct me if I am wrong since I was wrong about `field.data` but do we not need to cast float before we do isnumeric() since `field.data` are strings? That's what i do now so it's:
```python
        if "e" in field.raw_data[0].lower() or math.isnan(float(field.data[0])):
            raise validators.StopValidation(self.message)

```

petechd: Ok, thanks, the only edge case is when you have a whole number field and you enter decimal, it errors as the field data becomes `None` so I have to guard against it by checking if field data is not `None`.

MebinAbraham: I see that you have done `if field.data`, we should update it to be explicit here, so `if field.data is not None and ...`

tests/app/forms/validation/test_number_range_validator.py
```diff
@@ -57,6 +57,7 @@ def test_number_range_validator_raises_ValidationError(
         maximum_exclusive=max_excl,
     )
     mock_field.data = value
+    mock_field.raw_data = [value]
```
MebinAbraham, line 60: This file can be revertded?

tests/app/forms/validation/test_number_validator.py
```diff
@@ -26,6 +23,20 @@ def test_number_validator_raises_StopValidation(
     assert error_messages["INVALID_NUMBER"] == str(exc.value)
 
 
+@pytest.mark.usefixtures("gb_locale")
+def test_number_validator_NaN_raises_StopValidation(
+    number_check, mock_form, mock_field
+):
+    value = "NaN"
```
MebinAbraham, line 30: Can you not just do this in the one above?. Just always set the data given that is the real scenario. Then you don't need a dedicated test here.

petechd: Yeah but in above scenario Decimal will be casted on "a"  and try except will stop on exception but you cannot cast decimal on "a" in preparation before you run number check, but at the point where isnan checked Decimal is casted on NaN already. So I had to  created this separate. Basically you cannot do
```python
mock_field.data = Decimal(value)
```
 in:
```python
@pytest.mark.parametrize(
    "value",
    ([None], [""], ["a"], ["2E2"]),
)
@pytest.mark.usefixtures("gb_locale")
def test_number_validator_raises_StopValidation(
    number_check, value, mock_form, mock_field
):
    mock_field.raw_data = value
    mock_field.data = Decimal(value)

    with pytest.raises(StopValidation) as exc:
        number_check(mock_form, mock_field)

    assert error_messages["INVALID_NUMBER"] == str(exc.value)
```
when the value is "a"

MebinAbraham: Fair enough. 👍

app/forms/validators.py
```diff
@@ -183,6 +185,7 @@ def __call__(
             field.raw_data[0]
             .replace(numbers.get_group_symbol(flask_babel.get_locale()), "")
             .replace(" ", "")
+            .replace("_", "")
```
MebinAbraham, line 188: We seem to have an inconsistent amount of call like these, I suggest make a helper method that, does this given an input, replace group symbol, whitespace and `_` then return the value. Then we only have a single place to update, this is done in like 4 places now.

tests/app/forms/validation/test_number_range_validator.py
petechd: Removed/reverted now. Thanks!

app/forms/validators.py
petechd: Added, thanks!

Yuyuutsu (approved):
Changes LGTM !

petechd:
Thanks, I've moved all helper methods to form helpers module now.

app/helpers/form_helpers.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+
+def sanitise_mobile_number(data: str) -> str:
+    data = re.sub(r"[\s.,\t\-{}\[\]()/]", "", data)
+    return re.sub(r"^(0{1,2}44|\+44|0)", "", data)
+
+
+def format_playback_value(
+    value: Union[float, Decimal], currency: Optional[str] = None
```
kylelawsonAND, line 26: Union[float, Decimal] -> float | Decimal

Optional[str] -> str | None

kylelawsonAND (changes requested):

app/forms/validators.py
```diff
@@ -49,15 +54,15 @@ def __call__(
...
+            Decimal(sanitise_number(number=field.raw_data[0]))
         except (ValueError, TypeError, InvalidOperation, AttributeError) as exc:
             raise validators.StopValidation(self.message) from exc
 
-        if "e" in field.raw_data[0].lower():
+        if "e" in field.raw_data[0].lower() or math.isnan(
+            Decimal(
+                sanitise_number(number=field.raw_data[0])
+            )  # number is sanitised to guard against inputs like `,NaN_` etc
+        ):
```
MebinAbraham, line 65: How come this changed?. What broke it?.

app/forms/validators.py
```diff
@@ -49,15 +54,15 @@ def __call__(
...
-            )
+            Decimal(sanitise_number(number=field.raw_data[0]))
         except (ValueError, TypeError, InvalidOperation, AttributeError) as exc:
             raise validators.StopValidation(self.message) from exc
 
-        if "e" in field.raw_data[0].lower():
+        if "e" in field.raw_data[0].lower() or math.isnan(
+            Decimal(
+                sanitise_number(number=field.raw_data[0])
+            )  # number is sanitised to guard against inputs like `,NaN_` etc
```
MebinAbraham, line 64: The comment shouldn't be here. Add doc strings to the sanitise number method instead.

app/forms/validators.py
petechd: Test percentage schema has integer with zero decimals input that, when it arrives at `NumberCheck` validator, has already `field.data` value set as `None`, so it was passing validation when you entered `"NaN"`. We have to use `raw_data` here?

app/forms/validators.py
petechd: I was thinking this comment only applies here when the value entered is `",NaN_"` or similar. If we don't  use sanitise method here validation would not stop, but it is such a rare case I preferred to leave comment.

MebinAbraham: Fair point.

MebinAbraham (approved):

app/forms/validators.py
MebinAbraham, line 65: Minor, this can be simplified, since you do the same call already.

```py
        try:
            # number is sanitised to guard against inputs like `,NaN_` etc
            number = Decimal(sanitise_number(number=field.raw_data[0]))
        except (ValueError, TypeError, InvalidOperation, AttributeError) as exc:
            raise validators.StopValidation(self.message) from exc

        if "e" in field.raw_data[0].lower() or math.isnan(number):
            raise validators.StopValidation(self.message)
```

petechd: Yeah, I was suspecting some potential refactor here but couldn't spot it like you. Makes sense, no point calling it twice.

app/helpers/form_helpers.py
petechd: Done now, thanks.

kylelawsonAND (approved):

PR #1112, opened 2023-05-25
Looping 2.5 | List Item Progress Displayed on List Collector Page

kylelawsonAND:
This incremental PR forms part of the[ Looping 2.5](https://trello.com/c/FYXB2EYJ) feature which is currently in [draft here](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1106).
This PR addresses the requirement for list item completion to be shown to on the List Collector page. List items that have repeating blocks and that are complete must show a green check mark on their row of the List Collector, all other items including complete list items which do not have repeating blocks, should have no icon. 

### Changes
**jinja_filters**: `map_summary_item_config` and `map_list_collector_config` now no longer expect an icon string to be given as a parameter. `map_list_collector_config` now expects a boolean parameter to indicates whether an icon should be rendered for each list item. 
The icon displayed is the check mark icon identified by "check", if the list item is complete and has repeating blocks.

**list_context**: `ListContext` now expects a `Location` object as a constructor parameter which represents the location of the list collector block that it is the context of. 
`_build_list_items_context ` will fetch the list collector block via the schema and determine if repeating blocks are present, this is used to add a boolean field to the individual list item context indicate it has repeating blocks.
Additionally the list item context dict has a boolean field to represent the list item's completeness by accessing the progress store. 

**list_collector_block**: The `list_summary_element` of `ListCollectorBlock` now passes a `Location` object to the `ListContext` it creates. Section Id is based on the Id of the section id of the location attribute given to this object at construction. list name and block id are taken from the list_collector_block variable which contains the schema element representing this list collector. The location attribute cannot be used as the list collector location as section summaries create this object and pass their own location. 

**question**: Passes it's `current_location` attribute as the list collector location to the `ListContext` it creates, as this Question(BlockHandler) represents the list collector currently rendered. 

**unrelated_question**:  Passes it's `current_location` attribute as the list collector location to the `ListContext` it creates.

**list-summary.html**: Passes True to render the dynamic list item icons now instead of "person" to always render the person icon. 

**tests/app/conftest.py**: Moved the location fixture here to be used by the questionnaire tests as well as contexts tests.

**tests/app/questionnaire/conftest.py**: Moved the location fixture from here to the tests/app level to be used by the questionnaire tests as well as contexts tests.

**test_jinja_filters**: Amending tests which expected an icon string to now expect the render icon bool. 

**test_list_context**: Many tests now have a `ProgressStore` fixture passed to them to be used to construct the tested `ListContext`, additionally they have the location fixture passed in as well to support the new constructor param. 
Some assertions now also assert that the `"is_complete"` and `"repeating_blocks" `bool flags are also present in the built list summaries/contexts. 

**test_section_summary**: 
Many tests now have a `ProgressStore` fixture passed to them to be used to construct the tested `ListContext`.
Some assertions now also assert that the `"is_complete"` and `"repeating_blocks" `bool flags are also present in the built list summaries/contexts. 

### How to review
Make use of the test schema `test_list_collector_repeating_blocks_section_summary.json` to ensure the check icon is shown on fully complete list items on the List Collector page. Try out other test schemas that make use of List Collectors and check that there is no check mark present on the list item rows, only on the schema that has repeating blocks.
All existing unit and integration tests should be working please flag if they are not, functional tests may fail. Anything failing with be fixed in the upcoming PRs. 

### Up next

- Unit testing
- Functional testing
- Additional test schemas
- Type hinting files/methods touched by this feature impl


kylelawsonAND:

app/views/contexts/summary/list_collector_block.py
```diff
@@ -107,6 +107,9 @@ def list_summary_element(self, summary: Mapping[str, Any]) -> dict[str, Any]:
         list_summary_context = self.list_context(
             list_collector_block["summary"],
             for_list=list_collector_block["for_list"],
+            list_collector_location=Location(section_id=self._location.section_id,
```
katie-gardner, line 110: It looks like this passes various attributes of the list collector block to ListContext, which calls the build_list_items_context which then looks up the block again - could you just pass down the list collector block instead?

app/views/contexts/list_context.py
```diff
@@ -44,6 +48,7 @@ def __call__(
     def _build_list_items_context(
```
katie-gardner, line 48: enforcing keyword args for this might be useful

katie-gardner (changes requested):
Functionally working nicely, few queries so far

tests/app/test_jinja_filters.py
```diff
@@ -318,14 +318,14 @@ def test_map_list_collector_config_no_actions():
...
-    output = map_list_collector_config(list_items, "icon")
+    output = map_list_collector_config(list_items, False)
 
     expected = [
         {
             "rowItems": [
                 {
                     "actions": [],
-                    "iconType": "icon",
+                    "iconType": None,
```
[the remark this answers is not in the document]
kylelawsonAND: Afraid not, just tried there, the result contains the key with None value, so omitting causes assertion to fail

app/views/contexts/summary/list_collector_block.py
kylelawsonAND: Brilliant point, will have a go at amending to that!

MebinAbraham: I don't mind either approach, ideally a function should be given just what it needs.
- If you pass the block, then you can remove a bunch of these args
- Otherwise, pass in whether it has repeating and a section id?.

templates/partials/summary/list-summary.html
```diff
@@ -3,7 +3,7 @@
 {% if list.editable %}
   {% set rows = map_list_collector_config(
     list.list_items,
-    "person",
+    True,
     _("Change"),
     _("Change details for {item_name}"),
     _("Remove"),
```
MebinAbraham, line 9: Minor and existing, I suggest keyword args.

kylelawsonAND: Added kwargs in this tempate 👍

app/views/contexts/list_context.py
kylelawsonAND: Enforced 👍

MebinAbraham (approved):

PR #1120, opened 2023-06-01
Add validation of calculated sum of dynamic answers

petechd:
Adds group validation of dynamic, list collector items based answers and static ones.  Sum of answers that are collected by a looped question are checked against hardcoded total or answer source, typical for calculation.

### How to review
Check new schema, functional tests and new unit tests.

### Checklist


petechd:
[Looping 3.5 - Validating dynamic answers based on a list collector (3 SP)](https://trello.com/c/WPWjw3Vp/5324-looping-35-validating-dynamic-answers-based-on-a-list-collector-3-sp)

berroar:
Can you update this so that it is raised against the feature-prepop branch rather than main 👍

app/forms/questionnaire_form.py
```diff
@@ -386,11 +386,25 @@ def _get_period_limits(
         return minimum, maximum
 
     def _get_formatted_calculation_values(
-        self, answers_list: Sequence[str]
+        self, answers_sequence: Sequence[str]
     ) -> list[str]:
+        answers_list = list(answers_sequence)
+        block_id = self.location.block_id if self.location else None  # type: ignore
+        # type ignore self.location always exists at this point
+        list_name = (
```
berroar, line 394: Do we only want to call this method if the block has dynamic answers, so use the `self.block_has_dynamic_answer` to check if we need to make this call?

app/questionnaire/questionnaire_schema.py
```diff
@@ -734,6 +726,18 @@ def is_answer_in_repeating_section(self, answer_id: str) -> bool | None:
     def is_answer_dynamic(self, answer_id: str) -> bool:
         return answer_id in self._dynamic_answer_ids
 
+    def get_list_name_for_dynamic_answer(self, block_id: str) -> str | None:
+        return self.get_block(block_id).get("question", {}).get("dynamic_answers", {}).get("values", {}).get("identifier", {})  # type: ignore
```
berroar, line 730: With the above this can be simplified as we will only be checking dynamic blocks

berroar: Also just a minor point I would move the type ignore message above the line of code

app/questionnaire/questionnaire_schema.py
```diff
@@ -734,6 +726,18 @@ def is_answer_in_repeating_section(self, answer_id: str) -> bool | None:
...
+        # type ignore block always exists at this point
+
+    def block_has_dynamic_answer(self, block_id: str) -> bool:
+        return (
+            "dynamic_answers" in self.get_block(block_id).get("question", {})  # type: ignore
```
kylelawsonAND, line 735: If this is being type ignored as the block will always exist, this might as well be simplified to` return "dynamic_answers" in self.get_block(block_id).get("question", {})  # type: ignore`

app/forms/questionnaire_form.py
petechd: Yeah, makes sense, no need to do this for all.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -496,7 +496,11 @@ def remove_dependent_blocks_and_capture_dependent_sections(self) -> None:
             section_key,
             blocks_to_remove,
         ) in self.dependent_block_id_by_section_key.items():
-            if section_key not in self.started_section_keys():
+            if section_key not in self.started_section_keys() and not any(
```
berroar, line 499: Maybe worth adding a comment here to explain why this additional check has been added?

app/forms/questionnaire_form.py
```diff
@@ -386,8 +386,24 @@ def _get_period_limits(
...
+        answers_list = list(answers_sequence)
+        block_id = self.location.block_id if self.location else None
+        if block_id and self.schema.block_has_dynamic_answer(block_id):
+            list_name = (
+                self.schema.get_list_name_for_dynamic_answer(block_id)
+                if block_id
+                else None
+            )
+            list_item_ids = self.list_store[list_name] if list_name else None
+            for answer_id in answers_list:
```
katie-gardner, line 400: I'd avoid extending a list whilst iterating over it, and suggest changing this so that instead of making answers_list as a copy of answers sequence at the top, instead initialise it as an empty list, and then iterate over the sequence and add values based off that, ie.
```
answers_list = []
... 
for answer_index, answer_id in answers_sequence:
   if self.schema.is_answer_dynamic(answer_id) and list_item_ids:
     answers_list.extend(f"{answer_id}-{list_item_id}" for list_item_id in list_item_ids)
   else:
     answers_list.append(answer_id)

Comment (petechd) [reply app/forms/questionnaire_form.py:400]: Well, this needs to take into account scenarios when block doesn't have dynamic answers lol, otherwise it works great and I agree this is better. We can't remove replace since it is removing number separators from our answers?

File: app/forms/questionnaire_form.py [hunk 6]
```diff
@@ -386,8 +386,24 @@ def _get_period_limits(
         return minimum, maximum
 
     def _get_formatted_calculation_values(
-        self, answers_list: Sequence[str]
+        self, answers_sequence: Sequence[str]
     ) -> list[str]:
+        answers_list = list(answers_sequence)
+        block_id = self.location.block_id if self.location else None
+        if block_id and self.schema.block_has_dynamic_answer(block_id):
```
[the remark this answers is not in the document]
petechd: According to type hints we have self.location can be one of None, Location, RelationshipLocation. so yeah block_id call can cause error when location is None.

app/forms/questionnaire_form.py
katie-gardner: yeah good point - your solution will definitely be better for the no dynamic answers case - and now you've got best of both, nice 👍 
so yeah the .replace comment doesn't make sense anymore

app/forms/questionnaire_form.py
```diff
@@ -386,8 +386,27 @@ def _get_period_limits(
...
     def _get_formatted_calculation_values(
-        self, answers_list: Sequence[str]
+        self, answers_sequence: Sequence[str]
     ) -> list[str]:
+        answers_list: list[str] = []
+        block_id = self.location.block_id if self.location else None
+        if block_id and self.schema.block_has_dynamic_answer(block_id):
+            list_name = (
+                self.schema.get_list_name_for_dynamic_answer(block_id)
+                if block_id
```
berroar, line 396: Do you need the if here given we check for block above?

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -496,7 +496,14 @@ def remove_dependent_blocks_and_capture_dependent_sections(self) -> None:
...
             blocks_to_remove,
         ) in self.dependent_block_id_by_section_key.items():
-            if section_key not in self.started_section_keys():
+            # We don't want to skip current iteration if the block has dynamic answers,
+            # if block has dynamic answers we don't use list_item_id from section_key set, so the location is removed in
+            # self.remove_completed_location and dependencies updated
+            if section_key not in self.started_section_keys() and not any(
+                block_id
+                for block_id in blocks_to_remove
+                if self._schema.block_has_dynamic_answer(block_id)
```
berroar, line 505: Was thinking rather than doing this check on every block we could add a method to questionnaire schemas to populate a set of block_ids containing dynamic answers upfront and use that here and below instead?

berroar:
Do we need to update the schema to demonstrate this working across sections?

app/questionnaire/questionnaire_store_updater.py
petechd: I guess, it makes sense. We invoke it 3x so far.

app/forms/questionnaire_form.py
petechd: My bad, now I can see it, seems like we don't need this at all.

app/questionnaire/questionnaire_store_updater.py
petechd: Added now, thanks!

app/questionnaire/questionnaire_schema.py
```diff
@@ -1248,6 +1254,22 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
                 if item["for_list"] == list_name and item.get("item_anchor_answer_id"):
                     return f"#{str(item['item_anchor_answer_id'])}"
 
+    def update_dependencies_for_dynamic_answers(
+        self, *, question: Mapping, block_id: str
+    ) -> None:
+        if dynamic_answers := question.get("dynamic_answers"):
```
katie-gardner, line 1260: in this if statement, if you did `self.dynamic_answers_parent_block_ids.add(block_id)`
I think you would then not need the `_populate_dynamic_answers_parent_block_ids` method

petechd:
> Do we need to update the schema to demonstrate this working across sections?

Added section and hub, so all should be covered now..

app/questionnaire/questionnaire_schema.py
petechd: Added now, thanks!

app/forms/questionnaire_form.py
```diff
@@ -386,8 +386,23 @@ def _get_period_limits(
...
-        self, answers_list: Sequence[str]
+        self, answers_sequence: Sequence[str]
     ) -> list[str]:
+        answers_list: list[str] = []
+        block_id = self.location.block_id if self.location else None
+        if block_id and block_id in self.schema.dynamic_answers_parent_block_ids:
+            list_name = self.schema.get_list_name_for_dynamic_answer(block_id)
+            list_item_ids = self.list_store[list_name] if list_name else None
+            for answer_id in answers_sequence:
+                if self.schema.is_answer_dynamic(answer_id) and list_item_ids:
```
katie-gardner, line 397: and then you wont need to check `and list_item_ids`

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -505,7 +512,9 @@ def remove_dependent_blocks_and_capture_dependent_sections(self) -> None:
             for block_id in blocks_to_remove:
                 location = Location(
                     section_id=section_id,
-                    list_item_id=list_item_id,
+                    list_item_id=None
```
katie-gardner, line 515: I reckon the change I've done for looping-4 would work here after all, because then the dynamic-answers block won't have a list-item-id associated to it, which was the problem I had too, so then you shouldn't need to manually modify the location here

katie-gardner: I think the only change that would be needed is the bit in questionnaire_schema which does this 
```
     for block_answer_id in answer_ids_for_block:
            self._answer_dependencies_map[block_answer_id] |= {
                self._get_answer_dependent_for_block_id(
                    block_id=block_id, for_list=list_name
                )
                if self.is_block_in_repeating_section(block_id)
                # non-repeating blocks such as dynamic-answers could depend on the list
                else self._get_answer_dependent_for_block_id(block_id=block_id)
            }
```

app/views/handlers/list_action.py
```diff
@@ -92,6 +92,20 @@ def handle_post(self):
                 self.current_location.section_id, self.current_location.list_item_id
             )
             self.questionnaire_store_updater.remove_dependent_blocks_and_capture_dependent_sections()
+            self.current_location.list_item_id = (
+                # list_item_id has to be removed in order to update section completeness, otherwise in self._update_section_completeness() call
+                # section key won't be accepted and section status will remain "COMPLETED", list_item_id is removed if list collector block dependencies
+                # are present, a specific scenario for dynamic answers
+                None
+                if self.questionnaire_store_updater.dependent_block_id_by_section_key[
```
katie-gardner, line 100: if schema list item id worked above it might work here too?

katie-gardner (changes requested):
Some suggestions but functionally works great! 👍

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -496,7 +496,14 @@ def remove_dependent_blocks_and_capture_dependent_sections(self) -> None:
...
             blocks_to_remove,
         ) in self.dependent_block_id_by_section_key.items():
-            if section_key not in self.started_section_keys():
+            # We don't want to skip current iteration if the block has dynamic answers,
+            # if block has dynamic answers we don't use list_item_id from section_key set, so the location is removed in
+            # self.remove_completed_location and dependencies updated
+            if section_key not in self.started_section_keys() and not any(
+                block_id
+                for block_id in blocks_to_remove
+                if block_id in self._schema.dynamic_answers_parent_block_ids
```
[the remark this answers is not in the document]
katie-gardner: though actually I think if you changed the questionnare_schema to make the dynamic-answer dependencies not have a list item id, then the section key would already be in `self.started_section_keys()` and then possibly this check would not be required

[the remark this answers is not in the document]
petechd: Wow, this is great!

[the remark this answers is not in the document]
petechd: Great, every change makes the code cleaner now!

app/questionnaire/questionnaire_store_updater.py
petechd: What an improvement, great stuff. Thanks.

app/questionnaire/questionnaire_schema.py
```diff
@@ -335,20 +336,14 @@ def _populate_answer_dependencies(self) -> None:
                     self._update_answer_dependencies_for_calculations(
                         question["calculations"], block_id=block["id"]
                     )
+                    self.update_dependencies_for_dynamic_answers(
```
katie-gardner, line 339: just occured to me, if this needs to run when question is calculated and when it isn't could you put this update before the if block and then you don't need it twice?

petechd: Makes absolute sense, refactored now, thanks!

kylelawsonAND (approved):
Can't pick anything else out in code + works as intended - approval once all other comments resolved 👌 Nice one!

app/questionnaire/questionnaire_schema.py
katie-gardner: at a glance I'm not sure what has gone wrong with the test when this was updated, I didn't think the answer dependencies for calculations would depend on going first? if it did.... oops

katie-gardner: functionally it still works fine 🤔

katie-gardner: test passes locally too

katie-gardner (approved):
Looking great! 👍

berroar (approved):

Yuyuutsu (approved):

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=validating-dynamic-based-on-list-collector
```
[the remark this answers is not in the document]
petechd: Done now, thanks.

PR #1122, opened 2023-06-02
Support optional question guidance and descriptions

berroar:

Supports optional descriptions and contents lists in Question description and guidance fields. Previously, Question descriptions and guidance would have still been rendered for None values. This change is needed to support prepop data where we may not always have values for descriptions and guidance fields.

### How to review
Use the test schema `test_optional_guidance_and_description.json` to test the change.

### Checklist


berroar:
[Support for optional question guidance and description (2 SP)](https://trello.com/c/GOW1kdjI/5708-support-for-optional-question-guidance-and-description-2-sp)

templates/partials/question.html
```diff
@@ -58,12 +58,16 @@
...
   {%- if question.guidance -%}
     {% set contents = question.guidance.contents %}
-    {% call onsPanel({
-      "id": "question-guidance-" ~ question.id,
-      "classes": "ons-u-mb-m"
-    }) %}
-      {% include 'partials/contents.html' %}
-    {% endcall %}
+    {% for item in contents %}
+      {%- if item["title"] or item["description"] or item["list"] -%}
```
katie-gardner, line 62: would `item.title`, `item.description` and `item.list` not work here?

templates/partials/question.html
```diff
@@ -58,12 +58,16 @@
 {% set question_guidance %}
   {%- if question.guidance -%}
     {% set contents = question.guidance.contents %}
-    {% call onsPanel({
-      "id": "question-guidance-" ~ question.id,
-      "classes": "ons-u-mb-m"
-    }) %}
-      {% include 'partials/contents.html' %}
-    {% endcall %}
+    {% for item in contents %}
```
katie-gardner, line 61: I think this has broken the `test_question_guidance.json` schema, its rendering the block 4 times now, I don't think the loop will work here, might need to be in contents?

katie-gardner (changes requested):

templates/partials/question.html
katie-gardner: since this got through, might need to update a functional test to cover this?

kylelawsonAND: Yep just found this issue too

berroar: Added a fix, just need to add a test

schemas/test/en/test_optional_guidance_and_description.json
```diff
@@ -0,0 +1,216 @@
...
+                                    {
+                                        "id": "answer-1",
+                                        "label": "Enter any description text that you want to be displayed",
+                                        "max_length": 20,
+                                        "mandatory": false,
+                                        "type": "TextField"
+                                    },
+                                    {
+                                        "id": "answer-2",
+                                        "label": "Enter any description text that you want to be displayed",
```
katie-gardner, line 52: Maybe change description to `Enter any alternative text you want to be displayed` or have question guidance to mention that its first non empty item that gets shown? might just be as I hadn't seen the block before but I assumed if one of them was empty that it would make a difference, but both need to be

It was clear enough after trying though so don't mind if you just resolve this comment

kylelawsonAND (approved):
LGTM 👍

templates/partials/question.html
katie-gardner: thats really interesting I didn't actually realise it was different - https://jinja.palletsprojects.com/en/3.0.x/templates/#variables

It should only be a difference in the order it tries to find the variable, I tried adding 
```
{{ item["title"]|pprint }}
<br/>
{{ item.title|pprint }}
```
above this and it printed the same my end, but in any case, square brackets look for the item before the attribute which isn't there so what you have is better anyway 👀 , my bad

Yuyuutsu (approved):
Changes look good to me and functionally tested 👍

templates/partials/question.html
```diff
@@ -58,12 +59,19 @@
...
+      {%- if item["title"] or item["description"] or item["list"] -%}
```
petechd, line 63: Non blocker but I've just checked and contents.html also supports "definition" field and it is accepted as a field in contents in validator but can it be used here? Should we check for item["definition"]? Does guidance support "definition"?

berroar: Updated the if statement as discussed 👍

tests/functional/spec/question_description.spec.js
```diff
@@ -1,8 +1,35 @@
...
 describe("Question description", () => {
   it("Given a question description has been set in the schema as an array, When it is rendered, Then it is displayed correctly as multiple paragraph attributes", async () => {
     await browser.openQuestionnaire("test_question_description.json");
     await expect(await $(NameBlockPage.questionTitle()).getHTML()).to.contain("<p>Answer the question</p><p>Go on</p>");
   });
 });
+
+describe("Optional question description and guidance", () => {
+  it("Given a question description has been set in the schema, When the value to be displayed is None, Then it is not rendered on the page", async () => {
+    await browser.openQuestionnaire("test_optional_guidance_and_description.json");
```
Yuyuutsu, line 17: Would we want a before() hook when launching the questionnaire?

mgdon650:
Looks good to me

tests/functional/spec/question_description.spec.js
berroar: Don't think I need that as there it's all being done within one test, would need before or before each if i had multiple cases for the same schema

rmccar (approved):

PR #1128, opened 2023-06-06
Feat looping 4.5

katie-gardner:
This PR adds the functionality that allows calculated summaries to use a list of repeating answers. Calculated summaries now support consisting of a single dynamic answer, or both dynamic and static answers where each dynamic answer resolves to a list of answers for each list item.

see the [Validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/201) for update to schemas to allow a single answer when it is dynamic

**Note**: Currently if you add/edit/remove an item from the list whilst on the section summary, you need to update and dynamic answers and reconfirm any affected calculated summaries, but the survey will take you through each block in the section until you reach it, rather than routing directly back to these two. The functionality to take the user to the next incomplete block in the section is part of a different card (linked to trello ticket) but you can use the dump to verify that the progress change is correct. The remove block needs to reset progress of the calculated summary, in case the dynamic answers aren't updated

### Technical details
`questionnare_schema.py` has been modified so that whilst dependencies for a question with dynamic answers are being added, that question, along with the `remove` block, are cached in an internal map. This then allows future calculated summaries which depend on those dynamic answers to use that map to additionally depend on the `remove` block. This is so that deleting an item on the list resets the progress of the calculated summary, and makes user re-confirm.

`questionnaire_store_updater.py` also required updating so that whilst list items and answers are being removed this triggers dependencies for the `remove` block to be re-evaluated, so that the changes to the schema class will work.

`evaluate_sum` in the operations now allows for some of its arguments to be lists, which it will recursively resolve

`value_source_resolver` has been updated so that whilst answers are being resolved, if the answer being resolved is dynamic but there is no list item id, then its for a list of dynamic answers, so resolve those instead. 

### How to review
Read through the trello card to familiarise with the concept and check that each scenario in the attached scenarios google sheet is working (Except the routing - see above). Check that the integration and functional tests (coming soon) have covered the scenarios.

### Checklist


katie-gardner:
[Looping 4 - Calculated summary using a list of repeating answers (5 SP)](https://trello.com/c/jXpOixe8/5320-looping-4-calculated-summary-using-a-list-of-repeating-answers-5-sp)

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
```
katie-gardner, line 2: Reminder to revert before merge

app/questionnaire/questionnaire_schema.py
```diff
@@ -378,6 +379,22 @@ def _update_answer_dependencies_for_calculated_summary_dependency(
             calculated_summary_block
         )
         for answer_id in calculated_summary_answer_ids:
+            # for any dynamic answer ids based on a list, the calculated summary needs to depend on the remove block for the list
```
berroar, line 382: Minor and not sure how the other devs feel about this, but maybe we could make the doc string a bit more comprehensive to cover the comments added here. Would make it more readable for me but not sure what others think 🤔 Same applies to some of the other method updated elsewhere in this PR.

katie-gardner: yeah absolutely 👍 I'll happily go back through and make the docstrings a bit stronger

berroar:
One minor quirk I noticed when testing the `test_new_calculated_summary_repeating_answers_only` schema, is that when I complete everything and click continue on the section summary, If I click previous on the Submit page it takes me back to the Calculated Summary page rather than the section summary page I was just on, not sure if this is expected 🤔

katie-gardner:
> One minor quirk I noticed when testing the `test_new_calculated_summary_repeating_answers_only` schema, is that when I complete everything and click continue on the section summary, If I click previous on the Submit page it takes me back to the Calculated Summary page rather than the section summary page I was just on, not sure if this is expected 🤔

Well spotted - it looks like an existing minor bug where list collector summary blocks aren't on the routing path when you click previous so `router.py` finds the previous block to be whatever the one before that was. The same thing happens on the existing schema `test_list_collector_section_summary` - we could raise another card for this?

app/questionnaire/questionnaire_schema.py
berroar: Looks good to me! 👍

schemas/test/en/test_new_calculated_summary_repeating_and_static_answers.json
```diff
@@ -0,0 +1,726 @@
...
+            "type": "string"
+        },
+        {
+            "name": "ru_name",
+            "type": "string"
+        }
+    ],
+    "questionnaire_flow": {
+        "type": "Hub",
+        "options": {}
```
petechd, line 26: Since big "Choose another section to complete" title on the hub doesn't make sense now, we should only make it visible(accessible) when the first section is complete, so:
```suggestion
        "options": {
            "required_completed_sections": [
                "section-1",
            ]
        }
```

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
```diff
@@ -0,0 +1,180 @@
...
+import CalculatedSummarySpendingPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/calculated-summary-spending.page.js";
+import CalculatedSummaryVisitsPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/calculated-summary-visits.page.js";
+import ListCollectorAddPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/list-collector-add.page";
+import DynamicAnswerPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/dynamic-answer.page";
+import Section1SummaryPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/section-1-summary.page";
+import ExtraSpendingMethodBlockPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/extra-spending-method-block.page";
+import ListCollectorRemovePage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/list-collector-remove.page";
+
+describe("Calculated summary with repeating answers", () => {
+  const currencyInputs = 'input[class="ons-input ons-input--text ons-input-type__input ons-input-number--w-20"]';
```
petechd, line 14: Can you make use of existing "inputs()" instead of defining new ones here?

app/questionnaire/value_source_resolver.py
```diff
@@ -100,12 +100,41 @@ def _resolve_list_item_id_for_value_source(
...
+    ) -> Generator[AnswerValueTypes, None, None]:
+        # Type ignore: the answer block will exist at this stage
+        question = self.schema.get_block_for_answer_id(answer_id).get("question", {})  # type: ignore
+
+        if not (dynamic_answers := question.get("dynamic_answers")):
+            return
+
+        values = dynamic_answers["values"]
+        if values["source"] == "list":
+            for list_item_id in self.list_store[values["identifier"]].items:
```
petechd, line 115: Does it not work the same without `.items`?

tests/integration/questionnaire/test_questionnaire_calculated_summary.py
```diff
@@ -186,3 +186,94 @@ def test_calculated_summary_value_sources_across_sections_repeating(self):
...
+        self.assertInBody(
+            "We calculate the total visits to the shop to be 12. Is this correct?"
+        )
+        self.assertNotInBody("How much do you spend on groceries at Tesco?")
+        self.assertNotInBody("How much do you spend on other items at Tesco?")
+        self.assertInBody("How many days a week do you shop at Tesco?")
+        self.post()
+        self.post()
+        self.post()
+        self.post({"weekly-car-trips-answer": "15"})
```
petechd, line 270: Can we add a functional test to show that we use answer from section 1 in section 2, this seems to be testing very general behaviour, why was it done in integration tests?

app/questionnaire/value_source_resolver.py
```diff
@@ -100,12 +100,41 @@ def _resolve_list_item_id_for_value_source(
             else None
         )
 
+    def resolve_dynamic_answers(
```
petechd, line 103: Does this need unit testing?

schemas/test/en/test_new_calculated_summary_repeating_and_static_answers.json
katie-gardner: Done 👍

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
```diff
@@ -0,0 +1,180 @@
+import HubPage from "../../../base_pages/hub.page";
+import AnySupermarketPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/any-supermarket.page.js";
+import ListCollectorPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/list-collector.page.js";
+import ExtraSpendingBlockPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/extra-spending-block.page.js";
+import CalculatedSummarySpendingPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/calculated-summary-spending.page.js";
+import CalculatedSummaryVisitsPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/calculated-summary-visits.page.js";
+import ListCollectorAddPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/list-collector-add.page";
+import DynamicAnswerPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/dynamic-answer.page";
+import Section1SummaryPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/section-1-summary.page";
```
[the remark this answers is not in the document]
katie-gardner: Section 2 is used now but still happy with that change - done 👍

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
katie-gardner: Didn't notice that option! great idea, fixed

app/questionnaire/value_source_resolver.py
katie-gardner: it does indeed, good point, removed .items

app/questionnaire/value_source_resolver.py
katie-gardner: Good idea - I decided it would probably be best to make this a private method but then update the existing test for value source resolver to cover this.

Annoyingly because MagicMock() is truthy, the is_answer_dynamic function keeps returning true in tests where its called by a Mock, so I had to mock out quite a few but all should be working now

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -164,16 +167,21 @@ def remove_list_item_and_answers(self, list_name: str, list_item_id: str) -> Non
 
         self._progress_store.remove_progress_for_list_item_id(list_item_id=list_item_id)
 
+        if remove_block_id := self._schema.get_remove_block_id_for_list(list_name):
```
berroar, line 170: Maybe better/clearer to handle the dependency capture in a separate method? As now this has been updated the title of this method is misleading in that it is no longer just removing from the progress store?

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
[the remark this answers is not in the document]
petechd: If it's used now it makes sense I guess so just keep it.

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
```diff
@@ -0,0 +1,200 @@
...
+    );
+    await assertSummaryValues(["3", "7"]);
+    await $(CalculatedSummaryVisitsPage.submit()).click();
+
+    await expect(await $(SummaryPage.supermarketsListLabel(1)).getText()).to.equal("Lidl");
+    await expect(await $(SummaryPage.supermarketsListLabel(2)).getText()).to.equal("Sainsburys");
+    await expect(await $(SummaryPage.supermarketsListLabel(3)).isExisting()).to.be.false;
+  });
+
+  it("Given I proceed to the second section, the calculated summary of days I visit the supermarket is used as a maximum to the first question", async () => {
```
berroar, line 181: Maybe worth adding an additional test to cover the status of the section being partially complete if you complete this section but go back and change the answer to the used as a maximum?

berroar (commented):
Looks good just some minor comments 👍

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
```diff
@@ -173,8 +173,28 @@ describe("Calculated summary with repeating answers", () => {
...
 
-    await expect(await $(Section1SummaryPage.supermarketsListLabel(1)).getText()).to.equal("Lidl");
-    await expect(await $(Section1SummaryPage.supermarketsListLabel(2)).getText()).to.equal("Sainsburys");
-    await expect(await $(Section1SummaryPage.supermarketsListLabel(3)).isExisting()).to.be.false;
+    await expect(await $(SummaryPage.supermarketsListLabel(1)).getText()).to.equal("Lidl");
+    await expect(await $(SummaryPage.supermarketsListLabel(2)).getText()).to.equal("Sainsburys");
+    await expect(await $(SummaryPage.supermarketsListLabel(3)).isExisting()).to.be.false;
+  });
+
+  it("Given I proceed to the second section, the calculated summary of days I visit the supermarket is used as a maximum to the first question", async () => {
```
petechd, line 181: Description would probably need to change to something like "...Then the correct error message is displayed" or we need to change the assertions? Check `guidancePanel(guidanceIndex)` or whatever it is that shows the right text?

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
```diff
@@ -173,8 +173,28 @@ describe("Calculated summary with repeating answers", () => {
...
+    await expect(await $(SupermarketTransportPage.singleErrorLink()).getText()).to.contain("Enter an answer less than or equal to 10");
+    await $(SupermarketTransportPage.weeklyCarTrips()).setValue(9);
+    await $(SupermarketTransportPage.submit()).click();
+    await expect(await browser.getUrl()).to.contain(SupermarketTransportCostPage.pageName);
+  });
+
+  it("Given I reach the final block, the calculated summary of dynamic answers is piped in correctly", async () => {
+    await $(SupermarketTransportCostPage.weeklyTripsCost()).setValue(30);
+    await $(SupermarketTransportCostPage.submit()).click();
+    await expect(await $("body").getText()).to.have.string("Total weekly supermarket spending: £380.00");
```
petechd, line 195: Maybe we should use this instead? `import CalculatedSummaryPipingPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/calculated-summary-piping.page.js";
`

petechd: So it is more explicit?

petechd:
You don't have massive gaps in test coverage but can we add one unit test in `tests/app/views/contexts/test_calculated_summary_context.py` to checks if `_remove_unwanted_questions_answers` transforms the block correctly? That it keeps the answers we want to keep and transforms the block correctly?

Yuyuutsu (commented):
Functionally tested, works as expected in the scenarios. Still looking through the changes 👍

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
katie-gardner: I can't see a way to do the assertion with the actual Page, but I've asserted that we're on that page just before to try and make that clearer - if you know another way I'd happily change it

tests/integration/questionnaire/test_questionnaire_calculated_summary.py
katie-gardner: I've added the test you suggested into `test_calculated_summary_context` so that the blocks being filtered out is more explicit than the integration test that got deleted - 100% coverage either way now

app/questionnaire/questionnaire_store_updater.py
katie-gardner: I like this idea 👍 it splits very cleanly into the removal and dependencies so made the new method as suggested

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
katie-gardner: done 👍

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_and_static_answers.spec.js
katie-gardner: Changed 👍

berroar (approved):
LGTM 👍

tests/integration/questionnaire/test_questionnaire_calculated_summary.py
```diff
@@ -186,3 +186,32 @@ def test_calculated_summary_value_sources_across_sections_repeating(self):
...
+        self.post(
+            {
+                f"cost-of-transport-{list_item_ids[0]}": "100",
+                f"cost-of-transport-{list_item_ids[1]}": "200",
+            }
+        )
+        self.assertInBody(
+            "We calculate the total monthly spending on public transport to be £300.00. Is this correct?"
+        )
+        self.post()
```
berroar, line 215: Do we need these post assert posts? 😅

katie-gardner: Great point 😅 - we do not 
deleted

app/questionnaire/value_source_resolver.py
```diff
@@ -100,12 +100,35 @@ def _resolve_list_item_id_for_value_source(
...
     ) -> ValueSourceEscapedTypes | ValueSourceTypes:
+        """resolves answer value by first checking if the answer is dynamic whilst not in a repeating section,
+        which indicates that it is a dynamic answer resolving to a list. Otherwise, retrieve answer value as normal.
+        """
         list_item_id = self._resolve_list_item_id_for_value_source(value_source)
         answer_id = value_source["identifier"]
 
+        # if not in a repeating section and the id is for a list of dynamic answers, then return the list of values
+        if not list_item_id and self.schema.is_answer_dynamic(answer_id):
+            return list(self._resolve_dynamic_answers(answer_id))
```
petechd, line 130: Correct me if I am wrong here, but if the only usage of `_resolve_dynamic_answers()` is here and we want a list then make that method return a list.

katie-gardner: Fair point - updated this, thanks 👍

petechd: I mean, generator is more universal but given that we only use it once. It's not some blocker change.

katie-gardner: Yeah I think you're right, yielding looked quite clean but theres no point in the generator if its its just going to be turned straight into a list

petechd (approved):

Yuyuutsu (approved):

katie-gardner:
[Looping 4.5 - Calculated summary using dynamic answers (3 SP)](https://trello.com/c/8njslssz/5721-looping-45-calculated-summary-using-dynamic-answers-3-sp)

mgdon650:
Looks fine to me :)

PR #1132, opened 2023-06-08
Type hinting for Views/Handlers

mgdon650:
As part of an ongoing effort to enforce type hinting throughout the codebase, the PR aims to introduce type hinting throughout the routes module and propagate these hints through the classes and files referenced by the routes module.

### How to review
All existing unit and functional tests pass in the same way they do on main.
All type hinting added conforms to[ type hinting guidelines.](https://github.com/ONSdigital/eq-questionnaire-runner/blob/main/doc/python-type-hinting.md)

### Checklist
* [ ] Unit tests pass
* [ ] Functional Tests Pass
* [ ] Adheres to typing guidelines

### DIFF
--- a/app/data_models/list_store.py
+++ b/app/data_models/list_store.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import random
 from functools import cached_property
 from string import ascii_letters
-from typing import Iterable, Iterator, Optional, TypedDict
+from typing import Iterable, Iterator, Optional, TypedDict, overload
 
 from structlog import get_logger
 
@@ -44,7 +44,15 @@ class ListModel:
     def __iter__(self) -> Iterator[str]:
         yield from self.items
 
+    @overload
     def __getitem__(self, list_item_index: int) -> str:
+        ...  # pragma: no cover
+
+    @overload
+    def __getitem__(self, list_item_index: slice) -> list[str]:
+        ...  # pragma: no cover
+
+    def __getitem__(self, list_item_index: slice | int) -> str | list[str]:
         return self.items[list_item_index]
 
     def __len__(self) -> int:
--- a/app/data_models/progress_store.py
+++ b/app/data_models/progress_store.py
@@ -3,6 +3,7 @@ from typing import Iterable, Iterator, MutableMapping, Optional
 
 from app.data_models.progress import Progress, ProgressDictType
 from app.questionnaire.location import Location
+from app.utilities.types import LocationType
 
 SectionKeyType = tuple[str, Optional[str]]
 
@@ -160,7 +161,7 @@ class ProgressStore:
 
         return []
 
-    def add_completed_location(self, location: Location) -> None:
+    def add_completed_location(self, location: LocationType) -> None:
         section_id = location.section_id
         list_item_id = location.list_item_id
 
@@ -185,7 +186,7 @@ class ProgressStore:
 
             self._is_dirty = True
 
-    def remove_completed_location(self, location: Location) -> bool:
+    def remove_completed_location(self, location: LocationType) -> bool:
         section_key = (location.section_id, location.list_item_id)
         if (
             section_key in self._progress
--- a/app/forms/questionnaire_form.py
+++ b/app/forms/questionnaire_form.py
@@ -26,6 +26,7 @@ from app.questionnaire.relationship_location import RelationshipLocation
 from app.questionnaire.rules.rule_evaluator import RuleEvaluator
 from app.questionnaire.value_source_resolver import ValueSourceResolver
 from app.utilities.mappings import get_flattened_mapping_values
+from app.utilities.types import LocationType
 
 logger = logging.getLogger(__name__)
 
@@ -456,7 +457,7 @@ def get_answer_fields(
     list_store: ListStore,
     metadata: MetadataProxy | None,
     response_metadata: MutableMapping,
-    location: Location | RelationshipLocation | None,
+    location: LocationType | None,
     progress_store: ProgressStore,
 ) -> dict[str, FieldHandler]:
     list_item_id = location.list_item_id if location else None
@@ -584,20 +585,23 @@ def _get_error_id(id_: str) -> str:
 
 
 def _clear_detail_answer_field(
-    form_data: MultiDict, question_schema: QuestionSchemaType
+    form_data: ImmutableMultiDict | MultiDict, question_schema: QuestionSchemaType
 ) -> MultiDict[str, Any]:
     """
     Clears the detail answer field if the parent option is not selected
     """
+    mutable_form_data = (
+        MultiDict(form_data) if isinstance(form_data, ImmutableMultiDict) else form_data
+    )
+
     for answer in question_schema.get("answers", []):
         for option in answer.get("options", []):
             if "detail_answer" in option and option["value"] not in form_data.getlist(
                 answer["id"]
             ):
-                if isinstance(form_data, ImmutableMultiDict):
-                    form_data = MultiDict(form_data)
-                form_data[option["detail_answer"]["id"]] = ""
-    return form_data
+                mutable_form_data[option["detail_answer"]["id"]] = ""
+
+    return mutable_form_data
 
 
 def generate_form(
@@ -608,9 +612,9 @@ def generate_form(
     metadata: MetadataProxy | None,
     response_metadata: MutableMapping,
     progress_store: ProgressStore,
-    location: Location | RelationshipLocation | None = None,
+    location: LocationType | None = None,
     data: dict[str, Any] | None = None,
-    form_data: MultiDict[str, Any] | None = None,
+    form_data: MultiDict | None = None,
 ) -> QuestionnaireForm:
     class DynamicForm(QuestionnaireForm):
         pass
--- a/app/views/contexts/calculated_summary_context.py
+++ b/app/views/contexts/calculated_summary_context.py
@@ -10,7 +10,6 @@ from app.jinja_filters import (
     format_unit,
     get_formatted_currency,
 )
-from app.questionnaire import Location
 from app.questionnaire.questionnaire_schema import (
     QuestionnaireSchema,
     get_calculated_summary_answer_ids,
@@ -20,6 +19,7 @@ from app.questionnaire.rules.rule_evaluator import RuleEvaluator
 from app.questionnaire.schema_utils import get_answer_ids_in_block
 from app.questionnaire.value_source_resolver import ValueSourceResolver
 from app.questionnaire.variants import choose_question_to_display, transform_variants
+from app.utilities.types import LocationType
 from app.views.contexts.context import Context
 from app.views.contexts.summary.calculated_summary_block import NumericType
 from app.views.contexts.summary.group import Group
@@ -36,7 +36,7 @@ class CalculatedSummaryContext(Context):
         metadata: MetadataProxy | None,
         response_metadata: MutableMapping,
         routing_path: RoutingPath,
-        current_location: Location,
+        current_location: LocationType,
         return_to: str | None = None,
         return_to_block_id: str | None = None,
     ) -> None:
--- a/app/views/contexts/summary/block.py
+++ b/app/views/contexts/summary/block.py
@@ -2,10 +2,11 @@ from typing import Any, Mapping, MutableMapping, Optional
 
 from app.data_models import AnswerStore, ListStore, ProgressStore
 from app.data_models.metadata_proxy import MetadataProxy
-from app.questionnaire import Location, QuestionnaireSchema
+from app.questionnaire import QuestionnaireSchema
 from app.questionnaire.rules.rule_evaluator import RuleEvaluator
 from app.questionnaire.value_source_resolver import ValueSourceResolver
 from app.questionnaire.variants import choose_variant
+from app.utilities.types import LocationType
 from app.views.contexts.summary.question import Question
 
 
@@ -19,7 +20,7 @@ class Block:
         metadata: Optional[MetadataProxy],
         response_metadata: MutableMapping,
         schema: QuestionnaireSchema,
-        location: Location,
+        location: LocationType,
         return_to: Optional[str],
         return_to_block_id: Optional[str] = None,
         progress_store: ProgressStore,
@@ -74,7 +75,7 @@ class Block:
         metadata: Optional[MetadataProxy],
         response_metadata: MutableMapping,
         schema: QuestionnaireSchema,
-        location: Location,
+        location: LocationType,
         return_to: Optional[str],
         return_to_block_id: Optional[str],
         progress_store: ProgressStore,
--- a/app/views/contexts/thank_you_context.py
+++ b/app/views/contexts/thank_you_context.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import Any, Optional, Union
+from typing import Any, Optional
 
 from flask import url_for
 from flask_babel import lazy_gettext
@@ -39,7 +39,9 @@ def build_thank_you_context(
     else:
         submission_text = lazy_gettext("Your answers have been submitted.")
     context_metadata = build_submission_metadata_context(
-        survey_type, submitted_at, metadata.tx_id
+        survey_type,
+        submitted_at,
+        metadata.tx_id,
     )
     return {
         "hide_sign_out_button": True,
@@ -54,8 +56,8 @@ def build_thank_you_context(
 
 def build_view_submitted_response_context(
     schema: QuestionnaireSchema, submitted_at: datetime
-) -> dict[str, Union[bool, str]]:
-    view_submitted_response: dict[str, Union[bool, str]] = {
+) -> dict[str, bool | str]:
+    view_submitted_response: dict[str, bool | str] = {
         "enabled": schema.is_view_submitted_response_enabled
     }
 
@@ -78,7 +80,7 @@ def build_census_thank_you_context(
     metadata: MetadataProxy,
     confirmation_email_form: Optional[EmailForm],
     form_type: str,
-) -> dict[str, Union[bool, str, None]]:
+) -> dict[str, bool | str | None]:
     context = {
         "display_address": metadata["display_address"],
         "form_type": form_type,
--- a/app/views/handlers/block.py
+++ b/app/views/handlers/block.py
@@ -1,9 +1,9 @@
 from datetime import datetime, timezone
 from functools import cached_property
-from typing import MutableMapping, Optional, Union
+from typing import Mapping, MutableMapping, Optional, Union
 
 from structlog import get_logger
-from werkzeug.datastructures import ImmutableDict
+from werkzeug.datastructures import ImmutableDict, ImmutableMultiDict
 
 from app.data_models import QuestionnaireStore
 from app.questionnaire.location import InvalidLocationException, Location
@@ -12,7 +12,9 @@ from app.questionnaire.questionnaire_schema import QuestionnaireSchema
 from app.questionnaire.questionnaire_store_updater import QuestionnaireStoreUpdater
 from app.questionnaire.relationship_location import RelationshipLocation
 from app.questionnaire.router import Router
+from app.questionnaire.routing_path import RoutingPath
 from app.utilities import safe_content
+from app.utilities.types import LocationType
 
 logger = get_logger()
 
@@ -23,9 +25,9 @@ class BlockHandler:
         schema: QuestionnaireSchema,
         questionnaire_store: QuestionnaireStore,
         language: str,
-        current_location: Union[Location, RelationshipLocation],
+        current_location: LocationType,
         request_args: MutableMapping,
-        form_data: MutableMapping,
+        form_data: ImmutableMultiDict,
     ):
         self._schema = schema
         self._questionnaire_store = questionnaire_store
@@ -50,19 +52,19 @@ class BlockHandler:
             )
 
     @property
-    def current_location(self):
+    def current_location(self) -> LocationType:
         return self._current_location
 
     @property
-    def return_to(self):
+    def return_to(self) -> str | None:
         return self._return_to
 
     @property
-    def return_to_block_id(self):
+    def return_to_block_id(self) -> str | None:
         return self._return_to_block_id
 
     @cached_property
-    def questionnaire_store_updater(self):
+    def questionnaire_store_updater(self) -> QuestionnaireStoreUpdater:
         return QuestionnaireStoreUpdater(
             self._current_location,
             self._schema,
@@ -72,7 +74,7 @@ class BlockHandler:
         )
 
     @cached_property
-    def placeholder_renderer(self):
+    def placeholder_renderer(self) -> PlaceholderRenderer:
         return PlaceholderRenderer(
             self._language,
             answer_store=self._questionnaire_store.answer_store,
@@ -85,7 +87,7 @@ class BlockHandler:
         )
 
     @cached_property
-    def router(self):
+    def router(self) -> Router:
         return Router(
             schema=self._schema,
             answer_store=self._questionnaire_store.answer_store,
@@ -95,12 +97,12 @@ class BlockHandler:
             response_metadata=self._questionnaire_store.response_metadata,
         )
 
-    def is_location_valid(self):
+    def is_location_valid(self) -> bool:
         return self.router.can_access_location(
             self._current_location, self._routing_path
         )
 
-    def get_previous_location_url(self):
+    def get_previous_location_url(self) -> str | None:
         return self.router.get_previous_location_url(
             self._current_location,
             self._routing_path,
@@ -109,7 +111,7 @@ class BlockHandler:
             self._return_to_block_id,
         )
 
-    def get_next_location_url(self):
+    def get_next_location_url(self) -> str:
         return self.router.get_next_location_url(
             self._current_location,
             self._routing_path,
@@ -118,7 +120,7 @@ class BlockHandler:
             self._return_to_block_id,
         )
 
-    def handle_post(self):
+    def handle_post(self) -> None:
         self._set_started_at_metadata()
         self.questionnaire_store_updater.add_completed_location()
         self.questionnaire_store_updater.remove_dependent_blocks_and_capture_dependent_sections()
@@ -126,7 +128,7 @@ class BlockHandler:
         self.questionnaire_store_updater.update_progress_for_dependent_sections()
         self.questionnaire_store_updater.save()
 
-    def _get_routing_path(self):
+    def _get_routing_path(self) -> RoutingPath:
         return self.router.routing_path(
             section_id=self._current_location.section_id,
             list_item_id=self._current_location.list_item_id,
@@ -134,7 +136,7 @@ class BlockHandler:
 
     def _update_section_completeness(
         self, location: Optional[Union[Location, RelationshipLocation]] = None
-    ):
+    ) -> None:
         location = location or self._current_location
 
         self.questionnaire_store_updater.update_section_status(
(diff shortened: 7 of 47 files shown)


Pipfile
```diff
@@ -72,6 +72,7 @@ google-cloud-tasks = "*"
 simplejson = "*"
 markupsafe = "*"
 pdfkit = "*"
+mypy = "*"
```
katie-gardner, line 75: You can revert this change - not needed

Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "65c2887dc528a01d63be44fc18d26e792c56501ddfb961c4cfb01a7ce7c925b7"
```
katie-gardner, line 4: No need for package changes so this file can be reverted entirely

app/views/contexts/calculated_summary_context.py
```diff
@@ -26,6 +26,7 @@
     QuestionnaireSchema,
```
katie-gardner, line 26: wont let me highlight above, but as you have changed the types to no longer use `Union` and `Optional` you also need to remove those from the typing imports

mypy.ini
```diff
@@ -6,30 +6,35 @@ warn_no_return = False
...
-[mypy-app.views.handlers.view_submitted_response_context]
-disallow_untyped_defs = True
-warn_return_any = True
-no_implicit_optional = True
+;[mypy-app.views.handlers.submit]
+;disallow_untyped_defs = True
+;warn_return_any = True
+;no_implicit_optional = True
+
+;[mypy-app.views.handlers.view_submitted_response_context]
```
katie-gardner, line 34: You can delete all these commented out ones now

katie-gardner (changes requested):
Having a proper read through now but just picked out the main things that spring to mind on a quick skim

app/views/handlers/content.py
```diff
@@ -28,24 +29,25 @@ def rendered_block(self):
...
+                self._schema.get_individual_response_list(),  # type: ignore
+                self._current_location.list_item_id,  # type: ignore
                 self._questionnaire_store,
             )
             if self._is_block_first_block_in_individual_response()
             else None,
         }
 
-    def _get_content_title(self, transformed_block):
+    def _get_content_title(self, transformed_block: dict) -> str | None:
```
katie-gardner, line 45: the transform_block is passed in by the Content class which is using transformed_variants to get it - so it'll be an `ImmutableDict`

app/views/handlers/question.py
```diff
@@ -135,25 +141,28 @@ def _get_answers_for_question(self, question_json) -> dict[str, Any]:
...
-    def _is_list_just_primary(self, list_items, list_name):
+    def _is_list_just_primary(self, list_items: list[str], list_name: str) -> bool:
         return (
             len(list_items) == 1
             and list_items[0]
             == self._questionnaire_store.list_store[list_name].primary_person
         )
 
-    def _get_answer_action(self):
+    def _get_answer_action(self) -> dict[str, str] | None:
```
katie-gardner, line 165: action can also have a property called "params" which is a dictionary so it would be `dict[str, str | dict]` which isn't really worth writing so you can change it to just `dict | None`

app/views/handlers/confirm_email.py
```diff
@@ -108,43 +110,51 @@ def question_schema(self) -> QuestionSchemaType:
...
+            self._session_store.session_data,  # type: ignore
+            # type ignore: metadata will be populated as we reach this stage
+            self._questionnaire_store.metadata,  # type: ignore
             self._schema,
         )
 
         try:
-            return current_app.eq["cloud_tasks"].create_task(
+            # type ignore: mypy not aware of eq attribute but it is a cloud task publisher
+            cloud_task_publisher: CloudTaskPublisher = current_app.eq["cloud_tasks"]  # type: ignore
```
katie-gardner, line 156: I've just realised this could be a `CloudTaskPublisher` or a `LogCloudTaskPublisher` see the setup.py, theres two potential options around line 370, I know we looked at this one, missed that 😁

app/views/handlers/block.py
```diff
@@ -135,24 +136,26 @@ def _update_section_completeness(
...
+    def _get_safe_page_title(self, page_title: str) -> str:
         page_title = self._schema.get_single_string_value(page_title)
         return safe_content(page_title)
 
-    def _resolve_custom_page_title_vars(self) -> MutableMapping:
+    def _resolve_custom_page_title_vars(self) -> dict[str, int]:
         list_item_position = self._questionnaire_store.list_store.list_item_position(
-            self.current_location.list_name, self.current_location.list_item_id
+            # Type ignore: no list_item_id or list_name is available yet
+            self.current_location.list_name,
```
katie-gardner, line 153: This comment doesn't quite make sense, do you mean other way round? you may need an ignore above as well

app/views/contexts/summary/block.py
```diff
@@ -74,7 +75,7 @@ def get_question(
         metadata: Optional[MetadataProxy],
         response_metadata: MutableMapping,
         schema: QuestionnaireSchema,
-        location: Location,
+        location: Location | RelationshipLocation,
```
katie-gardner, line 78: this affects summary/question.py so the change needs to be there too

katie-gardner (changes requested):
Still some errors which I think are the files you're looking at but great stuff! some comments so far 👍

app/views/handlers/confirm_email.py
mgdon650: I think it's CloudTaskPublisher as its calling Task from app.cloud_tasks.cloud_task_publishers.CloudTaskPublisher .

berroar:
[Type hints: Add type hints to the Handlers module (2 SP)](https://trello.com/c/aRYDKMTZ/5709-type-hints-add-type-hints-to-the-handlers-module-2-sp)

berroar:
Still 16 issues flagged by mypy that need to be resolved atm - mainly related to the changes for Location/Relationship Location

app/views/handlers/question.py
```diff
@@ -135,25 +141,28 @@ def _get_answers_for_question(self, question_json) -> dict[str, Any]:
...
         list_items = self._questionnaire_store.list_store[list_name].items
         section_id = self._schema.get_section_id_for_block_id(block_id)
 
         if self._is_list_just_primary(list_items, list_name) or not list_items:
             return Location(
-                section_id=section_id, block_id=block_id, list_name=list_name
+                # type ignore: section_id is always valid when this is called
+                section_id=section_id,  # type: ignore
+                block_id=block_id,
+                list_name=list_name,  # type: ignore
```
katie-gardner, line 155: you've got too many type ignores now, the `list_name` shouldn't need it

app/views/handlers/content.py
```diff
@@ -7,7 +8,7 @@
 
 class Content(BlockHandler):
     @cached_property
-    def rendered_block(self):
+    def rendered_block(self) -> dict:
```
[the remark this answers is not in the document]
katie-gardner: still outstanding

app/views/handlers/question.py
katie-gardner: still outstanding

app/views/handlers/content.py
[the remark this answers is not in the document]
katie-gardner: sorry @mgdon650 - I must have been looking at the transformed_block rather than the mutable version being returned, ignore this one

app/questionnaire/relationship_router.py
```diff
@@ -29,7 +30,7 @@ def __init__(
         self.unrelated_no_answer_values = unrelated_no_answer_values or []
         self.path = self._relationships_routing_path()
 
-    def can_access_location(self, location: RelationshipLocation) -> bool:
+    def can_access_location(self, location: Location | RelationshipLocation) -> bool:
```
katie-gardner, line 33: when would this be a `Location`?

app/views/handlers/submission.py
```diff
@@ -29,32 +34,37 @@ def get_receipting_metadata(metadata: MetadataProxy) -> dict:
...
 
-    def submit_questionnaire(self):
+    def submit_questionnaire(self) -> None:
         payload = self.get_payload()
 
         message = json_dumps(payload)
 
         encrypted_message = encrypt(
-            message, current_app.eq["key_store"], KEY_PURPOSE_SUBMISSION
+            message, current_app.eq["key_store"], KEY_PURPOSE_SUBMISSION  # type: ignore
```
katie-gardner, line 59: some missing comments for the #type ignores in this one

katie-gardner (changes requested):

app/questionnaire/relationship_router.py
katie-gardner: ahh.. I see the issue, the relationship router is part of relationship question inheriting from question from blockhandler where the location is either. Hmmm, not sure if better to stay as is or type ignore

mgdon650: Will keep to RelationshipLocation and type ignore the error to keep consistent to what we did yesterday

app/views/contexts/thank_you_context.py
```diff
@@ -75,12 +78,12 @@ def build_view_submitted_response_context(
...
 
 def build_census_thank_you_context(
-    metadata: MetadataProxy,
+    metadata: MetadataProxy | dict,
     confirmation_email_form: Optional[EmailForm],
     form_type: str,
 ) -> dict[str, Union[bool, str, None]]:
     context = {
-        "display_address": metadata["display_address"],
+        "display_address": metadata["display_address"],  # type: ignore
```
kylelawsonAND, line 86: Missing type ignore explanation

app/views/handlers/list_add_question.py
```diff
@@ -37,7 +36,7 @@ def handle_post(self):
 
         return super().handle_post()
 
-    def _resolve_custom_page_title_vars(self) -> MutableMapping:
+    def _resolve_custom_page_title_vars(self):  # type: ignore
```
kylelawsonAND, line 39: This is returning `dict[str, int]`?

app/views/handlers/list_action.py
```diff
@@ -1,39 +1,43 @@
...
+    def _get_routing_path(self) -> RoutingPath:
         return self.router.routing_path(section_id=self.parent_location.section_id)
 
-    def is_location_valid(self):
+    def is_location_valid(self) -> bool:
         can_access_parent_location = self.router.can_access_location(
             self.parent_location, self._routing_path
         )
 
         if (
```
berroar, line 31: Can be simplified to something like and not sure you need the ignore here?
```suggestion
         return bool(
            can_access_parent_location
            and self._current_location.list_name == self.parent_block["for_list"]
        )
```

app/views/handlers/list_action.py
```diff
@@ -50,20 +54,21 @@ def get_previous_location_url(self):
...
+    def get_next_location_url(self) -> str:
         if self._return_to == "section-summary":
             if self.router.can_display_section_summary(
                 self.parent_location.section_id, self.parent_location.list_item_id
             ):
                 return self.get_section_summary_url()
 
         if self.router.is_block_complete(
-            block_id=self.parent_location.block_id,
+            # type ignore: in this instance block_id is not none
```
berroar, line 70: I guess the comment should say why?

app/views/handlers/list_add_question.py
berroar, line 39: How come this one was changed? No comment here for explanation?

app/views/contexts/thank_you_context.py
mgdon650: I'm not sure why this is here as there is no error when i removed it. Will change

app/views/handlers/list_remove_question.py
```diff
@@ -3,46 +3,51 @@
...
             )
             self.capture_dependent_sections_for_list(self.parent_block["for_list"])
 
         return super().handle_post()
 
     def individual_response_enabled(self) -> bool:
-        return (
-            self.parent_block["for_list"] == self._schema.get_individual_response_list()
-        )
+        if self.parent_block["for_list"] == self._schema.get_individual_response_list():
```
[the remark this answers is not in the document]
berroar: Also why have we reverted from the simplified version?

app/views/handlers/question.py
```diff
@@ -35,7 +38,8 @@ def form(self):
                 metadata=self._questionnaire_store.metadata,
                 response_metadata=self._questionnaire_store.response_metadata,
                 location=self._current_location,
-                form_data=self._form_data,
+                # type ignore: this return is only passed to if there is form data.
```
berroar, line 41: Would move the comment above the call here

berroar (dismissed):

app/views/handlers/block.py
```diff
@@ -143,24 +144,26 @@ def _update_section_completeness(
...
 
-    def _get_safe_page_title(self, page_title):
+    def _get_safe_page_title(self, page_title: Mapping | str) -> str:
         page_title = self._schema.get_single_string_value(page_title)
         return safe_content(page_title)
 
     def _resolve_custom_page_title_vars(self) -> MutableMapping:
         list_item_position = self._questionnaire_store.list_store.list_item_position(
-            self.current_location.list_name, self.current_location.list_item_id
+            # Type ignore: list_item_id and list_name are populated at this stage
```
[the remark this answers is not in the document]
mgdon650: Its a linting issue. Its moving comments about the place

app/views/handlers/list_remove_question.py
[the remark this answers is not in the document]
mgdon650: It was throwing a "Returning Any from function declared to return bool" error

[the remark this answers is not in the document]
katie-gardner: since mypy doesn't know what the types for the immutable dictionary are, and so doesn't know how the `__eq__ `method is implemented, but casting for the sake of type hinting is avoided, but type hinting documentation states to only type ignore where its unavoidable, could this be a middle ground?
```
list_name: str = self.parent_block["for_list"]
return list_name == self._schema.get_individual_response_list()
```
I'm wondering if we need to document some sort of agreement on this type of scenario where it's obviously a boolean, and the above is longer than the simplified version, and we _can_ avoid a type ignore, but maybe don't want to in this case?

app/views/handlers/list_add_question.py
mgdon650: It was a linting error. Have updated after Kyles comment

app/views/contexts/thank_you_context.py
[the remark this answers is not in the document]
mgdon650: Think this was linting again.

app/views/handlers/confirmation_email.py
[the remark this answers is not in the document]
mgdon650: Updated. Will appear in next commit

app/views/handlers/confirmation_email.py
```diff
@@ -69,10 +68,8 @@ def get_page_title(self):
 
     @staticmethod
     def is_limit_reached(session_data: SessionData) -> bool:
-        return (
-            session_data.confirmation_email_count
-            >= current_app.config["CONFIRMATION_EMAIL_LIMIT"]
-        )
+        conf_email_limit: int = current_app.config["CONFIRMATION_EMAIL_LIMIT"]
```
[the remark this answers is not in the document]
mgdon650: 👍

app/views/handlers/individual_response.py
```diff
@@ -121,7 +121,8 @@ def __init__(
         self._form_data = form_data
         self._answers: dict[str, Any] | None = None
         self._list_item_id = list_item_id
-        self._list_name = self._schema.get_individual_response_list()
+        # Type ignore: in individual_response for_list is required which is where list_name comes from
```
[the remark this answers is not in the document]
mgdon650: I have tried changing the code but the unit tests are failing which would mean more of an investigation into the tests/code which would be time consuming and not as critical of an issue considering individual_response isn't used.

Will leave as errors for now so that they do not get missed.

app/views/handlers/individual_response.py
```diff
@@ -194,8 +195,9 @@ def router(self) -> Router:
         )
 
     @cached_property
-    def individual_section_id(self) -> str | None:
-        return self._schema.get_individual_response_individual_section_id()
+    def individual_section_id(self) -> str:
+        # Type ignore: Current usage of this method occur when individual section ID exists. Required element
```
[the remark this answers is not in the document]
mgdon650: same as my response above

app/views/handlers/individual_response.py
```diff
@@ -701,8 +715,16 @@ def handle_post(self) -> Response:
...
-        self._list_name = schema.get_individual_response_list()
+    def __init__(
+        self,
+        schema: QuestionnaireSchema,
+        questionnaire_store: QuestionnaireStore,
+        language: str,
+        request_args: dict[str, str],
+        form_data: ImmutableMultiDict[str, str],
+    ):
+        # Type ignore:  in individual_response for_list is required which is where list_name comes from.
```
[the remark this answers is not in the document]
mgdon650: same as above

app/views/handlers/list_action.py
mgdon650: Nice! Have changed this

app/views/handlers/list_remove_question.py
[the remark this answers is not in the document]
berroar: Yeah probably need to discuss this, I added this comment thinking more about not amending code for the sake of type hints, but can see where you are coming from with this hint technically being avoidable . Maybe one to raise in a tech session then and as a team we can agree whether we need to update the doc.

app/views/contexts/thank_you_context.py
```diff
@@ -39,7 +39,10 @@ def build_thank_you_context(
     else:
         submission_text = lazy_gettext("Your answers have been submitted.")
     context_metadata = build_submission_metadata_context(
-        survey_type, submitted_at, metadata.tx_id
+        survey_type,
+        submitted_at,
+        # type ignore: tx_id always exists at this point
+        metadata.tx_id,  # type: ignore
```
katie-gardner, line 45: If metadata is a `dict` then calling `metadata.tx_id` is going to error, might be one for @berroar to confirm as I think the change was something you both talked about? would it be better to change `metadata = get_metadata(current_user) or {}` into defaulting to another proxy or just type ignoring it?

tests/app/test_jinja_filters.py
```diff
@@ -153,9 +153,6 @@ def test_format_percentage(percentage, formatted_percentage):
         ("duration-year", 100, "short", "100 bl", "cy"),
         ("duration-hour", 100, "long", "100 awr", "cy"),
         ("duration-year", 100, "long", "100 mlynedd", "cy"),
-        ("mass-metric-ton", 100, "long", "100 tonnes", "en_GB"),
-        ("mass-metric-ton", 1, "long", "1 tonne", "en_GB"),
-        ("mass-metric-ton", 100, "short", "100 t", "en_GB"),
```
katie-gardner, line 158: I think you've deleted this by accident

app/views/handlers/submission.py
```diff
@@ -56,14 +56,20 @@ def submit_questionnaire(self) -> None:
...
+            KEY_PURPOSE_SUBMISSION,
         )
-
+        # type ignore: metadata must be present at time of submission
         additional_metadata = get_receipting_metadata(self._metadata)  # type: ignore
 
+        # type ignore: current_app can return empty Local Proxy. Similar to other files, this is ignored.
         submitted = current_app.eq["submitter"].send_message(  # type: ignore
             encrypted_message,
+            # type ignore: case_id exists at this point
```
katie-gardner, line 70: I think it would be more correct to say that the metadata exists, in a single type ignore comment, as those are attributes that will always exist on the metadata provided the metadata is not None

app/views/handlers/submission.py
```diff
@@ -56,14 +56,20 @@ def submit_questionnaire(self) -> None:
...
             case_id=self._metadata.case_id,  # type: ignore
+            # type ignore: tx_id exists at this point
```
katie-gardner, line 72: (likewise)

katie-gardner (changes requested):
Just 4 minor comments from the changes

tests/app/test_jinja_filters.py
mgdon650: Hmm weird, I didn't even touch that myself....

app/data_models/list_store.py
```diff
@@ -44,7 +44,15 @@ def __eq__(self, other: object) -> bool:
     def __iter__(self) -> Iterator[str]:
         yield from self.items
 
+    @overload
     def __getitem__(self, list_item_index: int) -> str:
+        ...  # pragma: no cover
+
+    @overload
+    def __getitem__(self, list_item_index: slice) -> list[str]:
```
MebinAbraham, line 52: Where do we use a non int look up?

app/data_models/progress_store.py
```diff
@@ -160,7 +161,7 @@ def get_completed_block_ids(
 
         return []
 
-    def add_completed_location(self, location: Location) -> None:
+    def add_completed_location(self, location: Location | RelationshipLocation) -> None:
```
MebinAbraham, line 164: Given `Location | RelationshipLocation` used like 40 times, I suggest using an type alias.

app/views/contexts/thank_you_context.py
MebinAbraham: This method should expect a metadata proxy. It is not valid to call it without one. Therefore, we should change `metadata = get_metadata(current_user) or {}` to 
`metadata: MetadataProxy= get_metadata(current_user)  # type ignore`

app/views/handlers/confirmation_email.py
```diff
@@ -26,12 +25,13 @@ def __init__(
...
         schema: QuestionnaireSchema,
-        page_title: Optional[str] = None,
-        serialised_email: Optional[str] = None,
+        page_title: str | None = None,
+        serialised_email: str | None = None,
     ):
         if not self.is_enabled(schema):
             raise ConfirmationEmailNotEnabled
 
+        # type ignore: session_data is populated at login therefore won't be None for confirmation email
```
MebinAbraham, line 34: Minor, use caps `# Type ignore:` as per doc and for consistency. Applies to all.

app/views/handlers/confirmation_email.py
```diff
@@ -41,11 +41,11 @@ def __init__(
         self._serialised_email = serialised_email
 
     @property
-    def page_title(self):
+    def page_title(self) -> str | LazyString | None:
```
MebinAbraham, line 44: Do you need `LazyString` when you got `str`?

app/views/handlers/content.py
```diff
@@ -28,24 +30,25 @@ def rendered_block(self):
             list_item_id=self._current_location.list_item_id,
         )
 
-    def get_context(self):
+    def get_context(self) -> dict | None:
```
MebinAbraham, line 33: This never returns None?

app/views/handlers/individual_response.py
```diff
@@ -133,7 +133,7 @@ def __init__(
     @cached_property
     def _list_model(self) -> ListModel:
         # Type ignore: Current usages of this cached property occur when List Name will exist and be not None
```
MebinAbraham, line 135: This comment can be removed?

app/views/handlers/question.py
```diff
@@ -19,14 +19,18 @@
 
 class Question(BlockHandler):
     @staticmethod
-    def _has_redirect_to_list_add_action(answer_action):
-        return answer_action and answer_action["type"] == "RedirectToListAddBlock"
+    def _has_redirect_to_list_add_action(answer_action: Mapping | None) -> bool:
+        return (
+            answer_action is not None
```
MebinAbraham, line 24: Did this need changing?. Previous one would pass with empty dict, this would error on KeyError.

app/views/handlers/question.py
```diff
@@ -35,7 +39,7 @@ def form(self):
                 metadata=self._questionnaire_store.metadata,
                 response_metadata=self._questionnaire_store.response_metadata,
                 location=self._current_location,
-                form_data=self._form_data,
+                form_data=self._form_data,  # type: ignore
```
MebinAbraham, line 42: This warning is valid?

app/views/handlers/submission.py
MebinAbraham: Each individual type ignore can also be removed, if you add a line like:

```py
metadata: MetadatProxy = self._metadata  # type: ignore
```

Then using metadata.case_id tx_id etc won't raise a mypy error..

app/views/handlers/question.py
```diff
@@ -116,7 +121,7 @@ def get_next_location_url(self):
             self._return_to_block_id,
         )
 
-    def _get_answers_for_question(self, question_json) -> dict[str, Any]:
+    def _get_answers_for_question(self, question_json: dict) -> dict:
```
MebinAbraham, line 124: Arguments should be abstract. Unless this needs dict specific methods, use `Mapping`.

app/views/handlers/question.py
mgdon650: Mypy didn't like it but instead of changing the code I could have written the following line actually. 

```return bool(answer_action and answer_action["type"] == "RedirectToListAddBlock")```

File: app/data_models/list_store.py (hunk 29 above)
Comment (mgdon650) [reply app/data_models/list_store.py:52]: In :

```app/views/handlers/relationships/unrelated_question.py:44: error: Incompatible return value type (got "Union[str, List[str]]", expected "List[str]")  [return-value]
```

app/views/handlers/confirmation_email.py
mgdon650: When we were going through this, mypy didn't like that lazy_gettext("Confirmation email") returned a LazyString hence why we included it. Re-visiting it now, it doesn't throw and error so will remove it

MebinAbraham (approved):
Approving with some minor comment, will leave it to the other reviews to check the remaining ones on my behalf.

app/views/handlers/calculated_summary.py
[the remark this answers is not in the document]
mgdon650: I think it's something to do with the merge. Because that was deleted ages ago? Reviewing the commit this morning as some other odd bits that have overwritten too.

app/views/handlers/confirmation_email.py
```diff
@@ -41,11 +41,11 @@ def __init__(
         self._serialised_email = serialised_email
 
     @property
-    def page_title(self):
+    def page_title(self) -> str | None:
```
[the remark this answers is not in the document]
katie-gardner: Have checked this is fixed 👍

app/views/contexts/thank_you_context.py
```diff
@@ -39,7 +39,10 @@ def build_thank_you_context(
     else:
         submission_text = lazy_gettext("Your answers have been submitted.")
     context_metadata = build_submission_metadata_context(
-        survey_type, submitted_at, metadata.tx_id
+        survey_type,
+        submitted_at,
+        # type ignore: tx_id always exists at this point
```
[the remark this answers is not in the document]
katie-gardner: Checked and these have been mass updated

app/views/contexts/thank_you_context.py
katie-gardner: @mgdon650 - this one is still outstanding I think you can actually just delete the type ignore? now you've fixed the other place it shouldn't be required

app/views/handlers/list_remove_question.py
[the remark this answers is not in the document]
katie-gardner: Will make a note about updating this in tech session and resolve for now

app/forms/questionnaire_form.py
```diff
@@ -584,19 +585,22 @@ def _get_error_id(id_: str) -> str:
...
         for option in answer.get("options", []):
             if "detail_answer" in option and option["value"] not in form_data.getlist(
                 answer["id"]
             ):
-                if isinstance(form_data, ImmutableMultiDict):
-                    form_data = MultiDict(form_data)
-                form_data[option["detail_answer"]["id"]] = ""
+                mutable_form_data[option["detail_answer"]["id"]] = ""
+
     return form_data
```
katie-gardner, line 604: you've switched the variable name to `mutable_form_data` but forgotten to return that instead

app/views/handlers/individual_response.py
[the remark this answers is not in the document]
katie-gardner: I think this one is controversial - `get_individual_response_list` is a method on `questionnaire_schema` which can indeed return None, for most schemas, however its only ever used here where it can't. I think the method ought to move, but I don't think its in the scope of the ticket, so will resolve based on it being a largely unused file and it leaves the typing in `questionnaire_schema` technically more correct... I think

katie-gardner (approved):

app/views/handlers/individual_response.py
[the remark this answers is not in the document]
kylelawsonAND: I think it's worth leaving as is since the scope of this ticket would grow further - my suggestion is resolve this comment, but make a ticket to investigate it as it's own issue. Should hopefully be a small one to pick up at a later date.

app/views/handlers/individual_response.py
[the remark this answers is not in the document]
kylelawsonAND: Add as part of the investigation ticket as mentioned in prev comment.

app/views/handlers/individual_response.py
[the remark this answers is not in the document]
kylelawsonAND: same suggestion as prev

kylelawsonAND (approved):

PR #1138, opened 2023-06-13
Bug Fix: `get_width_for_number` function when using value sources

rmccar:
This PR creates a min max map from answers that have a min and/or max set from another question, compares the answer ids to that and if there is a match it will get the correct min max value from the correct source.

### How to review
 - Launch test_numbers.json schema
 - Enter a min / max and submit
 - Compare the size of the range question inputs between this branch and the main branch

### Checklist


rmccar:
[Bug: get_width_for_number function does not work when using value sources](https://trello.com/c/joHAGFtd/5688-bug-getwidthfornumber-function-does-not-work-when-using-value-sources)

app/jinja_filters.py
```diff
@@ -256,9 +256,25 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
...
 def get_width_for_number(answer: AnswerType) -> Optional[int]:
     allowable_widths = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 15, 20, 30, 40, 50]
+    min_value = None
+    max_value = None
 
-    min_value = answer.get("minimum", {}).get("value", 0)
-    max_value = answer.get("maximum", {}).get("value", MAX_NUMBER)
+    schema = g.get("schema")
+
+    if answer.get("minimum", {}) and isinstance(answer["minimum"]["value"], dict):
```
katie-gardner, line 264: @petechd I was thinking about this from the conversation in tech session - what if it was something on these lines?

```
min_value = answer.get("minimum", {}).get("value")
while isinstance(min_value, dict):
    source = schema.get_answers_by_answer_id(min_value["identifier"])
    min_value = source[0].get("minimum", {}).get("value")
```

and same for max_value - it doesn't account for finding multiple variant answers, but as a general idea I think it'd be possible to do this without the `min_and_max_map` dictionary
(also that check might need to be more careful than just seeing if its a dictionary, but just an initial idea)

petechd: I am all for this idea, as it seems better, but I am not sure if we can do something that is different from the initial ask on the card.

katie-gardner: ahh I hadn't checked the comment about making it at schema load time, perhaps suggested that way for performance? in which case yeah this might not be viable

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +95,50 @@ def __init__(
         self._populate_answer_dependencies()
         self._populate_when_rules_section_dependencies()
         self._populate_calculated_summary_section_dependencies()
+        self._populate_min_max_map()
 
     @cached_property
     def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
         return ImmutableDict(self._answer_dependencies_map)
 
+    # pylint: disable=too-many-nested-blocks
```
berroar, line 104: We shouldn't just add a pylint ignores for things like this, we should try to break the method down or separate it out?

berroar (changes requested):

app/questionnaire/questionnaire_schema.py
katie-gardner: since this function call comes after `self._answers_by_id` has been set - could you iterate over that instead? then it  could just be 
```
for answer_id, answer in self._answers_by_id.items():
  if (answer_type := ....
````

Comment (petechd) [reply app/questionnaire/questionnaire_schema.py:104]: > since this function call comes after `self._answers_by_id` has been set - could you iterate over that instead? then it could just be
> 
> ```
> for answer_id, answer in self._answers_by_id.items():
>   if (answer_type := ....
> ```

This will simplify it a lot.

File: tests/app/test_jinja_filters.py [hunk 3]
```diff
@@ -253,10 +255,22 @@ def test_format_duration(duration, formatted_duration, app):
             {"maximum": {"value": 123456789012345678901123456789012345678901234567890}},
             None,
         ),
+        ({"maximum": {"value": {"identifier": "set-maximum", "source": "answers"}}}, 5),
+        (
+            {"minimum": {"value": {"identifier": "set-minimum", "source": "answers"}}},
+            15,
```
petechd, line 261: Input field getting default 15 when no maximum is bad (when resolved minimum is 5), we should decide how to change the logic to avoid that, or is minimum and maximum always required?

app/jinja_filters.py
```diff
@@ -256,28 +256,25 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
...
-        ) in schema.min_and_max_map:
-            max_value = schema.min_and_max_map[identifier] or None
-    if not min_value:
-        min_value = answer.get("minimum", {}).get("value", 0)
-    if not max_value:
-        max_value = answer.get("maximum", {}).get("value", MAX_NUMBER)
-
-    min_value_width = len(str(min_value))
-    max_value_width = len(str(max_value))
+    for item in minMaxList:
```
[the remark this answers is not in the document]
katie-gardner: though maybe a function that accepts "minimum" or "maximum" would be simpler than using a dictionary

app/jinja_filters.py
```diff
@@ -253,15 +253,27 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
...
+    min_max_value = None
+
+    if answer.get(min_max, {}) and isinstance(answer[min_max]["value"], dict):
+        if (
+            identifier := answer[min_max]["value"].get("identifier", None)
+        ) in schema.min_and_max_map:
+            min_max_value = schema.min_and_max_map[identifier] or None
+
+    if not min_max_value:
+        min_max_value = answer.get(min_max, {}).get("value", default_value)
```
katie-gardner, line 267: and if you returned in the if statement you wouldn't need to check `min_max_value` here
it could be just 
`return answer.get(...`

app/jinja_filters.py
```diff
@@ -253,15 +253,27 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
     return {"should_wrap_with_fieldset": should_wrap_with_fieldset}
 
 
+def get_min_max_values(min_max: str, answer: AnswerType, default_value: int) -> int:
```
katie-gardner, line 256: could do `min_max: Literal["minimum", "maximum"]` for extra safety

app/jinja_filters.py
```diff
@@ -253,15 +253,26 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
...
 
-    min_value = answer.get("minimum", {}).get("value", 0)
-    max_value = answer.get("maximum", {}).get("value", MAX_NUMBER)
-
-    min_value_width = len(str(min_value))
-    max_value_width = len(str(max_value))
+    min_value_width = get_min_max_values("minimum", answer, 0)
+    max_value_width = get_min_max_values("maximum", answer, MAX_NUMBER)
 
     width = min_value_width if min_value_width > max_value_width else max_value_width
```
Yuyuutsu, line 277: Just a suggestion, since this isn't part of the change. Could we use  `width = max(min_value_width, max_value_width)` instead. Makes it a bit clearer for me, but I don't mind if you prefer the if statement.

katie-gardner (changes requested):
I've realised this doesn't work if you have 
answer1 -> minimum = 10
answer2 -> minimum = answer1
answer3 -> minimum = answer2

because `min_max_map` ends up having an immutable dictionary as the value for answer3
but since the answers should be in order, if you find an answer that depends on another answer, it should already be in the map

so for answer3, you could just look up the value for answer2 in `min_max_map` - will look at this more closely tomorrow

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +95,39 @@ def __init__(
...
         self._populate_calculated_summary_section_dependencies()
+        self._populate_min_max_map()
 
     @cached_property
     def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
         return ImmutableDict(self._answer_dependencies_map)
 
+    def _populate_min_max_map(self) -> None:
+        for answer_id, answer in self._answers_by_id.items():
+            if (answer_type := answer[0].get("type")) and answer_type not in [
```
katie-gardner, line 106: this works for answers that are 1 reference deep, to include others, maybe you could have a populate minmax for answer function similar to the filter one and do something like this
```
if value := answer.get(min_max, {}).get("value"):
    if isinstance(value, int):
        self.min_and_max_map[answer["id"]] = value
    elif isinstance(value, dict):
        if value.get("source") == "answers":
            self.min_and_max_map[answer["id"]] = self.min_and_max_map[value["identifier"]]
```
which might make other part simlar, or if having every answer is overkill, skip the int case

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +95,39 @@ def __init__(
...
+                "Date",
+                "MonthYearDate",
+                "YearDate",
+            ]:
+                answer_to_search_for_min_max = self.get_answers_by_answer_id(answer_id)[
```
berroar, line 111: Why are we searching for answers by id here when are already in the context of the answer from the loop above?

kylelawsonAND: +1

app/jinja_filters.py
```diff
@@ -253,17 +253,28 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
...
 
 
+def get_min_max_values(
+    min_max: Literal["minimum", "maximum"], answer: AnswerType, default_value: int
+) -> int:
+    schema = g.get("schema")
+
+    if answer.get(min_max, {}) and isinstance(answer[min_max]["value"], dict):
+        if (
+            identifier := answer[min_max]["value"].get("identifier", None)
```
kylelawsonAND, line 263: .get("identifier", None) can just be .get("identifier") as .get() returns None by default.

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +95,36 @@ def __init__(
...
 
     @cached_property
     def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
         return ImmutableDict(self._answer_dependencies_map)
 
+    def _populate_min_max_map(self) -> None:
+        for answer_id, answers in self._answers_by_id.items():
+            for min_max in ["minimum", "maximum"]:
+                if (
+                    value := answers[0].get(min_max, {}).get("value")
```
katie-gardner, line 108: the `value := answers[0].get(min_max, {}).get("value")` needs to be in brackets - at the moment its assigning value to `True` as its setting it equal to the whole expression

If there were multiple options for an answer_id (detailed options maybe?) does this need to iterate over all of those instead of `answer[0]`?

petechd: This code @katie-gardner suggested was already working, we probably need to revert to it and just tackle complexity by delegating to some helper method:
```python
    def _populate_min_max_map(self) -> None:
        for answer_id, answers in self._answers_by_id.items():
            if (answer_type := answers[0].get("type")) and answer_type not in [
                "Date",
                "MonthYearDate",
                "YearDate",
            ]:
                for min_max in ["minimum", "maximum"]:
                    if value := answers[0].get(min_max, {}).get("value"):
                        if isinstance(value, int):
                            self.min_and_max_map[answer_id] = str(value)
                        elif isinstance(value, dict) and value:
                            if (
                                value.get("source") == "answers"
                                and value["identifier"] in self.min_and_max_map
                            ):
                                self.min_and_max_map[answer_id] = self.min_and_max_map[
                                    value["identifier"]
                                ]
```
and then test would be:

```python
def test_schema_min_max_populate():
    schema = load_schema_from_name("test_numbers")
    assert schema.min_and_max_map == {
        "set-minimum": "1000",
        "set-maximum": "10000",
        "test-range": "10000",
        "test-range-exclusive": "10000",
        "test-min": "-123",
        "test-max": "1234",
        "test-min-exclusive": "123",
        "test-max-exclusive": "1234",
        "test-percent": "100",
        "test-decimal": "10000",
        "other-answer": "10000",
    }
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +95,37 @@ def __init__(
...
+    def _append_to_min_max_map(
+        self, min_max: Literal["minimum", "maximum"], answer_id: str, answers: list
+    ) -> None:
+        if value := answers[0].get(min_max, {}).get("value"):
+            if isinstance(value, int):
+                self.min_and_max_map[answer_id] = str(value)
+            elif isinstance(value, dict) and value:
+                if (
+                    value.get("source") == "answers"
+                    and value["identifier"] in self.min_and_max_map
```
MebinAbraham, line 113: 1. When would `value["identifier"]` not be in `self.min_and_max_map`?
2. Why would `value["identifier"]` not be in `self.min_and_max_map`?
This is going to skip it, instead of working it out?

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +95,37 @@ def __init__(
...
+                ):
+                    self.min_and_max_map[answer_id] = self.min_and_max_map[
+                        value["identifier"]
+                    ]
+
+    def _populate_min_max_map(self) -> None:
+        for answer_id, answers in self._answers_by_id.items():
+            if (answer_type := answers[0].get("type")) and answer_type not in [
```
MebinAbraham, line 121: `type` is mandatory so you don't need to use `get`

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +95,37 @@ def __init__(
...
+                "Date",
+                "MonthYearDate",
+                "YearDate",
+            ]:
```
MebinAbraham, line 125: Why these types?. Explicitly check types that allow min and max?. This would check things like Checkbox etc.

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +95,37 @@ def __init__(
         self._populate_answer_dependencies()
         self._populate_when_rules_section_dependencies()
         self._populate_calculated_summary_section_dependencies()
+        self._populate_min_max_map()
```
MebinAbraham, line 98: Maybe name it, `populate_numeric_ranges` or `populate_mix_max_for_numeric_answers`?

app/jinja_filters.py
```diff
@@ -253,17 +253,28 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
     return {"should_wrap_with_fieldset": should_wrap_with_fieldset}
 
 
+def get_min_max_values(
+    min_max: Literal["minimum", "maximum"], answer: AnswerType, default_value: int
+) -> int:
+    schema = g.get("schema")
```
MebinAbraham, line 259: This should be inside the if statement, only fetch if needed.

MebinAbraham (commented):
Can we update the PR title to be meaningful.

app/questionnaire/questionnaire_schema.py
rmccar: You're right, this was added yesterday to get all the schemas to pass. I took it out and investigated it and it turns out there is an MBS schema that has an answer thats referenced that doesn't actually have a maximum value so I think its an issue with that schema that needs to be fixed.

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +96,41 @@ def __init__(
...
         self._populate_when_rules_section_dependencies()
         self._populate_calculated_summary_section_dependencies()
+        self._populate_mix_max_for_numeric_answers()
 
     @cached_property
     def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
         return ImmutableDict(self._answer_dependencies_map)
 
+    def _append_to_min_max_map(
+        self, min_max: Literal["minimum", "maximum"], answer_id: str, answers: list
```
katie-gardner, line 106: since you're only ever using `answers[0]` you could pass this into the function instead of the list - minor, not fussed

(Why is it that we only need `answers[0]`? I did try editing the schema to have multiple detail option answers and this still worked but not sure if this would get hit by some edge case somewhere)

app/questionnaire/questionnaire_schema.py
```diff
@@ -81,6 +82,7 @@ def __init__(
         ] = defaultdict(set)
         self._language_code = language_code
         self._questionnaire_json = questionnaire_json
+        self.min_and_max_map: dict[str, str] = {}
```
katie-gardner, line 85: Might be more consistent with the other attributes if this was private and there was a cached property that returned an immutable version of the map

app/questionnaire/questionnaire_schema.py
MebinAbraham: We have a feature called "variants" which means we can have multiple variants for 1 answer. The variant to use is resolved at run time, it is possible 1 answer id might have 2 answers where only 1 has a min/max or a combination. Therefore, the best option is probably to use the largest range. i.e. if answer 1 has no min/max, then the default min limit is 0, the default max is system limit, but if answer 2 has min/max set to -1000 and 1000, we should be using the min from answer 2 (as it is longer) and max from answer 1. Let me know if it needs further explanation.

rmccar: I will need some more logic for that then and for that I think I will still need to pass the list in

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +96,71 @@ def __init__(
...
+                    and value["identifier"] in self._min_and_max_map
+                ):
+                    if longest_string and len(
+                        str(self._min_and_max_map[value["identifier"]][min_max])
+                    ) > len(longest_string):
+                        longest_string = str(
+                            self._min_and_max_map[value["identifier"]][min_max]
+                        )
+                    else:
+                        longest_string = str(
```
berroar, line 138: The code in the if/else here seems to be doing the same thing? Is the if/else needed?

app/jinja_filters.py
```diff
@@ -253,17 +253,27 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
...
 
 
+def get_min_max_values(
+    min_max: Literal["minimum", "maximum"], answer: AnswerType, default_value: int
+) -> int:
+    if answer.get(min_max, {}) and isinstance(answer[min_max]["value"], dict):
+        schema = g.get("schema")
+        if (
+            identifier := answer[min_max]["value"].get("identifier")
+        ) in schema.min_and_max_map:
```
MebinAbraham, line 263: When would the identifier not be in `schema.min_and_max_map`?. You need to instead check it is an answer source.

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +96,75 @@ def __init__(
...
 
     @cached_property
     def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
         return ImmutableDict(self._answer_dependencies_map)
 
+    @cached_property
+    def min_and_max_map(self) -> ImmutableDict[str, str | dict[str, str]]:
+        return ImmutableDict(self._min_and_max_map)
+
+    def _set_min_max_value_for_answer(
```
MebinAbraham, line 109: Function is misnamed, it doesn't set anything?. `get_mix_max...`

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +96,75 @@ def __init__(
...
+        if isinstance(value, int):
+            if longest_string and len(str(value)) > len(longest_string):
+                return str(value)
+            elif not longest_string:
+                return str(value)
+
+        elif isinstance(value, dict) and value:
+            if (
+                value.get("source") == "answers"
+                and value["identifier"] in self._min_and_max_map
```
MebinAbraham, line 121: When is the identifier not in min max map?. You add it in the create_min_max map.

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +96,75 @@ def __init__(
...
+    ) -> None:
+        longest_string = None
+        for answer in answers:
+            value = answer.get(min_max, {}).get("value")
+            returned_longest_string = self._set_min_max_value_for_answer(value, min_max, longest_string)
+            if returned_longest_string:
+                longest_string = returned_longest_string
+
+        if not self._min_and_max_map.get(answer_id, {}):
+            self._min_and_max_map[answer_id] = {}
```
MebinAbraham, line 149: You can avoid this by using a `defaultdict`

app/questionnaire/questionnaire_schema.py
```diff
@@ -94,11 +96,75 @@ def __init__(
...
+                if longest_string and len(
+                    str(self._min_and_max_map[value["identifier"]][min_max])
+                ) > len(longest_string):
+                    return str(
+                        self._min_and_max_map[value["identifier"]][min_max]
+                    )
+                elif not longest_string:
+                    return str(
+                        self._min_and_max_map[value["identifier"]][min_max]
+                    )
```
MebinAbraham, line 132: Any reason for this to return the string instead of the length of the string?. Doing so means you can simplify some of this logic.

app/questionnaire/questionnaire_schema.py
rmccar: It was setting the longest string value but I changed it to return it instead. Renamed

app/questionnaire/questionnaire_schema.py
rmccar: Its because Im also using it then to set the value on line 147

app/questionnaire/questionnaire_schema.py
```diff
@@ -99,11 +103,64 @@ def __init__(
...
+                            self._min_and_max_map[value["identifier"]][min_max]
+                        )
+                    if longest_string == 0:
+                        longest_string = int(
+                            self._min_and_max_map[value["identifier"]][min_max]
+                        )
+
+        if longest_string:
+            self._min_and_max_map[answer_id][min_max] = longest_string  # type: ignore
+        else:
```
katie-gardner, line 149: could simplify to `self._min_and_max_map[answer_id][min_max] = longest_string or len(default_min_max)`

app/questionnaire/questionnaire_schema.py
```diff
@@ -99,11 +103,64 @@ def __init__(
...
         self._populate_calculated_summary_section_dependencies()
+        self._populate_min_max_for_numeric_answers()
 
     @cached_property
     def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
         return ImmutableDict(self._answer_dependencies_map)
 
+    @cached_property
+    def min_and_max_map(self) -> ImmutableDict[str, dict[str, int]]:
+        return ImmutableDict(self._min_and_max_map)
```
katie-gardner, line 114: This makes the top level immutable, but as it is a dictionary of dictionaries the values could still be modified, you'd need to use the `make_immutable` helper for this to be properly protected

app/questionnaire/questionnaire_schema.py
```diff
@@ -99,11 +103,64 @@ def __init__(
...
+
+    def _create_min_max_map(
+        self,
+        min_max: Any,
```
katie-gardner, line 118: `Literal["minimum", "maximum"]` instead of Any?

app/questionnaire/questionnaire_schema.py
```diff
@@ -99,11 +103,64 @@ def __init__(
...
+        answer_id: str,
+        answers: list,
```
katie-gardner, line 120: think this could be more specific with `list[dict]`

katie-gardner (changes requested):
Looks to all be working as expected! Changed an answer to a few levels of min/max reference and that worked too - just some comments with a few places I think the code can be simplified

app/questionnaire/questionnaire_schema.py
```diff
@@ -99,11 +103,64 @@ def __init__(
...
+        longest_string: int = 0
+        for answer in answers:
+            value = answer.get(min_max, {}).get("value")
+
+            if isinstance(value, int):
+                if longest_string and len(str(value)) > longest_string:
+                    longest_string = len(str(value))
+                if not longest_string:
+                    longest_string = len(str(value))
+
```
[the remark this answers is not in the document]
rmccar: longest_string can be because 0 is falsey

app/questionnaire/questionnaire_schema.py
```diff
@@ -99,11 +103,64 @@ def __init__(
...
+                        )
+
+        if longest_string:
+            self._min_and_max_map[answer_id][min_max] = longest_string  # type: ignore
+        else:
+            self._min_and_max_map[answer_id][min_max] = len(default_min_max)  # type: ignore
+
+    def _populate_min_max_for_numeric_answers(self) -> None:
+        for answer_id, answers in self._answers_by_id.items():
+            if (answer_type := answers[0]["type"]) and answer_type in [
```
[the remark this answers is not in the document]
rmccar: Is there an advantage to changing this to a set?

app/questionnaire/questionnaire_schema.py
```diff
@@ -99,11 +103,64 @@ def __init__(
...
+                        longest_string = int(
+                            self._min_and_max_map[value["identifier"]][min_max]
+                        )
+                    if longest_string == 0:
+                        longest_string = int(
+                            self._min_and_max_map[value["identifier"]][min_max]
+                        )
+
+        if longest_string:
+            self._min_and_max_map[answer_id][min_max] = longest_string  # type: ignore
```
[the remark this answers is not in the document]
rmccar: Removed lines anyway because of the below comment

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
katie-gardner: There is! 
It uses hash lookups instead of iterating over the list so offers slightly better performance 
e.g. the set one is about 3x faster
```
import timeit
print(timeit.timeit("9 in [1, 2, 3, 4, 5]"))
print(timeit.timeit("9 in {1, 2, 3, 4, 5}"))
```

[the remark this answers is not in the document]
rmccar: Cool 👍

app/questionnaire/questionnaire_schema.py
katie-gardner: Nice - with the `make_immutable` helper you can now remove the `ImmutableDict` call inside, and just pass it `self._min_and_max_map`

app/jinja_filters.py
katie-gardner: +1

katie-gardner (approved):
My comments addressed and agree with some of the remaining ones but otherwise all good from me 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -116,6 +121,47 @@ def placeholder_transform_section_dependencies_by_block(
...
+        return make_immutable(self._min_and_max_map)
+
+    def _create_min_max_map(
+        self,
+        min_max: Literal["minimum", "maximum"],
+        answer_id: str,
+        answers: list[ImmutableDict],
+        default_min_max: str,
+    ) -> None:
+        longest_value_length: int = 0
```
katie-gardner, line 135: non blocker, happened to just notice, you don't need the int type hint here as mypy will infer it

app/jinja_filters.py
```diff
@@ -253,17 +253,27 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
     return {"should_wrap_with_fieldset": should_wrap_with_fieldset}
 
 
+def get_min_max_values(
```
MebinAbraham, line 256: This should be renamed to like `get_mix_max_value_width`

MebinAbraham, line 256: A doc string explaining why we are doing this would help.

app/jinja_filters.py
MebinAbraham: Still outstanding.

app/questionnaire/questionnaire_schema.py
```diff
@@ -116,6 +121,47 @@ def placeholder_transform_section_dependencies_by_block(
...
+        for answer in answers:
+            value = answer.get(min_max, {}).get("value")
+
+            if isinstance(value, int):
+                longest_value_length = max(longest_value_length, len(str(value)))
+
+            elif isinstance(value, dict) and value:
```
MebinAbraham, line 142: Use a generic check like `Mapping`.

app/questionnaire/questionnaire_schema.py
```diff
@@ -116,6 +121,47 @@ def placeholder_transform_section_dependencies_by_block(
...
+        min_max: Literal["minimum", "maximum"],
+        answer_id: str,
+        answers: list[ImmutableDict],
+        default_min_max: str,
+    ) -> None:
+        longest_value_length: int = 0
+        for answer in answers:
+            value = answer.get(min_max, {}).get("value")
+
+            if isinstance(value, int):
```
MebinAbraham, line 139: Could this be floats?

app/questionnaire/questionnaire_schema.py
```diff
@@ -116,6 +121,47 @@ def placeholder_transform_section_dependencies_by_block(
...
+                longest_value_length = max(longest_value_length, len(str(value)))
+
+            elif isinstance(value, dict) and value:
+                if value.get("source") == "answers":
+                    longest_value_length = max(
+                        longest_value_length,
+                        int(self._min_and_max_map[value["identifier"]][min_max]),
```
MebinAbraham, line 146: Why is the `int` call needed?. If it is for type hinting sakes, it should be removed.

app/questionnaire/questionnaire_schema.py
```diff
@@ -116,6 +121,49 @@ def placeholder_transform_section_dependencies_by_block(
...
+        min_max: Literal["minimum", "maximum"],
+        answer_id: str,
+        answers: Iterable[ImmutableDict],
+        default_min_max: str,
+    ) -> None:
+        longest_value_length = 0
+        for answer in answers:
+            value = answer.get(min_max, {}).get("value")
+
+            if isinstance(value, float):
```
MebinAbraham, line 139: My comment wasn't that this is a float, it was that could it also be a float. Looking at validator, it is int or string (string is invalid), I would change to support int and float, as we can make validator allow float, which is edge case but still valid.

Right now setting to a string breaks validator, which is a gaps in validator that can be fixed separately.

app/questionnaire/questionnaire_schema.py
```diff
@@ -116,6 +121,49 @@ def placeholder_transform_section_dependencies_by_block(
...
+    def _populate_min_max_for_numeric_answers(self) -> None:
+        for answer_id, answers in self._answers_by_id.items():
+            if answers[0]["type"] in {
+                "Currency",
+                "Duration",
+                "Number",
+                "Percentage",
+                "Unit",
+            }:
+                self._create_min_max_map("minimum", answer_id, answers, len("0"))
```
MebinAbraham, line 162: Just pass `1` instead of calling `len` on `"0"`?

app/questionnaire/questionnaire_schema.py
```diff
@@ -116,6 +121,49 @@ def placeholder_transform_section_dependencies_by_block(
     def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
         return ImmutableDict(self._answer_dependencies_map)
 
+    @cached_property
+    def min_and_max_map(self) -> Any:
+        return make_immutable(self._min_and_max_map)
+
+    def _create_min_max_map(
```
MebinAbraham, line 128: It doesn't look like we are taking into account decimals.

If min was 0, but decimal was 2, right now it would return min length as 1 but it should be 3 as it would be `0.00`. Same applies to max.

app/questionnaire/questionnaire_schema.py
```diff
@@ -116,6 +121,55 @@ def placeholder_transform_section_dependencies_by_block(
...
+        longest_value_length = 0
+        for answer in answers:
+            value = answer.get(min_max, {}).get("value")
+
+            if isinstance(value, float | int):
+                value_length = len(str(value))
+                decimal_places = answer.get("decimal_places")
+
+                if decimal_places:
+                    value_length = value_length + decimal_places
```
MebinAbraham, line 144: Don't you need an extra one for the decimal point itself?

app/questionnaire/questionnaire_schema.py
```diff
@@ -116,6 +121,55 @@ def placeholder_transform_section_dependencies_by_block(
...
+
+    def _populate_min_max_for_numeric_answers(self) -> None:
+        for answer_id, answers in self._answers_by_id.items():
+            if answers[0]["type"] in {
+                "Currency",
+                "Duration",
+                "Number",
+                "Percentage",
+                "Unit",
+            }:
```
MebinAbraham, line 167: Minor, I think it is worth making this inline set a constant called "NUMERIC_ANSWER_TYPES".

tests/app/test_jinja_filters.py
```diff
@@ -258,10 +260,22 @@ def test_format_duration(duration, formatted_duration, app):
...
-    assert get_width_for_number(answer) == width
+def test_get_width_for_number(answer, width, app):
+    with app.app_context():
+        schema = QuestionnaireSchema({})
+        schema.min_and_max_map = {
+            "set-maximum": {"maximum": 5, "minimum": 0},
+            "set-minimum": {"maximum": 15, "minimum": 5},
+        }
+        g.schema = schema
+        assert get_width_for_number(answer) == width
```
MebinAbraham, line 278: This doesn't take into account chained values does i.e i.e 3 needs 2 and 2 needs 1?

app/jinja_filters.py
```diff
@@ -249,17 +249,34 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
...
+    min_max: Literal["minimum", "maximum"], answer: AnswerType, default_value: int
+) -> int:
+    """
+    This function gets the minimum and maximum value accepted for a question.
+    Which then allows us to use that value to set the width of the textbox to suit that min and max.
+    """
+
+    if (
+        answer.get(min_max, {})
+        and isinstance(answer[min_max]["value"], dict)
```
MebinAbraham, line 262: Minor I would make the check generic, i.e is instance Mapping

tests/app/test_jinja_filters.py
```diff
@@ -615,13 +676,14 @@ def test_get_formatted_address(address_fields, formatted_address):
     ],
 )
 def test_other_config_numeric_input_class(
-    answer_schema_number, max_value, expected_width
+    answer_schema_number, max_value, expected_width, app
 ):
-    if max_value:
-        answer_schema_number["maximum"] = {"value": max_value}
+    with app.app_context():
```
katie-gardner, line 681: does this need `app_context` ?

katie-gardner (commented):
Looks like all is working 👍 just few minor comments

app/questionnaire/questionnaire_schema.py
```diff
@@ -133,8 +133,9 @@ def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
         return ImmutableDict(self._answer_dependencies_map)
 
     @cached_property
-    def min_and_max_map(self) -> Any:
-        return make_immutable(self._min_and_max_map)
+    # Type ignore: safe to assume _min_and_max_map exists
```
katie-gardner, line 136: Type ignore is that we can assume the return type rather than existence, but otherwise looks good

katie-gardner (approved):

app/jinja_filters.py
```diff
@@ -251,7 +251,7 @@ def should_wrap_with_fieldset_processor() -> dict[str, Callable]:
 
 def get_min_max_value_width(
     min_max: Literal["minimum", "maximum"], answer: AnswerType, default_value: int
-) -> int:
+) -> dict[str, dict[str, int]]:
```
katie-gardner, line 254: this was correct before

rmccar: Yeah but because I removed the cast to int I had to change this because the return from this is calling that function

katie-gardner: Ah! The error is caused by something else
when you do 
`schema = g.get("schema")` mypy doesn't know the type of schema, so schema then fails to pick up the type of `min_and_max_map` which is why the mypy error is now 
`Returning Any from a function declared to return int`

The return type was correct as `int`
but to fix the error you just need to do
`schema: QuestionnaireSchema = g.get("schema")`

rmccar: Got it cheers

berroar (approved):

PR #1140, opened 2023-06-15
Update Prepop branch with the latest changes from main

kylelawsonAND:
Updating the Prepop feature branch with the latest changes from main. 
### How to review
Check all tests are passing and main changes are fully reflected.


app/questionnaire/questionnaire_schema.py
```diff
@@ -1271,6 +1271,10 @@ def is_summary_with_calculation(summary_type: str) -> bool:
     return summary_type in {"GrandCalculatedSummary", "CalculatedSummary"}
 
 
+def is_summary_with_calculation(summary_type: str) -> bool:
```
katie-gardner, line 1274: this got duplicated during merge

katie-gardner (changes requested):

app/questionnaire/questionnaire_schema.py
kylelawsonAND: Removed 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -1109,7 +1113,9 @@ def _get_section_and_block_ids_dependencies_for_progress_source_and_answer_ids_f
                     dependencies_ids_for_progress_value_source["blocks"][section_id] = {
                         identifier: OrderedSet()
```
katie-gardner, line 1114: I know this is a merge and this was already there, but I think this should be changed to 
```
dependencies_ids_for_progress_value_source["blocks"][section_id] = {
                        identifier: OrderedSet([current_section_id])
                    }

katie-gardner: it would fix the problem you're having with `.append` (needs to be `add`)

kylelawsonAND: looks like the statement is duplicated, one with passing in the current_section_id and one not, will remove the one without

kylelawsonAND: done 👍

PR #1145, opened 2023-06-20
Loading of supplementary data

katie-gardner:
This PR implements the creation of the new supplementary data store and the loading of sds data into that store whenever a new questionnaire session is launched (and any existing supplementary dataset id doesn't match)

It handles removing old supplementary data if the new data is different (or there is no new data) and creating or updating the list store for any supplementary data lists.

Tests are included to:

1. check that the sds api endpoint is called if (and only if) an `sds_dataset_id` is present in the launch claims metadata, and this id is different from any previous stored one.
2. check the serialisation/deserialisation of the supplementary data store and how data is retrieved
3. check that the questionnaire store appropriately updates the list store/answer store when supplementary data is added/updated/removed.

### Implementation decisions
The supplementary data, ideally will be fetched in `session.py` however this needs to happen between the store being created/fetched from storage and the metadata for it being overridden, which all happens in the authenticator. Since it needs the metadata from the login function, a context seemed like the way to go

### How to review
Read through the trello card to understand the requirements and check that the structure of the supplementary data store and how it will be interacted with is suitable 
Check that the test coverage is suitable for the changes

### How to test functionally
Following the guidance in run-mock-sds-endpoint.md start the mock sds api and pick the `test_supplementary_data` schema. 

1. Save the generated response id somewhere
4. Use the second dataset id from the read-me for the sds dataset id (that one has lists)
5. put a breakpoint on the final line of the login endpoint in `session.py`
6. On opening the survey and hitting that breakpoint, the newly created questionnaire store should have the correct supplementary data, and the list store should also contain the "products" list
7. Using another breakpoint on the first line of `_set_questionnaire_supplementary_data` if you relaunch the survey, with the same response id as before, but using the other supplementary dataset id
    * at first breakpoint the questionnaire store from before is loaded, so will have the old supplementary data, and a list store with "products"
    * on hitting the second breakpoint, the update will be complete, the list store will be empty, and the supplementary data will match that of `supplementary_data_no_repeat.json`
8. Use a break point on `return data` in `converter_v2.py` to check that the supplementary data and list mappings are included correctly.

### Checklist


katie-gardner:
[Loading of supplementary data (5 SP)](https://trello.com/c/Z9JGNTX2/5697-loading-of-supplementary-data-5-sp)

app/data_models/questionnaire_store.py
```diff
@@ -75,6 +141,7 @@ def serialize(self) -> str:
         data = {
             "METADATA": self._metadata,
             "ANSWERS": list(self.answer_store),
+            "SUPPLEMENTARY_DATA": self.supplementary_data_store.serialize(),
```
berroar, line 144: I notice now in the questionnaire store within the SUPPLEMENTARY_DATA field we have a supplementary_data key as per the image here - is this nesting necessary?
Screenshot 2023-06-22 at 14 26 34

berroar: Forgot the list_items below, maybe this key could be reamed to "data"?

schemas/test/en/test_supplementary_data.json
```diff
@@ -22,7 +22,8 @@
         },
         {
             "name": "sds_dataset_id",
-            "type": "string"
+            "type": "string",
+            "optional": true
```
berroar, line 26: Can't recall what we said about this - it was for the test case right? Can it now be reverted? I think if the sds_dataset_id is exists for a given form type it won't be optional

katie-gardner: Yeah I added this for coverage of the case where an sds_dataset_id is provided, the questionnaire store state is saved, and then its loaded and the new metadata doesn't contain an id, and so the old data needs deleting - if thats a case that will never happen I can revert this and slightly tweak the code not to worry about that scenario 👍 thoughts?

app/data_models/questionnaire_store.py
katie-gardner: good idea 👍  will go with data instead

katie-gardner: done

schemas/test/en/test_supplementary_data.json
berroar: Yeah I can understand checking for when the prepop data for a given id has been changed, but I'm not sure about the scenario above, as the form type would still require that sds_dataset_id, but maybe worth leaving this for now and double checking this with @MebinAbraham

kylelawsonAND (approved):
All working as expected, can see the data loaded into the new store and list store and can. see it being overwritten.
Impl looks grand - really nice work!

tests/app/data_model/test_questionnaire_store.py
```diff
@@ -112,3 +116,138 @@ def test_questionnaire_store_raises_when_writing_to_metadata(questionnaire_store
...
+        self.store = QuestionnaireStore(questionnaire_store.storage)
+        self.store.set_supplementary_data(supplementary_data)
+        assert "products" in self.store.supplementary_data_store.list_mappings
+        supplementary_list_item_ids = list(
+            self.store.supplementary_data_store.list_mappings["products"].values()
+        )
+        # check list mapping ids match list store ids
+        self.assert_list_store_data("products", supplementary_list_item_ids)
+
+    def test_updating_supplementary_data(
```
berroar, line 145: I can see there is a test for this but when I functionally test by launching a survey with our current repeating mock data -> make a copy of the response id -> update the mock data so that an additional list item is added -> relaunch the mock endpoint -> relaunch the survey with same response id -> I still only see two items in the /dump/debug window when I would expect to see three 😕

tests/app/submitter/test_convert_payload_0_0_3.py
```diff
@@ -1535,3 +1562,63 @@ def test_payload_dynamic_answers(version):
...
+
+    full_routing_path = [
+        RoutingPath(
+            ["dynamic-answer"],
+            section_id="section",
+        )
+    ]
+
+    supplementary_data = {
+        "survey_name": "supermarket test survey",
```
berroar, line 1578: Really minor but better to use one of the known supplementary data keys rather than survey_name?

tests/app/data_model/test_questionnaire_store.py
katie-gardner: Great spot - but this is actually what we want. If the sds dataset id has not changed, then it won't call the endpoint again for a data update. If you want to see 3 you'd need to adjust the mock data and endpoint to have a different sds dataset id

berroar: Ah i see I had `Any new list items in the reloaded data needs adding to the store.` in my head from the trello card but I see it is only when the id has changed 👍

berroar (approved):
Functionally tested, only the outstanding question but everything else looks good 🎉

mgdon650:
Looks good to me and works as should 👍

schemas/test/en/test_supplementary_data.json
MebinAbraham: A form type will either have it or not. There won't ever be a "maybe" case, so we don't need this?. If form type that had it, won't remove it midway. No reqs for that. Only for changing datasets.

app/data_models/answer_store.py
```diff
@@ -137,15 +137,17 @@ def remove_answer(
 
         return False
 
-    def remove_all_answers_for_list_item_id(self, list_item_id: str) -> None:
-        """Remove all answers associated with a particular list_item_id.
+    def remove_all_answers_for_list_item_ids(
+        self, list_item_ids: Iterable[str]
+    ) -> None:
```
MebinAbraham, line 142: I suggest instead of making this an iterable that is sometimes called with a single item of a list, use variable length args, i.e `*list_item_ids`? The is can be called with 1 or more list item ids.

app/data_models/supplementary_data_store.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Used to retrieve supplementary data in a similar style to AnswerStore
+        the identifier is the top level key for static data, and the name of the list for list items
+        selectors are used to reference nested data
+
+        For example if you wanted the identifier for the first item in "some_list"
+        it would be get_data(identifier="some_list", selectors=["identifier"], list_item_id=list_item_id-1)
+        """
+        if not (value := self._data_map.get((identifier, list_item_id))):
+            return None
```
MebinAbraham, line 94: I need to double-check this, but we might need to return the stored value instead of None, i.e. if it was an empty list, return that. For now, do that and we'll revert if not needed, depends on how we make use of it in placeholders. However, regardless of that, this can't check the value like this, as this would return None for values such as `0` or `False`.

app/data_models/supplementary_data_store.py
```diff
@@ -0,0 +1,121 @@
...
+        and list mappings of the form { list_name: { identifier : list_item_id }}
+        """
+        self._raw_data = supplementary_data or {}
+        self._list_mappings = list_mappings or {}
+        # use shallow copy of the data, as items will be popped off
+        self._data_map = self._build_map({**self._raw_data})
+
+    @cached_property
+    def raw_data(self) -> ImmutableDict:
+        return ImmutableDict(self._raw_data)
```
MebinAbraham, line 37: It would still be mutable right, as this wouldn't make the nested dicts an ImmutableDict?. We have a `make_immutable` helper.

app/routes/session.py
```diff
@@ -129,17 +134,41 @@ def login() -> Response:
...
+        questionnaire_store.metadata.survey_metadata["sds_dataset_id"]
+        if questionnaire_store.metadata and questionnaire_store.metadata.survey_metadata
+        else None
+    )
+
+    if not (new_sds_dataset_id := metadata["sds_dataset_id"]):
+        if existing_sds_dataset_id:
+            # remove what is already there
+            questionnaire_store.set_supplementary_data({})
+        return
```
MebinAbraham, line 157: As mentioned in the other comment, this could result in orphaned list items and answers.

app/routes/session.py
```diff
@@ -129,17 +134,41 @@ def login() -> Response:
...
+        # no need to fetch again
+        return
+
+    supplementary_data = get_supplementary_data(
+        # Type ignore: survey_id and either ru_ref or qid are required for schemas that use supplementary data
+        dataset_id=new_sds_dataset_id,
+        unit_id=metadata["ru_ref"] or metadata["qid"],  # type: ignore
+        survey_id=metadata["survey_id"],  # type: ignore
+    )
+    questionnaire_store.set_supplementary_data(supplementary_data["data"])
```
MebinAbraham, line 169: Can we add a log line to `get_supplementary_data` or after `set`, to say sd fetced. We can use this for monitoring to see how often we call the sds etc.

MebinAbraham (commented):
Few minor points, otherwise LGTM.

schemas/test/en/test_supplementary_data.json
katie-gardner: Removed the optional key and the code that was accounting for this and testing it 👍 
That makes sense

app/data_models/list_store.py
```diff
@@ -215,6 +215,10 @@ def add_list_item(self, list_name: str, primary_person: bool = False) -> str:
 
         return list_item_id
 
+    def delete_list(self, list_name: str) -> None:
+        """Deletes the entire list. Currently, this is only used when old supplementary data is being overwritten"""
```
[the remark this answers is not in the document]
katie-gardner: Done 👍

app/data_models/questionnaire_store.py
```diff
@@ -28,13 +29,12 @@ def __init__(
         self.version = version
         self._metadata: MutableMapping = {}
         # self.metadata is a read-only view over self._metadata
-        self.metadata: Optional[MetadataProxy] = (
-            MetadataProxy.from_dict(self._metadata) if self._metadata else None
-        )
+        self.metadata: MetadataProxy | None = None
```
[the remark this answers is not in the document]
katie-gardner: Yep, `self._metadata` was always none when that line of code is run so the `MetadataProxy.from_dict...` was redundant, all happens in `set_metadata` instead

app/data_models/questionnaire_store.py
```diff
@@ -63,10 +63,76 @@ def set_metadata(self, to_set: MutableMapping) -> QuestionnaireStore:
...
+        In the case that existing supplementary data is being replaced with new data: any list items in the old data
+        but not the new data are removed from the list store and related answers are deleted
+        :param new_data - the new supplementary data for comparison
+        """
+        deleted_list_item_ids: set[str] = set()
+        for list_name, mappings in self.supplementary_data_store.list_mappings.items():
+            if list_name not in new_data.get("items", {}):
+                self.list_store.delete_list(list_name)
+                deleted_list_item_ids.update(mappings.values())
+            else:
```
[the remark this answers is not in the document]
katie-gardner: Reversed - was just the order I naturally thought about the two cases but happy to swap round

app/data_models/questionnaire_store.py
```diff
@@ -63,10 +63,76 @@ def set_metadata(self, to_set: MutableMapping) -> QuestionnaireStore:
...
+                for identifier, list_item_id in mappings.items():
+                    if identifier not in [
+                        item["identifier"] for item in new_data["items"][list_name]
+                    ]:
```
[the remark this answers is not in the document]
katie-gardner: Good spot - fixed 👍

app/data_models/questionnaire_store.py
```diff
@@ -63,10 +63,76 @@ def set_metadata(self, to_set: MutableMapping) -> QuestionnaireStore:
...
+            list_name: self._create_supplementary_list(list_name, list_data)
+            for list_name, list_data in to_set.get("items", {}).items()
+        }
+
+        self.supplementary_data_store = SupplementaryDataStore(
+            supplementary_data=to_set, list_mappings=list_mappings
+        )
+
+    def _create_supplementary_list(
+        self, list_name: str, list_data: list[dict]
```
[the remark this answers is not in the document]
katie-gardner: Done

app/data_models/supplementary_data_store.py
katie-gardner: Ah 🤦‍♀️ complete oversight on the falsey values, thanks, fixed this 👍

app/data_models/supplementary_data_store.py
katie-gardner: Yep, good point, have switched to using `make_immutable` instead

app/data_models/supplementary_data_store.py
```diff
@@ -0,0 +1,121 @@
...
+
+        For example if you wanted the identifier for the first item in "some_list"
+        it would be get_data(identifier="some_list", selectors=["identifier"], list_item_id=list_item_id-1)
+        """
+        if not (value := self._data_map.get((identifier, list_item_id))):
+            return None
+
+        if selectors:
+            # for nested data, index with each selector
+            for selector in selectors:
```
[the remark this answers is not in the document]
katie-gardner: Nice 👍

app/routes/session.py
katie-gardner: Sure 👍

app/data_models/answer_store.py
katie-gardner: Yep - I like that, done 👍

app/routes/session.py
```diff
@@ -129,17 +134,39 @@ def login() -> Response:
...
+        # no need to fetch again
+        return
+
+    supplementary_data = get_supplementary_data(
+        # Type ignore: survey_id and either ru_ref or qid are required for schemas that use supplementary data
+        dataset_id=new_sds_dataset_id,
+        unit_id=metadata["ru_ref"] or metadata["qid"],  # type: ignore
+        survey_id=metadata["survey_id"],  # type: ignore
+    )
+    logger.info("supplementary data fetched")
```
MebinAbraham, line 166: Can we add the survey id and dataset id to the log line, ru_ref/qid should already be bound.

katie-gardner: Yep - they are already bound, added those two extras 👍

MebinAbraham (approved):
Nice work 👍

katie-gardner:
> Changes look good, but functionally is there a way to do this via vs code or is it easier to just use pycharm here?

You can use the dump/debug endpoint to check what happens if you keep the same response id but alter the sds_dataset_id 🙂

PR #1146, opened 2023-06-26
Fix `mass-metric-ton` units handling

petechd:
`mass-metric-ton` is no longer supported for `en_GB` and related locales, but still present in business schema and allowed in validator. Until removed from schema and validator definitions we want to substitute `mass-ton` for `mass-metric-ton` and vice versa, before and after unit formatting.

### How to review
New test cases in existing unit tests.

### Checklist

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -76,8 +76,8 @@ def format_unit(
     length: UnitLengthType = "short",
 ) -> str:
     # mass-metric-ton no longer supported for en_GB and related locales, but still present in business schema and allowed in validator,
-    # until removed from schema we substitute mass-ton for mass-metric-ton and vice versa, before/after unit formatting
-    measurement_unit = "mass-ton" if unit == "mass-metric-ton" else unit
+    # until removed from schema we substitute mass-tonne for mass-metric-ton before format unit
+    measurement_unit = "mass-tonne" if unit == "mass-metric-ton" else unit
 
     formatted_unit: str = units.format_unit(
         value=value,
@@ -85,8 +85,7 @@ def format_unit(
         length=length,
         locale=flask_babel.get_locale(),
     )
-    if unit == "mass-metric-ton":
-        return formatted_unit.replace("tn", "mt").replace("ton", "metric ton")
+
     return formatted_unit
 
 
@@ -100,8 +99,8 @@ def format_unit_input_label(unit: str, unit_length: UnitLengthType = "short") ->
     """
     unit_label: str
     # mass-metric-ton no longer supported for en_GB and related locales, but still present in business schema and allowed in validator,
-    # until removed from schema we substitute mass-ton for mass-metric-ton and vice versa, before/after unit formatting
-    measurement_unit = "mass-ton" if unit == "mass-metric-ton" else unit
+    # until removed from schema we substitute mass-tonne for mass-metric-ton before format unit
+    measurement_unit = "mass-tonne" if unit == "mass-metric-ton" else unit
 
     if unit_length == "long":
         unit_label = units.format_unit(
@@ -118,9 +117,6 @@ def format_unit_input_label(unit: str, unit_length: UnitLengthType = "short") ->
             length=unit_length,
             locale=flask_babel.get_locale(),
         ).strip()
-
-    if unit == "mass-metric-ton":
-        unit_label = unit_label.replace("tn", "mt").replace("ton", "metric ton")
 
     return unit_label
 
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -76,7 +76,7 @@ def format_unit(
     length: UnitLengthType = "short",
 ) -> str:
     # mass-metric-ton no longer supported for en_GB and related locales, but still present in business schema and allowed in validator,
-    # until removed from schema we substitute mass-ton for mass-metric-ton and vice versa before/after unit formatting
+    # until removed from schema we substitute mass-ton for mass-metric-ton and vice versa, before/after unit formatting
     measurement_unit = "mass-ton" if unit == "mass-metric-ton" else unit
 
     formatted_unit: str = units.format_unit(
@@ -100,7 +100,7 @@ def format_unit_input_label(unit: str, unit_length: UnitLengthType = "short") ->
     """
     unit_label: str
     # mass-metric-ton no longer supported for en_GB and related locales, but still present in business schema and allowed in validator,
-    # until removed from schema we substitute mass-ton for mass-metric-ton and vice versa before/after unit formatting
+    # until removed from schema we substitute mass-ton for mass-metric-ton and vice versa, before/after unit formatting
     measurement_unit = "mass-ton" if unit == "mass-metric-ton" else unit
 
     if unit_length == "long":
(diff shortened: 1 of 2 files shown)


petechd:
[Bug: Runner error handling `mass-metric-ton` units](https://trello.com/c/clHTvgL6/5760-bug-runner-error-handling-mass-metric-ton-units)

app/jinja_filters.py
```diff
@@ -93,22 +99,29 @@ def format_unit_input_label(unit: str, unit_length: UnitLengthType = "short") ->
     :param (str) unit_length length of unit text, can be one of short/long/narrow
     """
     unit_label: str
+    # mass-metric-ton no longer supported for en_GB and related locales, but still present in business schema and allowed in validator,
+    # until removed from schema we substitute mass-ton for mass-metric-ton and vice versa, before/after unit formatting
+    measurement_unit = "mass-ton" if unit == "mass-metric-ton" else unit
+
     if unit_length == "long":
         unit_label = units.format_unit(
```
katie-gardner, line 107: Could this be replaced with 
```
unit_label = format_unit(
  value=2,
  unit=unit,
  length=unit_length
).replace("2, ", "")
```
and likewise for the label below so that the hack only has to be in one place?

katie-gardner (changes requested):

app/jinja_filters.py
petechd: I am not sure, that "2" there is only to get some extra whitespace for long unit. How to make it only call format unit once?

katie-gardner: so I think you could do this
```
unit_label: str

    if unit_length == "long":
        unit_label = format_unit(value=2, unit=unit, length=unit_length).replace(
            "2 ", ""
        )
    else:
        # Type ignore: We pass an empty string  as the value so that we just return the unit label
        unit_label = format_unit(
            value="", unit=unit, length=unit_length  # type: ignore
        ).strip()

    return unit_label
```

katie-gardner: and then `format_unit` will have already made the substitution

app/jinja_filters.py
```diff
@@ -75,12 +75,18 @@ def format_unit(
...
+
     formatted_unit: str = units.format_unit(
         value=value,
-        measurement_unit=unit,
+        measurement_unit=measurement_unit,
         length=length,
         locale=flask_babel.get_locale(),
     )
+    if unit == "mass-metric-ton":
+        return formatted_unit.replace("tn", "mt").replace("ton", "metric ton")
```
katie-gardner, line 89: I installed babel 2.11.0 and ran 
```
print(units.format_unit(1, "mass-metric-ton", locale=Locale('en', 'GB')))
print(units.format_unit(1, "mass-metric-ton", length="short", locale=Locale('en', 'GB')))
```
which outputted 
`1 tonne`
`1 t`
these replacements aren't quite what it used to be

app/jinja_filters.py
```diff
@@ -75,12 +75,18 @@ def format_unit(
     value: int | float | Decimal,
     length: UnitLengthType = "short",
 ) -> str:
+    # mass-metric-ton no longer supported for en_GB and related locales, but still present in business schema and allowed in validator,
+    # until removed from schema we substitute mass-ton for mass-metric-ton and vice versa, before/after unit formatting
+    measurement_unit = "mass-ton" if unit == "mass-metric-ton" else unit
```
katie-gardner, line 80: I think instead we need to replace `mass-metric-ton` with `mass-tonne` and then the second replacement after calling `units.format_unit` won't be required

app/jinja_filters.py
katie-gardner: (see other comment) https://github.com/ONSdigital/eq-questionnaire-runner/pull/1146#discussion_r1242015494

app/jinja_filters.py
katie-gardner: both functions were doing `locale = flask_babel.get_locale()` so the refactor shouldn't affect that

app/jinja_filters.py
petechd: But only `mass-ton` is supported in validator atm, so no schema with `mass-tonne` is possible.

app/jinja_filters.py
petechd: But we use 2.12.1 already.

katie-gardner: Yep, so we've upgraded to 2.12.1 which has broken the schema, because we can't call format unit with `mass-metric-ton` anymore, so I installed the most recent version of babel that did still support this and checked what the output would be, which is `tonne` and `t`
So if it was changed to `metric-ton` now, this is different to what users would have seen when we were using a prior version of babel where this schema worked

petechd: Ok, I get it now, so it also adds to the confusion as in schema we are using metric tons but want tonne/t output in runner.

app/jinja_filters.py
petechd: Well, this will simplify it a lot in both.

app/jinja_filters.py
katie-gardner: yeah looking at `qfs_0002` it uses `tonnes` in the description but `mass-metric-ton` so that would mirror this too
```
"label": "Total quantity of heavy fuel oil purchased in tonnes",
"description": "Please provide your figures rounded to the nearest whole number. Enter &#39;0&#39;, rather than leave blank.",
"q_code": "122",
"unit": "mass-metric-ton",
"unit_length": "short",
"decimal_places": 0
```

app/jinja_filters.py
```diff
@@ -93,20 +98,18 @@ def format_unit_input_label(unit: str, unit_length: UnitLengthType = "short") ->
     :param (str) unit_length length of unit text, can be one of short/long/narrow
     """
     unit_label: str
+    # mass-metric-ton no longer supported for en_GB and related locales, but still present in business schema and allowed in validator,
+    # until removed from schema we substitute mass-tonne for mass-metric-ton before format unit
+    unit = "mass-tonne" if unit == "mass-metric-ton" else unit
```
katie-gardner, line 103: an advantage to this refactor is that you don't need this line anymore, because `format_unit` is already doing the replacement of `mass-metric-ton` with `mass-tonne`

app/jinja_filters.py
petechd: Ok, makes perfect sense, sorry, was hard to digest the first time.

app/jinja_filters.py
petechd: Oh, yeah, sorry distracted.

app/jinja_filters.py
petechd: I will double check what you did with 2.11 but these are terrific simplifications thanks tom your findings!

app/jinja_filters.py
katie-gardner: np! I realised after saying it looked almost the same so I should have added that extra info as a note 😅

berroar (approved):

Yuyuutsu (approved):
Functionally tested, and changes lgtm !

kylelawsonAND (approved):

PR #1148, opened 2023-06-27
Bug Fix: Decimals are only showing first 3 decimal places

mgdon650:

Numbers with more than 3 decimal places are being rounded to only show the first 3 places on summary pages and in placeholders.

Have changed formatted_number and format_unit to ensure that rounding does not take place and where necessary, all decimal places are displayed.

### How to test/review
Run ```make test-unit``` to ensure they pass. Test any schemas that accept decimals and ensure that there is no rounding taking place. You may need to edit "decimal_places:" within the schema files to allow for extra decimals to be entered e.g below:

Take any schema that shows answers at the end and with number/unit questions, such as test_unit_patterns.json, and edit one of the answer fields to have

```"decimal_places": 6```

enter a value for the field with 6 decimal places and continue to the survey summary.

It should show all 6 decimal places on the summary and anywhere else the answer is used if the max has been set at 6 (should always show the number of decimal places the user has entered).

### CODE REVIEW

File: app/jinja_filters.py [hunk 1]
```diff
@@ -78,13 +78,13 @@ def format_unit(
     # mass-metric-ton no longer supported for en_GB and related locales, but still present in business schema and allowed in validator,
     # until removed from schema we substitute mass-tonne for mass-metric-ton before format unit
     measurement_unit = "mass-tonne" if unit == "mass-metric-ton" else unit
-
+    formatted_value = value if value in [1, 2, ""] else 0
```
katie-gardner, line 81: instead of using `format_unit` like this to get the unit name with a dummy value and then replace the value, I think it would be better to use the babel function that `format_unit` calls to get the unit name

Then you could just do
```
unit_name = units.get_unit_name(
    measurement_unit=measurement_unit,
    length=length,
)
return f"{value}{unit_name}"
```

katie-gardner (changes requested):

app/jinja_filters.py
katie-gardner: In validator we verify that the maximum number of decimal places allowed is 6, so instead of the above, I think this could be another valid way to achieve the same thing
```
formatted_unit = units.format_unit(
   value=value,
   measurement_unit=measurement_unit,
   length=length,
   format="#.######"
)
```
possibly cleaner but I guess its still just another workaround for them not offering the quantization option that the other format methods do

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -58,16 +59,23 @@ def test_format_percentage(value, expected, transformer):
 @pytest.mark.parametrize(
     "unit, value, unit_length, expected",
     (
-        ("millimeter", Decimal(0.123), "short", "0.123 mm"),
+        (
+            "millimeter",
+            Decimal(0.123),
+            "short",
+            "0.1229999999999999982236431606 mm",
```
katie-gardner, line 66: why is this the expected result?

app/jinja_filters.py
MebinAbraham: When you provide a format, is locale not ignored?

MebinAbraham: Also, why do we do `formatted_value = value if value in [1, 2, ""] else 0`, that is confusing.

app/jinja_filters.py
```diff
@@ -35,7 +35,7 @@ def format_number(value: Union[int, Decimal, float]) -> str:
     formatted_number: str
     if value or value == 0:
         formatted_number = numbers.format_decimal(
-            value, locale=flask_babel.get_locale()
+            value, locale=flask_babel.get_locale(), decimal_quantization=False
```
MebinAbraham, line 38: `decimal_quantization` doesn't fix it for leading zeros. If a user enters 1.15000, it will render 1.15 even though we store 1.15000 and send downstream 1.15000.

petechd:
[Bug: Decimals are only showing first 3 decimal places](https://trello.com/c/1qgmlDAK/5689-bug-decimals-are-only-showing-first-3-decimal-places)

app/jinja_filters.py
petechd: So something similar to this would be valid?:
```python
    decimal_part = None

    if value:
        split_value = str(value).split(".")
        value = int(split_value[0])
        decimal_part = split_value[1] if len(split_value) > 1 else None

        value=value,
        measurement_unit=measurement_unit,
        length=length,
    )
    if decimal_part:
        formatted_unit_list = formatted_unit.split(" ")
        formatted_unit_list[0] = f"{formatted_unit_list[0]}.{decimal_part}"
        return " ".join(formatted_unit_list)
    return formatted_unit
```
but then, as Mick mentioned, if it was 1001.1111, it would become `1,001.1111 kilometre` not `1,001.1111 kilometres`..

katie-gardner: This might fail with a length of `narrow` where the format would be `1,001km`

There might be a way to do this with without the split, looking at the library, if you used one of the babel functions to get the precision of the number, you could split the pattern on "." and add that number of 0s to it

e.g. 
`numbers.format_decimal("1230.2400", locale="en", format="#,###.0000")` preserves the trailing zeros

tests/app/questionnaire/test_placeholder_transforms.py
mgdon650: its to do with Decimal(). i ran a test on it separately and that is the number it gives

MebinAbraham: It needs to works with decimals as that's what we store for answers not floats. It would work if Decimal quantisation is True. If we provide our own format we don't need to set it to False?

app/jinja_filters.py
petechd: Hacky, but former gurus of this time already introduced similar hacky code before:
```python
    if unit_length == "long":
        unit_label = format_unit(
            value=2, unit=unit, length=unit_length, locale=locale
        ).replace("2 ", "")
```

MebinAbraham: That is okay, and not brittle, when we just mutate the string, like we are doing, it becomes brittle and prone to error, I think we need to try do it via providing our own format but only modifying the decimal bit. Don't think there is a perfect solution, we just need to go with least prone to error.

MebinAbraham: > Ah true, a dynamic replacement would be better, the workaround here might be viable to some extent [python-babel/babel#90](https://github.com/python-babel/babel/issues/90) with something like `pattern = pattern.replace(".#", ".######")` but this isn't going to fix trailing zeroes, I can't immediately see a way to do that with the babel library other than formatting the integer part and manually adding the decimal part

This is our best bet I think. We just need to them adjust to fix trailing zeros.

tests/app/questionnaire/test_placeholder_transforms.py
MebinAbraham: FYI, the precision issue is because Decimal is being initialised with a float, it should be `Decimal("0.123").

app/utilities/decimal_places.py
```diff
@@ -1,11 +1,11 @@
 from decimal import Decimal
 
 import flask_babel
-from babel import Locale, numbers, units
+from babel import Locale, numbers
 
 
 def unit_dec(value: int | Decimal | float):
-    decimal_places = len(str(value).split(".")[-1])
+    decimal_places = len(str(value).split(".")[1])
```
katie-gardner, line 8: if value is an `int` this is going to break - you'll need to check it has decimal places first
an `isinstance` check won't be enough, as you could have something like `Decimal("222")`

There are functions for decimals that can get the number of places regardless of if its whole or not like `abs(Decimal("2.00").as_tuple().exponent)` but I quite like the simplicity of len, so probably just need to check the string has a `.`

MebinAbraham: We could use the `get_decimal_precision` method to do this. Happy with either.

katie-gardner: `get_decimal_precision` won't account for trailing zeroes, otherwise I'd agree

MebinAbraham: Ah fair point.

MebinAbraham (commented):
We need to also handle currency which has slightly different behaviour. Currency defaults to 2 decimal. But we can't set currency decimal to 1 if the user entered it to 1 decimal. For that case, we need to check the schema defined decimal and only use less than 2 if schema says so.

app/jinja_filters.py
```diff
@@ -32,13 +38,11 @@ def strip_tags(value: str) -> Markup:
 
 @blueprint.app_template_filter()
 def format_number(value: Union[int, Decimal, float]) -> str:
-    formatted_number: str
+    locale = flask_babel.get_locale()
+
     if value or value == 0:
```
katie-gardner, line 43: if value is `int | Decimal | float`.... then this check should be redundant, I suspect it can actually be none and this ought to be `if value is not None`
(and types fixed, or if types are correct, delete this)

app/jinja_filters.py
```diff
@@ -75,11 +86,12 @@ def format_unit(
...
     length: UnitLengthType = "short",
 ) -> str:
-    formatted_unit: str = units.format_unit(
+    locale = flask_babel.get_locale()
+    formatted_unit: str = custom_format_unit(
         value=value,
         measurement_unit=unit,
         length=length,
-        locale=flask_babel.get_locale(),
+        locale=locale,
```
katie-gardner, line 94: I thought it was good as in the currency one with this just passed in instead of making a variable

app/jinja_filters.py
```diff
@@ -494,8 +506,11 @@ def __init__(  # noqa: C901, R0912 pylint: disable=too-complex, too-many-branche
                 for option in value
             ]
         elif answer_type == "currency":
+            decimal_places = answer.get("decimal_places", None)
```
katie-gardner, line 509: Don't need the None default, `.get` already does that implicitly

app/questionnaire/placeholder_transforms.py
```diff
@@ -41,9 +44,21 @@ def __init__(
 
     input_date_format = "%Y-%m-%d"
 
-    def format_currency(self, number: Decimal | float, currency: str = "GBP") -> str:
-        formatted_currency: str = format_currency(number, currency, locale=self.locale)
-        return formatted_currency
+    def format_currency(
+        self,
+        number: Decimal | float,
```
katie-gardner, line 49: the tests for this use ints as a number, should be in the type hint too?

app/questionnaire/placeholder_transforms.py
```diff
@@ -41,9 +44,21 @@ def __init__(
...
+        currency: str = "GBP",
+        decimal_limit: int | None = None,
+    ) -> str:
+        if number or number == 0:
```
katie-gardner, line 53: again, I think either the typing should be fixed and this should be `if number is not None` or the check deleted

app/questionnaire/placeholder_transforms.py
```diff
@@ -124,11 +139,9 @@ def format_possessive(self, string_to_format: str) -> str:
 
         return string_to_format
 
-    def format_number(self, number: int | Decimal | str) -> str:
+    def format_number(self, number: int | Decimal | float) -> str:
         if number or number == 0:
```
katie-gardner, line 143: Appreciate these were already there, but same again - `if number is not None`

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,94 @@
...
+    decimal_limit: int | None = None,
+) -> str:
+    decimal_separator = numbers.get_decimal_symbol(locale)
+    decimal_places = (
+        len(str(value).split(f"{decimal_separator}")[1])
+        if decimal_separator in str(value)
+        else 0
+    )
+
+    # get local pattern
```
katie-gardner, line 36: comment typo `locale`

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,94 @@
...
+    parsed_locale = Locale.parse(locale)
+    number_format = parsed_locale.currency_formats["standard"]
+    number_format.frac_prec = (0, decimal_places)
+
+    # if decimal_limit is less than 2 then return the value the user entered
+    if decimal_limit and decimal_limit < 2:
```
katie-gardner, line 42: this might need to be `if decimal_limit is not None` will 0 work?

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,94 @@
+from decimal import Decimal
+from typing import Literal, TypeAlias
+
+from babel import Locale, numbers, units
+
+UnitLengthType: TypeAlias = Literal["short", "long", "narrow"]
+
+
+def custom_format_decimal(value: int | Decimal | float, locale: Locale | str) -> str:
```
katie-gardner, line 9: Need unit tests for all the public methods in this file

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,94 @@
...
+        return numbers.format_currency(
+            number=value,
+            currency=currency,
+            format=number_format,
+            locale=locale,
+            currency_digits=False,
+        )
+
+    currency_digits = decimal_places < 2
+    return numbers.format_currency(
```
katie-gardner, line 52: I tried 
`print(custom_format_currency(Decimal("123.45", "GBP", "en_GB", 1))` and got back `£123.45`
Im not sure if this is doing the right thing?

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,94 @@
...
+        currency_digits=currency_digits,
+    )
+
+
+def custom_format_unit(
+    value: int | float | Decimal,
+    measurement_unit: str,
+    locale: Locale | str,
+    length: UnitLengthType = "short",
+):
```
katie-gardner, line 66: My understanding is that this adds on the unit but keeps decimals the same, docstring would be helpful

katie-gardner (changes requested):
I think there are a few errors with the custom functions that will be spotted once tests have been added for them, so will resume review after that

app/questionnaire/placeholder_transforms.py
MebinAbraham: I can't see a scenario where it would be valid to pass None. The schema should only call format if it knows it has something to format. it is Author's responsibility to ensure that is handled via schema logic. If it got None and we rendered an empty string, it wouldn't throw an error in prod but is ultimately hiding a flawed schema. So if when removing this, we see errors in prod, it would mean schemas aren't guarding this scenario and was incorrectly passing.

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -7,36 +7,57 @@
...
-        ("11.99", "GBP", "£11.99"),
-        ("11000", "USD", "US$11,000.00"),
-        (0, None, "£0.00"),
-        (0.00, None, "£0.00"),
+        (2, "GBP", 6, "£2.00"),
+        ("11", "GBP", 2, "£11.00"),
+        ("11.99", "GBP", 2, "£11.99"),
+        ("11000", "USD", 2, "US$11,000.00"),
+        (0, None, 2, "£0.00"),
+        (0.00, None, 2, "£0.00"),
```
MebinAbraham, line 17: This is the only test with trailing zero but default decimal for GBP would be 2 hence this doesn't prove trailing zero work. You should have one for `Decimal("1.00000"), GBP, 6, "£1.00000")`

MebinAbraham, line 17: Why are these set to None?. We are defaulting to GBP in the test body but why?.

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -7,36 +7,57 @@
...
+        (11000, "USD", 2, "US$11,000.00"),
+        (Decimal("123.45"), "GBP", 1, "£123.4"),
+        (Decimal("2.14564"), "GBP", 6, "£2.14564"),
+        (Decimal("2.1"), "GBP", 1, "£2.1"),
+        (Decimal("2.1"), "GBP", None, "£2.10"),
+        (Decimal("3000"), "JPY", 0, "JP¥3,000"),
+        (Decimal("1.1"), "GBP", 6, "£1.10"),
+        (Decimal("3000.445"), "GBP", 6, "£3,000.445"),
+        (Decimal("3000.44545"), "GBP", None, "£3,000.44545"),
+        ("", "", "", ""),
```
MebinAbraham, line 27: Why do we have this last one?

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -7,36 +7,57 @@
...
     ),
 )
-def test_format_currency(number, currency, expected, transformer):
-    assert transformer().format_currency(number, currency or "GBP") == expected
+def test_format_currency(
```
MebinAbraham, line 30: Grouping the tests by decimal length and currency would help.

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -7,36 +7,57 @@
...
+        ("123.40", "123.40"),
+        (123434.7678, "123,434.7678"),
+        ("123.45678", "123.45678"),
+        ("2344.6533", "2,344.6533"),
         ("1000", "1,000"),
         ("10000", "10,000"),
         ("100000000", "100,000,000"),
         (0, "0"),
-        (0.00, "0"),
+        (Decimal("0.00"), "0.00"),
```
MebinAbraham, line 53: We should have another with trailing decimal zeros > 3 since that is what the bug was for.

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -58,16 +79,26 @@ def test_format_percentage(value, expected, transformer):
...
+        ),
+        ("centimeter", 123, "short", "123 cm"),
+        ("kilometer", 123, "long", "123 kilometres"),
+        ("kilometer", 3, "long", "3 kilometres"),
+        ("kilometer", 1.2, "long", "1.2 kilometres"),
+        ("kilometer", 1.2345, "long", "1.2345 kilometres"),
         ("mile", "123", "short", "123 mi"),
         ("mile", "123", "narrow", "123mi"),
         ("mile", "123", None, "123 mi"),
+        ("mile", "123.456789", None, "123.456789 mi"),
```
MebinAbraham, line 96: Same as above, demo trailing decimal zeros.

tests/app/test_jinja_filters.py
```diff
@@ -57,25 +58,55 @@ def test_get_formatted_currency_with_no_value():
...
-        (0.00, "0"),
+        (Decimal("0.00"), "0.00"),
         ("", ""),
         (None, ""),
         (Undefined(), ""),
     ),
 )
-def test_format_number(number, formatted_number):
-    assert format_number(number) == formatted_number
+def test_format_number(number, formatted_number, app):
```
MebinAbraham, line 107: Test jinja filters and placeholder transforms have the same code, we should share the `parametrize` argument between the two instead of duplicating it, as it insures they both test the same scenarios.

tests/functional/spec/features/units.spec.js
```diff
@@ -54,4 +54,23 @@ describe("Units", () => {
...
+    await $(SetVolumeUnitsBlockPage.litres()).setValue(1.123);
+    await $(SetVolumeUnitsBlockPage.hectolitres()).setValue(1.1234);
+    await $(SetVolumeUnitsBlockPage.megalitres()).setValue(1.12345);
+    await $(SetVolumeUnitsBlockPage.submit()).click();
+    await $(SetWeightUnitsBlockPage.submit()).click();
+    await expect(await $(SubmitPage.cubicCentimetres()).getText()).to.equal("1.1 cm³");
+    await expect(await $(SubmitPage.cubicMetres()).getText()).to.equal("1.12 m³");
+    await expect(await $(SubmitPage.litres()).getText()).to.equal("1.123 l");
+    await expect(await $(SubmitPage.hectolitres()).getText()).to.equal("1.1234 hl");
+    await expect(await $(SubmitPage.megalitres()).getText()).to.equal("1.12345 Ml");
```
MebinAbraham, line 74: I suggest updating one of the number to have > 3 trailing decimal zeros.

tests/functional/spec/numbers.spec.js
```diff
@@ -94,5 +94,9 @@ describe("Number validation", () => {
 
       await expect(await browser.getUrl()).to.contain(SubmitPage.pageName);
     });
+
+    it("When a number with more than 3 decimal places has been entered, Then it should be displayed correctly on the summary", async () => {
+      await expect(await $(SubmitPage.testDecimal()).getText()).to.equal("£11.12345");
```
MebinAbraham, line 99: I suggest updating one of the number to have > 3 trailing decimal zeros.

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,97 @@
...
+
+from babel import Locale, numbers, units
+
+UnitLengthType: TypeAlias = Literal["short", "long", "narrow"]
+
+
+def custom_format_decimal(value: int | Decimal | float, locale: Locale | str) -> str:
+    """
+    Function is used to format decimal numbers avoiding the decimal_quantization=True/False constraints and preserving the number input by the user.
+    e.g 123.430 stays as 123.430 (if schema allows).
```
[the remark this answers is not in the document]
berroar: Updated this 👍

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,97 @@
...
+    number_format = get_number_format(value, locale)
+
+    return numbers.format_decimal(
+        value,
+        format=number_format,
+        locale=locale,
+    )
+
+
+def custom_format_currency(
```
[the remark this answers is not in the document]
berroar: Added doc strings 👍

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,97 @@
...
+        measurement_unit=measurement_unit,
+        length=length,
+        format=number_format,
+        locale=locale,
+    )
+
+    return formatted_unit
+
+
+def get_number_format(value: int | float | Decimal, locale: Locale | str) -> str:
```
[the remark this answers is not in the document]
berroar: Added doc strings 👍

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -7,36 +7,57 @@
...
     (
-        ("11", "GBP", "£11.00"),
-        ("11.99", "GBP", "£11.99"),
-        ("11000", "USD", "US$11,000.00"),
-        (0, None, "£0.00"),
-        (0.00, None, "£0.00"),
+        (2, "GBP", 6, "£2.00"),
+        ("11", "GBP", 2, "£11.00"),
+        ("11.99", "GBP", 2, "£11.99"),
+        ("11000", "USD", 2, "US$11,000.00"),
```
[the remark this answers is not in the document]
berroar: Updated these 👍

tests/app/questionnaire/test_placeholder_transforms.py
berroar: I have removed these as not sure either why GBP was being defaulted in the test (have also removed that) 👍

tests/app/questionnaire/test_placeholder_transforms.py
berroar: Removed 👍

app/utilities/decimal_places.py
berroar: Added tests for these methods now 👍

app/utilities/decimal_places.py
berroar: Should be fixed now 👍

app/utilities/decimal_places.py
berroar: Added docstrings to all the methods in this file 👍

app/jinja_filters.py
berroar: Removed the check now based on Mebin's comment below

app/questionnaire/placeholder_transforms.py
berroar: Updated now assuming that None's are not valid 👍

tests/app/questionnaire/test_placeholder_transforms.py
```diff
@@ -7,36 +7,57 @@
...
+        (0, None, 2, "£0.00"),
+        (0.00, None, 2, "£0.00"),
+        (11000, "USD", 2, "US$11,000.00"),
+        (Decimal("123.45"), "GBP", 1, "£123.4"),
+        (Decimal("2.14564"), "GBP", 6, "£2.14564"),
+        (Decimal("2.1"), "GBP", 1, "£2.1"),
+        (Decimal("2.1"), "GBP", None, "£2.10"),
+        (Decimal("3000"), "JPY", 0, "JP¥3,000"),
```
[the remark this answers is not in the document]
berroar: Added 👍

tests/app/questionnaire/test_placeholder_transforms.py
berroar: Added 👍

tests/app/questionnaire/test_placeholder_transforms.py
berroar: Added 👍

tests/app/questionnaire/test_placeholder_transforms.py
berroar: Added 👍

tests/app/test_jinja_filters.py
berroar: Sharing the parametrize now 👍

tests/functional/spec/features/units.spec.js
berroar: Updated 👍

tests/functional/spec/numbers.spec.js
berroar: Updated 👍

tests/app/questionnaire/test_placeholder_transforms.py
berroar: Grouping now 👍

app/questionnaire/placeholder_transforms.py
MebinAbraham: Worth checking if there are any places that call this, where answer could genuinely be None (shouldn't be any if schema logic is right?)

berroar: I've scanned through the business schemas checking the `format_currency` and `format_number` methods (don't think any use `format_unit` at the moment) and it seems all the conditional `format_currency` transforms are using mandatory questions and the ones that aren't have default `0` values so think we are ok there.

For `format_number` I found instances in the `1_0112` and `1_0215` Retail Sales schemas where there is a conditional question asking for the number of full-time employees , which is then being used in a transform. This does now fall over when not answered given the above changes. All other instances i found seemed to be mandatory, only found these instances but possible I may have missed some.

tests/app/test_jinja_filters.py
```diff
@@ -3,7 +3,7 @@
 
 import pytest
 import simplejson as json
-from jinja2 import Undefined
+from _decimal import Decimal
```
katie-gardner, line 6: from `decimal` instead of `_decimal` ?

tests/app/views/contexts/test_section_summary_context.py
```diff
@@ -368,6 +368,7 @@ def test_context_for_section_summary_with_list_summary_and_first_variant(
                                                             "unit": None,
                                                             "unit_length": None,
                                                             "value": 123,
+                                                            "decimal_places": None,
```
katie-gardner, line 371: I find it odd that `"decimal_places": None` has been added everywhere, can it be taken out now that things have been fixed? or was this needed?

katie-gardner (changes requested):
Functionally looks like its almost all working now 👍 just wondering if theres still a chance to slightly simplify the currency variant further

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,113 @@
...
+    decimal_places = len(str(value).split(".")[1]) if "." in str(value) else 0
+
+    # get locale pattern
+    parsed_locale = Locale.parse(locale)
+    number_format = parsed_locale.currency_formats["standard"]
+
+    # If set, the number of decimals displayed is limited based on the value of the `decimal_limit` parameter.
+    # If the number of decimal places entered by the user is less than the `decimal_limit` then we should display the
+    # number of decimals as entered by the user. Otherwise, we should display the number of decimals as entered by the user.
+    if decimal_limit:
```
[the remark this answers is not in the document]
katie-gardner: It gets confusing because no decimal limit defaults to whatever the user enters, but I'm wondering if thats what we should be doing - if there is no decimal limit, should currency not default to 2?

app/jinja_filters.py
```diff
@@ -31,30 +37,30 @@ def strip_tags(value: str) -> Markup:
...
-def get_formatted_currency(value: Union[float, Decimal], currency: str = "GBP") -> str:
-    if value or value == 0:
-        formatted_currency: str = numbers.format_currency(
-            number=value, currency=currency, locale=flask_babel.get_locale()
-        )
-        return formatted_currency
-
-    return ""
+def get_formatted_currency(
+    value: float | Decimal,
```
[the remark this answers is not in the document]
berroar: Updated 👍

app/utilities/decimal_places.py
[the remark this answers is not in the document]
berroar: Updated with the code from the above comment, but simplifying further with `currency_digits=decimal_limit is None,` doesn't seem to work.

tests/app/test_jinja_filters.py
berroar: Updated 👍

tests/app/views/contexts/test_section_summary_context.py
berroar: They seem to be still needed as they are part of the generated context

app/utilities/decimal_places.py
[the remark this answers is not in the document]
katie-gardner: Ahh... yeah if you have 100, and decimal places 6, it would not change it to £100.00 in that instance, so I see why the 2 was needed. Since every currency format uses 2 for frac_prec[1] this seems safe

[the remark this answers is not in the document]
MebinAbraham: Will discuss with Donna tomorrow however food for thought: Author by default sets decimals to 0, therefore, a normal currency field will not allow decimal and display will use no decimal places since schema limit is 0 by default. Therefore, we probably should treat not defined as 0 and only use the locale limit of 2 if the user entered 1 decimal place. So £1.1 would display as £1.10 unless schema limit is 1.

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,102 @@
...
+    decimal_limit: int | None = None,
+) -> str:
+    """
+    This function provides a wrapper for the numbers `format_currency` method, generating the
+    number format (including the desired number of decimals).
+
+    The number of decimals displayed is based on the value entered by the user, the decimal limit set in the schema
+    and the locale.
+    """
+    decimal_places = len(str(value).split(".")[1]) if "." in str(value) else 0
```
MebinAbraham, line 37: This would mean other decimal symbols like commas won't work. Can you not get the decimal symbol using the locale?

berroar: Yes it is possible to get the separator with the locale, was doing that previously, but finding it difficult to get this working as the format_decimal method only accepts decimal and float types, so not sure it's possible to pass types to it with other decimal symbols i.e `,` 😕. Would the value already be formatted by the time it gets to these methods?

MebinAbraham (commented):
The ``_format_total` method of calculated summary and grand calculated summary context is now aware of the decimal limit, so the total is always 2 decimal places. That needs passing in the decimal places limit so that it can behaviour consistent with the answer display.

Secondly, when getting the decimal places for the CS/GCS, you need to get the max decimal places among all answers since we don't enforce all answers that make up the total must have the same decimal limit. New tests will need to be added for this.

katie-gardner (approved):
Working as expected, with and without transforms, currency defaults to min 2dp when limit is above and shows more in higher cases, nicely done 👍 
just added a couple of super minor things, non blocking

app/questionnaire/questionnaire_schema.py
```diff
@@ -960,15 +960,30 @@ def get_answer_format_for_calculated_summary(
...
             "currency": first_answer.get("currency"),
+            "decimal_places": decimal_limit,
         }
 
+    def get_decimal_limit(self, answer_ids: list[str]) -> int:
+        decimal_limit = 0
+        for answer_id in answer_ids:
+            for answer in self.get_answers_by_answer_id(answer_id):
+                if (
+                    answer.get("decimal_places")
```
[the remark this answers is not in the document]

app/utilities/decimal_places.py
```diff
@@ -40,29 +40,18 @@ def custom_format_currency(
     parsed_locale = Locale.parse(locale)
```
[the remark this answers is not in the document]
berroar: This relates to the comment above: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1148#discussion_r1301523857. I was using the locale delimiter initially but it breaks because by the time we reach this method the delimiter will always have been formatted and will be `.`

app/views/contexts/calculated_summary_context.py
```diff
@@ -307,11 +316,15 @@ def _get_answer_format(self, groups: Iterable[Mapping]) -> Tuple[dict, list]:
     @staticmethod
     def _format_total(
         *,
-        answer_format: Mapping[str, Literal["short", "long", "narrow"]],
+        answer_format: Mapping[str, Literal["short", "long", "narrow"] | int],
```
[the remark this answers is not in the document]
berroar: Using generic now 👍

app/views/contexts/calculated_summary_context.py
```diff
@@ -298,7 +298,16 @@ def _get_answer_format(self, groups: Iterable[Mapping]) -> Tuple[dict, list]:
                             "unit": answer.get("unit"),
                             "unit_length": answer.get("unit_length"),
                             "currency": answer.get("currency"),
+                            "decimal_places": 0,
                         }
+
+                        if (
+                            answer.get("decimal_places")
```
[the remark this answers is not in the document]
katie-gardner: It could be done without extra nesting if you keep the ifs combined
```
if (decimal_places := answer.get("decimal_places")) and decimal_places > answer_format["decimal_places"]:
  answer_format["decimal_places"] = decimal_places
```
however, I think the function is clear too, resolving 👍

app/questionnaire/placeholder_transforms.py
katie-gardner: Just checking before resolving @berroar - was there a follow up for these two somewhere?

app/forms/validators.py
```diff
@@ -128,39 +128,64 @@ def __call__(
         field: Union[DecimalFieldWithSeparator, IntegerFieldWithSeparator],
     ) -> None:
         value: Union[int, Decimal] = field.data
+        decimal_places = 0
```
MebinAbraham, line 131: We should set this to None, as the `custom_format_currency` will set to 0.
How come it is not needed for `format_number`?.

app/forms/validators.py
```diff
@@ -128,39 +128,64 @@ def __call__(
...
+        if field.type == "DecimalFieldWithSeparator":
```
MebinAbraham, line 132: Check using `isinstance` instead of the literal string type.

app/forms/validators.py
```diff
@@ -128,39 +128,64 @@ def __call__(
...
-            error_message = self.validate_minimum(value) or self.validate_maximum(value)
+            error_message = self.validate_minimum(
+                value, decimal_places
+            ) or self.validate_maximum(value, decimal_places)
             if error_message:
                 raise validators.ValidationError(error_message)
 
-    def validate_minimum(self, value: NumType) -> Optional[str]:
+    def validate_minimum(
+        self, value: NumType, decimal_places: int | None
```
MebinAbraham, line 143: Minor, enforce keyword only args?. or at least for decimal places.

app/forms/validators.py
```diff
@@ -424,9 +449,20 @@ def __call__(
         is_valid, message = self._is_valid(condition, total, target_total)
 
         if not is_valid:
+            decimal_places = (
+                2
+                if isinstance(target_total, int)
```
MebinAbraham, line 454: Why do we default to 2 here?.

app/jinja_filters.py
```diff
@@ -34,30 +40,30 @@ def strip_tags(value: str) -> Markup:
...
-        formatted_currency: str = numbers.format_currency(
-            number=value, currency=currency, locale=flask_babel.get_locale()
-        )
-        return formatted_currency
-
-    return ""
+def get_formatted_currency(
+    value: int | float | Decimal,
+    currency: str = "GBP",
+    decimal_limit: int | None = None,
```
MebinAbraham, line 58: We should be consistent with the naming for clarity, some are called limit, some are called places.

app/views/contexts/calculated_summary_context.py
```diff
@@ -298,20 +298,34 @@ def _get_answer_format(self, groups: Iterable[Mapping]) -> Tuple[dict, list]:
...
     @staticmethod
     def _format_total(
         *,
-        answer_format: Mapping[str, Literal["short", "long", "narrow"]],
+        answer_format: Mapping,
         total: NumericType,
     ) -> str:
         if answer_format["type"] == "currency":
-            return get_formatted_currency(total, answer_format["currency"])
+            return get_formatted_currency(
```
MebinAbraham, line 324: `get_formatted_currency` is just a wrapper on `custom_format_currency` which doesn't add anything extra. Therefore, I suggest dropping this one and renaming `custom_format_currency` to `get_formatted_currency`.

app/views/contexts/calculated_summary_context.py
```diff
@@ -298,20 +298,34 @@ def _get_answer_format(self, groups: Iterable[Mapping]) -> Tuple[dict, list]:
...
+                value=total,
+                currency=answer_format["currency"],
+                decimal_limit=answer_format.get("decimal_places") or 0,
```
MebinAbraham, line 327: The `or 0` is redundant?. ``custom_format_currency` already handles the None.

tests/app/test_jinja_filters.py
```diff
@@ -28,6 +28,108 @@
...
+        (11000, "USD", 2, "US$11,000.00"),
+        (Decimal("11.99"), "GBP", 2, "£11.99"),
+        (2, "GBP", 6, "£2.00"),
+        (Decimal("1.1"), "GBP", 6, "£1.10"),
+        (Decimal("1.10"), "GBP", 6, "£1.10"),
+        (Decimal("1.100"), "GBP", 6, "£1.100"),
+        (Decimal("1.1000"), "GBP", 6, "£1.1000"),
+        (Decimal("2.14564"), "GBP", 6, "£2.14564"),
+        (Decimal("1.00000"), "GBP", 6, "£1.00000"),
+        (Decimal("3000.445"), "GBP", 6, "£3,000.445"),
```
MebinAbraham, line 54: Minor, but like the ones below, it would be clearer if this was organise by like input length currency type etc.

tests/app/utilities/test_decimal_places.py
```diff
@@ -0,0 +1,129 @@
...
+        (100000000, "en_GB", "100,000,000"),
+        (0, "en_GB", "0"),
+        (0, "de_DE", "0"),
+        (Decimal("0.00"), "en_GB", "0.00"),
+        (Decimal("0.000"), "en_GB", "0.000"),
+        (Decimal("0.000"), "es_ES", "0,000"),
+        (Decimal("0.00000"), "en_GB", "0.00000"),
+        (Decimal("0.00000"), "es_ES", "0,00000"),
+    ),
+)
```
MebinAbraham, line 43: Why is this different to `TEST_FORMAT_NUMBER_PARAMS`, they both use `custom_format_decimal`. The setting of locale can be done by the test, so you could use this for both?.

tests/app/utilities/test_decimal_places.py
```diff
@@ -0,0 +1,129 @@
...
+        (Decimal("1.20000"), "kilometer", "de_DE", "long", "1,20000 Kilometer"),
+        (Decimal("1.2"), "kilometer", "pl_PL", "long", "1,2 kilometra"),
+        (Decimal("1.2345"), "kilometer", "en_GB", "long", "1.2345 kilometres"),
+        (Decimal("1.2345"), "kilometer", "es_ES", "long", "1,2345 kilómetros"),
+        (123, "mile", "en_GB", "short", "123 mi"),
+        (123, "mile", "en_GB", "narrow", "123mi"),
+        (123, "mile", "en_US", "narrow", "123mi"),
+    ),
+)
+def test_custom_format_unit(
```
MebinAbraham, line 87: Same as above, why do we have this and `TEST_FORMAT_UNIT_PARAMS`. If you handle the locale you have a single test set for both?.

tests/app/utilities/test_decimal_places.py
```diff
@@ -0,0 +1,129 @@
...
+        (2, "GBP", "en_GB", 6, "£2.00"),
+        (Decimal("2.14564"), "GBP", "en_GB", 6, "£2.14564"),
+        (Decimal("1.1"), "GBP", "en_GB", 6, "£1.10"),
+        (Decimal("1.10"), "GBP", "en_GB", 6, "£1.10"),
+        (Decimal("1.100"), "GBP", "en_GB", 6, "£1.100"),
+        (Decimal("1.1000"), "GBP", "en_GB", 6, "£1.1000"),
+        # (Decimal("3000.445"), "GBP", "en_GB", 6, "£3,000.445"),
+    ),
+)
+def test_custom_format_currency(
```
MebinAbraham, line 124: Same as above, why do we have this and `TEST_FORMAT_CURRENCY_PARAMS`. If you handle the locale you have a single test set for both?.

app/forms/validators.py
berroar: Updated 👍

app/views/contexts/calculated_summary_context.py
berroar: Updated 👍

app/forms/validators.py
berroar: Set to None now 👍

app/views/contexts/calculated_summary_context.py
berroar: Updated now 👍

tests/app/utilities/test_decimal_places.py
berroar: Updated 👍

tests/app/utilities/test_decimal_places.py
berroar: Updated 👍

tests/app/utilities/test_decimal_places.py
berroar: Updated 👍

app/forms/validators.py
berroar: Have reverted this temporarily, there is an issue in `_get_target_total_and_currency` which i think is always returning an int when calculating the target total. So that method needs to be fixed

app/jinja_filters.py
berroar: This awkward because was trying to differentiate between number of decimal places and the decimal limit set in the schema, but in some instances we use the number of decimal places as the limit

app/forms/validators.py
berroar: isinstance evaluates as False for field.type as DecimalFieldWithSeparator (evaluates True for str)

app/jinja_filters.py
MebinAbraham: I would name it relevant to the usage. If it is used to work out the number of places, then it should be called limit, if it is used as the literal places, then use places.

app/forms/validators.py
MebinAbraham: It shouldn't be `isinstance` on `field.type`, it should be `isinstance` on field itself.

app/forms/validators.py
MebinAbraham: The total should return based on no. of places we need to display.

berroar: Yeah I get the logic, like it's just that that method at the moment doesn't return a decimal so at the moment it is not possible to figure out the correct number of decimals to be displayed. So looks like that `_get_target_total_and_currency` needs to be refactored

app/forms/validators.py
berroar: Updated 👍

app/forms/validators.py
MebinAbraham: Is there a lot to refactoring?. Don't you just need to return the max decimal place for the answers to calculate?.
It would be done in `_validate_calculated_question` since that is where the `SumCheck` validator is called, and you have calculation and question in scope.

Same thing applies to calc/grand calc summaries. I think it is an issue there also, as it currently get the answer format for the first answer. Do we have any validation to ensure all answers must have same no. of decimal places?.

berroar: No there is nothing stopping answers having different decimal limits which are then being used in calc summaries etc, added a schema for this too 👍

berroar: Added a fix for the initial issue now 👍

app/forms/questionnaire_form.py
```diff
@@ -203,27 +204,32 @@ def _get_target_total_and_currency(
...
+                result["target_total"] = self.value_source_resolver.resolve(calculation["value"])  # type: ignore
             else:
-                calculation_value = calculation["value"]
-            currency = question.get("currency")
-            return calculation_value, currency
+                result["target_total"] = calculation["value"]
+            result["currency"] = question.get("currency")
+            result["decimal_places"] = self.schema.get_decimal_limit(
+                calculation["answers_to_calculate"]
+            )
```
MebinAbraham, line 218: If you do this at the very start, you don't need to duplicate it below for non value one.

app/forms/validators.py
MebinAbraham: Great stuff.

app/forms/questionnaire_form.py
berroar: Updated 👍

app/questionnaire/placeholder_parser.py
```diff
@@ -178,6 +180,23 @@ def _parse_transforms(
...
+                    )
+
+                if (
+                    transform["transform"] == "format_currency"
+                    and arg_value.get("source") == "calculated_summary"
+                ):
+                    answer_ids = get_calculated_summary_answer_ids(self._schema.get_block(arg_value["identifier"]))  # type: ignore
+                    transform_args["decimal_limit"] = self._schema.get_decimal_limit(
+                        answer_ids
+                    )
```
MebinAbraham, line 198: We should review this implementation as this isn't pretty fragile. We should avoid hard coding logic like this in general. We might have other cases in future in which based on the argument, we need to do custom logic, We should leave it up to the transform to handle any logic special to them. Placeholder parser should be generic. It should be done similar to `TRANSFORMS_REQUIRING_ROUTING_PATH`. 

What is essentially happening is, some transforms might need to use the raw arguments to work out some info needed for the final transform. We could pass in these unresolved argument to the transform then the transform can do whatever special logic it wants. It means each transform handles any special logic to them and not placeholder parser. I suggest introducing a constant like `TRANSFORMS_REQUIRING_UNRESOLVED_ARGUMENTS` then if transform name is in them, it should also pass along the raw arguments which the transform can use to handle the special cases.

If anything is not clear let me know.

app/questionnaire/placeholder_parser.py
```diff
@@ -178,6 +180,23 @@ def _parse_transforms(
 
                 transform_args[arg_key] = resolved_value
 
+                if (
+                    transform["transform"] == "format_currency"
+                    and arg_value.get("source") == "answers"
```
MebinAbraham, line 185: The source could be `previous_transform` which wouldn't currently work. This would be a tricky one, and not something we could easily handle as the transformed value could go through any number of transformation. Therefore, we might have to make it a an exception and document in the decision log and say format currency when used with previous transform won't take into account the schema limit.
The alternative would be to enforce author to set the decimal places explicitly. (We will likely do this, if this ever becomes a problem)

app/helpers/form_helpers.py
```diff
@@ -21,9 +22,15 @@ def sanitise_mobile_number(data: str) -> str:
     return re.sub(r"^(0{1,2}44|\+44|0)", "", data)
 
 
-def format_playback_value(value: float | Decimal, currency: str | None = None) -> str:
+def format_playback_value(
+    value: float | Decimal, currency: str | None = None, decimal_limit: int | None = 0
```
MebinAbraham, line 26: If decimal limit is defaulted to 0, then you can remove the None.

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,105 @@
...
+        locale=locale,
+    )
+
+
+def get_formatted_currency(
+    *,
+    value: float | Decimal,
+    currency: str = "GBP",
+    locale: str | None = None,
+    decimal_limit: int | None = 0,
```
MebinAbraham, line 30: If decimal limit is defaulted to 0, then you can remove the None.

app/forms/questionnaire_form.py
```diff
@@ -283,13 +291,19 @@ def _validate_calculated_question(
...
+        decimal_places: int | None,
     ) -> bool:
         messages = None
         if "validation" in question:
             messages = question["validation"].get("messages")
 
         validator = SumCheck(messages=messages, currency=currency)
 
+        if decimal_places:
+            formatted_total = Decimal(format(target_total, f".{decimal_places}f"))
```
MebinAbraham, line 303: I don't follow why this is being done. Can you clarify.

app/questionnaire/placeholder_parser.py
berroar: Updated this 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -1024,15 +1024,28 @@ def get_answer_format_for_calculated_summary(
...
 
+    def get_decimal_limit(self, answer_ids: list[str]) -> int:
+        decimal_limit = 0
+        for answer_id in answer_ids:
+            for answer in self.get_answers_by_answer_id(answer_id):
+                if decimal_places := answer.get("decimal_places"):
+                    if decimal_places > decimal_limit:
+                        decimal_limit = decimal_places
+
+        return decimal_limit
```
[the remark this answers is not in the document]

app/helpers/form_helpers.py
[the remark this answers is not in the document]
berroar: Updated 👍

app/forms/questionnaire_form.py
```diff
@@ -203,27 +204,34 @@ def _get_target_total_and_currency(
...
         question: QuestionSchemaType,
-    ) -> Optional[tuple[Union[Calculation, AnswerValueTypes], Optional[str]]]:
-        calculation_value: Union[Calculation, AnswerValueTypes]
-        currency: Optional[str]
+    ) -> dict:
+        result = {
+            "decimal_places": self.schema.get_decimal_limit(
+                calculation["answers_to_calculate"]
+            ),
+            "currency": None,
```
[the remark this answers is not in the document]
berroar: Updated 👍

app/forms/questionnaire_form.py
berroar: Forgot to remove this and pass it to the method, fixed now 👍

app/questionnaire/placeholder_parser.py
```diff
@@ -178,6 +180,9 @@ def _parse_transforms(
 
                 transform_args[arg_key] = resolved_value
 
+                if transform["transform"] in TRANSFORMS_REQUIRING_UNRESOLVED_ARGUMENTS:
+                    transform_args["raw_args"] = arg_value
```
MebinAbraham, line 184: This would only ever set 1 value. It needs to set the args itself, not just the value. Needs to be moved outside the for loop.

app/questionnaire/placeholder_transforms.py
```diff
@@ -41,8 +47,29 @@ def __init__(
...
-    def format_currency(self, number: Decimal | float, currency: str = "GBP") -> str:
-        formatted_currency: str = format_currency(number, currency, locale=self.locale)
+    def format_currency(
+        self,
+        number: int | Decimal | float,
+        currency: str = "GBP",
+        decimal_limit: int | None = None,
+        raw_args: Mapping | None = None,
+    ) -> str:
+        if raw_args:
```
MebinAbraham, line 57: We need to doc string what raw args are used for and when they might not exist etc.

app/questionnaire/placeholder_transforms.py
```diff
@@ -41,10 +47,39 @@ def __init__(
...
-        formatted_currency: str = format_currency(number, currency, locale=self.locale)
+    def format_currency(
+        self,
+        number: int | Decimal | float,
+        raw_args: Mapping,
+        currency: str = "GBP",
+    ) -> str:
+        """
+        The raw arguments for the transform are required here, in addition to the formatted number, as custom logic is required
+        to calculate the correct number of decimals based on the source of the transform.
```
MebinAbraham, line 58: It should mention, the decimal only takes into account if source is answer or calc summary without any previous transform.

We need to also capture that in the decision logs.

app/questionnaire/placeholder_parser.py
```diff
@@ -153,12 +156,15 @@ def _parse_transforms(
...
         transformed_value: TransformedValueTypes = None
-
         for transform in transform_list:
             transform_args: MutableMapping = {}
             value_source_resolver = self._get_value_source_resolver_for_transform(
                 transform
             )
+
+            if transform["transform"] in TRANSFORMS_REQUIRING_UNRESOLVED_ARGUMENTS:
+                transform_args["raw_args"] = transform["arguments"]
```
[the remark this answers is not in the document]
berroar: Yeah I added an integration test that test the asserts the currency value on the playback page

schemas/test/en/test_calculated_and_grand_calculated_summary_decimals.json
```diff
@@ -0,0 +1,271 @@
+{
```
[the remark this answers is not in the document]
berroar: Just felt that adding the new schema made it easier to create simple tests for calculated and grand calculated summaries to demo decimals - in one schema, with the need to update existing schemas for calc summary and grand calc summary separately and having to update any other tests that we be affected by the change.

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    Generates the number format based on the value entered by the user and the locale
+
+    Format follows the number formats as specified in the babel docs e.g: '#,##0.###'
+
+    Uses the decimal places set by the user with frac_prec to ensure that trailing zeroes
+    are not dropped and that the correct number of decimal places as entered by the user are displayed
+    after formatting.
+    """
+    decimal_places = len(str(value).split(".")[1]) if "." in str(value) else 0
```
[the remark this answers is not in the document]
berroar: Added a method 👍

katie-gardner (commented):
Functionally looks to all be working the way I'd expect, with other schemas, modifying decimals to try and hit edge cases, and the new one, a few small comments/questions

app/questionnaire/placeholder_parser.py
berroar: Added to the decision log now 👍

app/questionnaire/placeholder_transforms.py
berroar: Added to the decision log now 👍

schemas/test/en/test_calculated_and_grand_calculated_summary_decimals.json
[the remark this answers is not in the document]
MebinAbraham: Fair enough, was a non blocker and probably appropriate given the different cases anyway.

app/questionnaire/questionnaire_schema.py
```diff
@@ -36,6 +36,7 @@
 DependencyDictType: TypeAlias = dict[str, OrderedSet[str]]
 
 TRANSFORMS_REQUIRING_ROUTING_PATH = {"first_non_empty_item"}
+TRANSFORMS_REQUIRING_UNRESOLVED_ARGUMENTS = ["format_currency"]
```
MebinAbraham, line 39: Minor for consistency, this and the arg name should match, both should either be raw arg or unresolved arg.

app/questionnaire/placeholder_parser.py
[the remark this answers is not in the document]
MebinAbraham: We should be adding unit tests first then integration test if needed. But won't block in this instance.

schemas/test/en/test_calculated_and_grand_calculated_summary_decimals.json
```diff
@@ -0,0 +1,271 @@
...
+            "groups": [
+                {
+                    "id": "group",
+                    "title": "Total a range of values",
+                    "blocks": [
+                        {
+                            "type": "Question",
+                            "id": "first-number-block",
+                            "question": {
+                                "id": "first-number-question",
```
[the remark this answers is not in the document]
katie-gardner: Change is exactly what was hoping for 👌 - can we have the same for the 5th and 6th question?

tests/app/test_jinja_filters.py
```diff
@@ -18,16 +19,147 @@
...
     strip_tags,
 )
 from app.questionnaire.questionnaire_schema import QuestionnaireSchema
+from app.utilities.decimal_places import get_formatted_currency
 from app.utilities.schema import load_schema_from_name
 
+TEST_FORMAT_CURRENCY_PARAMS = (
+    "value, currency, locale_string, decimal_limit, expected_result",
+    [
+        (Decimal("2.1"), "GBP", "en_GB", None, "£2"),
```
MebinAbraham, line 35: As discussed in tech session, I don't think this is correct. This would cause problems with the previous transform scenario. If the value has a decimal, then the expected value should never be cropped. We can safely assume, if the value has a decimal, then they were allowed to enter that.
For this I would expect the answer to be £2.10

app/utilities/decimal_places.py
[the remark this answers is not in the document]
MebinAbraham: I think point 1 is still outstanding.

tests/app/test_jinja_filters.py
MebinAbraham: This example would fail currently:

```py
def test_multiple_metadata_transform_format_currency_placeholder(
    mock_renderer, mock_schema, mock_location
):
    placeholder_list = [
        {
            "placeholder": "total",
            "transforms": [
                {
                    "transform": "add",
                    "arguments": {
                        "lhs": Decimal("1.23"),
                        "rhs": Decimal("1"),
                    },
                },
                {
                    "transform": "format_currency",
                    "arguments": {
                        "number": {"source": "previous_transform"}
                    },
                },
            ],
        }
    ]

    metadata = get_metadata()

    parser = PlaceholderParser(
        language="en",
        answer_store=AnswerStore(),
        list_store=ListStore(),
        metadata=metadata,
        response_metadata={},
        schema=mock_schema,
        renderer=mock_renderer,
        progress_store=ProgressStore(),
        location=mock_location,
        supplementary_data_store=SupplementaryDataStore(),
    )

    placeholders = parser(placeholder_list)

    assert placeholders["total"] == "£2.23"
```

```sh
tests/app/questionnaire/test_placeholder_parser.py:53 (test_multiple_metadata_transform_format_currency_placeholder)
'£2' != '£2.23'

Expected :'£2.23'
Actual   :'£2'
```

app/utilities/decimal_places.py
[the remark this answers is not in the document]
berroar: Updated now 👍

schemas/test/en/test_calculated_and_grand_calculated_summary_decimals.json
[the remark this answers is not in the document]
berroar: Oops missed those, added now 👍

app/questionnaire/questionnaire_schema.py
berroar: Updated 👍

tests/app/test_jinja_filters.py
berroar: Updated as discussed, and added the above test cases 👍

tests/app/test_jinja_filters.py
```diff
@@ -32,20 +32,41 @@
...
-        (Decimal("2.1"), "GBP", "en_GB", None, "£2"),
-        (Decimal("3000.44545"), "GBP", "en_GB", None, "£3,000"),
-        (Decimal("123.1234"), "GBP", "en_GB", 0, "£123"),
+        # This test case assumes that the number of decimal places entered by the
+        # user has already been validated before reaching the get_formatted_currency method.
+        # When there is no decimal limit set, providing the user has entered a decimal, then the currency local
+        # currency precision will take precedence if the number of decimal places entered by the user is less than the
+        # currency precision value. If the number of decimal places entered by the user is greater than the currency
+        # precision value then we will display the number of decimal places as entered by the user
+        # The Jordanian Dinar is used as an example in this test case as the currency precision is set to .000
```
katie-gardner, line 41: Nicely explained 👍

katie-gardner (commented):
Tried whatever I can to break it, and it never crops any decimals, the only weird case I could find was this one
Screenshot 2023-09-07 at 14 37 50
calculated summaries will call `_set_decimal_places` for each answer in turn, ensuring the decimals are the maximum of all answer, grand calculated summary context doesn't do that though so if the first calculated summary has 1dp and the second calculated summary has more but the user doesn't enter them, the GCS displays the 1dp, I think this was our "should never get here case" but in this instance we do

(line 121 of the GCS context ` # validator ensures all calculated summaries are of the same type, so the first can be used for the format`)
Could get the max dp of the calculated summaries? such an edge case though...

app/utilities/decimal_places.py
```diff
@@ -45,17 +46,26 @@ def get_formatted_currency(
...
+        decimal_limit is not None
+        and currency_precision > decimal_limit >= decimal_places
+    ) or (decimal_limit is None and not decimal_places):
+        currency_digits = False
+    else:
+        currency_digits = decimal_places < currency_precision
 
     # The number of decimals displayed is limited based on the value of the `decimal_limit` parameter, if not set, defaults to 0.
     # If the number of decimal places entered by the user is less than the `decimal_limit` then we should display the
     # number of decimals as entered by the user.
```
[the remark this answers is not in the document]
berroar: Updated this 👍

berroar:
> Tried whatever I can to break it, and it never crops any decimals, the only weird case I could find was this one Screenshot 2023-09-07 at 14 37 50 calculated summaries will call `_set_decimal_places` for each answer in turn, ensuring the decimals are the maximum of all answer, grand calculated summary context doesn't do that though so if the first calculated summary has 1dp and the second calculated summary has more but the user doesn't enter them, the GCS displays the 1dp, I think this was our "should never get here case" but in this instance we do
> 
> (line 121 of the GCS context ` # validator ensures all calculated summaries are of the same type, so the first can be used for the format`) Could get the max dp of the calculated summaries? such an edge case though...

Added a fix for this as suggested 👍

app/forms/validators.py
```diff
@@ -424,9 +449,20 @@ def __call__(
         is_valid, message = self._is_valid(condition, total, target_total)
 
         if not is_valid:
+            decimal_limit = decimal_limit or (
+                None
+                if isinstance(target_total, int)
+                else str(target_total)[::-1].find(".")
```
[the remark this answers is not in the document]
katie-gardner: based off how answer store works 2 would only ever end up here as an integer, so I'll resolve this

app/questionnaire/questionnaire_schema.py
```diff
@@ -1033,15 +1034,36 @@ def get_answer_format_for_calculated_summary(
...
         return {
             "type": first_answer["type"].lower(),
             "unit": first_answer.get("unit"),
             "unit_length": first_answer.get("unit_length"),
             "currency": first_answer.get("currency"),
+            "decimal_places": decimal_limit,
         }
 
+    def get_grand_calculated_summary_decimal_limit(
+        self, calculated_summary_ids: list[str]
```
katie-gardner, line 1050: I'd favour this being in `grand_calculated_summary_context.py` over here since its only used there and specific to it (maybe it could be inline as well) but this does fix the issue 🙌

katie-gardner (approved):
Functionally all working in the many cases I've tried! minor suggestion we could take or leave

app/questionnaire/placeholder_transforms.py
```diff
@@ -41,10 +47,40 @@ def __init__(
...
         return formatted_currency
 
+    def _get_decimal_limit(self, unresolved_arguments: Mapping) -> int | None:
+        decimal_limit = None
+        source = unresolved_arguments["number"].get("source")
+        identifier = unresolved_arguments["number"].get("identifier")
+        if source == "answers":
+            decimal_limit = self.schema.get_decimal_limit([identifier])
+
+        if source == "calculated_summary":
```
MebinAbraham, line 76: This should be `elif`.

app/views/contexts/calculated_summary_context.py
```diff
@@ -298,20 +294,34 @@ def _get_answer_format(self, groups: Iterable[Mapping]) -> Tuple[dict, list]:
                             "unit": answer.get("unit"),
                             "unit_length": answer.get("unit_length"),
                             "currency": answer.get("currency"),
+                            "decimal_places": 0,
```
MebinAbraham, line 297: The default should be None. Handling 0 will be done by the format method.

app/views/contexts/calculated_summary_context.py
```diff
@@ -298,20 +294,34 @@ def _get_answer_format(self, groups: Iterable[Mapping]) -> Tuple[dict, list]:
...
                     answer_value = answer.get("value") or 0
                     values_to_calculate.append(answer_value)
 
         return answer_format, values_to_calculate
 
+    @staticmethod
+    def _set_decimal_places(answer: Mapping, answer_format: dict) -> None:
+        if decimal_places := answer.get("decimal_places"):
+            if decimal_places > answer_format["decimal_places"]:
+                answer_format["decimal_places"] = decimal_places
```
MebinAbraham, line 311: Once the above is done, this will need updating to handle decimal place 0 as it will be falsely. I think it would be better to do it with a single var in the main method but will leave it up to you as long as the None is default.

```py
    def _get_answer_format(self, groups: Iterable[Mapping]) -> Tuple[dict, list]:
        ...
        decimal_limits = []

        for group in groups:
            for block in group["blocks"]:
                ...
                for answer in question["answers"]:
                    ...

                    if (decimal_places := answer.get("decimal_places")) is not None:
                        decimal_limits.append(decimal_places)

        answer_format["decimal_places"] = max(decimal_limits, default=None)
        return answer_format, values_to_calculate
```

app/questionnaire/questionnaire_schema.py
```diff
@@ -1033,15 +1034,36 @@ def get_answer_format_for_calculated_summary(
...
+            decimal_limits.append(self.get_decimal_limit(answer_ids))
+        return max(decimal_limits)
+
+    def get_decimal_limit(self, answer_ids: list[str]) -> int:
+        decimal_limits: list[int] = [
+            answer.get("decimal_places", 0)
+            for answer_id in answer_ids
+            for answer in self.get_answers_by_answer_id(answer_id)
+        ]
+        return max(decimal_limits)
```
MebinAbraham, line 1065: We should return None if not set for any. If none of them set it, this would return 0 but we should return None. Will have a chain reaction of some callers.

MebinAbraham (commented):
Looks good, just some minor comments. 👍

app/views/contexts/calculated_summary_context.py
katie-gardner: Agree - prefer single var in method too

app/questionnaire/questionnaire_schema.py
katie-gardner: @MebinAbraham as we discussed I didn't do this but changed the name to `get_decimal_limit_from_sources` with a minor refactor - was this what you had in mind?

petechd (approved):

app/questionnaire/questionnaire_schema.py
MebinAbraham: I was thinking sources as in the values sources itself, since this only handles calc summary and passed in ids instead of sources it should be specific to calc summary, like `get_decimal_limit_from_calculated_summaries`

app/utilities/decimal_places.py
```diff
@@ -0,0 +1,124 @@
...
+        locale=locale,
+    )
+
+
+def get_formatted_currency(
+    *,
+    value: float | Decimal,
+    currency: str = "GBP",
+    locale: str | None = None,
+    decimal_limit: int | None = 0,
```
MebinAbraham, line 31: Default should be None
```suggestion
    decimal_limit: int | None = None,
```

MebinAbraham (approved):
:tada: 👍

app/questionnaire/questionnaire_schema.py
katie-gardner: Ahh ok - renamed in the interests of getting it across 👍

app/helpers/form_helpers.py
```diff
@@ -21,9 +22,17 @@ def sanitise_mobile_number(data: str) -> str:
...
 
 
-def format_playback_value(value: float | Decimal, currency: str | None = None) -> str:
+def format_playback_value(
+    value: float | Decimal,
+    currency: str | None = None,
+    decimal_limit: int | None = None,
+) -> str:
+    # For playback of values we set the decimal limit based on the number of decimal places
+    # in the given value rather than any limit that has been set in the schema as we do for other methods
```
Yuyuutsu, line 31: Is it okay to use` """ """ `rather then `#` so the comment "flows" better when reading.

katie-gardner: Yep - good spot, have fixed the ones in the integration tests too

PR #1152, opened 2023-07-03
Decrypt Supplementary Data form SDS API

kylelawsonAND:
It is required for prepop that the data payload retrieved from the SDS API be encrypted to ensure the transmission of the data is secure. 
This PR implements decryption of the SDS payload, as well as updating the mock SDS API to encrypt the payload. The encryption mechanism is symmetric, as such both SDS (and the mock) and EQ will share an encryption key to encrypt and decrypt the payload. This is distinct from submission and authentication which uses asymmetric public and private keys. 

### Changes

**app/keys.py**: Adding a new key type for supplementary data decryption.

**app/routes/errors.py**: Handling the new exception `MissingSupplementaryDataKey` by displaying the 500 internal server error page. 

**app/services/supplementary_data.py**: Adding new exception `MissingSupplementaryDataKey`.
Method `get_supplementary_data` first checks that the Supplementary Data decryption key exists in the `KeyStore`, if it does not, it throws a `MissingSupplementaryDataKey` to render the internal server error page. This method also now calls the new method `decrypt_supplementary_data` after decoding the successful SDS API response payload.
New method `decrypt_supplementary_data` added which expects a MutableMapping object which should be the decoded payload from a successful SDS API request. Throws a Validation exception if either the `encryption_key_id` or `data` JSON key-values are not present. Using the key in the `KeyStore` specified with the `encryption_key_id` in `supplementary_data`, the value for the key `data` is decrypted using `sdc.crypto` helper methods. This decrypted object is set as the value for `data`, and the modified `supplementary_data` is returned. 
New method `get_key_store` added so the KeyStore can be retrieved without code duplication across methods in this file. 

**app/setup.py**:  `validate_required_keys` called for the Authenication key purpose, ensuring at start-up, the public and private keys used for Authentication, are present in the executing environment's keys file. 

**dev-keys.yml**: Adding a **development only** key for use by both EQ and SDS mock to decrypt and encrypt respectively. Generated using the method for private keys specified [here](https://confluence.ons.gov.uk/display/SDC/Provision+new+keys). 

**scripts/mock_data/**: Adding the `encryption_key_id` key-value to the SDS data payloads. 

**scripts/mock_sds_endpoint.py**: New method `encrypt_mock_data` which takes a mock sds data message and encrypted the value of `data` using helper methods from `sdc.crytpo` along with the new encryption key added to `dev-keys.yml`. 

**tests/app/services/test_request_supplementary_data.py**: Amended existing supplementary data tests to mock an SDS API response to contain an encrypted payload, and to then later assert the decrypted payload is present. Where required additional patches have been added to mock the method `decrypt_supplementary_data` to return the decrypted payload. 
Added tests for `decrypt_supplementary_data` to test successful decryption and failure paths in isolation. 
Added test for exception when the encryption key does not exist in the `KeyStore`.

**tests/integration/integration_test_case.py**: Adding the supplementary data key to the app set up `KeyStore`.

**tests/integration/routes/test_session.py**: Added tests for invalid supplementary data being returned form the API call, and missing encryption key. Happy path decryption is covered by existing tests. 
Removed duplicated code which checked that the 500 internal server error page had rendered for many tests. 

**tests/jwt-test-keys/sdc-sds-supplementary_data-encryption-private-v1.pem**: The supplementary data key that was generated and is used throughout these changes. 

### How to review
Existing supplementary data payloads should be successfully loading into the survey when launched.
Place breakpoints in the mock SDS API where encryption takes place, and verify the data is encrypted. 
Place a breakpoint on line 74 of `supplementary_data.py` and observe that an encrypted payload is delivered, and is decrypted and interpreted as a python object.


app/services/supplementary_data.py
```diff
@@ -78,6 +91,24 @@ def get_supplementary_data(*, dataset_id: str, unit_id: str, survey_id: str) ->
     raise SupplementaryDataRequestFailed
 
 
+def decrypt_supplementary_data(supplementary_data: MutableMapping) -> Mapping:
+    if (encryption_key_id := supplementary_data.get("encryption_key_id")) and (
```
MebinAbraham, line 95: Having a second look, the `kid` would be in the header (after deserialisation) therefore we don't actually need this property right?. If so, we should consider changing the approach to not include this in the response as it is redundant.

Can you confirm this. If so, I'll chat to the other about dropping this key.

app/services/supplementary_data.py
```diff
@@ -78,6 +91,24 @@ def get_supplementary_data(*, dataset_id: str, unit_id: str, survey_id: str) ->
...
+        encrypted_data := supplementary_data.get("data")
+    ):
+        key = (
+            get_key_store()
+            .get_private_key_by_kid(KEY_PURPOSE_SDS, encryption_key_id)
+            .as_jwk()
```
MebinAbraham, line 101: If the above is done, we can just use the `decrypt` method.

kylelawsonAND:
[SDS data decryption - Runner implementation (3 SP)](https://trello.com/c/9f9jyLOu/5707-sds-data-decryption-runner-implementation-3-sp)

app/services/supplementary_data.py
kylelawsonAND: Confirmed can get the kid out of the jwe header so encryption_key_id can be dropped. 
Updated code and tests to reflect

app/services/supplementary_data.py
kylelawsonAND: Done 👍

app/routes/errors.py
```diff
@@ -191,8 +195,10 @@ def too_many_feedback_requests(
 
 
 @errors_blueprint.app_errorhandler(SupplementaryDataRequestFailed)
+@errors_blueprint.app_errorhandler(MissingSupplementaryDataKey)
+@errors_blueprint.app_errorhandler(InvalidSupplementaryData)
 def supplementary_data_request_failed(
-    exception: SupplementaryDataRequestFailed,
+    exception: SupplementaryDataRequestFailed | MissingSupplementaryDataKey,
```
MebinAbraham, line 201: Should also add `InvalidSupplementaryData`

app/services/supplementary_data.py
```diff
@@ -30,7 +33,20 @@ def __str__(self) -> str:
...
+
+
+class InvalidSupplementaryData(Exception):
+    def __str__(self) -> str:
+        return "Supplementary data invalid"
+
+
 def get_supplementary_data(*, dataset_id: str, unit_id: str, survey_id: str) -> dict:
+    if not get_key_store().get_key(purpose=KEY_PURPOSE_SDS, key_type="private"):
+        raise MissingSupplementaryDataKey()
```
MebinAbraham, line 48: Non blocker, if you are not passing anything into the exception, you don't need to instantiate it. https://docs.python.org/3/tutorial/errors.html#raising-exceptions
```suggestion
        raise MissingSupplementaryDataKey
```

dev-keys.yml
```diff
@@ -87,3 +87,62 @@ keys:
       V7ZuRp//4vIzEONqqyLUFt1JZKuCwWQB2sRjKWpqRB07C0o243XefD5f7844nmXq
       H0xKMFRYjDqhFdTXFWVPnk/tu/MxoSTdK0FfsgUnjs+TGvN5TnBW
       -----END RSA PRIVATE KEY-----
+  df88fdad2612ae1e80571120e6c6371f55896696:
+    platform: sdc
+    purpose: supplementary_data
+    service: sds
+    type: private
+    use: encryption
```
MebinAbraham, line 95: Minor and non blocker, fancy adding the new attributes to the old keys. All new keys and prod keys have all these properties.

scripts/mock_data/supplementary_data_no_repeat.json
```diff
@@ -1,8 +1,9 @@
 {
   "dataset_id": "c067f6de-6d64-42b1-8b02-431a3486c178",
   "survey_id": "123",
+  "encryption_key_id": "df88fdad2612ae1e80571120e6c6371f55896696",
```
MebinAbraham, line 4: To be removed.

scripts/mock_data/supplementary_data_with_repeat.json
```diff
@@ -1,6 +1,7 @@
 {
   "dataset_id": "34a80231-c49a-44d0-91a6-8fe1fb190e64",
   "survey_id": "123",
+  "encryption_key_id": "df88fdad2612ae1e80571120e6c6371f55896696",
```
MebinAbraham, line 4: Can be removed.

scripts/mock_data/supplementary_data_with_repeat_v2.json
```diff
@@ -1,6 +1,7 @@
 {
   "dataset_id": "6b378962-f0c7-4e8c-947e-7d24ee1b6b88",
   "survey_id": "123",
+  "encryption_key_id": "df88fdad2612ae1e80571120e6c6371f55896696",
```
MebinAbraham, line 4: Can be removed.

app/services/supplementary_data.py
```diff
@@ -78,6 +96,22 @@ def get_supplementary_data(*, dataset_id: str, unit_id: str, survey_id: str) ->
     raise SupplementaryDataRequestFailed
 
 
+def decrypt_supplementary_data(supplementary_data: MutableMapping) -> Mapping:
+    if encrypted_data := supplementary_data.get("data"):
+        try:
+            decrypted_data = JWEHelper.decrypt(
```
katie-gardner, line 102: Why does this use the `JWEHelper.decrypt` over the `sdc.crypto.decrypter.decrypt` which wraps that method? Authenticator uses the decrypter one so I'm just wondering why this one doesn't, was the check on the tokens not needed here?

app/services/supplementary_data.py
```diff
@@ -78,6 +96,22 @@ def get_supplementary_data(*, dataset_id: str, unit_id: str, survey_id: str) ->
...
+                encrypted_data, key_store=get_key_store(), purpose=KEY_PURPOSE_SDS
+            )
+            supplementary_data["data"] = json.loads(decrypted_data)
+            return supplementary_data
+        except InvalidTokenException as e:
+            raise InvalidSupplementaryData from e
+
+    raise ValidationError(
+        "Supplementary data response cannot be decrypted without the key 'data'"
```
katie-gardner, line 111: Do we go with this or something like 'No data to decrypt'?

scripts/mock_data/supplementary_data_no_repeat.json
katie-gardner: How come this isn't needed?

scripts/mock_sds_endpoint.py
```diff
@@ -1,21 +1,32 @@
...
 
+import yaml
 from flask import Flask, Response, request
+from sdc.crypto.jwe_helper import JWEHelper
+from sdc.crypto.key_store import KeyStore
+
+from app.keys import KEY_PURPOSE_SDS
 
 app = Flask(__name__)
+with open("dev-keys.yml", encoding="UTF-8") as keys_file:
```
katie-gardner, line 11: should this use environment variable instead of the filename directly? I guess its a mock endpoint so it doesn't really matter

tests/app/services/test_request_supplementary_data.py
```diff
@@ -1,20 +1,29 @@
...
 TEST_SDS_URL = "http://test.domain/v1/unit_data"
+EXPECTED_SDS_DECRYPTION_VALIDATION_ERROR = (
+    "Supplementary data response cannot be decrypted without the key 'data'"
+)
 
-mock_supplementary_data_payload = {
+decrypted_mock_supplementary_data_payload = {
     "dataset_id": "44f1b432-9421-49e5-bd26-e63e18a30b69",
     "survey_id": "123",
+    "encryption_key_id": "df88fdad2612ae1e80571120e6c6371f55896696",
```
katie-gardner, line 26: based off above comments, maybe this key isn't needed? though I am unsure where it comes in instead of this

tests/app/services/test_request_supplementary_data.py
```diff
@@ -47,6 +56,27 @@
     },
 }
 
+encrypted_mock_supplementary_data_payload = {
+    "dataset_id": "44f1b432-9421-49e5-bd26-e63e18a30b69",
```
katie-gardner, line 60: since we have a fair few bits of json now, it might be nice to pull these out into a `conftest.py` file for the services folder and keep the tests cleaner looking

tests/app/services/test_request_supplementary_data.py
```diff
@@ -145,9 +179,13 @@ def test_get_supplementary_data_retries_transient_error(app: Flask, mocker):
...
         )
 
+        mocker.patch(
+            "app.services.supplementary_data.decrypt_supplementary_data",
+            return_value=decrypted_mock_supplementary_data_payload,
+        )
         mocker.patch(
             "app.services.supplementary_data.validate_supplementary_data",
-            return_value=mock_supplementary_data_payload,
+            return_value=decrypted_mock_supplementary_data_payload,
```
katie-gardner, line 188: this might not work - but now that you're patching the decrypt to have the correct data, do you still need to patch the `validate_supplementary_data` method?

tests/app/services/test_request_supplementary_data.py
```diff
@@ -185,3 +223,47 @@ def test_get_supplementary_data_max_retries(app: Flask, mocker):
...
+        assert result == decrypted_mock_supplementary_data_payload
+
+
+def test_decrypt_supplementary_data_raises_validation_error_when_encrypted_payload_missing_data(
+    app: Flask,
+):
+    with app.app_context():
+        with pytest.raises(ValidationError) as e:
+            decrypt_supplementary_data(mock_supplementary_data_payload_missing_data)
+        assert EXPECTED_SDS_DECRYPTION_VALIDATION_ERROR in str(e.value)
```
katie-gardner, line 242: super minor, instead of casting, for an exact match you could assert the expected error is one of the items in e.value.messages

app/services/supplementary_data.py
```diff
@@ -90,3 +124,8 @@ def validate_supplementary_data(
         )
     except ValidationError as e:
         raise ValidationError("Invalid supplementary data") from e
+
+
+def get_key_store() -> KeyStore:
```
katie-gardner, line 129: minor - do we need a method for one dict lookup?

tests/integration/routes/test_session.py
```diff
@@ -1,19 +1,35 @@
...
+from tests.app.services.test_request_supplementary_data import TEST_SDS_URL
 from tests.integration.integration_test_case import IntegrationTestCase
 
 TIME_TO_FREEZE = datetime(2020, 1, 1, 12, 0, 0, tzinfo=timezone.utc)
 EQ_SESSION_TIMEOUT_SECONDS = 45 * 60
 BUSINESS_URL = ACCOUNT_SERVICE_BASE_URL
 SOCIAL_URL = ACCOUNT_SERVICE_BASE_URL_SOCIAL
+TEST_SDS_URL = "http://localhost:5003/v1/unit_data"
+
+mock_supplementary_data_payload_missing_data = {
```
katie-gardner, line 22: same as above I think these would be suitable candidates for a conftest function

katie-gardner (changes requested):
Some minor queries/comments but functionally working nicely! looks great

tests/app/services/test_request_supplementary_data.py
katie-gardner: or equal to e.value.messages[0]

app/services/supplementary_data.py
kylelawsonAND: That method expects a public key to exist as well when decoding, all we need is the decryption part as this isn't signed - just a symmetric key scenario here, unless Mebin's above comment resolves to going back to asymmetric.

katie-gardner: ahhhh ok - makes sense, thanks 👍

scripts/mock_data/supplementary_data_no_repeat.json
kylelawsonAND: The encryption key id is encoded within the JWE token header, so can be decoded and used to select the key to use. Mebin is in the process of getting this change reflected on the SDS side.

scripts/mock_sds_endpoint.py
```diff
@@ -44,5 +55,15 @@ def load_mock_sds_dataset_metadata(survey_id: str, period_id: str):
     return Response(status=404)
 
 
+def encrypt_mock_data(mock_data):
+    key = keys.get_private_key_by_kid(
+        purpose=KEY_PURPOSE_SDS, kid=mock_data.get("encryption_key_id")
+    )
+    mock_data["data"] = JWEHelper.encrypt_with_key(
+        json.dumps(mock_data["data"]), key.kid, key.as_jwk()
+    )
```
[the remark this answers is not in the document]
kylelawsonAND: Will stand by for an update 👍

tests/app/services/test_request_supplementary_data.py
kylelawsonAND: Reason given in Mebin's comment above

app/services/supplementary_data.py
kylelawsonAND: Shut up dumb Kyle from the past - I have removed it and just passed the key store (which can be fetched in get_supplementary_data), into the decrypt_supplementary_data method that gets called

app/routes/errors.py
kylelawsonAND: Added 👍

scripts/mock_data/supplementary_data_with_repeat.json
kylelawsonAND: Woops, good spot

scripts/mock_data/supplementary_data_with_repeat_v2.json
kylelawsonAND: Done 👍

app/services/supplementary_data.py
kylelawsonAND: Changed message 👍

tests/app/services/test_request_supplementary_data.py
kylelawsonAND: Great point! Have removed that mock

tests/app/services/test_request_supplementary_data.py
kylelawsonAND: Have changed to ` assert EXPECTED_SDS_DECRYPTION_VALIDATION_ERROR in e.value.messages`

tests/integration/routes/test_session.py
kylelawsonAND: Done 👍

katie-gardner (commented):
Subject to taking another look if the encryption changes, all my comments resolved, looks great 👍

Yuyuutsu (commented):
Had a look through and LGTM. Will have a relook once we get an update on the symmetric and asymmetric 👍

dev-keys.yml
```diff
@@ -87,3 +99,62 @@ keys:
...
+      kDLafMb+Wzd8WcFkeJ/2X9gOs4yhNJ9EWnRUD6kJU3bG1yYze54wk84/7A5TP6GE
+      chbvdYanO4ZvQu9yLq5U9tIj+bL2PoiYa24780nOlFKPa9XWyuZEaRXMPKbsQmKC
+      xc+A6xGbchdG6Vy4MgCnQcXeT1H+2C8CggEBAMvwlPifgjCk2v7ejVPvA0K63aii
+      l8Acv6wtvELeQGHLg/ZmzbHNZG9mExhuHASeIEn/1p64dAUQ6NTvHq78bgZ4sa2M
+      U6surGKebDwcbgMZe5lN6jZhem9DHWyXCLco3FHWFj6bNMuBfOdUemVi4bEukcrM
+      ItfML9fhBIKva5aAWej+lEQydTImhLOfgXbEse7ic+ZWVddpuqNCq137Mcbc1q0t
+      QChlDkY6+uOdYa3xmSCHtDc64ymzBR6XAQHQnX9fqfLWdz3Ytk9s+H34Hrn6mJZ/
+      /MRZOZimld2tdshUdEQ+Kaf4wVLcEwJuW5/NeMbba+iIA/sPqHRkYmVs1v8=
+      -----END RSA PRIVATE KEY-----
+    version: v1
```
MebinAbraham, line 160: Can we also add this property, it seems it was missed.

kylelawsonAND: Good spot! Added, just made them all v1

scripts/mock_sds_endpoint.py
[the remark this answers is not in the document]
kylelawsonAND: Fixed 👍

katie-gardner (approved):

PR #1156, opened 2023-07-05
Add list collector content pages

petechd:
This adds new list collector content block, two new classes `ListCollectorContentBlock` and `ListCollectorContent` (for list context), as well as new `ListContentContext` for summaries. content block uses new `ListCollectorContent` class which inherits from `Question` and uses latest `repeating blocks` feature first introduced [here](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1106). More details on the card.

Some extra changes were made to methods checking section completeness for the hub, before (for a section) we didn't take into account repeating blocks when checking it. Also, section progress was not being checked for repeating blocks when generating the link for the hub.

### How to review
New integration and functional tests were added. Some changes in `generate_pages.py` were needed.

### Checklist


petechd:
[Support for non question list collector (3 SP)](https://trello.com/c/acXLlgAj/5704-support-for-non-question-list-collector-3-sp)

app/views/contexts/summary/group.py
```diff
@@ -168,6 +171,38 @@ def _build_blocks_and_links(
 
                     self.placeholder_text = list_summary_element["empty_list_text"]
 
+            elif block["type"] == "ListCollectorContent":
+                list_collector_content_section: ImmutableDict | None = (
```
katie-gardner, line 175: This is identical to the "ListCollector" case except for the `if view_submitted_response` block, I think it could be collapsed into one with that if block only applying for a ListCollector

app/views/contexts/summary/list_collector_content_block.py
```diff
@@ -0,0 +1,148 @@
...
+from app.views.contexts.summary.block import Block
+
+
+class ListCollectorContentBlock:
+    def __init__(
+        self,
+        routing_path_block_ids: Iterable[str],
+        answer_store: AnswerStore,
+        list_store: ListStore,
+        progress_store: ProgressStore,
```
katie-gardner, line 20: again I think duplication could be lower, either reusing existing block or both inheriting from a base class that already has things like `_get_related_anwers`

app/views/handlers/list_repeating_question.py
```diff
@@ -29,9 +53,8 @@ def get_next_location_url(self) -> str:
                 return_to_answer_id=self._return_to_answer_id,
                 return_to_block_id=self._return_to_block_id,
             )
-            return repeating_block_url
 
-        return super().get_next_location_url()
+        return None
```
katie-gardner, line 57: you shouldn't need the return None, that's implicitly what it does anyway

app/views/handlers/list_collector_content.py
```diff
@@ -0,0 +1,75 @@
...
+
+from flask import url_for
+
+from app.views.contexts.list_content_context import ListContentContext
+from app.views.handlers.question import Question
+
+
+class ListCollectorContent(Question):
+    def __init__(self, *args):
+        self._is_adding = False
```
katie-gardner, line 11: do we need `is_adding` ?

app/views/handlers/list_repeating_question.py
```diff
@@ -14,13 +14,37 @@ def get_next_location_url(self) -> str:
...
             return self.get_section_summary_url()
 
+        if first_incomplete_location := self.get_first_incomplete_repeating_block_url(
+            self.current_location.list_item_id
+        ):
+            return first_incomplete_location
+
+        if (
+            not first_incomplete_location
+            and self.parent_block["type"] == "ListCollectorContent"
```
katie-gardner, line 24: Hmmm, I'm guessing the issue here is that the ListCollectorContent block is being marked as complete as soon as you go past it, and therefore unlike list collectors, you don't go back to the parent and say that you don't have more items to add?

I think this could be problematic as we're bypassing an incomplete block being marked as complete by just routing back to it each time - I reckon you probably need to do something like list collectors `_is_list_collector_complete` so that this works properly

katie-gardner (changes requested):
Haven't finished my review yet but putting up comments so far - I think theres some opportunity to reduce duplication but functionally pretty sound 👍 

Only problem at the moment, is that I can enter the url for the section summary manually and skip to it before ive completed the repeating blocks - see my comment about the list repeating question url routing

app/questionnaire/questionnaire_schema.py
```diff
@@ -697,6 +698,20 @@ def get_repeating_list_for_section(self, section_id: str) -> str | None:
         if repeat := self.get_repeat_for_section(section_id):
             return repeat.get("for_list")
 
+    def get_repeating_blocks_list_for_section(self, section_id: str) -> str | None:
+        # type ignore block always exists at this point
+        if blocks := self.get_blocks_for_section(self.get_section(section_id)):  # type: ignore
```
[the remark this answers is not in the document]
petechd: Wow, correct!

app/questionnaire/questionnaire_schema.py
```diff
@@ -697,6 +698,20 @@ def get_repeating_list_for_section(self, section_id: str) -> str | None:
...
+            for block in blocks:
+                if block.get("repeating_blocks", []):
+                    return block.get("for_list")
+
+    def get_repeating_blocks_block_type(self, section_id: str) -> str | None:
```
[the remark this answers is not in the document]
petechd: I mean, we talked with @berroar  about it and came to a conclusion that you can't have more than one list collector content block in a section but yeah same applies to list collector with repeating blocks I guess so, yeah, makes sense to make it more universal.

app/questionnaire/router.py
```diff
@@ -518,15 +518,27 @@ def get_enabled_section_keys(
...
+            ):
+                for list_item_id in self._list_store[repeating_blocks_list]:
+                    yield (section_id, list_item_id)
+
+                yield (section_id, None)
+
             else:
-                section_key = (section_id, None)
-                yield section_key
+                yield (section_id, None)
```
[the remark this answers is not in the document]
petechd: Of course 🤦

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=list-collector-content-pages
```
katie-gardner, line 2: reminder to revert

tests/functional/generate_pages.py
```diff
@@ -840,6 +840,17 @@ def process_block(
                 page_filename=f'{repeating_block["id"]}-repeating-block.page.js',
             )
 
+    if block["type"] == "ListCollectorContent":
```
katie-gardner, line 843: you could do `if block["type"] in {"ListCollectorContent", "ListCollector"}: ...`
and then you can delete the for loop just above this one as they are duplicates

tests/integration/questionnaire/test_questionnaire_list_collector_content.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.post({"any-companies-or-branches-answer": "Yes"})
+
+        self.add_company("Company A", "123", "No")
+
+        self.post({"any-other-companies-or-branches-answer": "Yes"})
+
+        self.add_company("Company B", "456", "No")
+
+        self.post({"any-other-companies-or-branches-answer": "No"})
```
katie-gardner, line 154: Non issue - don't think we need the gaps every other line

tests/integration/questionnaire/__init__.py
```diff
@@ -12,6 +12,15 @@ def add_person(self, first_name: str, last_name: str):
         number_of_people = self.number_of_people()
         return self.get_list_item_ids()[number_of_people - 1]
 
+    def add_company(self, name: str, number: str, authorised_insurer: str):
+        self.post(
```
katie-gardner, line 16: Since this is only used by `test_questionnaire_list_collector_content.py` and is specific to that test, i think it would be more suitable there

app/views/contexts/list_content_context.py
```diff
@@ -0,0 +1,68 @@
+from typing import Any, Generator, Mapping, Optional
+
+from werkzeug.datastructures import ImmutableDict
+
+from app.views.contexts.context import Context
+
+
+class ListContentContext(Context):
+    def __call__(
```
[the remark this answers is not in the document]
petechd: Removed completely now.

tests/functional/generate_pages.py
petechd: What about list operations?
```python
list_operations = ["add", "edit", "remove"]
```
we don't want them for `ListCollectorContent`...

katie-gardner: yep exactly, so you'd leave that part alone, but iterate over the repeating blocks for either type of list collector, so end result would be
```
if block["type"] == "ListCollector":
    list_operations = ["add", "edit", "remove"]
    for list_operation in list_operations:
        process_block(
            block[f"{list_operation}_block"],
            dir_out,
            schema_data,
            spec_file,
            relative_require,
            page_filename=f'{block["id"]}-{list_operation}.page.js',
        )

if block["type"] in {"ListCollectorContent", "ListCollector"}:
    for repeating_block in block.get("repeating_blocks", []):
        process_block(
            repeating_block,
            dir_out,
            schema_data,
            spec_file,
            relative_require,
            page_filename=f'{repeating_block["id"]}-repeating-block.page.js',
        )
```

app/views/contexts/summary/group.py
petechd: Well, there is one other difference, it initialises different class, but I tried to make it work and reduce duplication.

tests/functional/generate_pages.py
petechd: Ok, understood.

app/views/contexts/summary/list_collector_content_block.py
petechd: Ok, made a base class now that both inherit from, thanks!

petechd:
> Haven't finished my review yet but putting up comments so far - I think theres some opportunity to reduce duplication but functionally pretty sound 👍
> 
> Only problem at the moment, is that I can enter the url for the section summary manually and skip to it before ive completed the repeating blocks - see my comment about the list repeating question url routing

Seems like addressing that issue you mentioned in https://github.com/ONSdigital/eq-questionnaire-runner/pull/1156#discussion_r1254528460 solved the problem, when you try to access summary now you get redirected to list collector content block, thanks for giving me the hints on how to fix this.

app/views/handlers/list_repeating_question.py
petechd: Addressed now, used exactly this with a small tweak (to get the list name) and it works like a charm! thanks.

tests/integration/questionnaire/test_questionnaire_list_collector_content.py
petechd: I am not sure what to do here, I just followed convention from other files like `tests/integration/questionnaire/test_questionnaire_list_collector.py`

app/views/handlers/list_collector_content.py
```diff
@@ -0,0 +1,82 @@
...
+
+        return self.router.get_next_location_url(
+            self._current_location,
+            self._routing_path,
+            self._return_to,
+            self._return_to_answer_id,
+            self._return_to_block_id,
+        )
+
+    def handle_post(self):
```
MebinAbraham, line 66: If we inherit from list collector and didn't override, would it not work?.

templates/partials/question.html
```diff
@@ -14,6 +14,9 @@
 {% set question_instruction = format_paragraphs(question.instruction) %}
 {% set question_error = form.question_errors[question.id] %}
 
+{%- if content.content -%}
+  <h1>{{ content.content.title }}</h1>
+{%- endif -%}
```
MebinAbraham, line 19: How come this was added?

templates/partials/list-content.html
```diff
@@ -0,0 +1,21 @@
...
+  {%- if content.list and content.list.list_items -%}
+    {% set list = content.list %}
+    <div class="ons-u-mb-m">
+      {% include 'partials/summary/list-summary.html' %}
+    </div>
+  {%- endif -%}
+
+  {% for description in contents %}
+    <p>{{ description["description"] }}</p>
+  {% endfor %}
```
MebinAbraham, line 19: All contents should work, this is only doing paragraphs. We should just be calling out to `partials/contents.html`.

Yuyuutsu (commented):
Functionally tested - looks good to me .
Had a look at the code , will look again after other changes have been made 👍 
 Just think when you dont enter any companies it looks a bit funny but I dont think its part of this card
image

app/views/contexts/summary/group.py
```diff
@@ -141,24 +144,37 @@ def _build_blocks_and_links(
...
+                            routing_path_block_ids=routing_path_block_ids,
+                            answer_store=answer_store,
+                            list_store=list_store,
+                            progress_store=progress_store,
+                            metadata=metadata,
+                            response_metadata=response_metadata,
+                            schema=schema,
+                            location=location,
+                            language=language,
+                        )
```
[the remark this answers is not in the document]
petechd: Added and it's working great.

templates/partials/list-content.html
petechd: Using partial now.

tests/integration/questionnaire/test_questionnaire_list_collector_content.py
katie-gardner: Ah ok 👍 looks like it varies, the `test_questionnaire_interstitial` one was what I'd been doing, could be one for team chat on what everyone finds easiest to read, or you can just resolve, not a big deal

templates/partials/block.html
```diff
@@ -2,8 +2,11 @@
...
 <div class="block" id="{{content.block.id}}">
 
-  {% if 'question' in block %}
-      {% set question = block['question'] %}
-      {% include 'partials/question.html' %}
-  {% endif %}
+{%-  if block['question'] -%}
+  {% set question = block['question'] %}
+  {% include 'partials/question.html' %}
+{%- elif content["list"] and content["list"] -%}
```
katie-gardner, line 8: the `content["list"]` check is done twice

katie-gardner (commented):
Prior comments addressed and a couple added but still working nicely 👍 
Will re-review when comments around inheriting to reduce code are in, think its a good idea

app/views/handlers/list_collector_content.py
MebinAbraham: It would error as is yes, but we easily update `_get_answer_action` to handle question not existing, or you can override the method in this content file to return None.

templates/partials/question.html
petechd: We were experimenting with templates and forgot to remove it after, fixed now.

petechd:
> Functionally tested - looks good to me . When you dont enter any companies it looks a bit funny but I dont think its part of this card. Had a look at the code , will look again after other changes have been made 👍

That is the ugly consequence of not adding any list items, It won't look any different so it is expected behaviour, thanks for checking.

app/views/handlers/list_collector_content.py
petechd: But then list collector inherits from Question class `handle_post()` which expects `form.data` which we don't have/send in `Contents` block so it breaks there and at this point it is probably better to keep overriding `handle_post`?

app/views/contexts/hub_context.py
```diff
@@ -157,6 +157,60 @@ def _get_row_for_section(
...
+                        section_title,
+                        CompletionStatus.NOT_STARTED,
+                        self.get_section_url(
+                            section_id,
+                            repeating_blocks_list_item_id,
+                            CompletionStatus.NOT_STARTED,
+                        ),
+                        row_id,
+                    )
+
```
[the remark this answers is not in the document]
petechd: That was needed before we introduced that method checking repeating blocks statuses, not needed now so removed!

app/questionnaire/router.py
```diff
@@ -518,15 +518,27 @@ def get_enabled_section_keys(
...
+                    yield (section_id, list_item_id)
+
+            elif (
+                repeating_blocks_list := self._schema.get_repeating_blocks_list_for_section(
+                    section_id
+                )
+            ) and (
+                self._schema.get_list_collector_type_for_section(section_id)
+                == "ListCollectorContent"
+            ):
```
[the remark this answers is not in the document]
petechd: Cause we only yield this section id, list item id keys for repeating if the parent list collector is of Content type (which doesn't track repeating blocks progress otherwise).

app/views/contexts/summary/list_collector_content_block.py
```diff
@@ -0,0 +1,57 @@
...
+            section_id=self._location.section_id,
+            has_repeating_blocks=bool(list_collector_block.get("repeating_blocks")),
+            return_to="section-summary",
+            content_definition=ImmutableDict({}),
+        )
+
+        return {
+            "title": rendered_summary["title"],
+            "type": rendered_summary["type"],
+            "empty_list_text": None,
```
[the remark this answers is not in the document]
petechd: We reverted this so it uses default empty list text, thanks!

app/views/handlers/list_collector_content.py
```diff
@@ -0,0 +1,82 @@
+from functools import cached_property
+
+from flask import url_for
+
+from app.views.contexts import ListContext
+from app.views.handlers.question import Question
+
+
+class ListCollectorContent(Question):
```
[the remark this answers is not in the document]
petechd: Inheriting from list collector now so there is less individual methods in the new class.

app/questionnaire/router.py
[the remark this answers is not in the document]
MebinAbraham: I'm still not following, which is content special?. It shouldn't be different to normal list collector bar the display?.

app/views/handlers/list_collector_content.py
```diff
@@ -0,0 +1,82 @@
...
+                summary_definition=self.rendered_block["summary"],
+                content_definition=self.rendered_block["content"],
+                section_id=self.current_location.section_id,
+                for_list=self.rendered_block["for_list"],
+                return_to=self._return_to,
+                has_repeating_blocks=bool(self.repeating_block_ids),
+            ),
+        }
+
+    def get_next_location_url(self):
```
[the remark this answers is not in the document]
MebinAbraham: If you make `_get_answer_action` graceful, you can just call the inherited `get_next_location_url` without needed to override here.

app/views/handlers/list_collector_content.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def _is_list_collector_content_complete(self):
+        list_name = self._schema.get_repeating_blocks_list_name_for_section(
+            self.current_location.section_id
+        )
+        return not self.get_first_incomplete_repeating_block_location(
+            repeating_block_ids=self.repeating_block_ids,
+            section_id=self.current_location.section_id,
+            list_name=list_name,
+        )
```
MebinAbraham, line 70: This is duplicating the parent `_is_list_collector_complete` method.

app/views/handlers/list_collector_content.py
MebinAbraham: I think there is a higher risk in these diverging. We don't need to make form conditional, you just need to use the power of inheritance to separate out the common methods. The current code is also inconsistent with list collector which again is the risk we are trying to avoid. List collector only does any of these if the collector is complete. Here is does all but `_update_section_completeness` if collector isn't complete. The current code already doesn't revaluate the path.

It could be something like:
```py
        if self._is_list_collector_complete():
                section_id=self._current_location.section_id,
            )
            return super(Question, self).handle_post()
```

Calling Questions super does the trick but can be prone to error if we add an inheritance in the middle of it, so it can also be done by you guy making the content of handle post of block.py a common method you can call here. See what the devs prefer.

This will break a test, but the likely issue lies somewhere else and not here as the logic for handle post bar the form handling should be the same between the normal and content list collectors.

app/views/handlers/question.py
```diff
@@ -79,6 +79,10 @@ def rendered_block(self):
         )
 
         self._set_page_title(page_title)
+
+        if transformed_block["type"] == "ListCollectorContent":
+            return transformed_block
```
MebinAbraham, line 84: Since this is `question.py`, just I suggest checking if `question` exists, if not return the block. No need to couple it to the type.

templates/listcollectorcontent.html
```diff
@@ -0,0 +1 @@
+{% include 'question.html' %}
```
MebinAbraham, line 1: The logic in question was updated to handle content. Just call `list-content.html` here?.

templates/partials/block.html
```diff
@@ -2,8 +2,11 @@
...
-  {% if 'question' in block %}
-      {% set question = block['question'] %}
-      {% include 'partials/question.html' %}
-  {% endif %}
+{%-  if block['question'] -%}
+  {% set question = block['question'] %}
+  {% include 'partials/question.html' %}
+{%- elif content["list"] -%}
+  {% include 'partials/list-content.html' %}
+{%-  endif -%}
```
MebinAbraham, line 10: Based on the above, revert this?

app/questionnaire/router.py
```diff
@@ -518,15 +518,32 @@ def get_enabled_section_keys(
...
+                repeating_blocks_list := self._schema.get_repeating_blocks_list_name_for_section(
+                    section_id
+                )
+            ) and (
+                # First list collector must be present at this point
+                self._schema.get_first_list_collector_for_section(section_id).get(  # type: ignore
+                    "type"
+                )
+                == "ListCollectorContent"
+            ):
```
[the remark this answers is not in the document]
MebinAbraham: If I'm not mistaken, I have a feeling this might be a misuse to fix the issue mentioned in https://github.com/ONSdigital/eq-questionnaire-runner/pull/1156#discussion_r1263756040.

app/questionnaire/router.py
```diff
@@ -518,15 +518,32 @@ def get_enabled_section_keys(
...
                 for list_item_id in self._list_store[repeating_list]:
-                    section_key: ProgressKeyType = (section_id, list_item_id)
-                    yield section_key
+                    yield section_id, list_item_id
+
+            elif (
+                repeating_blocks_list := self._schema.get_repeating_blocks_list_name_for_section(
+                    section_id
+                )
```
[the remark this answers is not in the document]
petechd: Yes, this was missed when we first introduced repeating blocks. Something that was supposed to be added earlier.

app/views/contexts/list_context.py
```diff
@@ -38,12 +40,17 @@ def __call__(
...
+        list_context = {
             "list": {
                 "list_items": list_items,
                 "editable": any([edit_block_id, remove_block_id]),
-            }
+            },
+            "content": content_definition,
         }
+        if content_definition:
+            list_context["content"] = content_definition
```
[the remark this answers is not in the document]
petechd: Ok, removed.

app/views/contexts/summary/list_collector_content_block.py
```diff
@@ -0,0 +1,54 @@
...
+        )
+
+        return {
+            "title": rendered_summary["title"],
+            "type": rendered_summary["type"],
+            "empty_list_text": rendered_summary.get("empty_list_text"),
+            "list_name": rendered_summary["for_list"],
+            "related_answers": related_answers,
+            **list_summary_context,
+        }
```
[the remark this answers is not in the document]
petechd: I am not sure here, these are the two options you get, with and without label:

Screenshot 2023-07-17 at 10 56 09
Screenshot 2023-07-17 at 10 57 05

Which one should we use?

app/questionnaire/questionnaire_schema.py
```diff
@@ -1048,7 +1072,7 @@ def _block_for_answer(self, answer_id: str) -> ImmutableDict | None:
 
         if (
             parent_block
-            and parent_block["type"] == "ListCollector"
+            and is_list_collector_block_editable(parent_block)
```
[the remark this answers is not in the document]
petechd: There are no question/answers in `ListCollectorContent` hence we only check for `ListCollector` here, since it is used to check `_block_for_answer`.

app/questionnaire/router.py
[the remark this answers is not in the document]
petechd: Removed now, thanks! It wasn't needed.

templates/partials/block.html
petechd: Done!

templates/listcollectorcontent.html
petechd: Upated!

app/questionnaire/router.py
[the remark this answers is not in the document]
berroar: Removed the whole check now as wasn't required given the other changes 👍

app/views/contexts/summary/list_collector_content_block.py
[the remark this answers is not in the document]
MebinAbraham: Pass the `item_label` if we have one. It is optional.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
MebinAbraham: Ya but the answer might be a repeating blocks answer. In which case, this wouldn't work.

app/views/contexts/summary/group.py
```diff
@@ -1,15 +1,19 @@
...
 from app.survey_config.link import Link
 from app.views.contexts.summary.block import Block
 from app.views.contexts.summary.calculated_summary_block import CalculatedSummaryBlock
 from app.views.contexts.summary.list_collector_block import ListCollectorBlock
+from app.views.contexts.summary.list_collector_content_block import (
+    ListCollectorContentBlock,
+)
 
 
 class Group:
```
katie-gardner, line 19: As Mebin commented on the looping pr, we've now got a gap where a CalculatedSummary of repeating questions from a ListCollectorContent block won't work, I think it just needs a check similar to below for which class to initialise the `list_collector_block` with on line 103, and the `get_repeating_block_related_answer_blocks` to be on the base block class

However as this needs testing and probably another schema, maybe it should be separated out onto another card?

katie-gardner (changes requested):
Few extra minor comments from recent changes, but still looks good 👍 

If complete both sections, go back to section 1, add a new list item and then return to the Hub, section 2 is still marked as complete, despite now having incomplete list items, and it remains complete even when I start adding answers for them

Appreciate the section 1 list collector was just to make testing this content block possible, so this might not be an issue that needs fixing but flagging it just to double check?

MebinAbraham:
> If complete both sections, go back to section 1, add a new list item and then return to the Hub, section 2 is still marked as complete, despite now having incomplete list items, and it remains complete even when I start adding answers for them
> 
> Appreciate the section 1 list collector was just to make testing this content block possible, so this might not be an issue that needs fixing but flagging it just to double check?

I could be used that way, hence this is a gap that we would need to fix. We have logic to revaluate progress of section that depend on a "list" source but not parts of a list i.e repeating blocks. That was just missed when we implemented it.

app/views/handlers/list_action.py
```diff
@@ -67,10 +67,13 @@ def get_next_location_url(self):
...
-            block_id=self.parent_location.block_id,
-            section_id=self.parent_location.section_id,
-            list_item_id=self.parent_location.list_item_id,
+        if (
+            self.router.is_block_complete(
+                block_id=self.parent_location.block_id,
+                section_id=self.parent_location.section_id,
+                list_item_id=self.parent_location.list_item_id,
+            )
+            and not self.parent_block["type"] == "ListCollectorContent"
```
MebinAbraham, line 76: Why are we checking the type here?.

app/views/handlers/list_collector_content.py
MebinAbraham: This is still outstanding. There is still code that shouldn't exist now.

app/views/handlers/list_repeating_question.py
```diff
@@ -19,6 +19,18 @@ def get_previous_location_url(self):
...
+            self.parent_block["id"]
+        )
+        if not edit_block:
+            block_id = self._request_args.get("previous")
+            return self._get_location_url(
+                block_id=block_id,
+                return_to=self._return_to,
+                return_to_answer_id=self._return_to_answer_id,
+                return_to_block_id=self._return_to_block_id,
+            )
```
MebinAbraham, line 32: How come this is at this level?. This will skip the other logic routes if we don't have an edit block which is incorrect. This should be at the end. We also don't have a `previous` arg?. If we can't edit at the end, we should return to parent.

app/views/handlers/question.py
```diff
@@ -154,21 +158,22 @@ def _is_list_just_primary(self, list_items, list_name):
...
+                submitted_answer = self.form.data[answer["id"]]
+
+                for option in answer.get("options", {}):
+                    action = option.get("action")
+
+                    if action and (
+                        option["value"] == submitted_answer
+                        or option["value"] in submitted_answer
+                    ):
+                        return action
```
MebinAbraham, line 176: We should avoid this deep nesting but investing the logic and existing early. Do you even need this change once the change to handle post are done to call the parent of Question?

MebinAbraham:
> If complete both sections, go back to section 1, add a new list item and then return to the Hub, section 2 is still marked as complete, despite now having incomplete list items, and it remains complete even when I start adding answers for them

Should be fixed once we do https://github.com/ONSdigital/eq-questionnaire-runner/pull/1156#discussion_r1273893928

app/questionnaire/router.py
```diff
@@ -518,15 +518,27 @@ def get_enabled_section_keys(
...
+            ) and (
+                self._schema.get_list_collector_type_for_section(section_id)
+                == "ListCollectorContent"
+            ):
+                for list_item_id in self._list_store[repeating_blocks_list]:
+                    yield (section_id, list_item_id)
+
+                yield section_id, None
```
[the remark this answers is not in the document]
MebinAbraham: This was reverted. How come?.

tests/functional/spec/list_collector_content.spec.js
```diff
@@ -0,0 +1,129 @@
...
+      await fillInListCollectorSection();
+      await expect(await browser.getUrl()).to.contain(HubPage.url());
+      await expect(await $(HubPage.summaryRowState("section-list-collector-contents")).getText()).to.contain("Not started");
+      await $(HubPage.submit()).click();
+      await expect(await browser.getUrl()).to.contain(ListCollectorContentPage.url());
+    });
+    it("When I get to the list collector content page, Then the relevant content is displayed.", async () => {
+      await fillInListCollectorSection();
+      await $(HubPage.submit()).click();
+      await expect(await $(await $("body")).getHTML()).to.contain(
```
MebinAbraham, line 26: Do we not have selectors for content pages already?

tests/integration/questionnaire/test_questionnaire_list_collector_content.py
MebinAbraham: I don't we need gaps after every line, it looks very verbose. My preference is gaps after each logical chunk. Thoughts?

templates/partials/list-content.html
MebinAbraham: We should add more properties that contents support. It wasn't using it yet no tests failed. Even though it has now changes, no tests verify it.

tests/integration/questionnaire/test_questionnaire_list_collector_content.py
katie-gardner: I'd prefer that too

app/views/contexts/summary/group.py
katie-gardner: Its on `ListCollectorBlock` but I think it needs to be on `ListCollectorBaseBlock` so that `ListCollectorContent` can also make use of it

app/questionnaire/questionnaire_schema.py
```diff
@@ -1423,3 +1431,8 @@ def get_calculation_block_ids_for_grand_calculated_summary(
         calculation_block=grand_calculated_summary_block,
         source_type="calculated_summary",
     )
+
+
+def is_list_collector_block_editable(block: Mapping) -> bool | None:
+    # Type will always exist at this point
+    return block["type"] == "ListCollector"  # type: ignore
```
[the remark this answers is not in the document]
petechd: I went with `bool()` in the end, I heard before we should avoid casting types using type hints and `:`.

app/views/contexts/summary/list_collector_block.py
```diff
@@ -151,18 +95,6 @@ def get_repeating_block_related_answer_blocks(
                 blocks.extend(answer_blocks)
```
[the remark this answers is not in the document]
petechd: I am not sure about this one, this method is only used in context of generating display of calculated summary and grand calculated summary, so we should only move it to base class if content list collector supports calculated summary?

app/views/contexts/summary/list_collector_content_block.py
```diff
@@ -0,0 +1,52 @@
...
+        rendered_summary = self._placeholder_renderer.render(
+            data_to_render=summary, list_item_id=self._location.list_item_id
+        )
+
+        if list_collector_blocks_on_path:
+            repeating_blocks = list_collector_block.get("repeating_blocks", [])
+            related_answers = self._get_related_answer_blocks_by_list_item_id(
+                list_model=current_list, repeating_blocks=repeating_blocks
+            )
+            item_label = self._schema.get_item_label(
```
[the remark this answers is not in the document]
petechd: I am not sure here as this will mean we won't need the child classes and everything will be done in the base class.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
MebinAbraham: Yes, we should. If we know a type will always be x, it can be type ignored, without casting to appease mypy. We should update the type hiniting doc if it is not there.

app/views/contexts/summary/list_collector_block.py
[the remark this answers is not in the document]
katie-gardner: On my[ supplementary data branch ](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1173) I've got a schema which uses repeating blocks to ask questions about a supplementary list and then it has a calculated summary of those answers, so I think we'll want the content block to support calculated/grand calculated summaries too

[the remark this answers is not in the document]
MebinAbraham: Just for additional info/context, list collector content block needs to support everything that a list collector can do bar add, edit and remove and that it is a content block instead of a question. Other than that, it should behave identically. Hence, the importance to keep code shared as much as possible.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
katie-gardner: ah my bad, the doc doesn't explicitly say don't cast, but agree with that being a sensible addition, and ignoring this occurance 👍

app/views/handlers/question.py
petechd: This block question check is still needed, so just returning early now and nesting is reduced.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
petechd: Added `ListCollectorContent` as possible type and using one single const for all checks to reduce duplication.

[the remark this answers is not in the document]
petechd: Added a check here using a constant now that we reuse in other places too.

app/views/handlers/list_collector_content.py
petechd: I think I reduced it as much as I could now, without breaking it.

tests/functional/spec/list_collector_content.spec.js
petechd: There is `definitionContent()` selector but we cannot use it here. I narrowed it by using the id, so `expect(await $("#main-content").getText())`, can possibly narrow it down more by doing `expect(await $("#main-content > div > div > p").getText())` but do we want to?

app/questionnaire/router.py
```diff
@@ -518,15 +518,13 @@ def get_enabled_section_keys(
...
+                section_id
+            ):
                 for list_item_id in self._list_store[repeating_list]:
-                    section_key: ProgressKeyType = (section_id, list_item_id)
-                    yield section_key
+                    yield section_id, list_item_id
             else:
-                section_key = (section_id, None)
-                yield section_key
+                yield section_id, None
```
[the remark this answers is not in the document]
petechd: I did some heavy refactoring after @katie-gardner comment and possibly didn't realise this now allows submitting incomplete questionnaires.

app/views/handlers/list_action.py
petechd: If repeating block is complete, parent list collector content block is also complete and this will try to get you to next location url (if this check is removed first "if" will run as all conditions will be valid) **but** we want to get to the second "return" and go to parent location url which will be the list collector content block. So this was added just to have the correct journey.

app/views/handlers/list_repeating_question.py
petechd: Some confusion here, my mistake, moved it to the bottom now and using parent as previous, also removed type ignore and fixed some test that was incorrect.

tests/integration/questionnaire/test_questionnaire_list_collector_content.py
petechd: So I grouped the journeys together, assertions together and method calls like "self.launchSurvey()" separate.

app/views/contexts/summary/group.py
petechd: Added now so it's also `ListCollectorBlock` ones, thanks. `get_repeating_block_related_answer_blocks` moved to base block now.

app/views/contexts/summary/list_collector_block.py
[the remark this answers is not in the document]
petechd: Ok, added now. Thanks.

MebinAbraham (changes requested):
There are quite a few places that assumes that add/edit/remove exist. You would need to go through the code base and find occurrences we use those blocks and handle them not existing.

petechd:
> There are quite a few places that assumes that add/edit/remove exist. You would need to go through the code base and find occurrences we use those blocks and handle them not existing.

From what I've investigated it seems like these (add, remove, edit) blocks are mentioned (assumed to exist) a lot in questionnaire schema class methods but all of it is .get() extracted so seems like we have safe fallbacks there and don't process blocks that don't exist and don't return anything worst case, another example is questionnaire store updater module where we do .get() them or otherwise don't update dependencies. Whenever it is called in other places (outside questionnaire schema module) we check the type is ListCollectorBlock (eg. in convert payload, variants.py), so we do have safety measures in place. Wouldn't these extensive unit and integration tests break? Maybe I should add some extra scenarios/journeys to integration tests? Do you know any places where it does assume these blocks exist and break?

app/views/contexts/summary/list_collector_content_block.py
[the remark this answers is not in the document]
katie-gardner: Yeah now that you've reduced got shared methods I think this won't be needed, so all good 👍

app/views/handlers/list_collector_content.py
```diff
@@ -9,14 +9,12 @@ def _get_additional_view_context(self) -> Any:
         return self.rendered_block["content"]
 
     def handle_post(self) -> None:
-        self._set_started_at_metadata()
-        self.questionnaire_store_updater.add_completed_location()
         if self._is_list_collector_complete():
-            self.questionnaire_store_updater.add_completed_location()
```
katie-gardner: just for my understanding - why isn't this needed here?

katie-gardner (commented):
My comments addressed and basically all good from me 👍 
Just one issue - which I think is where I commented my confusion, I'm now seeing repeating block sections being in progress on the dump despite having finished them, so the ticks aren't showing up despite them being completed. The other issue with progress of the section not updating I understand will be done in another card? And I tried adding a calculated summary which all rendered properly, and I the minor progress issue I think would be fixed by the overall one

I also tried to find places that assume add edit exist but agree it does look like most are guarded and it didn't break when I dropped the list collector - only one I can find is `get_add_block_for_list_collector` - think that one needs fixing (and that is widely used so should fix a bunch)

Once main is merged in and you have cards like using supplementary data non list items ill try it with my card to double check

MebinAbraham:
> I'm now seeing repeating block sections being in progress on the dump despite having finished them, so the ticks aren't showing up despite them being completed. 

Is this if 2 section have list collectors with repeating blocks for the same list but you edit one?. Can you elobrate?.

> The other issue with progress of the section not updating I understand will be done in another card?

Can you celebrate?. Not sure which card we are referring to or issue.

katie-gardner:
> > I'm now seeing repeating block sections being in progress on the dump despite having finished them, so the ticks aren't showing up despite them being completed.
> 
> Is this if 2 section have list collectors with repeating blocks for the same list but you edit one?. Can you elobrate?.
> 
> > The other issue with progress of the section not updating I understand will be done in another card?
> 
> Can you celebrate?. Not sure which card we are referring to or issue.

Yep, if section 1 has a list collector, section 2 has lit collector content
Complete both sections
Add a new list item to section 1

Section 2 is still complete but the repeating blocks are not, and in fact, all of them revert to in progress, not just those for the new list item
I can press save and continue from the list collector content, and then it will take me to the repeating blocks for the new list item, I do those, and then that list item has a tick, but the other two don't, and are stuck as in progress despite being complete

I might have mis-remembered, I thought we were going to have a follow up card that handled the progress of the section  with a list collector content block not updating when you add a new list item, but I think it was "Questionnaire should not be marked as complete until repeating blocks are completed" that I was thinking of, which might not cover all of this case?

MebinAbraham:
> Yep, if section 1 has a list collector, section 2 has lit collector content Complete both sections Add a new list item to section 1
 
> Section 2 is still complete but the repeating blocks are not, and in fact, all of them revert to in progress, not just those for the new list item I can press save and continue from the list collector content, and then it will take me to the repeating blocks for the new list item, I do those, and then that list item has a tick, but the other two don't, and are stuck as in progress despite being complete

If section 2 was complete to begin with, then when I add a new item in section 1, I can see section 2 is still complete but goes to summaries with No answers provided for the new item.

I think we should address that here given the committed schema demos this behaviour. Should be simple enough as we just need to get all list collectors that depend on the list instead of the first list collector.

@petechd

MebinAbraham:
> > There are quite a few places that assumes that add/edit/remove exist. You would need to go through the code base and find occurrences we use those blocks and handle them not existing.
> 
> From what I've investigated it seems like these (add, remove, edit) blocks are mentioned (assumed to exist) a lot in questionnaire schema class methods but all of it is .get() extracted so seems like we have safe fallbacks there and don't process blocks that don't exist and don't return anything worst case, another example is questionnaire store updater module where we do .get() them or otherwise don't update dependencies. Whenever it is called in other places (outside questionnaire schema module) we check the type is ListCollectorBlock (eg. in convert payload, variants.py), so we do have safety measures in place. Wouldn't these extensive unit and integration tests break? Maybe I should add some extra scenarios/journeys to integration tests? Do you know any places where it does assume these blocks exist and break?

- `_update_answer_dependencies_for_list_source` would error as it is expecting add block to exist. Other than that, it looks okay.

tests/functional/spec/list_collector_content.spec.js
MebinAbraham: We tackle it separetly.

app/questionnaire/router.py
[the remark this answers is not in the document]
MebinAbraham: There is a gap here but we will tackle it in a different PR

app/views/contexts/summary/group.py
```diff
@@ -170,7 +184,12 @@ def _build_blocks_and_links(
                     section_id=section["id"],  # type: ignore
                     list_name=block["for_list"],
                 ):
-                    list_collector_block = ListCollectorBlock(
+                    list_collector_block_class = (
+                        ListCollectorBlock
+                        if is_list_collector_block_editable(block)
+                        else ListCollectorContentBlock
+                    )
+                    list_collector_block = list_collector_block_class(  # type: ignore
```
MebinAbraham, line 192: Why does this type exist?

app/views/contexts/summary/group.py
```diff
@@ -181,14 +200,19 @@ def _build_blocks_and_links(
                         location=location,
                         language=language,
                         return_to=return_to,
+                        return_to_block_id=return_to_block_id,
                     )
 
-                    list_summary_element = list_collector_block.list_summary_element(
+                    # Type ignore: base block check not required here as both child blocks have list_summary_element
+                    list_summary_element = list_collector_block.list_summary_element(  # type: ignore
```
MebinAbraham, line 207: This is valid, you need to fix the typing for the above lines when initialisation the class to fix this.

app/views/handlers/list_collector_content.py
```diff
@@ -0,0 +1,20 @@
...
+    def handle_post(self) -> None:
+        if self._is_list_collector_complete():
+            self._routing_path = self.router.routing_path(
+                section_id=self._current_location.section_id,
+                list_item_id=self._current_location.list_item_id,
+            )
+            return super(Question, self).handle_post()
+        self.questionnaire_store_updater.add_completed_location()
+        self.questionnaire_store_updater.update_progress_for_dependent_sections()
+        self.questionnaire_store_updater.save()
```
MebinAbraham, line 20: To fix x, this will need 
```
        if self._is_list_collector_complete():
                section_id=self._current_location.section_id,
            )
            super(Question, self).handle_post()
```

then in list repeating question you want to do the same logic as the list add question:
Create a helper method and reuse it in add and repeating question block.
```
        # Clear the answer from the confirmation question on the list collector question
        answer_ids_to_remove = self._schema.get_answer_ids_for_block(
            # Type ignore: for parent_location block_id is not none
            self.parent_location.block_id  # type: ignore
        )
        self.questionnaire_store_updater.remove_answers(answer_ids_to_remove)
        self.questionnaire_store_updater.remove_completed_location(self.parent_location)
```

MebinAbraham (changes requested):

app/views/contexts/summary/group.py
petechd: Removed now, after making some changes.

app/views/contexts/summary/group.py
petechd: Fixed the type, ignore removed.

app/views/handlers/question.py
```diff
@@ -157,7 +161,10 @@ def _is_list_just_primary(self, list_items: list[str], list_name: str) -> bool:
             == self._questionnaire_store.list_store[list_name].primary_person
         )
 
-    def _get_answer_action(self) -> dict | None:
+    def _get_answer_action(self) -> Any:
+        if not self.rendered_block.get("question"):
+            return
```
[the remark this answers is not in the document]
petechd: We use list collectors own `get_next_location_url()` method as content list collector inherits that, and then that supers `Question` `get_next_location_url()`.

app/views/contexts/summary/list_collector_content_block.py
```diff
@@ -0,0 +1,52 @@
...
+                self._section["id"], current_list.name
+            )
+
+        list_summary_context = self.list_context(
+            list_collector_block["summary"],
+            for_list=list_collector_block["for_list"],
+            section_id=self._location.section_id,
+            has_repeating_blocks=bool(list_collector_block.get("repeating_blocks")),
+            return_to="section-summary",
```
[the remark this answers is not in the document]
petechd: Not hardcoded anymore.

app/views/handlers/list_repeating_question.py
```diff
@@ -50,10 +49,18 @@ def get_previous_location_url(self) -> str:
...
             self.parent_block["id"]
         )
+        if not edit_block:
+            block_id = self.parent_block["id"]
+            return self._get_location_url(
+                block_id=block_id,
+                return_to=self._return_to,
+                return_to_answer_id=self._return_to_answer_id,
+                return_to_block_id=self._return_to_block_id,
+            )
```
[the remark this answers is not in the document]
petechd: Added, thanks.

app/questionnaire/router.py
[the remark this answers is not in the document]
petechd: This one is removed now.

app/questionnaire/router.py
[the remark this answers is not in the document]
petechd: From what I remember, this was removed cause we refactored the code to use some common method that was introduced in Katie's PR.

petechd:
Logic in list collector content handler is now aligned with list collector. As discussed with @MebinAbraham  the existing issue affecting both types of list collectors (section progress key not being added to progress store when no question block prior to list collector) needs to be addressed separately.

app/views/contexts/summary/group.py
```diff
@@ -107,7 +113,18 @@ def _build_blocks_and_links(
                 if parent_list_collector_block_id not in routing_path_block_ids:
                     continue
 
-                list_collector_block = ListCollectorBlock(
+                list_collector_block_class: Type[
+                    ListCollectorBlock | ListCollectorContentBlock
```
katie-gardner, line 117: type hint not needed - mypy will infer it from the classes

app/views/handlers/list_collector.py
```diff
@@ -52,8 +52,7 @@ def get_next_location_url(self) -> str:
 
         return super().get_next_location_url()
 
-    def get_context(self) -> dict[str, dict]:
-        question_context = super().get_context()
+    def _get_list_context(self) -> dict:
```
katie-gardner, line 55: I think the type hint is still `dict[str, dict]`?

app/views/handlers/list_collector_content.py
```diff
@@ -0,0 +1,17 @@
+from typing import Any
+
+from app.views.handlers.list_collector import ListCollector
+from app.views.handlers.question import Question
+
+
+class ListCollectorContent(ListCollector):
+    def _get_additional_view_context(self) -> Any:
```
katie-gardner, line 8: missing proper return type

app/views/handlers/list_collector.py
```diff
@@ -65,18 +64,21 @@ def get_context(self) -> dict[str, dict]:
...
+            for_list=self.list_name,
+            edit_block_id=self.rendered_block.get("edit_block", {}).get("id"),
+            remove_block_id=self.rendered_block.get("remove_block", {}).get("id"),
+            return_to=self._return_to,
+            section_id=self.current_location.section_id,
+            has_repeating_blocks=bool(self.repeating_block_ids),
+        )
+
+    def _get_additional_view_context(self) -> dict:
+        return super().get_context()
```
katie-gardner, line 78: I'd find it helpful to have a docstring on this one, mentioning what it does in the case of a list collector - initially I wasn't sure why this was added instead of using the super method in `get_context` but now I see its so it can be overwritten in list collector content

app/views/contexts/summary/list_collector_base_block.py
```diff
@@ -0,0 +1,200 @@
...
+            self._schema,
+            self._answer_store,
+            self._list_store,
+            self._progress_store,
+            self._metadata,
+            self._response_metadata,
+            self._supplementary_data_store,
+        )
+
+    def _list_collector_block_on_path(self, for_list: str) -> list:
```
katie-gardner, line 73: I think this should be returning `list[ImmutableDict]`

app/views/contexts/summary/list_collector_base_block.py
```diff
@@ -0,0 +1,200 @@
...
+            if list_collector_block["id"] in self._routing_path_block_ids
+        ]
+
+    def _list_collector_block(
+        self, for_list: str, list_collector_blocks_on_path: list
+    ) -> ImmutableDict:
+        list_collector_blocks = list(
+            self._schema.get_list_collectors_for_list(self._section, for_list=for_list)
+        )
+        # Type ignore: self._schema.get_list_collectors_for_list can return None
```
katie-gardner, line 90: Appreciate this was already there, but the type ignore has an incorrect comment, `list_collector_blocks` will be a list of Immutable dictionaries and the reason for the type ignore is that `list_collector_blocks_on_path` is of type `list` - so if the type of that is just corrected this ignore can be removed

katie-gardner (changes requested):
Working subject to the issue that will be resolved on another card 👍 - few minor thoughts added

app/questionnaire/questionnaire_schema.py
```diff
@@ -701,7 +704,10 @@ def get_driving_question_for_list(
 
     def get_remove_block_id_for_list(self, list_name: str) -> str | None:
         for block in self.get_blocks():
-            if block["type"] == "ListCollector" and block["for_list"] == list_name:
+            if (
+                is_list_collector_block_editable(block)
```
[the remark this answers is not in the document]
petechd: Yeah we don't check the block["type"] anymore and re-use existing method, is there a double check here?

app/views/contexts/summary/group.py
petechd: "ListCollectorBaseBlock" has no attribute "list_summary_element" when removed as mypy thinks we are using parent class and the method is not in that class.

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
katie-gardner: No.. 🤦‍♀️ my bad 😆 must have misread this

app/views/handlers/list_collector_content.py
petechd: This might return any as content key might not always be there in rendered block but we got around it by accessing using .get() method.

petechd:
@katie-gardner all addressed, except two where we are not sure if the changes are necessary or keeping it as it is is necessary (but we left comments). Thanks.

app/views/contexts/calculated_summary_context.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         metadata: MetadataProxy | None,
         response_metadata: MutableMapping,
         routing_path: RoutingPath,
-        current_location: LocationType,
+        current_location: Location,
```
MebinAbraham, line 44: Should this have changed?. Couple of others like this, bad merge?. These were changed in the type hinting PR.

app/views/handlers/question.py
```diff
@@ -82,6 +82,10 @@ def rendered_block(self) -> dict:
         )
 
         self._set_page_title(page_title)
+
+        if not transformed_block.get("question"):
+            return transformed_block
```
MebinAbraham, line 87: Add a comment why we do this, as it might be confusing why question.py wouldn't have question. Ideally it could be to block.py

app/views/handlers/question.py
```diff
@@ -162,6 +166,9 @@ def _is_list_just_primary(self, list_items: list[str], list_name: str) -> bool:
         )
 
     def _get_answer_action(self) -> dict | None:
+        if not self.rendered_block.get("question"):
```
MebinAbraham, line 169: Same here, a comment would help.

templates/partials/list-content.html
```diff
@@ -0,0 +1,19 @@
+
+{% set title = content.title %}
+{% set contents = content.contents %}
+
+{% call onsQuestion({
+  "id": question.id,
+  "title": title,
```
MebinAbraham, line 7: Why do we use `onsQuestion`?. question won't exist, hence the id will be not be defined.

tests/functional/spec/list_collector_content.spec.js
```diff
@@ -0,0 +1,139 @@
...
+      await $(ResponsiblePartyQuestionPage.yes()).click();
+      await $(ResponsiblePartyQuestionPage.submit()).click();
+      await expect(await browser.getUrl()).to.contain(ListCollectorContentPage.url());
+    });
+    it("When I get to the list collector content page, Then the relevant content is displayed.", async () => {
+      await fillInListCollectorSection();
+      await $(HubPage.submit()).click();
+      await $(ResponsiblePartyQuestionPage.yes()).click();
+      await $(ResponsiblePartyQuestionPage.submit()).click();
+      await expect(await $("#main-content").getText()).to.contain(
```
MebinAbraham, line 31: Do we not have selectors?

tests/functional/generate_pages.py
```diff
@@ -854,6 +854,17 @@ def process_block(
                 page_filename=f'{repeating_block["id"]}-repeating-block.page.js',
             )
 
+    if block["type"] in {"ListCollector", "ListCollectorContent"}:
+        for repeating_block in block.get("repeating_blocks", []):
```
MebinAbraham, line 858: This is duplicating the above, list collector is already handled. Handle repeating block for both, then handle add edit remove for list collector only

templates/listcollectorcontent.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'layouts/_questionnaire.html' %}
+{% from "components/question/_macro.njk" import onsQuestion %}
+{% import 'macros/helpers.html' as helpers %}
+
+{% set save_on_signout = true %}
```
MebinAbraham, line 5: You need to set `{% set continue_button_text = _("Continue") %}`

At the moment the button text is "Save and continue" which is not what we wabt. Test should be updated to verify.

templates/partials/list-content.html
petechd: No need, removed.

templates/partials/list-content.html
```diff
@@ -0,0 +1,19 @@
...
+  {%- if content.list and content.list.list_items -%}
+    {% set list = content.list %}
+    <div class="ons-u-mb-m">
+      {% include 'partials/summary/list-summary.html' %}
+    </div>
+  {%- endif -%}
+
+  {% include 'partials/contents.html' %}
+
+{% endcall %}
```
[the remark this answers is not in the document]
petechd: Only used once so removing the extra partial.

templates/partials/list-content.html
```diff
@@ -0,0 +1,19 @@
...
+  "title": title,
+}) %}
+  {%- if content.list and content.list.list_items -%}
+    {% set list = content.list %}
+    <div class="ons-u-mb-m">
+      {% include 'partials/summary/list-summary.html' %}
+    </div>
+  {%- endif -%}
+
+  {% include 'partials/contents.html' %}
```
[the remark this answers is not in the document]
petechd: We've added guidance, definition. Description was already there. Added test assertions as well.

app/views/handlers/question.py
```diff
@@ -24,7 +24,7 @@ def _has_redirect_to_list_add_action(answer_action: Mapping | None) -> bool:
 
     @cached_property
     def form(self) -> QuestionnaireForm:
-        question_json = self.rendered_block["question"]
+        question_json = self.rendered_block.get("question", {})
```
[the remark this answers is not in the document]
petechd: No, we cannot revert this cause when we submit LCC page it still calls this method, similar to normal LC.

app/views/contexts/calculated_summary_context.py
petechd: Yeah this is from that big merge, fixed now, thanks!

MebinAbraham (approved):

rmccar (approved):

app/questionnaire/questionnaire_schema.py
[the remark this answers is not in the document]
katie-gardner: I think this comment got re-opened by someone so I'm guessing thats to remove `bool(...)` and type ignore instead? I've no strong feeling whether its updated or this just resolved

PR #1149, opened 2023-06-28
Remove update_section_completeness call in list actions

petechd:
### What is the context of this PR?
This removes `update_section_completeness` call that adds list items progress for child blocks which wasn't done in the past and wasn't required for looping 3.5. The only time childs are used is for repeating blocks.

### How to review
Check that all tests still pass.

### Checklist
* [ ] New static content marked up for translation
* [ ] Newly defined schema content included in eq-translations repo

PR #1157, opened 2023-07-05
Looping 4 - Calculated Summary with repeating blocks

katie-gardner:
This PR adds the functionality that allows calculated summaries to use repeating block answers that resolve to a list of answers. 

Calculated summaries now support consisting of a single repeating answer, or both repeating and static answers where each repeating answer resolves to a list of answers for each list item. By repeating answer, this means either dynamic, or an answer for a repeating block.

This required a minor change to validator - [corresponding PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/206)

Note that when editing a list item from the section summary that is only used by the second calculated summary in the test schema - this will take you to the first calculated summary and then the second. Jumping to the second calculated summary will be covered by the card that takes you to the next incomplete block, rather than just stepping through the section until it is complete.

### Technical details
`placeholder_render.py` has been modified to suffix any repeating question blocks/answers with their list item id. This is just so the HTML outputted by the jinja templates is valid. To see this working, inspect element on the summary page of a list collector with repeating blocks and note that the div ids are the same for each list item id. With this change they should now have their list item id suffix (Just like dynamic answers would)

`questionnaire_schema.py` now stores a list of repeating block answer ids and a map of repeating blocks to their list, so that the value source resolver has the ability to check if an answer is for a repeating block, and if so, where to get the list of answers from

`value_source_resolver.py` now checks if an answer is either dynamic, or for a repeating block and for either case, calls a shared method to get the list of answers for each list item id.

`group.py` now checks if the block being rendered is a repeating question. It needs to do this before checking if the block is on the path, as list repeating questions won't be on it, its a property of the list collector instead. This will call a new method on `ListCollectorBlock` to return a list of blocks for the question, one copy for each list item id

`list_collector_block.py` Now accepts `return to` and `return_to_block_id` as this is now used outside of section summaries and also by calculated summaries. It has a new method to take a repeating question block and return a list of rendered copies for each list item id. It also renders related answers to ensure that any answer labels using placeholders show correctly.

`list_action.py`, `list_edit_question.py`, `list_repeating_question.py` have had various updates so that on a repeating block, the previous link will take you to the previous block or the edit block, and continuing will take you to the next incomplete repeating block, or back to a calculated/section summary if the block is complete and `return_to` is populated.

Unit tests have been added to cover the changes to value source resolver
integration tests have been added to cover the changes to routing on a list collector with repeating blocks
functional tests to cover any scenarios on the trello ticket that weren't covered by looping 4.5

### How to test functionally
Use the new `test_new_calculated_summary_repeating_blocks.json` schema to add list items with repeating blocks that then feed into two separate calculated summaries, one that also uses a static answer, and one that only has a single repeating answer. 
Check that only the correct answers show up and that routing works as expected when adding/removing/editing list items.

### How to review
Read through the trello card to familiarise with the concept and check that each scenario in the attached scenarios google sheet is working. Check the test coverage is suitable

### Checklist


scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=looping-4-calculated-summary-repeating-blocks
```
katie-gardner, line 2: Reminder to revert before merge

app/views/contexts/summary/list_collector_block.py
```diff
@@ -92,7 +96,9 @@ def list_summary_element(self, summary: Mapping[str, Any]) -> dict[str, Any]:
             remove_block_id = list_collector_block["remove_block"]["id"]
             add_link = self._add_link(summary, list_collector_block)
             repeating_blocks = list_collector_block.get("repeating_blocks", [])
-            related_answers = self._get_related_answers(current_list, repeating_blocks)
+            related_answers = self._get_related_answers(
+                list_model=current_list, repeating_blocks=repeating_blocks
```
petechd, line 100: Non blocker, If list_model is an instance of `ListModel` class maybe we should rename the var in get_related_answers so it is `current_list` too or rename `current_list` to `list_model`, just for clarity?
Especially when we call list store below and var is list_model again:
```python
     list_model = self._list_store[list_name]
```

petechd (commented):
I didn't functionally test it yet, just had a glance. Will continue tomorrow.

app/questionnaire/questionnaire_schema.py
```diff
@@ -778,6 +812,9 @@ def is_answer_in_repeating_section(self, answer_id: str) -> bool | None:
     def is_answer_dynamic(self, answer_id: str) -> bool:
         return answer_id in self._dynamic_answer_ids
 
+    def is_answer_for_repeating_block(self, answer_id: str) -> bool | None:
```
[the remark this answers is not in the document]
katie-gardner: Nope 😄 fixed

app/questionnaire/questionnaire_schema.py
```diff
@@ -821,6 +858,13 @@ def get_add_block_for_list_collector(
             ]
             return add_block
 
+    def get_edit_block_for_list_collector(
+        self, list_collector_id: str
+    ) -> ImmutableDict | None:
+        if list_collector := self.get_block(list_collector_id):
```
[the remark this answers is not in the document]
katie-gardner: Added a type ignore for the list collector block 👍 I've left return as optional though as with the introduction of your ticket the edit block could become optional so saves remembering to do it in a merge?

app/questionnaire/questionnaire_schema.py
```diff
@@ -361,6 +380,17 @@ def _populate_answer_dependencies(self) -> None:
                                 option["detail_answer"], block_id=block["id"]
                             )
 
+    def _update_dependencies_for_repeating_blocks(self, block: ImmutableDict) -> None:
```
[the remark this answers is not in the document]
katie-gardner: the type hint for this would be `ImmutableDict[str, list | dict | str]` which doc suggests we don't include and just leave at top level

app/views/contexts/summary/list_collector_block.py
katie-gardner: Yeah that seems like a good switch 👍 changed to use `list_model` as that matches the other stuff there better I think

app/views/handlers/list_edit_question.py
```diff
@@ -13,6 +13,15 @@ def is_location_valid(self):
             return False
         return True
 
+    def get_next_location_url(self):
+        if self._is_returning_to_section_summary():
+            return self.get_section_summary_url()
+
+        if url := self._get_first_incomplete_repeating_block_url():
+            return url
```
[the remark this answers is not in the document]
katie-gardner: I'd put it in repeating question but I also need this in `ListEditQuestion` because if you are on the edit block for a list collector with incomplete repeating blocks, you also need to jump to the url of the first incomplete one, and both inherit from ListAction so I think thats where it needs to go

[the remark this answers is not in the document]
kylelawsonAND: You might be able to have `ListRepeatingQuestion` inherit `ListEditQuestion` - I originally had it doing this in the Looping 2.5 impl but there was were calls preventing compatibility in `ListAction` `handle_post` which was called by `ListEditQuestion`. 
However this was removed in[ this PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1149). 
This means you could call `ListEditQuestion.handle_post` safely from `ListRepeatingQuestion.handle_post` now. They also share identical `get_next_location_url`.
So this can then be declared only in ListEditQuestion and remove some duplicated code.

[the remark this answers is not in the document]
katie-gardner: yes! that will totally do it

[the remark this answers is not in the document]
katie-gardner: Done 👍  and removed the method as Pete suggests since its back to only being used once, thanks both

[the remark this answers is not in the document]
kylelawsonAND: Looks good to me, consider my comments on this thread resoved 👍

kylelawsonAND (approved):
All looks good to me now after resolving the ListRepeatingQuestion/ListEditQuestion inheritance ✅

app/questionnaire/questionnaire_schema.py
```diff
@@ -82,7 +82,8 @@ def __init__(
         ] = defaultdict(set)
         self._language_code = language_code
         self._questionnaire_json = questionnaire_json
-        self._repeating_block_ids: list[str] = []
+        self._repeating_block_to_list_map: dict[str, str] = {}
```
MebinAbraham, line 85: I think we need to prefix repeating block with `list` otherwise it might get confusing as we have repeating sections, and technically all blocks inside that is a repeating block, but we are specifically talking about repeating blocks for a list. Thoughts?

app/questionnaire/value_source_resolver.py
```diff
@@ -100,36 +100,58 @@ def _resolve_list_item_id_for_value_source(
             else None
         )
 
+    def _resolve_repeating_answers_for_list(
+        self, *, answer_id: str, list_name: str
+    ) -> list[AnswerValueTypes]:
+        """Return the list of answers in answer store that correspond to the given list name and dynamic/repeating answer_id"""
+        answer_values: list[AnswerValueTypes] = []
+        for list_item_id in self.list_store[list_name]:
+            if answer_value := self._get_answer_value(
```
MebinAbraham, line 109: I see the existing code for dynamic was like this, but this wouldn't work for falsey values like `0`.

app/questionnaire/value_source_resolver.py
```diff
@@ -100,36 +100,58 @@ def _resolve_list_item_id_for_value_source(
...
+                answer_id=answer_id,
+                list_item_id=list_item_id,
+                assess_routing_path=False,
```
MebinAbraham, line 112: This shouldn't be set, or set to None?

app/views/contexts/section_summary_context.py
```diff
@@ -181,6 +181,7 @@ def _custom_summary_elements(
                     schema=self._schema,
                     location=self.current_location,
                     language=self._language,
+                    return_to="section-summary",
```
MebinAbraham, line 184: How come we default this here, we might now always have a section summary. We should have no summary, a section summary or a list summary.

app/views/contexts/summary/group.py
```diff
@@ -151,6 +171,7 @@ def _build_blocks_and_links(
                         schema=schema,
                         location=location,
                         language=language,
+                        return_to="section-summary",
```
MebinAbraham, line 174: Same as above.

app/views/contexts/summary/group.py
```diff
@@ -88,6 +88,26 @@ def _build_blocks_and_links(
         blocks = []
 
         for block in group_schema["blocks"]:
+            if block["type"] == "ListRepeatingQuestion":
+                # list repeating questions aren't themselves on the path, the parent list collector is checked in ListCollectorBlock
+                list_collector_block = ListCollectorBlock(
```
MebinAbraham, line 93: > group.py now checks if the block being rendered is a repeating question. It needs to do this before checking if the block is on the path, as list repeating questions won't be on it, its a property of the list collector instead. This will call a new method on ListCollectorBlock to return a list of blocks for the question, one copy for each list item id

I'm not following why we are still doing this. Is this logic calc summary specific?. We handle list collector already which should cover repeating block answers, why do we need this here now?.

This adds repeating blocks even if the list collector isn't for it isn't on the path.

app/views/handlers/list_action.py
```diff
@@ -18,7 +18,10 @@ def parent_location(self):
         )
 
     def _get_routing_path(self):
-        return self.router.routing_path(section_id=self.parent_location.section_id)
+        return self.router.routing_path(
+            section_id=self.parent_location.section_id,
+            list_item_id=self.parent_location.list_item_id,
```
MebinAbraham, line 23: I think this might lead to confusion, as a parent wouldn't have a list item id?. So either don't set it, or set it to `None`. Otherwise, it makes me think that there is a scenario in which the parent will have a list item id. Worth adding doc strings also.

schemas/test/en/test_new_calculated_summary_repeating_blocks.json
```diff
@@ -0,0 +1,393 @@
...
+                                            "source": "answers",
+                                            "identifier": "transport-additional-cost"
+                                        }
+                                    ]
+                                }
+                            }
+                        },
+                        {
+                            "type": "CalculatedSummary",
+                            "id": "calculated-summary-count",
```
MebinAbraham, line 359: We should demo this calculated summary being used as a value source. From the card:

> As per the scenario, it must be possible to also reference the total in future blocks for piping/validation/logic (routing/skip/enabled). You can already use calculated summary totals, this is to make sure it still works when using repeating questions

app/questionnaire/questionnaire_schema.py
katie-gardner: Yeah that makes sense, I'll change it to `_list_names_by_list_repeating_block` to incorporate the list and keep it similar to other dict names

app/views/contexts/section_summary_context.py
katie-gardner: `ListCollectorBlock` previously would default to section-summary in `list_summary_element` and now that its a configurable param I need to add it, I think from `section_summary_context.py` this is valid to pass in?

app/views/contexts/summary/group.py
katie-gardner: Removed this one and used instance attribute 👍

app/views/contexts/summary/group.py
katie-gardner: Yep - its calc summary specific at the moment, we handle list collectors, but we don't handle list repeating questions, and a calculated summary won't necessarily use all the repeating blocks and answers, so I can't use the existing "ListCollector" case as this is rendering summary elements including all questions rather than the specific ones that have been passed to group in the calc summary case

katie-gardner: Will fix the issue with list collectors that arent on the path

app/views/handlers/list_action.py
```diff
@@ -80,9 +83,7 @@ def handle_post(self):
         )
 
         if self.questionnaire_store_updater.is_dirty():
-            self._routing_path = self.router.routing_path(
-                self.current_location.section_id, self.current_location.list_item_id
-            )
+            self._routing_path = self._get_routing_path()
```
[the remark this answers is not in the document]
katie-gardner: When it was working out which block to route to next, it used list item id in the routing path, which then meant it checked every block in the section with (block_id, list_item_id) rather than (block_id, None) which would be the correct key for a non-repeating section, my thought was that the routing path ought to be based on the list collector rather than its children, but I'm happy to look into another way of fixing this issue if it shouldn't

app/views/contexts/section_summary_context.py
MebinAbraham: Ya, it was more around hard coding it. It might be an existing issue.

app/views/handlers/list_action.py
MebinAbraham: Ya, a list collector in a repeating section shouldn't be allowed.

app/views/handlers/list_action.py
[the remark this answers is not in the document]
MebinAbraham: I'm not sure that I fully understand the scenario. Might be a good one to discuss/demo.

app/views/contexts/summary/group.py
MebinAbraham: Same here, not sure if I fully follow. Isn't this going to add repeating blocks to the list of blocks for all summaries?. It should only be when it is a calculated summary right?.

app/questionnaire/value_source_resolver.py
katie-gardner: Fixed and updated test 👍

app/views/contexts/summary/group.py
katie-gardner: Added a comment as per our chat to say that this relates to calculated summary types and not section/final summary types

app/views/handlers/list_action.py
katie-gardner: removed list_item_id and added comment 👍

app/views/handlers/list_action.py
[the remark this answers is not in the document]
katie-gardner: As discussed have removed the list item id from the method this calls and left this as is

app/views/handlers/list_edit_question.py
```diff
@@ -13,6 +15,31 @@ def is_location_valid(self):
             return False
         return True
 
+    def get_next_location_url(self):
+        """
+        Unless editing from the summary page, If there are repeating blocks and not all are complete, go to the next one
+        """
+        if self._is_returning_to_section_summary():
```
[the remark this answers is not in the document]
katie-gardner: Yep good point - if you use `test_list_collector_repeating_blocks_section_summary` and go to the section-summary, then edit a list item or repeating block answer, previous takes you to the summary

if you do it from the /submit/ page and use previous, you go back to the section summary rather than the /submit/ page as the url had the wrong return to location

Now it exits out to the list collector, because final-summary doesn't have the same handling, will have a look at generalising that to other types of summary

schemas/test/en/test_new_calculated_summary_repeating_blocks.json
katie-gardner: Ah I figured it might just be the same as the similar test for dynamic answers, but happy to do for this one too, added another section and functional test to cover 👍

app/questionnaire/placeholder_renderer.py
```diff
@@ -181,6 +182,29 @@ def _handle_and_resolve_dynamic_answers(
                 else:
                     data_to_render_mutable |= updated_value
 
+    def _handle_repeating_block_id_suffix(
```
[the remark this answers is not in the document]
katie-gardner: I have to go through dynamic answer pointers twice to keep those two separate but assuming that's OK then happy to do that, done

app/questionnaire/schema_utils.py
```diff
@@ -18,9 +18,7 @@ def find_pointers_containing(
         if search_key in input_data:
             yield pointer or ""
         for k, v in input_data.items():
-            if (isinstance(v, dict)) and search_key in v:
-                yield pointer + "/" + k if pointer else "/" + k
-            elif isinstance(v, (list, tuple, dict)):
+            if isinstance(v, (list, tuple, dict)):
```
katie-gardner, line 21: I've done this because in the instance you have something like
`{"question": {"id": "...", "answers": [{"id": }]}}`
it'll see that the search key is in question, and yield the pointer, and not then go through the answers
now will do both

app/questionnaire/placeholder_renderer.py
```diff
@@ -181,6 +182,29 @@ def _handle_and_resolve_dynamic_answers(
...
+        self, *, data_to_render_mutable: dict, list_item_id: str | None
+    ) -> None:
+        """
+        If the data to render is a repeating question, the block-ids/question-ids/answer-ids need suffixing with list_item_id
+        This is so the page renders with valid HTML and avoids duplicate ids for divs when the same block is shown for each list item
+        """
+        if (
+            not list_item_id
+            or data_to_render_mutable.get("id") not in self._schema.repeating_block_ids
```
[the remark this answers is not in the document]
katie-gardner: As suggested, moving it into `Block` did the trick and left placeholders alone
conveniently, `Block` is only used from summaries anyway, so I didn't have to check anything (except being a repeating block)

app/questionnaire/placeholder_renderer.py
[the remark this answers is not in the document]
katie-gardner: Now reverted this since the suffixing is elsewhere

app/questionnaire/schema_utils.py
MebinAbraham: Good spot. Found out the hard way? 💥

katie-gardner: just a bit 😄

app/questionnaire/questionnaire_schema.py
MebinAbraham, line 383: Worth updating name to `_update_dependencies_for_list_repeating_blocks`. Applies to all references of repeating blocks.

app/views/contexts/section_summary_context.py
MebinAbraham: Still outstanding?.

app/views/contexts/summary/block.py
```diff
@@ -111,10 +116,23 @@ def get_question(
...
-    def serialize(self) -> dict[str, Any]:
-        return {
-            "id": self.id,
-            "title": self.title,
-            "number": self.number,
-            "question": self.question,
-        }
+    def _handle_id_suffixing(self, block: dict) -> dict:
+        """
+        If the block is repeating, the block id, as well as any other ids (e.g. question, answer) need updating to be suffixed with list_item_id
```
MebinAbraham, line 121: Worth being explicit, such as `If the block a list collector repeating block then ...`. We want to avoid confusion with repeating inside a repeating section.

app/views/contexts/summary/group.py
MebinAbraham: The paths are still not checked?. It is still adding the blocks regardless?.

app/views/handlers/list_repeating_question.py
```diff
@@ -2,36 +2,64 @@
...
                 return_to_answer_id=self._return_to_answer_id,
                 return_to_block_id=self._return_to_block_id,
             )
-            return repeating_block_url
 
-        return super().get_next_location_url()
+        edit_block = self._schema.get_edit_block_for_list_collector(
+            self.parent_block["id"]
+        )
+        return url_for(
```
MebinAbraham, line 54: Depending on which merges first, this or Pete's. We need to handle edit block not existing and return to parent if not.

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_blocks.spec.js
```diff
@@ -0,0 +1,160 @@
...
+import TransportRepeatingBlock1Page from "../../../generated_pages/new_calculated_summary_repeating_blocks/transport-repeating-block-1-repeating-block.page.js";
+import TransportRepeatingBlock2Page from "../../../generated_pages/new_calculated_summary_repeating_blocks/transport-repeating-block-2-repeating-block.page.js";
+import ListCollectorPage from "../../../generated_pages/new_calculated_summary_repeating_blocks/list-collector.page.js";
+import CalculatedSummarySpendingPage from "../../../generated_pages/new_calculated_summary_repeating_blocks/calculated-summary-spending.page.js";
+import CalculatedSummaryCountPage from "../../../generated_pages/new_calculated_summary_repeating_blocks/calculated-summary-count.page.js";
+import { assertSummaryValues, repeatingAnswerChangeLink } from "../../../helpers";
+import HubPage from "../../../base_pages/hub.page";
+import FamilyJourneysPage from "../../../generated_pages/new_calculated_summary_repeating_blocks/family-journeys.page";
+
+describe("Feature: Calculated Summary using Repeating Blocks", () => {
```
MebinAbraham, line 15: I can't see any, but we would need a test that show the calculated summary value calculation when the list collector is subsequently removed from the path.

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_blocks.spec.js
```diff
@@ -0,0 +1,160 @@
...
+    await assertSummaryValues(["£100.00", "£0.00", "£265.00", "£100.00", "£50.00"]);
+    await $(CalculatedSummarySpendingPage.submit()).click();
+    await expect(await browser.getUrl()).to.contain(CalculatedSummaryCountPage.pageName);
+    await expect(await $(CalculatedSummaryCountPage.calculatedSummaryTitle()).getText()).to.contain(
+      "We calculate the total journeys made per month to be 14. Is this correct?"
+    );
+    await assertSummaryValues(["2", "12"]);
+  });
+
+  it("Given I enter section 2, the calculated summary from section 1 is correctly used to validate my answer", async () => {
```
MebinAbraham, line 147: We should do one of these once a list item is added/removed. We need to also assert the status of the hub to verify the dependencies are working properly.

app/views/contexts/section_summary_context.py
katie-gardner: Would it not be "section-summary" for `section_summary_context` ? 
I've not got an instance attribute I can set for this and the function doesn't accept it as a parameter, so I've added the return to here as ListCollectorBlock doesn't default it to "section-summary" anymore

I could change `SectionSummaryContext` to accept a return to parameter though I think it would still be hardcoded somewhere
I might have missed what you had in mind?

MebinAbraham: 🤦 My bad, was looking at the wrong file. Yes this is correct since it is the section summary context.

app/views/contexts/summary/group.py
katie-gardner: Fixed 👍

app/questionnaire/value_source_resolver.py
```diff
@@ -42,10 +47,14 @@ def _is_answer_on_path(self, answer_id: str) -> bool:
         return True
 
     def _is_block_on_path(self, block_id: str) -> bool:
-        return (
-            self.routing_path_block_ids is not None
-            and block_id in self.routing_path_block_ids
-        )
+        if block_id in self.schema.repeating_block_ids:
+            # repeating blocks aren't on the path, so check the parent list collector
```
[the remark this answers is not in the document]
katie-gardner: Done 👍

app/questionnaire/value_source_resolver.py
```diff
@@ -42,10 +47,14 @@ def _is_answer_on_path(self, answer_id: str) -> bool:
...
+            list_name = self.schema.list_names_by_list_repeating_block_id[block_id]
+            # Type ignore: section and list collector will both exist if the block is repeating
+            section: ImmutableDict = self.schema.get_section_for_block_id(block_id)  # type: ignore
+            list_collector_block: ImmutableDict = self.schema.get_list_collector_for_list(section, list_name)  # type: ignore
+            return list_collector_block["id"] in (self.routing_path_block_ids or [])
```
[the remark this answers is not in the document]
katie-gardner: done 👍

app/questionnaire/questionnaire_schema.py
katie-gardner: updated all except the ones in the list collector classes themself 👍

app/views/contexts/summary/block.py
katie-gardner: Updated as per suggestion

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_blocks.spec.js
katie-gardner: added

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_blocks.spec.js
katie-gardner: Run into a problem with this one, currently calculated summary progress doesn't update by removing an answer from the path, only changing one, looking into this

katie-gardner: As per chat, list collector on/off path doesn't directly update CS progress but add/removing items does as I've changed the extra dependency on the remove block to be for add as well

Added a test for Hub progress when adding a new item and a test for correct values on the CS when list collector is removed from the path

app/views/handlers/list_edit_question.py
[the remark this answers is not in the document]
katie-gardner: Now fixed, test added

berroar (commented):
Only minor thing I noticed when functionally testing is that when you edit an answer in one of the the repeating blocks from the calculated summary, on the return to the Calculated summary page the screen is not anchored on the question that was edited? Is this expected? It works for our normal Calculated Summary schema.

tests/functional/spec/features/calculated_summary/new_calculated_summary_repeating_blocks.spec.js
```diff
@@ -0,0 +1,203 @@
...
+    await $(TransportRepeatingBlock1Page.transportCost()).setValue(0);
+    await $(TransportRepeatingBlock1Page.transportAdditionalCost()).setValue(265);
+    await $(TransportRepeatingBlock1Page.submit()).click();
+    await $(TransportRepeatingBlock2Page.transportCount()).setValue(2);
+    await $(TransportRepeatingBlock2Page.submit()).click();
+    await $(ListCollectorPage.no()).click();
+    await $(ListCollectorPage.submit()).click();
+  });
+
+  it("Given I have a calculated summary using some answers from a repeating block and static answers, I see the correct items and total.", async () => {
```
MebinAbraham, line 47: The format for these should be Given ... When .. Then ...

app/questionnaire/questionnaire_schema.py
```diff
@@ -528,10 +564,17 @@ def _update_answer_dependencies_for_list_source(
...
         # removing an item from a list will require any dependent calculated summaries to be re-confirmed, so cache dependencies
         if remove_block_question := self.get_remove_block_id_for_list(list_name):
-            remove_block_id = self.get_first_answer_id_for_block(remove_block_question)
-            self._list_collector_dynamic_answer_dependencies[block_id] = remove_block_id
+            remove_block_answer_id = self.get_first_answer_id_for_block(
+                remove_block_question
+            )
+            self._list_dependent_block_additional_dependencies[block_id].add(
+                remove_block_answer_id
+            )
```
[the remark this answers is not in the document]
katie-gardner: Validator will throw an error for the schema if any answer for the remove block isn't a radio answer with the `RemoveListItemAndAnswers` action, so I think the first one will remove it, and any others would too (though not sure if two answers with that action would actually work) 
Is the remove block meant to support multiple answers?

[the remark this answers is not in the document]
MebinAbraham: Ah yes, I was thinking multiple options which wouldn't be a problem. It doesn't look like validator prevents multiple answers. It just checks if it radio and 1 option at least has the action, which doesn't rule out multiple answers but happy to leave as is.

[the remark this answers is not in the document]
katie-gardner: Updated to get all ids on the safe side 👍 , and also it keeps it consistent with the other method

katie-gardner:
> Only minor thing I noticed when functionally testing is that when you edit an answer in one of the the repeating blocks from the calculated summary, on the return to the Calculated summary page the screen is not anchored on the question that was edited? Is this expected? It works for our normal Calculated Summary schema.

Well spotted - have now fixed this, on the old list collector summaries the return to answer was taking you to the first instance of the copied answer id, rather than the specific one being edited, so have added the list item id into the return to answer id, similar to how its been used to suffix the answer id

It feels slightly ugly, as the dynamic answers are not handled by the `Block` class, but already have their ids changed in the renderer, and dynamic answers need the suffix to be on the anchor as well. I considered using the answer_id/original_answer_id route for other list scenarios like repeating blocks but I would still then need dynamic answer logic in the answer class to work out what the anchor is
it seemed easier to just say, suffix the return to answer id, but only if id hasn't already been modified. Open to opinions 

Tests expanded for changes to `Answer` and to check the urls for dynamic/repeating blocks

app/questionnaire/questionnaire_schema.py
```diff
@@ -340,6 +357,9 @@ def _populate_answer_dependencies(self) -> None:
                 self._update_answer_dependencies_for_summary(block)
                 continue
 
+            if block["type"] == "ListCollector" and block.get("repeating_blocks"):
```
MebinAbraham, line 360: @petechd FYI, you will need to update this when it merges.

app/questionnaire/value_source_resolver.py
```diff
@@ -100,36 +114,61 @@ def _resolve_list_item_id_for_value_source(
...
+        """Return the list of answers in answer store that correspond to the given list name and dynamic/repeating answer_id"""
+        answer_values: ResolvedAnswerList = []
+        for list_item_id in self.list_store[list_name]:
+            answer_value = self._get_answer_value(
+                answer_id=answer_id, list_item_id=list_item_id
+            )
+            if answer_value is not None and self.escape_answer_values:
+                answer_values.append(escape_answer_value(answer_value))
+            else:
+                answer_values.append(answer_value)
```
MebinAbraham, line 129: With this logic, the else values might contain `None` if self.escape is false. Both should return same list, 1 should just be escaped.

app/views/contexts/summary/block.py
```diff
@@ -12,22 +15,24 @@
...
+        return_to: str | None,
+        return_to_block_id: str | None = None,
         progress_store: ProgressStore,
         language: str,
     ) -> None:
         self.id = block_schema["id"]
         self.title = block_schema.get("title")
         self.number = block_schema.get("number")
+        self.location = location
+        self.schema = schema
```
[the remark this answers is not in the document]
MebinAbraham: Some left in the `__init__` part.

app/views/contexts/summary/list_collector_block.py
```diff
@@ -128,6 +134,21 @@ def list_summary_element(self, summary: Mapping[str, Any]) -> dict[str, Any]:
             **list_summary_context,
         }
 
+    def repeating_block_question_element(self, block: ImmutableDict) -> list[dict]:
```
MebinAbraham, line 137: I think the naming is not quite right. Should this just not be something like `get_repeating_block_answers_blocks`. I think the existing `get_releated_answers` is misleading. It implies you get back a list of answers. We should rename to something like `get_related_answer_blocks_by_list_item`.

app/views/handlers/list_repeating_question.py
MebinAbraham: @petechd FYI.

app/questionnaire/value_source_resolver.py
MebinAbraham: > would there be any other use for resolving a list where it might be relevant to include the Nones

I can't think of any. Possibly in the future.

> it occured to me the original _resolve_answer_value_source would return none if the answer_value is None

That makes sense given we have to return something, None being the default, but when returning a list of items, i think it is fine to only return if we have a value. We haven't got a need for the other case, and given the existing logic for resolving list was only adding to the list if we had an answer (even though it was naively checking the truthiness since it would treat 0 as invalid), it only returned if it has a value.

katie-gardner: Yeah that makes sense to me - will update 👍

app/views/contexts/summary/list_collector_block.py
katie-gardner: One day I'll master naming variables 🙃 , definitely prefer those suggestions 👍  , updated answers as suggested and swapped this one to `get_repeating_block_related_answer_blocks` to keep inline with the second (but can go with original too, dont mind)

app/views/contexts/summary/block.py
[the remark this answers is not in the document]
katie-gardner: sorted 👍

app/views/contexts/summary/list_collector_block.py
MebinAbraham: Happy with the change 👍

tests/functional/generate_pages.py
```diff
@@ -60,7 +60,7 @@
     browser.openQuestionnaire(schema);
   });
 
-  it("Given..., When..., Then...", () => {
+  it("Given..., When..., Then I see...", () => {
```
MebinAbraham, line 63: Bad replace.

katie-gardner: What should this have been? Unedited?

MebinAbraham: Ya, the `Then I see` was for your test example.

MebinAbraham:
Dependencies for remove is not working as expected. There is a small gap with how we have implemented dependencies but we do have pretty much everything we need to fix it.

So using `test_new_calculated_summary_repeating_blocks`. 

1. Complete first section (add any amount of items)
2. Complete section 2 which uses the total from section 1.
3. If you go add a new item or change a value, it will change the status of section 2 to In progress which is correct. However, this does not work for remove.

Repeat the above but instead of adding an item, remove an item and see that Section 2's status is unchanged.

tests/functional/generate_pages.py
katie-gardner: Ah sorry, I misinterpreted the prior comment to mean for all of them - reverted 👍

MebinAbraham:
Another two scenarios that I don't think is quite right.

**Scenario 1:**

So using `test_new_calculated_summary_repeating_blocks`.

1. Get to section summary of section 1 (add 2 items).
2. Change your answer to `How often do you travel by ...?` (which is used on the second calculated summary).
3. It takes you to the first calculated summary instead of the second?. The first doesn't have anything to do with the answer you changed?.

**Scenario 2:**
Might be an existing issue.

1. Continue from step 3 of the previous scenario.
2. When you click previous it takes you to the parent list collector?. (because you can't access the section summary?). I would expect it to return to the repeating block I just edited in step 2.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -165,25 +165,32 @@ def remove_list_item_data(self, list_name: str, list_item_id: str) -> None:
...
-        and their progress updated.
-        (E.g. dynamic-answers depending on the add block could have validation, so need revisiting if an item is removed)
+        If an item is removed from a list, dependencies of any child blocks need to be captured and their progress updated.
+        (E.g. dynamic-answers depending on a child block could have validation, so need revisiting if an item is removed)
         """
         if remove_block_id := self._schema.get_remove_block_id_for_list(list_name):
-            answer_ids = self._schema.get_answer_ids_for_block(remove_block_id)
-            for answer_id in answer_ids:
-                self._capture_block_dependencies_for_answer(answer_id)
+            self._capture_block_dependencies_for_all_answers_in_block(remove_block_id)
```
MebinAbraham, line 172: Minor, we should move this below. I don't see a reason to handle remove like this and add/repeating differently. The one below covers multiple list collectors also.

Could be something like:
```py
child_blocks = (
    list_collector.get("add_block"),
    list_collector.get("remove_block"),
    *list_collector.get("repeating_blocks", []),
)

for child_block in child_blocks:
    if child_block:
        self._capture_block_dependencies_for_all_answers_in_block(
            child_block["id"]
        )
```

MebinAbraham (approved):
LGTM 👍

berroar (approved):

petechd (approved):

PR #1159, opened 2023-07-10
Bug Fix: None language code defaults to "en"

kylelawsonAND:
This PR implements a fix for answer type `mass-metric-tonnes` displaying the tooltip of "metric tons" when it should be "tonnes" regardless of whether the language being set is "en" or "en_GB". 
A fix was implemented in a previous PR to ensure `get_locale` returns "en_GB" instead of "en" when "en" is set as the `language_code`. This however assumed there would always be a `language_code` set, if `None` is provided as the `language_code`, `get_locale` will return "en".
This PR fixes this issue by ensuring `None` `language_code` defaults to `DEFAULT_LANGUAGE_CODE` (en).

### Changes
**session.py**: `cookie_session["language_code"]` defaults to `DEFAULT_LANGUAGE_CODE` if `metadata.language_code` is `None`.

**create_token.py**: New integration test helper method `create_token_v2_with_none_language_code` which takes a schema name and theme, creates the relevant default runner claims, deletes the language code and returns a generated token. 

**test_questionnaire_locale**: New integration test class which has a new test to open a new `test_unit_patterns` survey without a language code in the claims and asserting that the `mass-metric-tonnes` answer tooltip shows "tonnes" and not "metric tons", indicating that a missing language code defaults to "en". 

### How to review
Select either survey `qfs_0002` or `test_unit_patterns` in launcher, and set language to "not set"
Navigate to the answer with type `mass-metric-tonnes`.
Hover over the tooltip and ensure it displays "tonnes", not "mass-metric-tons". 

The new integration test passes, as do all other existing tests.

### DIFF
--- a/app/routes/session.py
+++ b/app/routes/session.py
@@ -23,6 +23,7 @@ from app.helpers.template_helpers import (
 )
 from app.questionnaire import QuestionnaireSchema
 from app.routes.errors import _render_error_page
+from app.settings import DEFAULT_LOCALE
 from app.utilities.metadata_parser import validate_runner_claims
 from app.utilities.metadata_parser_v2 import (
     validate_questionnaire_claims,
@@ -126,7 +127,7 @@ def login() -> Response:
             "account_service_log_out_url"
         )
 
-    cookie_session["language_code"] = metadata.language_code
+    cookie_session["language_code"] = metadata.language_code or DEFAULT_LOCALE
 
     return redirect(url_for("questionnaire.get_questionnaire"))
 
--- a/app/routes/session.py
+++ b/app/routes/session.py
@@ -22,8 +22,8 @@ from app.helpers.template_helpers import (
     render_template,
 )
 from app.questionnaire import QuestionnaireSchema
+from app.questionnaire.questionnaire_schema import DEFAULT_LANGUAGE_CODE
 from app.routes.errors import _render_error_page
-from app.settings import DEFAULT_LOCALE
 from app.utilities.metadata_parser import validate_runner_claims
 from app.utilities.metadata_parser_v2 import (
     validate_questionnaire_claims,
@@ -127,7 +127,7 @@ def login() -> Response:
             "account_service_log_out_url"
         )
 
-    cookie_session["language_code"] = metadata.language_code or DEFAULT_LOCALE
+    cookie_session["language_code"] = metadata.language_code or DEFAULT_LANGUAGE_CODE
 
     return redirect(url_for("questionnaire.get_questionnaire"))
 
--- a/tests/integration/create_token.py
+++ b/tests/integration/create_token.py
@@ -163,5 +163,18 @@ class TokenGenerator:
 
         return self.generate_token(payload_vars)
 
+    def create_token_v2_with_none_language_code(
+        self, schema_name, theme="default", **extra_payload
+    ):
+        payload_for_theme = (
+            PAYLOAD_V2_SOCIAL if theme == "social" else PAYLOAD_V2_BUSINESS
+        )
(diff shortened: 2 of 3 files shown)


app/routes/session.py
```diff
@@ -126,7 +127,7 @@ def login() -> Response:
             "account_service_log_out_url"
         )
 
-    cookie_session["language_code"] = metadata.language_code
+    cookie_session["language_code"] = metadata.language_code or DEFAULT_LOCALE
```
MebinAbraham, line 130: This should be `DEFAULT_LANGUAGE_CODE` not `DEFAULT_LOCALE`

tests/integration/questionnaire/test_questionnaire_locale.py
```diff
@@ -0,0 +1,21 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestSession(IntegrationTestCase):
+    def test_none_language_code_defaults_to_en_GB(self):
```
MebinAbraham, line 5: Minor, but we now have 1 test via Integration test and another in functional test. Can we not just update the existing JS test to cover this scenario?.

kylelawsonAND: We need code coverage, so opted for int test

tests/integration/questionnaire/test_questionnaire_locale.py
```diff
@@ -0,0 +1,21 @@
...
+        )
+        self.get(url=f"/session?token={token}")
+
+        # Skip to the mass-metric-tonnes answer
+        self.post()
+        self.post()
+        self.post()
+        self.post()
+
+        # Then the cookie_session[language_code] will have defaulted to DEFAULT_LOCALE (en_GB), and the tooltip will show "tonnes" not "metric tons"
```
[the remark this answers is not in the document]
kylelawsonAND: Fixed 👍

app/routes/session.py
kylelawsonAND: Fixed 👍

katie-gardner: oops, my bad on the ticket

tests/integration/create_token.py
```diff
@@ -163,5 +163,18 @@ def create_token_with_schema_url(self, schema_name, schema_url, **extra_payload)
 
         return self.generate_token(payload_vars)
 
+    def create_token_v2_with_none_language_code(
+        self, schema_name, theme="default", **extra_payload
+    ):
+        payload_for_theme = (
```
katie-gardner, line 169: Appreciate this mirrors `create_token_v2` but I think for this one theres no need to have the "theme" parameter and the conditional logic based on theme

kylelawsonAND: I had it that way originally, I opted to keep it for extensibility in any future tests

kylelawsonAND: Removed 👍

app/routes/session.py
```diff
@@ -126,7 +127,7 @@ def login() -> Response:
             "account_service_log_out_url"
         )
 
-    cookie_session["language_code"] = metadata.language_code
+    cookie_session["language_code"] = metadata.language_code or DEFAULT_LANGUAGE_CODE
```
MebinAbraham, line 130: The only problem with this is that it doesn't fix it for people unless they relaunch. I think it is fine to leave as it and accept that. Not worth defaulting in get locale also.

katie-gardner (approved):
Steps to reproduce now show correct unit code, test fails without the fix, passes with the fix, looks good, nice 👍

MebinAbraham (approved):
LGTM

PR #1169, opened 2023-07-21
Using supplementary data non list items

katie-gardner:
This PR implements the usage of supplementary data (non list items) in a survey. This adds a new option for value sources of `supplementary_data` which has an identifier refering to a top level key of the supplementary data, and an optional list of selectors for accessing nested data

the changes are supported by [the corresponding validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/209)

### Background (read before you review)
To get this working, the only real change, is adding the `SupplementaryDataStore` to `ValueSourceResolver` and adding one additional method to resolve supplementary data. Unfortunately the number of dependencies for `ValueSourceResolver` is very high. There are many different functions and classes which initialise the class, and many functions and classes that in turn depend on those. It ended up being around 93 different functions/classes that needed the store to be added, which get called in about 330 different places

To ease the burden of the developers reviewing this PR, I have split it into 3 distinct commits. 
[Add SupplementaryDataStore dependency](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1169/commits/ece4f5ff0d552fd284c169b76d4d2ed53a4d0051)
The first commit is purely adding the supplementary data store as a parameter everywhere that it needs to go, tests, classes, functions, etc. Functionally no change, tests still pass, 100% coverage

[Using Supplementary data non list items](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1169/commits/419174dfbbfa60bb717c7b573333ce7fa322c1cc)
The second commit is adding the actual functionality to be able to use supplementary data, which is just a change to value source resolver and an adjustment to the test data being used by the mock endpoint, to show it working.

[Testing Supplementary Data non list items](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1169/commits/0c0b42a9aa29ae4dbf0b15a3d315fa475ab1b5af)
The final commit is the tests. The functional and integration tests will only work if the mock endpoint is running. (These will fail in the pipeline until we have replaced this with a dockerised version)

A couple of other commits since for PR comments/Linting
[More tests](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1169/commits/cf7cc35d18526b3b1dcfcefc12825f83f7dc4410)

### How to review
Read through the trello card to understand the requirements and check that the test schema and tests have covered what is asked for.

In terms of reviewing code, I would suggest skim reading the first commit, then checking the other two in depth.

### How to test functionally
Following the guidance in run-mock-sds-endpoint.md start the mock sds api and pick the `test_supplementary_data_non_list_items` schema. 
The dropdown for sds dataset id will show two different ids, these are the same structure but with different data, check that placeholders render correctly with each of them. You should also be able to preserve your response id, relaunch the survey with the updated dataset which has some different data and see your original answers against the new placeholders.

### Checklist


tests/app/questionnaire/rules/test_rule_evaluator.py
```diff
@@ -5,7 +5,12 @@
 from freezegun import freeze_time
```
MebinAbraham, line 5: Rule evaluator has tests for the different source type, we should do the same for SDS.

app/submitter/convert_payload_0_0_1.py
```diff
@@ -25,6 +30,7 @@ def convert_answers_to_payload_0_0_1(
     schema: QuestionnaireSchema,
     full_routing_path: Iterable[RoutingPath],
     progress_store: ProgressStore,
+    supplementary_data_store: SupplementaryDataStore,
```
MebinAbraham, line 33: This is missing from the doc strings. There are also a few others.

schemas/test/en/test_supplementary_data_non_list_items.json
```diff
@@ -0,0 +1,503 @@
...
+                                            "identifier": "total_uk_sales"
+                                        },
+                                        "answers_to_calculate": ["sales-bristol-answer", "sales-london-answer"],
+                                        "conditions": ["less than", "equals"]
+                                    }
+                                ],
+                                "answers": [
+                                    {
+                                        "id": "sales-bristol-answer",
+                                        "label": "Bristol Sales",
```
MebinAbraham, line 258: We only recently added placeholder in label, worth adding supplementary data here.

schemas/test/en/test_supplementary_data_non_list_items.json
```diff
@@ -0,0 +1,503 @@
+{
```
MebinAbraham, line 1: It would be good to demo https://github.com/ONSdigital/eq-questionnaire-runner/pull/1122 this using supplementary data. Either in the file that was introduced there or this one.

tests/app/questionnaire/test_dynamic_answer_options.py
```diff
@@ -1,7 +1,12 @@
 # pylint: disable=redefined-outer-name
```
MebinAbraham, line 1: Would be good to demo this with SDS.

schemas/test/en/test_supplementary_data_non_list_items.json
```diff
@@ -0,0 +1,503 @@
...
+            "type": "string"
+        }
+    ],
+    "questionnaire_flow": {
+        "type": "Hub",
+        "options": {
+            "required_completed_sections": ["introduction-section"]
+        }
+    },
+    "sections": [
```
MebinAbraham, line 38: Should add a repeating section and a repeating block to demo you can pipe in non list item data inside repeating blocks/sections.

I don't think the schema needs to be `non_list_items`. For list items, we will need repeats, so we can extend the same survey instead of duplicating most of it when we support list item values. Thoughts?

MebinAbraham (commented):
Just a quick initial pass. Will finish off Monday. Only some minor comments atm.

Nice work though, nice and easy to review also.

schemas/test/en/test_supplementary_data_non_list_items.json
katie-gardner: Good idea - will rename so it can just be extended and add in the repeating section

MebinAbraham: Right now it doesn't seem to check whether list item id should be used which means inside a repeating section/block, non-repeating data won't work. You would need the similar check as answers sources to work out if we are in repeating section etc before using the list item id.

app/submitter/convert_payload_0_0_1.py
katie-gardner: Added 👍 I couldn't actually find any others?

schemas/test/en/test_supplementary_data_non_list_items.json
katie-gardner: Removed the guidance from one of the datasets so can be tested with and without 👍

tests/app/questionnaire/test_dynamic_answer_options.py
katie-gardner: Added dynamic date options 👍

tests/app/questionnaire/rules/test_rule_evaluator.py
katie-gardner: Added a couple of extra tests, not sure how much extra ill be testing by adding it throughout so just picked a couple if that seems suitable?

schemas/test/en/test_supplementary_data_non_list_items.json
katie-gardner: Renamed to `all_variations` (but open to a better name if anyone has one), and its now got a repeating section.

The data is working in a repeating section because it doesn't make use of list item id, so at the moment its the other way round, in a repeating section repeating data will not work. I can look to add these checks and test but thinking that'd probably be more within the scope of the data with list items ticket?

MebinAbraham: Do we even need `all_variantions`?.

> The data is working in a repeating section because it doesn't make use of list item id, so at the moment its the other way round, in a repeating section repeating data will not work. I can look to add these checks and test but thinking that'd probably be more within the scope of the data with list items ticket?

Sounds good. Happy for it to be tackled in the next card. 👍

katie-gardner: I'd ditch it but we have an empty schema with that name, and the mocking for tests like `test_session` gets a bit ugly if I use the new schema as it can't handle the introduction block, maybe that one could be `_empty` or actually - if I just enable the hub from the start instead of after the intro that might also sort it

app/submitter/convert_payload_0_0_1.py
MebinAbraham: Thought I saw more warning, maybe not.

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=supplementary-data-value-source
```

tests/app/questionnaire/rules/test_rule_evaluator.py
MebinAbraham: That is plenty.

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -784,3 +791,45 @@ def test_progress_values_source_throws_if_no_location_given():
...
+            None,
+        ),
+    ],
+)
+def test_supplementary_data_value_source_non_list_items(
+    supplementary_data_store_with_data, value_source, expected_result
+):
+    value_source_resolver = get_value_source_resolver(
+        supplementary_data_store=supplementary_data_store_with_data
+    )
```
MebinAbraham, line 826: This should have a variation which does this same thing in the context of a repeat. So pass location with a list item id set. You can just use another parameterise.

schemas/test/en/test_supplementary_data_non_list_items.json
MebinAbraham: Im a bit confused as to why we need the empty one?.

katie-gardner: It made the tests that don't care about the actual supplementary data simpler - but I agree, I'll get rid of it and fix the tests

MebinAbraham (commented):
Functionally all good 👍 
Other than the remaining comments, LGTM.

schemas/test/en/test_supplementary_data_all_variations.json
```diff
@@ -0,0 +1,934 @@
+{
```
MebinAbraham, line 1: We should enable view submitted response and ensure that it renders fine.

Can enable using 

```json
"post_submission": {
    "view_response": true
},
```

tests/functional/spec/features/supplementary_data/supplementary_data_non_list_items.spec.js
```diff
@@ -0,0 +1,130 @@
...
+import TradingPage from "../../../generated_pages/supplementary_data_all_variations/trading.page";
+import ListCollectorPage from "../../../generated_pages/supplementary_data_all_variations/list-collector.page";
+import AnyEmployeesPage from "../../../generated_pages/supplementary_data_all_variations/any-employees.page";
+import Section2Page from "../../../generated_pages/supplementary_data_all_variations/section-2-summary.page.js";
+import AddEmployeePage from "../../../generated_pages/supplementary_data_all_variations/list-collector-add.page.js";
+import Section3Page from "../../../generated_pages/supplementary_data_all_variations/section-3-summary.page.js";
+import LengthOfEmploymentPage from "../../../generated_pages/supplementary_data_all_variations/length-of-employment.page.js";
+import SupermarketTransportPage from "../../../generated_pages/new_calculated_summary_repeating_and_static_answers/supermarket-transport.page";
+
+describe("Using supplementary data non list items", () => {
```
MebinAbraham, line 18: Can we enable the section summary and it ensure that it renders fine, worth doing view submitted response also.

katie-gardner: Done 👍 - I've noticed that, the section context is build via:
`_get_refactored_groups` turning `group-3` into `group-3-0`
and then the summary context builds all the groups, which creates duplicates of `group-3-0` and then `set_unique_group_ids` sets this to `group-3-0` and `group-3-0-1`

This probably isn't quite what is wanted on the summary page, but it may be preferable for this to be fixed in a follow up ticket (? ), it just means that for my functional tests, the last two selectors use an odd group content index

MebinAbraham: Thanks, I see that it is done for view submitted but can you add it for section summary also?. 

> I've noticed that, the section context is built via:
_get_refactored_groups turning group-3 into group-3-0
and then the summary context builds all the groups, which creates duplicates of group-3-0 and then set_unique_group_ids sets this to group-3-0 and group-3-0-1

> This probably isn't quite what is wanted on the summary page, but it may be preferable for this to be fixed in a follow up ticket (? ), it just means that for my functional tests, the last two selectors use an odd group content index

Agreed. It complicates the selectors however let's not spend that on that now. That is some tech debt we can progress in the near future.

tests/app/questionnaire/test_value_source_resolver.py
katie-gardner: Added 👍

MebinAbraham (approved):
Nice one

MebinAbraham (approved):
Nice one

kylelawsonAND (approved):
Code looks sound, testing scenarios work as expected, same with coverage scenarios. 
PR description is fantastic too 👌 👌

PR #1172, opened 2023-07-27
Authenticate SDS Requests with GCP

kylelawsonAND:
It is intended that requests made to SDS must be authenticated to restrict and control API access.
This PR implements a mechanism for fetching OIDC credentials for a given source as defined by a service implementation. In this PR, local and GCP services have been implemented. 

### How to review
**Local**

1. Run SDS runner dependencies (launcher, mock SDS etc).
2. Launch an SDS survey.
3. You should enter into the survey as normal.
4. Check runner logs and confirm that before fetching SDS data there are logs indicating that it Runner would fetch OIDC credentials. 

**GCP**

1. Log into the GCP console in chrome.
2. Navigate to [this GCP project ](https://console.cloud.google.com/welcome?project=composite-haiku-394009), or alternatively, create a project in eq-sandbox, add a service account and client id under APIs and Services > Credentials.
3. Download the service account json file and place it somewhere in the project.
4. Set the env vars:
```
SDS_OAUTH2_CLIENT_ID=<the Client Id in the GCP project>
GOOGLE_APPLICATION_CREDENTIALS=<path to the service account json (you may need it wrapped in quotes)>
```
7. Edit .development.env such that `OIDC_TOKEN_BACKEND=gcp`
8. Launch an SDS survey.
9. You should enter into the survey as normal.
10. Check runner logs and confirm that before fetching SDS data there are logs indicating that it Runner has fetched OIDC credentials from GCP.


Pipfile
```diff
@@ -73,6 +73,9 @@ simplejson = "*"
 markupsafe = "*"
 pdfkit = "*"
 ordered-set = "*"
+cachetools = "*"
+types-cachetools = "*"
+types-pytz = "*"
```
MebinAbraham, line 78: These should be added to dev dependencies.

app/oidc/local_oidc.py
```diff
@@ -0,0 +1,37 @@
...
+class LocalOIDCCredentialsService(OIDCCredentialsService):
+    @refresh_oidc_credentials
+    @ttl_cache(maxsize=None, ttl=TTL)
+    def get_credentials(
+        self,
+        *,
+        iap_client_id: str,
+    ) -> Credentials:
+        logger.info("generating local oidc credentials")
+        logger.info(f"using IAP Client ID {iap_client_id}")
```
MebinAbraham, line 34: Use a single log line and avoid passing data via f-strings. The IAP client id can be attached by passed it as keyword arguments to the logger. Applies to the other instance also.

app/oidc/oidc.py
```diff
@@ -0,0 +1,66 @@
...
+from google.auth.credentials import Credentials
+from google.auth.transport.requests import Request
+from structlog import get_logger
+
+from app.settings import OIDC_TOKEN_LEEWAY_IN_SECONDS
+
+P = ParamSpec("P")
+
+logger = get_logger()
+TTL = 3600 - OIDC_TOKEN_LEEWAY_IN_SECONDS  # 1 hour minus leeway in seconds
```
MebinAbraham, line 16: For clarity, we should assign 3600 to a constant or add comments. i.e `OIDC_TOKEN_VALIDITY_IN_SECONDS`

app/oidc/oidc.py
```diff
@@ -0,0 +1,66 @@
...
+    @functools.wraps(func)
+    def wrapper_refresh_oidc_credentials(
+        *args: P.args, **kwargs: P.kwargs
+    ) -> Credentials:
+        credentials = func(*args, **kwargs)
+        expiry = (
+            pytz.utc.localize(dt=credentials.expiry)
+            if is_naive(credentials.expiry)
+            else credentials.expiry
+        )
```
MebinAbraham, line 50: We don't need to localise it but I'm not sure what it is conditional anyway, the lib should be consistent as it is either timezone aware or not.

But either way, you can just add the timezone to the datetime instead of localising it as the lib always passes a UTC date regardless of it is it naive, therefore you can just call `.replace(tzinfo=timezone.utc)` on the date.

```py
expiry = credentials.expiry.replace(tzinfo=timezone.utc)
if expiry < get_expiry_from_ttl(TTL):
   ...
```

app/oidc/oidc.py
```diff
@@ -0,0 +1,66 @@
...
+        if expiry < get_expiry_from_ttl(TTL):
+            logger.info("refreshing oidc credentials")
```
MebinAbraham, line 52: Worth adding the client id.

app/oidc/gcp_oidc.py
```diff
@@ -0,0 +1,35 @@
...
+from google.auth.transport.requests import Request
+from google.oauth2.id_token import fetch_id_token_credentials
+from structlog import get_logger
+
+from app.oidc.oidc import TTL, OIDCCredentialsService, refresh_oidc_credentials
+
+logger = get_logger()
+
+
+class GCPCredentialsService(OIDCCredentialsService):
```
MebinAbraham, line 12: OIDC should be in the name for clarity.

app/settings.py
```diff
@@ -146,6 +146,14 @@ def utcoffset_or_fail(date_value, key):
 
 SDS_API_BASE_URL = os.getenv("SDS_API_BASE_URL")
 
+OIDC_TOKEN_BACKEND = os.getenv("OIDC_TOKEN_BACKEND")
+
+OIDC_TOKEN_LEEWAY_IN_SECONDS = int(os.getenv("OIDC_TOKEN_LEEWAY_IN_SECONDS", "300"))
+
+SDS_OAUTH2_CLIENT_ID = os.getenv(
+    "SDS_OAUTH2_CLIENT_ID", "test-client-id.apps.googleusercontent.com"
```
MebinAbraham, line 154: Test client id shouldn't be committed. Defaults should be prod first.

app/services/supplementary_data.py
```diff
@@ -64,6 +66,13 @@ def get_supplementary_data_v1(
         backoff_factor=SUPPLEMENTARY_DATA_REQUEST_BACKOFF_FACTOR,
     )
 
+    # Type ignore: oidc_credentials_service is a singleton of this application
+    oidc_credentials_service: OIDCCredentialsService = current_app.eq["oidc_credentials_service"]  # type: ignore
+    credentials = oidc_credentials_service.get_credentials(
+        iap_client_id=SDS_OAUTH2_CLIENT_ID
+    )
```
MebinAbraham, line 73: It would be clearer to raise an exception if we don't have a client id than the request failing due to passing None.

app/oidc/oidc.py
kylelawsonAND: Annoyingly its the google auth lib which gives a utc datetime for the expiry on credentials, but its naive. So to do the calculations I needed to add it, we can replace like you say, or localize it.
The reason I opted to localize is to guarantee the decorator can handle the credentials as it will error if it tries to calculate validity with naive and aware dates.

app/oidc/gcp_oidc.py
kylelawsonAND: I did have this, and it was `GCPOIDCCredentialService` which looked like an model number or something, shall we go with that, or change the impl part to be a suffix like `OIDCCredentialServiceGCP` and `OIDCCredentialServiceLocal`?

app/settings.py
kylelawsonAND: Good point, should i remove the default entirely for now then since we won't know prod default?

app/oidc/oidc.py
MebinAbraham: I understood why you needed to localise but not why it was conditional `if naive` (it should always be naive or not). Was it because the local one wasn't naive?.

The reason I suggested `.replace` is, localise implies it could be timezone aware which isn't utc but we want to convert to utc, but that is never the case. So since it is always utc, we just need to add tzinfo, for the sake of comparing two timezone aware objects.

kylelawsonAND: Oh wait sorry you weren't meaning to move it out of the decorator - ignore me, will use replace each time instead of check naivety and localize 👍

app/oidc/gcp_oidc.py
MebinAbraham: 😄 True. Yes go with `OIDCCredentialServiceGCP` and `OIDCCredentialServiceLocal`

app/settings.py
MebinAbraham: Yes. We won't expose it in the public source anyway even if it is not strictly sensitive. We will pass it in when deploying the app.

app/oidc/oidc.py
kylelawsonAND: The way we'd get the client id in the context of the decorator would be to decode the token, at that point I'm not sure that's worthwhile?

kylelawsonAND: Have defaulted SDS_OAUTH2_CLIENT_ID to None to address another comment, so will just use that, in local the log will show None which isn't really an issue

app/oidc/oidc.py
```diff
@@ -0,0 +1,66 @@
...
+            credentials.refresh(Request())
```
MebinAbraham, line 53: Given this refresh is called outside the function that has the cache wrapper, it isn't clear how exactly cache is working given we are not calling the function.

We should document that this is working because `credentials` references the same cached object in memory, therefore when refreshing this credentials object, it is actually the same object the cache holds. We should also document this must be used with in-memory cache as things like redis wouldn't work with this pattern.

I do like the decorator pattern but I'm not sure I'm a fan of the cache working because it happens to update the same object in memory. If no one else has a concerned with this, I'm happy to yield (just make sure we have through tests using the `cache_info()` method and good documentation).

app/oidc/oidc.py
kylelawsonAND: Done 👍

app/oidc/oidc.py
kylelawsonAND: Done 👍

.functional-tests.env
```diff
@@ -29,3 +29,4 @@ COOKIE_SETTINGS_URL=#
 EQ_SUBMISSION_CONFIRMATION_BACKEND=log
 VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS=35
 SDS_API_BASE_URL=http://localhost:5003
+OIDC_TOKEN_BACKEND=log
```
MebinAbraham, line 32: Should be local?

kylelawsonAND: Woops good catch - fixed 👍

app/oidc/oidc.py
MebinAbraham: Doesn't look like it, it is binding the `SDS_OAUTH2_CLIENT_ID` explicit which won;t be true when the same thing is used for CIR. Just pass `args` and `kwargs` to the logger info, actually. As `kwargs` will contain the iap client id in the current case. Either pass both, or enforce one of positional / keyword only arguments.

kylelawsonAND: Binding kwargs to the log 👍

app/oidc/gcp_oidc.py
kylelawsonAND: Renamed 👍

app/oidc/oidc.py
kylelawsonAND: Yeah the existence of the TTL cache ensures that if a token goes unused for over the credential's TTL, it is discarded behind the scenes. 
The reason for the refresh when requesting credentials, ensures that even if a returned cached result has 0.1 seconds of lifetime left, the requesting function has credentials which are guaranteed to be valid for the proceeding usage.  
It also assumes the credentials acquired from the decorated function have this refresh method, so should only decorate functions which return google OIDC credentials in whatever implementation that may be.

MebinAbraham: The usage of why we refresh was clear, but how the cache was being updated with the refreshed credentials isn't. If it has expired then refreshing is correct.

My main point being:

> We should document that this is working because credentials references the same cached object in memory, therefore when refreshing this credentials object, it is actually the same object the cache holds. We should also document this must be used with in-memory cache as things like redis wouldn't work with this pattern. 

The fine detail of the cache working before it is referring the same object in memory is the gotcha. We need to make this clear through some comments.

MebinAbraham: For context for others, the current pattern doesn't quite work as expected. Since the credentials are refreshed outside the function, the ttl of the cache is never extended and we will end up referring the cache at 55 mins, then again at 60 mins enough though the refresh at 55 minutes, should have extended it for another 55 minutes.

Details:
- At 55 minutes, the cache is classed as expired based on our check, so the credential gets refreshed without calling the function. So credential is not valid for another hours (55 mins based on our check)
- A request after would call the function and generate the credential again because the TTL cache has expired even though we refreshed the credentials moments ago.

This is hard to see at the moment cause the cache ttl and our expiry check is using the same TTL. If the TTL cache was 60 mins and our check was 55 minutes, this become evident.

MebinAbraham:
Reminder to deploy the new env vars to PROD via TF before this before.

app/oidc/oidc.py
katie-gardner: As talked about - have removed the additional decorator in favour of simplicity and just have the ttl cache of 55 minutes now

tests/app/oidc/test_gcp_oidc.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+    mock_token_fetch.return_value.refresh.assert_called_once_with(mock_request())
+
+    # credentials are cached
+    oidc_credentials_service.get_credentials(iap_client_id=TEST_SDS_OAUTH2_CLIENT_ID)
+    assert mock_token_fetch.call_count == 1
+    assert mock_token_fetch.return_value.refresh.call_count == 1
+
+    # re-fetched when cache expires
+    oidc_credentials_service.get_credentials.cache_clear()
```
MebinAbraham, line 26: We should have a test with mocked TTLs, this current one asserts that the cache works as expected if the cache is cleared, but this line is not something we execute. So if the lib had a bug and TTLs weren't expiring this test would still pass, but credentials would end up as expired and not refresh.

katie-gardner: I ran into some issues with this one, we can't mock time with freezegun because the TTLCache has a cached version of the time function, I've replaced the timer with something that is updated by freezegun and checks that the cache behaves as we want it to at the 30, 55 min, 109 min, 110 min markers, let me know if thats the sort of thing you were thinking/thoughts

tests/app/oidc/test_gcp_oidc.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@freeze_time("2023-01-01T12:00:00")
+@patch("app.oidc.gcp_oidc.Request", Mock)
+@patch("app.oidc.gcp_oidc.fetch_id_token_credentials")
+def test_oidc_credentials_service_gcp_ttl(mock_token_fetch):
+    """
+    by default, TTLCache uses a cached version of time.monotonic for the timer
+    which means that mocking time with freezegun doesn't work, as it won't affect the timer
+    to work around this and test the caching, we can replace the timer with a time.monotonic lambda
```
MebinAbraham, line 32: I would link tot his open issue: https://github.com/spulec/freezegun/issues/477

tests/app/oidc/test_gcp_oidc.py
```diff
@@ -0,0 +1,62 @@
...
+    oidc_credentials_service.get_credentials(iap_client_id=TEST_SDS_OAUTH2_CLIENT_ID)
+    assert mock_token_fetch.call_count == 1
+    assert mock_token_fetch.return_value.refresh.call_count == 1
+
+    for datetime_str, expected_call_count in [
+        ("2023-01-01T12:30:00", 1),  # still valid after 30 minutes
+        ("2023-01-01T12:55:00", 2),  # becomes invalid after 55 minutes
+        ("2023-01-01T13:49:00", 2),  # new fetch is valid for a further 55 minutes
+        ("2023-01-01T13:50:00", 3),  # then also becomes invalid
+    ]:
```
MebinAbraham, line 53: I suggest some edits to comments for clarity:

```py
for datetime_str, expected_call_count in [
    ("2023-01-01T12:30:00", 1),  # still valid after 30 minutes (valid until 12:55)
    ("2023-01-01T12:55:00", 2),  # becomes invalid after 55 minutes (this new call makes it valid until 13:50)
    ("2023-01-01T13:49:00", 2),  # this fetch is valid for another 1 minute until 13:50
    ("2023-01-01T13:50:00", 3),  # then at 13:50 it becomes invalid so new call is made
]:
```

Optional: `datetime_str` could also be something like `datetime_to_invoke_at`, to imply that is when we call the function.

tests/app/oidc/test_gcp_oidc.py
```diff
@@ -0,0 +1,62 @@
...
+    which means that mocking time with freezegun doesn't work, as it won't affect the timer
+    to work around this and test the caching, we can replace the timer with a time.monotonic lambda
+    """
+    oidc_credentials_service = OIDCCredentialsServiceGCP()
+
+    # overwrite the timer
+    oidc_credentials_service.get_credentials = ttl_cache(
+        maxsize=None,
+        ttl=TTL,
+        timer=lambda: time.monotonic(),  # pylint: disable=unnecessary-lambda
```
MebinAbraham, line 40: The linting is genuine, you don't need a lambda, just pass it a callable, i.e `time.monotonic`, so `time=time.monotonic`

MebinAbraham (commented):
Just need to retest functionally, otherwise looking good.

tests/app/oidc/test_gcp_oidc.py
MebinAbraham: Exactly what I was looking for, nice one. The custom ttl wrapper isn't great, but that is fine given it is an issue with the lib and not in our control.

app/oidc/gcp_oidc.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class OIDCCredentialsServiceGCP(OIDCCredentialsService):
+    @staticmethod
+    @ttl_cache(maxsize=None, ttl=TTL)
+    def get_credentials(*, iap_client_id: str) -> Credentials:
+        """
+        The credentials are valid for an hour, and we use a ttl cache of 55 minutes
+        this is to ensure that even in the edge case where the timers on the cache and the credentials don't quite align,
+        the 5 minutes leeway should be more than enough to cover it and guarantee safety
```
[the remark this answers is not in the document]
katie-gardner: done

tests/app/oidc/test_gcp_oidc.py
katie-gardner: Good shout - added

app/oidc/gcp_oidc.py
```diff
@@ -0,0 +1,36 @@
...
+
+logger = get_logger()
+
+
+class OIDCCredentialsServiceGCP(OIDCCredentialsService):
+    @staticmethod
+    @ttl_cache(maxsize=None, ttl=TTL)
+    def get_credentials(*, iap_client_id: str) -> Credentials:
+        """
+        The credentials are valid for an hour, and we use a ttl cache of this minus a configurable leeway
```
MebinAbraham, line 20: Same here, the creds are valid for `OIDC_TOKEN_VALIDITY_IN_SECONDS`

app/oidc/gcp_oidc.py
```diff
@@ -0,0 +1,36 @@
...
+        the OIDC_TOKEN_LEEWAY_IN_SECONDS should be more than enough to cover it and guarantee safety
+        """
+        logger.info("fetching OIDC credentials from GCP", iap_client_id=iap_client_id)
+
+        request = Request()
+        target_audience = iap_client_id
+
+        credentials = fetch_id_token_credentials(
+            audience=target_audience, request=request
+        )
```
MebinAbraham, line 31: Minor, since it is keyword args, the vars don't add much?.
```py
        credentials = fetch_id_token_credentials(
            audience=iap_client_id, request=Request()
        )
```

tests/app/oidc/test_gcp_oidc.py
```diff
@@ -0,0 +1,65 @@
...
+    mock_token_fetch.assert_called_once_with(
+        audience=TEST_SDS_OAUTH2_CLIENT_ID, request=mock_request()
+    )
+    mock_token_fetch.return_value.refresh.assert_called_once_with(mock_request())
+
+
+@freeze_time("2023-01-01T12:00:00")
+@patch("app.oidc.gcp_oidc.Request", Mock)
+@patch("app.oidc.gcp_oidc.fetch_id_token_credentials")
+def test_oidc_credentials_service_gcp_ttl(mock_token_fetch):
```
MebinAbraham, line 28: Minor, you don't need `gcp` in naming, the file is specifically for gcp. Can even drop `oidc_credentials_service` as this file is about that. My suggestion would be `test_get_credentials` and `test_get_credentials_ttl` or `test_ttl`.

app/oidc/gcp_oidc.py
katie-gardner: request is used twice, but got rid of the other 👍

tests/app/oidc/test_gcp_oidc.py
katie-gardner: Sure thing, done

app/oidc/gcp_oidc.py
```diff
@@ -0,0 +1,35 @@
...
+
+TTL = OIDC_TOKEN_VALIDITY_IN_SECONDS - OIDC_TOKEN_LEEWAY_IN_SECONDS
+
+logger = get_logger()
+
+
+class OIDCCredentialsServiceGCP(OIDCCredentialsService):
+    @staticmethod
+    @ttl_cache(maxsize=None, ttl=TTL)
+    def get_credentials(*, iap_client_id: str) -> Credentials:
```
MebinAbraham, line 18: Something to consider is a retry, we pretty much retry all our external requests.

tests/app/oidc/test_gcp_oidc.py
```diff
@@ -0,0 +1,65 @@
...
+from mock import Mock, patch
+
+from app.oidc.gcp_oidc import TTL, OIDCCredentialsServiceGCP
+
+TEST_SDS_OAUTH2_CLIENT_ID = "TEST_SDS_OAUTH2_CLIENT_ID"
+
+
+@patch("app.oidc.gcp_oidc.Request")
+@patch("app.oidc.gcp_oidc.fetch_id_token_credentials")
+def test_get_credentials(mock_token_fetch, mock_request):
```
MebinAbraham, line 14: One for thought, it might be good to have an example of the previous cache test one you had for the purpose of if an expection is raised then it is not cached. It might be a given. So happy to leave depending on your thoughts.

MebinAbraham (commented):
Functionally all good 👍

MebinAbraham (commented):
Functionally all good 👍

app/oidc/gcp_oidc.py
MebinAbraham: You can pass in a session object to the `Request(session=session)`. Here you could reuse our `get_retryable_session()` method and configure it appropriately. I haven't tested if this works out of the box.

For now, I'm happy for us to reuse:

```py
SUPPLEMENTARY_DATA_REQUEST_BACKOFF_FACTOR = 0.2
SUPPLEMENTARY_DATA_REQUEST_MAX_RETRIES = 2  # Totals no. of request should be 3. The initial request + SUPPLEMENTARY_DATA_REQUEST_MAX_RETRIES
SUPPLEMENTARY_DATA_REQUEST_TIMEOUT = 3
SUPPLEMENTARY_DATA_REQUEST_RETRY_STATUS_CODES = [
    408,
    429,
    500,
    502,
    503,
    504,
]
```

We can tune later.

MebinAbraham: Can we check if Timeout (should be) and Connection errors are retried.

katie-gardner: have done some digging and the google auth library has a retry built into it, in each case it ends up in `_token_endpoint_request_no_throw` and - if the initial request fails it goes into this
```
retries = _exponential_backoff.ExponentialBackoff()
    for _ in retries:
        request_succeeded, response_data, retryable_error = _perform_request()
        if request_succeeded or not retryable_error:
            return request_succeeded, response_data, retryable_error

    return False, response_data, retryable_error
```
which by default attempts 3 times with an exponential backoff iterator

So I am thinking we probably don't need to add our own on top?

katie-gardner: `retryable_error` will return true if it errors and the status code is internal error, service unavailable or request timeout, so we should be covered on the necessary ones

MebinAbraham: For context for other, the above is not used when calling the metadata server, but it does look like GCP does retry up to 5 times when calling the metadata server. It does seem to only retry Transport errors which seems appropriate given it is not an external call. Therefore, as discussed, just some tests to verify the retry is working.

katie-gardner: I've pushed up an attempt at this - I'm not sure how much value we're going to get from this test given the lib should have tests and the amount I ended up needing to mock to get this to work
or maybe I've missed something obvious 😄 
is this on the lines you were thinking or did you have a different idea of how it would look?

katie-gardner: I suppose its only 4 patches which isn't too bad

MebinAbraham: Will have a look, but the purpose was mainly to ensure that there is a retry than retry works. The lib tests would check it, but our test would fail if Google were to remove the retry, i.e lets say they expect users to pass in a retry strategy. We had similar issues with GCS.

katie-gardner: Ah cool - makes sense, this should cover it having the retries

MebinAbraham: LGTM.

petechd (approved):

petechd:
@katie-gardner ignore all other validation errors, it must be wrong unit, like "mile" instead of "length-mile" in that schema.

tests/app/oidc/test_gcp_oidc.py
```diff
@@ -0,0 +1,100 @@
+import time
```
MebinAbraham, line 1: The `tests`/app/oidc` folder needs an `__ini__.py` file

tests/app/oidc/test_gcp_oidc.py
katie-gardner: Happy with that - added a simple one that checks the cache stays empty if it fails

MebinAbraham (approved):

PR #1173, opened 2023-07-28
Using supplementary data list items

katie-gardner:
This PR implements the usage of supplementary data list items in a survey. This is an extension to the [non list items pr 
](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1169)
The main functional changes in this ticket, are:

- altering the error handling in `supplementary_data_store.py` so that nested data can be optional, such as the `guidance_exclude` block, but still raising an error if you try to index non-dictionary non-none data with a selector. 
- Adding a method to check if the data being referenced is repeating.
- Updating `value_source_resolver.py` to make sure that a list item id is used if the data being referenced is repeating.

Other than that, the test data has been combined to reduce the number of files, and unit tests and functional tests have been added.

### How to review
Read through the trello card and check that the test schema allows for testing the mentioned requirements
Check the changes to value source resolver and supplementary data store and that the changes are suitably tested

### How to test functionally
Exactly as with the other PR: following the guidance in run-mock-sds-endpoint.md start the mock sds api and pick the `test_supplementary_data` schema. 
The dropdown for sds dataset id will show two different ids, these are the same structure but with different data, check that placeholders render correctly with each of them. You should also be able to preserve your response id, relaunch the survey with the updated dataset which has some different data and see your original answers against the new placeholders. In this instance, check that the list items have updated correctly and any dynamic answers or repeating sections are updated/removed/added accordingly

Test any unhappy path you can think of, and make sure the functional test passes locally (removing the skip)

### Checklist
* [ ] The new functional test passes if you uncomment the .skip and have the mock endpoint running


app/data_models/supplementary_data_store.py
```diff
@@ -94,17 +94,31 @@ def get_data(
...
         """
+        if self.is_data_repeating(identifier) and not list_item_id:
+            raise InvalidSupplementaryDataSelector(
+                f"Cannot reference items from `{identifier}` outside a repeat"
+            )
+
         value = self._data_map.get((identifier, list_item_id))
-        # for nested data, index with each selector
+        # for nested data, index with each selector, or return None if there is no data to index
         for selector in selectors or []:
```
petechd, line 104: Just having a quick glance while reviewing mock sds. I know it's not this PR but can you explain the logic here? if `selectors` is false then you iterate through empty list so don't iterate at all?

katie-gardner: It was a suggestion made on a previous PR to reduce the amount of nesting, so that I wouldn't need
```
if selectors:
 for selector in selectors:
   # extra indent
```
and could have the one liner instead which I think is quite tidy, but I don't mind too much what we go with, can update if preferred

app/data_models/supplementary_data_store.py
```diff
@@ -94,17 +94,31 @@ def get_data(
...
+            if value is None:
+                return None
             if not isinstance(value, dict):
```
MebinAbraham, line 107: Existing, but I suggest updating this to `Mapping` as code isn't dict specific.

app/questionnaire/placeholder_transforms.py
```diff
@@ -56,9 +56,12 @@ def format_date(self, date_to_format: str, date_format: str) -> str:
...
     @staticmethod
-    def format_list(list_to_format: Sequence[str]) -> str:
-        formatted_list = "<ul>" + "".join(f"<li>{item}</li>" for item in list_to_format)
-        formatted_list += "</ul>"
+    def format_list(list_to_format: Sequence[str] | None) -> str:
+        formatted_list = (
+            f"<ul>{''.join(f'<li>{item}</li>' for item in list_to_format)}</ul>"
+            if list_to_format
+            else ""
+        )
```
MebinAbraham, line 64: Would future-proof this if you wrap `list_to_format` in `self.remove_empty_from_list`

app/questionnaire/value_source_resolver.py
```diff
@@ -88,25 +88,29 @@ def _get_answer_value(
...
-        if list_item_id:
-            return list_item_id
+        if value_source["source"] == "supplementary_data":
+            return (
+                self.list_item_id
+                if self.supplementary_data_store.is_data_repeating(
+                    value_source["identifier"]
+                )
+                else None
+            )
```
MebinAbraham, line 113: Any reason the behaviour was changed?. Previously it would get to the final `if` even when one of these resolved to None, whereas, with this change it will return None and not get to the final check.

MebinAbraham (commented):
I've raised a card to look at removing a lot of our `$("body").getText())` usages, as we seem to be using it more and more. We'll add selectors where appropriate when we got a bit more time.

Looks good. Nice one. We need to remember to swap the list collector block type once the other PR merges.

app/questionnaire/placeholder_transforms.py
katie-gardner: Good shout - done

app/questionnaire/value_source_resolver.py
katie-gardner: I changed it as I don't think they can resolve to None, a `list_item_selector` must have an identifier that is a non-empty string, so I made it consistent with the pattern I added for `supplementary_data` which does need to return None if its None, thinking this would help readability too.  Can adjust if I've missed a case though

katie-gardner: At the time I was thinking that it was not possible for that path to do anything different... im questioning it now..
If value source was location, then identifier has to be `to_list_item_id` or `list_item_id` and if the getattr returns None, and it then goes into the final check, it will fail `is_repeating_answer` so return None anyway, and I think similarly for the list case, so functionally I believe the behaviour is consistent

MebinAbraham: Yes, I think it would be the same. 👍

MebinAbraham (approved):
👍

petechd (approved):

Yuyuutsu (approved):
Looks good to me 🖖

PR #1174, opened 2023-08-01
Remove non-JS GA

MebinAbraham:
Since we currently don't have a non-js flow for accepting cookies, we should not be enabling GA when JS is off.
The PR removes the non-js GA tracking script.

### How to review
Ensure:
- the non-JS GA script is not in the page source when GTM is enabled but JS is disabled.
- when the GTM variables are provided, and JS is enabled, the gtm scripts are loaded as expected.
- no `frame-src` CSP errors are seen in the console.

### Checklist


Yuyuutsu (approved):
Compared with staging branch. `Google tag manager (no script)` has been removed and I can still see the `ons-js-enabled` . LGTM !

rmccar (approved):

PR #1180, opened 2023-08-08
update WebdriverIO to use internal chrome driver

Yuyuutsu:
This PR is updating the WebdriverIO packages for the new [ChromedriverIO v8.14](https://github.com/webdriverio/webdriverio/releases/tag/v8.14.0) release which uses the new [Chrome for Testing](https://developer.chrome.com/blog/chrome-for-testing/). WebdriverIO now manages the drivers rather than us, so we can now remove `chromedriver` and `wdio-chromedriver-service`. 
### How to review
Run Functional tests locally and you should now see a new chrome browser called Chrome for Testing. 

### Checklist

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -138,10 +138,6 @@ jobs:
       EQ_RUN_FUNCTIONAL_TESTS_HEADLESS: True
     steps:
       - uses: actions/checkout@v3
-      - name: Update to latest stable Chrome
-        run: |
-          wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
-          sudo apt install ./google-chrome-stable_current_amd64.deb
       - uses: actions/setup-node@v3
         with:
           node-version: "18.14.0"
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
         "format": "yarn format:tests && yarn format:test-schemas",
         "format:tests": "prettier \"tests/functional/**/*.js\" --write && eslint --fix \"tests/functional/**/*.js\"",
         "format:test-schemas": "prettier \"schemas/test/*/*.json\" --write && eslint --fix \"schemas/test/*/*.json\"",
-        "wdio": "wdio run ./wdio.conf.js"
+        "wdio": "wdio run ./tests/functional/wdio.conf.js"
     },
     "devDependencies": {
         "@babel/core": "^7.22.9",
@@ -33,7 +33,6 @@
         "@wdio/mocha-framework": "^8.14.0",
         "@wdio/spec-reporter": "^8.14.0",
         "chai": "^4.3.7",
-        "chromedriver": "^115.0.0",
         "eslint": "^8.46.0",
         "eslint-cli": "^1.1.1",
         "eslint-config-standard": "^17.1.0",
@@ -52,7 +51,6 @@
         "prettier": "^3.0.1",
         "typescript": "^5.1.6",
         "uuid": "^9.0.0",
-        "wdio-chromedriver-service": "^8.1.1",
         "webdriverio": "8.14.3"
     },
     "prettier": {}
--- a/tests/functional/wdio.conf.js
+++ b/tests/functional/wdio.conf.js
@@ -53,6 +53,7 @@ export const config = {
   capabilities: [
     {
       browserName: "chrome",
+      browserVersion: "stable",
       // If outputDir is provided WebdriverIO can capture driver session logs
       // it is possible to configure which logTypes to include/exclude.
       // excludeDriverLogs: ['*'], // pass '*' to exclude all driver session logs
@@ -116,7 +117,7 @@ export const config = {
   // Services take over a specific job you don't want to take care of. They enhance
   // your test setup with almost no effort. Unlike plugins, they don't add new
   // commands. Instead, they hook themselves up into the test process.
-  services: ["chromedriver"],
+  services: [],
 
   // Framework you want to run your specs with.
   // The following are supported: Mocha, Jasmine, and Cucumber
--- a/yarn.lock
+++ b/yarn.lock
@@ -1176,11 +1176,6 @@
   dependencies:
     defer-to-connect "^2.0.1"
 
-"@testim/chrome-version@^1.1.3":
-  version "1.1.3"
-  resolved "https://registry.yarnpkg.com/@testim/chrome-version/-/chrome-version-1.1.3.tgz#fbb68696899d7b8c1b9b891eded9c04fe2cd5529"
-  integrity sha512-g697J3WxV/Zytemz8aTuKjTGYtta9+02kva3C1xc7KXB8GdbfE1akGJIsZLyY/FSh2QrnE+fiB7vmWU3XNcb6A==
-
 "@tootallnate/quickjs-emscripten@^0.23.0":
   version "0.23.0"
   resolved "https://registry.yarnpkg.com/@tootallnate/quickjs-emscripten/-/quickjs-emscripten-0.23.0.tgz#db4ecfd499a9765ab24002c3b696d02e6d32a12c"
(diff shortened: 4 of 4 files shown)


tests/functional/wdio.conf.js
```diff
@@ -45,14 +45,15 @@ export const config = {
   // and 30 processes will get spawned. The property handles how many capabilities
   // from the same test should run tests.
   //
-  maxInstances: 2,
+  maxInstances: 1,
```
MebinAbraham, line 48: Did we have to drop this?.

package.json
```diff
@@ -52,7 +51,6 @@
         "prettier": "^3.0.1",
         "typescript": "^5.1.6",
         "uuid": "^9.0.0",
-        "wdio-chromedriver-service": "^8.1.1",
         "webdriverio": "8.14.3"
```
MebinAbraham, line 54:
```suggestion
        "webdriverio": "^8.14.3"
```

tests/functional/wdio.conf.js
Yuyuutsu: My change didnt save earlier when adding the commit. Just pushed before ur comment 😂

MebinAbraham: I'm not following, did it need to be lowered?.

Yuyuutsu: No, 1 Max instance works fine. Unless we wanted to keep 2 ? Was just saying the instance change didnt get pushed when I initially git added

MebinAbraham: Yes, it should be 2 unless it can't be. Lowering it to 1 is going to twice as long to run.

Yuyuutsu: Oh I understand now, reverted. Will have a look at the actions results again

Yuyuutsu: Reverting back to 1 instance because tests are failing too frequently, needing to retry multiple times

MebinAbraham: In that case, I would set this to use an env var and default it to 2 so locally it isn't 1. Then in Actions it should override to use 1. Similar to how we do `EQ_RUN_FUNCTIONAL_TESTS_HEADLESS`.

MebinAbraham: Also, why does it fail at 2, is it to due with the issue I linked previously or some other reason?. Is it something that we can't get around?.

petechd (commented):
Tested functionally and it works as expected!

tests/functional/wdio.conf.js
Yuyuutsu: Had a look for a while now cannot seem to find a reason. Will revert back to two because it is working locally fine.

berroar (approved):

MebinAbraham (commented):
Do we need the webdriver install in actions or can that be removed?

petechd:
> Do we need the webdriver install in actions or can that be removed?

Removed install from actions now.

katie-gardner (approved):
Changes look sound and runs exactly as expected locally 👍

MebinAbraham:
> > Do we need the webdriver install in actions or can that be removed?
> 
> Removed install from actions now.

Doesn't look like it has?

.github/workflows/pull_request.yml
```diff
@@ -43,9 +43,6 @@ jobs:
         with:
           path: ${{ steps.get-yarn-cache.outputs.dir }}
           key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
-      - name: Install yarn deps
-        if: steps.cache-yarn.outputs.cache-hit != 'true'
-        run: yarn
```
MebinAbraham, line 48: Why was this removed?. If we clear cache of Actions, this would break the actions jobs.

petechd: Reverted now, thanks.

.functional-tests.env
```diff
@@ -30,3 +30,4 @@ EQ_SUBMISSION_CONFIRMATION_BACKEND=log
 VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS=35
 SDS_API_BASE_URL=http://localhost:5003
 OIDC_TOKEN_BACKEND=local
+EQ_MAX_INSTANCES=1
```
MebinAbraham, line 33: This is vague and makes me think it controls runner's no. of instances. I would rename to something like `EQ_FUNCTIONAL_TEST_MAX_INSTANCES`.

petechd:
> > > Do we need the webdriver install in actions or can that be removed?
> > 
> > 
> > Removed install from actions now.
> 
> Doesn't look like it has?

Removing install step doesn't have any adverse effect on actions so getting rid of it now.

.functional-tests.env
petechd: @MebinAbraham I had to slightly change it now so it's a bool and ternary used to get its value, but at least it seems like all flakiness is gone. Maybe we should rename this var now?

tests/functional/wdio.conf.js
```diff
@@ -45,14 +45,15 @@ export const config = {
   // and 30 processes will get spawned. The property handles how many capabilities
   // from the same test should run tests.
   //
-  maxInstances: 2,
+  maxInstances: parseInt(process.env.EQ_FUNCTIONAL_TEST_MAX_INSTANCES) || 2,
```
MebinAbraham, line 48: Currently, it makes the env mandatory. This way, it is not mandatory.
```suggestion
  maxInstances: parseInt(process.env.EQ_FUNCTIONAL_TEST_MAX_INSTANCES || 2),
```

.functional-tests.env
```diff
@@ -30,3 +30,4 @@ EQ_SUBMISSION_CONFIRMATION_BACKEND=log
 VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS=35
 SDS_API_BASE_URL=http://localhost:5003
 OIDC_TOKEN_BACKEND=local
+EQ_FUNCTIONAL_TEST_MAX_INSTANCES=1
```
MebinAbraham, line 33: Is this needed here?

package.json
```diff
@@ -52,8 +51,7 @@
         "prettier": "^3.0.1",
         "typescript": "^5.1.6",
         "uuid": "^9.0.0",
-        "wdio-chromedriver-service": "^8.1.1",
-        "webdriverio": "8.14.3"
+        "webdriverio": "^8.14.3"
```
MebinAbraham, line 54: Update the yarn lock as 8.15.0 is out and it fixes the max instance issue. https://github.com/webdriverio/webdriverio/pull/10952
See if it is still flaky after upgrading.

.functional-tests.env
petechd: OK, added just in case but will remove now.

package.json
petechd: Still very flaky on two instances after this update, had two runs resulting in multiple suites failing on multiple tests, reverted it back to one and it pass all tests on first try..

MebinAbraham: Have we looked at why it might be happending?.

petechd: I did few more test runs with 2 instances and is similar kind of flakiness today as 1 instance so we could possible keep it as 2?

MebinAbraham: Sorry I'm not following, are you saying 1 is flaky as well?. Have we identified why it is flaky?.

petechd: They are both flaky but, from checking the logs, when using one instance we get same type and amount of errors and failed builds as on main branch (same errors/behaviour as on main), when we increase the number of instances we get failed tests but no logs indicating what the problem is, just failed tests.

MebinAbraham: Right okay, on 1 it is no flakier than main but on 2, it is error for other reasons.
So we should try to understand why 2 is flaky and if we can't get to the bottom of it, proceed with 1, and add a TODO comment to revisit and up to 2.

petechd: Seems like 1 is slightly less flaky than main but hard to capture and measure this. That TODO you mention is already added, I will try to get some meaningful error logs from 2.

MebinAbraham (approved):

package.json
MebinAbraham: Let's get this moving, as it blocks local development. Can you raise a card to look into this bug?. We can tackle it separately.

petechd: OK, will merge it and raise the card, thanks.

PR #1183, opened 2023-08-11
Bug Fix: Schemas with dynamic answers blocks in separate section to the list collector not loading

berroar:

Fixes a bug where schemas wouldn't load when the dynamic answer block was not in the same section as its dependent list collector.

### How to review

Use the `test_dynamic_answers_list_source` schema. Check that the schema is able to be launched with the newly added section containing dynamic answers.

### Checklist


berroar:
[Dynamic answers does not work when it is not the same section as the list collector](https://trello.com/c/JqiRm8Nt/5816-dynamic-answers-does-not-work-when-it-is-not-the-same-section-as-the-list-collector)

app/questionnaire/questionnaire_schema.py
```diff
@@ -555,14 +559,17 @@ def _update_answer_dependencies_for_list_source(
...
         """
         # Type ignore: section will always exist at this point, same with optional returns below
-        section: ImmutableDict = self.get_section_for_block_id(block_id)  # type: ignore
-        list_collector: ImmutableDict = self.get_list_collector_for_list(  # type: ignore
-            section=section,
+        section_id = self._list_name_to_section_id_origin_map[list_name]
+
+        section = self.get_section(section_id)
+
+        list_collector = self.get_list_collector_for_list(
```
katie-gardner, line 566: Non blocking
Since you now have this map - could `get_list_collector_for_list` be improved by including section lookup instead of it needing to be looked up and passed in everywhere its being used?

katie-gardner (approved):
Nice, fixes the issue as expected, and looks good 👍 
One optional idea

app/questionnaire/questionnaire_schema.py
```diff
@@ -302,6 +303,9 @@ def _get_blocks_by_id(self) -> dict[str, ImmutableDict]:
                     "PrimaryPersonListCollector",
                     "RelationshipCollector",
                 ):
+                    self._list_name_to_section_id_origin_map[
+                        block["for_list"]
+                    ] = self._parent_id_map[group["id"]]
```
MebinAbraham, line 308: Im not sure if the naming is accurate. This wouldn't be the origin, it would be the last section to use this list name. If 3 sections used the same list name, the id would be for section 3.

I don't think we have any validation to prevent a list name being used by multiple list collectors as long as they are in separate sections?. Therefore, this would need to hold a list of section id, then you need to loop over each list collector for each section and handle dependencies.

MebinAbraham: Related to https://github.com/ONSdigital/eq-questionnaire-validator/pull/205#discussion_r1273889828

app/questionnaire/questionnaire_schema.py
```diff
@@ -62,7 +62,7 @@ def __init__(
     ):
         self._parent_id_map: dict[str, str] = {}
         self._list_name_to_section_map: dict[str, list[str]] = {}
```
MebinAbraham, line 64: I suggest updating this existing one to something like `self._section_dependenices_by_list_name`.  Open to other suggestions.

app/questionnaire/questionnaire_schema.py
```diff
@@ -62,7 +62,7 @@ def __init__(
...
-        self._list_name_to_section_id_origin_map: dict[str, str] = {}
+        self._list_name_to_section_id_origin_map: dict[str, list] = {}
```
MebinAbraham, line 65: I suggest updating this to something like `self._list_collector_section_ids_by_list_name`. Open to other suggestions.

app/questionnaire/questionnaire_schema.py
```diff
@@ -303,9 +303,14 @@ def _get_blocks_by_id(self) -> dict[str, ImmutableDict]:
...
-                        block["for_list"]
-                    ] = self._parent_id_map[group["id"]]
+                    if self._list_name_to_section_id_origin_map.get(block["for_list"]):
+                        self._list_name_to_section_id_origin_map[
+                            block["for_list"]
+                        ].append(self._parent_id_map[group["id"]])
+                    else:
+                        self._list_name_to_section_id_origin_map[block["for_list"]] = [
+                            self._parent_id_map[group["id"]]
+                        ]
```
MebinAbraham, line 313: Use a default dict?

app/questionnaire/questionnaire_schema.py
```diff
@@ -907,23 +913,29 @@ def get_questions(self, question_id: str) -> list[ImmutableDict] | None:
...
-        )
+                blocks.extend(
+                    block
+                    for block in self.get_blocks_for_section(section)
+                    if block["type"] == collector_type and block["for_list"] == for_list
+                )
+
+        return blocks
 
     def get_list_collector_for_list(
```
MebinAbraham, line 933: This method should get `get_list_collectors_for_list` plural.

app/questionnaire/questionnaire_schema.py
```diff
@@ -907,23 +913,29 @@ def get_questions(self, question_id: str) -> list[ImmutableDict] | None:
...
         self, for_list: str, primary: bool = False
-    ) -> ImmutableDict | None:
-        # Type ignore: Section will exist at this point
-        section: ImmutableDict = self.get_section(self._list_name_to_section_id_origin_map[for_list])  # type: ignore
+    ) -> list[ImmutableDict]:
+        sections = self._list_name_to_section_id_origin_map[for_list]
 
-        return next(self.get_list_collectors_for_list(section, for_list, primary))
+        return self.get_list_collectors_for_list(sections, for_list, primary)
```
MebinAbraham, line 938: This name needs changes as the above comment will clash with this. This can be `get_list_collectors_for_list_for_sections`.

app/questionnaire/value_source_resolver.py
```diff
@@ -56,8 +56,14 @@ def _is_block_on_path(self, block_id: str) -> bool:
...
-            list_collector_block: ImmutableDict = self.schema.get_list_collector_for_list(list_name)  # type: ignore
-            return list_collector_block["id"] in self.routing_path_block_ids
+            section: ImmutableDict = self.schema.get_section_for_block_id(block_id)  # type: ignore
+            list_collector_blocks: ImmutableDict = self.schema.get_list_collector_for_list(list_name)  # type: ignore
+
+            for list_collector_block in list_collector_blocks:
+                if section["id"] == self.schema.get_section_id_for_block_id(
+                    list_collector_block["id"]
+                ):
+                    return list_collector_block["id"] in self.routing_path_block_ids
```
MebinAbraham, line 66: This doesn't seem quite right. You probably that the method call where you pass in list name and section, instead of fetching all then filtering.

app/views/contexts/summary/list_collector_block.py
```diff
@@ -71,9 +71,7 @@ def list_summary_element(self, summary: Mapping) -> dict:
         list_model = self._list_store[summary["for_list"]]
 
         list_collector_blocks = list(
-            self._schema.get_list_collectors_for_list(
-                self._section, for_list=summary["for_list"]
-            )
+            self._schema.get_list_collector_for_list(for_list=summary["for_list"])
         )
```
MebinAbraham, line 75: This should fetch the list collector for the current section only?

app/questionnaire/questionnaire_schema.py
berroar: Updated 👍

app/questionnaire/questionnaire_schema.py
berroar: Updated 👍

app/questionnaire/questionnaire_schema.py
berroar: Updated 👍

app/questionnaire/questionnaire_schema.py
berroar: Updated 👍

app/views/contexts/summary/list_collector_block.py
berroar: Updated 👍

app/questionnaire/value_source_resolver.py
```diff
@@ -55,10 +55,13 @@ def _is_block_on_path(self, block_id: str) -> bool:
...
-            # Type ignore: section and list collector will both exist if the block is repeating
+            # Type ignore: list collector will exist if the block is repeating
             section: ImmutableDict = self.schema.get_section_for_block_id(block_id)  # type: ignore
-            list_collector_block: ImmutableDict = self.schema.get_list_collector_for_list(section, list_name)  # type: ignore
-            return list_collector_block["id"] in self.routing_path_block_ids
+            list_collector_blocks: list = self.schema.get_list_collectors_for_list(for_list=list_name, section_id=section["id"])  # type: ignore
+
+            for list_collector_block in list_collector_blocks:
+                if list_collector_block["id"] in self.routing_path_block_ids:
+                    return True
```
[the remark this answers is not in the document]
katie-gardner: appreciate this is a horrible example and I cannot think of any reason why a schema would do this, but there might be other ways this edge case could happen

yarn.lock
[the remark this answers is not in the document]
berroar: Oops 🙈 undone now!

app/questionnaire/value_source_resolver.py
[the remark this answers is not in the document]
berroar: Yeah wasn't sure about this, i think the previous logic was just checking the first list collector within the section

app/questionnaire/questionnaire_schema.py
```diff
@@ -912,7 +909,7 @@ def get_questions(self, question_id: str) -> list[ImmutableDict] | None:
         """
         return self._questions_by_id.get(question_id)
 
-    def get_list_collectors_for_list(
+    def get_list_collectors_for_list_for_sections(
         self, sections: list, for_list: str, primary: bool = False
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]
katie-gardner: ah - the return of the function needs it too, but rest are covered 👍

app/questionnaire/value_source_resolver.py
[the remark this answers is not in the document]
katie-gardner: Yeah I think it might be awkward to check in the current form, might be easier if we have something like `self._list_names_by_list_repeating_block_id` referring to the specific list collector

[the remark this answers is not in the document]
berroar: Sounds like a good idea 👍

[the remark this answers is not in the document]
berroar: Refactored now based on your suggestion 👍 although not even sure it needs a new method - just using the parent id map to lookup the list collector?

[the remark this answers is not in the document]
katie-gardner: yeah agreed, since its only once this looks like the way to go, nice

app/questionnaire/questionnaire_schema.py
berroar: Updated to handle multiple list collectors in separate sections as discussed, in-line with other PR referenced above. A new ticket will be created to add tests for the scenarios where we have multiple list collectors.

app/questionnaire/questionnaire_schema.py
```diff
@@ -61,7 +61,10 @@ def __init__(
         self, questionnaire_json: Mapping, language_code: str = DEFAULT_LANGUAGE_CODE
     ):
         self._parent_id_map: dict[str, str] = {}
-        self._list_name_to_section_map: dict[str, list[str]] = {}
+        self._section_dependencies_by_list_name: dict[str, list[str]] = {}
+        self._list_collector_section_ids_by_list_name: dict[str, list] = defaultdict(
```
katie-gardner, line 65: Minor -`list[str]` ?

katie-gardner (approved):
New naming is clear and all the paths I've tested are working as expected 👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -905,31 +909,36 @@ def get_questions(self, question_id: str) -> list[ImmutableDict] | None:
...
     def get_list_collectors_for_list(
-        section: Mapping, for_list: str, primary: bool = False
-    ) -> Generator[ImmutableDict, None, None]:
-        collector_type = "PrimaryPersonListCollector" if primary else "ListCollector"
+        self, for_list: str, primary: bool = False, section_id: str | None = None
+    ) -> list[ImmutableDict]:
+        if section_id:
+            sections = [section_id]
+        else:
+            sections = self._list_collector_section_ids_by_list_name[for_list]
```
MebinAbraham, line 936: Minor could be ternary.

MebinAbraham (approved):

app/questionnaire/questionnaire_schema.py
berroar: Added 👍

app/questionnaire/questionnaire_schema.py
berroar: Added 👍

PR #1184, opened 2023-08-14
Bug Fix: Editing list items 404 when using multiple list collectors with repeating blocks

petechd:
This updates `list_edit_question` and `list_repeating_question` so repeating blocks being used to find the first incomplete one are limited to the particular list now.

### How to review
New assertions and logic added (existing journey and brand new one), check functional test changes locally as it is set to skipped now in build.

### Checklist


petechd:
[Bug: Multiple list collectors using repeating blocks can lead to page not found errors](https://trello.com/c/CprkFM8G/5820-bug-multiple-list-collectors-using-repeating-blocks-can-lead-to-page-not-found-errors)

app/views/handlers/list_repeating_question.py
```diff
@@ -67,7 +67,10 @@ def get_previous_location_url(self) -> str:
     def handle_post(self) -> None:
         self.questionnaire_store_updater.add_completed_location(self.current_location)
         if not self.get_first_incomplete_list_repeating_block_location_for_list_item(
-            repeating_block_ids=self.repeating_block_ids,
+            repeating_block_ids=[
```
katie-gardner, line 70: this change is going to be needed for the `get_previous_location_url` as well, currently it will use the index of the repeating block against all repeating blocks, not just those for the list - so I reckon it should be a reusable property, see comment on list edit question

app/views/handlers/list_edit_question.py
```diff
@@ -24,7 +24,10 @@ def get_next_location_url(self) -> str:
             return url
 
         if first_incomplete_block := self.get_first_incomplete_list_repeating_block_location_for_list_item(
-            repeating_block_ids=self._schema.list_collector_repeating_block_ids,
+            repeating_block_ids=[
```
katie-gardner, line 27: Since `ListRepeatingQuestion` inherits from this, `repeating_block_ids` could be a cached property so you don't have to define this multiple times

app/views/handlers/list_repeating_question.py
```diff
@@ -67,7 +67,10 @@ def get_previous_location_url(self) -> str:
     def handle_post(self) -> None:
         self.questionnaire_store_updater.add_completed_location(self.current_location)
```
katie-gardner, line 68: can't comment above, but as the current `self.repeating_block_ids` is incorrect, I think fix/remove it (probably remove)

katie-gardner (changes requested):
Fixes the issue as on the card 👍 
However.... found another! If you take the `test_list_collector_repeating_blocks_section_summary` and copy and paste the section with the list collector, replacing `companies` with `businesses` or anything so that there are two list collectors with repeating blocks then

1. Complete first list collector with repeating blocks
2. Go into second list collector with repeating blocks
3. Add a list item
4. On the first repeating block click previous

You get a page not found error, because in `list_repeating_question.py` the `get_previous_location` method uses all repeating blocks in the schema to get the `repeating_block_index` which in this case is 3 because there was a previous list collector with repeating blocks, and so it tries to go to the previous repeating block of `companies-repeating-block-2` which isn't valid for the new list collector

Think its just a matter of making an identical fix to the one you've already made, but we should probably have a test for this

petechd:
@katie-gardner I've added schema with two list collectors with repeating blocks in two different sections and integration test that breaks without these changes in the code. I'm not sure about extending existing schemas as having an extra schema with two sections with repeating blocks is good to have as separate thing for testing in my opinion.

katie-gardner:
> @katie-gardner I've added schema with two list collectors each with repeating blocks in two different sections and integration test that breaks without these changes requested above in the code (both the first bug and the second bug you've just discovered re-surface in this test). I'm not sure about extending existing schemas as having an extra schema with two sections with repeating blocks is good to have as separate thing for testing in my opinion.

Nice one, issue fixed 👍 
The `test_list_collector_repeating_blocks_with_hub` schema is an exact copy of the `test_list_collector_repeating_blocks_section_summary` except with Hub instead of Linear. I think it would be helpful for your new schema to use Hub instead of Linear so we can test unhappy path cases swapping between the sections, so I'm all good with having two schemas, but I would delete the `hub` one because now we have 3 schemas which are almost identical and there is nothing the hub one can test that you can't test with your new schema and just sticking to the first section.

petechd:
> > @katie-gardner I've added schema with two list collectors each with repeating blocks in two different sections and integration test that breaks without these changes requested above in the code (both the first bug and the second bug you've just discovered re-surface in this test). I'm not sure about extending existing schemas as having an extra schema with two sections with repeating blocks is good to have as separate thing for testing in my opinion.
> 
> Nice one, issue fixed 👍 The `test_list_collector_repeating_blocks_with_hub` schema is an exact copy of the `test_list_collector_repeating_blocks_section_summary` except with Hub instead of Linear. I think it would be helpful for your new schema to use Hub instead of Linear so we can test unhappy path cases swapping between the sections, so I'm all good with having two schemas, but I would delete the `hub` one because now we have 3 schemas which are almost identical and there is nothing the hub one can test that you can't test with your new schema and just sticking to the first section.

Ok, refactored existing "hub" schema as requested above, fixed tests. Thanks.

schemas/test/en/test_list_collector_repeating_blocks_two_list_collectors.json
```diff
@@ -0,0 +1,828 @@
...
+                                ]
+                            }
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "id": "section-businesses",
+            "title": "General insurance business",
```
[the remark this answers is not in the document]
petechd: Done, thanks!

katie-gardner (approved):
Both issues fixed and all works as expected 👍

tests/functional/spec/features/supplementary_data/supplementary_data.spec.js
```diff
@@ -433,7 +457,32 @@ describe.skip("Using supplementary data", () => {
     await expect(await $(ViewSubmittedResponsePage.productReportingContent(1)).$$(summaryItems)[3].getText()).to.equal("Value of sales from other categories");
     await expect(await $(ViewSubmittedResponsePage.productReportingContent(1)).$$(summaryValues)[0].getText()).to.equal("£110.00");
     await expect(await $(ViewSubmittedResponsePage.productReportingContent(1)).$$(summaryValues)[1].getText()).to.equal("£220.00");
-    await expect(await $(ViewSubmittedResponsePage.productReportingContent(1)).$$(summaryValues)[2].getText()).to.equal("No answer provided");
+    await expect(await $(ViewSubmittedResponsePage.productReportingContent(1)).$$(summaryValues)[2].getText()).to.equal("£440.00");
     await expect(await $(ViewSubmittedResponsePage.productReportingContent(1)).$$(summaryValues)[3].getText()).to.equal("£330.00");
   });
 });
+describe.skip("Using supplementary data (editing list item bug)", () => {
```
MebinAbraham, line 464: The description shouldn't mention the bug, it should test some feature.

tests/functional/spec/features/supplementary_data/supplementary_data.spec.js
```diff
@@ -353,6 +354,29 @@ describe.skip("Using supplementary data", () => {
     await expect(await $(Section4Page.employmentStart()).getText()).to.contain("10 October 1999");
     await $(Section4Page.submit()).click();
   });
+  it("Given I add some additional employees via a list collector, When I edit newly added supplementary list item, Then I safely return to list collector page", async () => {
```
MebinAbraham, line 357: They are not supplementary list items, if they are normal list collectors?.

MebinAbraham (changes requested):
The new schema change (extra section) has nothing to do with supplementary data?. We should add them as part of test list collector repeating block schema instead of repurposing supplementary data schema.

tests/functional/spec/features/supplementary_data/supplementary_data.spec.js
```diff
@@ -433,7 +457,32 @@ describe.skip("Using supplementary data", () => {
...
+    await $(IntroductionBlockPage.submit()).click();
+    await $(HubPage.summaryRowLink("section-3")).click();
+    await $(AnyAdditionalEmployeesPage.yes()).click();
+    await $(AnyAdditionalEmployeesPage.submit()).click();
+    await $(AddAdditionalEmployeePage.employeeFirstName()).setValue("Joe");
+    await $(AddAdditionalEmployeePage.employeeLastName()).setValue("Bloggs");
+    await $(AddAdditionalEmployeePage.submit()).click();
+    await $(ListCollectorAdditionalPage.listEditLink(1)).click();
+    await $(AddAdditionalEmployeePage.submit()).click();
+    await expect(await browser.getUrl()).to.contain(ListCollectorAdditionalPage.pageName);
```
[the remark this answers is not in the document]
petechd: Just normal list collector flow so I will remove it here and do everything in repeating blocks normal schema as you said in a comment below.

petechd:
> The new schema change (extra section) has nothing to do with supplementary data?. We should add them as part of test list collector repeating block schema instead of repurposing supplementary data schema.

Correct, I will move the tests to repeating blocks and hub schema.

petechd:
> Functionally all good 👍

Moved the test to repeating blocks with hub schema spec now.

tests/functional/spec/features/supplementary_data/supplementary_data.spec.js
petechd: Removed this test now and added to repeating blocks spec.

tests/functional/spec/features/supplementary_data/supplementary_data.spec.js
petechd: Removed now.

tests/integration/questionnaire/test_questionnaire_list_collector_repeating_blocks.py
```diff
@@ -474,3 +474,25 @@ def test_edit_repeating_block_from_submit_page_returns_to_submit_page(
...
+    ):
+        self.launchSurvey("test_list_collector_repeating_blocks_with_hub")
+        self.post({"responsible-party-answer": "Yes"})
+
+        # Add some items and progress to second list collector
+        self.add_three_companies()
+        self.post({"any-other-companies-or-branches-answer": "No"})
+        self.post({"any-other-trading-details-answer": "No other details"})
+        self.post()
+        self.get("/questionnaire/sections/section-businesses/")
```
MebinAbraham, line 489: This tests needs more well-defined comments, it is not obvious why it is doing this in this way. i.e when coming back to it, it is not clear why we just randomly go to some specific URL. I suggest adding comments around why it means, i.e this is the second list collector section etc.

tests/functional/spec/features/repeating_blocks/list_collector_repeating_blocks.spec.js
```diff
@@ -320,11 +321,36 @@ describe("Given a journey through the list collector with repeating blocks, in a
     await $(CompaniesRepeatingBlock2Page.authorisedTraderUkRadioYes()).click();
     await $(CompaniesRepeatingBlock2Page.submit()).click();
   });
-
-  it("The section can now be submitted and and hub completed.", async () => {
+  it("Given I add additional company via a list collector, When I edit newly added list item, Then I safely return to list collector page", async () => {
```
MebinAbraham, line 324: This doesn't match the spec style or flow. The other `it` blocks are not `Given ..` styles.

I think we need to revisit this spec, I find it pretty unclear. It is doing a lot of implicit asserts. `it` blocks should end with an expect statement, but there is an expert then a bunch of actions which sometime is implicitly testing a journey, and sometime is just to progress to the next bit.

petechd: I refactored test's description to match the flow of this spec (using same conventions) but yeah, we need to revisit this to make it match all our other specs.

tests/integration/questionnaire/test_questionnaire_list_collector_repeating_blocks.py
petechd: Added a lot of descriptions so the journey should be clear, changed that specific URL to just another `post` so it's clearer as well.

katie-gardner (approved):
Functionally everything working and change to test looks good too (subject to revisiting in general) confirmed both issues mentioned are caught by the test if either change is reverted, and now pass 👍

berroar (approved):

MebinAbraham (approved):

PR #1187, opened 2023-08-17
Dockerise mock sds

katie-gardner:
This PR spins up the mock SDS service alongside runners other supporting services to enable easy deployment and functional testing with supplementary data.

### How to review
Check the requirements in the card are satisfied and that the commands in the readme files work as expected. Check the previously skipped tests are no longer skipped and check runner and launcher can talk to the mock SDS service after it has been launched with the extended `make dev-compose-up` command

### Checklist
* [ ] Functional tests using supplementary data are running and passing


README.md
```diff
@@ -115,6 +115,12 @@ make dev-compose-up-linux
 docker run -e SURVEY_RUNNER_SCHEMA_URL=http://docker.for.mac.host.internal:5000 -it -p 8000:8000 onsdigital/eq-questionnaire-launcher:latest
```
MebinAbraham, line 115: We need to set SDS API URL the same way here?.

.development.env
```diff
@@ -27,5 +27,5 @@ CDN_ASSETS_PATH=/design-system
 ADDRESS_LOOKUP_API_URL=https://whitelodge-ai-api.census-gcp.onsdigital.uk
 COOKIE_SETTINGS_URL=#
 EQ_SUBMISSION_CONFIRMATION_BACKEND=log
-SDS_API_BASE_URL=http://localhost:5003
-OIDC_TOKEN_BACKEND=local
\ No newline at end of file
+SDS_API_BASE_URL=http://sds:5003
```
MebinAbraham, line 30: The issue with updating this is that, this file can now not be used when running runner locally. This expects runner to be running in docker.

I suggest reverting this but in docker-compose.yml set `SDS_API_BASE_URL: "http://sds:5003"` inline which will take precedence. Then, if you use this file for local runner or when running in docker, both cases work.

.functional-tests.env
```diff
@@ -28,5 +28,5 @@ ADDRESS_LOOKUP_API_URL=https://whitelodge-ai-api.census-gcp.onsdigital.uk
 COOKIE_SETTINGS_URL=#
 EQ_SUBMISSION_CONFIRMATION_BACKEND=log
 VIEW_SUBMITTED_RESPONSE_EXPIRATION_IN_SECONDS=35
-SDS_API_BASE_URL=http://localhost:5003
+SDS_API_BASE_URL=http://sds:5003
```
MebinAbraham, line 31: Although this is mainly only used by action, the comment above is appropriate here also.

.development.env
katie-gardner: Ahh makes sense, have made that change and switched the dev compose files to use the same approach as the runner url

README.md
katie-gardner: Added 👍

.development.env
MebinAbraham: Doesn't look like `docker-compose-yml` was updated.

docker-compose.yml
```diff
@@ -2,47 +2,56 @@ version: '3'
...
+            - eq-env
+
+    eq-questionnaire-runner:
+        image: onsdigital/eq-questionnaire-runner:latest
+        build: ./
+        env_file:
+            - ${RUNNER_ENV_FILE}
+        environment:
+            DATASTORE_EMULATOR_HOST: datastore:8432
+            EQ_REDIS_HOST: "redis"
```
MebinAbraham, line 22: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1187#discussion_r1297461180 needs to be done here as well.

katie-gardner: Thanks 👍 added it here

.development.env
katie-gardner: oops - updated as per other comment

MebinAbraham (approved):

berroar (approved):

petechd (approved):

PR #1191, opened 2023-08-22
Update packages and version to Python 3.11

petechd:
This updates packages to their latest versions and Python to 3.11. Some version constraints were removed (flake8 and flask) as part of it. Minor tweaks were needed in order to make it compatible. Check the comments below. Github actions also updated where possible.

### How to review
Run all scripts, check if everything is working as expected. Run tests and verify that the project still functions correctly with the updated package versions. Run a benchmark to ensure the package update has no impact on performance.

### Checklist

### Performance Graph

performance_graph


petechd:
Regarding removed line from integration test: The X-XSS Protection header has been deprecated and will be replaced by the Reflected-XSS directive in the Content Security Policy.

petechd:
No newer actions runner image for linux so we need to keep using Ubuntu-22.04 and with old Python 3.10.12 version.

scripts/extract_translation_templates.py
```diff
@@ -22,8 +22,8 @@ def get_template_content(filename, ignore_context=False):
     with open(filename, encoding="UTF-8") as file:
         return list(
             filter(
-                lambda l: all(
-                    not l.startswith(param) for param in line_beginnings_to_ignore
+                lambda line: all(
```
petechd, line 25: Latest Flake8 highlights this var as too ambiguous.

petechd:
[Update Runner packages and version to Python 3.11 (2 SP)](https://trello.com/c/WFO5Lbxs/5725-update-runner-packages-and-version-to-python-311-2-sp)

katie-gardner:
> > The X-XSS Protection header, that is gone from the header in test, has been deprecated and will be replaced by the Reflected-XSS directive in the Content Security Policy (not in the headers yet).
> 
> Just leaving a comment for devs to consider. This needs some thorough checking of it to make sure it has not regressed and that the new way is enforced.

From looking at https://github.com/wntrblm/flask-talisman/pull/28/files and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection this seems sound - if a change is needed to the csp, could the policy be left on in `setup.py` until its there?

petechd:
> > > The X-XSS Protection header, that is gone from the header in test, has been deprecated and will be replaced by the Reflected-XSS directive in the Content Security Policy (not in the headers yet).
> > 
> > 
> > Just leaving a comment for devs to consider. This needs some thorough checking of it to make sure it has not regressed and that the new way is enforced.
> 
> From looking at https://github.com/wntrblm/flask-talisman/pull/28/files and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection this seems sound - if a change is needed to the csp, could the policy be left on in `setup.py` until its there?

I've added a new card to investigate it and for now set it to "enabled" so all tests and behaviour stay the same.

katie-gardner (approved):
Changes look good as does benchmark 👍

rmccar (approved):

PR #1194, opened 2023-08-29
Route to first incomplete location when returning to summary

katie-gardner:
This PR adds the functionality to return to the first incomplete block in a section when return to parameters are present and the section is incomplete. This way when returning to a section summary or final summary, if they are not immediately accessible, the user will be routed to each incomplete block in turn (and only the incomplete ones) before being returned to the section summary as soon as possible. This makes the behaviour consistent with calculated and grand calculated summaries, which already do this.

To make this work technically, the method that was being used by calculated and grand calculated summaries, is no longer called in the specific return to functions, but instead in the main `get_return_to_url` function, whenever return parameters are present and the section is not complete. (With the one exception of the grand calculated summary being in a different section, so it does need to call the function itself)

Also fixed a minor issue where the `return_to=calculated_summary` was not being removed when the user landed back on the calculated summary page

### How to review
Check that the updates to the tests thoroughly cover the scenarios on the card, and that no other unhappy path journeys have been missed. Make sure the behaviour is consistent between the different summary types

### How to test functionally
I'd suggest using the schemas 

- `list_collector_repeating_blocks_section_summary` - this has both a section summary and final summary. Check that if you edit your answers from either summary page, you return as expected
- `grand_calculated_summary_cross_section_dependencies` - If you edit answers from the summary pages in the earlier sections that affect calculated summaries. Regardless of whether you edited the answer from the calculated summary or section summary, the continue button should take you to each incomplete block only, and then back to the summary. You can use the new question `tv-block-question` to check that when you change an answer, any already completed blocks are skipped when returning to the summary.
- `new_calculated_summary_repeating_section` If you edit your answers, you should jump to incomplete blocks only before returning to the summary. 

### Checklist


katie-gardner:
[Return to section/final summary doesn't jump straight back to first incomplete location (2 SP)](https://trello.com/c/Fsr0Q9K1/5794-return-to-section-final-summary-doesnt-jump-straight-back-to-first-incomplete-location-2-sp)

berroar (approved):
Changes look good, tested functionally and all works 👍

petechd (approved):

app/questionnaire/router.py
```diff
@@ -262,7 +262,6 @@ def get_return_to_location_url(
                 return_to_block_id=return_to_block_id,
                 location=location,
                 routing_path=routing_path,
-                is_for_previous=is_for_previous,
```
MebinAbraham, line 265: How come this was removed?

katie-gardner: Not needed anymore, as I took the call to `get_url_for_inaccessible_location` out of the calculated summary function and called it in the parent function instead when the section isn't complete (and that was the only one that needed the parameter)

MebinAbraham (approved):
👍

PR #1196, opened 2023-09-01
Refactor progress store

petechd:
All the refactors required in progress store are done as part of this PR plus one extra method I could find was renamed (not mentioned on the card).
`SectionKey` type was defined in Location module because of circular import issues. It was added to Relationship Location as well, they are interchangeable in most places in the code.

### How to review
Check if nothing was missed, refactoring was done using Refactor option in IDE and double checked/manually searched through to minimise any errors.

### Checklist


petechd:
[Refactor progress store to improve clarity (1 SP)](https://trello.com/c/3Ppe2Qny/5837-refactor-progress-store-to-improve-clarity-1-sp)

Yuyuutsu (approved):
Looked through editor and see all changes made where needed. LGTM 👍

app/questionnaire/router.py
```diff
@@ -116,9 +114,7 @@ def can_display_section_summary(
     ) -> bool:
         return bool(
             self._schema.get_summary_for_section(section_id)
-        ) and self._progress_store.is_section_or_repeating_blocks_progress_complete(
-            section_id, list_item_id
-        )
+        ) and self._progress_store.is_section_complete(section_id, list_item_id)
```
berroar, line 117: How come we don't use the section key here? Looks like some more methods could be updated to use the new key? For example methods like `_get_first_incomplete_section_key` and `_get_last_complete_section_key`. Think there might be a few more too from my first check.

berroar (dismissed):

app/questionnaire/router.py
petechd: I've added all I could find and it's a lot of extra modules refactored.

app/data_models/progress_store.py
```diff
@@ -170,45 +152,39 @@ def update_section_or_repeating_blocks_progress_completion_status(
...
     ) -> str:
         """
         Return the CompletionStatus of the Section or Repeating Blocks for a list item,
         specified by the given section_id and list_item_id.
         Returns NOT_STARTED if the progress does not exist
         """
-        progress_key = SectionKey(section_id, list_item_id)
-        if progress_key in self._section_and_repeating_blocks_progress:
-            return self._section_and_repeating_blocks_progress[progress_key].status
+        progress_key = section_key
```
berroar, line 162: No need to assign this var, can just use section_key

app/data_models/progress_store.py
```diff
@@ -170,45 +152,39 @@ def update_section_or_repeating_blocks_progress_completion_status(
...
-    ) -> list[str]:
+    def get_completed_block_ids(self, section_key: SectionKey) -> list[str]:
         """
         Return the block ids recorded as part of the progress for the Section or Repeating Blocks
         for list item specified by the given section_id and list_item_id
         """
-        progress_key = SectionKey(section_id, list_item_id)
-        if progress_key in self._section_and_repeating_blocks_progress:
-            return self._section_and_repeating_blocks_progress[progress_key].block_ids
+        progress_key = section_key
```
berroar, line 185: Should the instances of progess_key in this file now be section_key?

app/data_models/progress_store.py
```diff
@@ -221,18 +197,16 @@ def add_completed_location(self, location: LocationType) -> None:
         list_item_id = location.list_item_id
 
         completed_block_ids = self.get_completed_block_ids(
-            section_id=section_id, list_item_id=list_item_id
+            SectionKey(section_id=section_id, list_item_id=list_item_id)
```
katie-gardner, line 198: Here you can use `location.section_key` instead of creating one

app/views/contexts/summary_context.py
```diff
@@ -87,7 +87,7 @@ def build_summary_item(
             response_metadata=self._response_metadata,
             current_location=location,
             routing_path=self._router.routing_path(
-                section_id, list_item_id=list_item_id
+                SectionKey(section_id=section_id, list_item_id=list_item_id)
```
katie-gardner, line 90: could use `location.section_key`

app/views/handlers/block.py
```diff
@@ -132,8 +132,10 @@ def handle_post(self) -> None:
 
     def _get_routing_path(self) -> RoutingPath:
         return self.router.routing_path(
-            section_id=self._current_location.section_id,
-            list_item_id=self._current_location.list_item_id,
+            SectionKey(
```
katie-gardner, line 135: could use `self._current_location.section_key`

app/views/handlers/list_collector_content.py
```diff
@@ -9,7 +10,9 @@ def _get_additional_view_context(self) -> dict:
     def handle_post(self) -> None:
         if self._is_list_collector_complete():
             self._routing_path = self.router.routing_path(
-                section_id=self._current_location.section_id,
-                list_item_id=self._current_location.list_item_id,
+                SectionKey(
+                    section_id=self._current_location.section_id,
```
katie-gardner, line 14: `self.current_location.section_key`

app/views/handlers/question.py
```diff
@@ -234,8 +234,10 @@ def handle_post(self) -> None:
             # In order to support progress value source references of the previous block
             self.questionnaire_store_updater.add_completed_location()
             self._routing_path = self.router.routing_path(
-                section_id=self._current_location.section_id,
-                list_item_id=self._current_location.list_item_id,
+                SectionKey(
+                    section_id=self._current_location.section_id,
+                    list_item_id=self._current_location.list_item_id,
```
katie-gardner, line 239: same again

app/data_models/progress_store.py
```diff
@@ -39,22 +38,19 @@ def __init__(
...
         self._is_routing_backwards: bool = False
-        self._section_and_repeating_blocks_progress: MutableMapping[
-            SectionKey, Progress
-        ] = self._build_map(in_progress_sections_and_repeating_blocks or [])
+        self._progress: MutableMapping[SectionKey, Progress] = self._build_map(
+            progress or []
+        )
 
     def __contains__(
         self, section_and_repeating_blocks_progress_key: SectionKey
```
katie-gardner, line 51: I think the name for this one ought to be updated to inline with the others

app/data_models/progress_store.py
```diff
@@ -121,47 +112,38 @@ def section_and_repeating_blocks_progress_keys(
...
+            return section_keys
 
         return [
             progress_key
-            for progress_key in progress_keys
+            for progress_key in section_keys
             if any(section_id in progress_key for section_id in section_ids)
         ]
 
     def update_section_or_repeating_blocks_progress_completion_status(
```
katie-gardner, line 130: update this function name too?

app/data_models/progress_store.py
petechd: Based on our conversations in tech sessions I think they are supposed to stay `progress keys`.

app/data_models/progress_store.py
petechd: I was not sure about touching it but it makes sense to remove this, done, thanks!

katie-gardner (changes requested):

app/data_models/progress_store.py
berroar: Yeah wasn't sure just slightly confusing when there are lines like `progress_key = SectionKey(section_id, list_item_id)`

app/data_models/progress_store.py
```diff
@@ -39,22 +38,19 @@ def __init__(
...
         self, section_and_repeating_blocks_progress_key: SectionKey
     ) -> bool:
-        return (
-            section_and_repeating_blocks_progress_key
-            in self._section_and_repeating_blocks_progress
-        )
+        return section_and_repeating_blocks_progress_key in self._progress
 
     @staticmethod
     def _build_map(
```
berroar, line 56: This build map function should also be updated to return a section key

app/questionnaire/router.py
```diff
@@ -137,9 +131,9 @@ def get_next_location_url(
         Get the next location in the section. If the section is complete, determine where to go next,
         whether it be a summary, the hub or the next incomplete location.
         """
-        is_section_complete = (
-            self._progress_store.is_section_or_repeating_blocks_progress_complete(
-                location.section_id, location.list_item_id
+        is_section_complete = self._progress_store.is_section_complete(
+            SectionKey(
+                section_id=location.section_id, list_item_id=location.list_item_id
```
MebinAbraham, line 136: Can use `location.section_key`

app/questionnaire/router.py
```diff
@@ -268,9 +264,9 @@ def get_return_to_location_url(
             return url
 
         if is_section_complete is None:
-            is_section_complete = (
-                self._progress_store.is_section_or_repeating_blocks_progress_complete(
-                    location.section_id, location.list_item_id
+            is_section_complete = self._progress_store.is_section_complete(
+                SectionKey(
+                    section_id=location.section_id, list_item_id=location.list_item_id
```
MebinAbraham, line 269: Can use `location.section_key`

app/questionnaire/router.py
```diff
@@ -528,25 +521,23 @@ def _get_enabled_section_keys(
...
 
     def _get_first_incomplete_section_key(self) -> tuple[str, str | None] | None:
         for section_id, list_item_id in self._get_enabled_section_keys():
-            if not self._progress_store.is_section_or_repeating_blocks_progress_complete(
-                section_id, list_item_id
+            if not self._progress_store.is_section_complete(
+                SectionKey(section_id=section_id, list_item_id=list_item_id)
             ):
-                return section_id, list_item_id
+                return SectionKey(section_id=section_id, list_item_id=list_item_id)
```
MebinAbraham, line 533: This is unpacking the key just to create it again.
This can be simplified to:
```py
        for section_key in self._get_enabled_section_keys():
            if not self._progress_store.is_section_complete(section_key):
                return section_key
```

app/questionnaire/router.py
```diff
@@ -528,25 +521,23 @@ def _get_enabled_section_keys(
...
-    def _get_last_complete_section_key(self) -> tuple[str, str | None] | None:
+    def _get_last_complete_section_key(self) -> SectionKey | None:
         for section_id, list_item_id in list(self._get_enabled_section_keys())[::-1]:
-            if self._progress_store.is_section_or_repeating_blocks_progress_complete(
-                section_id, list_item_id
+            if self._progress_store.is_section_complete(
+                SectionKey(section_id=section_id, list_item_id=list_item_id)
             ):
-                return section_id, list_item_id
+                return SectionKey(section_id=section_id, list_item_id=list_item_id)
```
MebinAbraham, line 540: Same as above, can be simplified:
```py
        for section_key in list(self._get_enabled_section_keys())[::-1]:
            if self._progress_store.is_section_complete(section_key):
                return section_key
```

MebinAbraham (commented):
Updating routing path seems to have created more code 🤔 

We should be consistent in how we call `SectionKey`, some are keyword arguments, some aren't. I think we should pick one style. Keyword arg is safer, but the latter is shortly. I don't have a real preference as long as it is consistent.

katie-gardner:
> Updating routing path seems to have created more code 🤔
> 
> We should be consistent in how we call `SectionKey`, some are keyword arguments, some aren't. I think we should pick one style. Keyword arg is safer, but the latter is shorter. I don't have a real preference as long as it is consistent. See if others do.

Also happy to roll with either. I think its quick to pick up that section key always goes section first then list item so I'd be slightly in favour of not using kwargs, and I'd also be in favour of `list_item_id` defaulting to None so it doesn't have to be manually passed in everywhere theres a non-list related section.

katie-gardner:
> > I'd also be in favour of `list_item_id` defaulting to None so it doesn't have to be manually passed in everywhere theres a non-list related section.
> 
> Can a single attribute of a namedtuple be defaulted?. Thought they both need the same default?. If it's possible, then I'm in favour.

Yep! I think it's less nice to make work with `collections.namedtuple` but for the typed one it works just like dataclasses

MebinAbraham:
> > > I'd also be in favour of `list_item_id` defaulting to None so it doesn't have to be manually passed in everywhere theres a non-list related section.
> > 
> > 
> > Can a single attribute of a namedtuple be defaulted?. Thought they both need the same default?. If it's possible, then I'm in favour.
> 
> Yep! I think it's less nice to make work with `collections.namedtuple` but for the typed one it works just like dataclasses

Ah awesome, wasn't aware of that 👍

app/questionnaire/dependencies.py
```diff
@@ -48,8 +48,9 @@ def get_routing_path_block_ids_by_section_for_dependent_sections(
         if key in sections_to_ignore:
             continue
 
-        if key in progress_store.started_section_and_repeating_blocks_progress_keys():
-            routing_path = path_finder.routing_path(*key)
+        if key in progress_store.started_section_keys():
+            # Type ignore: SectionKey from location should be used here, but cannot as the types SectionKey type is used throughout this module
+            routing_path = path_finder.routing_path(key)  # type: ignore
```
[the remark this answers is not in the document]
petechd: SectionKey is already imported from types module when here I need the one from location. Had to add it to location cause I was getting circular imports error so SectionKey is defined in two places now as the other one is imported to a lot of places.

app/questionnaire/path_finder.py
```diff
@@ -33,22 +33,22 @@ def __init__(
...
-    ) -> RoutingPath:
+    def routing_path(self, section_key: SectionKey) -> RoutingPath:
         """
         Visits all the blocks in a section and returns a path given a list of answers.
         """
         routing_path_block_ids: list[str] = []
-        current_location = Location(section_id=section_id, list_item_id=list_item_id)
-        section = self.schema.get_section(section_id)
+        current_location = Location(
+            section_id=section_key.section_id, list_item_id=section_key.list_item_id
```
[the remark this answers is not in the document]
katie-gardner: yeah `Location(*section_key)` would result in `Location(section_id=section_key.section_id, block_id=section_key.list_item_id)`
but you could unpack it with correct kwargs by doing `Location(**section_key._asdict())`

petechd:
> > Updating routing path seems to have created more code 🤔
> > We should be consistent in how we call `SectionKey`, some are keyword arguments, some aren't. I think we should pick one style. Keyword arg is safer, but the latter is shorter. I don't have a real preference as long as it is consistent. See if others do.
> 
> Also happy to roll with either. I think its quick to pick up that section key always goes section first then list item so I'd be slightly in favour of not using kwargs, and I'd also be in favour of `list_item_id` defaulting to None so it doesn't have to be manually passed in everywhere theres a non-list related section.

I will not use use kwargs and default list item id to None.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -446,15 +446,18 @@ def _evaluate_dependents(
         if is_path_complete is None:
             is_path_complete = self._router.is_path_complete(
                 self._router.routing_path(
-                    dependent_section.section_id,
-                    list_item_id=dependent_section.list_item_id,
+                    SectionKey(
+                        section_id=dependent_section.section_id,
+                        list_item_id=dependent_section.list_item_id,
+                    ),
```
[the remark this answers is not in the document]
petechd: Given the time constraints I would leave this one out.

app/questionnaire/dependencies.py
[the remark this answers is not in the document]
katie-gardner: I reckon we could drop the one from `types` and just use the one in `Location`

app/views/handlers/individual_response.py
```diff
@@ -589,12 +589,17 @@ def handle_post(self) -> Response | None:
 
     def _update_section_completeness(self) -> None:
         if not self._questionnaire_store.progress_store.get_completed_block_ids(
-            section_id=self.individual_section_id, list_item_id=self._list_item_id
+            section_key=SectionKey(
```
katie-gardner, line 592: Non blocker, if this was `section_key := ` instead of being a kwarg, you could reuse it below instead of re-defining it

app/views/handlers/question.py
```diff
@@ -304,16 +303,17 @@ def get_first_incomplete_list_repeating_block_location_for_list_item(
         list_item_id: str,
         list_name: str,
     ) -> Location | None:
-        if self._questionnaire_store.progress_store.is_section_or_repeating_blocks_progress_complete(
-            section_id=section_id, list_item_id=list_item_id
+        if self._questionnaire_store.progress_store.is_section_complete(
+            SectionKey(section_id, list_item_id)
```
katie-gardner, line 307: non blocker - could use walrus operator here and reuse below

tests/app/questionnaire/test_questionnaire_store_updater.py
```diff
@@ -1157,7 +1156,9 @@ def test_dependent_sections_completed_dependant_blocks_removed_and_status_update
     }
 
     assert dependent_block_id in progress_store.get_completed_block_ids(
-        section_id=dependent_section_key[0], list_item_id=dependent_section_key[1]
+        section_key=SectionKey(
+            section_id=dependent_section_key[0], list_item_id=dependent_section_key[1]
```
katie-gardner, line 1160: I think this could have been `SectionKey(*dependent_section_key)` same for others in the file

app/questionnaire/router.py
```diff
@@ -313,18 +303,20 @@ def _get_return_to_for_grand_calculated_summary(
...
+                not self._progress_store.is_section_complete(
+                    SectionKey(section_key.section_id)
                 )
                 or grand_calculated_summary_section not in self.enabled_section_ids
             ):
                 return None
             routing_path = self._path_finder.routing_path(
-                section_id=grand_calculated_summary_section
+                SectionKey(
+                    section_id=grand_calculated_summary_section, list_item_id=None
```
katie-gardner, line 318: minor - don't need the `= None` anymore

app/views/handlers/list_action.py
```diff
@@ -50,7 +50,9 @@ def get_section_or_final_summary_url(self) -> str | None:
         if (
             self._return_to == "section-summary"
             and self.router.can_display_section_summary(
-                self.parent_location.section_id, self.parent_location.list_item_id
+                SectionKey(
+                    self.parent_location.section_id, self.parent_location.list_item_id
```
katie-gardner, line 54: `self.parent_location.section_key`

app/views/handlers/list_action.py
```diff
@@ -67,11 +69,13 @@ def get_next_location_url(self) -> str:
...
             return url
 
-        if self.router.is_block_complete(
+        if self._questionnaire_store.progress_store.is_block_complete(
             # Type ignore: block_id would exist at this point
             block_id=self.parent_location.block_id,  # type: ignore
-            section_id=self.parent_location.section_id,
-            list_item_id=self.parent_location.list_item_id,
+            section_key=SectionKey(
+                self.parent_location.section_id,
```
katie-gardner, line 76: `self.parent_location.section_key`

katie-gardner (approved):
Couple more places we might be able to shave off a few lines or re-use a var but nothing major from me 👍

rmccar (approved):

app/questionnaire/dependencies.py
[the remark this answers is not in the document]
MebinAbraham: I didn't release we had two. We should only have one.

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
MebinAbraham: Not a fan of using the private method in this case.

app/utilities/types.py
```diff
@@ -1,11 +1,22 @@
...
 
 class SectionKey(NamedTuple):
+    section_id: str
+    list_item_id: str | None = None
+
+
+class DependentSection(NamedTuple):
     section_id: str
     list_item_id: str | None
+    is_complete: bool | None
```
MebinAbraham, line 18: These could be defaulted which would be result in around ~10 calls not needing to pass these. Left it out to minimise changes. Happy to address if reviewers are happy.

app/questionnaire/dependencies.py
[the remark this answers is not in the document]
MebinAbraham: Fixed, kept the one in `types.py` since that was the one used throughout.

app/questionnaire/questionnaire_store_updater.py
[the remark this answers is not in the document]
MebinAbraham: Fixed.

katie-gardner (commented):
On the note of completion status, I'm vaguely thinking an enum would suit it better than a dataclass, as it means we're using `status: str` everywhere where we could be using
`status: CompletionStatus` if we had instead used an enum
```
class CompletionStatus(str, Enum):
    COMPLETED = "COMPLETED"
    IN_PROGRESS = "IN_PROGRESS"
    NOT_STARTED = "NOT_STARTED"
    INDIVIDUAL_RESPONSE_REQUESTED = "INDIVIDUAL_RESPONSE_REQUESTED"
```
Not sure what thoughts are on this? If we did this and moved it to `progress.py` it could be used by progress and progress dict

MebinAbraham:
> > On the note of completion status, I'm vaguely thinking an enum would suit it better than a dataclass, as it means we're using `status: str` everywhere where we could be using `status: CompletionStatus` if we had instead used an enum
> > ```
> > class CompletionStatus(str, Enum):
> >     COMPLETED = "COMPLETED"
> >     IN_PROGRESS = "IN_PROGRESS"
> >     NOT_STARTED = "NOT_STARTED"
> >     INDIVIDUAL_RESPONSE_REQUESTED = "INDIVIDUAL_RESPONSE_REQUESTED"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Not sure what thoughts are on this? If we did this and moved it to `progress.py` it could be used by progress and progress dict (and then no need for an extra method to iterate over the values)
> 
> Ya noticed it but didn't change it here to minimise changes but happy to if reviewers are happy given this is around the refactor of it so would be a good time to do it. I'm happy to if rest are.

Alternatively, I'm happy to do the change as part of the same card but I can raise it as a separate PR to make the change obvious and easy to review. That way it doesn't block this PR.

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
katie-gardner: Came across it in [this answer](https://stackoverflow.com/questions/26180528/convert-a-namedtuple-into-a-dictionary)  mentioning its a [documented method](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict), and underscore is just for avoiding conflicts rather than indicating it being private.

That said, Its not all that readable, so I agree with not doing it - just for interest

katie-gardner:
> > > On the note of completion status, I'm vaguely thinking an enum would suit it better than a dataclass, as it means we're using `status: str` everywhere where we could be using `status: CompletionStatus` if we had instead used an enum
> > > ```
> > > class CompletionStatus(str, Enum):
> > >     COMPLETED = "COMPLETED"
> > >     IN_PROGRESS = "IN_PROGRESS"
> > >     NOT_STARTED = "NOT_STARTED"
> > >     INDIVIDUAL_RESPONSE_REQUESTED = "INDIVIDUAL_RESPONSE_REQUESTED"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Not sure what thoughts are on this? If we did this and moved it to `progress.py` it could be used by progress and progress dict (and then no need for an extra method to iterate over the values)
> > 
> > 
> > Ya noticed it but didn't change it here to minimise changes but happy to if reviewers are happy given this is around the refactor of it so would be a good time to do it. I'm happy to if rest are.
> 
> Alternatively, I'm happy to do the change as part of the same card but I can raise it as a separate PR to make the change obvious and easy to review. That way it doesn't block this PR.

Yeah makes sense to minimise crowding this one 👍  happy with that approach or in another mini followup if others agree

app/utilities/types.py
Yuyuutsu: Happy for a them to be defaulted if other devs are 👍

app/views/handlers/individual_response.py
```diff
@@ -322,11 +323,10 @@ def _render_block(self) -> dict[str, Any]:
         )
 
     def _update_section_status(self, status: str) -> None:
-        self._questionnaire_store.progress_store.update_section_or_repeating_blocks_progress_completion_status(
+        self._questionnaire_store.progress_store.update_section_status(
             # Type ignore: Current usages of this method occur when Individual Section ID exists and is not None
```
Yuyuutsu, line 327: This is not needed anymore since type ignore got removed

MebinAbraham: Done

app/questionnaire/router.py
```diff
@@ -313,18 +303,18 @@ def _get_return_to_for_grand_calculated_summary(
         grand_calculated_summary_section: str = (
             self._schema.get_section_id_for_block_id(return_to_block_id)  # type: ignore
         )
-        if grand_calculated_summary_section != location.section_id:
+        if grand_calculated_summary_section != section_key.section_id:
             # the grand calculated summary is in a different section which will have a different routing path
             # but don't go to it unless the section is enabled and the current section is complete
```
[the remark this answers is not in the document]
MebinAbraham: Fixed the grammar. Can't use `"""` at this level, that has to be the first comment i.e doc strings. At this level, it will think it is just a string.

Yuyuutsu (approved):
Looks good to me 🖖

app/utilities/types.py
katie-gardner: Agreed, also prefer that 👍 
& thanks for explaining - it seemed odd its never True but I can't see a natural True/False replacement that makes sense so see why that is - maybe just worth a comment on the class to point that out.

katie-gardner (approved):
Looking good 👍

app/utilities/types.py
MebinAbraham: Addressed in [`f0e62fb` (#1196)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1196/commits/f0e62fb583d789b08cb17aeeb0108f8c81351d6a) & [`8428314` (#1196)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1196/commits/8428314c97ea119fb96dc962f34926f928f6c5bd)

MebinAbraham:
Some more uses of the `section_key` property can be used, which has been done in [`6bd90e5` (#1196)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1196/commits/6bd90e5f3278c43409f903e6ddceb180798074de)

Yuyuutsu (approved):

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
MebinAbraham: @katie-gardner @Yuyuutsu There are a few places where unpacking would be neater, what if we added our own public `asdict()` wrapper?.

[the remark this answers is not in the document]
katie-gardner: Sure - I'd be all for that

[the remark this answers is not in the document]
Yuyuutsu: Yeah that sounds good

app/views/handlers/question.py
```diff
@@ -262,8 +259,8 @@ def capture_dependent_sections_for_list(self, list_name: str) -> None:
                 continue
             self.questionnaire_store_updater.dependent_sections.add(
                 DependentSection(
-                    section_id=section_id,
-                    list_item_id=list_item_id,
+                    section_id,
+                    list_item_id,
                     is_complete=None,
```
katie-gardner, line 264: can drop this `is_complete`

app/questionnaire/path_finder.py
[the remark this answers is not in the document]
MebinAbraham: Done in [`b1f5f77` (#1196)](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1196/commits/b1f5f77c9d66c50882f668f7a4c9607eccc4dd06)

katie-gardner (approved):
🎉

Yuyuutsu (approved):

PR #1198, opened 2023-09-06
Add validation for supplementary data lists property

katie-gardner:
This PR adds the runner validation that checks any list defined in a schema, is either populated by a list collector with add/edit/remove blocks, or it is populated by supplementary data. [The corresponding validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/213) ensures the schema lists are populated by a list collector or in the `supplementary_data.lists` property. This then ensures it is in the actual supplementary data fetched.

### How to review
Check the validation is working as expected and covered by tests. Either running the mock SDS locally and changing the datasets to not contain required lists, or editing `test_supplementary_data` to require different lists for the repeating sections/dynamic answers etc, check that an error occurs whenever lists are missing

### Checklist


katie-gardner:
[Support for repeating sections / list collector content blocks without a list collector (2 SP)](https://trello.com/c/cyFoALKE/5786-support-for-repeating-sections-list-collector-content-blocks-without-a-list-collector-2-sp)

app/questionnaire/questionnaire_schema.py
```diff
@@ -135,10 +147,18 @@ def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
...
     def min_and_max_map(self) -> ImmutableDict[str, ImmutableDict[str, int]]:
         return make_immutable(self._min_and_max_map)  # type: ignore
 
+    @cached_property
+    def lists_populated_by_list_collector(self) -> frozenset[str]:
+        return make_immutable(self._lists_populated_by_list_collector)  # type: ignore
+
+    @cached_property
+    def schema_dependent_lists(self) -> frozenset[str]:
+        return make_immutable(self._schema_dependent_lists)  # type: ignore
```
MebinAbraham, line 160: Did you not bother with the type ignore comment since the first property explains it or was it missed?.

app/questionnaire/questionnaire_schema.py
```diff
@@ -106,6 +116,8 @@ def __init__(
         )
         self._list_names_by_list_repeating_block_id: dict[str, str] = {}
         self._repeating_block_answer_ids: set[str] = set()
+        self._lists_populated_by_list_collector: set[str] = set()
+        self._schema_dependent_lists: set[str] = set()
```
MebinAbraham, line 120: I'm not sure what `schema_dependent_lists` actually mean?. It seems to indicate more than it is. Maybe that is just me. We are already in the context of the schema, so can this just be like `list_names` or something similar?. The `dependent` bit confuses me and makes me think there is more to it?.

app/questionnaire/questionnaire_schema.py
```diff
@@ -933,9 +952,9 @@ def get_add_block_for_list_collector(
                 "ListCollector": "add_block",
                 "PrimaryPersonListCollector": "add_or_edit_block",
             }
-            add_block: ImmutableDict = list_collector[
-                add_block_map[list_collector["type"]]
-            ]
+            add_block: ImmutableDict | None = list_collector.get(
+                add_block_map.get(list_collector["type"])
+            )
```
MebinAbraham, line 957: 👍

app/routes/session.py
```diff
@@ -169,9 +175,29 @@ def _set_questionnaire_supplementary_data(
...
+def _validate_supplementary_data_lists(
+    *, supplementary_data: dict, schema: QuestionnaireSchema
+) -> None:
+    """
+    Validates that any lists the schema requires (dynamic answers, list collector content blocks, repeating sections etc.) are covered:
+    either by a supplementary data list, or a list populated by a list collector
+    """
+    supplementary_lists = set(supplementary_data["data"].get("items", {}).keys())
+    populated_lists = supplementary_lists | schema.lists_populated_by_list_collector
+    if not schema.schema_dependent_lists <= populated_lists:
```
MebinAbraham, line 194: I might be wrong here but I'm not following why we do this.
The schema will have a `supplementary_data.lists` property, so don't we just need to check `supplementary_data.lists` exists within `supplementary_lists`?. Everything else is checked by validator?

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=support-supplementary-lists-property
```

app/questionnaire/questionnaire_schema.py
MebinAbraham: See if it is needed after https://github.com/ONSdigital/eq-questionnaire-runner/pull/1198#discussion_r1317457814

app/utilities/make_immutable.py
```diff
@@ -7,6 +7,8 @@
 def make_immutable(data: Any) -> Any:
     if isinstance(data, abc.Hashable):
         return data
+    if isinstance(data, set):
+        return frozenset(data)  # set items must be hashable
```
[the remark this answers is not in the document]
katie-gardner: Ah good point - though I've now removed the change since I don't need it anymore and won't get the coverage

app/routes/session.py
```diff
@@ -139,11 +141,15 @@ def login() -> Response:
...
-    questionnaire_store: QuestionnaireStore, metadata: MetadataProxy
+    *,
+    questionnaire_store: QuestionnaireStore,
+    metadata: MetadataProxy,
+    schema: QuestionnaireSchema,
 ) -> None:
     """
     If the survey metadata has an sds dataset id, and it either doesn't match what it stored, or there is no stored supplementary data
     then fetch it and add it to the store
+    This includes verification that the supplementary data lists cover any schema dependent lists which aren't populated by a list collector
```
berroar, line 152: Do we want to update this doc string now you removed that `schema_dependent_lists` var?

berroar (approved):
Looks good, tested functionally 👍

app/routes/session.py
katie-gardner: Yeah good idea - have amended that comment and shortened it a bit

app/routes/session.py
```diff
@@ -169,9 +174,27 @@ def _set_questionnaire_supplementary_data(
...
         sds_dataset_id=new_sds_dataset_id,
     )
+    # ensure any required lists for the schema are included in the supplementary data
+    _validate_supplementary_data_lists(
+        supplementary_data=supplementary_data, schema=schema
+    )
     questionnaire_store.set_supplementary_data(supplementary_data["data"])
 
 
+def _validate_supplementary_data_lists(
```
MebinAbraham, line 184: I think we should do the validation when `validate_supplementary_data_v1` runs. Thoughts?

MebinAbraham (commented):
Looks good, just a minor suggestion.

MebinAbraham (approved):

app/routes/session.py
MebinAbraham: Happy this is resolved.

app/routes/session.py
katie-gardner: Yep 👍  - agree its better there, and makes the testing better too, moved it all over

app/services/supplementary_data.py
```diff
@@ -53,9 +57,7 @@ def get_supplementary_data_v1(
 
     supplementary_data_url = f"{current_app.config['SDS_API_BASE_URL']}/v1/unit_data"
 
-    parameters = {"dataset_id": dataset_id, "identifier": identifier}
-
-    encoded_parameters = urlencode(parameters)
+    encoded_parameters = urlencode({"dataset_id": dataset_id, "identifier": identifier})
```
katie-gardner, line 60: pylint wanted fewer local vars

app/routes/session.py
```diff
@@ -139,11 +141,14 @@ def login() -> Response:
...
+    questionnaire_store: QuestionnaireStore,
+    metadata: MetadataProxy,
+    schema: QuestionnaireSchema,
 ) -> None:
     """
     If the survey metadata has an sds dataset id, and it either doesn't match what it stored, or there is no stored supplementary data
-    then fetch it and add it to the store
+    then fetch it, verify any schema supplementary lists are included in the fetched data, and add it to the questionnaire store
     """
     if not (new_sds_dataset_id := metadata["sds_dataset_id"]):
```
MebinAbraham, line 153: Just noticed a problem. Since we do the validation after this check, if the schema list changes, but the datasets hadn't, we wouldn't validate again, hence it could be invalid but still pass our checks. Edge and would only happen if they update mid-exercise. i.e Author updates the schema, but the dataset id hasn't changed yet and a new list was added/renamed.

katie-gardner: ahh yes... won't be able to avoid that with the list validation in the parser - could always revert the last commit so its called in `_set_questionnaire_supplementary_data` and additionally validate in the cases of no data / `existing == new` ?

MebinAbraham: Ya we'll need to cover this just in case.

app/routes/session.py
katie-gardner: Now reverted as per other comment

app/routes/session.py
katie-gardner: Have made the switch, will now validate every launch regardless of new data or not

app/routes/session.py
```diff
@@ -139,23 +141,33 @@ def login() -> Response:
...
     )
 
-    if existing_sds_dataset_id == new_sds_dataset_id:
-        # no need to fetch again
+    if (
+        not (new_sds_dataset_id := metadata["sds_dataset_id"])
+        or existing_sds_dataset_id == new_sds_dataset_id
+    ):
+        # no need to fetch: either no supplementary data or it hasn't changed, just validate lists
+        _validate_supplementary_data_lists(
```
MebinAbraham, line 167: Longer term we can check if the cir instrument id changed as well (i.e schema has changed)

app/routes/session.py
```diff
@@ -169,9 +181,26 @@ def _set_questionnaire_supplementary_data(
...
 
 
+def _validate_supplementary_data_lists(
+    *, supplementary_data: dict, schema: QuestionnaireSchema
+) -> None:
+    """
+    Validates that any lists the schema requires (which are those in the supplementary_data.lists property)
+    are included in the supplementary data
+    """
+    supplementary_lists = set(supplementary_data.get("items", {}).keys())
```
MebinAbraham, line 197: Minor, either drop `set` as `keys` returns a view object which support set operations. Alternatively, drop `keys()` and keep `set` as in Py3 you don't need to say `keys()` in this case.

app/routes/session.py
```diff
@@ -169,9 +181,26 @@ def _set_questionnaire_supplementary_data(
...
+    if missing := schema.supplementary_lists - supplementary_lists:
+        raise ValidationError(
+            f"Supplementary data does not include the following lists required for the schema: {', '.join(missing)}"
+        )
```
MebinAbraham, line 201: If it is keys in schema that aren't in SDS, then this needs to be flipped?.

app/routes/session.py
katie-gardner: Oh nice - didn't realise keys would support it out the box, done 👍

app/routes/session.py
katie-gardner: I think I've got this the right way round? missing is { any supplementary lists the schema requires that are not in the actual supplementary data lists }

katie-gardner: If it looks wrong from the naming happy to modify

MebinAbraham: Nope, all good. I misread 👍

MebinAbraham (approved):

PR #1210, opened 2023-09-20
Grand calculated summary inside a repeating section

katie-gardner:
This PR adds support for using a Grand Calculated Summary within a repeating section. Within a repeating section a grand calculated summary is allowed to reference other calculated summaries within the repeating section, or non-repeating calculated summaries from other sections.

The [corresponding validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/215) ensures that a GCS can only reference a repeating CS if both reside in the same repeating section

(Note that the schema is actually valid without the validator changes, as those have added restrictions rather than enabled something new, however this PR is pointing at the validator branch to demo that the validator changes are compatible with these ones)

**For this iteration, a repeating GCS using non-repeating calculated summaries with repeating answers is not supported.**

### How to review
It was decided that the refactor of `return_to` parameters into a class would be done separately in a follow on ticket, so it is not a separate piece of work here but to simplify review, the three main commits are

1. [Add return to list parameters](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1210/commits/5129aa086918e4669d7d2c7c07f23ef1fd7e595c)
2. [Changes required to support grand calculated summaries in repeats](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1210/commits/a52c2176ab2a8b704355e3bc33c8155006f1960b)
3. [Functional tests](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1210/commits/bd89e2bc06d794b0243136b53e8db3d45dfe12ee)

the `return_to` parameters have been extended to include `return_to_list_item_id` and `return_to_list_name` so that when using the change links to route to a non-repeating calculated summary and its answers, the continue button has sufficient information to return to the repeating section at the correct list item. These needed to filter through to answer/block/question/calculated summary contexts and handlers

### How to test functionally
Use the `test_grand_calculated_summary_inside_repeating_section` schema and follow the guidance in the first section. Namely, do not add items to the list, that functionality is in the follow on card. But do check that the change links from the repeating sections for vehicles work as expected, you can change answers in both the repeating and non-repeating calculated summaries and return to the correct location, and that progress updates as expected when updating answers in the first section.

### Checklist


katie-gardner:
[Grand Calculated Summary: Support within repeating sections (5 SP)](https://trello.com/c/gP4302Vl/5733-grand-calculated-summary-support-within-repeating-sections-5-sp)

app/views/contexts/summary/calculated_summary_block.py
```diff
@@ -30,20 +31,34 @@ def __init__(
...
 
         self.id = block_schema["id"]
         self.title = block_schema["calculation"]["title"]
         self._return_to = return_to
         self._return_to_block_id = return_to_block_id
+        self._return_to_list_name = return_to_list_name
+        self._return_to_list_item_id = return_to_list_item_id
         self._block_schema = block_schema
         self._schema = schema
+        if location and self._schema.is_block_in_repeating_section(self.id):
```
berroar, line 56: Do we need the check for location here? Can it be None?

katie-gardner: good spot - it can't, removing 👍

berroar (approved):
Tested, all looks good to me 👍

app/views/handlers/block.py
```diff
@@ -44,6 +44,8 @@ def __init__(
         self._return_to = request_args.get("return_to")
         self._return_to_answer_id = request_args.get("return_to_answer_id")
         self._return_to_block_id = request_args.get("return_to_block_id")
+        self._return_to_list_name = request_args.get("return_to_list_name")
```
MebinAbraham, line 47: Do we actually need to pass this?. This is deterministic if we have `return_to_list_item_id`. We should avoid adding anything to the URL that we don't strictly need. From a quick scan, the only place we actually need list name is `_get_return_to_for_grand_calculated_summary`, but in there, that can be worked out dynamically as you have access to the list store in router.py. You just need a map in list store for list item id to list name look up.

app/views/contexts/calculated_summary_context.py
```diff
@@ -97,8 +103,14 @@ def build_view_context(self) -> dict[str, dict]:
         block_id: str = self.current_location.block_id  # type: ignore
         block: ImmutableDict = self._schema.get_block(block_id)  # type: ignore
 
-        calculated_section: dict = self._build_calculated_summary_section(block)
-        calculation = block["calculation"]
+        rendered_block = self._placeholder_renderer.render(
+            data_to_render=block, list_item_id=self.current_location.list_item_id
+        )
```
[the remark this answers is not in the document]
katie-gardner: Sure, I like this approach, good idea 👍

app/views/handlers/block.py
katie-gardner: Good point, we don't, with the lookup in the list store it can indeed be removed, so have done that. I went with going through the lists as think it should be ok in this specific case

app/views/contexts/calculated_summary_context.py
```diff
@@ -116,24 +128,24 @@ def build_view_context(self) -> dict[str, dict]:
...
     def _build_formatted_summary(
         self,
         *,
-        block: ImmutableDict,
         groups: Iterable[Mapping],
-        calculation: ImmutableDict,
+        calculation: Mapping,
         formatted_total: str,
         summary_type: str,
+        block_type: str,
```
MebinAbraham, line 145: Can we not remove both these types and instead use `self.rendered_block["type"]`, `yes block type is hyphen separated but could be handled dynamically instead of passing them in?.

app/data_models/list_store.py
```diff
@@ -178,6 +178,12 @@ def _list_item_ids(self) -> list[str]:
 
         return ids
 
+    def get_list_name_for_list_item_id(self, list_item_id: str) -> str:
+        for list_name in self._lists:
+            if list_item_id in self[list_name].items:
+                return list_name
+        raise ValueError(f"list_item_id {list_item_id} not found in any lists")
```
MebinAbraham, line 185: If I mess with the url and change id to be invalid, this will result in a 500 error. We could raise 404, but I think we should just return None and in any place where None isn't acceptable, it can raise an exception.

app/views/contexts/summary/calculated_summary_block.py
```diff
@@ -74,9 +87,12 @@ def _build_link(self) -> str:
         return url_for(
             "questionnaire.block",
             block_id=self.id,
+            list_name=self._list_name,
+            list_item_id=self._list_item_id,
             return_to=self._return_to,
             return_to_answer_id=self.id,
             return_to_block_id=self._return_to_block_id,
+            return_to_list_item_id=self._return_to_list_item_id,
```
MebinAbraham, line 95: If `return_to_list_item_id` is the same as `list_item_id` i.e we are changing an answer in the same repeating section then we shouldn't add this to the url, as it is not needed.

MebinAbraham (commented):
Couple of comments otherwise LGTM.

app/views/contexts/calculated_summary_context.py
katie-gardner: Sure - made it a separate string method so it can always be picked up elsewhere if needed

app/data_models/list_store.py
katie-gardner: Done 👍

app/views/contexts/summary/calculated_summary_block.py
katie-gardner: Yeah that seems sensible, altered it here to only use the param if not in a repeating section, and made the change to router so we don't use it now inside the same repeat

MebinAbraham:
@katie-gardner ignore.

app/questionnaire/router.py
```diff
@@ -299,34 +309,44 @@ def _get_return_to_for_grand_calculated_summary(
...
             # the grand calculated summary is in a different section which will have a different routing path
             # but does not go to it unless the section is enabled and the current section is complete
             if (
-                not self._progress_store.is_section_complete(
-                    SectionKey(section_key.section_id)
-                )
+                not self._progress_store.is_section_complete(section_key)
                 or grand_calculated_summary_section not in self.enabled_section_ids
             ):
                 return None
```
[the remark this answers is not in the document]
katie-gardner: Thanks for spotting this - have added an extra question depending on that first calculated summary so this can be manually & functionally tested and passed it to that extra method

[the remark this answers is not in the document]
MebinAbraham: We need to look at this wider as anchor/return to answer id is another culprit. Once we add a `ReturnLocation` it should because easily to spot and not miss.

MebinAbraham (approved):
👍

PR #1214, opened 2023-09-26
Add type hints to fields module

liamtoozer:
* Added type hinting to the `fields` module to align with other modules with type hinting in place. 
* Updates to field handlers to include optional return types for `UnboundField`
* Small tweaks to some unit tests to include enforced keyword args and more concrete types

### How to review
Check to see if the type hints added to the module make sense, and any advice on how to improve would be great

### Checklist

[Trello](https://trello.com/c/hFiUlVrZ)


app/forms/fields/year_date_field.py
```diff
@@ -26,11 +28,28 @@ def data(self):
...
+        **kwargs: Any,
+    ):
         form_class = get_form_class(validators)
-        super().__init__(form_class, **kwargs)
-
-    def process(self, formdata, data=None, extra_filters=None):
+        super().__init__(
+            form_class,
+            label=label,
+            description=description,
```
katie-gardner, line 43: How did you decide which kwargs to pull out of the dictionary and type over which ones to leave? just wondering since it differs between them

app/forms/fields/year_date_field.py
```diff
@@ -26,11 +28,28 @@ def data(self):
 
 
 class YearDateField(FormField):
-    def __init__(self, validators, **kwargs):
+    def __init__(
+        self,
+        *,
+        validators: Sequence[Callable],
+        label: str | None = None,
+        description: str,
```
katie-gardner, line 36: All the tests with this have it as an empty string, it could be `str | None = None` same as label since the library defaults it but im wondering if this was for a specific reason and if so all good

app/forms/fields/decimal_field_with_separator.py
```diff
@@ -15,13 +16,25 @@ class DecimalFieldWithSeparator(DecimalField):
...
+            **kwargs,
+        )
         self.data = None
 
-    def process_formdata(self, valuelist):
+    def process_formdata(self, valuelist: Sequence[str] | None = None) -> None:
         if valuelist:
             try:
-                self.data = Decimal(sanitise_number(valuelist[0]))
+                self.data = Decimal(sanitise_number(valuelist[0]))  # type: ignore # None type is safeguarded in if statement above
```
katie-gardner, line 38: same as my other comment about the type ignore, can be fixed in the init with `self.data: int | None = None`

app/forms/field_handlers/select_handlers.py
```diff
@@ -8,13 +7,9 @@
...
 from app.questionnaire.questionnaire_schema import InvalidSchemaConfigurationException
+from app.utilities.types import Choice, ChoiceWithDetailAnswer
 
-Choice = namedtuple("Choice", "value label")
-ChoiceWithDetailAnswer = namedtuple(
-    "ChoiceWithDetailAnswer", "value label detail_answer_id"
-)
-
-ChoiceType = Union[Choice, ChoiceWithDetailAnswer]
+ChoiceType: TypeAlias = Choice | ChoiceWithDetailAnswer
```
MebinAbraham, line 12: Worth moving this one to `types.py` also.

mypy.ini
```diff
@@ -51,6 +51,11 @@ disallow_untyped_defs = True
 warn_return_any = True
 no_implicit_optional = True
 
+[mypy-app.forms.fields.*]
```
MebinAbraham, line 54: Since all of the forms folder is now types, you can remove all the `form.` rules and replace it with a single:

```ini
[mypy-app.forms.*]
disallow_untyped_defs = True
warn_return_any = True
no_implicit_optional = True
```

katie-gardner (commented):
Looking good 👍 added a few initial comments with a couple of questions/ideas

app/forms/field_handlers/date_handlers.py
```diff
@@ -53,7 +53,9 @@ def validators(self) -> DateValidatorTypes:
         return validate_with
 
     def get_field(self) -> DateField:
-        return DateField(self.validators, label=self.label, description=self.guidance)
+        return DateField(
```
katie-gardner, line 56: I've just noticed `Field` from wtforms has a `__new__` method overriding the class with unbound field whenever `_form` isn't passed in... so this is going to return `UnboundField` and similar for other field types too 👀

app/forms/fields/year_date_field.py
MebinAbraham: Since we haven't removed `**kwargs` I don't think it is worth pulling these out, the kwargs keep the field extensible. If we were typing each one, then the kwargs don't provide any value. I would learn towards keeping the field generic and just using kwargs. See what others think.

app/forms/fields/select_field_with_detail_answer.py
```diff
@@ -27,11 +52,11 @@ def __iter__(self):
             opt.checked = checked
```
katie-gardner, line 52: `opt.process(None, value)` implies the `formdata` parameter elsewhere can also be None so it might be `MultiDict | None` where its defined

app/forms/fields/month_year_date_field.py
```diff
@@ -27,11 +29,28 @@ def data(self):
...
+            form_class,
+            label=label,
+            description=description,
+            **kwargs,
+        )
+
+    def process(
+        self,
+        formdata: MultiDict,
+        data: UnsetValue = unset_value,
```
MebinAbraham, line 51: data can be str or unsetvalue. For some reason mypy isn't picking up some cases.

app/forms/fields/date_field.py
```diff
@@ -1,13 +1,15 @@
...
+from werkzeug.datastructures import MultiDict
 from wtforms import Form, FormField, StringField
-from wtforms.utils import unset_value
+from wtforms.utils import UnsetValue, unset_value
 
 logger = logging.getLogger(__name__)
 
 
-def get_form_class(validators):
+def get_form_class(validators: Sequence[Callable]) -> Form:
```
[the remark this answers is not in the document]
MebinAbraham: The usage seems to be getting past mypy but IDE seems to pick it up 🤔

app/forms/fields/date_field.py
```diff
@@ -30,11 +32,28 @@ def data(self):
 
 
 class DateField(FormField):
-    def __init__(self, validators, **kwargs):
+    def __init__(
+        self,
+        *,
+        validators: Sequence[Callable],
```
MebinAbraham, line 38: Instead of using generic Callable, use our types where possible. There is validator types variables in the handlers files, for example here it could be `DateValidatorTypes` from `date_handlers.py`. You will most likely need to move them to `types.py` to avoid circular imports.

Applies to other fields also.

app/forms/fields/year_date_field.py
```diff
@@ -26,11 +28,28 @@ def data(self):
...
+            label=label,
+            description=description,
+            **kwargs,
+        )
+
+    def process(
+        self,
+        formdata: MultiDict,
+        data: UnsetValue = unset_value,
+        extra_filters: None = None,
```
katie-gardner, line 51: Documentation suggests `extra_filters` is a list of filter functions so I think this one may be `Sequence[Callable] | None` similar to some others

app/forms/validators.py
```diff
@@ -127,7 +127,7 @@ def __call__(
         form: "QuestionnaireForm",
         field: Union[DecimalFieldWithSeparator, IntegerFieldWithSeparator],
     ) -> None:
-        value: Union[int, Decimal] = field.data
+        value: Union[int, Decimal, None] = field.data
```
MebinAbraham, line 130: While we are here, update the use of `Union` in favour of `|`

app/forms/fields/year_date_field.py
katie-gardner: I would lean that way as well, unless specifically validating one/ its type creates issues, I think it's fine staying as kwargs

app/forms/field_handlers/date_handlers.py
liamtoozer: Just to confirm from conversations in the team - we'll update each of the 'getters' in the handlers module to return `UnboundField` type as well. Thanks for your help with this!

app/forms/fields/date_field.py
liamtoozer: Ah nice that makes sense, thanks 👍

app/forms/fields/year_date_field.py
liamtoozer: Thanks both 👍  Just to confirm after a quick team discussion, we've gone with the preferred way of keeping `kwargs`  for caller flexibility with the third party library, and only extracting a `kwarg` if the scope of the class calls for it. I think I've updated it in places where this should be the case now?

mypy.ini
liamtoozer: Updated, thanks 👍

MebinAbraham (approved):
👍

petechd (approved):
👍

katie-gardner (approved):
🎉

PR #1215, opened 2023-09-26
Feat: Grand calculated summary value sources

katie-gardner:
This PR adds support for using Grand Calculated Summaries as value sources. It goes alongside the [corresponding validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/216) which allows for this within a schema.

To do this, `value_source_resolver` has been updated and the existing method `_resolve_calculated_summary_value_source` has been used for grand calculated summaries too, as it worked directly out of the box without GCS needing its own. For clarity, some parameters have been renamed

placeholder transforms also needed to be updated so the custom logic for decimal places applies on a non grand calculated summary page that references a grand calculated summary value

Two schemas updated with a GCS inside and outside of a repeat to cover both cases, and functional tests updated accordingly 

## Note
As with validator, this builds on the [PR for GCS in repeating sections](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1210), so the PR has been pointed at main for the test suite, and now pointed at that branch for review, it will be merged into main on approval

### How to review
Use `test_grand_calculated_summary_inside_repeating_section.json` and `test_grand_calculated_summary_repeating_answers.json` to test and make sure the grand calculated summary values are piped in correctly. In particular on the repeating answers one, make sure that if you remove calculated summaries from the path using the options in section 2 that the GCS being used as a value source is correctly updated.

### Checklist


katie-gardner:
[Grand Calculated Summary: Ability to use as a Value source (3 SP)](https://trello.com/c/QG98iwvm/5585-grand-calculated-summary-ability-to-use-as-a-value-source-3-sp)

app/questionnaire/placeholder_transforms.py
```diff
@@ -73,6 +77,14 @@ def _get_decimal_limit(self, unresolved_arguments: Mapping) -> int | None:
             decimal_limit = self.schema.get_decimal_limit_from_calculated_summaries(
                 [identifier]
             )
+        elif source == "grand_calculated_summary":
+            # Type ignore: Validator will have checked the id so the block is guaranteed to exist
+            grand_calculated_summary_block: ImmutableDict = self.schema.get_block(identifier)  # type: ignore
```
petechd, line 82: Why do we cast the type here? To avoid type ignore on line 85?

app/questionnaire/value_source_resolver.py
```diff
@@ -319,8 +319,8 @@ def resolve(
     ) -> ValueSourceEscapedTypes | ValueSourceTypes:
         source = value_source["source"]
 
-        if source == "calculated_summary":
-            return self._resolve_calculated_summary_value_source(
+        if source in {"calculated_summary", "grand_calculated_summary"}:
```
MebinAbraham, line 322: Unit test needs to be added to ValueSourceResolver to test GCS.

MebinAbraham (commented):
I see that "dependencies" weren't handled, was that on purpose?. Right now, GCS values that aren't on the path would still be used. If you check for the reference to `calculated_summary` in the code base, you can see all the places we need to update to handle dependencies.

However, regardless, we should do it in a separate PR/card, to make the review easier. The card does mention only handling GCS values that are on the path, however we probably didn't I think it through and the sizing on the card should be higher even if tackled in a separate PR.

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=grand-calculated-summary-value-source
```

MebinAbraham (approved):
Happy with the initial build.

Note for other reviewers: Dependencies have not been implemented in this PR, so you may get an answer not on the path, and validation won't get revisted etc.

app/questionnaire/value_source_resolver.py
```diff
@@ -230,20 +230,20 @@ def _resolve_list_value_source(self, value_source: Mapping) -> int | str | list:
...
         """
-        calculated_summary_block: ImmutableDict = self.schema.get_block(value_source["identifier"])  # type: ignore
-
-        if not self._is_block_on_path(calculated_summary_block["id"]):
+        summary_block: ImmutableDict = self.schema.get_block(value_source["identifier"])  # type: ignore
+        if not self._is_block_on_path(summary_block["id"]):
             return None
 
-        calculation = calculated_summary_block["calculation"]
+        calculation = summary_block["calculation"]
```
petechd, line 245: This confused me a little bit at first. Can you change it to something like `# old calculated summary block's calculation object may have answers_to_calculate instead of operation`? Or am I getting it wrong?

app/questionnaire/value_source_resolver.py
```diff
@@ -230,20 +230,20 @@ def _resolve_list_value_source(self, value_source: Mapping) -> int | str | list:
...
 
-    def _resolve_calculated_summary_value_source(
+    def _resolve_summary_with_calculation(
         self, value_source: Mapping, *, assess_routing_path: bool
     ) -> IntOrDecimal | None:
-        """Calculates the value for the 'calculation' used by the provided Calculated Summary.
+        """Calculates the value for the 'calculation' used by the provided Calculated or Grand Calculated Summary.
 
-        The caller is responsible for ensuring the provided Calculated Summary and its answers are on the path,
+        The caller is responsible for ensuring the provided Summary and its components are on the path
```
petechd, line 238: Summary is not really an actual block type, so shouldn't be capitalised? I can only find details like this, otherwise the code changes look good.

app/questionnaire/value_source_resolver.py
katie-gardner: Yep good spot, it should be operation instead of calculation, have updated the comment so it hopefully is a bit clearer?

petechd: Yeah, thanks. It wasn't some big issue, non-blocking change really.

petechd (approved):

MebinAbraham (approved):

PR #1219, opened 2023-09-28
Add calculated summary negative demo

VirajP1002:

This PR demonstrates that Runner is able to handle negative values. Within the calculated summary schema, a minimum value was added to each of the answers to provide compatibility for handling this.

### How to review

- Run functional tests to ensure all they pass
- Check the changes in the schema

### Checklist


tests/functional/spec/features/calculated_summary/calculated_summary_test_case.js
```diff
@@ -422,6 +422,51 @@ class TestCase {
...
+
+      await $(SixthNumberBlockPage.sixthPercent()).setValue(23);
+      await $(SixthNumberBlockPage.sixthNumber()).setValue(45);
+      await click(SixthNumberBlockPage.submit());
+
+      await expect(browser).toHaveUrlContaining(CurrencyTotalPlaybackPage.pageName);
+    });
+    it("Given I have a range of positive and negative values to enter, when I enter the negative values, then the total of the summary should be correct", async () => {
+      await expect(await $(CurrencyTotalPlaybackPage.firstNumberAnswer()).getText()).toContain(expectedAnswerValue);
+      await expect(await $(CurrencyTotalPlaybackPage.calculatedSummaryTitle()).getText()).toContain(
```
katie-gardner, line 463: This assertion actually checks the 3 substrings separately, "We calculate..." and expectedTotal and "Is this correct"
I think it would be better to check a single formatted string
```
await expect(await $(CurrencyTotalPlaybackPage.calculatedSummaryTitle()).getText()).toContain(
 `We calculate the total of currency values entered to be ${expectedTotalValue}. Is this correct?`
);
```
and then change the test to pass in the expectedTotalValue as "-£6.00" rather than "-6"

tests/functional/spec/features/calculated_summary/calculated_summary.spec.js
```diff
@@ -15,3 +15,18 @@ describe("Feature: Calculated Summary", () => {
     CalculatedSummaryTestCase.testCrossSectionDependencies("test_new_calculated_summary_cross_section_dependencies.json");
   });
 });
+
+describe("Feature: Calculated Summary with negative values", () => {
+  describe("Given I enter a negative value in the first section", () => {
+    CalculatedSummaryTestCase.testNegative("test_calculated_summary.json", -1, 2, 3, 0, 4, "-£1.00");
```
katie-gardner, line 21: I don't think you need to include verifying `-£1.00` in the test as its not part of the summary, just a single answer, I think the 4 answers and expected total suffice  (would change the total to be a formatted string though, see below)

tests/functional/spec/features/calculated_summary/calculated_summary_test_case.js
```diff
@@ -422,6 +422,51 @@ class TestCase {
       );
     });
   }
+
+  testNegative(schema, firstAnswerValue, secondAnswerValue, thirdAnswerValue, fourthAnswerValue, expectedTotalValue, expectedAnswerValue) {
```
katie-gardner, line 426: Paired with my comment above, I think you can remove the `expectedAnswerValue` as a param

tests/functional/spec/features/calculated_summary/calculated_summary_test_case.js
```diff
@@ -422,6 +422,51 @@ class TestCase {
...
+      await $(FifthNumberBlockPage.fifthNumber()).setValue(78.91);
+      await click(FifthNumberBlockPage.submit());
+
+      await $(SixthNumberBlockPage.sixthPercent()).setValue(23);
+      await $(SixthNumberBlockPage.sixthNumber()).setValue(45);
+      await click(SixthNumberBlockPage.submit());
+
+      await expect(browser).toHaveUrlContaining(CurrencyTotalPlaybackPage.pageName);
+    });
+    it("Given I have a range of positive and negative values to enter, when I enter the negative values, then the total of the summary should be correct", async () => {
```
katie-gardner, line 461: I might rephrase as "Given I have entered a range of positive and negative values, When I reach the calculated summary, Then the total is correct" since this particular `it` is after you've entered them

schemas/test/en/test_calculated_summary.json
```diff
@@ -51,7 +51,11 @@
                                         "mandatory": true,
                                         "type": "Currency",
                                         "currency": "GBP",
-                                        "decimal_places": 2
+                                        "decimal_places": 2,
```
katie-gardner, line 54: Looks good 👍 I think it would be worth making this same change to `test_new_calculated_summary` so that they are still in sync, and then maybe have half your test cases use the new variant too

katie-gardner (commented):
Looking good! a few minor comments

schemas/test/en/test_calculated_summary.json
VirajP1002: Thanks Katie, with regards to adding the changes into `test_new_calculated_summary`, would I write my tests in the same file as I'm slightly unsure about the second part of the comment

tests/functional/spec/features/calculated_summary/calculated_summary_test_case.js
VirajP1002: That makes sense, I can add that in.

tests/functional/spec/features/calculated_summary/calculated_summary.spec.js
VirajP1002: I thought that the "-£1.00" would be used to check if the summary page contains it, meaning that the summary page "allows" negatives and it displays correctly. I attached an image to where I thought it would check.


Screenshot 2023-09-28 at 15 47 27

katie-gardner: Ahhh I see, yeah I think thats a good idea, what about changing it to a list, `expectedAnswerValues` so that with your other test cases where the negative value is not the first answer but another its verified too. Theres a helper method `assertSummaryValues` which you could use for this

VirajP1002: Yep, that worked and looks good!

petechd:
[Update the calculated summary schema to demo negatives (1 SP)](https://trello.com/c/BbLbQXCc/5807-update-the-calculated-summary-schema-to-demo-negatives-1-sp)

petechd (approved):

MebinAbraham (approved):
👍

tests/functional/spec/features/calculated_summary/calculated_summary.spec.js
MebinAbraham: Happy this is resolved.

tests/functional/spec/features/calculated_summary/calculated_summary_test_case.js
MebinAbraham: Happy this is resolved.

tests/functional/spec/features/calculated_summary/calculated_summary_test_case.js
MebinAbraham: Happy this is resolved. It has not been removed but is now an array. It could be dynamically calculated but left as is since it is minor.

PR #1223, opened 2023-10-06
Return to list collector summary when navigating back

liamtoozer:
For schemas with a list collector & a summary, after clicking 'continue' from the list collector summary details page and then clicking 'previous' to return back to that same page, the user is currently returned to the previous list collector block instead of the list collector summary page.

### How to review
Follow the steps stated in the Trello ticket - use `test_list_collector_list_summary` to complete the survey up to the list collector summary. Click continue again so you’re on the "Check your answers and submit" page, then click previous. You should now be returned to the list collector section summary page instead.

[Trello](https://trello.com/c/J0fMiFa9)


liamtoozer:
[Bug: List collector section summaries are skipped when navigating backwards from submission page](https://trello.com/c/J0fMiFa9/5753-bug-list-collector-section-summaries-are-skipped-when-navigating-backwards-from-submission-page)

tests/app/questionnaire/test_router.py
```diff
@@ -1781,6 +1781,30 @@ def test_last_block_not_on_path(self):
         assert completed_block_not_on_path == last_completed_block_in_progress_store
         assert expected_location_url == last_location_url
 
+    @pytest.mark.usefixtures("app")
+    def test_list_collector_list_summary_returns_to_summary(self):
```
katie-gardner, line 1785: I'm thinking instead of `list_summary_returns_to_summary` that `final_summary_returns_to_section_summary` might give that extra bit of clarity to what the test is checking - open to other thoughts or ideas

app/questionnaire/router.py
```diff
@@ -70,6 +70,11 @@ def get_first_incomplete_location_in_questionnaire_url(self) -> str:
 
     def get_last_location_in_questionnaire_url(self) -> str | None:
         section_key = self._get_last_complete_section_key()
+        if section_key and self.can_display_section_summary(section_key):
```
katie-gardner, line 73: Optional non blocker - you could use the walrus operator and nest both paths under section key existing to save two checks

katie-gardner (approved):
Nice one 🎉  Doesn't break any other schemas I've tried, and fixes it for those list ones affected. Added a couple of tiny non blocking suggestions but nothing functional, working well.

app/questionnaire/router.py
liamtoozer: Ah cool! I'll push something up to see if I've understood correctly. Thanks 👍

berroar (approved):
Tested, looks good other than Katie's comments above 👍

app/questionnaire/router.py
katie-gardner: Yep - what you did was exactly what I was thinking!

katie-gardner (approved):

PR #1224, opened 2023-10-09
Update section progress on supplementary data change

katie-gardner:
This PR Fixes two issues that can occur if supplementary data changes when a questionnaire is partially complete

1) If a question depends on the length of a supplementary list, and a new supplementary dataset is used with a list item count that opens up the question, previously the section containing the question would remain completed. It now reverts to in progress. This is true when the supplementary dataset has added or removed items, both will work.

2) If a section contains a list collector with repeating blocks depending on supplementary data. If the supplementary data changes and a list item is removed, the section will revert to in progress.

Note that it does not fix the case where a list item is added and the repeating blocks need to be revisited.

### How to review
Check the changes to the questionnaire store updater make sense. A location wasn't accessible from the session endpoint so it was made optional. And since all occurrences of removing list item data were following it with a call to capture dependencies for the list, this has been moved into the remove list item data method itself.

Additionally the method to capture dependencies for a list has been moved from question to the questionnaire store updater, and the questionnaire store updater is now used to remove list items from the list store when supplementary data changes, rather than doing it by hand so that the dependencies are correctly captured.

### How to test functionally
Run the mock SDS locally on [the branch with an additional dataset](https://github.com/ONSdigital/eq-runner-mock-sds/pull/3)
Use the `test_supplementary_data` schema, and try swapping between the three datasets while clicking the use previous response ID button.

- If using the first supplementary dataset with 2 items, if you complete the production targets section and relaunch the survey with the second supplementary dataset with 3 items, the section should become incomplete.

- If using the second supplementary dataset with 3 items, if you complete the sales targets section, then relaunch the survey using the first supplementary dataset which has 2 items, you should see the section revert to incomplete

- If the product details section is completed whilst using the second supplementary dataset, when moving to the first supplementary dataset which contains fewer items, the section should revert to in progress and force you to reconfirm the calculated summaries within.

- (Note the reverse, swapping to the dataset with more items, does not update completeness of the section, this will be addressed in a follow on card)

### Checklist


katie-gardner:
[Bug: Section progress not reset when new supplementary data version has extra dynamic answers](https://trello.com/c/qdiKw74Y/5831-bug-section-progress-not-reset-when-new-supplementary-data-version-has-extra-dynamic-answers)

schemas/test/en/test_supplementary_data.json
```diff
@@ -1544,6 +1544,204 @@
                     ]
                 }
             ]
+        },
+        {
```
berroar, line 1548: Perhaps this is just me, but when I was going through the scenarios, it made it slightly clearer to test when I added a section summary to these new sections. As otherwise when going back to view the section via 'view answers' you can only see the interstitial page, which makes it confusing as to how many list items there are.

app/routes/session.py
```diff
@@ -184,7 +186,46 @@ def _set_questionnaire_supplementary_data(
...
+    router = Router(
+        schema=schema,
+        answer_store=questionnaire_store.answer_store,
+        metadata=questionnaire_store.metadata,
+        list_store=questionnaire_store.list_store,
+        progress_store=questionnaire_store.progress_store,
+        response_metadata=questionnaire_store.response_metadata,
+        supplementary_data_store=questionnaire_store.supplementary_data_store,
+    )
+    questionnaire_store_updater = QuestionnaireStoreUpdater(
```
berroar, line 216: I think I would lean toward having the non location questionnaire store updater class, rather than having the additional location checks, but curious to see what others think?

berroar (commented):
Think all the scenarios are working functionally 👍

PR #1226, opened 2023-10-09
Updating makefile commands

VirajP1002:
Added more bespoke commands to the Makefile to streamline development processes

### How to review
- Run `make validate-test-schema` to ensure the single schema is validated
- Run `make test-single-functional` & `make test-suite-functional:`to ensure the specified single spec/suite is run
- Run `make test-lint-js` to ensure both JS lint and JS tests are run
- Run `make test-lint-python` to ensure both Python lint and Python tests are run
- Run `make test-functional-headless` to ensure functional tests are run headless

### Checklist


VirajP1002:
[Add more bespoke commands to Makefile (1 SP)](https://trello.com/c/hsbPpYML/5681-add-more-bespoke-commands-to-makefile-1-sp)

Makefile
```diff
@@ -22,6 +24,8 @@ lint: lint-python
 lint-python:
 	pipenv run ./scripts/run_lint_python.sh
 
+test-lint-python: lint-python
```
berroar, line 27: Think this one is just running lint at the moment and not test? Also minor but as we always run lint before tests I would rename these to be lint-test-python etc

Makefile
```diff
@@ -38,9 +42,28 @@ test-unit:
 test-functional:
 	pipenv run ./scripts/run_tests_functional.sh
 
+test-functional-headless:
+	EQ_RUN_FUNCTIONAL_TESTS_HEADLESS='True' pipenv run ./scripts/run_tests_functional.sh
+
+test-single-functional:
```
berroar, line 48: I think we should update the readme to specify using the makefile commands where it references running single tests/test suites 👍

setup.cfg
```diff
@@ -6,3 +6,7 @@ filterwarnings=
     ignore:.*formatargspec.*:DeprecationWarning
     ignore:.*isAlive.*:PendingDeprecationWarning
 
+[flake8]
+extend-ignore = E501,E203
+exclude = venv,tests
```
berroar, line 11: How come we're ignoring tests here? Also, so we need to add venv here and above? Since we should be using pipenv to create the virtualenv?

berroar (changes requested):

Makefile
VirajP1002: Yes, that is correct, I've made those changes.

Makefile
VirajP1002: 👍

setup.cfg
VirajP1002: The reason we're ignoring is because we would then have to refactor a large amount of Python files and not ignoring them leads to lots of lint errors. From my understanding, I had to ignore venv here as well because of the way my project is set up otherwise I still get the bunch of lint errors.

berroar: I don't think we can exclude tests though as this will ignore the python test files that should be linted in the test directory

petechd: But did we lint tests before? If not we should create another PR to fix linting issues in tests?

.eslintignore
```diff
@@ -2,3 +2,123 @@ coverage/**
 node_modules/**
 dist/**
 src/index.html
+
+# Python temp file
+*.py[cod]
```
katie-gardner, line 7: Why did each of the `.ignore` files end up with the gitignore content in them?

scripts/run_lint_python.sh
```diff
@@ -34,5 +34,5 @@ display_result $? 1 "isort linting check"
 ./scripts/run_mypy.sh
 display_result $? 1 "Mypy type check"
 
-black --check . --exclude node_modules
+black --check . --exclude node_modules --exclude venv
```
katie-gardner, line 37: I think this change should be reverted, sorry If I caused confusion, I'd meant that if you use pipenv to make the virtualenv outside of the main directory then this wouldn't be an issue anymore

setup.cfg
petechd: We use this config for flake8 so maybe we should paste it into the new config:
```
[flake8]
max-line-length = 160
ignore = C815,C816,W503,E203
```

.pylintrc
```diff
@@ -48,7 +48,7 @@ ignore=CVS
 # ignore-list. The regex matches against paths and can be in Posix or Windows
 # format. Because '\' represents the directory delimiter on Windows systems, it
 # can't be used as an escape character.
-ignore-paths=
+ignore-paths=venv
```
katie-gardner, line 51: As with other comment I think this can be reverted, if pipenv is used with a venv outside the runner folder

setup.cfg
berroar: Think @petechd is correct, the config should be the same as above 👍

katie-gardner (changes requested):

.eslintignore
VirajP1002: When we run yarn lint and format, I believe it only checks the JS and JSON files (From package.json) so we can include the values in .ignore as it doesn't need to check over these values.

scripts/run_lint_python.sh
VirajP1002: No worries, I can remove those

Makefile
MebinAbraham, line 48: This might be better named as `test-functional-spec`

Makefile
```diff
@@ -38,9 +42,28 @@ test-unit:
...
+	yarn test_functional --spec ./tests/functional/spec/$(SPEC)
+
+test-suite-functional:
```
MebinAbraham, line 51: I suggest naming this `test-functional-suite`

.eslintignore
petechd: Yeah, we don't want to look for `.js` files everywhere in runner root, that was the idea behind it.

Makefile
```diff
@@ -38,9 +42,28 @@ test-unit:
...
+	yarn test_functional --suite $(SUITE)
+
+lint-js:
+	yarn lint
+
+lint-test-js: lint-js test-functional
+
+test-headless-lint-js: lint-js test-functional-headless
```
MebinAbraham, line 59: We need to be careful from making the make commands too verbose and adding combinations we won't use.
Can devs verify that all the commands we have are all useful.

Makefile
```diff
@@ -1,6 +1,8 @@
+include .functional-tests.env
```
MebinAbraham, line 1: Why do we need to do this?.

VirajP1002: It isn't needed, I forgot to remove it, I've removed it now

setup.cfg
MebinAbraham: Why is it in setup.cfg?. It is already configured in `tox.ini`, we only want 1 source of truth. Either is fine by me as long as there is only one. Is there some specific reason both are needed?

Makefile
VirajP1002: The command on 59 was specified on the ticket itself, however, when running both JS test + lint, I thought it would be appropriate to provide the option to allow the dev to run JS test + lint headless. If the name is an issue, I can think of a different one.

MebinAbraham:
Test slack comment. Ignore.

Makefile
katie-gardner: I think if we do `lint-test-js` it makes sense to also have a headless version, I'd make the names more consistent though, like `lint-test-js` and `lint-test-headless-js`. That said, I probably won't use these two

.eslintignore
katie-gardner: I'm all for adding more files and directories to avoid it checking things in the root, though the whole ignore seems overkill, did many of these contains js files?

katie-gardner (approved):
All the new commands working as intended, readme makes sense, nice one 🎉

setup.cfg
VirajP1002: I've made the changes, let me know if they are suitable

README.md
```diff
@@ -278,7 +282,7 @@ yarn test_functional
 This can be limited to a single spec using:
 
 ``` shell
-yarn test_functional --spec ./tests/functional/spec/exit.spec.js
+make test-functional-spec SPEC=<spec>
```
MebinAbraham, line 285: There is a yarn command above this which should be replaced.
There are multiple others which should be replaced with the `make` counterparts.

README.md
```diff
@@ -278,7 +282,7 @@ yarn test_functional
 This can be limited to a single spec using:
```
MebinAbraham, line 282: Worth mention here, it should include the path to the spec, not just the same.

README.md
```diff
@@ -291,7 +295,7 @@ Test suites are configured in the `wdio.conf.js` file.
 An individual test suite can be run using:
```
MebinAbraham, line 295: Existing issue but while we are here, worth mentioning the values you can use.

MebinAbraham (commented):
https://github.com/ONSdigital/eq-questionnaire-runner/blob/main/.github/workflows/pull_request.yml This should be updated to make use of any make commands we have available. There are references to yarn and direct calls to .sh files

.github/workflows/pull_request.yml
```diff
@@ -169,7 +169,7 @@ jobs:
       - name: Docker compose
         run: docker-compose --version && RUNNER_ENV_FILE=.functional-tests.env docker-compose up --build -d
       - name: Functional tests
-        run: ./scripts/run_tests_functional.sh ${{ matrix.suite }}
+        run: make test-functional-suite SUITE=${{ matrix.suite }}
```
katie-gardner, line 172: Nice 👍

Makefile
```diff
@@ -35,11 +39,32 @@ test:
...
+
+generate-spec:
+	pipenv run python -m tests.functional.generate_pages schemas/test/en/$(SCHEMA).json ./tests/functional/generated_pages/$(patsubst test_%,%,$(SCHEMA)) -r '../../base_pages' -s tests/functional/spec/$(SCHEMA).spec.js
 
 validate-test-schemas:
-	pipenv run ./scripts/validate_test_schemas.sh
+	./scripts/validate_test_schemas.sh
+
+validate-test-schema:
+	./scripts/validate_test_schemas.sh $(SCHEMA_PATH)$(SCHEMA)
```
katie-gardner, line 67: Super minor - but now that we have another command which adds the `.json` for you - I feel it'd be nice here too for consistency and also since they all end in `.json` I'd find it convenient to only need to type

`make validate-test-schema SCHEMA=test_address`

katie-gardner (approved):
Commands all working, looks good 👍 
Added one tiny suggestion, non-blocker

Makefile
VirajP1002: Sure, I can add that in 👍

petechd (approved):

Makefile
katie-gardner: I don't have a strong view - it was just so it would be consistent with the generate-spec command. My slight preference is consistency with how SCHEMA= is passed in but I'm still happy with it as is, and if not including the `.json` then I agree with dropping the second command

berroar (approved):

MebinAbraham (approved):
Happy with current state, left a non blocker comment.

Nice work 👍

Yuyuutsu (approved):

MebinAbraham (approved):

PR #1228, opened 2023-10-17
Add DataStores class to questionnaire store

petechd:
This PR encapsulates all our questionnaire store's data stores into single `DataStores` class. I've added the optional methods to initialise `ValueSourceResolver`, `RuleEvaluator`,`PlaceholderRenderer` to further reduce repetition. New class is kept in questionnaire store module for now, just to get it working in the first pass but happy to move it out.

### How to review
Running functional and unit tests plus some manual checks in live schemas and visual code checks should be sufficient.

### Checklist


petechd:
[Encapsulate questionnaire store data stores into a single class (3 SP)](https://trello.com/c/zpoYPyXJ/5802-encapsulate-questionnaire-store-data-stores-into-a-single-class-3-sp)

app/views/contexts/summary/question.py
```diff
@@ -31,54 +25,55 @@ def __init__(
...
         self.schema = schema
         self.answer_schemas = iter(question_schema.get("answers", []))
-        self.answer_store = answer_store
-        self.list_store = list_store
-        self.progress_store = progress_store
-        self.supplementary_data_store = supplementary_data_store
+        self.answer_store = data_stores.answer_store
+        self.list_store = data_stores.list_store
+        self.progress_store = data_stores.progress_store
+        self.supplementary_data_store = data_stores.supplementary_data_store
```
katie-gardner, line 45: the only place these attributes are used is in creating `PlaceholderRenderer` on line 305 
they could be removed and the datastores method can be used instead to initialise it

app/data_models/questionnaire_store.py
```diff
@@ -1,23 +1,111 @@
...
         EncryptedQuestionnaireStorage,
     )
 
 
+@dataclass
+class DataStores:
+    # self.metadata is a read-only view over self._metadata
+    metadata: MetadataProxy | None = None
+    response_metadata: MutableMapping = field(default_factory=dict)
+    list_store: ListStore = ListStore()
```
katie-gardner, line 37: We should avoid mutable default arguments, these need the default factory too, or a normal class with none defaults

app/data_models/questionnaire_store.py
```diff
@@ -1,23 +1,111 @@
...
+            supplementary_data_store=self.supplementary_data_store,
+        )
+
+    def placeholder_renderer(
+        self,
+        language: str,
+        schema: QuestionnaireSchema,
+        location: LocationType | None = None,
+        placeholder_preview_mode: bool | None = False,
+    ) -> placeholder_renderer.PlaceholderRenderer:
```
katie-gardner, line 94: it looks a bit odd to have the import name match the function name, with this class in a separate file I'd import `PlaceholderRenderer` directly

app/data_models/questionnaire_store.py
```diff
@@ -120,48 +204,55 @@ def _remove_old_supplementary_lists_and_answers(
...
         data = {
             "METADATA": self._metadata,
-            "ANSWERS": list(self.answer_store),
-            "SUPPLEMENTARY_DATA": self.supplementary_data_store.serialize(),
-            "LISTS": self.list_store.serialize(),
-            "PROGRESS": self.progress_store.serialize(),
-            "RESPONSE_METADATA": self.response_metadata,
+            "ANSWERS": list(self.data_stores.answer_store),
+            "SUPPLEMENTARY_DATA": self.data_stores.supplementary_data_store.serialize(),
+            "LISTS": self.data_stores.list_store.serialize(),
```
katie-gardner, line 244: Since this is serialising each thing in the data stores, I think a serialise method in the DataStores class which gets called here would make sense

app/data_models/questionnaire_store.py
```diff
@@ -120,48 +204,55 @@ def _remove_old_supplementary_lists_and_answers(
...
         return json_dumps(data)
 
     def delete(self) -> None:
         self._storage.delete()
         self._metadata.clear()
-        self.response_metadata = {}
-        self.answer_store.clear()
-        self.progress_store.clear()
+        self.data_stores.response_metadata = {}
+        self.data_stores.answer_store.clear()
```
katie-gardner, line 254: Likewise I think the datastores should have its own delete, so future changes are made there rather than externally

app/questionnaire/rules/rule_evaluator.py
```diff
@@ -30,13 +29,13 @@
 @dataclass
 class RuleEvaluator:
     schema: QuestionnaireSchema
-    answer_store: AnswerStore
-    list_store: ListStore
+    answer_store: answer_store.AnswerStore
+    list_store: list_store.ListStore
```
katie-gardner, line 33: Why has this been done?
I think it would be good here to remove these parameters and replace them with `data_stores` then the placeholder renderer below can come from the data stores

app/questionnaire/value_source_resolver.py
```diff
@@ -34,8 +34,8 @@ class ValueSourceResolver:
     schema: QuestionnaireSchema
     location: LocationType | None
     list_item_id: str | None
-    progress_store: ProgressStore
-    supplementary_data_store: SupplementaryDataStore
+    progress_store: progress_store.ProgressStore
+    supplementary_data_store: supplementary_data_store.SupplementaryDataStore
```
katie-gardner, line 38: Was some of this refactor done with a tool? Some of the changes look a bit odd

app/submitter/converter_v2.py
```diff
@@ -49,15 +49,15 @@ def convert_answers_v2(
...
-    list_store = questionnaire_store.list_store
-    progress_store = questionnaire_store.progress_store
-    supplementary_data_store = questionnaire_store.supplementary_data_store
+    response_metadata = questionnaire_store.data_stores.response_metadata
+    answer_store = questionnaire_store.data_stores.answer_store
+    list_store = questionnaire_store.data_stores.list_store
+    progress_store = questionnaire_store.data_stores.progress_store
+    supplementary_data_store = questionnaire_store.data_stores.supplementary_data_store
 
     survey_id = schema.json["survey_id"]
```
katie-gardner, line 62: on line 115, `get_payload_data` still has all the params, but it could be changed to just having `data_stores` and a large number of places calling it then don't need to pass each parameter from `questionnaire_store.data_stores` one by one

app/questionnaire/placeholder_parser.py
```diff
@@ -57,8 +58,8 @@ def __init__(
         response_metadata: MutableMapping,
         schema: QuestionnaireSchema,
         renderer: PlaceholderRenderer,
-        progress_store: ProgressStore,
-        supplementary_data_store: SupplementaryDataStore,
+        progress_store: ps.ProgressStore,
+        supplementary_data_store: sds.SupplementaryDataStore,
```
katie-gardner, line 62: Args need removing here too and replacing with data stores

app/questionnaire/value_source_resolver.py
petechd: This one is done to avoid circular import error.

app/views/contexts/summary/question.py
petechd: I don't think this one is possible as we initialise `PlaceholderRenderer` within the `DataStores` class. So can we pass `DataStores` instance inside itself?
If we can do:
```python
    def placeholder_renderer(
        self,
        language: str,
        schema: QuestionnaireSchema,
        location: LocationType | None = None,
        placeholder_preview_mode: bool | None = False,
    ) -> placeholder_renderer.PlaceholderRenderer:
        return placeholder_renderer.PlaceholderRenderer(
            language=language,
            data_stores=self,
            schema=schema,
            location=location,
            placeholder_preview_mode=placeholder_preview_mode,
        )
```
then I will refactor it as well.

app/data_models/questionnaire_store.py
petechd: Caused by circular import errors.

app/questionnaire/rules/rule_evaluator.py
petechd: Circular import error.

app/questionnaire/placeholder_parser.py
petechd: Similar issue to https://github.com/ONSdigital/eq-questionnaire-runner/pull/1228#discussion_r1362072105, I will try to refactor this more.

katie-gardner: Ahhh I see - that makes sense now, i think you can do the imports under `if TYPE_CHECKED` to avoid the circular errors

app/questionnaire/rules/rule_evaluator.py
katie-gardner: Same as other, ignore about replacing args, I think the imports can be fixed with `TYPE_CHECKED` - I see what you mean

app/views/contexts/summary/question.py
katie-gardner: I see what you mean with the other ones, makes sense to keep params if you've got those creation methods in data stores
I think this one can be fixed though if you use `placeholder_renderer = self.data_stores.placeholder_renderer` instead of initialising directly?

petechd: I've just checked/refactored a lot and it seems impossible because of all the circular imports issues for type hints.

petechd: > I think this one can be fixed though if you use `placeholder_renderer = self.data_stores.placeholder_renderer` instead of initialising directly?

Yeah, this should be possible to change.

app/data_models/questionnaire_store.py
```diff
@@ -1,23 +1,111 @@
...
+from app.utilities.types import LocationType, SupplementaryDataListMapping
 
 if TYPE_CHECKING:
     from app.storage.encrypted_questionnaire_storage import (  # pragma: no cover
         EncryptedQuestionnaireStorage,
     )
 
 
+@dataclass
+class DataStores:
```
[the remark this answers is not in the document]
petechd: I think I decided to keep it questionnaire store for the same reasons, we get circular import errors for AnswerStore, ListStore, ProgressStore, SupplementaryDataStore.

app/data_models/questionnaire_store.py
petechd: The only thing is, each of these `serialize` methods does something slightly different, depending on the store. For exa ple, `SuplementaryDataStore` one creates this object:
```python
return {
            "data": self._raw_data,
            "list_mappings": self._list_mappings,
        }
```

app/data_models/questionnaire_store.py
petechd: Same here, each `clear` does something different, depending on the store.

app/data_models/questionnaire_store.py
petechd: Ok, added now, thanks.

app/data_models/questionnaire_store.py
katie-gardner: I'd just wondered if it would make sense to have a `serialize` method in the datastores, which is essentially just this one, and saves you needing to do `self.data_stores.....serialize()` etc and makes it the responsibility of datastores, and the questionnaire store would then be able to just call that

app/data_models/questionnaire_store.py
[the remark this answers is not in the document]
petechd: This will cause circular import errors for all the stores.

app/data_models/questionnaire_store.py
petechd: I get what you mean, we could do all of it in `data_stores` but in both `serialize` and `_deserialize` we change `QuestionnaireStore`'s `self._metadata` so I am not sure if we want to separate it out? `serialize` also calls `QuestionnaireStore`'s `self.set_metadata` method.

app/data_models/questionnaire_store.py
petechd: Changed it to:
```python
    def delete(self) -> None:
        self._storage.delete()
        self._metadata.clear()
        self.data_stores.delete()
```

app/data_models/questionnaire_store.py
petechd: I had to do couple more refactors to move it to `data_stores`.

app/data_models/questionnaire_store.py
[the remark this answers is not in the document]
petechd: In separate file now, but had to migrate some methods so we don't get circular imports issue.

app/questionnaire/value_source_resolver.py
```diff
@@ -34,8 +34,8 @@ class ValueSourceResolver:
     schema: QuestionnaireSchema
     location: LocationType | None
     list_item_id: str | None
-    progress_store: ProgressStore
-    supplementary_data_store: SupplementaryDataStore
+    progress_store: progress_store.ProgressStore
```
Yuyuutsu, line 37: for consistency are we able to  import progress_store as ps like in `placeholder_renderer.py`

app/forms/questionnaire_form.py
```diff
@@ -492,31 +458,26 @@ def get_answer_fields(
...
-                    progress_store=progress_store,
-                    metadata=metadata,
-                    response_metadata=response_metadata,
-                    supplementary_data_store=supplementary_data_store,
+                    answer_store=data_stores.answer_store,
+                    list_store=data_stores.list_store,
+                    progress_store=data_stores.progress_store,
+                    metadata=data_stores.metadata,
+                    response_metadata=data_stores.response_metadata,
+                    supplementary_data_store=data_stores.supplementary_data_store,
```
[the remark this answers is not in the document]
petechd: It's because `PathFinder` is initialised in `PlaceholderParser` which is initialised in `PlaceholderRenderer` which is part of a helper method in `DataStores` class and we pass in separate stores as arguments at `DataStores` level. But we can recreate `DataStores` as one entity in parser to pass it to path finder.

app/data_models/questionnaire_store.py
```diff
@@ -120,48 +110,43 @@ def _remove_old_supplementary_lists_and_answers(
...
-        self.answer_store = AnswerStore(json_data.get("ANSWERS"))
-        self.list_store = ListStore.deserialize(json_data.get("LISTS"))
-        self.response_metadata = json_data.get("RESPONSE_METADATA", {})
+        self.data_stores.deserialize_stores(json_data, metadata=self._metadata)
 
     def serialize(self) -> str:
-        data = {
+        serialized_stores = self.data_stores.serialize_stores()
+        data = serialized_stores | {
             "METADATA": self._metadata,
```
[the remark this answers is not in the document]
katie-gardner: I see what you mean with the abstraction, though, mixed thoughts, it doesn't seem right to me to have a method thats
```python
def method(self):
    self.data_stores.x.a()
    self.data_stores.y.b()
    self.data_stores.z.c()
```
the entire functionality of it relates to the `data_stores` so I would have thought it makes sense to make the data stores responsible for it. That way if someone comes to change something about a data store, I think they're less likely to accidentally miss logic from another class, even if that does result in a fair bit of it moving.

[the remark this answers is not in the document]
katie-gardner: Though I guess this would be fixed by it being a view only add-on though and having each attribute still present on the store, so I think I'd also lean towards the attributes still being in this class

app/data_models/data_stores.py
```diff
@@ -0,0 +1,134 @@
...
+    )
+
+    def rule_evaluator(
+        self,
+        *,
+        schema: QuestionnaireSchema,
+        location: Location,
+        routing_path_block_ids: Iterable[str] | None = None,
+    ) -> rule_evaluator.RuleEvaluator:
+        return rule_evaluator.RuleEvaluator(
```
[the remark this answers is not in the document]
petechd: This one is gone from data stores/outdated comment.

app/data_models/data_stores.py
```diff
@@ -0,0 +1,134 @@
...
+            metadata=self.metadata,
+            response_metadata=self.response_metadata,
+            schema=schema,
+            progress_store=self.progress_store,
+            supplementary_data_store=self.supplementary_data_store,
+            location=location,
+            placeholder_preview_mode=placeholder_preview_mode,
+        )
+
+    def set_supplementary_store(
```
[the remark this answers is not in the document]
petechd: Same, helper methods removed from data stores.

app/data_models/data_stores.py
```diff
@@ -0,0 +1,134 @@
...
+    placeholder_renderer,
+    value_source_resolver,
+)
+from app.questionnaire.relationship_location import RelationshipLocation
+from app.questionnaire.rules import rule_evaluator
+from app.utilities.types import LocationType
+
+
+@dataclass
+class DataStores:
```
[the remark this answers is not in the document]
petechd: Data stores are now only grouping all the stores, no helper methods in this module.

app/views/contexts/summary/question.py
```diff
@@ -31,54 +26,61 @@ def __init__(
...
         self.schema = schema
+        self.data_stores = data_stores
         self.answer_schemas = iter(question_schema.get("answers", []))
-        self.answer_store = answer_store
-        self.list_store = list_store
-        self.progress_store = progress_store
-        self.supplementary_data_store = supplementary_data_store
+        self.answer_store = data_stores.answer_store
+        self.list_store = data_stores.list_store
+        self.progress_store = data_stores.progress_store
```
katie-gardner, line 46: these attributes aren't being used, so they could be deleted

app/views/contexts/summary/block.py
```diff
@@ -23,109 +15,71 @@ def __init__(
...
-            list_item_id=self.location.list_item_id if self.location else None,
-            use_default_answer=True,
-            progress_store=progress_store,
-            supplementary_data_store=supplementary_data_store,
-        )
+        # self._rule_evaluator = self.data_stores.rule_evaluator(
+        #     # Type ignore: location in rule_evaluator can be both Location or RelationshipLocation type but is only Location type here
+        #     schema=self.schema,
+        #     location=self.location,  # type: ignore
+        # )
```
katie-gardner, line 37: Couple of left behind commented out bits

app/views/contexts/summary/question.py
```diff
@@ -31,54 +26,61 @@ def __init__(
...
         )
         self.number = question_schema.get("number", None)
 
-        self.rule_evaluator = rule_evaluator
-        self.value_source_resolver = value_source_resolver
+        self._rule_evaluator = RuleEvaluator(
+            schema=self.schema,
+            data_stores=data_stores,
+            location=self.location,
+        )
```
katie-gardner, line 61: Nice, better place for these 👍

katie-gardner (commented):
Will do a second pass but from a first look only found two small things 👍

katie-gardner (changes requested):
Finished going through now - just a few more small things

app/questionnaire/router.py
```diff
@@ -87,9 +66,6 @@ def can_access_location(
         Checks whether the location is valid and accessible.
         :return: boolean
         """
-        if location.section_id not in self.enabled_section_ids:
-            return False
```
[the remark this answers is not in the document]
petechd: It's not covered in any of the tests (after adding the changes) and doesn't cause any adverse effects after removal in functional tests, it's hard to find this ever used..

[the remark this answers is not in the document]
katie-gardner: on main if I replace `return False` with raising an error, this test `test_cant_access_section_disabled` from `test_router` fails - it seems slightly concerning that the changes would stop a line being it if we don't know why, any idea why that test doesn't hit that line anymore with the changes?

tests/app/questionnaire/test_router.py
```diff
@@ -142,6 +129,7 @@ def test_is_complete(self):
 
     def test_is_not_complete(self):
         self.schema = load_schema_from_name("test_textfield")
+        self.data_stores.progress_store = ProgressStore()
```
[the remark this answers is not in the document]
petechd: It's because progress store is previously initialised and have value in test case class above it.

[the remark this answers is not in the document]
katie-gardner: Ah OK - why is that only happening as a result of the change?

app/questionnaire/router.py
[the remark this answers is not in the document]
petechd: Ok, it is a question of initialising a clean data store then possibly, since we are within the context of that parent test case? This seems to work (visits the code now):
```python
    def test_cant_access_section_disabled(self):
        self.schema = load_schema_from_name("test_section_enabled_checkbox")
        self.data_stores = DataStores()

        current_location = Location(
            section_id="section-2", block_id="section-2-block", list_item_id=None
        )
        can_access_location = self.router.can_access_location(
            current_location, routing_path=[]
        )

        assert not can_access_location
```

app/questionnaire/rules/rule_evaluator.py
```diff
@@ -30,43 +23,30 @@
...
-            answer_store=self.answer_store,
-            list_store=self.list_store,
-            metadata=self.metadata,
-            response_metadata=self.response_metadata,
-            schema=self.schema,
-            location=self.location,
-            progress_store=self.progress_store,
-            supplementary_data_store=self.supplementary_data_store,
+        renderer: placeholder_renderer.PlaceholderRenderer = (
+            placeholder_renderer.PlaceholderRenderer(
```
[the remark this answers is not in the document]
petechd: Seems like something I should remove after stripping down of data stores class, ordinary import works now.

app/questionnaire/router.py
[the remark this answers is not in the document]
katie-gardner: ahhh, yeah, makes sense
we can probably do something like this in `RouterTestCase` to make sure data stores is fresh every time
```python
class RouterTestCase:
    schema = None
    data_stores: DataStores

    @pytest.fixture(autouse=True)
    def setup(self):
        self.data_stores = DataStores()

    @cached_property
    def router(self):
        return Router(self.schema, self.data_stores)
```

app/views/contexts/calculated_summary_context.py
```diff
@@ -67,7 +52,7 @@ def rendered_block(self) -> dict:
         block_id: str = self.current_location.block_id  # type: ignore
         block: ImmutableDict = self._schema.get_block(block_id)  # type: ignore
 
-        return self._placeholder_renderer.render(
+        return self._placeholder_renderer.render(  # type: ignore
```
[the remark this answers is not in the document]
petechd: This one is a leftover from the previous refactor when I was asked to call `data_stores.placeholder_renderer` helper and it had different location type hints possibly, removing now.

tests/app/forms/test_questionnaire_form.py
```diff
@@ -1438,7 +1276,7 @@ def test_sum_calculated_field_value_source_calculated_summary_repeat_not_equal_v
         ] == schema.error_messages["TOTAL_SUM_NOT_EQUALS"] % {"total": "10"}
 
 
-def test_multi_calculation(app, answer_store, list_store):
+def test_multi_calculation(app, answer_store, data_stores):
```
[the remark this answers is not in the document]
petechd: Just dropped data_stores as param if that's ok? Makes more sense to me.

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -63,19 +59,14 @@ def get_calculation_block(
 
 def get_value_source_resolver(
     schema: QuestionnaireSchema = None,
-    answer_store: AnswerStore = AnswerStore(),
-    list_store: ListStore = ListStore(),
-    metadata: MetadataProxy = None,
-    response_metadata: Mapping = None,
+    data_stores: DataStores = DataStores(),
```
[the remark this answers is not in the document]
petechd: Removed... and in some other one as well (`test_rule_evaluator.py`).

app/questionnaire/router.py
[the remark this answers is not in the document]
petechd: Yeah, makes sense.

tests/app/questionnaire/test_router.py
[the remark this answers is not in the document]
petechd: Ok, so initialising up where the class gets defined. Sorted.

liamtoozer (commented):
Looks great from what I can see & understand! Just had a couple of learning questions but no worries if there's not enough time

app/questionnaire/placeholder_renderer.py
```diff
@@ -64,12 +51,16 @@ def get_plural_count(
...
         if source == "answers":
-            answer = self._answer_store.get_answer(source_id)
+            answer = self._data_stores.answer_store.get_answer(source_id)
             return answer.value if answer else None
         if source == "list":
-            return len(self._list_store[source_id])
+            return len(self._data_stores.list_store[source_id])
 
-        return self._metadata[source_id] if self._metadata else None
+        return (
```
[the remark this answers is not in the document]
petechd: The bracket here is just a syntactic sugar, something automated formatting does, just a multiline code, such brackets on one line means other things.

app/views/contexts/calculated_summary_context.py
```diff
@@ -249,15 +224,11 @@ def _get_evaluated_total(
...
-            self._metadata,
-            self._response_metadata,
+            data_stores=self._data_stores,
+            schema=self._schema,
             routing_path_block_ids=routing_path_block_ids,
-            location=self.current_location,
-            progress_store=self._progress_store,
-            supplementary_data_store=self._supplementary_data_store,
+            # Type ignore: location in rule_evaluator can be both Location or RelationshipLocation type but is only Location type here
+            location=self.current_location,  # type: ignore
```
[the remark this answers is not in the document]
petechd: Yeah, good catch. This one was ignored before when I still had helper methods in `DataStores` where location was either `Location` type or `RelationshipLocation`, location in `RuleEvaluator` is only `Location` type but I was executing it from data stores (data_stores.rule_evaluator) hence the conflict. Not needed after the refactor.

app/questionnaire/placeholder_renderer.py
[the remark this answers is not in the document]
liamtoozer: Ah okay cool, that makes sense, thanks 👍

app/views/contexts/calculated_summary_context.py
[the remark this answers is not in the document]
liamtoozer: Ah I see, thanks for explaining that makes sense 👍

app/questionnaire/router.py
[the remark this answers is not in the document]
katie-gardner: Yeah I think tests should not depend on each other so good stuff 👍

tests/app/questionnaire/test_placeholder_renderer.py
```diff
@@ -326,7 +326,7 @@ def get_placeholder_render_dynamic_answers(*, language="en", data_stores=DataSto
     )
 
 
-def get_placeholder_render(*, language="en", data_stores=DataStores()):
+def get_placeholder_render(*, language="en", data_stores=None):
     schema = MagicMock()
```
katie-gardner, line 330: if `data_stores` _can_ be None, would be good to cover for it, at the moment it could be called without being passed in, and then not work, with the other comment I'd imagined something like
```python
def x(data_stores=None):
    data_stores = data_stores or DataStores()
```
avoiding mutable default but allowing it not to be passed in, so all the calls to this don't need `get_placeholder_renderer(DataStores())`
atm we don't get the benefit from the default

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -59,7 +59,7 @@ def get_calculation_block(
 
 def get_value_source_resolver(
     schema: QuestionnaireSchema = None,
-    data_stores: DataStores = DataStores(),
+    data_stores: DataStores = None,
     location: Union[Location, RelationshipLocation] = Location(
```
katie-gardner, line 63: same with this one, if this function handles the default being None with `data_stores or DataStores()` then it doesn't need to be passed in everywhere

tests/app/questionnaire/test_placeholder_renderer.py
petechd: Done, thanks.

tests/app/questionnaire/test_value_source_resolver.py
petechd: Done.

katie-gardner (approved):
My comments all fixed 👍

app/data_models/data_stores.py
```diff
@@ -0,0 +1,21 @@
...
+from app.data_models.list_store import ListStore
+from app.data_models.metadata_proxy import MetadataProxy
+from app.data_models.progress_store import ProgressStore
+from app.data_models.supplementary_data_store import SupplementaryDataStore
+
+
+@dataclass
+class DataStores:
+    # self.metadata is a read-only view over self._metadata
+    metadata: MetadataProxy | None = None
```
katie-gardner, line 14: Non blocker - this comment doesn't apply any more, either we need `_metadata` here or reference questionnaire store

petechd: @katie-gardner not sure about the best wording for the final comment, can it be `# self.metadata is a read-only view over QuestionnaireStore's self._metadata` ?

petechd: It's something from before my time on the team so I never looked into details of this. If we add a setter, it will cause all these issues we discussed before.

MebinAbraham: I don't think the comment is needed, otherwise it would be needed everywhere we pass `MetadataProxy`. It also is unclear as it is reference another class' property. We could get rid of that property and keep unmodified version as part of MetadataProxy, it's not uncommon for a class to have "_raw_data". It means 2 classes don't need to store the same thing.

app/data_models/data_stores.py
```diff
@@ -0,0 +1,21 @@
...
+
+from app.data_models.answer_store import AnswerStore
+from app.data_models.list_store import ListStore
+from app.data_models.metadata_proxy import MetadataProxy
+from app.data_models.progress_store import ProgressStore
+from app.data_models.supplementary_data_store import SupplementaryDataStore
+
+
+@dataclass
+class DataStores:
```
MebinAbraham, line 12: What break if you made this frozen?

petechd:
```
app/data_models/questionnaire_store.py:58: error: Property "metadata" defined in "DataStores" is read-only  [misc]
app/data_models/questionnaire_store.py:80: error: Property "supplementary_data_store" defined in "DataStores" is read-only  [misc]
app/data_models/questionnaire_store.py:143: error: Property "progress_store" defined in "DataStores" is read-only  [misc]
app/data_models/questionnaire_store.py:145: error: Property "supplementary_data_store" defined in "DataStores" is read-only  [misc]
app/data_models/questionnaire_store.py:148: error: Property "answer_store" defined in "DataStores" is read-only  [misc]
app/data_models/questionnaire_store.py:149: error: Property "list_store" defined in "DataStores" is read-only  [misc]
app/data_models/questionnaire_store.py:150: error: Property "response_metadata" defined in "DataStores" is read-only  [misc]
app/data_models/questionnaire_store.py:166: error: Property "response_metadata" defined in "DataStores" is read-only  [misc]

```
and multitude of tests, but the fixing above will lead to another host of issues somewhere else, from what I recall (when I tried it).

MebinAbraham: That means, tests are written badly. What breaks outside of tests, what functionally breaks?.

petechd: From what I recall, for starters, all calls to metadata throughout the code, for example:
```python
schema_name := metadata.schema_name
```
or
```python
contextvars.bind_contextvars(
        tx_id=metadata.tx_id,
        ce_id=metadata.collection_exercise_sid,
    )
```
will be broken since making dataclass frozen won't allow us to access metadata this way anymore (but from what I remember, accessing by key like `metadata["id"]` will cause another host of issues). I didn't dig much further as even fixing metadata and all the tests was a lot of churn and looking somehow beyond the scope of the card.

MebinAbraham: Are key access like `metadata["id"]` disallowed when frozen, even though they are not being set?.

petechd: I deleted the last comment, until I finished investigating some more, but basically, a lot of issues with the metedata, the fact that it is in both `questionnaire_store` and `data_stores`, cannot overcome some obstacles when initiating DataStore whenever it needs to be re-instantiated. Here's the original comment:




> From what I recall, for starters, all calls to metadata throughout the code, for example:
```python
schema_name := metadata.schema_name
```
> or
```python
contextvars.bind_contextvars(
        tx_id=metadata.tx_id,
        ce_id=metadata.collection_exercise_sid,
    )
```
> will be broken since making dataclass frozen won't allow us to access metadata this way anymore (but from what I remember, accessing by key like `metadata["id"]` will cause another host of issues). I didn't dig much further as even fixing metadata everywhere and all the tests was a lot of churn and looking somehow beyond the scope of the card. We had a discussion about making datastore just a separate, mutable class for now (on the same day) and I didn't pursue the idea of making it frozen any further and lost the changes since then.

petechd: There is a lot of extra work making it work as frozen datastore, I am not sure if I should fully dedicate my time investigating it.

MebinAbraham: Alright, let's leave it for now, there is enough indication that it isn't straight forward, so we can tackle it separately.

petechd: Yeah, from what I checked so far, It needs some considerable time allocated for investigating, thinking through and implementing (it sort of branches off from this as a separate task).

app/data_models/data_stores.py
petechd: Removed comment.

berroar (approved):
Looks good 👍

app/data_models/questionnaire_store.py
```diff
@@ -72,7 +60,7 @@ def set_supplementary_data(self, to_set: MutableMapping) -> None:
         this updates ListStore to add/update any lists for supplementary data and stores the
         identifier -> list_item_id mappings in the supplementary data store to use in the payload at the end
         """
-        if self.supplementary_data_store.list_mappings:
+        if self.data_stores.supplementary_data_store.list_mappings:
```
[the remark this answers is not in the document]
petechd: Renamed them, they are `stores` now, thanks.

app/forms/questionnaire_form.py
```diff
@@ -331,27 +311,18 @@ def _get_period_range_for_single_date(
...
             schema=self.schema,
-            answer_store=self.answer_store,
-            list_store=self.list_store,
-            metadata=self.metadata,
-            response_metadata=self.response_metadata,
-            location=self.location,
-            progress_store=self.progress_store,
-            supplementary_data_store=self.supplementary_data_store,
+            # Type ignore: location in rule_evaluator can be both Location or RelationshipLocation type but is only Location type here
+            location=self.location,  # type: ignore
```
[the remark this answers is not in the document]
petechd: I needed this for location when used in helper methods in `DataStores` class when they were  still there, before we reverted, forgot to remove it after, thanks.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -206,10 +206,10 @@ def update_same_name_items(
         same_name_items = set()
         people_names: dict[str, str] = {}
 
-        list_model = self._questionnaire_store.list_store[list_name]
+        list_model = self._questionnaire_store.data_stores.list_store[list_name]
 
         for current_list_item_id in list_model:
-            answers = self._questionnaire_store.answer_store.get_answers_by_answer_id(
+            answers = self._questionnaire_store.data_stores.answer_store.get_answers_by_answer_id(
```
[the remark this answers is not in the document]
petechd: Used now... and in few others.

tests/app/data_model/test_questionnaire_store.py
```diff
@@ -23,24 +23,26 @@ def test_questionnaire_store_json_loads(
...
     store = QuestionnaireStore(questionnaire_store.storage)
     # Then
-    assert store.metadata == MetadataProxy.from_dict(basic_input["METADATA"])
-    assert store.response_metadata == basic_input["RESPONSE_METADATA"]
-    assert store.answer_store == AnswerStore(basic_input["ANSWERS"])
+    assert store.data_stores.metadata == MetadataProxy.from_dict(
+        basic_input["METADATA"]
+    )
+    assert store.data_stores.response_metadata == basic_input["RESPONSE_METADATA"]
+    assert store.data_stores.answer_store == AnswerStore(basic_input["ANSWERS"])
```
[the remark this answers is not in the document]
petechd: Refactored in places where this occurred in this file.

MebinAbraham (approved):
Do have mixed feelings about this change but in general I think the change is for the better. I think we've accepted that we have sacrificed some readability for maintainability/extendibility.


LGTM 👍

PR #1229, opened 2023-10-17
Create ReturnLocation object to store return params

liamtoozer:
This is a small refactor PR to prevent having to pass `return_to...` strings across different areas of runner. It now means that all `return_to...` attributes can be stored and passed around in an immutable `ReturnLocation` object, and will reduce the amount of parameters for some functions.

### How to review
Functionally, there should be no regression from these changes. Without breaking the existing tests, I used `ReturnLocation.to_dict()` to entirely unpack the object in some of [the more basic](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1229/files#diff-fff3ae9abad33e32a37673e46ae937e7e16842a53214e660e63257c503c6e3dbR204) `url_for()` calls that make use of `_anchor`, where appropriate.

For the more specific attributes needed in the `url_for()` calls, I've just adjusted to [pass the individual](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1229/files#diff-17106716ca87c4428dbe8ea4faee1d08fd638866502948cdedd16138c6aee428R72-R79) `ReturnLocation` attributes instead of unpacking the entire object with `to_dict()` - I couldn't see a way around this without adding more parameters in `to_dict()` calls, but would appreciate any steer/guidance on this if possible (and for the other changes).

### Checklist


liamtoozer:
[Create a ReturnLocation object to hold return to parameters (3 SP)](https://trello.com/c/T2xUsL8u/5893-create-a-returnlocation-object-to-hold-return-to-parameters-3-sp)

katie-gardner (commented):
Just going to test functionally but from a first pass this is looking good 👍

app/views/contexts/summary/calculated_summary_block.py
```diff
@@ -86,16 +82,18 @@ def __init__(
...
             block_id=self.id,
             list_name=self._list_name,
             list_item_id=self._list_item_id,
-            return_to=self._return_to,
+            return_to=self._return_location.return_to,
             return_to_answer_id=self.id,
-            return_to_block_id=self._return_to_block_id,
+            return_to_block_id=self._return_location.return_to_block_id,
             return_to_list_item_id=return_to_list_item_id,
             _anchor=self.id,
```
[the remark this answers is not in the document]
liamtoozer: Ah good spot! I'll have a look to see if anything similar happens elsewhere and will pop a card up for it

app/views/contexts/summary/group.py
```diff
@@ -25,8 +26,8 @@
 )
 
 
+# pylint: disable=too-many-locals
```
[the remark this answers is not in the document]
liamtoozer: I think I removed this initially for the method but then put it back at class-level by mistake. Good spot, thanks!

katie-gardner (approved):
Tested a variety of navigation with staging and this up side by side and found no functional difference, changes look sound 🎉

app/questionnaire/return_location.py
```diff
@@ -0,0 +1,27 @@
+from dataclasses import asdict, dataclass
+
+
+@dataclass(kw_only=True, frozen=True)
+class ReturnLocation:
+    """
+    Used to store return locations in the questionnaire.
+
+    return_to: The name of the section to return to
```
berroar, line 9: Not sure this is correct that it is the name of the section, the return_to is often a summary page or a calculated summary page etc

tests/app/questionnaire/test_return_location.py
```diff
@@ -0,0 +1,37 @@
+from app.questionnaire.return_location import ReturnLocation
+
+
+def test_return_location_to_dict():
+    return_location = ReturnLocation(
```
berroar, line 5: Minor but could reuse this object as it is the same for both tests? Or parametrize?

berroar (approved):
Looks good, seems to be working functionally!

app/questionnaire/return_location.py
```diff
@@ -0,0 +1,27 @@
...
+    return_to_block_id: The block_id of the block to return to
+    return_to_answer_id: The answer_id of the answer to return to
+    return_to_list_name: The list_name to return to if the location is associated with a list
+    return_to_list_item_id: The list_item_id to return to if the location is associated with a list
+    """
+
+    return_to: str | None = None
+    return_to_block_id: str | None = None
+    return_to_answer_id: str | None = None
+    return_to_list_name: str | None = None
```
katie-gardner, line 19: Just realised this is a param we didn't go with in the end - no need for `return_to_list_name` its only being used in your tests

app/questionnaire/return_location.py
```diff
@@ -0,0 +1,27 @@
...
+    return_to_list_item_id: str | None = None
+
+    def to_dict(self, anchor_return_to_answer_id: bool = False) -> dict:
```
MebinAbraham, line 22: Minor, the arg should read more like a boolean, maybe `answer_id_is_anchor` or `is_answer_id_anchor`, something along those lines.

app/questionnaire/return_location.py
liamtoozer: Ah, that's a good point, I will update it to `The name of the summary page to return to` which hopefully catches each name?

katie-gardner: I think the 'type' of summary page might be clearer

liamtoozer: Great, thanks both 👍

tests/app/questionnaire/test_return_location.py
```diff
@@ -0,0 +1,33 @@
...
+    assert TEST_RETURN_LOCATION_OBJECT.to_dict() == {
+        "return_to": "test-return-to",
+        "return_to_block_id": "test-return-to-block-id",
+        "return_to_answer_id": "test-return-to-answer-id",
+        "return_to_list_item_id": "return-to-list-item-id",
+    }
+
+
+def test_return_location_to_dict_with_anchor_return_to_answer_id():
+    return_location = ReturnLocation(
```
katie-gardner, line 21: Forgot the new shared test object for this one

PR #1232, opened 2023-10-20
Remove question text from GA event

berroar:

Updates the GA event in the `preview` template so that question text is no longer being passed through as a label.

### How to review
Check `Preview` functionality still works/nothing is broken by this change.

### Checklist


berroar:
[Prevent question text being sent to GA in preview feature](https://trello.com/c/2I7jJ8h6/5912-prevent-question-text-being-sent-to-ga-in-preview-feature)

petechd (approved):
Preview still works as expected.

templates/partials/introduction/preview.html
```diff
@@ -23,7 +23,6 @@ <h2 class="ons-u-mt-m">{{ intro.title }}</h2>
           "data-ga": "click",
           "data-ga-category": "Preview Survey",
           "data-ga-action": "Open panel",
-          "data-ga-label": question.question
```
rmccar, line 26: Don't we still want to send something in the `data-ga-label` field?

berroar: Didn't add one as wasn't sure it was required and looks like we don't always attach a label in other events? But can add something here if it is needed?

katie-gardner (commented):
Preview working as expected 👍 
Is `individual-response-guidance.html` subject to this problem too? also wondering how the `definition` in `defintion.html` is called and if that has been checked?

berroar:
> Preview working as expected 👍 Is `individual-response-guidance.html` subject to this problem too? also wondering how the `definition` in `defintion.html` is called and if that has been checked?

Individual Response was only used during Census and isn't being used any more, so that one shouldn't be an issue. Will check definition 👍

templates/partials/introduction/preview.html
rmccar: I think this is something we need to look into across eQ. We need to be consistent about what we send and what attributes we are setting. I think theres some stuff getting set that isn't even getting used too

berroar:
> Preview working as expected 👍 Is `individual-response-guidance.html` subject to this problem too? also wondering how the `definition` in `defintion.html` is called and if that has been checked?

@katie-gardner The definition title used in the GA label just takes the literal definition title from the schema (example in `test_question_definition`), but the title field doesn't support placeholders so I can't see why there would be any sensitive data here

templates/partials/introduction/preview.html
berroar: Yeah there is another ticket at the moment laying the groundwork for this, checkin what we are sending 👍

rmccar (approved):

PR #1233, opened 2023-10-20
Update to DS v66

rmccar:
Updates the DS version to 66.0.1 and fixes the breaking change where the form element has now been removed from the base page template by adding it in runner using the `bodyStart` and `bodyEnd` blocks insted of setting it in params.

### How to review
- Launch some schemas to see that runner works and looks as expected
- Check that the form element appears in the html of a question page correctly

### Checklist


rmccar:
[Update eQ to v66 of the DS](https://trello.com/c/2JnsOIIB/5932-update-eq-to-v66-of-the-ds)

Yuyuutsu (commented):
I can see the form changes 👍 . 
Runner looks good, only difference is that the space between the nav items has gotten smaller. As well as space with `save and exit survey text` with the icon has gotten smaller too.

katie-gardner (commented):
Working as before, looks fine to me 👍 
Only thing, is that with the changes the form has `novalidate="null"` instead of previously being `novalidate` was this intentional? Was this previously defaulting to true and that has now been lost?

rmccar:
> I can see the form changes 👍 . Runner looks good, only difference is that the space between the nav items has gotten smaller. As well as space with `save and exit survey text` with the icon has gotten smaller too. Is this expected?

I see the issue you are talking about. I compared with staging and there does seem to be extra space around the links in staging. I think this is down to the html not being minimised in staging so still has some whitespace around it which is causing this issue. Its the same on the DS theres whitespace around the html but I think this should be dealt with as a separate issue as its nothing to do with the DS version change. I have even tested it without the version change and the result is the same.

rmccar:
> Working as before, looks fine to me 👍 Only thing, is that with the changes the form has `novalidate="null"` instead of previously being `novalidate` was this intentional? Was this previously defaulting to true and that has now been lost?

You're right it should just be `novalidate` Ive updated this now

katie-gardner (approved):
Form now fixed 👍

PR #1235, opened 2023-10-24
Refactor dependencies splitting answers and lists

katie-gardner:
This PR has made some modifications and simplifications to the way runner handles dependencies, primarily, splitting up list and answer dependencies, to allow bug cards relating to incorrect evalulation of progress to be completed more easily.

(During the process of this PR this area was documented [on confluence](https://confluence.ons.gov.uk/display/SDC/eQ%3A+Questionnaire+Dependencies))

### Changes Summary
This PR has added two new maps to `questionnaire_schema`

- `list_dependenices`:
 Previously when something dependent on a list, it was treated as depending on the add block and the remove block of that list. Now that dependency is captured against the list name, allowing both supplementary lists and list collector lists dependencies to be handled the same. 

- `when_rules_section_dependencies_by_list`
Previously, when rules relating to answers and progress were captured on initialisation, but list dependencies were handled at run time in `questionnaire_store_updater`. This map replaces `_section_dependencies_by_list_name` which is what got populated at run time. It is now populated alongside answers and progress for consistency and simplicity.

A method `get_list_name_for_answer_id` was added. This is being used when handling dependencies of calculated summaries to determine if the answer is repeating. If the answer is repeating, it gets the list it repeats over, as the calculated summary will also need to depend on that.

Whilst observing calculated summary dependencies, I noticed `_get_calculated_summary_section_dependencies` was looking up the section for each answer individually, rather than just the section of the calculated summary, which it doesn't need to do, so this was simplfied.

`get_values_for_key` is now used less and was causing code coverage issues. So it was moved to utilities as it doesn't relate to questionnaire schema, and it now reuses a method that does almost the same thing.

### How to review
Review the changes, and the trello ticket and check that the changes make sense and have achieved the goal of the ticket. 

Functionally this should not change anything, except that the edit-block for a list no longer requires revisiting of dynamic answers (unless those answers specifically depend on the edit block)

Test using complex schemas like `test_progress_value_source_repeating_sections_chained_dependencies` and check there is no regression to progress updating.

### Checklist


app/questionnaire/questionnaire_store_updater.py
```diff
@@ -159,27 +161,35 @@ def remove_list_item_data(self, list_name: str, list_item_id: str) -> None:
...
-                *list_collector.get("repeating_blocks", []),
+        self._capture_block_dependencies_for_list(list_name)
+        section_ids = self._schema.get_when_rule_section_dependencies_for_list(
+            list_name
+        )
+        section_ids.update(
+            self._schema.list_collector_section_ids_by_list_name[list_name]
+        )
+
+        for (
```
berroar, line 179: Minor but can use the returned section_key here e.g. `for section_key in self.started_section_keys(section_ids=section_ids)`?

app/questionnaire/questionnaire_schema.py
```diff
@@ -138,9 +142,13 @@ def placeholder_transform_section_dependencies_by_block(
         return self._placeholder_transform_section_dependencies_by_block
 
     @cached_property
-    def answer_dependencies(self) -> ImmutableDict[str, set[AnswerDependent]]:
+    def answer_dependencies(self) -> ImmutableDict[str, set[Dependent]]:
         return ImmutableDict(self._answer_dependencies_map)
 
+    @cached_property
+    def list_dependencies(self) -> ImmutableDict[str, set[Dependent]]:
```
berroar, line 149: Do we want to add a test for this in test_questionnaire_schema as we do for some of the other dependency maps?

app/questionnaire/questionnaire_store_updater.py
katie-gardner: Yeah I prefer how this looks, thanks 👍

app/questionnaire/questionnaire_schema.py
katie-gardner: Good idea 👍 have added a test for it and I moved the existing test for the when rules one and renamed it given that ive changed the method

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -159,27 +161,30 @@ def remove_list_item_data(self, list_name: str, list_item_id: str) -> None:
...
+
+        for section_key in self.started_section_keys(section_ids=section_ids):
+            # Only add sections which are repeated sections for this list, or the section in which this list is collected
+            # Prevents list item progresses being added as dependants as these are captured by started_section_keys(section_ids=section_ids)
+            if (
+                not self._schema.get_repeat_for_section(section_key.section_id)
+                and section_key.list_item_id
+            ):
+                continue
+            self.dependent_sections.add(DependentSection(**section_key.to_dict()))
```
berroar, line 187: Not necessarily related to this change, but could use SectionKey in the DependentSection class, rather than passing the individual vars to keep things consistent. May be overkill though

katie-gardner: As in 
```python
class DependentSection(NamedTuple):
    section_key: SectionKey
    is_complete: bool | None = None
```
?
I agree that'd be nice here but I worry it'd make other initialisations of `DependentSection` worse, most of the use cases aren't with a section_key. I'd probably lean the other way in this case

berroar: Yeah that's what I meant, yeah I can see lots of places where it might be slightly worse, but kind of made sense to me as dependent section has a method to return a section key as well. Only other thing is that if we made this change there would be barely any difference between the DependentSection and SectionKey classes.

berroar:
Found quite an odd bug in the `test_progress_value_source_repeating_sections_chained_dependencies`, if you go into the first section, you cannot complete the question. Hopefully it's not just something locally for me 😅, but seems to work on Main

katie-gardner:
> Found quite an odd bug in the `test_progress_value_source_repeating_sections_chained_dependencies`, if you go into the first section, you cannot complete the question. Hopefully it's not just something locally for me 😅, but seems to work on Main

Oh no 😄 what does it do? keep redirecting you to same question? I'm not able to reproduce this, might need a demo if it persists

berroar:
> > Found quite an odd bug in the `test_progress_value_source_repeating_sections_chained_dependencies`, if you go into the first section, you cannot complete the question. Hopefully it's not just something locally for me 😅, but seems to work on Main
> 
> Oh no 😄 what does it do? keep redirecting you to same question? I'm not able to reproduce this, might need a demo if it persists

Yeah I can't seem to escape from the question page 😕

app/questionnaire/questionnaire_schema.py
```diff
@@ -914,6 +837,19 @@ def is_repeating_answer(
             or self.is_answer_dynamic(answer_id)
         )
 
+    def get_list_name_for_answer_id(self, answer_id: str) -> str | None:
+        """
+        Returns the name of the list if the answer is dynamic/repeating and none otherwise
+        """
+        block_id: str = self.get_block_for_answer_id(answer_id)["id"]  # type: ignore
```
petechd, line 844: We probably need to add a short description of type ignore here and before line 850.

app/questionnaire/questionnaire_schema.py
```diff
@@ -503,19 +500,9 @@ def _update_answer_dependencies_for_calculated_summary_dependency(
...
-                # Type ignore: answer_id is valid so block must exist
-                block_id: str = self.get_block_for_answer_id(answer_id)["id"]  # type: ignore
-                if block_id in self._list_dependent_block_additional_dependencies:
-                    for (
-                        list_block_id
-                    ) in self._list_dependent_block_additional_dependencies[block_id]:
-                        self._answer_dependencies_map[list_block_id].add(
-                            answer_dependent
-                        )
+            if list_name := self.get_list_name_for_answer_id(answer_id):
```
petechd, line 503: Should the name of the method change from `_update_answer_dependencies_for_calculated_summary_dependency` now since we update list dependecies map as well? Should the docstring change as well?

app/questionnaire/questionnaire_schema.py
```diff
@@ -503,19 +500,9 @@ def _update_answer_dependencies_for_calculated_summary_dependency(
...
+                # dynamic/repeating answers means the calculated summary also depends on the list those answers loop over
+                self._list_dependencies_map[list_name].add(answer_dependent)
             self._answer_dependencies_map[answer_id].add(answer_dependent)
 
     def _update_answer_dependencies_for_grand_calculated_summary(
```
petechd, line 508: Same here? Do we want to rename it so we mention list dependecies? Should it be general `dependencies`?

app/questionnaire/questionnaire_schema.py
petechd: Does it make sense to call it `_update_dependencies_for_calculated_summary_dependency`?

app/questionnaire/questionnaire_schema.py
```diff
@@ -616,54 +613,23 @@ def _update_answer_dependencies_for_value_source(
...
     ) -> None:
-        """Updates dependencies for a block depending on a list collector
-
-        This method also stores a map of { block_depending_on_list_source -> {add_block, remove_block} }, because:
-        blocks like dynamic_answers, don't directly need to depend on the add_block/remove_block,
-        but a block depending on the dynamic answers might (such as a calculated summary)
-        """
-        list_collectors = self.get_list_collectors_for_list(
-            for_list=list_name,
+        """Store the given block as a dependency of the list"""
```
petechd, line 616: If the functionality is really simple maybe docstring is not needed anymore? Should we used it more for describing our design decisions?

app/questionnaire/questionnaire_schema.py
katie-gardner: Yeah good point - updated the docstring too

app/questionnaire/questionnaire_schema.py
katie-gardner: Yep, good idea, have fixed this in quite a few places

app/questionnaire/questionnaire_schema.py
katie-gardner: Added one comment to cover both cases

app/questionnaire/questionnaire_schema.py
katie-gardner: Given it is now so simple, and is actually only used in one place, I've removed it entirely and put the code into the single place it is used directly

petechd (approved):

app/utilities/mappings.py
```diff
@@ -35,3 +35,8 @@ def _handle_sequence(value: Sequence) -> Generator[Mapping, None, None]:
                 yield from get_mappings_with_key(key, v, ignore_keys)
             if isinstance(v, Sequence):
                 yield from _handle_sequence(v)
+
+
+def get_values_for_key(key: str, data: Mapping | Sequence) -> Generator:
```
liamtoozer, line 40: Minor question (not related to this PR really, as I think it might have been there before the type hinting updates) - would the `Generator` yield type narrow down to a `Mapping`? E.g. something like `-> Generator[Mapping, None, None]`?

liamtoozer (commented):
I've tested with a few schemas and everything seems to work great for me!

app/utilities/mappings.py
katie-gardner: In this case its getting all the values out of a `Mapping` rather than the mappings themself like the one above it so for example if data is
```
data = { "store": [1, 2, 3], "test": { "store": 1 }, "nested_1": [{"nested_2": { "store": "xyz" }}] }
```
the get values for key would generate `[1, 2, 3]` then `1` then `"xyz"` so it could be anything and I'm not able to narrow down the type

liamtoozer: Ah I see, cool that makes sense - thanks 👍

app/questionnaire/questionnaire_store_updater.py
katie-gardner: Open for some team views on this - I have tried adding this throughout locally, I'm not keen to keep it due to the number of instances where we'd have something like `DependentSection(SectionKey(section_id, list_item_id))` 
Where repeating section dependencies are done by iterating through list store and using list item with section, so the two aren't already paired by a section key. 
But there isn't that much in it and if it makes more sense to everyone and they prefer it for following dependencies I can go with it

app/utilities/mappings.py
MebinAbraham, line 40: Minor, I suggest we enforce keyword only arg at least for data.

app/questionnaire/questionnaire_schema.py
```diff
@@ -431,94 +448,59 @@ def _get_answers_by_id(self) -> dict[str, list[ImmutableDict]]:
...
-        the calculated summary block needs to depend on the `remove_block` and `add_block` for the list
-        so that adding/removing items requires re-confirming the calculated summary
+        For a block that depends on a calculated summary block, add the block as a dependency of each of the calculated summary answers
+        Similarly if the calculated summary depends on a list, then add the block as a dependency of the list
         """
         calculated_summary_answer_ids = get_calculated_summary_answer_ids(
             calculated_summary_block
         )
-        answer_dependent = self._get_answer_dependent_for_block_id(
+        answer_dependent = self._get_dependent_for_block_id(
```
MebinAbraham, line 494: The variable should be renamed.

app/questionnaire/questionnaire_schema.py
```diff
@@ -603,67 +587,34 @@ def _update_answer_dependencies_for_value_source(
                 )
 
                 for answer_id_for_block in answer_ids_for_block:
-                    self._answer_dependencies_map[answer_id_for_block] |= {
-                        self._get_answer_dependent_for_block_id(
+                    self._answer_dependencies_map[answer_id_for_block].add(
+                        dependent := self._get_dependent_for_block_id(
                             block_id=block_id, answer_id=answer_id
                         )
```
MebinAbraham, line 593: Existing but this should be outside the for loop, it is the same for all answers.

app/questionnaire/questionnaire_schema.py
```diff
@@ -603,67 +587,34 @@ def _update_answer_dependencies_for_value_source(
...
-                        else self._get_answer_dependent_for_block_id(block_id=block_id)
-                    }
-                self._list_dependent_block_additional_dependencies[block_id] = set(
-                    answer_ids_for_block
-                )
-                # removing an item from a list will require any dependent calculated summaries to be re-confirmed, so cache dependencies
-            if remove_block_id := self.get_remove_block_id_for_list(list_name):
-                self._list_dependent_block_additional_dependencies[block_id].update(
-                    self.get_answer_ids_for_block(remove_block_id)
-                )
```
MebinAbraham, line 653: 🎉

app/questionnaire/questionnaire_schema.py
```diff
@@ -914,6 +822,20 @@ def is_repeating_answer(
             or self.is_answer_dynamic(answer_id)
         )
 
+    def get_list_name_for_answer_id(self, answer_id: str) -> str | None:
+        """
+        Returns the name of the list if the answer is dynamic/repeating and none otherwise
```
MebinAbraham, line 827: Worth being more explicit here, `if answer is dynamic, in a repeating block or section.`

app/questionnaire/questionnaire_schema.py
```diff
@@ -1375,16 +1313,16 @@ def _get_rules_section_dependencies(
...
+                    current_section_id, dependent_answer_ids
+                )
+            )
 
-                if section_id != current_section_id:
-                    self._when_rules_section_dependencies_by_answer[answer_id].add(
-                        current_section_id
-                    )
-                    section_dependencies.add(section_id)  # type: ignore
+            for list_name in dependent_list_names:
```
MebinAbraham, line 1322: The previous one was ignoring list dependencies for rules in `["question_variants", "content_variants"]` keys, now I'm wondering why.

app/questionnaire/questionnaire_schema.py
```diff
@@ -1407,51 +1345,28 @@ def _get_rules_section_dependencies(
...
-                        sources=calculated_summary_sources,
-                    )
-                )
+                section_dependencies: set[str] = set()
+                for source in sources:
+                    if source["source"] == "calculated_summary":
+                        section_dependencies.add(
+                            # Type ignore: Validator ensures a valid identifier so section will exist
+                            self.get_section_id_for_block_id(source["identifier"])  # type: ignore
+                        )
```
MebinAbraham, line 1364: Minor, any reason for the formatting change?. Previous one was done to reduce nesting.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -159,27 +161,30 @@ def remove_list_item_data(self, list_name: str, list_item_id: str) -> None:
...
+        section_ids.update(
+            self._schema.list_collector_section_ids_by_list_name[list_name]
+        )
+
+        for section_key in self.started_section_keys(section_ids=section_ids):
+            # Only add sections which are repeated sections for this list, or the section in which this list is collected
+            # Prevents list item progresses being added as dependants as these are captured by started_section_keys(section_ids=section_ids)
+            if (
+                not self._schema.get_repeat_for_section(section_key.section_id)
+                and section_key.list_item_id
```
MebinAbraham, line 184: Minor, would be more performant to flip this logic, check if we have list item id, if so, fetch repeat list name.

app/questionnaire/questionnaire_store_updater.py
MebinAbraham: I think in this specific bit, passing section key does look like a nice win-win but I agree that most initialisation don't have a section key and would therefore be worse of for readability as they become

```py
DependentSection(
	SectionKey(section_id, list_item_id)
)
```

So in this instance, my preference is as is.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -401,11 +413,23 @@ def update_answers(
             if self._update_answer(answer_id_to_use, list_item_id_to_use, answer_value):
                 self._capture_section_dependencies_for_answer(answer_id_to_use)
                 self._capture_block_dependencies_for_answer(answer_id_to_use)
-                self.capture_progress_section_dependencies()
+                answers_updated = True
+
+        if answers_updated:
```
MebinAbraham, line 418: Don't a var, can use `if self._answer_store.is_dirty:`?

app/utilities/mappings.py
```diff
@@ -35,3 +35,8 @@ def _handle_sequence(value: Sequence) -> Generator[Mapping, None, None]:
...
+    for mapping in get_mappings_with_key(key, data):
```
MebinAbraham, line 41: Same as above.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -131,12 +131,14 @@ def add_primary_person(self, list_name: str) -> str:
...
         # the location must be removed from the progress store.
         self.remove_completed_location(self._current_location)
+        self._capture_dependencies_for_list_change(list_name)
 
         return self._list_store.add_list_item(list_name, primary_person=True)
 
     def add_list_item(self, list_name: str) -> str:
         new_list_item_id = self._list_store.add_list_item(list_name)
         self.remove_completed_relationship_locations_for_list_name(list_name)
+        self._capture_dependencies_for_list_change(list_name)
```
MebinAbraham, line 141: This would be better done at the source, i.e handler, or SDS. Being here means the `_capture_dependencies_for_list_change` is called multiple times when SDS is adding list items, or removing list items, even if they are all for the same list name.

app/views/handlers/list_add_question.py
```diff
@@ -54,9 +54,6 @@ def handle_post(self) -> None:
         self.questionnaire_store_updater.update_answers(
             self.form.data, self._list_item_id
         )
-
-        self.capture_dependent_sections_for_list(self.parent_block["for_list"])
```
MebinAbraham, line 58: We probably want this here, see comment on `add_list_item` method.

app/views/handlers/primary_person_list_collector.py
```diff
@@ -35,21 +35,17 @@ def handle_post(self) -> None:
             self._primary_person_id = (
                 self.questionnaire_store_updater.add_primary_person(list_name)
             )
-
-            self.capture_dependent_sections_for_list(list_name)
```
MebinAbraham, line 39: Same as above.

app/views/handlers/primary_person_list_collector.py
```diff
@@ -35,21 +35,17 @@ def handle_post(self) -> None:
...
-
-            self.questionnaire_store_updater.update_same_name_items(
-                list_name, self.rendered_block.get("same_name_answer_ids")
-            )
             # This method could determine the current section's status incorrectly, as
             # the call to update the answer store takes place in
             # `super().handle_post()`. The section status will eventually get
             # determined correctly when the parent class' `update_section_status`
             # method is called.
-            self.capture_dependent_sections_for_list(list_name)
```
MebinAbraham, line 54: Same as above.

app/utilities/mappings.py
katie-gardner: Sure, whats your general view for enforcing? always do it for 2+? Is there somewhere we have/can add a general rule for this? 

The approach I have been taking is whenever there is more than one argument, enforce kwargs, unless theres only a couple of argument types and the types don't overlap, such as here, because mypy would pick that up. (Or exceptional cases like SectionKey where its very common and enforcing would bloat) but I'm happy to go with enforcing anytime its more than 1

app/questionnaire/questionnaire_schema.py
katie-gardner: I made the changes as I thought the extra function made it harder to follow and could be inlined
That said, this can be nested less. I've removed it by doing a comprehension instead.

app/questionnaire/questionnaire_store_updater.py
katie-gardner: Ah nice done 👍

app/utilities/mappings.py
katie-gardner: 👍 assuming this one means for `get_mappings_with_key`? done for both anyway

app/questionnaire/questionnaire_store_updater.py
katie-gardner: Yeah, I wanted to minimise what callers need to do, but agree this probably does get called too much if done like that - made it public and moved to callers 👍

app/questionnaire/questionnaire_schema.py
katie-gardner: Going to try an example as I'm also not sure why those would have been skipped

katie-gardner: It looks like this was probably done for performance benefit - the question variants have the same question-ids, block-ids, and answer-ids so the progress is not going to be affected by adding or removing to a list, the changed answer will just relate to a question/answer with different text etc. 
In which case I could probably do the same thing? It saves a few checks but are there edge cases where we do want to check those sections? currently i cant think of any so ignoring makes sense

Yuyuutsu (commented):
Concourse doc and changes look good to me 👍

app/utilities/mappings.py
MebinAbraham: We don't have a set rule (yet). My preference is when the argument order becomes unclear. I wasn't suggesting enforcing keyword args for `key` given the function name end with `key`, so I was thinking `get_values_for_key(key: str, *, data: Mapping | Sequence) -> ....`. This one was possibly fine due to function name, i.e name implies first should be key, then there is only 1 other.

We can come up with our preference for when we create a style guide. I agree 2+ is a good rule, or if function name doesn't make clear the arg order for the first two.

> because mypy would pick that up
Mypy won't pick up bad test cases in which the order is different as we don't enforce mypy for tests.

app/questionnaire/questionnaire_schema.py
MebinAbraham: Since it is not done at run time anymore, you can't ignore it without making it work at runtime or having two maps?. As only it shouldn't be ignored in all places?

MebinAbraham: Actually, it shouldn't matter, it doesn't affect completeness therefore should be safe to ignore variants. We don't reshow pages if "placeholders" change, this would be the same concept. Until they become a requirement, we can ignore variants.

app/utilities/mappings.py
katie-gardner: Have taken out the enforcing of key, I think that makes sense, good with 2+/ unclear

app/questionnaire/questionnaire_schema.py
katie-gardner: Have ignored them for when rules and added an extra test for it

MebinAbraham (approved):
👍

PR #1241, opened 2023-10-30
Update test_variants_question schema question to make Number type mandatory

liamtoozer:
When submitting empty/whitespace for a numeric response with the test schema `test_variants_question.json`,  runner emits an HTTP 500. This is because the number is needed for routing in a `ConfirmationQuestion` later on in the survey, so it cannot be empty

**Note**: Following on from a tech session discussion, this is a very narrow edge case that would not really add much benefit in protecting from schema validation. It would require significant effort to create a path to protect against this scenario

### How to review
Check that the question is mandatory and that page navigation works as expected. I've checked the other test schemas to see if this crops up elsewhere, but seems okay - might be worth double-checking in the review in case I've missed anything?

### Checklist


berroar (approved):
Fixes the issue 👍 Minor, but the journey for that schema is a bit strange, selecting "No" on the confirmation question `Are you over 16?` results in being stuck on the same page. So could update that but looks like the journey was always like that for this schema.

katie-gardner (approved):
Issue fixed 👍 
Agree its a slightly odd journey but not really in scope so all good from me either way

PR #1242, opened 2023-10-30
Support for multiple return to answer ids

petechd:
In this PR we modify `return_to_answer_id` to be a list of calculated summary ids and answer ids. The continue button forces new behaviour (returns to the calculated summary anchored at answer_id and pop it off the list so it becomes `return_to_answer_id=calculated_summary_id`). This change is similar to what we did to `return_to_block_id` and makes the page focus on the right elements on calculated summaries and grand calculated summaries.

### How to review
Existing tests were modified and some new assertions were added where applicable (description of the test changed as well).

### Checklist


petechd:
[Support for multiple return to answer ids](https://trello.com/c/CFWFZ4Ce/5797-support-for-multiple-return-to-answer-ids)

app/questionnaire/router.py
```diff
@@ -321,12 +321,19 @@ def _get_return_to_for_grand_calculated_summary(
...
+                return_to_answer_ids = return_location.return_to_answer_id.split(",")
+                return_to_answer_id = "".join(return_to_answer_ids.pop())
+
             return url_for(
                 "questionnaire.block",
                 block_id=return_location.return_to_block_id,
                 list_item_id=list_item_id,
                 list_name=list_name,
-                _anchor=return_location.return_to_answer_id,
+                _anchor=str(return_to_answer_id)
```
berroar, line 334: How come we need to cast to sting here? Could it not be simplified to `_anchor=return_to_answer_id or return_location.return_to_answer_id`?

tests/functional/spec/features/grand_calculated_summary/grand_calculated_summary_overlapping_answers.spec.js
```diff
@@ -40,7 +40,7 @@ describe("Feature: Grand Calculated Summary", () => {
       await click(GrandCalculatedSummaryShoppingPage.submit());
     });
 
-    it("Given I edit an answer that is only used in a single calculated summary, I am routed back to the calculated summary and then the grand calculated summary", async () => {
+    it("Given I edit an answer that is only used in a single calculated summary, I am routed back to the calculated summary and then the grand calculated summary and the correct fields are focused", async () => {
```
berroar, line 43: I know this was pre-existing but a lot of these tests should follow the Given, When, Then pattern, a lot seem to be Given, Given 😅

app/questionnaire/router.py
petechd: I was trying to fix type hint and tried different conditions so it best fits but forgot we don’t need this type conversion anymore after I found perfect solution for removing type ignores. Fixed now, thanks.

tests/functional/spec/features/grand_calculated_summary/grand_calculated_summary_overlapping_answers.spec.js
petechd: Yeah, I am not sure here, I touch only one, two assertion in each file and only one test description in all of spec files, does it mean I should change all of the test descriptions in a given spec? Keep it the same and create a separate task? Change the description to use "given, when, then" since it's one test I am editing and it will stand out?

katie-gardner (changes requested):
Have only functionally reviewed so far as found some problems in `test_grand_calculated_summary_overlapping_answers`

From `calculated-summary-1` if I click a change link, and then press continue, the url has `return_to_answer_id=q1-a1` but it should be None, because this goes back to the cs page anchored at q1-a1 so its not an answer to return to anymore

From the grand calculated summary, I was expecting `return_to_answer_id=q1-a1,calculated-summary-1` on the answer page after clicking two change links, similar to how the other params have a list, but instead I only have `q1-a1` on the answer, its not following the same pattern as `return_to` and `return_to_block_id`

petechd:
> Have only functionally reviewed so far as found a couple of issues in `test_grand_calculated_summary_overlapping_answers`
> 
> From `calculated-summary-1` if I click a change link, and then press continue, the url has `return_to_answer_id=q1-a1` but it should be None, because this goes back to the cs page anchored at q1-a1 so its not an answer to return to anymore
> 
> From the grand calculated summary, I was expecting `return_to_answer_id=q1-a1,calculated-summary-1` on the answer page after clicking two change links, similar to how the other params have a list, but instead I only have `q1-a1` on the answer, its not following the same pattern as `return_to` and `return_to_block_id` (which I think might be what causes the above)

First issue from your comment addressed, need to spend more time on the second.

petechd:
> Have only functionally reviewed so far as found a couple of issues in `test_grand_calculated_summary_overlapping_answers`
> 
> From `calculated-summary-1` if I click a change link, and then press continue, the url has `return_to_answer_id=q1-a1` but it should be None, because this goes back to the cs page anchored at q1-a1 so its not an answer to return to anymore
> 
> From the grand calculated summary, I was expecting `return_to_answer_id=q1-a1,calculated-summary-1` on the answer page after clicking two change links, similar to how the other params have a list, but instead I only have `q1-a1` on the answer, its not following the same pattern as `return_to` and `return_to_block_id` (which I think might be what causes the above)

The second issue fixed now (hopefully).

app/questionnaire/router.py
```diff
@@ -321,12 +321,19 @@ def _get_return_to_for_grand_calculated_summary(
             ),
             routing_path,
         ):
+            return_to_answer_id = None
+            if return_location.return_to_answer_id:
+                return_to_answer_ids = return_location.return_to_answer_id.split(",")
+                return_to_answer_id = "".join(return_to_answer_ids.pop())
```
katie-gardner, line 327: because of the comment below, on the calculated summary page, you've ended up with two answer ids in the `return_to_answer_id` but when returning to a grand calculated summary, this currently should never be in a position where there are 2 - so I think the changes to this function won't be needed if the other one is fixed

app/views/contexts/summary/answer.py
```diff
@@ -64,17 +64,24 @@ def _build_link(
...
+            return_to=return_location.return_to,
+            list_item_id=list_item_id,
+            is_in_repeating_section=is_in_repeating_section,
+        )
+        if (
+            return_location.return_to_answer_id
+            and return_to_answer_id
+            not in return_location.return_to_answer_id.split(",")
+        ):
+            return_to_answer_id += f",{return_location.return_to_answer_id}"  # type: ignore
```
katie-gardner, line 77: I think any extra logic should probably be in the `_return_to_answer_id` method

tests/app/questionnaire/test_router.py
```diff
@@ -1566,6 +1567,7 @@ def test_return_to_grand_calculated_summary_from_repeating_answer(
             return_to="grand-calculated-summary",
             block_id="calculated-summary-6",
             return_to_block_id="grand-calculated-summary-3",
+            return_to_answer_id="calculated-summary-6,calculated-summary-6",
```
katie-gardner, line 1570: 👀

tests/app/questionnaire/test_router.py
```diff
@@ -793,6 +793,7 @@ def test_return_to_grand_calculated_summary_from_answer(
             return_to="grand-calculated-summary",
             block_id="distance-calculated-summary-1",
             return_to_block_id="distance-grand-calculated-summary",
+            return_to_answer_id="distance-calculated-summary-1,distance-calculated-summary-1",
             _anchor="distance-calculated-summary-1",
```
katie-gardner, line 797: same comment again, this test looks off

katie-gardner (changes requested):
The return to answer id can still have two many items in it on a calc summary page, but have commented where I think causes it, looks correct on answer page now

tests/app/questionnaire/test_router.py
petechd: Removed return answer id now as it shouldn't have one in url?

tests/app/questionnaire/test_router.py
petechd: This issue disappears when I implement your code suggestion, so resolved.

app/questionnaire/router.py
```diff
@@ -369,18 +376,27 @@ def _get_return_to_for_calculated_summary(
...
-            # Type ignore: return_location.return_to will always be populated at this point
-            return_to_remaining = ",".join(return_location.return_to.split(",")[1:]) or None  # type: ignore
+            # Type ignore: return_location.return_to and return_location.return_to_answer_id will always be populated at this point
+            return_to = ",".join(return_location.return_to.split(",")[1:]) or None  # type: ignore
+            return_to_answer_id = (
+                ",".join(
+                    [return_location.return_to_answer_id.split(",")[0], block_id]  # type: ignore
+                )
+                if remaining
+                else None
```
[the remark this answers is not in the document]
petechd: Using your code suggestion, slightly modified, seems to solve all our issues.

app/questionnaire/router.py
petechd: Looks like the suggestion below fixed it.

app/views/contexts/summary/answer.py
petechd: Moved it there now.

app/views/contexts/summary/answer.py
```diff
@@ -64,17 +64,24 @@ def _build_link(
...
     ) -> str:
+        return_to_answer_id = self._return_to_answer_id(
+            return_to=return_location.return_to,
+            list_item_id=list_item_id,
+            is_in_repeating_section=is_in_repeating_section,
+        )
+        if (
+            return_location.return_to_answer_id
+            and return_to_answer_id
+            not in return_location.return_to_answer_id.split(",")
```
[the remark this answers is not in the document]
petechd: We ended up with the same calculated summary id (duplication) in `return_answer_id`.

tests/app/views/contexts/summary/test_answer.py
```diff
@@ -6,20 +6,24 @@
...
+        (None, None, False, None),
+        ("calculated-summary", "total", False, None),
+        ("section-summary", None, True, None),
+        (None, None, True, None),
+        ("calculated-summary", "total", True, None),
+        ("calculated-summary", "total", True, "calculated-summary-1"),
     ],
 )
-def test_create_answer(return_to, return_to_block_id, is_in_repeating_section):
+def test_create_answer2(
```
katie-gardner, line 20: I think you might have meant to change the function name back to not having the `2`

app/questionnaire/router.py
```diff
@@ -369,18 +369,25 @@ def _get_return_to_for_calculated_summary(
             return_to_block_id = ",".join(remaining) if remaining else None
 
             # remove first item and return the remaining ones
-            # Type ignore: return_location.return_to will always be populated at this point
-            return_to_remaining = ",".join(return_location.return_to.split(",")[1:]) or None  # type: ignore
+            # Type ignore: return_location.return_to and return_location.return_to_answer_id will always be populated at this point
+            return_to = ",".join(return_location.return_to.split(",")[1:]) or None  # type: ignore
+            anchor, *return_to_answer_ids = return_location.return_to_answer_id.split(",")  # type: ignore
+            return_to_answer_id = (
+                ",".join(return_to_answer_ids)  # type: ignore
```
katie-gardner, line 376: Do you definitely need this type ignore? I thought mypy might pick it up but I suppose if its coming from a line with a type ignore maybe thats why it doesn't

app/views/contexts/summary/answer.py
```diff
@@ -86,16 +88,24 @@ def _return_to_answer_id(
         return_to: str | None,
         list_item_id: str | None,
         is_in_repeating_section: bool,
+        return_to_answer_id: str | None,
     ) -> str | None:
         """
         If the summary page using this answer has repeating answers, but it is not in a repeating section,
         then the answer ids will be suffixed with list item id, so the return to answer id link also needs this to work correctly
         """
+        answer_id = None
```
katie-gardner, line 97: to fix the below this will need to be `: str | None`

app/views/contexts/summary/answer.py
```diff
@@ -86,16 +88,24 @@ def _return_to_answer_id(
...
                 and not self._original_answer_id  # original answer would mean id has already been suffixed
             ):
-                return f"{self.id}-{list_item_id}"
-            return self.id
+                answer_id = f"{self.id}-{list_item_id}"
+            else:
+                answer_id = self.id
+
+        if return_to_answer_id:
+            # Type ignore: return_to_answer_id will always be populated at this point
```
katie-gardner, line 109: The type ignore is for the wrong reasons, this will be complaining because `answer_id` could be None
You can fix this if you nest this block inside the `if return_to` 
it will only be populated if `return_to` is also populated so it makes sense to nest it and fixes the type hinting

katie-gardner (commented):
Functionally all working nicely now 👍 Just a couple minor comments
And I think there should be a test in `test_router` similar to the existing ones around multiple block ids for the multiple answer ids (might be able to just edit an existing one)

app/questionnaire/router.py
petechd: Turns out it wasn't needed,  thanks.

tests/app/views/contexts/summary/test_answer.py
petechd: Done, thanks.

app/views/contexts/summary/answer.py
petechd: I fixed the below by nesting the logic, so this one is not needed?

katie-gardner: I thought mypy might infer type as None and then not like you adding a string to it later, but if its not complaining then all good 👍

petechd:
> Functionally all working nicely now 👍 Just a couple minor comments And I think there should be a test in `test_router` similar to the existing ones around multiple block ids for the multiple answer ids (might be able to just edit an existing one)

Added two tests in `test_router` that hopefully covers it all now.

tests/app/questionnaire/test_router.py
```diff
@@ -798,6 +798,131 @@ def test_return_to_grand_calculated_summary_from_answer(
 
         assert expected_previous_url == next_location_url
```
katie-gardner, line 799: New tests look good, though the existing one above this is inaccurate now we have multiple return to answer ids, I think. it ought to have two answer ids in the return location like the return to and return to block, the test does work but it doesn't reflect what happens anymore

katie-gardner (commented):
Changes and coverage all good 👍 
The test_router tests `test_return_to_calculated_summary_from_incomplete_section`, `test_return_to_grand_calculated_summary_from_answer` and `test_return_to_grand_calculated_summary_from_answer_incomplete_section` and `test_return_to_grand_calculated_summary_from_repeating_answer` are all using multiple return tos and return to blocks, but not with your new answer change, and I think it would be good to fix so the tests are more similar to actual usage

petechd:
> Changes and coverage all good 👍 The test_router tests `test_return_to_calculated_summary_from_incomplete_section`, `test_return_to_grand_calculated_summary_from_answer` and `test_return_to_grand_calculated_summary_from_answer_incomplete_section` and `test_return_to_grand_calculated_summary_from_repeating_answer` are all using multiple return tos and return to blocks, but not with your new answer change, and I think it would be good to fix so the tests are more similar to actual usage

This one fixed now, thanks.

katie-gardner (approved):

tests/app/views/contexts/summary/test_answer.py
```diff
@@ -37,18 +41,20 @@ def test_create_answer(return_to, return_to_block_id, is_in_repeating_section):
...
-    if not is_in_repeating_section:
-        return_to_answer_id += "-answer-item-id"
-
-    if return_to and return_to_block_id:
-        query_string = f"?return_to={return_to}&return_to_answer_id={return_to_answer_id}&return_to_block_id={return_to_block_id}"
-    elif return_to:
-        query_string = (
-            f"?return_to={return_to}&return_to_answer_id={return_to_answer_id}"
-        )
+    if return_to and return_to_block_id and return_to_answer_id:
```
berroar, line 44: I find this test really hard to follow in it's current state, I know it had if statements previously but this should be avoided in tests. I would be in favour of moving the query string to the params e.g.
`@pytest.mark.parametrize(
    "return_to, return_to_block_id, is_in_repeating_section, return_to_answer_id, query_string",
    [
        (
            "section-summary",
            None,
            False,
            "answer-id",
            "?return_to=section-summary&return_to_answer_id=answer-id-answer-item-id,answer-id",
        ),
        (None, None, False, "answer-id", ""),
        (
            "calculated-summary",
            "total",
            False,
            "answer-id-answer-item-id",
            "?return_to=calculated-summary&return_to_answer_id=answer-id-answer-item-id,answer-id-answer-item-id&return_to_block_id=total",
        ),
        (
            "section-summary",
            None,
            True,
            "answer-id",
            "?return_to=section-summary&return_to_answer_id=answer-id,answer-id",
        ),
        (None, None, True, "answer-id-answer-item-id", ""),
        (
            "calculated-summary",
            "total",
            True,
            "answer-id",
            "?return_to=calculated-summary&return_to_answer_id=answer-id,answer-id&return_to_block_id=total",
        ),
        (
            "calculated-summary",
            "total",
            True,
            "calculated-summary-1",
            "?return_to=calculated-summary&return_to_answer_id=answer-id,calculated-summary-1&return_to_block_id=total",
        ),
    ],
)`

petechd: Done, thanks.

PR #1246, opened 2023-11-03
Add support for CIR

katie-gardner:
This PR adds support for CIR in Runner. The launch metadata now allows `cir_instrument_id` as a 3rd option (alongside `schema_name` and `schema_url` for launching a survey). Metadata validation has been updated to support it, and any schema fetched is cached just like those loaded by schema url.

Tests mocking the endpoint have been added

### How to review
Check that the changes to runner on launch are sufficient to support CIR, that this matches how runner receives the instrument id in `ons-schema-definitions` and that the tests add sufficient coverage.

### How to test functionally
The easiest way to test functionally, is run the [CIR Mock](https://github.com/ONSdigital/eq-runner-mock-cir/pull/1) locally, and in runner, go to `session.py` and manually change the token:
```python
# put this right after runner_claims = get_runner_claims(decrypted_token)
from hashlib import sha256
from uuid import UUID

language = runner_claims["language_code"]
schema_name = runner_claims.pop("schema_name")
schema_hash = sha256(f"{language}_{schema_name}".encode("utf-8")).hexdigest()
runner_claims["cir_instrument_id"] = str(UUID(schema_hash[:32]))
```
This will replace `schema_name` in the token with `cir_instrument_id` equal to the UUID that the CIR mock would have generated based off language and name for the same schema and so call the CIR mock to load the equivalent survey. You can monitor the locally running CIR mock to confirm this is the case.

### Checklist


app/utilities/metadata_parser_v2.py
```diff
@@ -102,9 +103,13 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
...
-        if data and not (data.get("schema_name") or data.get("schema_url")):
+        if data and not (
+            data.get("schema_name")
+            or data.get("schema_url")
+            or data.get("cir_instrument_id")
+        ):
             raise ValidationError(
-                "Neither schema_name or schema_url has been set in metadata"
+                "Neither schema_name, schema_url or cir_instrument_id has been set in metadata"
             )
```
MebinAbraham, line 113: We should update this to the check is that one and only one of them is provided.

katie-gardner: Good shout, done 👍

VirajP1002 (commented):
Functionally tested and expected behaviour achieved

app/utilities/schema.py
```diff
@@ -206,8 +235,12 @@ def load_schema_from_url(
         language_code=language_code,
         pid=pid,
     )
+    if cir_instrument_id:
```
berroar, line 238: Only minor, but we could keep this method generic if we set the parameters in the calling methods?

berroar (approved):
Tested functionally - able to launch with the mock CIR, all looks good 🎉

app/utilities/schema.py
katie-gardner: Yeah, I would prefer generic, my only holding back was trying to keep `test_load_schema_from_url` as is and if parameters as a dictionary is passed in its not hashable for the lru cache, and if I encode the params, I might as well join them up in the caller too - works with an Immutable dictionary though, is what I just pushed up what you had in mind? I had to add some param args in the tests, but outside tests I think this works

Another option is to have `**kwargs` and use this to form the parameters, but I don't like that because it means the existing implementation is then called like this which looks really odd
`load_schema_from_url(schema_url=url, language_code="en", language="en")` 

Or I could go ahead and pass in the constructed url to the function. Gone with the dict but on the fence with which is better, so all for opinions

app/utilities/metadata_parser_v2.py
```diff
@@ -100,12 +101,28 @@ class RunnerMetadataSchema(Schema, StripWhitespaceMixin):
...
+                ]
             )
+            if option_count == 0:
+                raise ValidationError(
+                    "Neither schema_name, schema_url or cir_instrument_id has been set in metadata"
+                )
+            if option_count > 1:
+                raise ValidationError(
+                    "Only one of schema_name, schema_url or cir_instrument_id should be specified in metadata"
+                )
```
MebinAbraham, line 125: It would be useful to know which ones were given. Currently, launching system would have to work out which is extra that us being able to tell them. i.e. Only one can be set but got x and y.

app/utilities/schema.py
```diff
@@ -114,11 +117,15 @@ def load_schema_from_metadata(
...
+            parameters=ImmutableDict({"language": language_code}),
+        )
         duration_in_milliseconds = (time.time() - start) * 1_000
 
         cache_info = (
-            load_schema_from_url.cache_info()  # pylint: disable=no-value-for-parameter
+            load_schema_from_url.cache_info()  # pylint: disable=missing-kwoa,too-many-function-args
         )
         logger.info(
             f"load_schema_from_url took {duration_in_milliseconds:.6f} milliseconds",
```
MebinAbraham, line 131: For the initial go live, it would be worth doing this for instrument id one also. Specifically the time, not cache info.

app/utilities/schema.py
```diff
@@ -196,18 +221,22 @@ def _load_schema_file(schema_name: str, language_code: str) -> Any:
 
 @lru_cache(maxsize=None)
 def load_schema_from_url(
-    schema_url: str, language_code: str | None
+    *, schema_url: str, language_code: str | None, parameters: ImmutableDict[str, str]
```
MebinAbraham, line 224: Instead of passing a dict, pass kwargs?.

app/utilities/schema.py
berroar: Yeah I see what you mean 🤔. What you've pushed is what I had in mind and it looks OK to me now, so happy with it as is 👍

app/utilities/schema.py
```diff
@@ -114,11 +117,15 @@ def load_schema_from_metadata(
         # This should not be used in production.
 
         start = time.time()
-        schema = load_schema_from_url(schema_url, language_code)
+        schema = load_schema_from_url(
+            schema_url=schema_url,
+            language_code=language_code,
+            parameters=ImmutableDict({"language": language_code}),
```
MebinAbraham, line 123: This is odd as we are passing the same thing in two ways.

Given CIR doesn't support languages and schema url is not used for prod and likely to be removed, I suggest remove the language parameter from the url and completely from the `load_schema_from_url` until we need it, for CIR it won't be as simple as a query param anyway.

You can this pass guid as a `**kwargs`, maybe name it `**query_params`

MebinAbraham (commented):
There are some log line where we bind schema name/url, those need update to now also bind CIR instrument ID. Unfortunately, the instrument id won't have enough context for us, we will need to think about binding something else in future

app/utilities/metadata_parser_v2.py
katie-gardner: Good idea, done

app/utilities/schema.py
katie-gardner: Added same thing minus the cache info 👍

app/utilities/schema.py
katie-gardner: I agree its odd, good to remove it but just to confirm, you're suggesting I change the existing
`constructed_schema_url = f"{schema_url}?language={language}"` to `constructed_schema_url = f"{schema_url}` in the case of load schema from url - even if not used in production, won't that break it? or does that not matter. What was the intended use case for it?

app/utilities/schema.py
katie-gardner: The reason I didn't was https://github.com/ONSdigital/eq-questionnaire-runner/pull/1246#discussion_r1393979158 but if we get rid of the language param for schema url then agree with this 👍

katie-gardner: Dropped this now since only the cir variant has params

app/utilities/schema.py
```diff
@@ -196,8 +223,12 @@ def _load_schema_file(schema_name: str, language_code: str) -> Any:
 
 @lru_cache(maxsize=None)
 def load_schema_from_url(
-    schema_url: str, language_code: str | None
+    *, schema_url: str, language_code: str | None
```
MebinAbraham, line 226: Similar to the previous PRs comment, happy for this to be as is, but could be enforced for language code only given function name is clear.

MebinAbraham (approved):

MebinAbraham:
> There are some log lines where we bind schema name/url, those need update to now also bind CIR instrument ID.

This is outstanding.

app/utilities/schema.py
katie-gardner: changed 👍 will leave existing calls as is though

app/utilities/schema.py
katie-gardner: Whole url now goes in instead so still generic but this issue avoided

katie-gardner:
> > There are some log lines where we bind schema name/url, those need update to now also bind CIR instrument ID.
> 
> This is outstanding.

Added these now - followed the pattern of adding a test for the flush endpoint but leaving individual response

app/routes/individual_response.py
```diff
@@ -65,6 +65,11 @@ def before_individual_response_request() -> Response | None:
     if schema_url := metadata.schema_url:
         contextvars.bind_contextvars(schema_url=schema_url)  # pragma: no cover
 
+    if cir_instrument_id := metadata.cir_instrument_id:
+        contextvars.bind_contextvars(
+            cir_instrument_id=cir_instrument_id
+        )  # pragma: no cover
```
MebinAbraham, line 71: Make the helper public and re-use `_bind_contextvars_schema_from_metadata`?

app/routes/flush.py
```diff
@@ -63,6 +63,9 @@ def flush_data() -> Response:
                 contextvars.bind_contextvars(schema_name=schema_name)
             if schema_url := metadata.schema_url:
                 contextvars.bind_contextvars(schema_url=schema_url)
+            if cir_instrument_id := metadata.cir_instrument_id:
+                contextvars.bind_contextvars(cir_instrument_id=cir_instrument_id)
```
MebinAbraham, line 67: Make the helper public and re-use `_bind_contextvars_schema_from_metadata`?. Worth moving to different helper file.

app/settings.py
```diff
@@ -146,6 +146,8 @@ def utcoffset_or_fail(date_value, key):
 
 SDS_API_BASE_URL = os.getenv("SDS_API_BASE_URL")
 
+CIR_API_BASE_URL = os.getenv("CIR_API_BASE_URL")
```
MebinAbraham, line 149: This needs adding to the README https://github.com/ONSdigital/eq-questionnaire-runner?tab=readme-ov-file#environment-variables
Looks like `SDS_API_BASE_URL` was previously missed, worth adding that also.

katie-gardner: Good spot, added both

app/routes/flush.py
katie-gardner: Put in a shared utilities method 👍

README.md
```diff
@@ -390,45 +390,47 @@ Once we have the translated .po files they can be added to the source code and u
...
+| WEB_SERVER_UWSGI_ASYNC_CORES              |                              | The number of cores to initialise when using "uwsgi-async" web server worker type                              |
+| DATASTORE_USE_GRPC                        | False                        | Determines whether to use gRPC for Datastore. gRPC is currently only supported for threaded web servers        |
+| ACCOUNT_SERVICE_BASE_URL                  | `https://surveys.ons.gov.uk` | The base URL of the account service used to launch the survey                                                  |
+| ONS_URL                                   | `https://www.ons.gov.uk`     | The URL of the ONS website where static content is sourced, e.g. accessibility info                            |
+| SDS_API_BASE_URL                          | `http://localhost:5003`      | The base URL of the SDS API used for fetching supplementary data                                               |
+| CIR_API_BASE_URL                          | `http://localhost:5004`      | The base URL of the CIR API used for fetching collection instruments                                           |
```
MebinAbraham, line 433: Doc defaults are prod first. Since these aren't defaulted, these dev defaults can be removed.

MebinAbraham (approved):

PR #1251, opened 2023-11-08
Remove code added to fix functional tests for older chromedriver version

liamtoozer:
We [recently moved](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1180/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L55-R54) to use the baked-in chromedriver in WebdriverIO for our functional tests. Since we're now using a `stable` release channel for chromedriver as a result of this, it's worth investigating to see if some of the workarounds can be removed. These were previously failing intermittently in the GitHub Actions builds.

### How to review
Check that the tests pass locally, but this is mostly to do with getting the GHA builds passing. 

I've tested the same suites run by the GHA locally several times, and all passed okay. I've also re-run the GHA a few times but can't get any of the suites to fail after removing the `waitForExist` calls. This PR might be a case of merging it in and reverting if we spot any flakiness in the future

### Checklist

[Trello](https://trello.com/c/sE17W41y/5751-investigate-removing-code-added-to-fix-functional-tests-for-chromedriver-v113-1-sp)


liamtoozer:
[Investigate removing code added to fix functional tests for Chromedriver v113 (1 SP)](https://trello.com/c/sE17W41y/5751-investigate-removing-code-added-to-fix-functional-tests-for-chromedriver-v113-1-sp)

berroar (approved):

VirajP1002 (approved):
Ran all the specified tests locally and ran GHA a few times which passed successfully!

PR #1253, opened 2023-11-09
Add confirmation email support to non census thank-you page

petechd:
This change needs to be implemented before we start removing census survey config class and census theme. At the moment, email confirmation is only available on census themed thank-you pages and only tested there, if we remove support for census theme, as a consequence of removing config, our coverage of confirmation email will drop to zero.
That's why we need to add email confirmation to different themed schemas and remove the logic around it in thank-you page py module. 

### How to review
Some tests had to be removed as a consequence of removing census theme email confirmation logic (census themed schemas use normal thank-you page template now).

### Checklist


petechd:
[Remove the Census survey configs (1 SP)](https://trello.com/c/x2HAtl6u/5903-remove-the-census-survey-configs-1-sp)

tests/integration/routes/test_confirmation_email.py
```diff
@@ -255,10 +255,9 @@ def test_thank_you_missing_email(self):
 
         # Then I get an error message on the thank you page
         self.assertInUrl("/submitted/thank-you/")
-        self.assertInBody("There is a problem with this page")
         self.assertInBody("Enter an email address")
         self.assertEqualPageTitle(
-            "Error: Thank you for completing the census - Confirmation email test schema"
+            "We’ve received your answers - Confirmation email test schema"
```
berroar, line 260: This test is kind of confusing now, nothing in it to suggest that having a missing email generates an error?

berroar (commented):
I think the schema `test_feedback_email_confirmation` also needs to be updated, not sure how the test are passing for that at the moment as it breaks when I run it locally 🤔

tests/integration/routes/test_confirmation_email.py
petechd: I've changed all error message tests that I've touched so they use more specific CSS selector of error panel, I hope you like it.

petechd:
> I think the schema `test_feedback_email_confirmation` also needs to be updated, not sure how the test are passing for that at the moment as it breaks when I run it locally 🤔

Fixed the schema, it uses "default" theme now, well spotted. I wouldn't worry about it too much as all census mentions are removed removed in the next PR anyway.

templates/thank-you.html
```diff
@@ -80,6 +80,15 @@ <h2 data-qa="view-submitted-response-title">{{ _("Get a copy of your answers") }
     {% endif %}
   {% endif %}
 
+  {% set form = content.form %}
```
MebinAbraham, line 83: Minor call this `confirmation_email_form` as we will have other form on this page soon (feedback).

petechd: I'm not able to make this change unless we change `error-panel.html` template (add some logic like `if confirmation_email_form`). `error-panel.html`  is universally used by all `forms` so I'm not sure if we want to litter it with extra logic. Workaround would be to create another `error-panel.html` just for emails, but I am not sure if we want that. For now I am changing `partials/email-form.html` to `partials/confirmation-email-form.html` as it has some `Confirmation` specific content and it's the least invasive.

katie-gardner (approved):
All works as I'd expect and can't find any regression elsewhere, nice 👍

berroar (approved):

templates/thank-you.html
MebinAbraham: I'm not seeing the issue. Why does `error-panel.html` need changing?.
The error template should be given `form` regardless of which parent template calls it.

Given we will have more than one form, I would call this `confirmation_email_form` in the context before it is even passed to the view. 

It is not an issue in `confirm-email.html` because that will only every have one form, so the line can stay unchanged.

MebinAbraham (commented):
The file `confirmation-email.html` seems to be unused. If so, it should be removed.

templates/thank-you.html
petechd: `confirmation-email.html` -> `error-panel.html. Check `test_confirmation_email.json` and get to send another email panel and send black email.

petechd:
> The file `confirmation-email.html` seems to be unused. If so, it should be removed.

Sort of answered here: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1253#discussion_r1400772939

templates/thank-you.html
petechd: `confirmation-email.html` will need to change for a scenario when you try to send another email from `/submitted/confirmation-email/send`, then you will get the error:
Screenshot 2023-11-22 at 07 56 09

petechd: Just some more context, if we don't modify error message (if we change the form name) then we will end up with an empty error panel. This is on second confirmation email.

petechd: Screenshot 2023-11-22 at 09 02 56

petechd:
[Add confirmation email support to non census thank-you page (2 SP)](https://trello.com/c/riWUSzK9/5963-add-confirmation-email-support-to-non-census-thank-you-page-2-sp)

templates/thank-you.html
MebinAbraham: Here are the changes needed to support multiple forms on the thank you page. No changes needed to any other html files. It was easier for me to push up the changes to a copy branch to demo the change. Only 12 lines of change.
https://github.com/ONSdigital/eq-questionnaire-runner/compare/add-confirmation-email-support-non-census-thank-you-page...add-confirmation-email-support-non-census-thank-you-page-multiple-forms?expand=1

Any questions, let me know.

petechd: That solves the whole problem, pushed up now, thanks!

katie-gardner (approved):
Change since last review looks good and all still working functionally 👍

MebinAbraham (approved):

PR #1258, opened 2023-11-14
Fix test number schema

berroar:
Fixes a bug in the test_numbers schema where the `detail-answer` question relied on an answer that was not mandatory.

### How to review
Check that you can only access the `detail-answer` question when the `test-range` question has been answered.

### Checklist


VirajP1002 (approved):

katie-gardner (approved):
Works as expected and no other cases in the schema of the same issue 👍

PR #1261, opened 2023-11-21
Fix misused test summary helper functions

katie-gardner:
During the Supplementary data progress bug fix, it was noticed that the helper functions `assertSummaryValues`, `assertSummaryItems` and `assertSummaryTitles` are being misused in numerous places, by being called but not awaited, so the assertion does not actually happen. This hid the inconsistent use of the selectors. Some places expected all summary values/items/titles, others a subset. For consistency these have been changed to just use the class name, which covers totals as well as items.

This PR has fixed the selectors and all misused occurrences of the functions *except for the supplementary data spec which will be fixed separately in the bug fix*

A follow up card has been raised to address this more widely in runner, but since I had already fixed these, raising it in this separate PR to avoid bloating my other one.

### How to review
Make sure you understand the issue, agree with the fix, and that the changes to the calculated summary spec assertions are correct. Other than that it was just a case of adding await calls, or adding the total into the summary values.  

### Checklist


liamtoozer (approved):
Tested these changes locally and all pass for me 👍

PR #1265, opened 2023-11-22
Supplementary data dependencies fix

katie-gardner:
This PR replaces [an old PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1224) that implemented a fix to progress not updating on supplementary data change using the new split dependencies. The other branch will be deleted upon the merge of this one.

It fixes the following issues:

1) If a question depends on the length of a supplementary list, and a new supplementary dataset is used with a list item count that opens up the question, previously the section containing the question would remain completed. It now reverts to in progress. This is true when the supplementary dataset has added or removed items, both will work.

2) If a section contains a list collector with repeating blocks depending on supplementary data. If the supplementary data changes and a list item is removed, the section will revert to in progress. This includes for repeating blocks which will require revisiting.

### How to review
The questionnaire store updater has been split into a base updater which does not have location and question, and the main questionnaire store updater now inherits from this. 

**To Make review easier, for now, the two questionnaire store updaters reside in the same file, this way the reviewer can see which methods have been removed from the base updater, and added to the new class**

After merge, a follow up will move it to its own file

### How to test functionally
Run the mock SDS locally on [the branch with an additional dataset](https://github.com/ONSdigital/eq-runner-mock-sds/pull/3)
Use the `test_supplementary_data` schema, and try swapping between the three datasets while clicking the use previous response ID button.

- If using the first supplementary dataset with 2 items, if you complete the production targets section and relaunch the survey with the second supplementary dataset with 3 items, the section should become incomplete.

- If using the second supplementary dataset with 3 items, if you complete the sales targets section, then relaunch the survey using the first supplementary dataset which has 2 items, you should see the section revert to incomplete

- If the product details section is completed whilst using the second supplementary dataset, when moving to the first supplementary dataset which contains fewer items, the section should revert to in progress and force you to reconfirm the calculated summaries within and fill incomplete repeating blocks.

- If you have completed the repeating section for the employees on dataset 1 or dataset 2 and you then use the 3rd dataset, you should see every previously completed repeat revert to in progress, because each repeat now has a new question which was unlocked by having 3 supplementary employees. (Any employees that exist only in the 3rd dataset will appear 'not started' since that section is new)

### Checklist


katie-gardner:
[Bug: Section progress not reset when new supplementary data version has extra dynamic answers](https://trello.com/c/qdiKw74Y/5831-bug-section-progress-not-reset-when-new-supplementary-data-version-has-extra-dynamic-answers)

berroar:
I guess for this scenario: `If you have completed the repeating section for the employees on dataset 1 or dataset 2 and you then use the 3rd dataset, you should see every previously completed repeat revert to in progress, because each repeat now has a new question which was unlocked by having 3 supplementary employees`

Only really minor, but maybe just clarify this is only the employees that exist in both datasets rather than every previously complete as the other new ones will be `Not started`. 

All scenarios behaving correctly for me 🎉

katie-gardner:
> I guess for this scenario: `If you have completed the repeating section for the employees on dataset 1 or dataset 2 and you then use the 3rd dataset, you should see every previously completed repeat revert to in progress, because each repeat now has a new question which was unlocked by having 3 supplementary employees`
> 
> Only really minor, but maybe just clarify this is only the employees that exist in both datasets rather than every previously complete as the other new ones will be `Not started`.
> 
> All scenarios behaving correctly for me 🎉

Cheers Rhys, updated PR comment to mention that

app/data_models/questionnaire_store.py
```diff
@@ -60,30 +63,54 @@ def set_metadata(self, to_set: MutableMapping) -> QuestionnaireStore:
 
         return self
 
-    def set_supplementary_data(self, to_set: MutableMapping) -> None:
+    def set_supplementary_data(
+        self,
+        *,
+        to_set: MutableMapping,
+        base_questionnaire_store_updater: BaseQuestionnaireStoreUpdater,
+    ) -> None:
```
MebinAbraham, line 71: We should avoid this circular dependency, this will complicate this further in future. It's odd to pass the QSU which in charge of updating the QS to QS itself.

Worth having a discussion as it might be complex to explain here.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -15,21 +15,18 @@
 from app.utilities.types import DependentSection, LocationType
 
 
-class QuestionnaireStoreUpdater:
-    """Component responsible for any actions that need to happen as a result of updating the questionnaire_store"""
+class BaseQuestionnaireStoreUpdater:
```
MebinAbraham, line 18: Minor, but would be clearer as `QuestionnaireStoreUpdateBase` to match our other base classes.

app/data_models/questionnaire_store.py
```diff
@@ -60,30 +63,54 @@ def set_metadata(self, to_set: MutableMapping) -> QuestionnaireStore:
...
         identifier -> list_item_id mappings in the supplementary data store to use in the payload at the end
         """
+        modified_lists: set[str] = set()
+
         if self._stores.supplementary_data_store.list_mappings:
-            self._remove_old_supplementary_lists_and_answers(new_data=to_set)
+            self._remove_old_supplementary_lists_and_answers(
+                new_data=to_set,
+                base_questionnaire_store_updater=base_questionnaire_store_updater,
+                modified_lists=modified_lists,
```
MebinAbraham, line 85: This is usually bad practice. Avoid mutation of arguments passed in. It would make be more accepted if it was needed across multiple methods, given modified lists are not needed external to this method, I suggest the methods returning the modified lists.

app/data_models/questionnaire_store.py
katie-gardner: Great point, this is way better, done 👍 
Moved the tests to the test file for the QSUB as well

app/data_models/questionnaire_store.py
katie-gardner: Makes sense, switched to returning them, thanks

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -160,6 +149,13 @@ def remove_list_item_data(self, list_name: str, list_item_id: str) -> None:
...
         self._progress_store.remove_progress_for_list_item_id(list_item_id=list_item_id)
 
+    def remove_list_data(self, list_name: str) -> None:
+        """Delete entire list and remove any associated answers"""
+        self._answer_store.remove_all_answers_for_list_item_ids(
+            *self._list_store[list_name].items
+        )
+        self._list_store.delete_list(list_name)
+
     def capture_dependencies_for_list_change(self, list_name: str) -> None:
```
MebinAbraham, line 159: In this method, there is a scenario in which a KeyError is raised.

```
section_ids.update(
            self._schema.list_collector_section_ids_by_list_name[list_name]
        )
```

This needs to be updated as this would key error if the schema had no list collectors.

Can you ensure all methods called from `set_supplementary_data` are graceful.

MebinAbraham: We can sort this in a new PR given it is existing, as this is need before Roofing Tiles schema can be launched.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -505,6 +439,11 @@ def _evaluate_dependent_of_dependents(
         )
         evaluated_dependents.append((dependent_section_id, list_item_id))
 
+    def is_current_section(self, section_id: str, list_item_id: str | None) -> bool:
```
petechd, line 442: Do we need to pass in arguments if this method doesn't do anything? Shouldn't we create another definition of `remove_dependent_blocks_and_capture_dependent_sections` in the child class that actually makes correct use of `is_current_section`?

petechd: This class' `remove_dependent_blocks_and_capture_dependent_sections` could be?:
```python
def remove_dependent_blocks_and_capture_dependent_sections(self) -> None:
        """Removes dependent blocks from the progress store."""
                    list_item_id=list_item_id,
                    block_id=block_id,
                )
                blocks_removed |= self._progress_store.remove_completed_location(
                    location
                )

            if blocks_removed:
 ```

File: app/questionnaire/questionnaire_store_updater.py [hunk 6]
```diff
@@ -74,7 +80,9 @@ def remove_completed_relationship_locations_for_list_name(
             for target in target_relationship_collectors:
                 block_id = target["id"]
                 section_id = self._schema.get_section_for_block_id(block_id)["id"]  # type: ignore
-                self.remove_completed_location(Location(section_id, block_id))
+                self._progress_store.remove_completed_location(
+                    Location(section_id, block_id)
+                )
```
MebinAbraham, line 85: We should either use the `self.remove_completed_location` wrapper, or remove it. I don't mind as long as it is consistent.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -104,7 +112,7 @@ def _update_relationship_question_completeness(self, list_name: str) -> None:
                         "id"
                     ]  # type: ignore
                     location = Location(section_id, collector["id"])
-                    self.add_completed_location(location)
+                    self._progress_store.add_completed_location(location)
```
MebinAbraham, line 115: Same as above.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -170,7 +166,7 @@ def capture_dependencies_for_list_change(self, list_name: str) -> None:
             list_name
         )
         section_ids.update(
-            self._schema.list_collector_section_ids_by_list_name[list_name]
+            self._schema.list_collector_section_ids_by_list_name.get(list_name, ())
```
MebinAbraham, line 169: I suggest using a `set` default than a `tuple` for clarity, even if it functionally works.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -550,3 +488,209 @@ def _get_chronological_section_dependents(self) -> list:
...
+        Used to set or update the supplementary data whenever the sds endpoint is called
+        (Which should be once per session, but only if the sds_dataset_id has changed)
+
+        this updates ListStore to add/update any lists for supplementary data and stores the
+        identifier -> list_item_id mappings in the supplementary data store to use in the payload at the end
+        """
+        list_mappings: dict[str, list[SupplementaryDataListMapping]] = {}
+        modified_lists: set[str] = set()
+
+        if self._questionnaire_store.data_stores.supplementary_data_store.list_mappings:
```
MebinAbraham, line 503: Given the other stores all have an attribute assigned, we should do the same for sds for consistency.

```py
self._answer_store = self._questionnaire_store.data_stores.answer_store
self._list_store = self._questionnaire_store.data_stores.list_store
self._progress_store = self._questionnaire_store.data_stores.progress_store
```

But given you are updating the literal instance, you will need to use `@property` to ensure there are no stale references. It can also be private like the others.

```py
@property
def _supplementary_data_store(self) -> SupplementaryDataStore:
    return self._questionnaire_store.data_stores.supplementary_data_store
```

MebinAbraham (commented):
Other than some minor comments, LGTM.

app/questionnaire/questionnaire_store_updater.py
katie-gardner: I'd removed the wrapper from the base updater, its only in the subclass, as theres multiple places still using the wrapper for the normal QSU, and I can't keep it in the base one due to lack of location - could look at updating callers but it is consistent within each class if thats enough? otherwise will look at passing in location to any callers

katie-gardner: OK cool, if its just 3 places, I'll pass it in and keep & use the wrapper without the default 👍

app/questionnaire/questionnaire_store_updater.py
katie-gardner: uses helper now 👍

app/questionnaire/questionnaire_store_updater.py
petechd: Yeah, I prefer your latest change, that method before, always returning `False` and all these ignores, was sort of like a `code smell`, hard to explain.

katie-gardner: Yeah I felt similarly about it and couldn't think of a better way at the time, but I agree so good to get rid of it, thanks for flagging 👍

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -550,3 +501,197 @@ def _get_chronological_section_dependents(self) -> list:
...
+                list_item_id = self.add_list_item(list_name)
+                modified_lists.add(list_name)
+            list_mappings.append(
+                SupplementaryDataListMapping(
+                    identifier=identifier, list_item_id=list_item_id
+                )
+            )
+        return list_mappings, modified_lists
+
+    def _remove_old_supplementary_lists_and_answers(
```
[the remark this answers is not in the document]
katie-gardner: I like your suggestion of outdated, to me that already implies existing, and that its no longer relevant, switched to `_remove_outdated_supplementary_lists_and_answers` if that sounds ok?

app/questionnaire/questionnaire_schema.py
```diff
@@ -454,6 +454,14 @@ def _populate_answer_and_list_dependencies(self) -> None:
                 self._update_dependencies_for_summary(block)
                 continue
 
+            if block["type"] == "ListCollectorContent" and block.get(
+                "repeating_blocks"
+            ):
+                # normal list collectors don't need this because the add/remove handlers manage revisiting repeating blocks
```
berroar, line 460: Really minor, but would say something like `ListCollector block types`, rather than `normal list collectors` to make this clearer and more explicit

tests/functional/spec/features/supplementary_data/supplementary_data.spec.js
```diff
@@ -320,6 +323,14 @@ describe("Using supplementary data", () => {
     ]);
     await assertSummaryValues(["100 kg", "200 kg", "50 kg", "300 kg", "£110.00", "£220.00", "£330.00"]);
     await click(Section6Page.submit());
+    await expect(await $(HubPage.summaryRowState("section-6")).getText()).toBe("Completed");
+  });
+
+  it("Given I am using a supplementary dataset with 2 items in the product list, When I enter the production targets section, Then I only see an interstitial block as the question is skipped", async () => {
```
berroar, line 329: Minor, and may be just me, but I think this (and maybe some of the other added tests in other files) should be more generic, so that the behaviour being tested is clear, rather than explicitly stating 2 items and the name of the list etc.

berroar (approved):
All LGTM, just some minor comments 🎉

app/questionnaire/questionnaire_schema.py
katie-gardner: Changed to "Editable" list collectors as I wonder if list collector block type might imply that its any of the list collector blocks and confuse people? but I think editable works and is more explicit than this. Seem ok?

tests/functional/spec/features/supplementary_data/supplementary_data.spec.js
katie-gardner: Yeah I see what you mean, I've gone through and updated some of the descriptions to try and make them a bit more generic/specifc to whats actually being tested rather than the schema setup. If any others jump out let me know. I didn't take out some mentions of product list and section names because I thought if I went much further it might then be confusing when different lists are testing different things, hopefully this is slightly better

app/questionnaire/questionnaire_store_updater.py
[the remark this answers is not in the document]
petechd: Yeah, it will just make it clearer for whoever looks at this code later, thanks.

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -526,19 +473,23 @@ def remove_dependent_blocks_and_capture_dependent_sections(self) -> None:
...
-                # can be removed as we do not need to re-evaluate progress as we already know the section would be incomplete.
-                dependent = DependentSection(section_id, list_item_id)
-                if dependent in self.dependent_sections:
-                    self.dependent_sections.remove(dependent)
-
-                self.dependent_sections.add(
-                    DependentSection(section_id, list_item_id, is_complete=False)
-                )
+            if blocks_removed:
+                self._capture_dependent_section(section_id, list_item_id)
```
MebinAbraham, line 477: Minor, and optional, could use `section_key`. Reduces the number of args, prevent passing incorrect order and section key only have attributes that are needed and no bloat.

MebinAbraham (approved):
👍

app/questionnaire/questionnaire_store_updater.py
katie-gardner: done 👍

app/questionnaire/questionnaire_store_updater.py
```diff
@@ -550,3 +499,193 @@ def _get_chronological_section_dependents(self) -> list:
...
+        )
+        # Type ignore: block id will exist when at any time this is called
+        self._capture_section_dependencies_progress_value_source_for_block(
+            section_id=self._current_location.section_id,
+            block_id=self._current_location.block_id,  # type: ignore
+        )
+
+    def _capture_dependent_section(self, section_key: SectionKey) -> None:
+        """Only capture the dependent section if it is not the current one"""
+        if not section_key == self._current_location.section_key:
```
MebinAbraham, line 690: We should use `!=` instead of `not ==`

katie-gardner: oops 👀

PR #1266, opened 2023-11-24
Add no return_to_answer_id handling when generating return_to for calculated summary

petechd:
Fixes the error we saw in our formal environment. It adds extra check for `return_to_answer_id` in a relevant place so we don't attempt to `.split()` it like before.

### How to review
Steps to described on the card. New test added and simple schema where it is easier to recreate.

### Checklist


petechd:
[Investigate 500 error in router.py in PROD](https://trello.com/c/gEUXCPbj/5964-ops-investigation)

katie-gardner (changes requested):
Can reproduce the error with the new schema following steps on trello card and the changes fix the issue 👍 

I think the schema is unclear without the card though, if adding it I think it would be helpful to simplify the titles in the schema and add guidance panels explaining how to test with it. That said, happy to defer in the interests of getting the fix merged and avoiding more prod errors

schemas/test/en/test_validation_sum_against_total_calculated_summary.json
```diff
@@ -0,0 +1,160 @@
+{
+    "mime_type": "application/json/ons/eq",
+    "language": "en",
```
katie-gardner, line 3: its also possible to replicate the error by adding a single Calculated Summary of breakdown 1-4 to the end of `test_validation_sum_against_value_source` which has a similar setup, which might be preferable to the extra schema, but could also detract from what it is actually there to test, not too fussed either way, just an idea if you did want to avoid the extra one, otherwise feel free to resolve

app/questionnaire/router.py
```diff
@@ -350,7 +350,15 @@ def _get_return_to_for_calculated_summary(
             # remove first item and return the remaining ones
             # Type ignore: return_location.return_to and return_location.return_to_answer_id will always be populated at this point
```
berroar, line 351: This type ignore string is no longer correct?

app/questionnaire/router.py
```diff
@@ -350,7 +350,15 @@ def _get_return_to_for_calculated_summary(
...
             return_to = ",".join(return_location.return_to.split(",")[1:]) or None  # type: ignore
```
MebinAbraham, line 352: Is this type ignore valid and guarded? We need to be sure otherwise this can also error.

app/questionnaire/router.py
MebinAbraham: Also, a note to self on being careful with type ignores. It was an issue highlighted by mypy that we chose to ignore and got wrong.

petechd: Yeah, it is guaranteed, this code only executes when higher up this `if` statement is valid:
```python
if return_location.return_to.startswith("calculated-summary")..
```

app/questionnaire/router.py
petechd: Yeah, it is guaranteed, this code only executes when higher up this `if` statement is valid:
```python
if return_location.return_to.startswith("calculated-summary")..
```

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -149,4 +156,37 @@ describe("Feature: Sum of grouped answers equal to validation against value sour
       await expect(await $(SecondBreakdownAnswerPage.errorNumber(1)).getText()).toBe("Enter answers that add up to 3");
     });
   });
+  describe("Given I complete a grouped answer validation survey and get to the final summary page", () => {
+    it("When I journey back and I end up with no answer_id anchor for the url, Then I should not see any errors", async () => {
```
berroar, line 160: I would update this to make the journey clearer and not reference the backend stuff i.e. answer id or anchor? Tricky, but something like `Given I edit a question from a Calculated Summary page, When I change the answer and there is a question that needs to be revisited before I can return to the Calculated Summary Page, Then I revisit the relevant page before I route back to the Calculated Summary page.`

berroar (approved):

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: Ok, always tricky, especially with such a convoluted journey.

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -51,7 +52,7 @@ describe("Feature: Sum of grouped answers equal to validation against value sour
 
       await answerBothBreakdownQuestions(["3", "3", "3", "3"], ["2", "2", "1", "1"]);
 
-      await expect(browser).toHaveUrlContaining(SubmitPage.pageName);
+      await expect(browser).toHaveUrlContaining(AnotherTotalPlaybackPage.pageName);
```
katie-gardner, line 55: The test is a `Then I should get to the summary page` which doesn't align, probably needs to be "then i get to the next calculated summary" or continue once more to the submit page

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -74,7 +77,7 @@ describe("Feature: Sum of grouped answers equal to validation against value sour
 
       await answerBothBreakdownQuestions(["6", "3", "3", "3"], ["3", "3", "2", "1"]);
 
-      await expect(browser).toHaveUrlContaining(SubmitPage.pageName);
+      await expect(browser).toHaveUrlContaining(AnotherTotalPlaybackPage.pageName);
```
katie-gardner, line 80: same comment as above

tests/functional/spec/features/validation/sum/value_source.spec.js
```diff
@@ -149,4 +156,37 @@ describe("Feature: Sum of grouped answers equal to validation against value sour
...
+      await $(TotalAnswerPage.total()).setValue("5");
+      await click(TotalAnswerPage.submit());
+
+      await answerBothBreakdownQuestions(["2", "1", "1", "1"], ["1", "2", "0", "0"]);
+
+      await click(AnotherTotalPlaybackPage.submit());
+
+      await expect(browser).toHaveUrlContaining(SubmitPage.pageName);
+
+      await $(SubmitPage.previous()).click();
```
katie-gardner, line 170: ah, I realise I did this in the screen recording, but these two steps aren't required, no need to go to the submit page and then press previous, can just go straight into editing breakdown 1

katie-gardner (approved):
Superficial comments only, schema change looks good 👍

tests/functional/spec/features/validation/sum/value_source.spec.js
petechd: Well spotted, thanks!

PR #1267, opened 2023-11-27
Add require-await rule for functional test linting

liamtoozer:
This PR introduces the `require-await` eslint rule for Runner's functional tests. 

Some of the functional tests in runner use asynchronous helper functions to assert behaviour, but do not await these functions, resulting in the check not actually happening. This is hiding potential bugs in functionality. 

* Added the eslint rule `require-await` and updated tests caught by this rule. Interestingly, this rule only picks up arrow functions missing an `await` in our functional tests.
* Updated other tests that had missing `await`s. These missing `await`s were reported by the JetBrains IDE code analysis tool.

### How to review
Try running the linter with `make lint-js` and the `require-await` rule should pass.

### Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -68,6 +68,7 @@
       {
         "blocks": "never"
       }
-    ]
+    ],
+      "require-await": "error"
   }
 }
--- a/tests/functional/spec/answer_action_redirect_to_list_add_block_checkbox.spec.js
+++ b/tests/functional/spec/answer_action_redirect_to_list_add_block_checkbox.spec.js
@@ -37,7 +37,7 @@ describe("Answer Action: Redirect To List Add Question (Checkbox)", () => {
       await expect(browser).toHaveUrlContaining(AnyoneLiveAtListCollector.pageName);
 
       const peopleExpected = ["Marcus Twin"];
-      checkItemsInList(peopleExpected, AnyoneLiveAtListCollector.listLabel);
+      await checkItemsInList(peopleExpected, AnyoneLiveAtListCollector.listLabel);
     });
 
     it('When the user click the "Previous" link from the list collector, Then, they are taken to the last complete block', async () => {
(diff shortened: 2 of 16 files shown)


katie-gardner (approved):
Looking good - command working locally and shows errors if I alter code incorrectly. 
Interesting find with the extra errors from pycharm, looks like require-await catches missing awaits but not redundant ones?
Potentially something like https://typescript-eslint.io/rules/await-thenable/ could work but since its a redundancy rather than bug/error I think this change is good as is 👍

katie-gardner (commented):
Just for interest
behaviour seems to be:
- Error if async function with no await in the body
- Error if await in the body of non async function

but doesn't appear to actually check if async functions are being used properly 😞 limitation of the rule I guess

liamtoozer:
>Looking good - command working locally and shows errors if I alter code incorrectly.
Interesting find with the extra errors from pycharm, still some there which seem to just be redundant awaits
Potentially something like https://typescript-eslint.io/rules/await-thenable/ could work but since its a redundancy rather than bug/error I think this change is good as is 👍

That typescript lint rule looks good! Definitely worth investigating that, along with any further tests that could be amended from pycharm's linting outputs as well 👍 

> Just for interest behaviour seems to be:
> 
> * Error if async function with no await in the body
> * Error if await in the body of non async function
> 
> but doesn't appear to actually check if async functions are being used properly 😞 limitation of the rule I guess

Ah, that's a shame but good to know, thanks. Hopefully, this new rule will at least catch some of the edge cases

MebinAbraham:
> Updated other tests that had missing awaits. These missing awaits were reported by the PyCharm IDE code analysis tool for JS.

Got some examples? And do we know why?.

liamtoozer:
> > Updated other tests that had missing awaits. These missing awaits were reported by the PyCharm IDE code analysis tool for JS.
> 
> Got some examples? And do we know why?

The errors that have been fixed in this PR that were raised by PyCharm have been added to PR description now 👍

Not sure why these calls have been missed, possibly a result of a larger refactor? I looked through some of the previous commits around the files but nothing jumped out.

MebinAbraham (approved):

liamtoozer:
This PR included changes from this PR - https://github.com/ONSdigital/eq-questionnaire-runner/pull/1261

PR #1268, opened 2023-11-28
Add support in repeating GCS for static CS with repeating answers

katie-gardner:
This PR adds support for a repeating Grand Calculated Summary to include a static Calculated Summary that has repeating answers. These can be repeating block answers or dynamic answers.

This is paired with the [corresponding validator PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/227) which disallows the use of a calculated summary which has repeating answers for the same lists. So if for example, a calculated summary included a sum of dynamic answers for vehicles. In a repeating section for vehicles that calculated summary could not then be included in a repeating grand calculated summary, as it would be unclear whether to resolve the repeating answer id to the specific vehicle instance, or the collection of answers, given it was from another section. 

I had to update `test_repeating_sections_with_hub_and_spoke.json` to stop validator from breaking. It has the same answer id in the add block for both list collectors but this breaks under the new validation (and is a bug anyway) raised another card for a fix to that properly in validator

### How to review
Look over the card to check all parts have been covered. Functionally the changes are very small, its just a change to value source resolver to support returning all the items unless in a repeat for the same list. Primarily the changes are test coverage.

### How to test functionally
Use the schema `test_grand_calculated_summary_inside_repeating_section`
- Ensure that the change links in the grand calculated summaries work both for repeating answers and dynamic answers
- Make sure that changing an answer which affects the other repeating section correctly updates progress
- Note the guidance about the validation block, there is a separate card for implementing grand calculated summary dependencies. But make sure the calculated summary dependencies are working as expected.

### Checklist


katie-gardner:
[Grand Calculated Summary in repeat: Support for non-repeating Calculated Summaries with repeating answers (5 SP)](https://trello.com/c/s7JzRKrJ/5892-grand-calculated-summary-in-repeat-support-for-non-repeating-calculated-summaries-with-repeating-answers-5-sp)

VirajP1002 (commented):
Functionally tested and it works as intended 👍

app/questionnaire/value_source_resolver.py
```diff
@@ -75,6 +75,23 @@ def _get_answer_value(
...
+            answer_list := self.schema.get_list_name_for_answer_id(answer_id)
+        ):
+            # if there is a current list, and it differs to the repeating answer one, return None
+            if (
+                self.location
+                and self.location.list_name
+                and self.location.list_name != answer_list
+            ):
+                return None
+            return self.list_item_id
```
petechd, line 93: Just a detail. It, sort of, suggests that the var is a list of answers, when it's a list name, so maybe `list_name`, `list_name_for_answer` or `list_name_anwer_id` or something similar?
```suggestion
        if self.list_item_id and (
            list_name_for_answer := self.schema.get_list_name_for_answer_id(answer_id)
        ):
            # if there is a current list, and it differs to the repeating answer one, return None
            if (
                self.location
                and self.location.list_name
                and self.location.list_name != list_name_for_answer
            ):
                return None
            return self.list_item_id
```

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=gcs-in-repeat-with-static-cs-repeating-answers
```
petechd, line 2: Reminder to revert

schemas/test/en/test_grand_calculated_summary_inside_repeating_section.json
```diff
@@ -216,6 +213,49 @@
...
+                                                    "value": {
+                                                        "source": "answers",
+                                                        "identifier": "cost-name"
+                                                    }
+                                                }
+                                            ]
+                                        },
+                                        "answers": [
+                                            {
+                                                "id": "cost-base",
```
petechd, line 237: Should we keep the same convention here and call it `cost-repeating-block-1-answer` or `cost-repeating-block-1-answer-base`. Only a detail.

berroar (commented):
Not sure if this is one of the known issues for submitting incomplete? But when you fill in the GCS inside the repeating question, and the the dependent breakdown question with the validation based on the GCS total. If you go from the Summary and change the `What is your monthly expenditure on fuel for your Car?` question total, you can revisit the summary without the breakdown question being revalidated (`gcs-breakdown-block`)? 
So then you get back to the section summary with something like: 
image

katie-gardner:
> Not sure if this is one of the known issues for submitting incomplete? But when you fill in the GCS inside the repeating question, and the the dependent breakdown question with the validation based on the GCS total. If you go from the Summary and change the `What is your monthly expenditure on fuel for your Car?` question total, you can revisit the summary without the breakdown question being revalidated (`gcs-breakdown-block`)? So then you get back to the section summary with something like: image

Yeah this will unfortunately be the case until my GCS dependencies card merges, anything that the GCS *depends on* will work as expected, but anything that *depends on the GCS* doesn't yet get revisited when the GCS changes, so the functional test here will need minor modification when the other merges, or whichever goes first

berroar (approved):
LGTM!

schemas/test/en/test_grand_calculated_summary_inside_repeating_section.json
katie-gardner: Sure, not against changing these to be explicit, done same for the dynamic one

app/questionnaire/value_source_resolver.py
```diff
@@ -75,6 +75,23 @@ def _get_answer_value(
         ):
             return answer.value
 
+    def _resolve_list_item_id_for_answer(self, answer_id: str) -> str | None:
```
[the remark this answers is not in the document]
katie-gardner: Fixed the function name 👍 I was thinking the answer is what I'd consider repeating rather than the id, and thought the docstring might be too verbose if I added "and the answer id is for a repeating answer" but up for tweaking it or adding if that makes it easier to follow

petechd (approved):

liamtoozer (commented):
Functionally tested and works great for me 👍

PR #1269, opened 2023-11-28
Update mypy and gevent to latest versions

petechd:
This unpins mypy and gevent libraries so they are updated to the latest and all consequent issues are resolved.

### How to review
For testing gevent run the benchmark, mypy will highlight new issues in build and lint.

### Checklist


app/questionnaire/questionnaire_schema.py
```diff
@@ -701,7 +701,7 @@ def get_title_for_section(self, section_id: str) -> str | None:
 
     def get_show_on_hub_for_section(self, section_id: str) -> bool | None:
         if section := self.get_section(section_id):
-            return section.get("show_on_hub", True)
+            return bool(section.get("show_on_hub", True))
```
katie-gardner, line 704: This type cast is unnecessary, type ignore instead for this one?

app/questionnaire/questionnaire_schema.py
```diff
@@ -713,7 +713,7 @@ def get_summary_title_for_section(self, section_id: str) -> str | None:
 
     def show_summary_on_completion_for_section(self, section_id: str) -> bool | None:
         if summary := self.get_summary_for_section(section_id):
-            return summary.get("show_on_completion", False)
+            return bool(summary.get("show_on_completion"))
```
katie-gardner, line 716: I think based off type hint guidance this ought to be the original code with a type ignore

app/questionnaire/questionnaire_schema.py
```diff
@@ -871,7 +871,7 @@ def get_repeating_blocks_for_list_collector(
         self, list_collector_id: str
     ) -> list[ImmutableDict] | None:
         if list_collector := self.get_block(list_collector_id):
-            return list_collector.get("repeating_blocks", [])
+            return list(list_collector.get("repeating_blocks", []))
```
katie-gardner, line 874: same again, I think the type casting is not needed, and it can be type ignored

app/submitter/converter.py
```diff
@@ -128,7 +128,7 @@ def build_metadata(metadata: MetadataProxy) -> MetadataType:
 
 def get_optional_payload_properties(
     metadata: MetadataProxy, response_metadata: MutableMapping
-) -> MetadataType:
+) -> dict:
```
katie-gardner, line 131: If this is a `MetadataType` can this instead by fixed by type hinting `payload: MetadataType = {}`? or is it genuinely a different type

app/submitter/converter_v2.py
```diff
@@ -98,7 +98,7 @@ def convert_answers_v2(
 
 def get_optional_payload_properties(
     metadata: MetadataProxy, response_metadata: MutableMapping
-) -> MetadataType:
+) -> dict:
```
katie-gardner, line 101: same as above

app/views/handlers/confirmation_email.py
```diff
@@ -75,5 +75,5 @@ def is_limit_reached(session_data: SessionData) -> bool:
     @staticmethod
     def is_enabled(schema: QuestionnaireSchema) -> bool:
         if submission_schema := schema.get_post_submission():
-            return submission_schema.get("confirmation_email", False)
+            return bool(submission_schema.get("confirmation_email"))
```
katie-gardner, line 78: Same as other type cast comment

app/views/handlers/feedback.py
```diff
@@ -195,7 +195,7 @@ def is_limit_reached(session_data: SessionData) -> bool:
     @staticmethod
     def is_enabled(schema: QuestionnaireSchema) -> bool:
         if submission_schema := schema.get_post_submission():
-            return submission_schema.get("feedback", False)
+            return bool(submission_schema.get("feedback"))
```
katie-gardner, line 198: Type cast

app/views/handlers/list_collector_content.py
```diff
@@ -4,7 +4,7 @@
 
 class ListCollectorContent(ListCollector):
     def _get_additional_view_context(self) -> dict:
-        return self.rendered_block.get("content", {})
+        return dict(self.rendered_block.get("content", {}))
```
katie-gardner, line 7: Same again

katie-gardner (changes requested):
All works but most the mypy fixes are type casting where I think we can get away without it, suggest replacing with ignores instead

app/submitter/converter.py
petechd: Cause otherwise the `|` operator will be complained about in mypy as we try to merge two different types..

petechd:
[Unpin mypy (1 SP)](https://trello.com/c/XVA0R6dt/5957-unpin-mypy-1-sp)

petechd:
[Unpin gevent (1 SP)](https://trello.com/c/XyEgGN1a/5958-unpin-gevent-1-sp)

app/submitter/converter.py
katie-gardner: Yeah you're totally right, now I've updated mypy, the only way this could be fixed is doing `payload.update(optional_properties)` instead, and since return types are better as concrete, I think what you've done is better anyway, the `MetadataType` doesn't seem correct, so I'm all for getting rid of it and using dict as you've done 👍 my bad

petechd: Still, good to hear your opinion and discuss it.

rmccar (approved):

katie-gardner:
@petechd Benchmark looks fine too
image

petechd:
> @petechd Benchmark looks fine too !

Thanks.

PR #1270, opened 2023-12-05
Grand calculated summary dependencies

katie-gardner:

### How to review

### Checklist

### DIFF
--- a/app/questionnaire/dependencies.py
+++ b/app/questionnaire/dependencies.py
@@ -7,7 +7,7 @@ from werkzeug.datastructures import MultiDict
 
 from app.data_models import ProgressStore
 from app.questionnaire import QuestionnaireSchema
-from app.questionnaire.questionnaire_schema import get_sources_for_type_from_data
+from app.questionnaire.questionnaire_schema import get_sources_for_types_from_data
 from app.utilities.mappings import get_flattened_mapping_values
 from app.utilities.types import LocationType, SectionKey
 
@@ -20,7 +20,7 @@ def get_routing_path_block_ids_by_section_for_dependent_sections(
     location: LocationType,
     progress_store: ProgressStore,
     path_finder: PathFinder,
-    source_type: str,
+    source_types: set[str],
     data: MultiDict | Mapping | Sequence,
     sections_to_ignore: list | None = None,
     ignore_keys: list[str] | None = None,
@@ -35,8 +35,8 @@ def get_routing_path_block_ids_by_section_for_dependent_sections(
         else get_flattened_mapping_values(dependent_sections)
     )
 
-    if dependents and not get_sources_for_type_from_data(
-        source_type=source_type, data=data, ignore_keys=ignore_keys
+    if dependents and not get_sources_for_types_from_data(
+        source_types=source_types, data=data, ignore_keys=ignore_keys
     ):
         return block_ids_by_section
 
@@ -55,7 +55,7 @@ def get_routing_path_block_ids_by_section_for_dependent_sections(
     return block_ids_by_section
 
 
-def get_routing_path_block_ids_by_section_for_calculated_summary_dependencies(
+def get_routing_path_block_ids_by_section_for_calculation_summary_dependencies(
     *,
     location: LocationType,
     progress_store: ProgressStore,
@@ -65,7 +65,12 @@ def get_routing_path_block_ids_by_section_for_calculated_summary_dependencies(
     ignore_keys: list[str] | None = None,
     schema: QuestionnaireSchema,
 ) -> dict[SectionKey, tuple[str, ...]]:
-    dependent_sections = schema.calculated_summary_section_dependencies_by_block[
+    """
+    If the current location depends on any calculated or grand calculated summaries,
+    for all the sections that those CS or GCS values depend on, get the blocks on the path for that section.
+    These routing path block ids are then used to ensure the CS/GCS only includes answers on the path
+    """
+    dependent_sections = schema.calculation_summary_section_dependencies_by_block[
         location.section_id
     ]
     return get_routing_path_block_ids_by_section_for_dependent_sections(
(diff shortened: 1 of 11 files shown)


katie-gardner:
[Grand Calculated Summary Value Source dependencies (5 SP)](https://trello.com/c/kzl2Hwfl/5911-grand-calculated-summary-value-source-dependencies-5-sp)

app/questionnaire/dependencies.py
```diff
@@ -20,7 +20,7 @@ def get_routing_path_block_ids_by_section_for_dependent_sections(
     location: LocationType,
     progress_store: ProgressStore,
     path_finder: PathFinder,
-    source_type: str,
+    source_types: set[str],
```
petechd, line 23: Why did you decide to use set here? There are arguments, like `sections_to_ignore` that are passed as a list and not modified in the method. maybe we should keep it consistent and use sets or lists for all arguments of this method that it applies to? I mean, there is no difference in how `sections_to_ignore` and `source_types` are used later?

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=grand-calculated-summary-value-sources-fix
```
petechd, line 2: Reminder👍

app/questionnaire/questionnaire_schema.py
```diff
@@ -562,6 +562,27 @@ def _update_dependencies_for_dynamic_options(
...
+        )
+        dependent = self._get_dependent_for_block_id(
+            block_id=block_id, answer_id=answer_id
+        )
+        for answer_id_for_block in answer_ids_for_block:
+            self._answer_dependencies_map[answer_id_for_block].add(dependent)
+            # if the answer is repeating, then the calculated summary also depends on the list it loops over
+            if list_name := self.get_list_name_for_answer_id(answer_id_for_block):
+                self._list_dependencies_map[list_name].add(dependent)
+
```
petechd, line 585: Maybe not necessary as already explained what kind of answer ids they are:
```suggestion
        for answer_id in answer_ids_for_block:
            self._answer_dependencies_map[answer_id].add(dependent)
            # if the answer is repeating, then the calculated summary also depends on the list it loops over
            if list_name := self.get_list_name_for_answer_id(answer_id):
                self._list_dependencies_map[list_name].add(dependent)

```

app/questionnaire/questionnaire_schema.py
```diff
@@ -1241,6 +1265,14 @@ def _get_section_and_block_ids_dependencies_for_progress_source_and_answer_ids_f
                 calculated_summary_block  # type: ignore
             )
             dependent_answer_ids.update(calculated_summary_answer_ids)
+        elif source == "grand_calculated_summary" and identifier:
```
petechd, line 1268: This begs for `identifier` being checked beforehand. Since you check for it in every condition?

app/questionnaire/questionnaire_schema.py
```diff
@@ -577,22 +598,24 @@ def _update_dependencies_for_value_source(
...
-                dependent = self._get_dependent_for_block_id(
-                    block_id=block_id, answer_id=answer_id
+            self._update_dependencies_for_calculated_summary_value_source(
+                calculated_summary_id=identifier, block_id=block_id, answer_id=answer_id
+            )
+
+        elif value_source["source"] == "grand_calculated_summary":
+            identifier = value_source["identifier"]
+            # Type ignore: validator will ensure identifier is valid
+            gcs_block: ImmutableDict = self.get_block(identifier)  # type: ignore
```
berroar, line 610: As per the other PR, there is mixture here of shortened and the full version of `gcs` in vars which could probably do with being made consistent throughout.

app/questionnaire/questionnaire_schema.py
katie-gardner: Yeah, I would have done this, but since there is an answer id being passed into the function I'm hesitant to overwrite it in the for loop as it could look like a mistake to someone revisiting this

app/questionnaire/questionnaire_schema.py
katie-gardner: Good idea - I'm adding this to the other PR (Simplify when rules) because in that one the rules that get checked all have identifier, so the check was never necessary at all, and if I do a check here it will just get deleted again the moment the other card merges

app/questionnaire/questionnaire_schema.py
```diff
@@ -1352,26 +1380,34 @@ def _get_rules_section_dependencies(
...
         and make sure all involved sections are added as dependencies for that block and its section.
 
         Since calculated summaries can only contain answers from the section they are in,
-        it is sufficient to check the section for the calculated summary only.
+        it is sufficient to check the section for the calculated summary only. For grand calculated summaries
+        Need to include the section of the gcs, and each cs it references.
         """
         for section in self.get_sections():
             for block in self.get_blocks_for_section(section):
                 sources = get_mappings_with_key(
```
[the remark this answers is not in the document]
katie-gardner: I can't immediately see a nice way of making that work, it would need to check the mappings in the block to find out if it contains value sources, so I think it sort of has to check them all, but did you have a particular idea? If so maybe a follow up as its existing?

app/questionnaire/questionnaire_schema.py
katie-gardner: Updated them all to full names

app/questionnaire/dependencies.py
katie-gardner: Yeah set is not needed, it can be more generic so I've changed it to Iterable, unlike sections to ignore theres no case of `source_types or []` or similar, so I think a totally generic type is probably what this should be

berroar (approved):
Tested functionally, changes LGTM 🎉!

liamtoozer (approved):
I tested this functionally with each of the schemas in the PR and works great for me 👍

PR #1271, opened 2023-12-05
Simplify When rule dependencies

katie-gardner:
Adding [GCS Dependencies](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1270) increased the complexity of the already very complicated tracking of when rule dependencies. The existing approach to doing this had 3 functions to manage when rules

#### _get_section_and_block_ids_dependencies_for_progress_source_and_answer_ids_from_rule
- Given a rule with a source and identifier, this would find the dependent answers, lists, and blocks/sections relating to progress. (ie. Make sure the section with the rule depends on the right blocks, sections, answers)
- This function passed those dependencies to its caller
&darr;
#### _get_rules_section_dependencies
- This one is iterating over a set of when rules, for each rule, if it is one with a 'source' and 'identifier' type call the above to fill up local dependency maps
- It had a nested approach, update those local dependency maps to include the results of calling this function again for each item within the rule
- At the end it returns those same dependency maps to its caller:
&darr;
#### _populate_when_rules_section_dependencies
- This is the parent, iterating over each section, and for each when rule in the section, get the dependency maps with the above function. It then uses those maps to populate each of the schema `_when_rule_x_dependencies_by_y` maps.

### Refactor 
If instead, the rules with sources & identifiers are gathered up front out of the when rules, then no nesting is required, neither is checking the rule, and if a function processing the rule updates the `_when_rule_x_dependencies_by_y` maps itself, then there is no need to pass multiple maps between functions. So it becomes easier to read

This was done as part of the GCS Dependencies ticket, but I didn't want to make it really difficult to review, so this refactor is in a separate PR based off the GCS branch

### How to review
Functionally this should not have any effect at all, try completing some unhappy paths with complex schemas based around when rules for regression testing. Check the changes make sense and if they make this part of questionnaire schema easier to understand and follow.

### Checklist


app/questionnaire/questionnaire_schema.py
```diff
@@ -1177,85 +1179,42 @@ def _populate_when_rules_section_dependencies(self) -> None:
...
+                    rule, current_section_id=section["id"]
+                )
 
-    def _populate_block_dependencies_for_progress_value_source(
-        self,
-        rule_block_dependencies_for_progress_value_source: dict[
-            str, DependencyDictType
-        ],
+    def _populate_dependencies_for_rule(
+        self, rule: Mapping, *, current_section_id: str
```
petechd, line 1198: Why keyword-only argument after the `rule`, not before?

katie-gardner: It was a preference Mebin put forward, where since it ends with `for_rule` it is intuitive that the rule argument would be first, then for any other args where the order is not clear enforce kwargs

petechd (approved):

berroar (approved):

app/questionnaire/questionnaire_schema.py
```diff
@@ -1190,127 +1192,83 @@ def _populate_when_rules_section_dependencies(self) -> None:
...
-
-            self._populate_block_dependencies_for_progress_value_source(
-                rule_block_dependencies_for_progress_value_source
-            )
+            for rule in rules:
+                self._populate_dependencies_for_rule(
+                    rule, current_section_id=section["id"]
+                )
 
-    def _populate_block_dependencies_for_progress_value_source(
```
liamtoozer, line 1226: V minor - the doc string in `_populate_when_rules_section_dependencies()` references this function and might need to be tweaked now?

liamtoozer (commented):
Functionally tested and works great for me 👍

app/questionnaire/questionnaire_schema.py
katie-gardner: ah great shout - have updated the docstring

PR #1277, opened 2023-12-08
Update a schema to demonstrate minimum/maximum supports floats

VirajP1002:
Update the JSON schema validator to accurately support decimal minimum and maximum constraints while disallowing string types. To demonstrate this, I changed a schema in runner to show that floats are allowed.

`test_metadata_routing.json` schema doesn't work and a card in the backlog has been created to address this. The change was to ensure schema validation passes.

### How to review
- Ensure the schema is correct & validates
- Ensure the functional tests (numbers.spec.js) pass

### Checklist


VirajP1002:
[Update validator to support float minimum/maximum](https://trello.com/c/bpi6I4wg/5822-update-validator-to-support-float-minimum-maximum)

schemas/test/en/test_dynamic_answers_list_source.json
```diff
@@ -156,7 +156,7 @@
                                             "type": "Number",
                                             "decimal_places": 2,
                                             "minimum": {
-                                                "value": 1001
+                                                "value": 1000.99
```
katie-gardner, line 159: I think it would be good to have a functional test for this to make sure it's working properly, with correct error message, i.e. you can enter 1000.99 but not 1000.98

VirajP1002: I've added a test in 👍

schemas/test/en/test_dynamic_answers_list_source.json
```diff
@@ -150,16 +150,16 @@
...
                                             "id": "set-maximum",
-                                            "description": "Maximum amount of spending at this supermarket, should be between 1001 and 10000",
+                                            "description": "Maximum amount of spending at this supermarket, should be between 1000.99 and 10000.98",
                                             "label": "Maximum Spending",
                                             "mandatory": true,
                                             "type": "Number",
                                             "decimal_places": 2,
                                             "minimum": {
-                                                "value": 1001
+                                                "value": 1000.99
```
MebinAbraham, line 159: We should do this in the `test_numbers.json` file.

tests/functional/spec/features/dynamic_answers_list_value_source.spec.js
```diff
@@ -162,6 +162,24 @@ describe("Dynamic answers list value source", () => {
     await expect(await $$(OnlineShoppingPage.labels())[1].getText()).toBe("Percentage of online shopping at Aldi");
     await expect(await $$(OnlineShoppingPage.labels()).length).toBe(4);
   });
+
+  it("Given the minimum value is set to 1000.99, When the maximum amount of spending entered should be between 1000.99 and 10000 and 1000.98 is entered, Then the value should not be accepted", async () => {
```
MebinAbraham, line 166: To build on the above, we should update `numbers.spec.js`. The test should be relevant to the schema it is in.

app/jinja_filters.py
```diff
@@ -250,18 +250,21 @@ def get_min_max_value_width(
     """
     This function gets the minimum and maximum value accepted for a question.
     Which then allows us to use that value to set the width of the textbox to suit that min and max.
+
+    Check if value is a Mapping if not "answers" source and returns default value because the widths
+    of calculated_summary/grand_calculated_summary and potentially supplementary_data and metadata
+    are calculated incorrectly
```
katie-gardner, line 256: Just re-reading I think the docstring could do with an update, it mentions the other types of sources are calculated incorrectly, which _used_ to be the case, which is why you've changed it - so I'd reword it as something like

"If the min or max for the answer is a value source but not an "answers" source, such as a calculated or grand calculated summary, use the length of the default value for the min and max width, as the actual min and max width cannot currently be determined"

app/questionnaire/questionnaire_schema.py
```diff
@@ -165,8 +166,8 @@ def _create_min_max_map(
         for answer in answers:
             value = answer.get(min_max, {}).get("value")
 
-            if isinstance(value, float | int):
-                value_length = len(str(value))
+            if isinstance(value, float | int | Decimal):
+                value_length = len(str(int(value)))
```
katie-gardner, line 170: Comment here too about decimals would be good

katie-gardner: (i.e. The value is being rounded to an integer as decimals are accounted for already in jinja filters)

app/jinja_filters.py
```diff
@@ -250,18 +250,20 @@ def get_min_max_value_width(
...
 
-    if (
-        answer.get(min_max, {})
-        and isinstance(answer[min_max]["value"], Mapping)
-        and answer[min_max]["value"].get("source") == "answers"
-    ):
-        schema: QuestionnaireSchema = g.get("schema")
-        identifier = answer[min_max]["value"].get("identifier")
-        return schema.min_and_max_map[identifier][min_max]
+    if answer.get(min_max, {}) and isinstance(answer[min_max]["value"], Mapping):
```
berroar, line 258: Maybe worth creating a var for `answer.get(min_max, {})` as it is used twice?

app/jinja_filters.py
```diff
@@ -250,18 +250,20 @@ def get_min_max_value_width(
...
+        if answer[min_max]["value"].get("source") == "answers":
+            schema: QuestionnaireSchema = g.get("schema")
+            identifier = answer[min_max]["value"].get("identifier")
+            return schema.min_and_max_map[identifier][min_max]
+        return len(str(default_value))
 
-    return len(str(answer.get(min_max, {}).get("value", default_value)))
+    # Factor out the decimals as it's accounted for in get_width_for_number
+    return len(str(int(answer.get(min_max, {}).get("value", default_value))))
```
berroar, line 266: I would do the int conversion first with the new var suggested above, as don't find this line very readable at the moment with multiple casts.

app/jinja_filters.py
```diff
@@ -250,18 +250,20 @@ def get_min_max_value_width(
...
-        and isinstance(answer[min_max]["value"], Mapping)
-        and answer[min_max]["value"].get("source") == "answers"
-    ):
-        schema: QuestionnaireSchema = g.get("schema")
-        identifier = answer[min_max]["value"].get("identifier")
-        return schema.min_and_max_map[identifier][min_max]
+    if answer.get(min_max, {}) and isinstance(answer[min_max]["value"], Mapping):
+        if answer[min_max]["value"].get("source") == "answers":
+            schema: QuestionnaireSchema = g.get("schema")
+            identifier = answer[min_max]["value"].get("identifier")
```
berroar, line 261: Should be able to use `answer[min_max]["value"]["identifier"]` as identifier is mandatory for answer value sources

app/jinja_filters.py
VirajP1002: I've added in some changes, let me know if these need to be changed to improve readability 👍

berroar (approved):

app/jinja_filters.py
```diff
@@ -250,18 +250,21 @@ def get_min_max_value_width(
...
+    if min_max_value and isinstance(answer[min_max]["value"], Mapping):
+        if answer[min_max]["value"].get("source") == "answers":
+            schema: QuestionnaireSchema = g.get("schema")
+            identifier = answer[min_max]["value"]["identifier"]
+            return schema.min_and_max_map[identifier][min_max]
+        return len(str(default_value))
+
+    # Factor out the decimals as it's accounted for in get_width_for_number
+    result = int(min_max_value.get("value", default_value))
+    return len(str(result))
```
MebinAbraham, line 267: This is suitable for now but we would have to resolve the length for calc/grand summary but using the max length of any answers that it makes up. Is there a card to do that, or was this solution not chosen as a temp solution?.

scripts/run_validator.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-tag=latest
+tag=support-float-max-or-min
```

tests/functional/spec/numbers.spec.js
```diff
@@ -13,6 +13,20 @@ describe("Number validation", () => {
...
   });
   describe("Given I am completing the test numbers questionnaire,", () => {
+    it("When a float less than the minimum value, Then the correct error message including the minimum value as a float is included", async () => {
+      await $(SetMinMax.setMinimum()).setValue(-1000.99);
+      await $(SetMinMax.setMaximum()).setValue(1000);
+      await click(SetMinMax.submit());
+      await expect(await $(SetMinMax.errorNumber(1)).getText()).toBe("Enter an answer more than or equal to -1,000.98");
+    });
+
+    it("When a float less than the maximum value, Then the correct error message including the maximum value as a float is included", async () => {
```
MebinAbraham, line 23: These don't read well. I would change it to something like `When a maximum value with decimals are used, Then ...`

VirajP1002: Updated 👍

tests/functional/spec/numbers.spec.js
```diff
@@ -13,6 +13,20 @@ describe("Number validation", () => {
     await browser.openQuestionnaire("test_numbers.json");
   });
   describe("Given I am completing the test numbers questionnaire,", () => {
+    it("When a maximum value with decimals is used, Then the correct error message including the minimum value as a float is included", async () => {
```
MebinAbraham, line 16: Are these tests not the wrong way around, the minimum is testing maximum and vice versa?.

I think it would help readability if they read something like the following:
`When a maximum value with decimals is used and I enter a value greater than the maximum, Then the error message display the maximum value I can enter including the decimal`

Would need to change for both.

VirajP1002: They were the wrong way around, thanks and I've updated them 👍

MebinAbraham (approved):

scripts/run_validator.sh
VirajP1002: Reverted

PR #1278, opened 2023-12-11
List collector content progress tests

katie-gardner:
This PR pulls across the functional fix to list collector content progress made in the [Supplementary data bugfix](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1265) and adds a test for it in the list collector content spec.

At the moment, due to the need to capture the dependency on the list collector content block when the list changes, and the need to capture dependencies of a list when an item is removed, there is an edge case where removing a list item requires the list collector content block to be re-visited. Since its an edge case, and there is a high likelihood the section would need revisiting anyway, its a single extra continue click for the user, and deemed expected behaviour. A comment has been added to the test & remove handler to highlight this should it be changed later

the test was being temperamental locally so I switched to use the `click` helper function and I updated the existing test to re-use the new helper function but otherwise its just two extra ones

### How to review
Check the test covers the change as per the card and attempt any unhappy path you can think of to try and submit the survey with incomplete repeating blocks. It should now be covered in all scenarios. 

It may make more sense for the test to use a `before` hook instead of before each and run in sequence, comment if so and I'll update.

### Checklist


katie-gardner:
[Questionnaire should not be marked as complete until repeating blocks are completed](https://trello.com/c/EPVLvS6U/5811-questionnaire-should-not-be-marked-as-complete-until-repeating-blocks-are-completed)

MebinAbraham (approved):

petechd (approved):

PR #1284, opened 2023-12-28
Update runner to use poetry package manager

Yuyuutsu:
Update Runner to use poetry package manager.
Other related changes made: 
- Packages have been updated. 
- Fixed any new type hinting errors found 
- Flake8 Config has been updated to accommodate new black update. [Black documentation here](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#:~:text=Flake8%23,E203%2C%20E704)

`Pylint` has been pinned due to problems with how `Pylint` interacts with `pylint_quotes`in versions`3.*.*`. There has been an issue raised and a PR to resolve the issue with `pylint_quotes`, but owner of the repoistory has not merged yet. See [issue here](https://github.com/edaniszewski/pylint-quotes/issues/29)

We will now use [poetry-plugin-dotenv](https://github.com/volopivoshenko/poetry-plugin-dotenv) plugin to load dotenvs.
Also added [poetry-plugin-up](https://github.com/MousaZeidBaker/poetry-plugin-up) plugin to manage `pyproject.toml ` dependencies The installation guide of the plugins have been added to the `ReadMe` file.

### How to review
- Github actions works as expected with Poetry
- Make Commands work as expected
- Dockerfile Builds
- if Runner application is working as expected
- Tests run as expected
- New Installation and setup works 
### Checklist


Yuyuutsu:
[Update runner to use poetry package manager (2 SP)](https://trello.com/c/9rGxXogW/5930-update-runner-to-use-poetry-package-manager-2-sp)

pyproject.toml
petechd, line 1: Should we unpin all these and keep them wildcarded like before?
Screenshot 2024-01-23 at 11 43 59

Yuyuutsu: Updated to use wildcard. However I have had to pin pylint due to an issue with astroid and pylint_quotes. An Issue has been raised and there is a fix ready. Just waiting on the owner of the repo to merge but it seems like the last time he was active was in 2021 😭. I have added this to the description of the PR

.github/workflows/pull_request.yml
```diff
@@ -22,13 +22,23 @@ jobs:
...
+          version: 1.7.1
+          virtualenvs-create: true
+          virtualenvs-in-project: false
+          virtualenvs-path: ~/.virtualenvs
+      - name: Cache Poetry virtualenv
+        uses: actions/cache@v3
+        id: cache-virtualenv
+        with:
+          path: ~/.virtualenvs
+          key: ${{ runner.os }}-${{ env.PYTHON_VERSION }}-virtualenvs-${{ hashFiles('poetry.lock') }}
```
MebinAbraham, line 37: We should cache using the `cache` key of `actions/setup-python@v4` step.  Poetry is supported. https://github.com/actions/setup-python#caching-packages-dependencies

.github/workflows/pull_request.yml
```diff
@@ -22,13 +22,23 @@ jobs:
...
       - name: Install virtual environment
         run: |
           sudo apt-get install libsnappy-dev
-          pipenv install --dev
+          poetry install --no-interaction
```
MebinAbraham, line 41: Poetry by default installs dev dependencies.  Use `--only main` to install the main deps only. Need to change references elsewhere also.

Dockerfile
```diff
@@ -14,11 +14,13 @@ ENV WEB_SERVER_UWSGI_ASYNC_CORES 10
...
-COPY Pipfile Pipfile
-COPY Pipfile.lock Pipfile.lock
+COPY pyproject.toml pyproject.toml
+COPY poetry.lock poetry.lock 
 
 RUN groupadd -r appuser && useradd -r -g appuser -u 9000 appuser && chown -R appuser:appuser .
-RUN pip install pipenv==2023.8.22 && pipenv install --deploy --system && make build
+RUN pip install "poetry==1.7.1"
+RUN poetry config virtualenvs.create false
+RUN poetry install --no-dev && make build
```
MebinAbraham, line 23: Use `--only main` instead of `--no-dev` as poetry no allows groups. If we have more than a dev group, i.e test group, this would install test group dependencies. Plus `--no-dev` is deprecated.

app/helpers/url_safe_serializer.py
```diff
@@ -6,4 +6,7 @@
 
 
 def url_safe_serializer() -> URLSafeSerializer:
-    return URLSafeSerializer(current_app.secret_key, salt=cookie_session[EQ_SESSION_ID])
+    secret_key = current_app.secret_key
+    if secret_key is None:
+        raise ValueError("Flask application secret key is not set.")
+    return URLSafeSerializer(secret_key, salt=cookie_session[EQ_SESSION_ID])
```
MebinAbraham, line 12: This shouldn't have changed. This would be too late. We should type ignore this but enfore secret key exists when loaded. So in `setup_secure_cookies` method.

pyproject.toml
```diff
@@ -1,2 +1,89 @@
...
\ No newline at end of file
+target_version = ['py38']
+
+[tool.poetry]
+name = "eq-questionnaire-runner"
+version = "0.1.0"
+description = "Python-Flask server side web application"
+authors = ["ONSDigital"]
+
+[tool.poetry.dev-dependencies]
```
MebinAbraham, line 10: Defining dev dependencies like this is deprecated. You need to use groups https://python-poetry.org/docs/master/managing-dependencies/. So here we should use `[tool.poetry.group.dev.dependencies]`

.github/workflows/pull_request.yml
```diff
@@ -93,18 +113,33 @@ jobs:
...
+        run: poetry install --no-interaction
       - name: Load templates
         run: make load-design-system-templates
       - name: Compile translations
         run: make translate
       - name: Link env vars
-        run: ln -sf .development.env .env
+        run: ln -sf .development.env .env 
+      - name: Load .env File
+        uses: xom9ikk/dotenv@v2
```
MebinAbraham, line 139: We should only use well maintained or recognised authors. This doesn't seem to be. It would be good if Actions doesn't differ from local. Can you not use the poetry dot env plugin instead?.

MebinAbraham (commented):
Is the PR description up to date?. Other than Actions, there is no mention of dotenv. Did you not need the poetry dotenv plugin?. If so, that needs documenting, if not, how is the dotenvs being loaded?.

MebinAbraham (commented):
Currently `make run` is erroring for me. There are some breaking changes that need resolving. `Error: Failed to find Flask application or factory in module 'app'. Use 'app:name' to specify one.
`

.github/workflows/pull_request.yml
Yuyuutsu: Poetry dot env plugin has an issue here with CI/CD. I think i miss read your message last week with the plugin name. I think you meant poetry-plugin-dotenv . I will double check with that plugin instead

Yuyuutsu:
> Is the PR description up to date?. Other than Actions, there is no mention of dotenv. Did you not need the poetry dotenv plugin?. If so, that needs documenting, if not, how is the dotenvs being loaded?.

Same as above. i think its because i used poetry dotenv plugin rather then poetry plugin dot env. Will update so it should work for everyone

app/helpers/url_safe_serializer.py
Yuyuutsu: Yeah wasnt sure to type ignore will remove test as well 👍

.github/workflows/pull_request.yml
MebinAbraham: So here, since it was installed the dev dependencies, this reference is correct.

.github/workflows/pull_request.yml
Yuyuutsu: Updated to use the Poetry plugin dotenv. Had to be separately installed since its a plugin and not part of the lock and toml file

.github/workflows/pull_request.yml
Yuyuutsu: Oh I thought it was just good practice because of examples I saw on the docs. Now Removed 👍

pyproject.toml
Yuyuutsu: Changed to `[tool.poetry.group.dev.dependencies]`

petechd:
The issue in your broken test could be `test_links_not_in_header_when_valid_session_theme_social` method definition duplicated in that class. You have two instances of it.

Yuyuutsu:
> test_links_not_in_header_when_valid_session_theme_social

Thanks, I'll have a look

Dockerfile
```diff
@@ -14,11 +14,13 @@ ENV WEB_SERVER_UWSGI_ASYNC_CORES 10
...
 ENV GUNICORN_CMD_ARGS -c gunicorn_config.py
 
-COPY Pipfile Pipfile
-COPY Pipfile.lock Pipfile.lock
+COPY pyproject.toml pyproject.toml
+COPY poetry.lock poetry.lock 
 
 RUN groupadd -r appuser && useradd -r -g appuser -u 9000 appuser && chown -R appuser:appuser .
-RUN pip install pipenv==2023.8.22 && pipenv install --deploy --system && make build
+RUN pip install "poetry==1.7.1"
```
liamtoozer, line 21: To prevent introducing additional image layers with the separate `RUN` statements here, could these commands be grouped sequentially like in a similar way that the pipenv commands were? So something like:
```
RUN pip install "poetry==1.7.1" && \
    poetry config virtualenvs.create false && \
    poetry install --only main && \
    make build
```
This might slightly reduce the image size also? No worries if this has already been tried though!

Yuyuutsu: Installation with a single `Run` statement has been added 👍

liamtoozer (approved):
Works great for me - tested locally and with a container 👍

petechd (approved):

pyproject.toml
MebinAbraham: After having a discussion with Rhys, we've decided that we should use the poetry default versioning scheme. So don't use `*`, instead use the poetry default of using `^` to allow minor version update.  We can also use https://github.com/MousaZeidBaker/poetry-plugin-up to help manage version updates. This should be added to the README.

pyproject.toml
```diff
@@ -1,2 +1,90 @@
 [tool.black]
-target_version = ['py37']
\ No newline at end of file
+target_version = ['py38']
```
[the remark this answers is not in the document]
Yuyuutsu: I dont think that was meant to be there. Removed 👍

.github/workflows/pull_request.yml
Yuyuutsu: Forgot to update. Was added 👍

pyproject.toml
Yuyuutsu: Updated Readme and Pyproject Toml File.

.github/workflows/pull_request.yml
```diff
@@ -93,18 +102,32 @@ jobs:
...
+          virtualenvs-path: ~/.virtualenvs
+      - uses: actions/setup-python@v4
+        id: cache-poetry
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          cache: "poetry"
+      - name: Install dotenv plugin
+        run: poetry self add poetry-plugin-dotenv
+      - name: Install Dependencies 
+        if: steps.cache-poetry.outputs.cache-hit != 'true'
```
MebinAbraham, line 120: This shouldn't be needed. See https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

Check if poetry action is compatible with this, if not we can install via `pipx`

.github/workflows/pull_request.yml
```diff
@@ -93,18 +102,32 @@ jobs:
...
-        run: pip install pipenv==2023.8.22
-      - name: Install virtual environment
-        run: pipenv install --dev
+      - name: Install poetry
+        uses: snok/install-poetry@v1
+        with:
+          version: 1.7.1
+          virtualenvs-create: true
+          virtualenvs-in-project: false
+          virtualenvs-path: ~/.virtualenvs
```
MebinAbraham, line 111: We shouldn't need to set these, these are the defaults. Unless there was a reason for doing so?. The default path uses the cache dir, this one doesn't.

pyproject.toml
```diff
@@ -1,2 +1,87 @@
-[tool.black]
-target_version = ['py37']
\ No newline at end of file
+[tool.poetry]
+name = "eq-questionnaire-runner"
+version = "0.1.0"
+description = "Flask server side application which provide a set of functionalities that go into schemas used to create surveys"
```
MebinAbraham, line 4:
```suggestion
version = "1.0.0"
description = "ONS Digital eQ Questionnaire Runner App"
```

README.md
```diff
@@ -85,18 +85,26 @@ Python versions can be changed with the `pyenv local` or `pyenv global` commands
...
 ```
 
-###### NOTE: You may encounter an issue with the `python-snappy` installation with `pipenv` above on Mac OSX. If so, install with the recommended [workaround command](https://github.com/andrix/python-snappy#frequently-asked-questions) instead:
-``` shell
-CPPFLAGS="-I/usr/local/include -L/usr/local/lib" pipenv install --dev
-```
+Install poetry, poetry dotenv plugin and install dependencies:
+
+''' shell
+curl -sSL https://install.python-poetry.org | python3 -
```
MebinAbraham, line 98: We should pin the poetry version. We will have to update it somewhat frequently. Some lock files generated from specific versions cannot be used by other version. Given we pin the version in actions and Dockerfile, we should pin here to ensure we generate the lock using the same version. 

Use the latest version `1.8.2`

Yuyuutsu: Updated all ares to now use `v1.8.2` and Readme updated to install pinned version of Poetry

.github/workflows/pull_request.yml
Yuyuutsu: I think I have updated it as you suggested. I was a bit confused so could you check if it was the change you wanted 👍

.github/workflows/pull_request.yml
```diff
@@ -19,16 +19,20 @@ jobs:
...
       - run: |
           echo "PYTHON_VERSION=$(cat .python-version)" >> $GITHUB_ENV
-      - uses: actions/setup-python@v4
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+        with:
+          version: 1.8.2
+          virtualenvs-create: true
+      - uses: actions/setup-python@v5
+        id: cache-poetry
```
MebinAbraham, line 28: id is now redundant, hence can be removed.

.github/workflows/pull_request.yml
```diff
@@ -46,24 +50,26 @@ jobs:
...
-        run: pip install pipenv==2023.8.22
-      - name: Install virtual environment
-        run: pipenv install --dev
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+        with:
+          version: 1.8.2
+          virtualenvs-create: true
+      - uses: actions/setup-python@v5
+        id: cache-poetry
```
MebinAbraham, line 65: id is now redundant, hence can be removed.

.github/workflows/pull_request.yml
```diff
@@ -93,18 +95,27 @@ jobs:
...
```
MebinAbraham, line 104: id is now redundant, hence can be removed.

.github/workflows/pull_request.yml
```diff
@@ -131,22 +142,24 @@ jobs:
...
       - run: |
           echo "PYTHON_VERSION=$(cat .python-version)" >> $GITHUB_ENV
-      - uses: actions/setup-python@v4
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+        with:
+          version: 1.8.2
+          virtualenvs-create: true
+      - uses: actions/setup-python@v5
+        id: cache-poetry
```
MebinAbraham, line 151: id is now redundant, hence can be removed.

.github/workflows/pull_request.yml
Yuyuutsu: Removed

.github/workflows/pull_request.yml
```diff
@@ -93,18 +95,27 @@ jobs:
...
         run: make load-design-system-templates
       - name: Compile translations
         run: make translate
       - name: Link env vars
-        run: ln -sf .development.env .env
+        run: ln -sf .development.env .env 
       - name: Running unit tests
-        run: make test-unit
+        run:     
+          make test-unit
```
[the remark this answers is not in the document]
Yuyuutsu: Removed

.github/workflows/pull_request.yml
Yuyuutsu: Removed

.github/workflows/pull_request.yml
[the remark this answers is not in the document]
Yuyuutsu: Hmm must have been accidental. Reverted 👍

MebinAbraham (approved):

pyproject.toml
```diff
@@ -1,2 +1,89 @@
 [tool.black]
-target_version = ['py37']
\ No newline at end of file
+target_version = ['py38']
+
+[tool.poetry]
+name = "eq-questionnaire-runner"
+version = "0.1.0"
+description = "Python-Flask server side web application"
+authors = ["ONSDigital"]
```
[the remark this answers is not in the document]
Yuyuutsu: Resloved in another comment

app/helpers/url_safe_serializer.py
Yuyuutsu: Resolved on another PR

PR #1287, opened 2024-01-03
Remove redundant awaits found in PyCharm and await-thenable linting

liamtoozer:
A spike investigation was carried out to look into additional linting with a TypeScript rule to prevent redundant `await`s in our functional tests. Although it was decided that we wouldn't be using it in the short-term, it did find some issues that could be mopped up. This is also the case for some issues that PyCharm flagged up during the investigation.

This PR therefore is a small removal of redundant `await`s found from the investigation.

### How to review
Ensure that the functional tests still pass correctly.


liamtoozer:
[Investigate redundant awaits in functional tests (1 Day)](https://trello.com/c/dj7gacWx/5970-investigate-redundant-awaits-in-functional-tests-1-day)

petechd (approved):

berroar (approved):

PR #1302, opened 2024-01-22
Runner test script improperly generating survey metadata

VirajP1002:
`create_token.py` had a method that was broken leading to metadata not being handled correctly in particular for any non-v1 `create_token` method. In order to fix this, logic was added to ensure additional metadata was placed in `survey_metadata` which is the correct level. Tests had to be added in `test_login` as the bug was non-existent without those tests in place.
 
A test suite for the `create_token.py` test script was created in order to demonstrate that a token was being generated correctly and that metadata was being placed at the correct level and the contents were correct.

### How to review
- Ensure the changes made are appropriate
- Ensure unit tests pass

### Checklist


VirajP1002:
[Bug: Runner test cases improperly generating survey metadata](https://trello.com/c/LhQAQEmV/5952-bug-runner-test-cases-improperly-generating-survey-metadata)

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,138 @@
...
+            "collection_exercise_sid": "789",
+            "response_id": "1234567890123456",
+            "language_code": "en",
+            "roles": [],
+            "account_service_url": ACCOUNT_SERVICE_URL,
+        },
+    },
+]
+
+"""
```
berroar, line 71: Would move this comment so it sits under the class, also minor I'd prob remove the v1/v2 stuff since there is already a ticket raised to remove it so won't be around for much longer.

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,138 @@
...
+    def test_supplementary_data_included_in_token(self):
+        token = self.token_generator.create_supplementary_data_token(
+            "test_checkbox.json", flag_1=True
+        )
+        with self.test_app.app_context():
+            decrypted_token = decrypt_token(token)
+            self.assertEqual(
+                decrypted_token, PAYLOAD_V2_SUPPLEMENTARY_DATA | decrypted_token
+            )
+            assert ({"flag_1": True}).items() <= decrypted_token.get(
```
berroar, line 124: Why are we checking for flag 1 in this test? Should that be separate to supplementary data?

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,138 @@
...
+                "survey_metadata"
+            ).get("data").items()
+
+    def test_metadata_is_removed_from_token(self):
+        for value in [
+            [self.token_generator.create_token_without_jti, "jti"],
```
berroar, line 130: Could make this a dict to make it more readable in the test below?

tests/integration/test_create_token.py
VirajP1002: Just to test any additional payload is added in at the correct level and I added it as a parameter where I called the `token_generator.create_supplementary_data_token`, happy to change to a different piece of metadata

berroar: I guess my point is that it is just confusing what is being tested here as the naming of the test suggests it is only checking that the supplementary data metadata is set

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,138 @@
...
+                    "case_ref": "1000000000000001",
+                    "qid": PAYLOAD_V2_SOCIAL.get("survey_metadata")
+                    .get("data")
+                    .get("qid"),
+                    "value": "Dummy Text",
+                },
+                "receipting_keys": ["qid"],
+            }
+
+    def test_supplementary_data_included_in_token(self):
```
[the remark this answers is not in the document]
VirajP1002: Updated 👍

tests/integration/test_create_token.py
VirajP1002: Updated

tests/integration/test_create_token.py
VirajP1002: Maybe to avoid a convoluted name, it maybe be easier to split the test into two, one being `test_supplementary_dataset_included_in_token` and the other being `test_structure_of_supplementary_dataset_token`

tests/integration/session/test_login.py
```diff
@@ -287,6 +299,16 @@ def test_login_with_invalid_questionnaire_claims_should_be_forbidden(self):
 
         self.assertStatusForbidden()
 
+    def test_login_with_invalid_questionnaire_claims_should_be_forbidden_v2(self):
```
petechd, line 302: Avoid adding two tests with the same name.

tests/integration/create_token.py
```diff
@@ -125,7 +125,11 @@ def _get_payload_with_params(
         payload_vars["response_expires_at"] = get_response_expires_at()
 
         for key, value in extra_payload.items():
-            payload_vars[key] = value
+            if payload_vars.get("survey_metadata"):
```
petechd, line 128: Possibly move this condtional from the loop so it's only checked once and have two loops on `if` and `else`? As in:
```python
if survey_metadata := payload_vars.get("survey_metadata").get("data"):
    for key, value:
        survey_metada[key] = value
else:
    for key, value:
         payload_vars[key] = value
```
Otherwise you create the survey metadata every single time? Also what happens to "data" object? Is it supposed to stay in payload_vars["survey_metadata"]?

tests/integration/create_token.py
```diff
@@ -125,7 +125,11 @@ def _get_payload_with_params(
...
+                survey_metadata = payload_vars["survey_metadata"].get("data")
+                survey_metadata[key] = value
```
petechd, line 130: Also, shouldn't this be:
```python3
    data = payload_vars["survey_metadata"].get("data")
    data[key] = value
``` 
for clarity? so:
```python3
data = payload_vars["survey_metadata"].get("data")
       for key, value in extra_payload.items():
                data[key] = value
````

File: tests/integration/create_token.py (hunk 6 above)
Comment (petechd) [reply tests/integration/create_token.py:128]: You can simplify it by doing:
```python
if survey_metadata := payload_vars.get("survey_metadata").get("data"):
    populate_with_extra_payload_items(extra_payload, survey_metadata)
else:
    populate_with_extra_payload_items(extra_payload, payload_vars)
```

VirajP1002: Updated 👍

tests/integration/create_token.py
VirajP1002: Makes sense, updated 👍

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,173 @@
...
+        token = self.token_generator.create_supplementary_data_token(
+            "test_checkbox.json"
+        )
+        with self.test_app.app_context():
+            decrypted_token = decrypt_token(token)
+            self.assertEqual(
+                decrypted_token, PAYLOAD_V2_SUPPLEMENTARY_DATA | decrypted_token
+            )
+
+    def test_structure_of_sds_metadata_in_token(self):
```
petechd, line 124: I wouldn't say this is sds specific, just name the test after what you are exactly testing here, so something like `test_extra_payload(or metadata)_items_are(is)_added_to_token`. Similar to what you are doing below, where you call it: `test_metadata_is_removed_from_token`

VirajP1002: Previously, I had one test which was the combination of `test_structure_of_sds_metadata_in_token` and `test_sds_metadata_included_in_token`. I had a previous comment from Rhys where the name of the old test was confusing as it was testing two things so I split these tests up to make the purposes of the tests clear. I understand your point and I'm happy to rename but the intended purpose of this test was to ensure the structure of `survey_metadata` matched with the `survey_metadata` in `PAYLOAD_V2_SUPPLEMENTARY_DATA`.

VirajP1002: If I changed it, a more appropriate name could be `test_additional_payload_nested_in_token`?

petechd: Yeah, this is a good and better name.

VirajP1002: Updated 👍

tests/integration/session/test_login.py
petechd, line 302: I am not sure about tests having same names (even worse, in the same module), but they are part of different classes, but still, should be something like `test_login_with_invalid_questionnaire_claims_should_be_forbidden_v2_post`.
But I know it wasn't your design so not your fault, I don't expect you to change it, so just a note.

tests/integration/session/test_login.py
```diff
@@ -11,6 +11,8 @@
 
 SCHEMA_PATH_MAP = get_schema_path_map(include_test_schemas=True)
 
+# pylint: disable=too-many-public-methods
```
petechd, line 14: Maybe just make some of the ones (that are being called in tests) private?

petechd: Even renaming single method, `cir_url_mock`  as `_cir_url_mock` will cause this pylint issue go away? But you could rename all the static methods.

tests/integration/session/test_login.py
VirajP1002: I'm happy to change this as it may be difficult for someone to navigate to find the correct tests

tests/integration/session/test_login.py
VirajP1002: Not sure this would work as I tried this and it failed both unit tests and linting

tests/integration/session/test_login.py
petechd: Yeah, that's what I was thinking, you will be send to the wrong method when traversing. Maybe lets wait for some other devs opinion on this?

tests/integration/session/test_login.py
petechd: Maybe you refactored it incorrectly? I renamed it to `_cir_url_mock` and both pylint (after removing too many public methods exclusion) and unit tests are happy. I am on your branch, I've made sure I re-installed all dependencies etc. What does it fail on for you?
EDIT: In your failed test a method cannot be found:
`AttributeError: 'TestLoginWithPostRequest' object has no attribute 'schema_url_mock'` so some bad refactoring.

VirajP1002: There was an `_` missing somewhere else too, updated 👍

petechd: Can you rename all remaining static methods (3 left?) so we have some consistency in the file?

VirajP1002: Renamed them 👍

petechd (approved):

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,173 @@
...
+test_parameters = [
+    {
+        "schema": "test_metadata_routing.json",
+        "theme": "default",
+        "additional_payload": {
+            "flag_1": 123,
+            "period_id": "abc",
+            "link": "https://example.com",
+        },
+        "payload": {
```
berroar, line 22: This can just be:
```suggestion
        "payload": PAYLOAD_V2_BUSINESS,
```
Same applies to social below

VirajP1002: Updated this, I also move `test_parameters` into the test it's used by to have consistency in the file as the last test does the same

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,134 @@
...
+                    "flag_1": True,
+                    "user_id": "abc",
+                    "date": "2016-05-12",
+                },
+                "payload": PAYLOAD_V2_SOCIAL,
+            },
+        ]
+        for value in test_parameters:
+            with self.subTest():
+                additional_payload = value.get("additional_payload")
```
berroar, line 42: The .gets in this file can just be bracket access as we know the values exist in the dict

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,134 @@
...
+from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class TestCreateToken(IntegrationTestCase, AppContextTestCase):
+    """
+    The purpose of this test class is to test the creation of a token (from create_token.py) to ensure
+    metadata in a decrypted token is nested/found in the correct level.
+    """
+
+    def test_payload_from_token(self):
```
berroar, line 17: Could be test payload structure from token or something more descriptive?

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,134 @@
...
+            with self.subTest():
+                additional_payload = value.get("additional_payload")
+                token = self.token_generator.create_token_v2(
+                    schema_name=value.get("schema"),
+                    theme=value.get("theme"),
+                    **additional_payload,
+                )
+                with self.test_app.app_context():
+                    decrypted_token = decrypt_token(token)
+                    self.assertEqual(
```
berroar, line 50: I'm not sure about the | here as it makes the test a bit confusing to follow, I'm not sure if maybe there is a nicer way of checking the expected values? I would also avoid multiple assertions where possible - the additional assertion below checks that the keys exist in the decrypted token but should we not check for an exact match with expected keys and values for the whole survey metadata dict, as there could be extra values we do not expect?

tests/integration/test_create_token.py
VirajP1002: Updated the ones for the test values where we know the values exist, didn't do the same for `decrypted_token`, I wasn't sure if those `.get`'s should be changed?

tests/integration/test_create_token.py
VirajP1002: Changed to a more suitable name 👍

tests/integration/test_create_token.py
VirajP1002: I've refactored the test where it is one big test, that checks the content & structure of the token is as expected from `decrypted_token` this will cover both assertions.

berroar: Think this is a lot clearer now 👍

tests/integration/test_create_token.py
berroar: Yeah even those ones can be updated as we expect them to exist as part of the test 👍

tests/integration/test_create_token.py
```diff
@@ -0,0 +1,203 @@
...
+        mock_uuid.return_value = 1001
+        mock_time.return_value = 1709054595.091798
+        mock_response_expiry_time.return_value = "2024-02-28T09:59:43.109276+00:00"
+        test_parameters = [
+            {
+                "schema": "test_metadata_routing.json",
+                "theme": "default",
+                "additional_payload": {
+                    "flag_1": 123,
+                    "period_id": "abc",
```
berroar, line 96: Minor but I would update the period_id to the kind of period_id we would expect e..g YYYYMM format

berroar (approved):
LGTM 👍

tests/integration/test_create_token.py
VirajP1002: Updated 👍

tests/integration/test_create_token.py
VirajP1002: Makes sense, updated 👍

PR #1305, opened 2024-01-23
Update to DS 67.0.5

rmccar:
Updates to DS version 67.0.5

### How to review
Load a few schemas and see if they still look and work as expected

### Checklist


rmccar:
[Update to DS version 67.0.5](https://trello.com/c/fCMalz7K/6017-update-to-ds-version-6705)

Yuyuutsu (approved):
Looks good, only difference between staging is that the header and footer link gap size. But I think that was an issue with how staging is deployed, if I remember right.

rmccar:
> Looks good, only difference between staging is that the header and footer link gap size. But I think that was already a known thing if I remember right.

Oh yeah that was down to the html minification I think we have a card to look into this

berroar (approved):

PR #1306, opened 2024-01-24
Bug fix: Backwards routing does not work when dependencies exist

MebinAbraham:
Fixes an issue with backwards routing due to us overwriting the reference to `routing_path_block_ids` when building the path. The variable `routing_path_block_ids` is passed to `_evaluate_routing_rules` by reference and mutated when routing backwards, however when dependent block ids exist, this reference is lost as we were re-assigning the variable: https://github.com/ONSdigital/eq-questionnaire-runner/blob/main/app/questionnaire/path_finder.py#L178-L181

This change ensures the reference is not overwritten, but instead we use a new variable and a copy of `routing_path_block_ids`. In future, we will look at rewriting pathfinder to not do such mutation which not only is unclear but prone to error as proven by this.

The updated schema simply moves the Confirmation question into the second section, which creates a when rule dependencies given the confirmation question's routing uses the first question. The additional dependencies are now populated because it is now in another section, previously it didn't need additional dependencies as it was parh of the same routing path.

### How to review
1. Use the updated test schema from this PR on `main`.
2. Get to the **Confirmation Question** (confirm-zero-employees-block) and select `No I need to correct this`
3. Confirm that on submit, the answer is removed, but you are not routed back. You should still be on the same page.
4. Checkout this branch
5. Repeat step 2
6. Confirm that on submit, you are routed back to the previous page (number-of-employees-total-block)

---

1. Test the scenario attached on the card and confirm it is no longer an issue.
2. Feel free to test any live surveys that have confirmation questions, which also has dependencies.

---
Since the issue is with a memory reference being overwritten, no new tests were needed other than updating the schema and changing some section ids in the existing test. You can ensure that the existing unit test `test_remove_answer_and_block_if_routing_backwards` is adequate but running this schema and unit test on `main` and you should see that the test fails as the routing paths do not match.

### Checklist


app/questionnaire/path_finder.py
```diff
@@ -224,16 +224,15 @@ def evaluate_skip_conditions(
         if not skip_conditions:
             return False
 
-        if when_rules_block_dependencies:
-            routing_path_block_ids = (
-                when_rules_block_dependencies + routing_path_block_ids
-            )
+        block_ids_for_dependencies = (
+            list(routing_path_block_ids) + when_rules_block_dependencies
+        )
```
MebinAbraham, line 229: This has no functional change, this was updated for consistency only.

app/questionnaire/path_finder.py
```diff
@@ -175,16 +175,16 @@ def _evaluate_routing_rules(
         routing_path_block_ids: list[str],
         when_rules_block_dependencies: list[str],
     ) -> int | None:
-        if when_rules_block_dependencies:
-            routing_path_block_ids = (
-                when_rules_block_dependencies + routing_path_block_ids
-            )
+        # Use `list` to create a shallow copy since routing_path_block_ids is mutated hence we don't to update its memory reference
```
berroar, line 178:
```suggestion
        # Use `list` to create a shallow copy since routing_path_block_ids is mutated hence we don't want to update its memory reference
```

app/questionnaire/path_finder.py
berroar: Do we need to cast `routing_path_block_ids` as a list here when it should already be a list as per the type hints?

berroar (approved):
Tested vs Main, changes fix the broken journey 👍

tests/app/questionnaire/test_path_finder.py
```diff
@@ -414,12 +414,17 @@ def test_remove_answer_and_block_if_routing_backwards():
                 "section_id": "default-section",
                 "list_item_id": None,
                 "status": CompletionStatus.COMPLETED,
+                "block_ids": ["route-backwards-block"],
```
petechd, line 417: Looks fine. Since this issue wasn't picked up by our existing tests shouldn’t we add some functional one to be safer in the future?

MebinAbraham: See the PR description around the test. This same test would fall on `main` because the routing path would be of different size.

app/questionnaire/path_finder.py
MebinAbraham: The casting was done to keep it consistent and also create a shallow copy in the event that this method ever mutated the `routing_path_block_ids`. Could remove casting, but it would be safer to keep. The change itself wasn't needed for this method was only changed for consistency.

petechd (approved):

PR #1316, opened 2024-02-09
Add automatic integration test generation script

petechd:
This adds python script which allows developers to complete a survey journey, and have their actions recorded and used to generate the python code that would execute that same journey as part of an integration test. It uses playwright library so the Pipfile had to be updated, some libraries were pinned until [this PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1284) merges. The `URLSafeSerializer` change was also taken from that PR to prevent the Mypy error. Also, duplicated test name removed. A new README has also been added to the `/tests/integration/` folder with details on how to run the script.

### How to review
Before you start make sure you re-installed the dependencies. Test it using the new makefile command. Check if generated integration test (modules) can be used in runner without any changes (except adding assertions).

### Checklist


berroar (commented):
Need to update the readme with some info on how to run the script, expected output etc

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,124 @@
...
+
+    if (
+        previous_request_method == ""
+        and survey_start
+        and request.url == "http://localhost:5000/questionnaire/"
+    ):
+        survey_start = False
+
+
+def request_handler(request: Request):
```
berroar, line 91: Noticed we type hinted some but not all methods?

petechd:
> Need to update the readme with some info on how to run the script, expected output etc

We were not sure about this since other scripts eg. validate-test-schema are not really explained, so we left it out.

berroar:
> > Need to update the readme with some info on how to run the script, expected output etc
> 
> We were not sure about this since other scripts eg. validate-test-schema are not really explained, so we left it out.

I feel in this instance it would be good for the dev to know that this exists and how it works, where it outputs the file etc. As I just discovered it overwrites the file if you create a test for the same schema on the second go, so would be good to call out stuff like that 👍

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,124 @@
...
+    for answer_id, answer_values in query_string.items():
+        if any(
+            answer_values
+        ):  # Only populate items if non-empty answers dict in request
+            if len(answer_values) > 1:  # Handle multiple values in list
+                items[answer_id] = answer_values
+            elif len(answer_values) == 1:
+                items[answer_id] = answer_values[0]  # A single item in list
+
+    if len(items) > 0:
```
berroar, line 55: Is this the same as doing if items? Then we wouldn't need the if below?

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,124 @@
...
+        survey_start = False
+
+
+def request_handler(request: Request):
+    global schema_name
+    matching_url_found = RUNNER_ROOT_URL in request.url
+
+    if matching_url_found:
+        process_request(request)
+    elif request.url != LAUNCHER_ROOT_URL and LAUNCHER_ROOT_URL in request.url:
```
berroar, line 97: Only minor but I think it breaks here if we use any of the sign out or save buttons, as the launcher root url is used as the account service url by default in launcher

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,124 @@
...
+        logger.info(f"self.post({items})")
+        return
+
+    if len(items) == 0:  # Empty post for no answers and non question pages
+        file.write("\n        self.post()")
+        logger.info("self.post()")
+
+
+# pylint: disable=global-statement
+def process_get(request, file):
```
berroar, line 66: I noticed when using post submission pages the `s` seems to be dropped from `submitted` e.g. `self.get("/ubmitted/view-response/")`

scripts/generate_integration_test.py
liamtoozer: Ah good spot, thanks!

Makefile
```diff
@@ -116,3 +116,8 @@ dev-compose-down-linux:
 
 profile:
 	pipenv run python profile_application.py
+
+generate-integration-test:
+	docker-compose -f docker-compose-dev-mac.yml up -d
```
berroar, line 121: Do we need this? If so, this should be a prerequisite in the readme for running the script, as it falls over now if it's already running?

Makefile
```diff
@@ -116,3 +116,8 @@ dev-compose-down-linux:
...
+	pipenv run python -m scripts.generate_integration_test
+	pipenv run black scripts/test_*
```
berroar, line 123: I get `Error: Invalid value for 'SRC ...': Path 'scripts/test_*' does not exist.` when trying to run this

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,141 @@
...
+                )
+                logger.info(f"self.launchSurvey('{schema_name}')")
+        else:
+            # capture launcher urls for sign-out, save etc
+            with open(f"{schema_name}.py", "a", encoding="utf-8") as file:
+                path = urlparse(request.url).path
+                file.write(generate_get_code_snippet(path))
+
+
+def generate_post_code_snippet(post_data: dict | str | None = "") -> str:
```
berroar, line 101: The post and get method could be merged as they return the same thing?

Makefile
petechd: Removed now, added something about it in the readme.

Makefile
petechd: Fixed now.

scripts/generate_integration_test.py
petechd: Merged now, one method.

rmccar (approved):

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,133 @@
...
+survey_start = True
+schema_name = ""
+
+LAUNCHER_ROOT_URL = "http://localhost:8000"
+RUNNER_ROOT_URL = "http://localhost:5000"
+
+
+# pylint: disable=global-variable-not-assigned
+def process_runner_request(request: Request) -> None:
+    global schema_name
```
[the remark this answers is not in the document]
liamtoozer: I think we tried to get around this, but we needed `schema_name` globally when we process the Runner request to open the file name after we grab it from the initial launcher request. In the Runner request, there's no `schema_name` to extract for use

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,133 @@
...
+                )
+                logger.info(f"self.launchSurvey('{schema_name}')")
+        else:
+            # capture launcher urls for sign-out, save etc
+            with open(f"./scripts/{schema_name}.py", "a", encoding="utf-8") as file:
+                path = f'"{urlparse(request.url).path}"'
+                file.write(generate_method_request("get", path))
+
+
+def generate_method_request(method: str, data: dict | str | None = "") -> str:
```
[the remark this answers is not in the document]
liamtoozer: Updated, thanks 👍

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,133 @@
...
+def process_get(request: Request, file: TextIO) -> None:
+    global previous_request_method
+    global survey_start
+
+    if (  # Logic to filter out the session and root questionnaire GET requests at the start
+        previous_request_method == "GET"
+        and "session?token" not in request.url
+        and request.url != f"{RUNNER_ROOT_URL}/questionnaire/"
+    ):
+        path = f'"{urlparse(request.url).path}"'
```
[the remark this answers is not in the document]
liamtoozer: I think this was so the self.post and self.get had the data in quotes. If we add the quotes in the `generate_method_request()` and data is `None`, it would end up with something like `self.post("")` instead of `self.post()`

tests/integration/README.md
```diff
@@ -0,0 +1,16 @@
...
+To generate the GET and POST requests of a user journey that can be run as an integration test, run the following make command from the project root folder:
+
+```shell
+make generate-integration-test
+```
+
+This will launch a browser pointing to the Launcher UI (make sure the application and supporting services are running). Now follow the below steps:
+
+1. Choose a schema and launch it - the schema name will be used for the name of the file
+1. Navigate through the survey. **Note:** all GETs and POSTs will be recorded
```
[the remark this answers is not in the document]
liamtoozer: Updated the readme with some extra bits of info, thanks

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def request_handler(request: Request) -> None:
+    if LAUNCHER_ROOT_URL in request.url:
+        process_launcher_request(request)
+    elif RUNNER_ROOT_URL in request.url:
+        process_runner_request(request)
+
+
+def run(pw: Playwright) -> None:
```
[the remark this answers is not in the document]
liamtoozer: We tried updating the ini file but couldn't get it working because when mypy [is invoked](https://github.com/ONSdigital/eq-questionnaire-runner/blob/main/scripts/run_mypy.sh#L3) it has the `app` folder. We added the script name in there instead as we couldn't find a graceful way around it?

tests/integration/README.md
berroar, line 1: Any reason why this lives in the tests directory rather than the scripts directory? On the flip side it seems odd that that the generated test file gets output to the scripts directory?

tests/integration/README.md
```diff
@@ -0,0 +1,30 @@
...
+### Details
+To speed up the process of generating integration tests for Runner, there is a dev-convenience script that records the GET and POST requests of a user journey
+and outputs this formatted in the style of an integration tests.
+
+There are a points to note for this script:
+
+* All GETs and POSTs will be recorded
+* The dev will need to manually add their own assertions in the generated test file
+* It does **not** handle dynamic answers because these are generated at runtime
+* It handles navigating back via clicking 'previous' links etc. To do this, some additional logic was needed to filter out the beginning of the survey journey,
```
berroar, line 14: Maybe a point about list_item_ids needing to be handled after the test has been generated?

scripts/README.md
```diff
@@ -0,0 +1,34 @@
...
+There are a few points to note for this script:
+
+* All GETs and POSTs will be recorded
+* The dev will need to manually add their own assertions in the generated test file
+* It does **not** handle dynamic answers because these are generated at runtime - you will need to update the output script to handle list_item_ids separately,
+  as they will not be known beforehand.
+* It handles navigating back via clicking 'previous' links etc. To do this, some additional logic was needed to filter out the beginning of the survey journey,
+  after the first GET request from launcher that makes the first GET request in Runner
+
+### Usage
```
berroar, line 19: The only other thing I can think is that if you try to generate a second test for the same schema, it will overwrite the first one. Is it worth calling out it is a one at a time thing?

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,137 @@
...
+RUNNER_ROOT_URL = "http://localhost:5000"
+TEST_TEMPLATE = """from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class {class_name}(IntegrationTestCase):
+    def {function_name}(self):
+        self.launchSurvey("{schema_name}")
+"""
+
+survey_journey: Dict[str, str | bool] = {
```
berroar, line 19: Should these be None rather than ""?

scripts/README.md
liamtoozer: That's a good point! Will update readme and the spike doc as it would be good for the script to maybe warn (or fail) in future

berroar (approved):
LGTM 👍

rmccar (approved):

scripts/generate_integration_test.py
```diff
@@ -50,49 +51,62 @@ def process_post(request: Request, file: IO) -> None:
...
                 )
-                logger.info(request)
-        else:
-            # capture launcher urls for sign-out, save etc
-            with open(output["file_name"], "a", encoding="utf-8") as file:
-                path = f'"{urlparse(request.url).path}"'
-                file.write(generate_method_request(method="get", data=path))
+            )
+            logger.info(request)
+    else:
```
MebinAbraham, line 105: If you are always going to check both cases, i.e if true and when false, then avoid using inversing via `not` as the first check as it makes it less clear.

```py
if True:
    do_stuff()
else:
    do_other_stuff()
```

Over:

```py
if not True:
   do_other_stuff()
else:
   do_stuff()
```

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,133 @@
+from typing import TextIO
+from urllib.parse import parse_qs, urlparse
+
+from playwright.sync_api import Playwright, Request, sync_playwright
+from structlog import get_logger
+
+logger = get_logger()
+
+
+previous_request_method = ""
```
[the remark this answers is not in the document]
liamtoozer: Updated the readme with extra info and added a short docstring to one of the functions to hopefully show intent a bit more, thanks. Let me know if I need to pop any more info there as I wasn't sure how in-depth to go

scripts/generate_integration_test.py
liamtoozer: I think I tried to put the 'first' action of the file creation up front, but agree it's more difficult to read with inversion. Definitely makes that more readable, thanks 👍

Yuyuutsu (commented):
Tested and it works as expected 👍 . Not sure if this is an issue for this PR but it seems the chromium page has a set size even when full screened.

liamtoozer:
> Tested and it works as expected 👍 . Not sure if this is an issue for this PR but it seems the chromium page has a set size whcih stays the same even when full screened.

Yeah good spot! It looks like chromium renders pages in a viewport by default, presumably for consistent testing when running headlessly 🤔? 

Updated to fullscreen with no viewport 👍

scripts/generate_integration_test.py
```diff
@@ -0,0 +1,146 @@
...
+logger = get_logger()
+
+LAUNCHER_ROOT_URL = "http://localhost:8000"
+RUNNER_ROOT_URL = "http://localhost:5000"
+
+TEST_TEMPLATE = """from tests.integration.integration_test_case import IntegrationTestCase
+
+
+class {class_name}(IntegrationTestCase):
+    def {function_name}(self):
```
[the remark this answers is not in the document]
MebinAbraham: As discussed, we don't need to do this given we are using it for test schemas which always has the `test` prefix.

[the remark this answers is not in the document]
liamtoozer: Will update the readme and the spike with this info, as it may trip someone up if they use a non-test schema

liamtoozer:
Added do-not-merge label: Investigating some issues with mypy, after Pipfile dependency update after adding playwright

MebinAbraham (approved):

PR #1318, opened 2024-02-16
Support piping repeating supplementary data outside a repeat

petechd:
This PR adds piping of a repeating supplementary data outside a repeating (previously we prevented this, normal list sources were still allowed to be piped). Supplementary data source list elements can be now added to elements like descriptions, guidances and wherever we use placeholders. Some extra logic had to be added since the previous list resolution was assumed to happen in a repeat context, so list_item_id from a list was always present.

### How to review
Two new transforms using supplementary data value source were added to the first Interstitial page in `test_supplementary_data` schema to showcase the changes. Few test cases added to existing unit test (testing the `get_data` method in Supplementary Data Store class).

### Checklist


petechd:
[Support piping repeating supplementary data outside a repeat (1 SP)](https://trello.com/c/aOgbTIth/6021-support-piping-repeating-supplementary-data-outside-a-repeat-1-sp)

rmccar (approved):

app/data_models/supplementary_data_store.py
```diff
@@ -114,11 +114,21 @@ def get_data(
...
-
+        list_items_mapping = self.list_lookup.get(identifier)
+        if (
+            not list_item_id and list_items_mapping
+        ):  # check if we are NOT in repeat, the other method, is_data_repeating(), cannot be used here as it always evaluates to True
+            values = []
+            for list_item_id_key in list_items_mapping:
+                list_item_id = list_items_mapping[list_item_id_key]
+                values.append(self.resolve_value(identifier, selectors, list_item_id))
+            return values
```
MebinAbraham, line 125: Few things.

- Use the list look up only when needed, i.e if we don't have list item id
- Use `.values` to avoid looping over the list, to then just fetch the values, that why `.values()` is useful
- This doesn't handle a value for one of the list not existing. We would need to handle that, as we handle it when fetching answers beloning to a list.
- Optional, use a comphrension

```py
        if not list_item_id and (
            list_item_ids := self.list_lookup.get(identifier, {}).values()
        ):
            return [
                value
                for _list_item_id in list_item_ids
                if (value := self.resolve_value(identifier, selectors, _list_item_id))
            ]

        return self.resolve_value(identifier, selectors, list_item_id)
```

You will need to update the test to ensure if the value is missing for one of the item, then it is not rendered. You will need to update the SDS dataset if we don't have an existing example we can use.

MebinAbraham (commented):
We should update the existing functional test to cover the new changes.

schemas/test/en/test_supplementary_data.json
```diff
@@ -58,11 +58,52 @@
                                 "contents": [
                                     {
                                         "title": "You have successfully loaded Supplementary data",
-                                        "description": "Press continue to proceed to the introduction",
+                                        "description": {
+                                            "text": "List of products: {products}",
```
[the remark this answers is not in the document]
petechd: With other list (employees) we run into the problem of first names and then last names (in case of employees) being rendered in sequence, so, for those not involved in this conversation before, list like: [`first name person a`, `first name person b`, `last name person a`, `last name person b`], as it first loops through the first then the second, that's how it was designed. As long as we are happy, for example, with only last names on the list, then I can add it, otherwise some bigger changes will be needed.

app/data_models/supplementary_data_store.py
```diff
@@ -114,11 +114,21 @@ def get_data(
...
         """
-        if self.is_data_repeating(identifier) and not list_item_id:
-            raise InvalidSupplementaryDataSelector(
-                f"Cannot reference items from `{identifier}` outside a repeat"
-            )
-
+        list_items_mapping = self.list_lookup.get(identifier)
+        if (
+            not list_item_id and list_items_mapping
+        ):  # check if we are NOT in repeat, the other method, is_data_repeating(), cannot be used here as it always evaluates to True
```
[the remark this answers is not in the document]
petechd: No need to mention about `is_data_repeating()` as it's not needed in this if statement so removing it now.

VirajP1002 (commented):
Works functionally as I can see the changes on the webpage the unit tests pass locally and changes make sense to me.

schemas/test/en/test_supplementary_data.json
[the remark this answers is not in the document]
MebinAbraham: Yes, that's fine.

app/data_models/supplementary_data_store.py
MebinAbraham: Re: the above code should check the `value is not None` instead of `if value`, otherwise it will discard falsey value such as `0`.

petechd: I had to leave list comprehension and make it normal loop, as it was not allowing me to add this logic. If I added `if value := .... and value is not None` it was evaluating it and adding `True` and `False` to the list of values, same with `if value := .... is not None`. So opted for normal for loop instead.

petechd:
> We should update the existing functional test to cover the new changes.

Added to existing functional test spec, also that unit test with value resolving to `None` added as well, had to create extra supplementary data fixture structure for this, as you predicted.

app/data_models/supplementary_data_store.py
```diff
@@ -114,11 +114,31 @@ def get_data(
         For example if you wanted the identifier for the first item in "some_list"
         it would be get_data(identifier="some_list", selectors=["identifier"], list_item_id=list_item_id-1)
         """
-        if self.is_data_repeating(identifier) and not list_item_id:
-            raise InvalidSupplementaryDataSelector(
-                f"Cannot reference items from `{identifier}` outside a repeat"
-            )
+        if not list_item_id and (
```
berroar, line 117: Only minor, but is there any reason why we moved away from the original if statement here which I think is slightly more readable? If we revert the for can be updated to `for _list_item_id in self.list_lookup.get(identifier).values()` as we will know the identifier is in the store

petechd: Whichever is more readable, I've added back the original one, thanks!

tests/functional/spec/features/supplementary_data/supplementary_data.spec.js
```diff
@@ -46,8 +46,13 @@ describe("Using supplementary data", () => {
       responseId,
     });
   });
+  it("Given I launch a survey using supplementary data, When I begin the interstitial block, Then I see the repeating supplementary data outside of the repeat piped in", async () => {
```
berroar, line 49: Applies to many tests but I think we should make our Given/When/Then's generic, so wouldn't say on the interstitial block. Maybe something like "When I am outside a repeating section, Then I am able to see the list of items relating to a given supplementary data list item on the page"

berroar (approved):

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -1165,53 +1180,52 @@ def test_supplementary_data_value_source_list_items(
     )
 
 
-def test_supplementary_data_invalid_selector_raises_exception(
-    supplementary_data_store_with_data,
+def test_supplementary_data_value_source_list_items_extra_item_not_resolving_when_no_value(
```
MebinAbraham, line 1183: I find this hard to understand. What is it testing?. `extra item not resolving when no value`?. It's not clear what that means.

app/data_models/supplementary_data_store.py
```diff
@@ -115,10 +115,28 @@ def get_data(
...
+                )
+                if value is not None:
+                    values.append(value)
+            return values
+
+        return self.resolve_value(
+            identifier=identifier, selectors=selectors, list_item_id=list_item_id
+        )
 
+    def resolve_value(
```
MebinAbraham, line 133: We should make this private. The public entry point is `get_data`.

tests/app/questionnaire/test_value_source_resolver.py
petechd: Added the initial description you used when requesting this test.

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -1165,53 +1180,52 @@ def test_supplementary_data_value_source_list_items(
     )
 
 
-def test_supplementary_data_invalid_selector_raises_exception(
-    supplementary_data_store_with_data,
+def test_supplementary_data_value_source_list_items_value_missing_not_rendered(
```
MebinAbraham, line 1183: `not_rendered` wouldn't be correct here as the test does not check that. `Rendering` can only be checked by testing Placeholder Renderer or a functional here. Here, it would be better to use `value_missing_ignored`.

MebinAbraham (approved):

app/data_models/supplementary_data_store.py
```diff
@@ -115,10 +115,28 @@ def get_data(
         it would be get_data(identifier="some_list", selectors=["identifier"], list_item_id=list_item_id-1)
         """
         if self.is_data_repeating(identifier) and not list_item_id:
```
MebinAbraham, line 117: We can revisit this, but this is technically flawed.
- If it is repeating, and we have a list item id then validator should have prevent that (it doesn't since we have added validation for then).
- Alternatively, since SDS is immutable, we could allow self reference, which would be a case of removing the `not list_item_id` check.

Right not, validator will allow something that runner will ignore. Happy to keep it as is for now and revisit if it becomes an issue.

petechd: Ideally, I would like to finish this off before I leave, maybe just revisit and improve after?

MebinAbraham: Ya, it's not something we will do now. Highlighted it for awareness that there is a small gap.

PR #1322, opened 2024-02-21
Update Runner Python Packages 

Yuyuutsu:
Update python packages. 
Changes Made:
- Fixed any new type hinting errors found 
- Flake8 Config has been updated to accommodate new black update. [Black docu
(https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#:~:text=Flake8%23,E203%2C%20E704)
- `Pylint` has been pinned due to problems with how `Pylint` interacts with `pylint_quotes`in versions`3.*.*`. There has been an issue raised and a PR to resolve the issue with `pylint_quotes`, but owner of the repoistory has not merged yet. See [issue here](https://github.com/edaniszewski/pylint-quotes/issues/29)
- `dynamodb()` test has been updated to use `mock_aws` due to [`moto v5.0.0`](https://github.com/getmoto/moto/blob/f9add957c9f8311f7e1aa927de7a83f2cb5adce2/CHANGELOG.md#L4)
- Google Publisher Mock has been updated

### How to review
- Check if tests and runner runs as expected
- Check if changes make sense

#### Main Branch
Baseline benchmark

#### Update Runner Python Package 
Update Runner Python Package branch

### Checklist


Yuyuutsu:
[Update runner to use poetry package manager (2 SP)](https://trello.com/c/9rGxXogW/5930-update-runner-to-use-poetry-package-manager-2-sp)

tests/app/test_setup.py
```diff
@@ -1,6 +1,9 @@
+from unittest.mock import MagicMock
```
berroar, line 1: We should get this from pytest mock rather than unitttest, think we tried to move away from unittest outside of our integration tests 👍

Yuyuutsu: updated to use: `from mock import MaigcMock`

berroar (commented):
Looks good, just needs a benchmark run as discussed 👍

berroar (changes requested):
Need to regenerate the piplock file as we have now updated the sdc-cryptography repo to fix some dependabot issues: https://github.com/ONSdigital/sdc-cryptography/releases/tag/1.1.4

berroar (approved):

Pipfile
```diff
@@ -8,7 +8,7 @@ verify_ssl = true
 mock = "*"
 pytest-cov = "*"
 jsonschema = "*"
-pylint = "*"
+pylint = "2.17.7"
```
MebinAbraham, line 11: Isn't the issue with version 3?. If so, the pin should be set to less than version 3.

app/helpers/url_safe_serializer.py
```diff
@@ -6,4 +6,5 @@
 
 
 def url_safe_serializer() -> URLSafeSerializer:
-    return URLSafeSerializer(current_app.secret_key, salt=cookie_session[EQ_SESSION_ID])
+    # Type Ignore: Changing secret key at this point would be too late.
```
MebinAbraham, line 9: This type ignore comment is confusing. The comment should be around the fact that we validate that it exists on app start up, hence it must exist at this point.

Pipfile
Yuyuutsu: updated

petechd (approved):

setup.cfg
```diff
@@ -10,4 +10,4 @@ filterwarnings=
 # Ignore node_modules and cloned repos when not in a virtual environment
 exclude = node_modules/*,tests/*,src/*
 max-line-length = 160
-ignore = C815,C816,W503,E203
+ignore = C815,C816,W503,E203,E704
```
liamtoozer, line 13: Out of scope for this PR, but it may be useful to have inline comments/links on why we're ignoring some of these rules?

VirajP1002 (commented):
LGTM! All the tests pass and that changes make sense to me!

liamtoozer (approved):
Works great for me! 👍

setup.cfg
liamtoozer: Ticket raised to look into it 👍

PR #1324, opened 2024-02-26
Add backup name env var handling in restore CI task

petechd:
Currently we don't handle situations where the BACKUP_NAME env var is not provided (and there is no default hardcoded). The script tries to extract export metadata with no backup name appended to it and it fails but the error is not clear. This updates the script so it warns that the var needs to be set.

### How to review
[Fly execute ](https://concourse-ci.org/tasks.html#running-tasks) the `restore_questionnaire_state.yaml` file, without providing the BACKUP_NAME var and check if it errors and you get the `"BACKUP_NAME variable must be set"` message. Fly execute with BACKUP_NAME and check if you don't get the same error.

### Checklist


petechd:
[Fix restore_questionnaire_state script backup_name env var handling](https://trello.com/c/S6boxa4D/6067-fix-restorequestionnairestate-script-backupname-env-var-handling)

ci/restore_questionnaire_state.yaml
```diff
@@ -17,6 +17,12 @@ run:
   args:
     - -exc
     - |
+      if [ -z "${BACKUP_NAME}" ]
+      then
+        echo "BACKUP_NAME variable must be set"
+        exit 1
+      fi
```
MebinAbraham, line 24: We should check all the variables. It would still error if other params were missing but BACKUP name was passed?.

petechd: The other vars have default values so it wouldn't error, if used in pipeline, I mean.

petechd: It's a loop now and checks all the vars (if they are set), thanks.

ci/restore_questionnaire_state.yaml
```diff
@@ -17,6 +17,17 @@ run:
   args:
     - -exc
     - |
+      VARS=('PROJECT_ID' 'BUCKET_NAME' 'BACKUP_NAME' 'FILE_PREFIX')
```
berroar, line 20: Really minor but i I'd call these `PARAMS` rather than vars to keep it consistent throughout the file.

ci/restore_questionnaire_state.yaml
```diff
@@ -17,6 +17,17 @@ run:
...
+
+      for VAR in "${VARS[@]}"
+      do
+        if [ -z "${!VAR}" ]
+        then
+          echo "$VAR variable must be set"
+          exit 1
+        fi
+      done
```
MebinAbraham, line 29: Minor, consider using bash expansion or reduce the num lines but have `do` and `then` on the same line.

```bash
PARAMS=('PROJECT_ID' 'BUCKET_NAME' 'BACKUP_NAME' 'FILE_PREFIX')
for param in "${PARAMS[@]}"; do
  : "${!param:?\"Error: $param must be set\"}"
done
``

Comment (petechd) [reply ci/restore_questionnaire_state.yaml]: Added, thanks.

File: ci/restore_questionnaire_state.yaml (hunk 2 above)
Comment (petechd) [reply ci/restore_questionnaire_state.yaml]: Added, thanks.

File: ci/restore_questionnaire_state.yaml [hunk 4]
```diff
@@ -17,6 +17,11 @@ run:
   args:
     - -exc
     - |
+      PARAMS=('PROJECT_ID' 'BUCKET_NAME' 'BACKUP_NAME' 'FILE_PREFIX')
+      for PARAM in "${PARAMS[@]}"; do
+      : "${!PARAM:?"Error: $PARAM must be set"}"
```
MebinAbraham, line 22: Minor should be indented.

MebinAbraham (approved):

berroar (approved):

ci/restore_questionnaire_state.yaml
petechd: Added indentation and tested.

PR #1330, opened 2024-03-06
Fix placeholder transform resolution issue in calculation summary pages

berroar:
Fixes an issue where placeholders were not being evaluated correctly on Calculation Summary pages (CS, GCS) as the dependency map `_placeholder_transform_section_dependencies_by_block` was not being being populated correctly for these block types. 

Previously, Calculation Summary blocks were only accounting for any placeholders set in the specific CS/GCS block in the schema, but Calculated Summary pages can render question text (containing placeholders) from their dependent questions, so the Calculation Summary page may have a dependency on another block which then may have dependencies on other sections. The logic has now been updated to account for this and update the dependency map correctly.

The code was added to support Grand Calculated Summaries but at the moment it is not possible to add any tests for this as unlike Calculated Summaries they do not render question text from their dependent blocks.

### How to review

Use the `test_placeholder_dependencies_with_calculation_summaries` schema to test the change.
1. Launch `test_placeholder_dependencies_with_calculation_summaries`
2. Don’t accept default reporting period, enter new values and complete section 2
3. Go back to Section 1 summary screen, change dates back to default values by answering ‘Yes, I can report for this period’ for the first question.
4. Go back to Section 2 and get to the calculated summary by changing one of the answers that are used on the CS.
5. Check that the dates set in the metadata are rendered on the Calculated Summary page, not the dates that were previously entered by the user.

### Checklist


berroar:
[Bug: Period Date changes not reflecting in piped questions](https://trello.com/c/UzvOhN8A/6070-bug-period-date-changes-not-reflecting-in-piped-questions)

app/questionnaire/questionnaire_schema.py
```diff
@@ -1385,27 +1385,47 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
...
-                if item.get("source") == "answers"
-            }
-            placeholder_dependencies = self._get_section_ids_for_answer_ids(
-                answer_ids=placeholder_answer_ids
-            )
-            if placeholder_dependencies:
-                # Type Ignore: At this point we section id  and block id cannot be None
-                section_id = self.get_section_id_for_block_id(block["id"])
-                self._placeholder_transform_section_dependencies_by_block[section_id][  # type: ignore
+            answer_ids: list = []
```
petechd, line 1404: I guess this is wrong cause the block cannot be CS and GCS at the same time, so:
```python
            if block["type"] == "GrandCalculatedSummary":
                answer_ids = self.get_answer_ids_for_grand_calculated_summary_id(
                    block["id"]
                )
            elif block["type"] == "CalculatedSummary":
                answer_ids = get_calculated_summary_answer_ids(block)
```
or even:
```python
            if block["type"] == "GrandCalculatedSummary":
                answer_ids = self.get_answer_ids_for_grand_calculated_summary_id(
                    block["id"]
                )
            elif block["type"] == "CalculatedSummary":
                answer_ids = get_calculated_summary_answer_ids(block)
            else:
               answer_ids = []
```
or something similar..

app/questionnaire/questionnaire_schema.py
```diff
@@ -1385,27 +1385,47 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
...
+            if block["type"] == "GrandCalculatedSummary":
+                answer_ids = self.get_answer_ids_for_grand_calculated_summary_id(
                     block["id"]
-                ].update(
-                    placeholder_dependencies
                 )
+            if block["type"] == "CalculatedSummary":
+                answer_ids = get_calculated_summary_answer_ids(block)
+
+            if answer_ids:
```
petechd, line 1412: `if answer_ids` and `if placeholder_answer_ids` are exclusive for a block so make them `if/elif`?

berroar: Updated 👍

petechd (approved):

tests/app/questionnaire/test_questionnaire_schema.py
```diff
@@ -1090,3 +1090,27 @@ def test_grand_calculated_summary_when_rule_dependencies():
...
+    Ensures that dependencies are captured correctly for calculation summary blocks using transforms.
+    In this schema the calculation summaries use placeholder transforms based on other blocks that have dependencies in the
+    reporting-period-section
+    """
+    schema = load_schema_from_name(
+        "test_placeholder_dependencies_with_calculation_summaries"
+    )
+
+    assert (
+        schema._placeholder_transform_section_dependencies_by_block  # pylint: disable=protected-access
```
MebinAbraham, line 1106: This should be using the public property instead of the private attribute.

app/questionnaire/questionnaire_schema.py
```diff
@@ -1385,27 +1385,48 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
...
                 )
+            elif block["type"] == "CalculatedSummary":
+                answer_ids = get_calculated_summary_answer_ids(block)
+            else:
+                answer_ids = []
+
+            if answer_ids:
+                dependent_blocks = [
+                    self.get_block_for_answer_id(answer_id) for answer_id in answer_ids
+                ]
```
MebinAbraham, line 1416: Answer ids might contain duplicate hence you will get duplicate block which is wasteful. Only fetch the blocks for unique answer ids.

app/questionnaire/questionnaire_schema.py
```diff
@@ -1385,27 +1385,48 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
                 if item["for_list"] == list_name and item.get("item_anchor_answer_id"):
                     return f"#{str(item['item_anchor_answer_id'])}"
 
+    def _update_placeholder_dependencies(
+        self, placeholder_answer_ids: set, block: ImmutableDict
```
MebinAbraham, line 1389: Type hings for answer ids needs to be stricter. Also, you don't need `placeholder` in the variable name, it is already in the context of placeholders.

app/questionnaire/questionnaire_schema.py
```diff
@@ -1385,27 +1385,48 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
                 if item["for_list"] == list_name and item.get("item_anchor_answer_id"):
                     return f"#{str(item['item_anchor_answer_id'])}"
 
+    def _update_placeholder_dependencies(
```
MebinAbraham, line 1388: For consistency with our other dependencies method, this should be something like `_update_dependencies_for_placeholders`

app/questionnaire/questionnaire_schema.py
```diff
@@ -1385,27 +1385,48 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
...
+        self, placeholder_answer_ids: set, block: ImmutableDict
+    ) -> None:
+        if placeholder_dependencies := self._get_section_ids_for_answer_ids(
```
MebinAbraham, line 1391: Existing, but it would be clearer if this was named `dependent_sections`

app/questionnaire/questionnaire_schema.py
```diff
@@ -1385,27 +1385,48 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
...
+            if block["type"] == "GrandCalculatedSummary":
+                answer_ids = self.get_answer_ids_for_grand_calculated_summary_id(
                     block["id"]
-                ].update(
-                    placeholder_dependencies
                 )
+            elif block["type"] == "CalculatedSummary":
+                answer_ids = get_calculated_summary_answer_ids(block)
+            else:
+                answer_ids = []
```
MebinAbraham, line 1411: Worth adding code comments to document that Calculation summaries may indirectly contain placeholder dependences as they may reference blocks which uses placeholder transform requiring the path.

app/questionnaire/questionnaire_schema.py
```diff
@@ -1470,3 +1491,15 @@ def get_calculated_summary_ids_for_grand_calculated_summary(
 
 def is_list_collector_block_editable(block: Mapping) -> bool:
     return bool(block["type"] == "ListCollector")
+
+
+def get_placeholder_answer_ids_for_transform(block: ImmutableDict) -> set[str]:
```
MebinAbraham, line 1496: `for_transform` implies you are passing it a transform object?. What about `get_placeholder_answer_ids_requiring_routing_path`

tests/app/questionnaire/test_questionnaire_schema.py
berroar: Updated 👍

app/questionnaire/questionnaire_schema.py

app/questionnaire/questionnaire_schema.py

app/questionnaire/questionnaire_schema.py

app/questionnaire/questionnaire_schema.py

app/questionnaire/questionnaire_schema.py

app/questionnaire/questionnaire_schema.py
```diff
@@ -1385,22 +1385,25 @@ def get_item_anchor(self, section_id: str, list_name: str) -> str | None:
                 if item["for_list"] == list_name and item.get("item_anchor_answer_id"):
                     return f"#{str(item['item_anchor_answer_id'])}"
 
-    def _update_placeholder_dependencies(
-        self, placeholder_answer_ids: set, block: ImmutableDict
+    def _update_dependencies_for_placeholders(
+        self, answer_ids: set[str], block: ImmutableDict
```
MebinAbraham, line 1389: Does this need to be `set` as per style guide make it asbtract, i.e Iterable[str]`?

MebinAbraham (approved):

Yuyuutsu (commented):
Changes look good to me 👍

PR #1336, opened 2024-03-14
Update heading level for introduction questions

petechd:
This updates the twisties for the ‘Informaton you need’ section on the Introduction page so they are heading 3 level  nowsince they are a sub heading.

### How to review
Open test introduction schema and use dev tools to check if intro questions elements are h3 now (elements with text "Employee pay frequency", "What you need to know", "Definition of innovation"). Check the amended functional test and the new introduction test page element I've added. 

### Checklist


tests/functional/spec/introduction.spec.js
```diff
@@ -6,7 +6,7 @@ describe("Introduction page", () => {
     await browser.openQuestionnaire(introductionSchema);
   });
 
-  it("Given I start a survey, When I view the introduction page, Then I should be able to see introduction information", async () => {
+  it("Given I start a survey, When I view the introduction page, Then I should be able to see introduction information displayed correctly", async () => {
```
berroar, line 9: Minor (and these and preexisting), but these are slightly ambiguous to me. For your test specifically I would use `When preview content is set on the introduction page, then the content headings should be displayed at the correct level` or something similar? But this has been added to the existing test and may not make sense to have it separate

berroar (approved):
Compared with staging, LGTM 👍

tests/functional/spec/introduction.spec.js
petechd: I will separate it out since your test description is on point.

tests/functional/spec/introduction.spec.js
```diff
@@ -18,6 +18,11 @@ describe("Introduction page", () => {
       "To take part, all you need to do is check that you have the information you need to answer the survey questions.",
     );
   });
+  it("Given I start a survey, When preview content is set on the introduction page, Then the content headings should be displayed at the correct level", async () => {
+    await browser.openQuestionnaire(introductionSchema);
+    await expect(await $(IntroductionPage.previewQuestions()).isDisplayed()).toBe(true);
+    await expect(await $(IntroductionPage.introQuestion()).getHTML()).toContain("h3");
```
MebinAbraham, line 24: Try to be more explicit. Currently, it would pass if anything in the HTML with this ID has the text `h3`.

Try something like this but using the selector in the look up.
```js
const introQuestionH3Selector = `${IntroductionPage.introQuestion()} h3`;
const h3Exists = await $(introQuestionH3Selector).isExisting();
await expect(h3Exists).toBe(true);
```

petechd: Ok, tried it, this is much more explicit so keeping it in. Thanks.

rmccar (approved):

Yuyuutsu (approved):
Tested and changes look good to me 👍

MebinAbraham (approved):

PR #1339, opened 2024-03-25
Update eQ to DS 68.0.2 for GA change

rmccar:
Updates eQ to version 68.0.2 of the DS to pull in the change for GA where custom events can be sent.

### How to review
Ensure:
- visually, Runner is as expected.
- Test GA custom events. 
	- Set up a dummy GA account for test purposes. Docs: https://support.google.com/analytics/answer/9304153?hl=en
	- Link (set `EQ_GOOGLE_TAG_ID` env var ) your Google tag ID with Runner (local or deployed)
	- Accept Cookies in the Cookie banner
	- Send some events via Runner like link clicks (ensure they are valid, i.e has the appropriate attributes set (cross-reference with our Confluence doc). **Make sure you don't have any adblock extensions active.**
	- Verify you can see the events in GA > Reports > Realtime (Events section in the bottom half of the page)

### Checklist


Yuyuutsu (commented):
Able to see the prefixes `ons_click` and `ons_error`. 
image

petechd (approved):

petechd:
Screenshot 2024-03-27 at 12 03 11

MebinAbraham (approved):
There is a bug that needs to be fixed, but I will approve this as it does work for most cases. A follow-up PR can fix the bug as a DS change is required.

Some of the elements aren’t tracked due to the way the JS in the DS is done. The JS checks if the target that was clicked has a data ga attribute however, when you have things like an anchor which has nested elements like span, the target is the span itself but the event won’t trigger as the attributes are set on the anchor.

```html
<a
  href="/questionnaire/non-mandatory-checkbox/?return_to=final-summary&amp;return_to_answer_id=non-mandatory-checkbox-answer#non-mandatory-checkbox-answer"
  class="ons-summary__button"
  data-qa="non-mandatory-checkbox-answer-edit"
  data-ga="click"
  data-ga-category="Summary"
  data-ga-action="Edit click"
>
  <span class="ons-summary__button-text" aria-hidden="true">
    Change
  </span>
  <span class="ons-u-vh">
    Change your answer for: What extra toppings would you like?
  </span>
</a>;
```

In the example above, the attributes are set on the anchor tag but the target the DS gets back when a click happens is span hence the tracking doesn’t work. This will need to be fixed in the DS.

liamtoozer (approved):
Works great for me 👍

PR #1348, opened 2024-04-05
Create test cases for rules that compare to a boolean 

Yuyuutsu:
This PR adds test cases for the `boolean_flag` when rule to show it works as expected.

### How to review

- Run functional tests and check if they work as expected
- Check Description given for the tests are acceptable

### Checklist

### DIFF
--- a/tests/functional/spec/features/routing/boolean.spec.js
+++ b/tests/functional/spec/features/routing/boolean.spec.js
@@ -1,0 +1,22 @@
+import Block1Page from "../../../generated_pages/metadata_routing/block1.page"
+import Block2Page from "../../../generated_pages/metadata_routing/block2.page"
+import Block3Page from "../../../generated_pages/metadata_routing/block3.page"
+import { click } from "../../../helpers";
+
+describe("Feature: Routing - Boolean Flag", () =>{
+    it("Given Boolean Flag is False, When I press continue, Then I should be taken to the Block2Page", async ()=> {
+        await browser.openQuestionnaire("test_metadata_routing.json",{
+            booleanFlag: false,
+        });
+        await click(Block1Page.submit());
+        await expect(browser).toHaveUrlContaining(Block2Page.pageName);
+    })
+
+    it("Given Boolean Flag is True, When I press continue, Then I should skip Block2Page and be taken to Block3Page", async() =>{
+        await browser.openQuestionnaire("test_metadata_routing.json",{
+            booleanFlag: true
+        });
+        await click(Block1Page.submit());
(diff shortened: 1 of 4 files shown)


VirajP1002 (commented):
Tests work and the descriptions look good to me! 🎉 Just out of curiosity, why is `booleanFlag` set to null? I tried set it to false in both `wdio.conf.js` and `jwt_helper` and the tests pass?

Yuyuutsu:
> Tests work and the descriptions look good to me! 🎉 Just out of curiosity, why is `booleanFlag` set to null? I tried set it to false in both `wdio.conf.js` and `jwt_helper` and the tests pass?

My assumption thought that since the `booleanFlag` is not always present when launching a schema, it should be null. Happy to change 👍

VirajP1002:
> > Tests work and the descriptions look good to me! 🎉 Just out of curiosity, why is `booleanFlag` set to null? I tried set it to false in both `wdio.conf.js` and `jwt_helper` and the tests pass?
> 
> My assumption thought that since the `booleanFlag` is not always present when launching a schema, it should be null. Happy to change 👍

That makes sense to me 👍

VirajP1002:
Just had a look again at the booleanFlag PR, and maybe we should set it to False because there's another boolean above that's set to True for consistency, and I guess the default for boolean_flag is False, so maybe we should do that? 
I also think that Null can be set to false behind the scenes in JS and that might cause confusion in types

tests/functional/spec/features/routing/boolean.spec.js
```diff
@@ -0,0 +1,22 @@
+import Block1Page from "../../../generated_pages/metadata_routing/block1.page";
+import Block2Page from "../../../generated_pages/metadata_routing/block2.page";
+import Block3Page from "../../../generated_pages/metadata_routing/block3.page";
+import { click } from "../../../helpers";
+
+describe("Feature: Routing - Boolean Flag", () => {
+  it("Given Boolean Flag is False, When I press continue, Then I should be taken to the Block2Page", async () => {
```
berroar, line 7: The descriptions for tests like this should be generic, and could be made more meaningful. So needs to be reworded so that it describes the behaviour of the feature i.e `Given I have a routing rule that uses a boolean flag` etc

VirajP1002 (approved):

tests/functional/spec/features/routing/boolean.spec.js
Yuyuutsu: Updated the `it` function descriptions

PR #1364, opened 2024-05-13
Type hints for Publisher Module 

VirajP1002:
- Added type hinting to the Publisher module to align with other modules with type hinting in place.

### How to review
- Ensure type hints are correct and if amendments are required

### Checklist


rmccar (approved):

app/publisher/publisher.py
```diff
@@ -13,22 +14,26 @@
 
 class Publisher(ABC):
     @abstractmethod
-    def publish(self, topic_id, message, fulfilment_request_transaction_id):
+    def publish(
+        self, topic_id: str, message: bytes, fulfilment_request_transaction_id: str
+    ) -> Any:
```
liamtoozer, line 19: Would the return type probably be `None`, instead of the generic `Any`?

app/publisher/publisher.py
```diff
@@ -13,22 +14,26 @@
...
 
 
 class PubSubPublisher(Publisher):
-    def __init__(self):
+    def __init__(self) -> None:
         self._client = PublisherClient()
         _, self._project_id = google.auth.default()
 
-    def _publish(self, topic_id, message) -> "publisher.futures.Future":
+    def _publish(self, topic_id: str, message: bytes) -> "publisher.futures.Future":
```
Yuyuutsu, line 28: Should we use the `Future` class rather than the string?
However it was there before so i could be totally wrong 😂

VirajP1002: When I was debugging that method, it passes the `topic_id` into the method called `topic_path` and that method returns a formatted string with `topic_id` as part of it, so I thought it would have to be a string otherwise. I also just saw that the type hint for that method has also set `topic_id` to str, which I completely missed 😂

VirajP1002: Happy for anyone to double check as I may be wrong also

Yuyuutsu: Oh no i meant the "publisher.futures.Future". Right now I think it is basially -> Any

VirajP1002: Ahh I see what you mean know, so I set it to `Future` and it seems fine, no issues with mypy or linting. I think it was set to that because the method it calls in `response: Future = self._client.publish(topic_path, message)` returns `"pubsub_v1.publisher.futures.Future"` so I'm happy to change, maybe we can see what the other devs think as I don't know if there was a specific reason to why it wasn't set to `Future` to begin with?

Yuyuutsu: Sounds good, ill resolve the issue now and if the devs dont agree, you cna just change it back 👍

Yuyuutsu (approved):

app/publisher/publisher.py
VirajP1002: Devs were happy to keep it as `Future` 👍

PR #1366, opened 2024-05-16
Add more context to list item tick icons

VirajP1002:
Tick icons on lists of items with completed questions for each item (which are not hubs) do not have enough context for screen readers. So it was decided to add an element (`iconVisuallyHiddenText`) that tells the user that an item is complete, i.e. you have answered all the necessary questions.

The screen reader now says "Section Complete" when it sees a tick icon in the correct context.

This PR also included an update to the DS from 70.0.2 --> 70.0.5

### How to review
- Ensure the changes made are suitable
- Launch `test_list_collector_content_page`, progress to where you have added a couple of companies and you are on the summary of those companies, complete the questions on one of the companies. When you have done that one of the companies should now have a tick against it. Turn VoiceOver on (Shortcut on Mac is `CMD+fn+f5`). Once VoiceOver is on ensure you refresh the page in order for the screen reader to read the correct elements. As the VoiceOver is going over the page, once it reads the tick icon, it should say "Completed".

### Checklist
* [ ]


VirajP1002:
In `helpers.js` I updated `checkItemsInList` to ignore the `Section complete` in the received String from the summary page, which works and only a few of the tests have to be tweaked. However, to make it dynamic and to prevent tests being changed, I can change the original `checkItemsInList` to use `toContain` instead of `toEqual`. Let me know your thoughts and any questions you may have. Happy to hop on a call to discuss.

app/jinja_filters.py
```diff
@@ -709,8 +709,11 @@ def map_list_collector_config(
             else None
         )
 
+        iconVisuallyHiddenText = "Section complete" if icon == "check" else None
```
berroar, line 712: Where did the wording "Section complete" come from for this? Feels a little bit confusing to me given you are already within the context of a section, and section is a term that may be referenced elsewhere

app/jinja_filters.py
```diff
@@ -709,8 +709,11 @@ def map_list_collector_config(
...
+
         row_item = {
             "iconType": icon,
+            "iconVisuallyHiddenText": iconVisuallyHiddenText,
```
berroar, line 716: We don't use camel case for python vars, but not sure we need a var for this anyway, since it is only used once?

VirajP1002: Updated the var name 👍 We need this var, just in case render_icon is set to False, we don't want the "section complete" to be present, and if it is set to True we want it to be present, so I set the var up to control this

app/jinja_filters.py
VirajP1002: The wording was just from the ticket itself

app/jinja_filters.py
berroar: But what I mean is it could just be `iconVisuallyHiddenText: "Section complete" if icon else None`? Since icon will always be either check or None?

app/jinja_filters.py
VirajP1002: As discussed, I updated the text to match the DAC report and will mention this in refinement as well as the ambiguity of what is "complete or not"

app/jinja_filters.py
VirajP1002: Updated 👍

tests/app/test_jinja_filters.py
```diff
@@ -1282,5 +1289,59 @@ def test_summary_item_config_with_list_collector_and_one_related_answer():
...
+        },
+        {
+            "item_title": "Joe Bloggs",
+            "list_item_id": "two",
+            "repeating_block": True,
+            "is_complete": False,
+        },
+    ]
+
+    output = map_list_collector_config(list_items, True, True)
```
berroar, line 1308: Minor but i would name the params here to make this clearer

VirajP1002: Updated 👍

tests/functional/spec/features/repeating_blocks/list_collector_repeating_blocks.spec.js
```diff
@@ -96,7 +96,7 @@ describe("List Collector Repeating Blocks", () => {
       await $(AnyOtherCompaniesOrBranchesPage.listEditLink(2)).click();
       await $(EditCompanyPage.companyOrBranchName()).setValue("Government");
       await click(EditCompanyPage.submit());
-      await expect(await $(AnyOtherCompaniesOrBranchesPage.listLabel(2)).getText()).toBe("Government");
+      await checkItemsInList(["ONS", "GOV", "Government"], AnyOtherCompaniesOrBranchesPage.listLabel);
```
berroar, line 99: How come this was updated? A bit confusing now as I would have expected Gov to be have edited to be Government but it was MOD 😅

berroar (approved):

liamtoozer (approved):
Looks great to me 👍

tests/functional/spec/features/repeating_blocks/list_collector_repeating_blocks.spec.js
VirajP1002: Updated 👍

PR #1393, opened 2024-06-12
Fix failing token decryption unit test

petechd:
This fixes the intermittent issue we get in one of our unit tests. Since we use the same app context for multiple tests that decrypt token, shared across the unit test class as a property, the solution is to tear down the app context between tests. 

### How to review
Run tests locally and in actions as part of build. To speed it up remove all irrelevant jobs from pull_request.yaml and only install python dependencies and unit tests.

### Checklist


VirajP1002:
I ran actions in a draft PR and it failed once out of 15 times, not sure if there's a further underlying issue? https://github.com/ONSdigital/eq-questionnaire-runner/actions/runs/9503668920/job/26234364306?pr=1394

petechd:
> I ran actions in a draft PR and it failed once out of 15 times, not sure if there's a further underlying issue? https://github.com/ONSdigital/eq-questionnaire-runner/actions/runs/9503668920/job/26234364306?pr=1394

Yeah I saw that and also had the same error again locally and in actions, so have to tackle it differently since setting up the app and tearing it down doesn't fix it.

petechd:
The unit test job in actions passed 15 consecutive times now after we've added the new fix.

liamtoozer (approved):

berroar (approved):

PR #1396, opened 2024-06-14
Ignore babel minor version dependabot bumps for now

liamtoozer:
Following a recent PR to pin the babel version to `v2.14.0`, we should temporarily prevent dependabot bumping the minor version to `v2.15.0` until the problem has been resolved because this is causing the build to fail.

### How to review
Fork this repo and trigger a dependabot PR from this branch - the babel minor version increment should be excluded and the build should pass

### Checklist

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -59,9 +59,10 @@ updates:
           - "black"
           - "djlint"
     ignore:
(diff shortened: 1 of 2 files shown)


rmccar (approved):

petechd (approved):
Tested in a fork, works like a charm now!

PR #1402, opened 2024-06-18
Add correct calculated summaries title placeholder rendering

petechd:
To fix the placeholder rendering in Calculated Summary titles we pass the rendered block as a parameter into Summary Context now, so the blocks from previous sections are also considered. Previously used cached rendered block method had to be removed.

### How to review
Calculated Summary title in a relevant schema was slightly modified to showcase the new correct placeholder rendering. Worth checking the schema and the new assertions in existing integration test that fails without the fix.

### Checklist


berroar (approved):
Fixes the bug, LGTM 👍

liamtoozer (approved):
Looks good to me, I couldn't recreate the issue 👍

PR #1421, opened 2024-07-02
Contain visible labels within accessible names

VirajP1002:
The `Start section`, `Continue with section` and `View answers` links in hubs were not fully accessible to voice activation users because visible labels of these links are not contained within their accessible names, for example ‘Start Household bills section’ doesn't exactly contain it's visible label of `Start section` which could be confusing.

In order to update this, the `aria_labels` within `SECTION_CONTENT_STATES` had to be updated to change the order of which the label/section name is placed. Previously, it was ‘Start {section_name} section’ and the new version is now `Start section: {section_name}`, the same goes for the other links.

Tests were updated accordingly, a missing unit test for an `IN_PROGRESS` section was added and functional tests included a check for the html to ensure the visually hidden text is as expected.

### How to review
- Ensure the changes made are suitable
- Check the changes to the functional tests and if more changes are required
- Check all tests (functional & unit) pass
- Open any schema with a hub and see if the voiceover works as intended. A good schema to test all variants of the updated visually hidden text is `test_grand_calculated_summary_cross_section_dependencies`. Answer one question of the `Household bills` section, then return to the hub, then complete the `Grand Calculated Summary` section resulting in a hub with all 3 link variants. Use the voiceover (Shortcut command: cmd+fn+f5) and make sure the text is read out correctly, i.e. `Start section: Household bills`. Feel free to test other schemas with a hub.

### Checklist


tests/functional/spec/features/hub_and_spoke/hub_and_spoke.spec.js
```diff
@@ -38,6 +38,9 @@ describe("Feature: Hub and Spoke", () => {
       await expect(await $(HubPage.summaryRowState("employment-section")).getText()).toBe("Not started");
       await expect(await $(HubPage.summaryRowState("accommodation-section")).getText()).toBe("Not started");
       await expect(await $(HubPage.summaryRowState("household-section")).getText()).toBe("Not started");
+      await expect(await $(HubPage.summaryRowLink("employment-section")).getHTML()).toContain("Start section: Employment");
+      await expect(await $(HubPage.summaryRowLink("accommodation-section")).getHTML()).toContain("Start section: Accommodation");
+      await expect(await $(HubPage.summaryRowLink("household-section")).getHTML()).toContain("Start section: Household residents");
```
petechd, line 43: The changes look good, except there are some situations (like the one above) when these checks shouldn't be part of the test or test description should change. For example here we assess the state of a section. Also, we could possibly narrow it down and use different method instead of `.getHTML`. Can we use anything more specific from the list [here](https://webdriver.io/docs/api/element/)?

petechd: For example `When a user first views the Hub, The Hub should be in a continue state` checks if the hub is in continue state but also some other things.

tests/functional/spec/features/hub_and_spoke/hub_and_spoke.spec.js
```diff
@@ -163,6 +171,7 @@ describe("Feature: Hub and Spoke", () => {
       await click(EmploymentStatusBlockPage.submit());
 
       await expect(await $(HubPage.summaryRowState("employment-section")).getText()).toBe("Completed");
+      await expect(await $(HubPage.summaryRowLink("employment-section")).getHTML()).toContain("View answers: Employment");
```
petechd, line 174: Another example here, the check is completely unrelated.

petechd (approved):

tests/functional/spec/features/hub_and_spoke/hub_and_spoke.spec.js
```diff
@@ -40,6 +40,12 @@ describe("Feature: Hub and Spoke", () => {
       await expect(await $(HubPage.summaryRowState("household-section")).getText()).toBe("Not started");
     });
 
+    it("When a user has a screen reader, The visually hidden text should read the name and state of each section in the hub", async () => {
```
berroar, line 43: I would consider updating this as to me it implies there is some specific logic in place for screen readers, but this is the case for all users.

VirajP1002: Rephrased to "When a user utilises a screen reader, The visually hidden text read aloud should be the state and name of each section in the hub"

berroar (approved):

PR #1430, opened 2024-07-12
Investigate flake8-quotes

VirajP1002:
In order to keep our libraries up to date, we have to search for alternative to `pylint-quotes` to allow us to unpin Pylint and upgrade to the latest version. Updating Pylint with `pylint-quotes` results in an error so `flake8-quotes` was proposed to be a replacement to allow us to update Pylint without any issues.

Changes include removing `pylint-quotes` and any references, installing `flake8-quotes`, upgrading Pylint to version `v3.2.5` and disabling `cyclic-import` errors.

### How to review
- Ensure changes made are appropriate
- Ensure linting works as previously
- Following the investigation where there is some inconsistency with local and GHA linting (as mentioned in the [spike doc](https://confluence.ons.gov.uk/display/SDC/Alternative+tool+to+pylint-quotes)) discuss if it's worth us moving to `flake8-quotes`

### Checklist


.pylintrc
```diff
@@ -335,10 +335,6 @@ single-line-class-stmt=no
 # else.
 single-line-if-stmt=no
 
-string-quote=double-avoid-escape
```
petechd, line 338: We would probably want to preserve this config for flake8 quotes, so I've checked the docs and found something like this:
Screenshot 2024-07-16 at 08 31 50

petechd: So our conventions are documented in this config and they are intentional.

VirajP1002: Updated 👍

petechd (approved):

Yuyuutsu (commented):
Changes look good to me, just one question:
do we have a card ready to fix the cyclic imports?

VirajP1002:
> Changes look good to me, just one question: do we have a card ready to fix the cyclic imports?

I've created a [ticket](https://jira.ons.gov.uk/browse/ECI-1362) and it's in tech analysis 👍

Yuyuutsu (approved):

PR #1438, opened 2024-07-23
Add list update fix for repeating sections

petechd:
Currently we update answer dependencies for answers referenced in calculated summary that is part of repeating sections. These are marked for removal when `capture_dependencies_for_list_change` is executed during list add and list remove actions. We want to avoid adding these kind of dependencies when checking calculated summary in repeating section.

### How to review
Check any BRES schema, check the card for exact scenarios.

### Checklist


berroar (commented):
Will need a test to ensure that this functionality works and we have confidence it won't break in future

schemas/test/en/test_calculated_summary_dependent_questions.json
```diff
@@ -147,6 +149,412 @@
...
+                                "id": "question-number-of-employees-working-at-this-additional-site",
+                                "title": "In June 2019, what was the number of full-time and part-time employees that your business paid from its payroll, for this site?",
+                                "guidance": {
+                                    "contents": [
+                                        {
+                                            "description": "<strong>Include:</strong>"
+                                        },
+                                        {
+                                            "list": [
+                                                "full-time, over 30 hours a week, and part-time, 30 hours or less, employees aged 16 years and over. This includes individuals <strong>working from home</strong> and those temporarily absent but still employed, for example those on parental leave or sick leave",
```
berroar, line 484: Only minor, but I would consider de-Bresing this schema, as a lot of these descriptions are quite wordy but are not relevant to the test.

berroar (approved):
Fixes the bug, LGTM 👍

schemas/test/en/test_calculated_summary_dependent_questions.json
petechd: I removed all guidances and descriptions and shortened questions, renamed group etc. It should be fine now.

Yuyuutsu (approved):
Changes looks good to me and bug is fixed 👍

PR #1441, opened 2024-07-24
Fix integration test generator script and makefile command

petechd:
This adds integration test generator script and makefile changes that we've discussed. Chromium can no longer be downloaded and installed on the latest operating systems for security reasons (it will result in  an error when using the makefile command). We need to use Chrome browser as a default and skip the download and install step in Makefile, since Chrome is based on Chromium anyway, we can use it by setting `channel` parameter. For more info check the docs.

### How to review
Check the changes and the behaviour before (you need one of the newer versions of your operating system).

### Checklist


berroar (approved):

VirajP1002 (approved):
Screenshot 2024-08-01 at 17 25 45

All your changes work well 👍

Just a separate minor point (which may be something considered in the past) is when I ran the script for the schema `test_checkbox`, the test generated was called `test_test_checkbox` and the same goes for the class name with 2 tests, is that something we want to considered avoiding?

PR #1460, opened 2024-08-08
Add custom gtag events

petechd:

### How to review

### Checklist


app/helpers/template_helpers.py
```diff
@@ -130,9 +130,14 @@ def data_layer_context(
             key: value for key in DATA_LAYER_KEYS if (value := cookie_session.get(key))
         }
         context = [*additional_context, schema_context]
+
         if tx_id_context:
             context.append(tx_id_context)
-        return context
+
+        flattened_context: dict[str, str] = {}
+        for d in context:
```
berroar, line 138: Can the context not be created flattened in the first place? Rather than doing this loop?

app/helpers/template_helpers.py
```diff
@@ -130,9 +130,14 @@ def data_layer_context(
             key: value for key in DATA_LAYER_KEYS if (value := cookie_session.get(key))
         }
         context = [*additional_context, schema_context]
```
berroar, line 132: Not added as part of this PR, but does the additional context method even do anything? Seems just to return an empty list?

app/helpers/template_helpers.py
petechd: I've reduced all the extra code and it's flat by default now.

app/helpers/template_helpers.py
petechd: Removed the pointless method.

app/helpers/template_helpers.py
```diff
@@ -119,20 +119,16 @@ def service_links_context(
...
         )
-        additional_context = self._survey_config.get_additional_data_layer_context()
         schema_context = {
             key: value for key in DATA_LAYER_KEYS if (value := cookie_session.get(key))
         }
-        context = [*additional_context, schema_context]
-        if tx_id_context:
-            context.append(tx_id_context)
-        return context
+        return tx_id_context | schema_context
```
liamtoozer, line 131: Just a general Python-y question - is using `|` the preferred way that we merge dicts? I've seen it like `{**tx_id_context, **schema_context}` so just wanted to ask for my understanding.

liamtoozer (approved):
Works great for me 👍  Tested in GA

app/helpers/template_helpers.py
petechd: I think it is the most modern way of merging dicts, introduced in 3.9 so very recently.

liamtoozer: Ah awesome, thanks for the explanation 👍

PR #1487, opened 2024-08-28
Add improvements to dependabot config

petechd:
For this PR I tried to use all the options for dependabot configuration available that are relevant to us, for explanation of each check the doc at https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

Things that were changed in order to improve our config:

- divided python/poetry packages into two separate ecosystems (I read in github PRs for depednabot-core project that support for this was added some time ago) so, from now on, we will only execute dependabot pipeline for prod packages (will trigger on label "python" and requires another PR in pipelines)

- added two new ecosystems to check (npm/node and github actions)

- defined a day of the week when we want to execute the check

- defined the interval, for prod dependencies it's weekly (we check and merge depednabot PRs manually bi-weekly atm), for dev packages monthly (all three ecosystems python-dev, node and github actions). Open to debate but worth remembering there are only three options for the interval: daily, weekly or monthly.

- defined the time of the check and the timezone so we don't execute it at random time and in UTC

- added versioning strategy so for python and npm (which do have version constraints for packages in their relevant files (pyproject.toml and package.json). We will update the version constraint to the next major (eg. from ^1.0.0 to ^2.0.0) when it changes, check explanation here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

- removed ignored major and minor dev packages releases setting for python dev packages, since they are now checked monthly (using the new config) and won't trigger dependabot staging pipeline anymore, we want to mark them for checks as well. Before we ignored major and minor releases which caused dependabot checks to trigger and recreate PRs daily

- added more relevant labels (dependabot staging pipeline will trigger on "python" label now, dev group/ecosystem will have "python-dev" label) (NEW LABELS NEED TO BE CREATED FIRST BEFORE MERGING RUNNER PR)

### How to review
You might have to fork the repo, clone the fork locally, revert some commits and then force push it up to the fork with the new config and check if it works (might have to change the day and the time in the config to your time and day). Remember to enable Dependabot in the forked repo settings.

### Checklist


petechd:
If the new labelling is approved this will require changes in Dependabot pipeline before merging.

petechd:
New labels in Runner repo need to be created before this is merged as well.

.github/dependabot.yaml
```diff
@@ -1,9 +1,56 @@
...
+      interval: "monthly"
+      day: "friday"
+      time: "08:00"
+      timezone: "Europe/London"
+    versioning-strategy: increase-if-necessary
+    labels:
+      - "dependencies"
+      - "node"
+
+  - package-ecosystem: "pip"
```
liamtoozer, line 26: I'm getting the below in my forked repo. Does the `pip` ecosystem need to be unique in the config?
```
Ecosystem 'pip' has overlapping directories.
```

petechd: We can get around this by setting `target-branch: main` for one ecosystem, allegedly, will try it now.

.github/dependabot.yaml
```diff
@@ -1,9 +1,58 @@
...
+
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "friday"
+      time: "08:00"
+      timezone: "Europe/London"
+    target-branch: main
```
liamtoozer, line 49: Is it worth popping in a comment for this line, as it's the workaround for having 2 pip ecosystems?

.github/dependabot.yaml
```diff
@@ -58,39 +122,3 @@ updates:
           - "playwright"
           - "black"
           - "djlint"
-    ignore:
-      # Temporarily pinned to v2.14.0 - problem for translations found in v2.15.0, see: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1384
```
liamtoozer, line 62: We might want to keep this in (hopefully only temporary) as the minor version bump for babel affects the tests?

liamtoozer (commented):
Looks great to me and the schedule for the dependabot makes sense 👍

berroar (commented):
Do we need to call out somewhere (in the readme?) that this needs config exists and needs to be updated when a new package is added. Otherwise there is a risk some could be missed?

petechd:
Will make the changes now, just a reminder that what we decide to label prod dependencies needs to be added to and triggering dependabot pipeline, so changed in the triggering resource in pipeline.

petechd:
> Do we need to call out somewhere (in the readme?) that this needs config exists and needs to be updated when a new package is added. Otherwise there is a risk some could be missed?

I was thinking maybe pyproject.toml and package.json? Just a comment there about dependabot config?

.github/dependabot.yaml
```diff
@@ -59,38 +124,6 @@ updates:
           - "black"
           - "djlint"
     ignore:
-      # Temporarily pinned to v2.14.0 - problem for translations found in v2.15.0, see: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1384
+      # "black" temporarily pinned to v2.14.0 - problem for translations found in v2.15.0, see: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1384
```
liamtoozer, line 127: `"black"` -> `"babel"`

liamtoozer (approved):

.github/dependabot.yaml
petechd: Done, thanks, might be because I am doing four different tasks simultaneously.

berroar (approved):

PR #1490, opened 2024-08-30
Update list collector content page to use new Summary Component

VirajP1002:

### How to review

### Checklist


petechd (approved):

berroar:
Only minor issue I notice is that the spacing is now slightly different on the content page, when compared with what it was previously (using test env for example).

Test on the left vs this branch:
Screenshot 2024-09-17 at 10 40 10

rmccar:
> Only minor issue I notice is that the spacing is now slightly different on the content page, when compared with what it was previously (using test env for example).
> 
> Item 2 is slightly indented under the first row.
> 
> Test on the left vs this branch: Screenshot 2024-09-17 at 10 40 10

Yeah the spacing on the list component is slightly different to the summary one, ill try run this by a designer. I think this is something that can be fixed (if needed) outside of this piece of work though

berroar:
> > Only minor issue I notice is that the spacing is now slightly different on the content page, when compared with what it was previously (using test env for example).
> > Item 2 is slightly indented under the first row.
> > Test on the left vs this branch: Screenshot 2024-09-17 at 10 40 10
> 
> Yeah the spacing on the list component is slightly different to the summary one, ill try run this by a designer. I think this is something that can be fixed (if needed) outside of this piece of work though. If any change is needed then this will be a DS change

Just the acceptance criteria explicitly calls out that it needs to be identical so worth checking

tests/functional/spec/features/supplementary_data/supplementary_data.spec.js
```diff
@@ -370,11 +370,11 @@ describe("Using supplementary data", () => {
   it("Given there is now an additional product, When I resume the Product Details Section, Then I start from the list collector content block and see the new product is incomplete", async () => {
     await $(HubPage.summaryRowLink("section-6")).click();
     await expect(browser).toHaveUrlContaining(ListCollectorProductsPage.pageName);
-    await checkListItemComplete(`dt[data-qa="list-item-1-label"]`);
-    await checkListItemComplete(`dt[data-qa="list-item-2-label"]`);
-    await checkListItemIncomplete(`dt[data-qa="list-item-3-label"]`);
+    await summaryItemComplete(`li[data-qa="list-item-1-label"]`);
+    await summaryItemComplete(`li[data-qa="list-item-2-label"]`);
+    await summaryItemIncomplete(`li[data-qa="list-item-3-label"]`);
```
berroar, line 375: This needs to call the list item method? Hence the failures?

rmccar: Good spot, missed that one

tests/functional/helpers.js
```diff
@@ -6,11 +6,18 @@ export const checkItemsInList = async (itemsExpected, listLabel) => {
   }
 };
 
-export const checkListItemComplete = async (listItemLabel) => {
-  await expect(await $(listItemLabel).$(`.ons-summary__item-title-icon.ons-summary__item-title-icon--check`).isExisting()).toBe(true);
+export const summaryItemComplete = async (summaryItemLabel) => {
```
berroar, line 9: Do we need to duplicate these methods with complete/incomplete, could just pass the bool in? Then we'd only need two?

berroar: Or even one if we passed the element?

rmccar: Done, I did think about that but there would be some logic needed for the selector so I decided to leave it as two as it seemed neater to me. Happy to make the change to one though if other people feel different

petechd: You trade readability for universality so there are pros and cons for both approaches

rmccar:
> > > Only minor issue I notice is that the spacing is now slightly different on the content page, when compared with what it was previously (using test env for example).
> > > Item 2 is slightly indented under the first row.
> > > Test on the left vs this branch: Screenshot 2024-09-17 at 10 40 10
> > 
> > 
> > Yeah the spacing on the list component is slightly different to the summary one, ill try run this by a designer. I think this is something that can be fixed (if needed) outside of this piece of work though. If any change is needed then this will be a DS change
> 
> Just the acceptance criteria explicitly calls out that it needs to be visually identical so worth checking

Raised a PR on the DS to fix this

tests/functional/helpers.js
rmccar:
```
export const itemComplete = async (itemLabel, listType, status) => {
  let selector;
  if (listType === "Summary") {
    selector = `.ons-summary__item-title-icon.ons-summary__item-title-icon--check`;
  } else if (listType === "List") {
    selector = `.ons-list__prefix.ons-list__prefix--icon-check`;
  }
  await expect(await $(itemLabel).$(selector).isExisting()).toBe(status);
};
```
This is the sort of thing we would be looking at just to give you an idea

app/jinja_filters.py
```diff
@@ -648,6 +648,31 @@ def map_summary_item_config_processor() -> dict[str, Callable]:
...
+
+    for index, value in enumerate(list_values, 1):
+        item: dict = {"text": value["item_title"]}
+
+        if value["is_complete"]:
+            item["iconType"] = "check"
+
+        item["attributes"] = {
+            "data-qa": f"list-item-{index}-label",
+            "data-list-item-id": value["list_item_id"],
```
berroar, line 668: Are we doing anything with `data-list-item-id`? Is it needed?

tests/functional/helpers.js
petechd: Why don't we just pass in the actual element instead of listType? So you don't need the "ifs"?

app/jinja_filters.py
petechd: Used in tests (integration).

berroar: But not for List Collector Content pages?

berroar: Don't think there is an integration test that calls this new method, so probably a sign that one needs adding

tests/functional/helpers.js
rmccar: Because the selectors are quite long I felt like passing them in every time wasn't the best for readability. I could do that by setting them to variables at the top of the test files I spose but I feel like doing all this to just remove one helper statement is a bit overkill

petechd: I mean, the function definition describes well what the parameters are, if we get rid of the "if" logic and not set the hardcoded vars we end up with only 1 or 2 lines of code for this reusable function. I prefer that, after some consideration.

rmccar: Yeah I understand that, my thoughts were just that it would be simple in this file but you'd be just passing the problem down to the test file. It would mean that we would need to pass in `.ons-summary__item-title-icon.ons-summary__item-title-icon--check` or `.ons-list__prefix.ons-list__prefix--icon-check`, the selector we already pass in and the status every time we used this function which seems less readable to me. We could set up consts in the test file to help with this but all in all it comes down to whether we call a different function (list vs summary) or we pass in an extra param (list vs summary) which would be a long one, in that way its much of a muchness when you are calling it but is less readable. It just feels like to me we are trying to make this function too generic

rmccar: Might be easier to chat about it tomorrow in tech session

app/jinja_filters.py
petechd: Ok, seems like we have a gap in our testing, probably because map_list_config is referenced in the jinja processor and used as a blueprint we didn't get the lower coverage (it goes undetected).

tests/functional/helpers.js
petechd: If we can identify same pattern in different functions and use one instead of two or four that do essentially the same task that is a good thing. But if it is less readable to people keep it separate, down to the devs.

app/jinja_filters.py
berroar: Still not sure `data-list-item-id` is needed for this as looks like some other list collector features use it as part of their integration tests, but we're not really doing anything with it here. I'd just remove it.

tests/app/test_jinja_filters.py
```diff
@@ -1339,5 +1340,50 @@ def test_map_list_collector_config_render_icon_set():
     assert output == expected
 
 
+def test_map_list_config():
+    list_values = [
```
berroar, line 1344: Minor but this is the wrong type at the moment, no need for the nested list, supposed to be a list of dicts.

tests/functional/helpers.js
berroar: I'd usually say more generic is better, but agree that it needs to be readable. The reason I commented here in the first place as the fact we had multiple methods meant we were actually using the wrong one initially (when we thought the test might be flaky), and that's always a risk when there's a bunch of methods doing pretty much the same thing, so more maintainable and clearer sometimes when there is only one.

rmccar: Yeah agreed. Happy to have the conversation and explain a bit more in tech session, I think that will make this discussion easier

berroar (approved):

petechd (approved):

PR #1500, opened 2024-09-06
Resolve Cyclic Imports 

Yuyuutsu:
This PR refactors VSR and RuleEvaluator to resolve most of the Cyclic imports found after upgrading Pylint to version `v3.2.5`.

### How to review
Read through the changes and check if runner works as expected.
Check if tests pass locally

### Checklist


app/forms/questionnaire_form.py
```diff
@@ -310,20 +310,25 @@ def _get_period_range_for_single_date(
         date_to: Mapping[str, dict],
     ) -> timedelta:
         list_item_id = self.location.list_item_id if self.location else None
-        value_source_resolver = ValueSourceResolver(
+        rule_evaluator = RuleEvaluator(
```
berroar, line 313: I'm not sure this makes sense at the moment, the rule evaluator now has a value source resolver but you're still passing both to the field handler?

app/questionnaire/resolver.py
```diff
@@ -0,0 +1,35 @@
...
+ResolverTypes = None | str | int | Decimal | list | dict
+ResolveTypes = Markup | list[Markup]
+IntOrDecimal = int | Decimal
+ResolvedAnswerList = list[ResolveTypes | None]
+
+ResolveEscapedTypes: TypeAlias = Markup | list[Markup]
+
+
+@dataclass
+class Resolver(ABC):
```
berroar, line 21: Do we need this abstract class? I don't think this makes sense since there is only one instance of a "Resolver" class? So all of the stuff from ValueSourceResolver would be in here if we did this/nothing would be overridden since they are be the same?

Yuyuutsu: It was implemented as a pre measure. I have updated VSR and RuleEvaluator, where RuleEvaluator uses type annotations for the type hinting.

app/forms/questionnaire_form.py
Yuyuutsu: Removed VSR 👍

app/forms/questionnaire_form.py
```diff
@@ -480,32 +480,47 @@ def get_answer_fields(
     if block_ids_by_section:
         block_ids = get_flattened_mapping_values(block_ids_by_section)
 
-    def _get_value_source_resolver(list_item: str | None = None) -> ValueSourceResolver:
+    def _get_value_source_resolver(
```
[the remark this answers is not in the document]
berroar: Probably worth having a chat about this approach with the team anyway.

petechd, line 483: If this private method does not use vars from the higher scope of the class implicitly (is only passing parameters explicitly) , there is no reason why to nest it inside another method of this class. Should be separated out from the parent method.

app/questionnaire/rules/rule_evaluator.py
```diff
@@ -50,6 +45,7 @@ def __post_init__(self) -> None:
         self.operations = Operations(
             language=self.language, schema=self.schema, renderer=renderer
         )
+        self.value_source_resolver = self.value_source_resolver
```
petechd, line 48: Does this statement make sense?

app/questionnaire/value_source_resolver.py
```diff
@@ -272,6 +272,14 @@ def _resolve_summary_with_calculation(
             data_stores=self.data_stores,
             location=self.location,
             routing_path_block_ids=self.routing_path_block_ids,
+            value_source_resolver=ValueSourceResolver(
```
petechd, line 275: We definitely don't want to instantiate another `ValueSourceResolver` inside the `ValueSourceResolver` class instance.

petechd: I agree with @berroar this probably needs some more thought and some kind of structured approach/plan/proposal.

app/questionnaire/rules/rule_evaluator.py
Yuyuutsu: Technically,  yes 😂 . Removed 👍

app/questionnaire/value_source_resolver.py
Yuyuutsu: Yeah this felt akward to do when refactoring, Will have a look

app/forms/questionnaire_form.py
Yuyuutsu: I think it still uses the higher scope of the function. If not I think the function would not be needed

app/questionnaire/path_finder.py
```diff
@@ -181,10 +182,18 @@ def _evaluate_routing_rules(
         )
 
         when_rule_evaluator = RuleEvaluator(
-            self.schema,
-            self.data_stores,
+            schema=self.schema,
+            data_stores=self.data_stores,
             location=this_location,
             routing_path_block_ids=block_ids_for_dependencies,
+            value_source_resolver=ValueSourceResolver(
```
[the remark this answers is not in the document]
Yuyuutsu: Will bring this up in Tech Session

liamtoozer (commented):
Great work 👍  I agree though that we might want to take a bit more time and look at some of the alternatives like breaking down some of the bigger classes into smaller ones to get around the cyclic imports. Appreciate this is a lot more complex though and might not be as straightforward as that. Maybe a spike ticket to look at some possible solutions, with this being one of them?

Yuyuutsu:
Yeah I agree, as this approach still has its issues and would be better if we can find another approach. - Closing PR

Yuyuutsu:
NEW Investigation card: https://jira.ons.gov.uk/browse/ECI-1498

PR #1507, opened 2024-09-10
Fix pub/sub unit test

liamtoozer:
A recent PR was raised to remove a failing Pub/Sub unit test (https://github.com/ONSdigital/eq-questionnaire-runner/pull/1504), which was asserting an API response exception from GCP directly. This was causing failures because GCP had changed their error message text.

Instead, we can just assert that the exception is raised in our function, and patch the pub/sub exception so that we don't query GCP directly

### How to review
Check that the test passes and that it makes sense

### Checklist


Yuyuutsu (approved):

tests/app/publisher/test_publisher.py
```diff
@@ -24,3 +30,19 @@ def test_publish(publisher, mocker):
...
+    with patch(
+        "app.publisher.publisher.PubSubPublisher._publish", return_value=mock_future
+    ):
+        with pytest.raises(PublicationFailed) as ex:
+            publisher.publish(
+                "test-topic-id",
+                b"test-message",
+                fulfilment_request_transaction_id=str(uuid4()),
+            )
+        assert "Test exception during publish" in str(ex.value)
```
petechd, line 48: Not sure now if we should do all this, we test the side effect really. Maybe the best approach, if we don't want to check the exception message, would be just check/assert if it raises exception? So like in here (for example):
```python
    with pytest.raises(KeyError):
        get_field_handler(
            answer_schema={"type": invalid_field_type},
            value_source_resolver=value_source_resolver,
            rule_evaluator=rule_evaluator,
            error_messages=error_messages,
        )
```
so just:
```python
with pytest.raises(PublicationFailed):
            publisher.publish(
                "test-topic-id",
                b"test-message",
                fulfilment_request_transaction_id=str(uuid4()),
            )
```

petechd: I mean we still raise the same exception, and that's what the test is about, that the function call raises a certain exception.

liamtoozer: I think I used the patch in the test because the `_publish()` was hitting the GCP API, when we only care that it returns an exception and that we handle it. If I remove the patch, it will mean that the unit test will always try to hit the GCP API, which might not be what we want?

Might be easier to show what I mean as I'm not great at explaining this 😆

liamtoozer: Following an external discussion - since the test was originally looking for a `403` response from GCP intentionally, we decided to keep the test as it was, but tweak the string returned so that it won't be as brittle

Yuyuutsu (approved):
Relook -  changes look good, and works as expected

petechd (approved):
Not much we can do here, except the changes you've made.

tests/app/publisher/test_publisher.py
liamtoozer: Following another discussion - we really shouldn't be having an external dependency in a unit test, so we'll patch it for now and just ensure that we test our custom exception is raised

petechd (approved):

berroar (approved):

PR #1535, opened 2024-10-15
Update npm dependencies, deprecated test functions, and add major version dependabot ignore

liamtoozer:
Following an investigation for a dependabot PR, the dependencies were bumped but in doing so caused multiple errors stating that usage of `toHaveUrlContaining()` used in our functional tests were now deprecated, so we'll need to move to whatever the new ways are. The main changes for this PR are as follows:

* Due to eslint v9 being incompatible with some of our other npm packages, I've added an ignore to the dependabot config until we investigate some of the alternatives
* I've also added an ignore to the next major version of `@wdio/local-runner`. Bumping to `v9` was causing assertion failures with `getHTML()` returning a string with indentation and new lines 
* Updated `wdio.conf.js` export keyword as was getting errors after the version bump
* Updated deprecated functional test functions
* Created new helper function in `helper.js` to handle URL string match assertions centrally
* Created new helper function in `hub_and_spoke.spec.js` to handle strict hub and spoke URL matching

### How to review
Lots of file changes, so a bit tricky to review. Might be worth a quick check of the file changes to see if you spot anything incorrect, followed by checking GitHub Actions to see if there is anything that crops up.

### Checklist


package.json
```diff
@@ -23,33 +23,33 @@
...
-        "@babel/register": "^7.22.5",
-        "@babel/runtime": "^7.23.2",
-        "@wdio/cli": "^8.14.4",
+        "@babel/core": "^7.25.8",
+        "@babel/plugin-transform-runtime": "^7.25.7",
+        "@babel/preset-env": "^7.25.8",
+        "@babel/register": "^7.25.7",
+        "@babel/runtime": "^7.25.7",
+        "@wdio/cli": "^9.2.1",
         "@wdio/local-runner": "^8.14.3",
```
liamtoozer, line 32: Intentionally excluded from major version bump. See [dependabot config comment](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1535/files#diff-ee0c08340ace9ac3ce3c78e2377968bca2376b8bfd9a7a48d96515f5490a902cR47)

package.json
```diff
@@ -23,33 +23,33 @@
...
-        "@wdio/mocha-framework": "^8.14.0",
-        "@wdio/spec-reporter": "^8.14.0",
-        "eslint": "^8.46.0",
+        "@wdio/mocha-framework": "^9.1.3",
+        "@wdio/spec-reporter": "^9.1.3",
+        "eslint": "^v8.57.1",
         "eslint-cli": "^1.1.1",
```
liamtoozer, line 36: All `eslint*` packages intentionally excluded from major version bump. See [dependabot config comment](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1535/files#diff-ee0c08340ace9ac3ce3c78e2377968bca2376b8bfd9a7a48d96515f5490a902cR42)

tests/functional/spec/journeys/hub_and_spoke/hub_and_spoke.spec.js
```diff
@@ -159,7 +159,7 @@ describe("Feature: Hub and Spoke", () => {
 
     it("When the user returns to the Hub and continues, Then they should progress to the next section", async () => {
       await click(EmploymentTypeBlockPage.submit());
-      await expect(browser).toHaveUrlContaining(HubPage.url());
+      await expect(browser).toHaveUrl(expect.stringContaining(HubPage.url()));
```
berroar, line 162: Strange that just below this we use a different method of asserting the URL 🤔 e.g.
const expectedUrl = await browser.getUrl();
await expect(expectedUrl).toContain(ProxyPage.url());

berroar: Also, I though this is an issue that predates this PR, I think we need to be more strict with the check for the hub page, because `/questionnaire` is in all the page titles

liamtoozer: Ah yeah that is done a bit differently 🤔  According the the official docs, using `.toHaveUrl()` includes waiting and retry (and seems to have more flexibility with params), so I would think that would be the preferable function to use? 

I think I'll update the other occurrences where we're using `await browser.getUrl()` while we're here in this PR

tests/functional/spec/checkbox.spec.js
```diff
@@ -57,7 +57,7 @@ describe('Checkbox with "other" option', () => {
     // When
     await $(MandatoryCheckboxPage.otherDetail()).setValue("Other Text");
     await click(MandatoryCheckboxPage.submit());
-    await expect(browser).toHaveUrlContaining(NonMandatoryCheckboxPage.pageName);
+    await expect(browser).toHaveUrl(expect.stringContaining(NonMandatoryCheckboxPage.pageName));
```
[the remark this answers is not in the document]
liamtoozer: Ooh nice, I completely forgot about `helpers.js`. Will update now, thanks 👍

tests/functional/spec/journeys/hub_and_spoke/hub_and_spoke.spec.js
liamtoozer: > Also, though this is an issue that predates this PR, I think we need to be more strict with the check for the hub page, because /questionnaire is in all the page titles

As discussed externally, a new helper function for strict checking of the [URL path was created for this](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1535/commits/6e413d1ac5711bf20e5a60b562f76a0420244605#diff-5624047e3f252f3afd475bc7b98f4607ae6f53887752347aca54d19ea7640b12R389-R393). Thanks for your help 👍

liamtoozer: > I think I'll update the other occurrences where we're using await browser.getUrl() while we're here in this PR

Now updated with in [this commit](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1535/commits/e1ae742eaffc3e6aecb24e8b5d9c85e92cdc2363). I've also updated the PR description to cover these changes. Thanks 👍

berroar (approved):
Will need tickets to investigate unpinning those dependencies, but otherwise LGTM 👍

Yuyuutsu (approved):
LGTM !

PR #1538, opened 2024-10-17
Update sds schema validation

VirajP1002:
To align with SDS & Author we need to update the validation rules for supplementary data schema versions within Runner in order to support future automation of the releases of new SDS schema versions.

`supplementary_data_parser` previously used a list of known versions but this will become unmanageable with several versions. Instead, we compare the `sds_schema_version` set in a questionnaire schema to the  `schema_version` found in the supplementary data payload. We pass `sds_schema_version` from `session.py` to `supplementary_data_parser` for it to be validated. The old logic has been removed and tests have been updated/added to show this change.

### How to review
- Ensure changes make sense
- Check unit, integration & functional tests pass
- Ensure schema validation passes (using the validator [PR](https://github.com/ONSdigital/eq-questionnaire-validator/pull/256))

### Checklist


app/routes/session.py
```diff
@@ -177,6 +177,7 @@ def _set_questionnaire_supplementary_data(
         dataset_id=new_sds_dataset_id,
         identifier=identifier,  # type: ignore
         survey_id=metadata["survey_id"],  # type: ignore
+        sds_schema_version=schema.json.get("sds_schema_version", None),
```
liamtoozer, line 180: As `.get()` implicitly returns `None` if the key isn't found, might not need to override the output here?

app/utilities/supplementary_data_parser.py
```diff
@@ -67,25 +58,33 @@ class SupplementaryDataMetadataSchema(Schema, StripWhitespaceMixin):
 
     @validates_schema()
     def validate_dataset_and_survey_id(  # pylint: disable=unused-argument
```
petechd, line 60: The function name should probably change since you also check "sds_schema_version" now.

app/utilities/supplementary_data_parser.py
```diff
@@ -67,25 +58,33 @@ class SupplementaryDataMetadataSchema(Schema, StripWhitespaceMixin):
...
                     "Supplementary data did not return the specified Dataset ID"
                 )
 
-            if data["survey_id"] != self.context["survey_id"]:
+            if payload["survey_id"] != self.context["survey_id"]:
                 raise ValidationError(
                     "Supplementary data did not return the specified Survey ID"
                 )
 
+            if (self.context["sds_schema_version"]) and (
```
petechd, line 74: The first part doesn't need to be in brackets as you don't do any operation there, only second part?

app/utilities/supplementary_data_parser.py
```diff
@@ -67,25 +58,33 @@ class SupplementaryDataMetadataSchema(Schema, StripWhitespaceMixin):
...
+                payload["data"]["schema_version"] != self.context["sds_schema_version"]
```
petechd, line 75: This could potentially lead to `KeyError` if the "schema_version" is not there, since we don't check payload[data] like we do check payload (`if payload:`). Is "schema_version" guaranteed to be in the payload[data]?

tests/integration/session/test_login.py
```diff
@@ -365,6 +366,25 @@ def test_login_token_with_invalid_cir_instrument_id_results_in_500(self):
...
+    )
+    def test_login_with_sds_schema_version_valid(
+        self, mock_get, mock_validate, mock_set
+    ):
+
+        self.launchSupplementaryDataSurvey(
+            schema_name="test_supplementary_data_with_sds_schema_version",
+            sds_dataset_id="203b2f9d-c500-8175-98db-86ffcfdccfa3",
+        )
+        # Then
```
petechd, line 382: No "Given/When/Then" structure in this test, only "Then".

tests/integration/session/test_login.py
```diff
@@ -365,6 +366,25 @@ def test_login_token_with_invalid_cir_instrument_id_results_in_500(self):
         # Then
         self.assertException()
 
+    @patch("app.routes.session.get_supplementary_data_v1")
+    @patch("app.routes.session._validate_supplementary_data_lists")
+    @patch(
+        "app.questionnaire.questionnaire_store_updater.QuestionnaireStoreUpdaterBase.set_supplementary_data",
+    )
+    def test_login_with_sds_schema_version_valid(
```
petechd, line 374: You probably need to move your testing to `test_supplementary_data_parser.py` where the class and methods you changed are tested.

schemas/test/en/test_supplementary_data_with_sds_schema_version.json
```diff
@@ -0,0 +1,192 @@
+{
```
liamtoozer, line 1: There's probably a good reason but just wanted to check - is it worth adding `sds_schema_version` to the existing `test_supplementary_data` schema instead?

berroar: I believe it was because it meant 2 out of the 3 test sds datasets wouldn't open with the schema, as one was set to `v2` and the other `v3`, which would mean it wouldn't match with the `v1` set in the schema.

liamtoozer: Ah cool that makes sense, thanks 👍

tests/integration/session/test_login.py
VirajP1002: The purpose of this test was just to check if the schema opens, I initially did a functional test, but then we decided it would be better to add it as an integration test. So we chose to put it in test_login as these just check if schemas can open?

app/utilities/supplementary_data_parser.py
VirajP1002: Instead, I moved the sds_schema_version validation in it's own function underneath, just because it makes the name look smaller in addition to the method name making sense 😅 and validation still passes.

tests/integration/session/test_login.py
VirajP1002: Updated

app/utilities/supplementary_data_parser.py
VirajP1002: Updated

tests/integration/session/test_login.py
berroar: I'm not sure that this is testing anything at the moment though cause if you change the sds_schema_version in your schema the suite still passes

app/utilities/supplementary_data_parser.py
petechd: Good, that new function will require some unit testing (if it's not already tested).

schemas/test/en/test_supplementary_data_with_sds_schema_version.json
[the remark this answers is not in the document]
VirajP1002: Add discussed we decided to remove the schema and the integration test

tests/integration/session/test_login.py
VirajP1002: As discussed we agreed that the unit test did the same as the integration test and so I removed it

app/utilities/supplementary_data_parser.py
liamtoozer: V minor - No worries if this has already been discussed, so feel free to ignore me 😆  but we now have 2 similar functions that validate things in the payload as follows: 

* existing `validate_dataset_and_survey_id()` which validates 2 things in the payload
* newly added `validate_sds_schema_version()` which validates 1 thing in the payload
 
would it be worth splitting `validate_dataset_and_survey_id` into separate validation functions, if we're making the validation modular? 

Personal preference, but I feel like it might make the most sense to keep all validation logic for this as one simple function (like it was originally in the PR I think?) and just rename it to something like `validate_payload()`?

app/services/supplementary_data.py
```diff
@@ -121,14 +126,19 @@ def decrypt_supplementary_data(
 
 
 def validate_supplementary_data(
-    supplementary_data: Mapping, dataset_id: str, identifier: str, survey_id: str
+    supplementary_data: Mapping,
+    dataset_id: str,
+    identifier: str,
+    survey_id: str,
+    sds_schema_version: str | None = None,
 ) -> dict:
```
petechd, line 134: Little detail here but for the return value type hint we recommend in our guide to at least give types of the dict keys and values.

tests/app/parser/test_supplementary_data_parser.py
```diff
@@ -54,6 +54,34 @@ def test_invalid_supplementary_data_payload_raises_error():
...
+            sds_schema_version="v6",
+        )
+
+    assert (
+        str(error.value)
+        == "{'_schema': ['The Supplementary Dataset Schema Version does not match the version set in the Questionnaire Schema']}"
+    )
+
+
+def test_valid_supplementary_dataset_version():
```
petechd, line 73: You can create some more readable code by just writing a test where valid dataset version does not raise an error by using some helper function like this one:
```python
from contextlib import contextmanager

@contextmanager
def not_raises(exception):
  try:
    yield
  except exception:
    raise pytest.fail("DID RAISE {0}".format(exception))
```

petechd: ...and then use it as `with not_raises(ValidationError):`

app/services/supplementary_data.py
VirajP1002: Updated these 👍

tests/app/parser/test_supplementary_data_parser.py
VirajP1002: Added this in 👍

petechd (approved):

app/utilities/supplementary_data_parser.py
VirajP1002: Updated this, I've placed everything into one function called `validate_payload()` as that seems like a better alternative

berroar (approved):

app/utilities/supplementary_data_parser.py
```diff
@@ -66,27 +57,37 @@ class SupplementaryDataMetadataSchema(Schema, StripWhitespaceMixin):
...
                     "Supplementary data did not return the specified Dataset ID"
                 )
 
-            if data["survey_id"] != self.context["survey_id"]:
+            if payload["survey_id"] != self.context["survey_id"]:
                 raise ValidationError(
                     "Supplementary data did not return the specified Survey ID"
                 )
 
+            if payload:
```
liamtoozer, line 74: Minor - As we're checking a bit further up, this extra `if payload:` check won't be needed

liamtoozer (commented):
~~I've just taken a look at the SDS models and it looks like the `schema_version` was recently removed: https://github.com/ONSdigital/sds/commit/92704136f9b48a29ec9b6c3d47fcb4052b462948.~~

~~I think this would mean we would not be able to validate using the returned `schema_version` in the payload any longer.~~

Not an issue for us as we're using the `schema_version` from the `data` which should still be okay as that matches our SDS schema definitions 👍

liamtoozer (approved):

app/utilities/supplementary_data_parser.py
VirajP1002: Removed 👍

PR #1554, opened 2024-11-22
Fix ruff linting issue UP007: Using | for type annotations

berroar:
Fixes all instances of the ruff linting issue UP007.  This is where we have been using `Optional[A]` and `Union[A, B]` where we should be using `A | None` or `A | B`.

The only exception is where we have type aliases using types in "". We can't use | so these lines have been ignored.

### How to review
Check all instances are fixed and that there are no linting errors.

### Checklist


app/data_models/answer.py
```diff
@@ -69,13 +74,13 @@ def escape_answer_value(value: str) -> Markup: ...  # pragma: no cover
 
 @overload
 def escape_answer_value(
-    value: Union[None, int, Decimal]
-) -> Union[None, int, Decimal]: ...  # pragma: no cover
+    value: None | int | Decimal,
```
petechd, line 77: Comma not needed?

berroar: This gets flagged on the lint check and added when reformatted by black 🤔

petechd: I guess it's not such a big problem, doesn't change the syntax as it's only a type hint.

petechd (approved):
So many outdated type hints! I swear we had a ticket to fix those.

VirajP1002 (approved):

PR #1570, opened 2025-01-07
Remove Indivuidual Response

Yuyuutsu:
Remove Individual Response (IR) Functionality from Runner
New Test was added for LogPublisher and some tests were removed too.

### How to review
Check if Runner works as expected. 
Tests Pass and changes make sense.  
Check if all references to IR have been removed.

### Checklist


tests/app/publisher/test_publisher.py
```diff
@@ -50,3 +51,17 @@ def test_resolving_message_raises_exception_on_error(publisher):
                 b"test-message",
                 fulfilment_request_transaction_id=str(uuid4()),
             )
+
+
+def test_log_publisher_publish():
```
berroar, line 56: How come we needed this new test?

.development.env
```diff
@@ -15,7 +15,6 @@ EQ_REDIS_HOST=localhost
 EQ_REDIS_PORT=6379
 EQ_SECRETS_FILE=dev-secrets.yml
 EQ_KEYS_FILE=dev-keys.yml
-EQ_INDIVIDUAL_RESPONSE_POSTAL_DEADLINE=2021-04-28T14:00:00+00:00
```
berroar, line 18: Removed the env vars here but they are still expected in the deploy_app.yaml file. Will also need to update the pipelines and terraform and test deploying runner with the vars removed.

berroar (changes requested):

.development.env
Yuyuutsu: Updated and Tested. New PRs have been added to the card:

- [Pipelines PR](https://github.com/ONSdigital/eq-pipelines/pull/658)
- [Terraform GCP Runner PR](https://github.com/ONSdigital/eq-terraform-gcp-runner/pull/41)

tests/app/publisher/test_publisher.py
Yuyuutsu: LogPublisher was covered in`test_individual_response.py`

app/data_models/progress_store.py
```diff
@@ -77,12 +77,11 @@ def is_routing_backwards(self) -> bool:
     def is_section_complete(self, section_key: SectionKey) -> bool:
         """
         Return True if the CompletionStatus of the Section or List Item specified by the given section_id and
-         list_item_id is COMPLETED or INDIVIDUAL_RESPONSE_REQUESTED, else False.
+         list_item_id is COMPLETED, else False.
         """
         return section_key in self.section_keys(
```
petechd, line 82: Since there are no multiple statuses now, the whole logic can be simplified here (so you don't look for an item(status) in a sequence of statuses, instead check if status is something, in this case, COMPLETED).

app/data_models/progress_store.py
```diff
@@ -125,12 +124,6 @@ def update_section_status(
                 self._progress[section_key].status = status
                 self._is_dirty = True
 
-        elif status == CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED:
```
petechd, line 128: This could be improved now, since there is no `elif` anymore, you can return early etc., for example:
```python
    def update_section_status(
        self, status: CompletionStatus, section_key: SectionKey
    ) -> bool:
        """
        Updates the status of the Section or Repeating Blocks for a list item specified by the key based on the given section id and list item id.
        """
        if section_key not in self._progress:
            return False
        if self._progress[section_key].status != status:
            self._progress[section_key].status = status
            self._is_dirty = True
            return True
```

app/views/contexts/hub_context.py
```diff
@@ -109,24 +95,13 @@ def get_row_context_for_section(
             ]
         }
 
-        if section_status in (
-            CompletionStatus.COMPLETED,
-            CompletionStatus.INDIVIDUAL_RESPONSE_REQUESTED,
-        ):
+        if section_status in (CompletionStatus.COMPLETED,):
```
petechd, line 98: Comma not needed anymore?

tests/integration/individual_response/test_individual_response.py
```diff
@@ -1,1336 +0,0 @@
-# pylint: disable=too-many-lines
```
petechd, line 1: You need to also remove: 
```js
individualResponseGuidance() {
    return '[data-qa="individual-response-url"]';
  }
```
from question.page.js

app/data_models/progress_store.py
Yuyuutsu: Updated 👍

app/views/contexts/hub_context.py
Yuyuutsu: Updated 👍

tests/integration/individual_response/test_individual_response.py
Yuyuutsu: Removed 👍

petechd (commented):
To fully test this we need to change the validator branch as well (the `tag=latest` line) and revert after the Validator changes are merged?

templates/sectionsummary.html
```diff
@@ -30,7 +30,8 @@ <h1>{{ content.summary.title }}</h1>
                 {% set add_link_text = summary.add_link_text %}
                 {% set empty_list_text = summary.empty_list_text %}
                 {% set list_title = summary.title %}
-                <div {{ 'class="ons-u-mb-l"' if not loop.last }} data-qa="{{ summary.list_name }}-list-summary">
```
petechd, line 33: Why was this change added? Seems unrelated to your changes?

Yuyuutsu: Oh I think was an accidental reformat. Reverted now 👍

petechd (approved):

PR #1579, opened 2025-01-14
Migrate DS to version 72.2.0

petechd:
This migrates Runner to version 72 of the Design System. Adds the latest changes to the html structure for view submitted response, errors and summaries templates.

### How to review
Run local instance of the application and check if templates are as expected.

### Checklist


berroar (commented):
Noticed some spacing differences when compared with Staging, but guessing they are expected @rmccar?

rmccar:
> Noticed some spacing differences when compared with Staging, but guessing they are expected @rmccar?

There are some spacing changes included in v72 with the new spacing grid changes so there probably would be some but would be good to go through the changes you've noticed just to be sure

tests/functional/spec/list_collector/list_collector_section_summary.spec.js
```diff
@@ -351,9 +356,5 @@ const answerUkBasedQuestion = async () => {
 };
 
 const companiesListRowItem = (row, index) => {
```
berroar, line 358: Index no longer being used here.
```suggestion
const companiesListRowItem = (row) => {
```

berroar (commented):
Have noticed some weirdness on summary pages, this for example when viewing a summary page in the supplementary data schema.

Looks like the list name is being appended to the row 🤔

Screenshot 2025-01-23 at 13 33 00

rmccar:
> > Have noticed some weirdness on summary pages, this for example when viewing a summary page in the supplementary data schema.
> > Looks like the list name is being appended to the row 🤔
> > Screenshot 2025-01-23 at 13 33 00
> 
> Having checked the above, it looks like this issue pops up when we have a list summary for a given section but we don't specify an `item_label` e.g. the config one from the example above.
> 
> ```
> "summary": {
>                 "show_on_completion": true,
>                 "items": [
>                     {
>                         "type": "List",
>                         "for_list": "products",
>                         "title": "Products",
>                         "empty_list_text": "There are no products"
>                     }
>                 ],
>                 "show_non_item_answers": true
>             },
> ```
> 
> Example of the same scenario with the `item_label` configured: Screenshot 2025-01-24 at 09 41 31
> 
> Setting an `item_label` is optional and a design choice however, so it looks like something needs to be fixed.

Ok great cheers for this, thats really useful to know, I can take this and test it in the DS

.design-system-version
```diff
@@ -1 +1 @@
-70.0.16
+72.2.0
```
rmccar, line 1: I think you may as well upgrade to 72.3.0

app/jinja_filters.py
```diff
@@ -482,11 +482,11 @@ def __init__(  # noqa: C901, R0912 pylint: disable=too-complex, too-many-branche
...
         ) and answer.get("label"):
-            self.rowTitle = answer["label"]
-            self.rowTitleAttributes = {"data-qa": answer["id"] + "-label"}
+            self.title = answer["label"]
+            self.titleAttributes = {"data-qa": answer["id"] + "-label"}
         else:
-            self.rowTitle = strip_tags(question["title"])
-            self.rowTitleAttributes = {"data-qa": question["id"]}
+            self.title = strip_tags(question["title"])
+            self.titleAttributes = {"data-qa": question["id"]}
```
rmccar, line 489: How come these names have changed? Its still referring to the row title

.design-system-version
petechd: I think we agreed in tech session that we will use 72.2.0.

rmccar: Yeah I know what you mean because it is creeping up but as this is likely to not impact this PR I thought it would be worth at least testing it. But that can be done in another PR I guess

rmccar (approved):

berroar (approved):

PR #1582, opened 2025-01-15
Validate schemas script fix

petechd:
This rewrites our validation script in Python to fix the "completion" bug. Should work exactly the same plus adds some logging instead of "echo" commands to make it more insightful.

### How to review
Add some faulty schemas and pass "./schemas" directory as an argument to check if all the functionalities work as expected. In Runner there are two "make" commands to run validation (not the validator itself, that's a separate command). They should both work as expected.

### Checklist


scripts/validate_test_schemas.py
berroar, line 1: I think we might be better off creating this script in the Validator repo and then it can be shared between runner and the schemas repo. Would mean one source of truth/only one file to update 👍 Just need to install it as a dev dependency as we do with translations elsewhere.

scripts/validate_test_schemas.py
```diff
@@ -0,0 +1,131 @@
...
+            check=True,
+        )
+        return schema_path, result.stdout
+    except subprocess.CalledProcessError as e:
+        logger.info(f"Error validating schema {schema_path}: {e}")
+        return schema_path, None
+
+
+def main():
+    # pylint: disable=global-statement, broad-exception-caught
```
berroar, line 45: Think we can avoid using globals by setting up a class for these vars.  Also makes sense with the comment above if we're importing this as a class from the Validator repo 👍

petechd: I moved them into the `main` now so we avoid globals, not sure if we want to move it into a class if we don't decide to move the script altogether.

petechd:
I had to remove the structlog and use standard library logging as it was no longer available in Github Actions (it was cached before so it worked). The logging string interpolation had to be refactored to pass the linting.

Yuyuutsu (approved):
Works as expected 👍

Yuyuutsu (approved):
Works as expected 👍

Makefile
```diff
@@ -67,10 +67,10 @@ generate-spec:
 	poetry run python -m tests.functional.generate_pages schemas/test/en/$(SCHEMA).json ./tests/functional/generated_pages/$(patsubst test_%,%,$(SCHEMA)) -r '../../base_pages' -s tests/functional/spec/$(SCHEMA).spec.js
 
 validate-test-schemas:
-	./scripts/validate_test_schemas.sh
+	poetry run python -m scripts.validate_test_schemas
 
 validate-test-schema:
-	./scripts/validate_test_schemas.sh $(SCHEMA_PATH)$(SCHEMA).json
+	poetry run python -m scripts.validate_test_schemas $(SCHEMA_PATH)
```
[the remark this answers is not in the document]
petechd: It looks like a mistake by me and if we change it back to 
```shell
poetry run python -m scripts.validate_test_schemas $(SCHEMA_PATH)$(SCHEMA).json
``` 
it will work as expected. So I might have to change it back.

VirajP1002 (approved):

scripts/validate_test_schemas.py
```diff
@@ -0,0 +1,137 @@
...
+            logging.error("\033[31mHTTP Status: %s\033[0m", response)
+            if checks != 1:
+                logging.info("Retrying...\n")
+                time.sleep(5)
+            else:
+                logging.info("Exiting...\n")
+                sys.exit(1)
+            checks -= 1
+        else:
+            checks = 0
```
berroar, line 43: Minor, but this could be made a bit simpler to remove the nesting e.g. something like
```suggestion
def check_connection():
    for connection_attempts in range(4, 0, -1):
        response = subprocess.run(
            [
                "curl",
                "-so",
                "/dev/null",
                "-w",
                "%{http_code}",
                "http://localhost:5002/status",
            ],
            capture_output=True,
            text=True,
            check=False,
        ).stdout.strip()

        if response == "200":
            return

        logging.error("\033[31m---Error: Schema Validator Not Reachable---\033[0m")
        logging.error("\033[31mHTTP Status: %s\033[0m", response)

        if connection_attempts == 1:
            logging.info("Exiting...\n")
            sys.exit(1)

        logging.info("Retrying...\n")
        time.sleep(5)
```

scripts/validate_test_schemas.py
```diff
@@ -0,0 +1,137 @@
...
+        schemas = [
+            os.path.join(file_path, f)
+            for f in os.listdir(file_path)
+            if f.endswith(".json")
+        ]
+        logging.info("--- Testing Schemas in %s ---", file_path)
+    else:
+        file_path = sys.argv[1]
+        schemas = [sys.argv[1]]
+        logging.info("--- Testing %s Schema ---", file_path)
```
berroar, line 58: Minor, can just be schema here rather than calling arg twice:
```suggestion
        schema = sys.argv[1]
        schemas = [schema]
        logging.info("--- Testing %s Schema ---", schema)
```

berroar (approved):
LGTM 👍

VirajP1002 (approved):
Changes look good to me 👍

PR #1601, opened 2025-02-04
Fix issue rendering list summaries in DS versions 72.1+

petechd:
We had to fix an issue in Runner template caused by the nesting changes in DOM for onsSummary component from DS. The PR in which these changes were added: https://github.com/ONSdigital/design-system/pull/3408. Pinning to the latest version of DS makes list summaries (list collector items with dynamic answers) unusable. The object we pass to the jinja template that uses onsSummary component had to be modified. Title element  (only for list collector content with related answers) needs to be at the top level now so looks like: 
```python
{"title": title, "itemsList": row_items}
```
for all the other scenarios rows stay as:
```python
{"itemsList": row_items}
```
This PR is based not on the `main` branch to make the review easier, but it contains all the previous migration to DS 72 changes. The modified DOM structure (different nesting) of list summaries can be viewed and compared with version 70 in dev tools, `test_supplementary_data` but also `test_list_colllector_repeating_blocks_with_summary` etc.

### How to review
Make sure that there are no other issues remaining, especially check all possible variants of list summaries (wherever list summary is being used, so normal list collectors, supplementary data, repeating blocks, repeating answers, dynamic answers, view submission response etc). Check if the existing tests are sufficient.

### Checklist

### DIFF
--- a/app/jinja_filters.py
+++ b/app/jinja_filters.py
@@ -753,7 +753,7 @@ def map_list_collector_config(
         title = item_label or item_name
 
         if related_answers:
-            # Supplementary Data (List Collector Content) list summaries don't use the row_item created above, as they are not editable
+            # List summaries driven by List Collector Content pages don't use the row_item created above, as they are not editable
             row_items_related: list = [row_item] if editable else []
 
             for block in related_answers[list_item["list_item_id"]]:


petechd:
Had to change the base  branch to main to trigger all the actions but for review we will change the base branch back so checking these changes will be easier.

app/jinja_filters.py
```diff
@@ -751,10 +750,12 @@ def map_list_collector_config(
         if item_label:
             row_item["valueList"] = [{"text": item_name}]
 
-        row_item["title"] = item_label or item_name
-        row_items: list = [row_item]
+        title = item_label or item_name
 
         if related_answers:
+            # Supplementary Data (List Collector Content) list summaries don't use the row_item created above, as they are not editable
```
berroar, line 756: Maybe something like this might be clearer?
```suggestion
            # List summaries driven by List Collector Content pages don't use the row_item created above, as they are not editable
```

app/jinja_filters.py
```diff
@@ -765,9 +766,18 @@ def map_list_collector_config(
                     use_answer_label=True,
                     item_name=item_name,
                 )
-                row_items.extend(summary_row.itemsList)
-
-        rows.append({"itemsList": row_items})
+                row_items_related.extend(summary_row.itemsList)
+            # Again, List Collector Content list summaries don't use the row_item with "title" and are structured differently
```
berroar, line 770: I'm not sure we need this comment since it kind of repeats the one above, was thinking some of these maybe more suited/better explained in docstrings, and might drive us to think of some things to refactor. Obviously not for this ticket though 👍

berroar (approved):

app/jinja_filters.py
petechd: It is pretty convoluted what happens here, I agree, docstring would be good, it's just that I didn't even look at the whole of this code (for this class) only the bottom part that was relevant for my task.

rmccar: Yeah I think this will need to be refactored at some point but as for this ticket its fine

rmccar (approved):

PR #1646, opened 2025-04-22
ECI 1603 Resolve Ruff lint error messages issues

Farhanam76:
### Motivation and Context
To resolve the ruff linting rules EM101 & EM102 to improve the code quality, making sure it aligns with best practice 

### What has changed
Assigned strings and f-strings to a variable and passed the variable to the exception 

### How to test?
Run the `make lint` command and check if there are any EM101 & EM102 ruff linting error messages

### Links
Jira ticket : https://jira.ons.gov.uk/browse/ECI-1603
Docs : https://docs.astral.sh/ruff/rules/raw-string-in-exception/  https://docs.astral.sh/ruff/rules/f-string-in-exception/


app/forms/validators.py
```diff
@@ -425,9 +425,10 @@ def __call__(
             try:
                 conditions.remove("equals")
             except ValueError as exc:
+                invalid_mutiple_conditions = "There are multiple conditions, but equals is not one of them. "
```
berroar, line 428: I think this change is breaking the test at the moment because it's split the message into multiple lines.

app/authentication/authenticator.py
```diff
@@ -179,7 +179,8 @@ def create_session_questionnaire_store(
 
 def decrypt_token(encrypted_token: str | None) -> dict[str, Any]:
     if not encrypted_token:
-        raise NoTokenException("Please provide a token")
+        missing_token = "Please provide a token"
```
berroar, line 182: I think rather than having a variable the line above exception, it might be better to have the error message string in the actual exception class - especially for things like this where there is only ever one error.

berroar (changes requested):

app/forms/validators.py
Farhanam76: I pushed the changes where I put in one string, but If you want to keep it as 2 separate strings as it originally was, I can put brackets around them

app/authentication/authenticator.py
Farhanam76: I pushed the new code changes where the error message is passed into the exception class

app/authentication/no_token_exception.py
```diff
@@ -1,5 +1,5 @@
 class NoTokenException(Exception):
-    def __init__(self, value: str | int) -> None:
+    def __init__(self, value: str | int = "Please provide a token") -> None:
```
berroar, line 2: I think the type hinting can be updated here as not sure we ever pass an int?

app/authentication/user.py
```diff
@@ -7,4 +7,5 @@ def __init__(self, user_id: str | None, user_ik: str | None) -> None:
             self.user_id = user_id
             self.user_ik = user_ik
         else:
-            raise ValueError("No user_id or user_ik found in session")
+            missing_user_session_error = "No user_id or user_ik found in session"
```
berroar, line 10: For some of these it maybe better to set the error string as an attribute of the class rather than setting it directly above the call, makes it a bit more readable and easier to extend in future.

app/utilities/metadata_parser_v2.py
```diff
@@ -106,12 +107,14 @@ def validate_schema_options(  # pylint: disable=no-self-use, unused-argument
...
+                missing_metadata_option = "Neither schema_name, schema_url or cir_instrument_id has been set in metadata"
                 raise ValidationError(
-                    "Neither schema_name, schema_url or cir_instrument_id has been set in metadata"
+                    missing_metadata_option
                 )
             if len(options) > 1:
+                error_msg = f"Only one of schema_name, schema_url or cir_instrument_id should be specified in metadata, but {', '.join(options)} were provided"
                 raise ValidationError(
-                    f"Only one of schema_name, schema_url or cir_instrument_id should be specified in metadata, but {', '.join(options)} were provided"
+                    error_msg
```
berroar, line 117: Only minor but we need to be consistent with the naming approach for the error message vars (and variables in general), to make they are easily understood. This one is quite generic compared to the rest. Also, we don't generally abbreviate in our var names so would be `message` rather than `msg`.

app/views/handlers/block_factory.py
```diff
@@ -56,22 +56,25 @@ def get_block_handler(
 ) -> Any:
     block = schema.get_block(block_id)
 
+    block_error_msg = (f"block id {block_id} is not valid for this schema",
```
berroar, line 59: At the moment this approach is not very readable, if we want go down this route it would be bette to use a dict so it is more explicit. It might be better to assign the dict at the module level also so we're not creating it every time this method is called. This also applies to other instances in this file.

berroar (changes requested):

app/authentication/no_token_exception.py
Farhanam76: I updated this code and removed the int

app/authentication/user.py
Farhanam76: I just pushed this change where it's an attribute of the class, and for some of the other files, I have made changes to follow the same approach as well.

app/utilities/metadata_parser_v2.py
Farhanam76: That file has been updated so that the variable names are now specific to the content of the strings. I’ve also updated other files, changing variable names like msg to message for clarity.

app/views/handlers/block_factory.py
Farhanam76: I followed the suggested approach using a dictionary, as I couldn’t assign the strings and f-strings to variables due to an error message about too many local variables.

app/questionnaire/placeholder_renderer.py
```diff
@@ -19,6 +19,7 @@ class PlaceholderRenderer:
     Renders placeholders specified by a list of pointers in a schema block to their final
     strings
     """
+    placeholder_not_found = "No placeholder found to render"
```
Yuyuutsu, line 22: As this is a constant I think it would be better if we made the variable upper case to `PLACEHOLDER_NOT_FOUND`

app/setup.py
```diff
@@ -331,9 +332,8 @@ def setup_redis(application):
 def setup_submitter(application):
     if application.config["EQ_SUBMISSION_BACKEND"] == "gcs":
         if not (bucket_name := application.config.get("EQ_GCS_SUBMISSION_BUCKET_ID")):
-            raise MissingEnvironmentVariable(
-                "Setting EQ_GCS_SUBMISSION_BUCKET_ID Missing"
-            )
+            missing_bucket_id_message = "Setting EQ_GCS_SUBMISSION_BUCKET_ID Missing"
+            raise MissingEnvironmentVariable(missing_bucket_id_message)
```
Yuyuutsu, line 336: Maybe we should put these variables in the MissingEvniromentVariable class. Then we can call them from the class. Any opinons from other devs?

app/setup.py
```diff
@@ -365,7 +367,8 @@ def setup_submitter(application):
         application.eq["submitter"] = LogSubmitter()
 
     else:
-        raise NotImplementedError("Unknown EQ_SUBMISSION_BACKEND")
+        unknown_submission_message = "Unknown EQ_SUBMISSION_BACKEND"
```
Yuyuutsu, line 370: Also for the NotImplementedErrors we should put in a dictionary too.

app/questionnaire/placeholder_renderer.py
Farhanam76: In this file and others, I've made the variable names uppercase.

app/setup.py
Farhanam76: I updated the code to place these variables in a dict.

app/setup.py
Farhanam76: I placed the variables in a class since it's a custom exception, but I can switch to a dict if others prefer that approach.

app/forms/questionnaire_form.py
```diff
@@ -337,7 +337,9 @@ def _get_period_range_for_single_date(
         )
 
         if not min_period_date or not max_period_date:
-            raise ValueError("Period range must have a start and end date")
+
+            invalid_period_range = "Period range must have a start and end date"
```
berroar, line 341: I think we need to be consistent throughout if we're gonna make them constants 👍

Farhanam76:
The variables are lowercase to avoid ruff lint errors requiring lowercase names in functions. I can put it at the module level if a constant variable is preferred. Some are class attributes, while others aren't. For complex classes with many methods, I place them below the method definitions for readability. Some I have placed are inside the if condition as they are part of a loop.

Yuyuutsu:
> The variables are lowercase to avoid ruff lint errors requiring lowercase names in functions. I can put it at the module level if a constant variable is preferred. Some are class attributes, while others aren't. For complex classes with many methods, I place them below the method definitions for readability. Some I have placed are inside the if condition as they are part of a loop.

I perfer them being constant variables if thats okay. I don't know what other devs think.

app/questionnaire/rules/operations.py
```diff
@@ -135,6 +135,7 @@ def resolve_date_from_string(
         offset: DateOffset | None = None,
         offset_by_full_weeks: bool = False,
     ) -> date | None:
+        invalid_negative_days_offset = "Negative days offset must be less than or equal to -7 when used with `day_of_week` offset"
```
Yuyuutsu, line 138: Could you capitalise this variable please

Yuyuutsu, line 138: I think we should make this a class level constant vairable

app/routes/questionnaire.py
```diff
@@ -72,13 +72,12 @@
 @questionnaire_blueprint.before_request
 @login_required
 def before_questionnaire_request() -> Response | None:
+    repeated_submission_error = "The Questionnaire has been previously submitted"
```
Yuyuutsu, line 75: I think we should make this a constant outside the function

app/routes/session.py
```diff
@@ -301,19 +300,22 @@ def get_signed_out() -> Response | str:
 
 
 def get_runner_claims(decrypted_token: Mapping[str, Any]) -> dict:
+    runner_claims_error = "Invalid runner claims"
```
Yuyuutsu, line 303: If we can i think we should put this outside the function as a Constant near the top of the module. Also I think we should rename this variable to `INVALID_RUNNER_CLAIMS_ERROR` to be more specific

app/routes/session.py
```diff
@@ -301,19 +300,22 @@ def get_signed_out() -> Response | str:
...
 
     except ValidationError as e:
-        raise InvalidTokenException("Invalid runner claims") from e
+        raise InvalidTokenException(runner_claims_error) from e
 
 
 def get_questionnaire_claims(
     decrypted_token: Mapping, schema_metadata: Iterable[Mapping[str, str]]
 ) -> dict:
+    questionnaire_claims_error = "Invalid questionnaire claims"
```
Yuyuutsu, line 314: Same for here

Yuyuutsu (approved):

liamtoozer (commented):
Changes looks good 👍  Just some very minor suggestions for naming etc

Just wondering if it's worth adding `_error_message` and `_ERROR_MESSAGE` suffixes for all of the vars & consts to show clear intent? Appreciate that may make things more verbose though, so happy to skip this if so

app/forms/validators.py
```diff
@@ -469,7 +468,8 @@ def _is_valid(
         if condition == "less than or equals":
             return total <= target_total, "TOTAL_SUM_NOT_LESS_THAN_OR_EQUALS"
 
-        raise NotImplementedError(f"Condition '{condition}' is not implemented")
+        invalid_condition = f"Condition '{condition}' is not implemented"
```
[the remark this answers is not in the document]
Farhanam76: For this one, the f-string uses a local variable defined within the method, so I kept it at the method level as I'm not sure if it can be moved to the class level?

app/questionnaire/placeholder_transforms.py
```diff
@@ -300,6 +300,7 @@ def format_ordinal(
         return f"{number_to_format}{indicator}"
 
     def get_ordinal_indicator(self, number_to_format: int) -> str:
+        invalid_language_code = f"Language code '{self.language}' not implemented."
```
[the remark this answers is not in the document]
Farhanam76: I’ve just updated the code it should now be placed right before the raised exception.

app/views/handlers/submit_questionnaire.py
```diff
@@ -11,14 +11,16 @@
 
 
 class SubmitQuestionnaireHandler:
+    SUBMIT_PAGE_MESSAGE = "Submit page not enabled"
```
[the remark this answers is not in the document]
Farhanam76: I've updated the code to reflect the new variable name. I also renamed the other variables to include the error_message suffix, aiming to keep the naming consistent without making it too verbose.

app/setup.py
```diff
@@ -293,7 +312,7 @@ def setup_storage(application):
     elif application.config["EQ_STORAGE_BACKEND"] == "dynamodb":
         setup_dynamodb(application)
     else:
-        raise NotImplementedError("Unknown EQ_STORAGE_BACKEND")
+        raise NotImplementedError(UNKNOWN_IMPLEMENTED_ERROR["unknown_storage_backend"])
```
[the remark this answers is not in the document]
Farhanam76: I've refactored them as constant variables instead of dict

app/forms/validators.py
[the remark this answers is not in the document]
liamtoozer: Ah good point, sorry I missed this

liamtoozer (approved):

PR #1649, opened 2025-04-25
Upgrade Node Version to v22.15.0

Yuyuutsu:
Runner currently uses Node v20 which will soon be going under maintenance support only.  This PR upgrades our node version to the latest stable release - v22.15.0

### How to review
Ensure `npm` commands work
Check actions and all places that Node v22.15.0 is being used

### Checklist


petechd:
You might need to try to solve the issue with the modules not recognised by Node 23:
`module is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.`

rmccar:
I think we should only update to v22. Only the even numbered versions of node go into LTS and the odd versions become unsupported after 6 months

berroar:
> You might need to try to solve the issue with the modules not recognised by Node 23: `module is not specified and it doesn't parse as CommonJS. Reparsing as ES module because module syntax was detected. This incurs a performance overhead.`

Are these warnings still present in v22?

berroar (approved):

PR #1652, opened 2025-04-29
Migrate to DS 72

petechd:
This introduces updates to migrate the project to Design System version 72. Relevant templates were refactored to align with the new Design System components and structure. The branch was updated to be compatible with the latest version of DS 72 (72.8.0).

### How to review
Refer to the migration guide attached to this task, check the release notes of the DS releases and familiarise yourself with any relevant comments on the card. 

### Checklist


rmccar (approved):

Yuyuutsu (approved):

tests/functional/spec/textarea.spec.js
```diff
@@ -43,7 +43,7 @@ describe("Textarea", () => {
 
   it("The user is unable to add more characters when the limit is reached", async () => {
```
rmccar, line 44: I think this test description will need to change with this update. It no longer reflects how the textarea is supposed to work

Yuyuutsu:
Noticed that the footer is not fully grey-ed out when the body does not have much content. For example in `checkbox.json`:

image

rmccar:
> > Noticed that the footer is not fully grey-ed out when the body does not have much content. For example in `checkbox.json`: image
> 
> This will be down to the changes to the grid, some padding and line height has been changed across all pages so that all element sit on the grid on whole pixel numbers. If you reduce the content on the main branch you will get the same issue. It just so happens that this length of content shows up this between the two branches. We can ask the question with the designers if they are happy with this, but this isn't really a bug. Good spot though!

In fact looking at it again now, it also shows up on main on this page. It just the white bit at the bottom of the page is much smaller

Screenshot 2025-09-10 at 15 33 51

PR #1667, opened 2025-05-21
Refactor supplementary data functional test suite

Farhanam76:
Currently, the supplementary data suite is impacted by flakiness. This PR shows that the functional test suite has been refactored to make it clear and easier to navigate the tests for future use.

### How to review
* [ ] Check that the JSON Schema files are properly split and formatted.
* [ ] Ensure the new functional test specs are split and reference the correct JSON schemas
* [ ] Ensure functional test specs pass.
* [ ] Check the README file is correctly updated.

### Checklist


tests/functional/spec/supplementary_data_interstitial/supplementary_data_with_interstitial_and_calculated_summary.spec.js
```diff
@@ -0,0 +1,136 @@
+import { click, assertSummaryTitles } from "../../helpers";
```
petechd, line 1: We don't name specs based on the schema they open; instead, we focus on the functionality being tested in the spec. For example, `conditional_question_title` opens `test_titles_radio_and_checkbox.json`, so it should be renamed to reflect what is being tested in this schema. Are these specs grouped thematically? The title should clearly indicate what is being tested.

petechd:
In it's current state it causes too much flakiness we need to discuss how to investigate it and reduce it.

PR #1675, opened 2025-06-13
Prevent 'main' major dependency version dependabot bumps

liamtoozer:
Dependabot is currently raising PRs when a `main` poetry dependency is bumped to a major version. Our `pyproject.toml` is set to only bump minor & patch, so this should be reflected in our dependabot config also. 

This PR will mean that dependabot will continue to raise PRs but only for minor and patch version bumps. 

Note: the dev dependencies are fine to keep bumping to major versions, as they will not impact the Runner build.

### How to review
Difficult to check but I'll manually retrigger the dependabot PRs to see what happens

### Checklist


berroar (approved):

liamtoozer:
Will close this for now to prevent additional overheads with devs investigating major version bumps separately. Our automated benchmarks should hopefully catch any major issues anyway. Can always revisit this if needed later on

PR #1685, opened 2025-06-27
Remove concurrency from functional tests (4 jobs)

petechd:
PR with the fixes to remove the flakiness from functional tests in builds. Two approaches were created: [one with functional tests split into four separate jobs](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1685), which is 50% faster but more prone to flakiness, and [another with a single job](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1709). This gives us two options to choose from.

- The matrix strategy was removed as a major contributor to flakiness.  
- In this PR, all functional tests are run consecutively and separated into multiple independent jobs (different configurations of suites used). The build time is now 15 minutes, the reliability of the build is lower than this single job PR: . Flakiness in tests occurs in the 6th consecutive run.  
- Hub and spoke supplementary data tests had to be separated out and made synchronous as they were a major cause of flakiness.  
- Two minor tests skipped: cookie banner delete not re-appearing test, supplemnatry data heading on one of the pages test. Brings the total of skipped tests up to 3. These two are notoriously flaky in most builds and impossible to fix. 
- Some other tests could be made synchronous, and Webdriver's Promises (e.g., `isDisplayed()`) could be forced to be synchronous by using the `browser.call()` (https://webdriver.io/docs/api/browser/call/) solution, but that didn't work for our tests.  
- The Docker Compose task and loading design system templates are also contributors to the flakiness (and occasionally break builds), but in this PR, the focus was on fixing the functional tests.  

Webdriver Sync vs. Async Mode and common issues: https://v7.webdriver.io/docs/sync-vs-async/

### How to review
Test the functional tests in build of this PR, branch off to make changes.

### Checklist


Yuyuutsu:
Works nice, flakiness is definitely reduced here.I prefer this option but if we want to prioritise  even lower flakiness happy with the single use apporach

PR #1724, opened 2025-08-07
EQS 241- Reference docker scripts to point to GAR

EmmanuelA619:
We lost our license to use Dockerhub as an org, we have stopped being able to push app images to anywhere other than Google's artifact registry. 

Docker compose scripts need to remove references to the ONS digital dockerhub org, and point at the artifact registry instead.

This PR is a change to point docker compose scripts in eQ applications to GAR rather than Dockerhub

### How to review
clone ONSdigital/eq-questionnaire-runner and checkout into the branch
From the root of the repo run the below to get eq-questionnaire-runner running:
RUNNER_ENV_FILE=.development.env docker compose up -d

To launch a survey, navigate to http://localhost:8000/

This will:
Pull all the services (launcher, sds, cir) from GAR
Start them in the background

https://github.com/ONSdigital/eq-questionnaire-runner/blob/main/README.md

### Checklist


docker-compose-schema-validator.yml
```diff
@@ -1,13 +1,13 @@
 services:
   ajv-validator:
-    image: onsdigital/eq-questionnaire-validator:${TAG}-ajv
```
liamtoozer, line 3: I think this `TAG` is being used by `./scripts/run_validator.sh`, which is called from a few different places, so changing this to always be `latest` might have an impact in other places, like with `make run-validator`

I'm guessing this file was never intended to be called directly without that wrapper script, so I'd be inclined to keep the `TAG` in for flexibility for now?

PR #1749, opened 2025-09-16
Pin Compromised NPM Dependencies

Yuyuutsu:
Starting September 8th 13:16 UTC it has been reported that the npm `debug` and `chalk` packages were compromised.
This PR is to pin any packages related to the compromised packages to prevent any upgrades to the compromised versions.

### How to review

- Check if all compromised packages have been pinned to a safe version
- Check front end tests work as expected

### Checklist


VirajP1002 (commented):
Looks good so far, just a quick question, I was wondering if `wrap-ansi-cjs` needs to updated here and added to the overrides? We weren't sure if it was supposed to be using `7.0.0`? Even though it's a safe version, it might update itself? But I guess if it's in the lock we might not need to? The same goes for `strip-ansi-cjs`

image

rmccar:
Do we not want to just remove all the `^` here? instead of using overrides for the specific packages that were compromised this time? This would still leave us vulnerable to other packages not in that list that were compromised

Yuyuutsu:
Discussed with Viraj and Farhana - Pinning each version specifically for every parent package, to maintain the same versions as before.

liamtoozer (dismissed):

farres1 (changes requested):

rmccar (approved):

package-lock.json
```diff
@@ -5885,10 +6447,11 @@
             }
         },
         "node_modules/error-ex": {
-            "version": "1.3.2",
-            "resolved": "https://registry.npmjs.org/error-ex/-/error-ex-1.3.2.tgz",
-            "integrity": "sha512-7dFHNmqeFSEt2ZBsCriorKnn3Z2pj+fd9kmI6QoWw4//DL+icEBfc0U7qJCisqrTsKTjw4fNFy2pW9OqStD84g==",
+            "version": "1.3.4",
```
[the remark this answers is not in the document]
Paul-Joel: 1.3.4 is a roll forward of 1.3.2, to enable the compromised version to be destroyed. 1.3.4 is safe to use.

package-lock.json
```diff
@@ -12796,10 +12870,11 @@
             }
         },
         "node_modules/strip-ansi/node_modules/ansi-regex": {
-            "version": "6.1.0",
-            "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-6.1.0.tgz",
-            "integrity": "sha512-7HSX4QQb4CspciLpVFwyRe79O3xsIZDDLER21kERQ71oaPodF8jL725AgJMFAYbooIqolJoRLuM81SpeUkpkvA==",
+            "version": "6.2.2",
```
VirajP1002, line 12873: This version for `ansi-regex` is above the compromised version which is 6.2.1 (According to the docs) does this need to be pinned to a lower version?

package-lock.json
```diff
@@ -12768,10 +12840,11 @@
...
-            "version": "7.1.0",
-            "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-7.1.0.tgz",
-            "integrity": "sha512-iq6eVVI64nQQTRYq2KtEg2d2uU7LElhTJwsH4YzIHZshxlgZms/wIc4VoDQTlG/IvVIrBKG06CrZnp0qv7hkcQ==",
+            "version": "7.1.2",
+            "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-7.1.2.tgz",
+            "integrity": "sha512-gmBGslpoQJtgnMAvOVqGZpEz9dyoKTCzy2nfz/n8aIFhN/jCE/rCmcxabB6jOOHV+0WNnylOxaxBQPSvcWklhA==",
             "dev": true,
+            "license": "MIT",
             "dependencies": {
                 "ansi-regex": "^6.0.1"
```
VirajP1002, line 12849: I think this is the reason to why `ansi-regex` is above compromised version

package-lock.json
VirajP1002, line 12873: This version for ansi-regex is above the compromised version which is 6.2.1 (According to the docs) does this need to be pinned to a lower version?

package-lock.json
```diff
@@ -12768,10 +12840,11 @@
             }
         },
         "node_modules/strip-ansi": {
-            "version": "7.1.0",
-            "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-7.1.0.tgz",
-            "integrity": "sha512-iq6eVVI64nQQTRYq2KtEg2d2uU7LElhTJwsH4YzIHZshxlgZms/wIc4VoDQTlG/IvVIrBKG06CrZnp0qv7hkcQ==",
+            "version": "7.1.2",
```
VirajP1002, line 12843: This version for `strip-ansi` is above the compromised version which is 7.2.1 (According to the docs) does this need to be pinned to a lower version? I think this is the reason to why ansi-regex is set to `^6.x.x`

package-lock.json
```diff
@@ -13700,9 +13766,9 @@
             }
         },
         "node_modules/webdriver/node_modules/chalk": {
-            "version": "5.4.1",
-            "resolved": "https://registry.npmjs.org/chalk/-/chalk-5.4.1.tgz",
-            "integrity": "sha512-zgVZuo2WcZgfUEmsn6eO3kINexW8RAE4maiQ8QNs8CtpPCSyMiYsULR3HQYkm3w8FIA3SberyMJMSldGsW+U3w==",
+            "version": "5.6.2",
```
VirajP1002, line 13769: This version of chalk is above 5.6.1 which is the compromised version, I think there are few more instances of this across the lock, do we need to manually pin in the package.json?

package-lock.json
```diff
@@ -3126,20 +3258,21 @@
...
             }
         },
         "node_modules/@wdio/local-runner/node_modules/@wdio/logger": {
             "version": "8.38.0",
             "resolved": "https://registry.npmjs.org/@wdio/logger/-/logger-8.38.0.tgz",
             "integrity": "sha512-kcHL86RmNbcQP+Gq/vQUGlArfU6IIcbbnNp32rRIraitomZow+iEoc519rdQmSVusDozMS5DZthkgDdxK+vz6Q==",
             "dev": true,
+            "license": "MIT",
             "dependencies": {
                 "chalk": "^5.1.2",
```
VirajP1002, line 3277: Is this chalk being set to `^5.1.2` at risk of being upgrade to "5.6.x"? If do we need to pin chalk (as mentioned in the previous comment)

package-lock.json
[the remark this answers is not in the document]
VirajP1002: Would this apply to the other comments below as we will need to delete them if that's the case. We were working under the assumption that any version above could be still at risk of being compromised. We did look on a few of their GitHub's and saw that they address vulnerabilities?

package-lock.json
Yuyuutsu: 7.1.2 looks like it resolves the venerability found in 7.1.1
https://github.com/chalk/strip-ansi/releases

package-lock.json
Yuyuutsu: Looks like v5.6.2 resolves the vunearbility in v5.6.1
https://github.com/chalk/chalk/releases/tag/v5.6.2

PR #1770, opened 2025-11-03
Upgrade WebdriverIO to v9.24.0

Yuyuutsu:
This PR upgrades all WDIO packages to `v9.24.0` and updates our functional tests to comply with the new upgrades.

#### Changes made

- `getHTML()` no longer takes a boolean flag. It now accepts an options object. The default options:
`({ includeSelectorTag: true, prettify: true })` were causing differences in the returned HTML, which led to assertion failures (formatting changes and missing alt attributes on logo tags). We have to explicitly set these values to false.

- `newWindow()` now needs to explicitly have a URL value. In v8, `newWindow()` implicitly opened `about:blank` when no URL was provided. In v9 this implicit default was removed, so we now explicitly pass 'about:blank' when opening a new window.

- SDS Tests in GA are failing when in the main test, but work when separated into their own job with "MAX_INSTANCE=1". Also the very first SDS launch always hits the “Sorry, there is a problem with this service” page. So a new helper function  - `openQuestionnaireWithRetry()`. I think we need an investigation on why SDS is failing like this in GA only.

- Chrome options also updated in an attempt to make tests more consistent.

- Babel dependencies have also been upgraded to latest versions

### How to review

- Run Functional tests locally on main and then on this branch and see if they all work as expected
- Check all test-functional commands work as expected
- Do you agree with the changes made? There are multiple routes we can take with these changes, happy to hear out any suggestions.

### Links
- https://webdriver.io/blog/2024/08/15/webdriverio-v9-release/
- https://webdriver.io/docs/api/element/getHTML/
- https://webdriver.io/docs/api/browser/newWindow/

### Checklist


berroar (commented):
Was just wondering with:
`Some tests became flaky due to reliance onbrowser.pause() to avoid race conditions. Therefore a new helper function has been added, waitForPageToLoad(), which waits for the page to fully load.`
Did we test without replacing these at all? I think they were initially added before we went back to running consecutively?

berroar (commented):
Observing a lot of `The data in this module is over two months old.  To ensure accurate Baseline data, please update: `npm i baseline-browser-mapping@latest -D` in the run logs.

tests/functional/wdio.conf.cjs
```diff
@@ -233,7 +253,7 @@ exports.config = {
           cirInstrumentId,
           booleanFlag,
         });
-        this.url(`/session?token=${token}`);
+        await this.url(`/session?token=${encodeURIComponent(token)}`);
```
Yuyuutsu: await to prevent any race conditions - flakiness
encodeURIcomponent for an added layer of security

tests/functional/spec/components/checkbox/mutually_exclusive/mutually_exclusive_currency.spec.js
```diff
@@ -5,7 +5,7 @@ import { click } from "../../../../helpers";
 describe("Component: Mutually Exclusive Currency With Single Checkbox Override", () => {
   beforeEach(async () => {
     await browser.openQuestionnaire("test_mutually_exclusive.json");
-    await browser.pause(100);
+
```
liamtoozer: Very, very (very) minor: Remove blank line to be consistent with other tests? I think there are a few files like this that might need to be updated too

liamtoozer (commented):
Changes look great and just wanted to call out great PR description as well 👍 

I think this is more to do with the bump to `wdio` v9, but I wonder if some of the new flakiness we're seeing more of in Actions now is because of this. I think we might need to spend a bit of extra time to prevent introducing extra flakiness for some of these tests, because I'm seeing a few different failures that weren't cropping up as often with `v8`. Appreciate that you're trying to tackle this already though 👍  Not sure what the solution is yet but we might need to park this and spawn off an investigation into some of the newer intermittent failures?

Also, just a very minor suggestion, I think we might benefit from calling `getHTML()` from inside a central wrapper/helper function, and defaulting with the `{ prettify: false }` object there instead. Otherwise, passing the object in multiple place might be a maintenance overhead if we need to refactor in future? Might help keep some of the assertions to look a bit cleaner?

EmmanuelA619 (approved):

PR #1848, opened 2026-03-23
EQS-726: Test Jasmine setup

petechd:
Test PR to preview how we would migrate to Jasmine. It was also created to check and ensure the functional tests specs are robust and less flaky. Some WebDriver config changes were added (global wait time increased).

### How to review
Check the test runs or re-run the actions to check for flakiness.
If you want to run locally you need to hardcode the re-worked specs into the specs array in WebDriver config: https://github.com/ONSdigital/eq-questionnaire-runner/blob/67de17bb98e5bda567ed65aa607f22c068c7bd49/tests/functional/wdio.conf.cjs#L19

#### How to rework a schema
Most of the time all you need to do is remove the "before" (move the schema open command into the "it" of the test or use "beforeEach" but you need to be careful using that, check the context). Expects don't need initial await. Check my changes for more info.

### Checklist


VirajP1002:
Following 10 runs of GHA, only 3 failed so there is still some flakiness with this solution. However, have you tried combining the suites like we used to before? 

Either way, still think this is a good alternative as Mocha is vulnerable and this allows us to update Webdriver.

petechd:
> Following 10 runs of GHA only 3 failed so there is still some flakiness with this solution. However, have you tried combining the suites like we used to before?
> 
> Either way, still think this is a good alternative as Mocha is vulnerable and this allows us to updated Webdriver.

Combining them made it flakier but some further WebDriver config tweaking could possibly help with the flakiness? Here we get latest WebDriver update and previous vulnerability fix for free so maybe worth pursuing this? Depends how flaky Mocha setup is after updating WebDriver?

liamtoozer (commented):
Good stuff 👍  Looks promising from the initial runs on my forked repo a few times - think it's definitely worth a try!

petechd:
Not needed anymore.

PR #1869, opened 2026-04-13
EQS-897: Add matrix shard config

petechd:
This PR speeds up our functional tests using matrix jobs. It includes a new configurable script to dynamically divide all functional tests into shards. Compared to using a matrix with suites, this is not hardcoded and can be easily customised. The additional flakiness introduced by parallel runs was reduced by adding new `waitUntil` calls and questionnaire reopening. Supplementary data tests had to be consolidated into a single file, as they were notoriously flaky when making multiple simultaneous Mock SDS requests and always failed after introducing shards.

### How to review
Check the changes and compare them with how functional tests were run previously. Feel free to tweak the shard settings to see if results can be improved, but bear in mind we are limited by costs and the next slowest job in the build (unit tests).

### Checklist


VirajP1002 (approved):
Happy with these changes 👍

scripts/split_functional_specs.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# Usage: ./scripts/split_functional_specs.sh <total_shards> <shard_number>
+set -e
+
+TOTAL_SHARDS=$1
```
liamtoozer: 1. Do we want to add any validation/checks for the inputs? I know it's very unlikely to happen, but if the args are empty, or any sort of string etc, the script will exit with a status `0` in Actions, and give the impression that the tests ran successfully even if this script errored?
2. Might be worth matching the var names with those from the Actions config or vice versa?

liamtoozer (approved):
This is a game-changer! Great work. Just a couple of minor, non-blocking questions

scripts/split_functional_specs.sh
petechd: Pushed the changes with both addressed, I hope it works, it's been a long time.

PR #1875, opened 2026-04-22
EQS-613 Resolve Dependency issues

Farhanam76:
Resolve issues identified by Deptry in eQ Questionnaire Runner project. Some of the issues raised by Deptry are false positives, which are: 

- grpcio
- uwsgi
- email-validator
- gevent

Doc:  https://officefornationalstatistics.atlassian.net/wiki/spaces/SDC/pages/186517100/eQ+Services+-+Deptry+report+for+Runner+repositories

Dependencies added to pyproject.toml have been set to fixed versions. These versions can be pinned in a separate ticket if needed.

### How to review

- Make sure the build passes
- Check that the unit test passes
- Make sure Runner can be run and works as expected, and that the dependency changes haven’t introduced any issues.
- Check if there are any other false positives that may have been missed.


pyproject.toml
```diff
@@ -89,6 +85,14 @@ gevent = "^25.9.1"
 babel = "==2.14.0"  # Temporarily pinned - problem for translations found in v2.15.0, see: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1384
 wtforms = "^3.2.1"
 htmlmin2 = "^0.1.13"
+werkzeug = "3.1.5"
+cryptography = "46.0.7"
+jwcrypto = "1.5.6"
+jinja2 = "3.1.6"
+urllib3 = "2.6.3"
+botocore = "1.37.23"
+python-dateutil = ">=2.9.0"
```
liamtoozer: Just a general question - wanted to check why this pin for `python-dateutil`?

pyproject.toml
```diff
@@ -89,6 +85,14 @@ gevent = "^25.9.1"
...
+playwright = "^1.42.0"
```
liamtoozer: I think this is another false positive - `playwright` is only used for our [integration test generator script](https://github.com/ONSdigital/eq-questionnaire-runner/blob/main/scripts/generate_integration_test.py#L4) - so this should stay as a dev-only dependency

pyproject.toml
```diff
@@ -89,6 +85,14 @@ gevent = "^25.9.1"
 babel = "==2.14.0"  # Temporarily pinned - problem for translations found in v2.15.0, see: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1384
 wtforms = "^3.2.1"
 htmlmin2 = "^0.1.13"
+werkzeug = "3.1.5"
```
liamtoozer: Should probably be following the same version pinning constraints as we do elsewhere here too 👍

liamtoozer (changes requested):
Looks good 👍  Just spotted another false-positive for `playwright`
 
I think we should ideally be keep the same pinning constraints for our newly added dependencies here, just as we do with other dependencies (bugfix & minor)?

Also, I think we should probably introduce `deptry` into our CI, now that we have done an initial pass here - I will raise a ticket for this

pyproject.toml
Farhanam76: Updated

pyproject.toml
Farhanam76: Updated

pyproject.toml
Farhanam76: I pinned it incorrectly, but it’s now updated to use the same pinning constraints as the other dependencies.

liamtoozer (approved):

EmmanuelA619 (approved):

PR #1929, opened 2026-07-01
EQS-493: Increase request limits

petechd:
We observed multiple `entity too large` errors in our environments. This is because the default limit for form data is set to 65 KB per request, and users were submitting large answers that exceeded this limit, causing 500 errors. Validation and sanitisation of form fields only occurs after the form is parsed, so we couldn't address this by changing our application logic. Instead, the werkzeug request wrapper configuration had to be adjusted through the app config.

### How to review
Check the new tests, any answers with input fields, where your answer is larger than about 65 kilobytes will cause a crash. This fixes all the `entity too large` issues we could see in our environments.

### Checklist


VirajP1002 (approved):

VirajP1002:
Works with max 120kb request, I tested with something larger and it crashed the code but it's unlikely that we'll receive a request of that size. We can always adjust if we need to 👍

app/setup.py
```diff
@@ -121,6 +121,12 @@ def create_app(  # noqa: C901  pylint: disable=too-complex, too-many-statements
 ):
     application = Flask(__name__, template_folder="../templates")
     application.config.from_object(settings)
+    application.config["MAX_CONTENT_LENGTH"] = (
```
liamtoozer: I was just checking the env vars and it looks like there is a var with the same name here: https://github.com/ONSdigital/eq-questionnaire-runner/blob/main/app/settings.py#L58. 

Not sure if that's the same var as this though? I'm not certain if its even used anyway 🤔

petechd: I checked it and it seems like changing this value to 120k bytes works, so will change it to 120 x 1024 so `122880 (bytes)` since the previous value was multiple of 1024 too (64 x 1024). And I will remove the other code.

petechd: Plus the more readable syntax introduced in newer Python so `122_880`.

liamtoozer (approved):

VirajP1002 (approved):
Re-reviewed 👍

PR #1936, opened 2026-07-07
EQS-1008: Update SDS Endpoint URL

oms09:

- Updates the runner to call the new SDS endpoint
- runner called sds endpoints on a query based param  '/v1/unit_data?dataset_id=...&identifier=...' now its using a path based endpoint '/datasets/{dataset_id}/unit-data/{identifier}'
- update the test for the new endpoint format
- refactored code to avoid repetition 

### How to review
- Run the supplementary data service unit tests (you can use poetry run pytest tests/app/services/test_request_supplementary_data.py for a specific test)
- Confirm the new SDS unit data endpoint returns 200 (you can use curl -i "http://localhost:5003/datasets/203b2f9d-c500-8175-98db-86ffcfdccfa3/unit-data/12345678901") 
- run mock sds/ runner then run specific functional test or all e.g. make test-functional-spec SPEC=supplementary_data/supplementary_data_with_list_collector.spec.js
- run launcher open supplementary_data test and confirm in terminal that new endpoint is called and get 200

### Checklist


oms09:
Needs mock sds merged in to pass func test

app/services/supplementary_data.py
```diff
@@ -58,13 +58,9 @@ def get_supplementary_data_v1(
     if not key_store.get_key(purpose=KEY_PURPOSE_SDS, key_type="private"):
         raise MissingSupplementaryDataKey
 
-    supplementary_data_url = f"{current_app.config['SDS_API_BASE_URL']}/v1/unit_data"
-
-    parameters = {"dataset_id": dataset_id, "identifier": identifier}
-
-    encoded_parameters = urlencode(parameters)
+    base_url = current_app.config["SDS_API_BASE_URL"].rstrip("/")
```
liamtoozer: I can see why it was added for extra safety, but not sure if adding `rstrip()` is worth it here. I don't think we've had an issue with a trailing `/` in our URL env vars?

app/services/supplementary_data.py
```diff
@@ -58,13 +58,9 @@ def get_supplementary_data_v1(
...
     constructed_supplementary_data_url = (
-        f"{supplementary_data_url}?{encoded_parameters}"
+        f"{base_url}/datasets/{dataset_id}/unit-data/{quote(identifier, safe='')}"
```
liamtoozer: Very minor question - we previously encoded `dataset_id` when this endpoint accepted query string params. Now we're switching to a path endpoint, is there a reason why `dataset_id` is being handled slightly differently, i.e. with no `quote()` wrapping? Only because as at this point in the Runner journey, both `dataset_id` and `identifier` these are inferred as `str` type?

liamtoozer (commented):
Changes look good - tested and works as expected 👍  Just a couple of minor questions

app/services/supplementary_data.py
oms09: good point I didn't quote wrap dataset_id because its expected to be a UUID. but yeah I'll encode both for consistency

app/services/supplementary_data.py
oms09: updated :+1:

EmmanuelA619 (approved):
Tested and checked that mock SDS endpoint returns 200.
All 15 supplementary-data service unit tests passed.
The manual supplementary-data survey launched successfully.
The functional spec started, but failed later in the journey due to the date/page-state issue, not while retrieving supplementary data. But this is because changes for mock SDS needs to be merged in

liamtoozer (approved):

PR #1959, opened 2026-08-07
Eqs 500 upgrade python version

ems-sato:

Upgrades Python from 3.13.5 to 3.14.6 (Docker image moved to `python:3.14-slim-bookworm`).

### What changed

- Bumped Python to 3.14 in `.python-version`, `pyproject.toml` (`python = "^3.14"`), `poetry.lock`, and the `Dockerfile`
- Set ruff `target-version` to `py314`
- Fixed mypy issues surfaced under 3.14 (see below)
- Reformatted code for `black` 26.x
- **Removed pylint** - approved by tech lead and team (see "Why remove pylint" below)
- Added ruff rules to cover checks pylint used to provide
- Pinned `jwcrypto` to 1.5.6 and `sdc-cryptography` to 1.2.1, since newer versions raised `InvalidJWEKeyLength` in the address lookup auth token. This is unrelated to the Python upgrade — a follow-up ticket may be needed to investigate properly

### Why remove pylint

Pylint crashes on Python 3.14 (astroid can't parse it). Pylint 4.0 fixes this, but our `pylint-absolute-imports` plugin only supports pylint <4 and hasn't been updated since 2021, so there's no way to upgrade while keeping it.

The plugin only checks one thing (no relative imports) and ruff already enforces the same rule via `TID252`, confirmed enabled and passing

Removing pylint also meant dropping the now-invalid `suggestion-mode` option from `.pylintrc`, and excluding `.venv` from the lint script (it was being scanned when the virtualenv lives inside the project directory).

Follow-up issues from removing pylint (mainly re-linting `tests/` and `scripts/`, which had been excluded from ruff) are handled separately in #1960.

### Mypy fixes

- 2 errors: `Timestamp` serialize/deserialize in `app_models.py` didn't match the marshmallow `Field` signature: fixed by widening the params to match the base class (annotation-only change)
- 3 errors: functions returning `Any` when a specific type was declared: fixed by wrapping return values in `str()`

### How to review

- Run all tests
- Docker image builds cleanly on `python:3.14-slim-bookworm`, including `uwsgi`


app/data_models/app_models.py
```diff
@@ -51,8 +52,9 @@ class Timestamp(fields.Field):
     # pylint: disable=unused-argument
     def _serialize(
         self,
-        value: datetime,
-        *args: list | None,
+        value: Any,
```
petechd: These `Any`s we should use as a last resort, according to our typehints rules (documented), especially when they had a type before. Does that even need to become `Any`? Seems to pass locally using the previous types, when using `make lint`?

app/translations/messages.pot
petechd: `messages.pot` is a template generated for static translations, why was it removed?

app/forms/field_handlers/dropdown_handler.py
```diff
@@ -27,7 +27,7 @@ def choices(self) -> Sequence[Choice]:
         ]
 
     def _get_placeholder_text(self) -> str:
-        return self.answer_schema.get("placeholder", self.DEFAULT_PLACEHOLDER)
+        return str(self.answer_schema.get("placeholder", self.DEFAULT_PLACEHOLDER))
```
petechd: This will be casting string on a string since:
placeholder will be resolved as a string, for example, in this object, it looks like that:
```
"placeholders": [
    {
        "placeholder": "person_name",
        "transforms": [
            {
                "transform": "concatenate_list",
                "arguments": {
                    "list_to_concatenate": [
                        {"source": "answers", "identifier": "first-name"},
                        {"source": "answers", "identifier": "last-name"},
                    ],
                    "delimiter": " ",
                },
            },
        ],
    },
],
```
it will be `"person_name"` string, otherwise, it will default to `lazy_gettext("Select an answer")`, which is also a string, so basically we shouldn't cast a string on a string for the sake of type hinting, we might have to:
- get to the root of the problem and fix the typehint where it happens
- type it as Any
- ignore the type hint error

petechd: Applies to all the other instances of this problem.

app/forms/field_handlers/select_handlers.py
```diff
@@ -22,7 +22,7 @@ def choices(self) -> Sequence[ChoiceType]:
 
     @property
     def dynamic_options_schema(self) -> dict[str, Any]:
-        return self.answer_schema.get("dynamic_options", {})
+        return dict(self.answer_schema.get("dynamic_options", {}))
```
petechd: Same as other instances of this, `dynamic_options` will always be a dict, even if empty, so no need to cast a dict on a dict, definition of dynamic options in validator:
```
dynamic_options": {
  "type": "object",
```

pyproject.toml
```diff
@@ -88,7 +85,7 @@ wtforms = "^3.2.1"
 htmlmin2 = "^0.1.13"
 werkzeug = "^3.1.5"
 cryptography = "==48.0.1" # Temporarily pinned - causing intermittent issues in Cloud
-jwcrypto = "^1.5.6"
+jwcrypto = "==1.5.6" # Temporarily pinned - InvalidJWEKeyLength with 1.5.8
```
petechd: 👍

pyproject.toml
```diff
@@ -193,6 +190,17 @@ extend-select = [
...
 [tool.ruff.lint.isort]
 case-sensitive = true
+
+[tool.ruff.lint.mccabe]
+max-complexity = 10
+
+[tool.ruff.lint.pylint]
+max-returns = 6
+max-branches = 12
+max-statements = 50
```
petechd: Do we want to add all of the other rules removed?
```
# Maximum number of arguments for function / method.
max-args=5

# Maximum number of attributes for a class (see R0902).
max-attributes=7

# Maximum number of boolean expressions in an if statement (see R0916).
max-bool-expr=5

# Maximum number of locals for function / method body.
max-locals=15

# Maximum number of parents for a class (see R0901).
max-parents=7

# Maximum number of public methods for a class (see R0904).
max-public-methods=20

# Minimum number of public methods for a class (see R0903).
min-public-methods=2

# Maximum cyclomatic complexity
max-complexity=10

# Maximum positional arguments
max-positional-arguments=12
```
There are more?

petechd:
Would be great if we edit the title so it follows the format.

app/translations/messages.pot
ems-sato: Yes, this was an oversight from me. I probably accidentally removed it and will restore this

app/forms/field_handlers/dropdown_handler.py
ems-sato: Yes, agreed. I will switch to `typing.cast(str, ...)` and to the same for the `dict()` instances

app/forms/field_handlers/select_handlers.py
ems-sato: Answered above

app/data_models/app_models.py
ems-sato: For this one mypy flagged it under 3.14 because the signature didnt match the marshmallow `Field` base class. Its annotation only and not a runtime change. If the old types passes for you locally could you check if `make lint` is running mypy against the 3.14 env? This error only appeared after upgrade for me.

pyproject.toml
ems-sato: These are the ones with direct ruff equivalents, set to the same thresholds as the old pylint config. The plugins absolute-imports check is covered by `TID252`
The rest of the pylint fallout is in [#1960](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1960/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) which is branched off this one

ems-sato: Actually, reviewing this I see that `max-parents` and `min-public-methods` have no ruff equivalent. Happy to raise a ticket for the remaining ones

pyproject.toml
ems-sato: Ticket created: https://officefornationalstatistics.atlassian.net/browse/EQS-1285

pyproject.toml
ems-sato: Ticket created to address the above pylint checks with no ruff equivalent: https://officefornationalstatistics.atlassian.net/browse/EQS-1286

app/forms/field_handlers/dropdown_handler.py
ems-sato: These are now solved using `typing.cast`

app/forms/field_handlers/select_handlers.py
ems-sato: These are now solved using `typing.cast`

app/forms/field_handlers/number_handler.py
```diff
@@ -53,7 +53,7 @@ def references(self) -> dict[str, Any]:
 
     @cached_property
     def max_decimals(self) -> int:
-        return self.answer_schema.get("decimal_places", 0)
+        return int(self.answer_schema.get("decimal_places", 0))
```
petechd: We might be better off just adding ignore here instead of hiding the problem by casting integer on on an integer, since `decimal places` is always an integer, since it is validated like this:
```json
"decimal_places": {
        "type": "integer",
        "description": "Number of decimal places allowed"
      },
```

app/forms/field_handlers/select_handlers.py
```diff
@@ -22,7 +22,7 @@ def choices(self) -> Sequence[ChoiceType]:
 
     @property
     def dynamic_options_schema(self) -> dict[str, Any]:
-        return self.answer_schema.get("dynamic_options", {})
+        return cast(dict[str, Any], (self.answer_schema.get("dynamic_options", {})))
```
petechd: Same as `decimal_places` add ignore here since it is already a map according to validator definition:
```json
"oneOf": [
          {
            "$ref": "../rules/operators/value/map.json"
          },
          {
            "$ref": "../value_sources.json#/any_value_source_except_location"
          }
        ]
```

app/forms/field_handlers/dropdown_handler.py
```diff
@@ -27,7 +27,9 @@ def choices(self) -> Sequence[Choice]:
         ]
 
     def _get_placeholder_text(self) -> str:
-        return self.answer_schema.get("placeholder", self.DEFAULT_PLACEHOLDER)
+        return cast(
+            str, self.answer_schema.get("placeholder", self.DEFAULT_PLACEHOLDER)
```
petechd: I would remove this, since it's just for the sake of type checks, since we validate if schema has this field as a string in validation, same as the other two. In this example, the validation definition is:
```json
"placeholder": {
        "description": "The placeholder text to be used for the select field.",
        "$ref": "../common_definitions.json#/non_empty_string"
      },
```

pyproject.toml
```diff
@@ -83,12 +80,12 @@ pdfkit = "^1.0.0"
 ordered-set = "^4.1.0"
 cachetools = "^6.2.2.0"
 gevent = "^25.9.1"
-babel = "==2.14.0"  # Temporarily pinned - problem for translations found in v2.15.0, see: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1384
+babel = "^2.18.0"  # Temporarily pinned - problem for translations found in v2.15.0, see: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1384
```
petechd: So this comment is not needed anymore?

app/jinja_filters.py
```diff
@@ -77,6 +77,17 @@ def format_unit(
         locale=flask_babel.get_locale(),
     )
 
+    # Some locales (e.g. cy) have no long-form plural patterns in CLDR for
+    # certain units, in which case Babel returns the raw unit key. Fall back
+    # to the short form, which carries the correct translation.
+    if formatted_unit.endswith(unit):
```
petechd: Probably worth changing the logic so we only run `custom_format_unit` once here instead of twice, like in the Census repo [lines 77-111](https://github.com/ONSdigital/census31-eq-questionnaire-runner/blob/a04e08b5e5e3cdf296adf651c6710ceb3a972b6f/app/utilities/decimal_places.py#L77). Something similar to:

```python
if length != "short" and measurement_unit in formatted_unit:
        formatted_unit = units.format_unit(
            value=value,
            measurement_unit=measurement_unit,
            length="short",
            # Type ignore: babel function has incorrect type hinting, NumberPattern is valid here
            format=number_format,  # type: ignore
            locale=locale,
        )
```

.pylintrc
```diff
@@ -1,613 +0,0 @@
...
-# for backward compatibility.)
-extension-pkg-whitelist=
-
-# Return non-zero exit code if any of these messages/categories are detected,
-# even if score is above --fail-under value. Syntax same as enable. Messages
-# specified are enabled, while categories only check already-enabled messages.
-fail-on=
-
-# Specify a score threshold under which the program will exit with error.
-fail-under=10
```
liamtoozer: I'm not sure of the implications of losing this rule is? Might be okay, but does `ruff` has any equivalent rule?

.pylintrc
```diff
@@ -1,613 +0,0 @@
...
-                  _source,
-                  _make
-# List of valid names for the first argument in a class method.
-valid-classmethod-first-arg=cls
-
-# List of valid names for the first argument in a metaclass class method.
-valid-metaclass-classmethod-first-arg=mcs
-
-
-[DESIGN]
```
liamtoozer: We are losing a lot of these linting checks in this `DESIGN` block like `max-args=5`, `max-attributes=7` etc by removing this file. I think we will need to carry these over into ruff?

pyproject.toml
```diff
@@ -83,12 +80,12 @@ pdfkit = "^1.0.0"
...
 wtforms = "^3.2.1"
 htmlmin2 = "^0.1.13"
 werkzeug = "^3.1.5"
 cryptography = "==48.0.1" # Temporarily pinned - causing intermittent issues in Cloud
-jwcrypto = "^1.5.6"
+jwcrypto = "==1.5.6" # Temporarily pinned - InvalidJWEKeyLength with 1.5.8
```
liamtoozer: I _think_ this is a transitive dependency from the `sdc-cryptography` repo, so maybe it can just be removed? If so, does this also mean we might not need to pin the `sdc-cryptography` dependency?

https://github.com/ONSdigital/sdc-cryptography/blob/master/pyproject.toml#L28C9-L28C10

.pylintrc
liamtoozer: Ah ignore me, I can see that the defaults are already set in the `ruff` rulesets for these so there was no need for any overrides 👍

app/forms/field_handlers/number_handler.py
ems-sato: Done, replaced `int()` cast with `# type:ignore[no-any-return]`

app/forms/field_handlers/select_handlers.py
ems-sato: Done, same as `decimal_places`

app/forms/field_handlers/dropdown_handler.py
ems-sato: Done, removed the cast and added `# type:ignore[no-any-return]`

pyproject.toml
ems-sato: Comment removed

app/jinja_filters.py
ems-sato: Moved fallback into `custom_format_unit` following the Census implementation. It reuses the same `number_format` and only retries when the length isnt already short. `jinja_filters.py` is back to a single call.

.pylintrc
ems-sato: Ruff doesnt use scores so any violation of a selected rule exits non-zero, so the behaviour should be equivalent. The only gap is the two checks with no ruff equivalent (max-parents, min-public-methods), I raised this ticket to tackle them: https://officefornationalstatistics.atlassian.net/browse/EQS-1286

pyproject.toml
ems-sato: I had a look, and Runner imports jwcrypto directly (`session_store.py`, `storage_encryption.py`, `address_lookup_api_helper.py`), so I've kept it as a direct dependency. Like you pointed out sdc-cryptography declares `jwcrypto = "*"`, so it doesn't constrain the version. Without this pin, 1.5.8 would resolve and the InvalidJWEKeyLength error would come back. For the same reason, the sdc-cryptography pin is independent of this one, and its 1.3.0 issue needs looking at separately. Raised a ticket for this too: https://officefornationalstatistics.atlassian.net/browse/EQS-1285

liamtoozer: Sounds good, thanks for raising a ticket for this 👍

liamtoozer (approved):

PR #1960, opened 2026-08-12
Eqs 500 ruff tests scripts

ems-sato:
Branched out of [EQS-500-Upgrade-Python-version](https://github.com/ONSdigital/eq-questionnaire-runner/tree/EQS-500-Upgrade-Python-version)
Removes the ruff exclude for tests/ and scripts/ so those directories are linted again, findings resolved

### Why
Pylint was dropped and it ran over the entire repo via find. Ruff was configured to skip tests/ and scripts/ which left those directories with no python linting

### Findings
Removing the exclusion surfaced 1708 findings across 280 files. Resolved as:

- 1574x `S101` (`assert` in tests): ignored via `per-file-ignores`. The rule exists because `assert` is stripped under `python -O`
- 68x auto-fixed by `ruff check --fix` 
- ~40 ignored by path : `LOG015`/`S603`/`S607` in `scripts/` (standalone CLI tools), `DTZ001`/`DTZ007` in rules tests (fixed reference dates for weekday arithmetic), `E501` in five files containing encrypted token fixtures and long expected URLs.
- 8 threshold suppressions (`C901`, `PLR0912`, `PLR0915`) translated from existing `# pylint: disable=` to `# noqa`.
- Fixed rather than suppressed: three leftover debug `print`s in `test_questionnaire_question_variants.py`; a shadowed `TEST_SDS_URL` import in `test_session.py`; an unused loop variable; an f-string exception message.

### Notes for review
- Ruff has no equivalent for pylint's inference-based checks (`no-member`, `undefined-variable`). Mypy covers most of that ground.
- `tests/functional/generate_pages.py` carries large suppressions inherited from pylint. Out of scope here, but  might need its own ticket.
- Dead `# pylint: disable=` comments remain across the test suite where ruff doesn't flag the same line. Left for a follow-up (~40 files).


app/forms/field_handlers/dropdown_handler.py
```diff
@@ -27,7 +27,7 @@ def choices(self) -> Sequence[Choice]:
         ]
 
     def _get_placeholder_text(self) -> str:
-        return self.answer_schema.get("placeholder", self.DEFAULT_PLACEHOLDER)
+        return str(self.answer_schema.get("placeholder", self.DEFAULT_PLACEHOLDER))
```
petechd: I would remove this, since it's just for the sake of type checks, since we validate if schema has this field as a string in validation, same as the other two. In this example, the validation definition is:
```json
"placeholder": {
        "description": "The placeholder text to be used for the select field.",
        "$ref": "../common_definitions.json#/non_empty_string"
      },
```

app/forms/field_handlers/number_handler.py
```diff
@@ -53,7 +53,7 @@ def references(self) -> dict[str, Any]:
 
     @cached_property
     def max_decimals(self) -> int:
-        return self.answer_schema.get("decimal_places", 0)
+        return int(self.answer_schema.get("decimal_places", 0))
```
petechd: We might be better off just adding ignore here instead of hiding the problem by casting integer on on an integer, since `decimal places` is always an integer, since it is validated like this:
```json
"decimal_places": {
        "type": "integer",
        "description": "Number of decimal places allowed"
      },
```

app/forms/field_handlers/select_handlers.py
```diff
@@ -22,7 +22,7 @@ def choices(self) -> Sequence[ChoiceType]:
 
     @property
     def dynamic_options_schema(self) -> dict[str, Any]:
-        return self.answer_schema.get("dynamic_options", {})
+        return dict(self.answer_schema.get("dynamic_options", {}))
```
petechd: Same as `decimal_places` add ignore here since it is already a map according to validator definition:
```json
"oneOf": [
          {
            "$ref": "../rules/operators/value/map.json"
          },
          {
            "$ref": "../value_sources.json#/any_value_source_except_location"
          }
        ]
```

app/forms/field_handlers/field_handler.py
```diff
@@ -49,7 +49,7 @@ def label(self) -> str | None:
 
     @cached_property
     def guidance(self) -> str:
-        return self.answer_schema.get("guidance", "")
+        return str(self.answer_schema.get("guidance", ""))
```
petechd: Same as the other (similar) comments.

pyproject.toml
petechd: No babel upgrade in this PR? If we do, check this comment: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1959/changes#r4059954684

petechd:
Do we enable `no_self_use - Ruff equivalent (PLR6301)` in this PR? Otherwise, we’re risking forgetting about `no_self_use`. Doesn't seem to be on the `Reinstate` task.

app/jinja_filters.py
```diff
@@ -465,7 +465,7 @@ def __init__(self, text: str, other: str | None = None) -> None:
 
 
 class SummaryRowItem:
-    def __init__(  # noqa: C901 pylint: disable=too-complex, too-many-branches
+    def __init__(  # noqa: C901, PLR0912
```
petechd: PLR0912 is `too-many-branches` but what happened to the `too-complex` check? Is that the `McCabe` check on another task?

tests/app/storage/conftest.py
```diff
@@ -42,7 +42,7 @@ def dynamodb():
 @pytest.fixture
 def mock_client(mocker):
     client = mocker.Mock()
-    client.transaction.return_value = contextlib.suppress()
+    client.transaction.return_value = contextlib.suppress()  # noqa: B022
```
petechd: Can we use an argument in `contextlib.suppress()` to make this `noqa` go away?

tests/app/mock_data_store.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, **kwargs):
 
     # pylint: disable=no-self-use
     def transaction(self):
-        return contextlib.suppress()
+        return contextlib.suppress()  # noqa: B022
```
petechd: Can we use `contextlib.nullcontext()` instead? To get rid of `# noqa: B022`?

tests/app/questionnaire/test_value_source_resolver.py
```diff
@@ -59,7 +59,7 @@ def get_calculation_block(
 def get_value_source_resolver(
     schema: QuestionnaireSchema = None,
     data_stores: DataStores = None,
-    location: Location | RelationshipLocation = Location(
+    location: Location | RelationshipLocation = Location(  # noqa: B008
```
petechd: Better to just use the recommended fix for this from the official doc?:
https://docs.astral.sh/ruff/rules/function-call-in-default-argument/
```python
def create_list() -> list[int]:
    return [1, 2, 3]
```


```python
def mutable_default(arg: list[int] = create_list()) -> list[int]:
    arg.append(4)
    return arg
```

fixed:
```python
def better(arg: list[int] | None = None) -> list[int]:
    if arg is None:
        arg = create_list()

    arg.append(4)
    return arg
```
The problem is the Location assignment in args here.

tests/app/parser/test_supplementary_data_parser.py
```diff
@@ -13,7 +13,7 @@ def not_raises(exception):
     try:
         yield
     except exception as validation_error:
-        raise pytest.fail(f"{validation_error} RAISED")
+        raise pytest.fail(f"{validation_error} RAISED")  # noqa: EM102, B904
```
petechd: Creating the f string a step above would fix it? Same for B904, using `from`? https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

tests/app/questionnaire/rules/test_rule_evaluator.py
```diff
@@ -46,7 +46,7 @@ def get_rule_evaluator(
     language="en",
     schema: QuestionnaireSchema = None,
     data_stores: DataStores = None,
-    location: Location | RelationshipLocation = Location(
+    location: Location | RelationshipLocation = Location(  # noqa: B008
```
petechd: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1960/changes#r4060158078

tests/integration/test_app_create.py
```diff
@@ -339,7 +339,7 @@ def test_setup_dynamodb(self):
     def test_invalid_storage(self):
         self._setting_overrides["EQ_STORAGE_BACKEND"] = "invalid"
 
-        with self.assertRaises(Exception):
+        with self.assertRaises(Exception):  # noqa: B017
```
petechd: Make exception more specific? 
```python
self.assertRaises(SomeSpecificException, foo)
```
ValueError? KeyError?

tests/app/views/contexts/test_submitted_response_context.py
```diff
@@ -126,7 +126,7 @@ def test_view_submitted_response_expired(
 def test_build_view_submitted_response_no_submitted_at(app: Flask):
     with app.app_context():
         questionnaire_store = fake_questionnaire_store({}, None)
-        with pytest.raises(Exception):
+        with pytest.raises(Exception):  # noqa: B017
```
petechd: Same as the other comment in this PR, what kind of specific exception? Can be checked in debug.

tests/app/parser/test_supplementary_data_parser.py
petechd: EM102 can be fixed by moving the f-string out: https://docs.astral.sh/ruff/rules/f-string-in-exception/.

tests/functional/generate_pages.py
```diff
@@ -1110,7 +1108,7 @@ def append_spec_page_import(context, spec_file):
                 template_spec.write(SPEC_EXAMPLE_TEST.substitute(schema_name))
     else:
         if os.path.isdir(args.SCHEMA):
-            for root, dirs, files in os.walk(args.SCHEMA):
+            for root, _dirs, files in os.walk(args.SCHEMA):
```
petechd: I know this is  a convention thing, but not to get people confused, just make `_dirs` a `_` so it is clear it's not used anywhere,  name with a `_` prefix indicates a private variable.

app/translations/messages.pot
```diff
@@ -1,1541 +0,0 @@
-# Translations template for PROJECT.
```
liamtoozer: This is used for translations and we won't want to clear this file out

tests/app/forms/test_custom_fields.py
```diff
@@ -67,7 +67,7 @@ def test_integer_field_inputs(mock_form, number_input, result):
 @pytest.mark.parametrize(
     "number_input, result",
     [
-        ("1_1,0", Decimal("110")),
+        ("1_1,0", Decimal(110)),
         ("1.10", Decimal("1.1")),
```
liamtoozer: Not sure why the other strings weren't pick up by the linter here? 🤔

tests/functional/generate_pages.py
```diff
@@ -564,8 +564,7 @@ def process_content(context, page_spec):
     page_spec.write(CONTENT_ITEM_GETTER.safe_substitute(context))
 
 
-# pylint: disable=too-many-locals
-def write_summary_spec(
+def write_summary_spec(  # noqa: C901
```
liamtoozer: Very minor, and no worries if this has already been tried, but `PLR0914` might be the better fit here? https://docs.astral.sh/ruff/rules/too-many-locals/

app/forms/field_handlers/dropdown_handler.py
ems-sato: Already fixed in #1959, which this branch is now based on

app/forms/field_handlers/number_handler.py
ems-sato: Already fixed in #1959, which this branch is now based on

app/forms/field_handlers/select_handlers.py
ems-sato: Already fixed in #1959, which this branch is now based on

pyproject.toml
ems-sato: Babel is upgraded to `^2.18.0` in #1959, which this branch is now based on

app/translations/messages.pot
ems-sato: Good catch. It is restored and matches the regenerated version from #1959

ems-sato:
> Do we enable `no_self_use - Ruff equivalent (PLR6301)` in this PR? Otherwise, we’re risking forgetting about `no_self_use`. Doesn't seem to be on the `Reinstate` task.

PLR6301 is a preview rule in ruff, so enabling it would mean turning on preview mode for this one check. I have added to this ticket: https://officefornationalstatistics.atlassian.net/browse/EQS-1286, alongside max-parents and min-public-methods so it isn't forgotten. I have also removed the stale `# pylint: disable=no-self-use` in `mock_data_store.py`.

app/forms/field_handlers/field_handler.py
ems-sato: Done, replaced with `type: ignore[no-any-return]`, the same as others in #1959

app/jinja_filters.py
ems-sato: `C901` is ruff's McCabe complexity rule, the equivalent of pylint's `too-complex`, so both original checks are still covered: `C901` for too-complex and `PLR0912` for too-many-branches.

tests/app/storage/conftest.py
ems-sato: Done, switched to `contextlib.nullcontext()`

tests/app/mock_data_store.py
ems-sato: Done, switched to contextlib.nullcontext()

tests/app/questionnaire/test_value_source_resolver.py
ems-sato: Used the other fix from the ruff docs: a module-level `DEFAULT_LOCATION` constant as the default. The `None` version broke two tests that pass `location=None`.

tests/app/questionnaire/rules/test_rule_evaluator.py
ems-sato: Done, same as `test_value_sorce_resolver.py` a module-level `DEFAULT_LOCATION` constant as the default.

tests/app/parser/test_supplementary_data_parser.py
ems-sato: Done, moved the f-string into a `msg` variable and added `from validation_error`, so both noqas are gone

tests/integration/test_app_create.py
ems-sato: Done, narrowed it to `NotImplementedError`, which is what `create_app` raises for an unknown `EQ_STORAGE_BACKEND`

tests/app/views/contexts/test_submitted_response_context.py
ems-sato: Done, narrowed it to `TypeError`.

tests/functional/generate_pages.py
ems-sato: Done, rename to `_`

tests/functional/generate_pages.py
ems-sato: Thank you for spotting this! I have checked and `PLR0914` is a preview rule, so ruff doesn't enforce it without preview mode. `C901` is what ruff actually raises here (complexity 11 > 10). I've added too-many-locals to https://officefornationalstatistics.atlassian.net/browse/EQS-1286 with the other pylint checks that ruff doesnt currently enforce.

tests/app/forms/test_custom_fields.py
ems-sato: That's `FURB157` (verbose-decimal-constructor). It only flags strings that are whole integers, where `Decimal(110)` is exact and the string is unnecessary. The rule deliberately skips quoted floats like `Decimal("1.1")`, because Decimal(1.1) would pick up float imprecision. The [rule docs ](https://docs.astral.sh/ruff/rules/verbose-decimal-constructor/) call this out.

tests/functional/generate_pages.py
liamtoozer: Ah okay great, happy to skip as it's preview-only, thanks!

tests/app/forms/test_custom_fields.py
liamtoozer: Makes sense, thanks

liamtoozer (approved):

PR #1963, opened 2026-08-12
Add Playwright integration

petechd:
Based on the PR in Census Runner: https://github.com/ONSdigital/census31-eq-questionnaire-runner/pull/82 by @ajmaddaford. This is functional tests migration to Playwright and Typescript. Like in the original PR, we preserve the existing behaviour while changing the automation framework. Compared to the Census Runner PR, these extra specs had to be added, since they didn't exist at the time of the fork:

- tests/functional/spec/supplementary_data/supplementary_data.spec.js
- tests/functional/spec/hub_and_spoke/hub_and_spoke_supplementary_data.spec.js
- tests/functional/spec/launch_with_cir.spec.js

In the original PR some adjustments were made to prevent flakiness. So some tests slightly changed but the assertions should be the same.

### How to review
Visually check the specs and reworked Makefile, generate_pages, pull_request.yaml (are all actions needed). Check if the Webdriver was completely removed. Use Andrew's PR as reference: https://github.com/ONSdigital/census31-eq-questionnaire-runner/pull/82.

### Checklist


.github/workflows/pull_request.yml
```diff
@@ -190,7 +194,7 @@ jobs:
         env:
          SPECS: ${{ env.SPECS }}
       - name: Run functional tests
-        run: make test-functional
+        run: npx playwright test --reporter list,html
```
VirajP1002: Minor, not sure if the `html` reporter is needed here? Since it generates a html report at the end? Do we need that in Actions or maybe in one of our make commands. 

Also, couldn't the make command `make test-functional-headless` be used here?

petechd: it's GitHub Actions (build) so we want to keep it and run in a non headless mode to mimic the real behaviour? Html would be useful if we want to upload and view the artifacts, with the test runs results, but for now I guess we can remove it.

.github/workflows/pull_request.yml
```diff
@@ -190,7 +194,7 @@ jobs:
         env:
          SPECS: ${{ env.SPECS }}
       - name: Run functional tests
-        run: make test-functional
+        run: npx playwright test --reporter list
```
ems-sato, line 197: Not a blocker so not sure if needed for now but playwright.config.ts line 34 sets `trace: ci ? 'on-first-retry' : 'retain-on-failure'`, so traces are being written to test-results/ and then thrown away at teardown. I don't think we will have a way to debug a shard, might be nice to add an upload step with `if: failure()` named by shard?

Makefile
ems-sato: generate-spec is still present and still passes tests/functional/spec/$(SCHEMA).spec.js, but generate_pages.py has dropped --spec_file. Census team (PR 82) removed generate-spec entirely, should we carry the deletion over here?

ems-sato (commented):
Nice work, I think it looks really good. I added a couple of comments and I will try to go through it again as I might have missed more.
Another point I was going to add would be if it would be really helpful if we could update the README to reflect the tests changes and match with the new workflow.

tests/functional/spec/summaries/grand_calculated_summary/grand_calculated_summary_overlapping_answers.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      const calculatedSummary4Page = new CalculatedSummary4Page(page)
+      const grandCalculatedSummaryShoppingPage = new GrandCalculatedSummaryShoppingPage(page)
+      const hubPage = new HubPage(page)
+      const introductionBlockPage = new IntroductionBlockPage(page)
+      const section1SummaryPage = new Section1SummaryPage(page)
+      await openQuestionnaire('test_grand_calculated_summary_overlapping_answers.json')
+      await introductionBlockPage.submit().click()
+
+      // grand calculated summary should not be enabled until section-1 complete
+      await expect(hubPage.summaryRowLink('section-3')).toBeHidden()
```
ems-sato, line 41: @Farhanam76 and I noticed that on main this was `isExisting()).toBe(false)`. `toBeHidden()` covers that (it passes when the element is absent too) but also passes when it is present and invisible, so it is slightly looser. Now, just wondering about consistency: the suite uses `toBeHidden()` in some places and `toHaveCount(0)` in others (e.g. radio_options_from_checkbox_answers.spec.ts) for the same original pattern. Is there a rule for which, or should we pick one pattern?

petechd: As per the migration guide: https://github.com/ONSdigital/census31-eq-questionnaire-runner/blob/f74d32ff906e1c169db8561bdf1991feb555f86e/doc/playwright-migration.md?plain=1#L22,  we use `.toBeHidden` for visibility checks now, probably due to Andrew also trying to `Stabilise historically flaky journeys`. `toHaveCount(0)` probably because there could be multiple elements on a list, since some elements can have a list (of children). You would need to check the (generated) page template for details (which elements can be a list). I prefer `toBeHidden` since we only care about what user can see, for our functional tests, not what exists on page's `html` or in the underlying logic, these are two different things. We only care about user here and `toBeHidden` means `Element is not visible` so that works well.
What I was thinking about is, we have two properties in CSS to hide an element on the page, `visibility` and `display`. To make something invisible (for the user), one can have value `hidden` and the other `none`. When we use the `display: none` element is removed from the layout but using `visibility: hidden` would mean it still exists so `isExisting()).toBe(true)` would still pass but element wouldn't be visible for the user. Same if we check `isExisting()).toBe(false)`, this check for `visibility: hidden` would fail but pass for `display: none`. So I prefer `toBeHidden()`.

Makefile
petechd: Since it's much faster to just generate all pages for all specs, I would remove it, I didn't even know this command existed and never used it. Removing now

.github/workflows/pull_request.yml
petechd: It's more for local runs, where we want to check the test-results before we push the changes up? In theory everything is supposed to run correctly without any flakiness before we commit the changes.

tests/functional/spec/summaries/grand_calculated_summary/grand_calculated_summary_overlapping_answers.spec.ts
petechd: toBeHidden():
- Ensures that locator either does not resolve to any DOM node, or resolves to a non-visible
- returns true if the element exists in the DOM but is not visible (e.g. display: none, visibility: hidden, opacity: 0)

await el.isExisting():
- Element exists in the DOM (visible or not)

.github/workflows/pull_request.yml
ems-sato: Fair point for local. My only thought was CI-only failures... `retries: 2` is CI-only so a shard that passes locally but fails on CI will be hard to debug. But again, not a blocker and if if it's something we want to look into, it could be done in a follow up ticket.

tests/functional/spec/summaries/grand_calculated_summary/grand_calculated_summary_overlapping_answers.spec.ts
ems-sato: Yes, that makes sense thank you. The `visibility: hidden` is a good point, `isExisting()).toBe(false)` would actually fail even though the user can't see the element. Agree `toBeHidden()` is the right default for functional tests and leave the `toHaveCount(0)` cases as they are, given they are about list children

tests/functional/spec/jwt_helper.ts
```diff
@@ -0,0 +1,270 @@
+import KJUR from 'jsrsasign'
```
ems-sato, line 1: `jwt_helper.ts` exists at both `tests/functional/` and `tests/functional/spec/`, and these have drifted, so the top-level one has `cirInstrumentId` and types `schema` as `string|null` and the spec one has neither. The Census PR removed the one under `spec`, should we do the same here?  Same for `helpers.ts` although for that one, both are files seems identical

package.json
```diff
@@ -12,50 +12,42 @@
...
-        "@babel/runtime": "7.29.2",
-        "@wdio/cli": "9.27.0",
-        "@wdio/local-runner": "8.46.0",
-        "@wdio/mocha-framework": "9.27.0",
-        "@wdio/spec-reporter": "9.27.0",
+        "@eslint/js": "8.57.1",
+        "@playwright/test": "^1.61.1",
+        "@typescript-eslint/eslint-plugin": "8.65.0",
+        "@typescript-eslint/parser": "8.65.0",
+        "@typescript/native": "npm:typescript@^7.0.2",
```
ems-sato, line 27: This might be intentional but just to double check: `@typescript/native` resolves to plain `typescript` here, and line 44 has `typescript` resolving to `@typescript/typescript6`, is this the right way around?

package.json
```diff
@@ -12,50 +12,42 @@
...
-        "@babel/plugin-transform-runtime": "7.29.0",
-        "@babel/preset-env": "7.29.2",
-        "@babel/register": "7.28.6",
-        "@babel/runtime": "7.29.2",
-        "@wdio/cli": "9.27.0",
-        "@wdio/local-runner": "8.46.0",
-        "@wdio/mocha-framework": "9.27.0",
-        "@wdio/spec-reporter": "9.27.0",
+        "@eslint/js": "8.57.1",
+        "@playwright/test": "^1.61.1",
```
ems-sato, line 24: Main pinned `typescript` at `6.0.2`, but lines 24, 27, and 44 all use `^` now. `@playwright/test` is the one I am thinking about because `npx playwright install` would download browsers matching whatever version gets installed, so a drift could get flakes that are hard to reproduce. Is it worth pinning it?

tests/functional/generate_pages.py
```diff
@@ -55,263 +48,269 @@
 
 SPEC_EXAMPLE_TEST = Template(r"""
 describe("Example Test", () => {
```
ems-sato, line 50: Just noticed these and they look unreachable now that generate-spec is gone, looks safe to clean these up as well. Same for `append_spec_page_import` and it's 4 call sites, all inside `if spec_file:` guards.

petechd: The `SPEC_PAGE_IMPORT` is still used? For example, `process_view_submitted_response(data, require_path, out_dir, spec_file)` runs with a spec_file so that `if spec_file:` still applies. Unless `process_schema`'s `spec_file` is always `None`? I can see it is always `None`. But I agree, `SPEC_PAGE_HEADER` and `SPEC_EXAMPLE_TEST` can go.

.github/workflows/pull_request.yml
petechd: If something passes on a retry then we don't want to read the debug logs as it is only flaky but not always breaking? If something is flaky we can try it locally without retries to investigate further? I guess, if something works on a retry in build then it is acceptable level of flakiness?

package.json
petechd: If I remember correctly, I kept it like that to spare the reviewers from checking all the linting issues fixes/changes (eslint) as this PR would grow exponentially and it's already too large. I wanted to update linting in another ticket just to keep the size of this PR down.

package.json
petechd: Typescript is also ^6.0.2? Only eslint is worth keeping pinned and updated as part of some next task.

tests/functional/generate_pages.py
petechd: Ok, removed all, including any mention of redundant `spec_file`.

package.json
ems-sato: Yes, that makes sense. Thanks for explaining, I completely agree with it

.github/workflows/pull_request.yml
ems-sato: Fair enough, that is reasonable. Happy to leave it as it is

package.json
ems-sato: It was a nitpicky comment and fair on Typescript, happy to keep for another task.

ems-sato (approved):
LGTM, one last comment from me is that it would be good to update README  to reflect the tests changes. Not a blocker so happy to approve.

petechd:
Had to remove the division into projects in the .ts file (and README mention) as it was not compatible with shards, run all suites for every worker.

Farhanam76 (approved):

liamtoozer (approved):
I've eyeballed the changes as much as I can, run the tests locally altogether and in isolation - all looking great 👍

EmmanuelA619 (approved):

oms09 (approved):

tests/functional/base_pages/calculated-summary.page.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  calculatedSummaryQuestion (): Locator {
+    return this.locator('[data-qa=calculated-summary-question]')
+  }
+
+  calculatedSummaryAnswer (): Locator {
+    return this.locator('[data-qa=calculated-summary-answer]')
+  }
+
+  summaryItems (): Locator {
```
VirajP1002: I don't think this method is used anywhere, might be worth removing?

tests/functional/base_pages/confirmation-email-sent.page.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  sendAnotherEmail (): Locator {
+    return this.locator('a[id="send-another-email"]')
+  }
+
+  feedback (): Locator {
+    return this.locator('.ons-feedback')
+  }
+
+  feedbackLink (): Locator {
```
VirajP1002: Same as above, I don't think this method is used anywhere?

tests/functional/base_pages/calculated-summary.page.ts
VirajP1002: Looks like it might be needed since it's mentioned in tests, for some reason the IDE shows no usage

VirajP1002 (approved):

PR #1967, opened 2026-08-28
Update README for corporate macbook

ems-sato:
As part of the Corporate MacBooks onboarding we had changes to the repo setup as homebrew is not longer available. This updates the the README so the local setup instructions work on the machine, using Miniconda + conda-forge for packages and Podman as the container runtime

### How to review
The following sections of the README were updates:
- Pre-requisites
- Setup
- Run with Docker
- Installing dependencies

Retrace the new steps on a corporate Macbook and confirm the setup works from a clean start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,21 +11,44 @@
 
 ## Run with Docker
 
-Install [Docker](https://www.docker.com/) for your system. Make sure that you've installed both docker and docker-compose packages, preferably using Homebrew:
+Install Podman for your system as the container runtime.
+
+Make sure the Podman machine started every time you want to use container images:
 
 ``` shell
-brew install docker
-brew install docker-compose
+podman machine start
 ```
 
-On MacOS install container runtimes, eg. [Colima](https://github.com/abiosoft/colima):
+The Makefile invokes `docker compose`. Podman is API-compativle, so provide a `docker` command that points at it:
+
 ```shell
-brew install colima
+mkdir -p ~/.local/bin
+ln -s "$(which podman)" ~/.local/bin/docker
+hash -r
 ```
 
-Make sure Colima is started every time you want to use Docker images:
+`~/.local/bin` must be on your `PATH`. Verify both the command and compose support:
+
 ```shell
-colima start
+docker --version
+docker compose version
+```
+
+If compose is unavailable, install a provider:
+
+``` shell
+conda install -c conda-forge podman-compose
+```
+
+NOTE - approach not yet settled. The symlink above is the documented default because it applies to every Make recipe, including those that call `docker build`.
+An alternative under discussing on changing the Makefile to call `docker-compose` instead of `docker compose`.
+This avoids per-machine setup but only covers the compose recipes, `docker build` targets would still need the symlink.
+
+If change is adopted, confirm the standalone binary is present:
+
+``` shell
+which docker-compose
+docker-compose version
 ```
 
 To get eq-questionnaire-runner running the following command will build and run the containers
@@ -61,11 +84,49 @@ git clone git@github.com:ONSdigital/eq-questionnaire-runner.git
 
 ### Pre-Requisites
 
-In order to run locally you'll need Node.js, snappy, pyenv, jq and wkhtmltopdf installed
+The following must be installed and working before you start:
+- Miniconda: Python, node and system package management (install from Self Service)
+- Podman: Container runtime for supporting services (machine created and running)
+- wkhtmltopdf: PDF generation (installed separately, see below)
+- gcloud: Pulling images from Google Artifact Registry
+
+Python, Node.js, Poetry, snappy and jq are all provided by the conda environment created in setup
+Verify each is available
 
 ``` shell
-brew install snappy npm pyenv jq wkhtmltopdf
+conda --version
+podman --version
+gcloud --version
+make --version
 ```
+
+If `conda` reports `command not found` after installing from Self Service, the installer did not write the conda block into `~/.zshrc`/
+Confirm the install is present and wire it in:
+
+``` shell
+ls -d /opt/miniconda3
+/opt/miniconda3/bin/conda init zsh
+```
+
+Open a new terminal tab and re-check `conda --version`
+
+Make sure that the Podman machine is running:
+
+``` shell
+podman machine list
+podman machine start
+```
+
+`wkhtmltopdf` is nor reliably available on conda-forge for macOS ARM, so it is installed outside the conda environment.
+Download the macOS `.pkg` from the wkhtmltopdf downloads page and run the installer.
+Note that wkhtmltopdf is an archived project and no longer receives updates.
+Check if it is correctly installed:
+
+``` shell
+which wkhtmltopdf
+```
+
+This should return a path (typically `/usr/local/bin/wkhtmltopdf)
 
 ### Setup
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -82,48 +139,77 @@ echo "local" > .application-version
...
+
+
+> Note: conda-forge does not publish every Node patch release. Where the exact `.nvmrc` version is unavailable, pin the closest available patch below it and note the substitution in `environment.yml`.
+
+If `.python-version` or `.nvmrc` change, update `environment.yml` to match.
+
+Create and activate the environment:
+
+``` shell
+conda env create -f environment.yml
```
petechd: This command has the `-n` attribute missing? 
```
conda create -n eq-questionnaire-runner -f environment.yml
```

README.md
```diff
@@ -82,48 +139,77 @@ echo "local" > .application-version
...
-```shell
-export PYENV_ROOT="$HOME/.pyenv"
 
-command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"
+Version can be changed by editing `environment.yml` and running `conda env update -f environment.yml --prune`
 
-eval "$(pyenv init -)"
+#### Poetry
+
+Poetry must install into the conda environment rather than creating its own virtualenv.
```
petechd: Maybe some mention how to install `poetry`? Since we explain how to install everything else?

README.md
```diff
@@ -82,48 +139,77 @@ echo "local" > .application-version
...
+These flags now apply automatically whenever the environment is active.
+
+#### Dependencies
+
+With the environment active, install the Python dependencies:
 
 ``` shell
-curl -sSL https://install.python-poetry.org | python3 - --version 2.1.2
-poetry self add poetry-plugin-dotenv
 poetry install
```
petechd: Maybe something about `poetry update`?

README.md
```diff
@@ -61,12 +80,50 @@ git clone git@github.com:ONSdigital/eq-questionnaire-runner.git
 
 ### Pre-Requisites
 
-In order to run locally you'll need Node.js, snappy, pyenv, jq and wkhtmltopdf installed
+The following must be installed and working before you start:
+- Miniconda: Python, node and system package management (install from Self Service)
+- Podman: Container runtime for supporting services (machine created and running)
+- wkhtmltopdf: PDF generation (installed separately, see below)
+- gcloud: Pulling images from Google Artifact Registry
```
petechd: It's not in Self Service as a package so we should mention it should be installed from source?

README.md
```diff
@@ -61,12 +80,50 @@ git clone git@github.com:ONSdigital/eq-questionnaire-runner.git
...
+Open a new terminal tab and re-check `conda --version`
+
+Make sure that the Podman machine is running:
+
+``` shell
+podman machine list
+podman machine start
+```
+
+`wkhtmltopdf` is nor reliably available on conda-forge for macOS ARM, so it is installed outside the conda environment.
```
petechd: A typo here.

README.md
```diff
@@ -11,27 +11,46 @@
...
+Make sure the Podman machine started every time you want to use container images:
 
 ``` shell
-brew install docker
-brew install docker-compose
+podman machine start
 ```
 
-On MacOS install container runtimes, eg. [Colima](https://github.com/abiosoft/colima):
+Podman is API-compatible with Docker, so provide a `docker` command that points at it. This is still needed for the plain `docker run` commands used later in this README (e.g. running launcher, SDS, CIR images):
```
petechd: Any need for `alias docker=podman` anymore in `.zshrc`? Or would it only confuse us?

Makefile
```diff
@@ -106,13 +106,13 @@ run-uwsgi-async: link-development-env
 	WEB_SERVER_TYPE=uwsgi-async poetry run ./run_app.sh
 
 dev-compose-up:
-	docker compose -f docker-compose-dev.yml pull eq-questionnaire-launcher
-	docker compose -f docker-compose-dev.yml pull sds
-	docker compose -f docker-compose-dev.yml pull cir
-	docker compose -f docker-compose-dev.yml up -d
+	docker-compose -f docker-compose-dev.yml pull eq-questionnaire-launcher
```
liamtoozer: Not sure if we need to add the hyphen here - `podman compose` and `docker compose` seem to work okay? As long as the alias or symlink from docker to podman is configured, it should be okay?

environment.yml
```diff
@@ -0,0 +1,10 @@
+name: eq-runner
+channels:
+  - conda-forge
+dependencies:
+  - python=3.13.5
```
liamtoozer: Just a heads-up that this will also need to be documented in our pinned versions Confluence page (just so we can keep tabs on dependencies)

README.md
```diff
@@ -11,27 +11,46 @@
...
 
-Make sure Colima is started every time you want to use Docker images:
+`~/.local/bin` must be on your `PATH`. Verify:
+
 ```shell
-colima start
+docker --version
+```
+
+The Makefile and the commands below use the standalone `docker-compose` binary (not the
```
liamtoozer: We might not need to do this, as per this comment: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1967/changes#r3948509228

README.md
```diff
@@ -11,27 +11,46 @@
 
 ## Run with Docker
```
liamtoozer: Very minor - does this section name need to change if we're using Podman?

README.md
```diff
@@ -11,27 +11,46 @@
...
+ln -s "$(which podman)" ~/.local/bin/docker
+hash -r
 ```
 
-Make sure Colima is started every time you want to use Docker images:
+`~/.local/bin` must be on your `PATH`. Verify:
+
 ```shell
-colima start
+docker --version
```
liamtoozer: This will return `docker version 5.8.4`, which is actually the podman version and not the docker version, which might cause some confusion? Not sure what is the best for this and open to suggestions, but I would probably vote for `alias docker='podman'` in shell profile?

README.md
liamtoozer: My vote would be for adding this as an `alias` but can maybe discuss as a team to see what's best for everyone?

README.md
```diff
@@ -11,27 +11,46 @@
...
+``` shell
+which docker-compose
+docker-compose version
 ```
 
 To get eq-questionnaire-runner running the following command will build and run the containers
 
 ``` shell
-RUNNER_ENV_FILE=.development.env docker compose up -d
+RUNNER_ENV_FILE=.development.env docker-compose up -d
```
liamtoozer: Hyphen might not be needed here and in the other places?

README.md
ems-sato: `environment.yml` sets `name: eq-runner`, no `-n` is not needed here

README.md
ems-sato: Sure, `poetry` comes via `environment.yml` so there is no separate install step but you are right, I will adjust to make it explicit

README.md
ems-sato: Yes, I can add that as a note

README.md
ems-sato: Yes, I got it confused. Will add that

README.md
ems-sato: I dont think we need it as we went for this option in the drop-in discussion: Change the Makefile to docker-compose. No per-machine setup, but this fixes only compose commands and docker build target would still need the symlink

README.md
```diff
@@ -82,50 +165,91 @@ echo "local" > .application-version
...
+
+> Note: conda-forge does not publish every Node patch release. Where the exact `.nvmrc` version is unavailable, pin the closest available patch below it and note the substitution in `environment.yml`.
+
+If `.python-version` or `.nvmrc` change, update `environment.yml` to match.
+
+Create and activate the environment:
+
+``` shell
+conda env create -f environment.yml
+conda activate eq-runner
```
liamtoozer: Not sure if it might be overkill for this readme, but may be worth mentioning that `conda init zsh` followed by `exec zsh` will need to been run as well otherwise the `conda activate eq-runner` might fail?

README.md
liamtoozer: Just wondered if this could be instead set in the `environment.yml` file in the `variables:` block?

README.md
```diff
@@ -82,50 +165,91 @@ echo "local" > .application-version
...
+on the notices cache, run **Repair ownership of user conda directory** in Self
+Service, then retry.
+
+#### Poetry
+
+Poetry must install into the conda environment rather than creating its own virtualenv.
+Set this on the environment so that no configuration file is left in the repository:
+
+``` shell
+conda env config vars set POETRY_VIRTUALENVS_CREATE=false
```
liamtoozer: Just wondered if this could be instead set in the `environment.yml` file in the `variables:` block?

liamtoozer (commented):
Great work 👍  Able to spin up Runner and is working great 🙌  Just some really minor questions/comments

README.md
liamtoozer: My vote would be for adding this as an `alias` but happy to go with the flow on this. I forgot we discussed this originally as a team 🫣

README.md
ems-sato: Yes! I dont know why I didnt think of that. Yes, that should be possible

README.md
ems-sato: Renamed to "Run with Podman"

Makefile
ems-sato: Yes, so both forms should work if you have the docker -> podman symlink. This was the option we picked at the drop-in session. The conditional approach in census31 team looks like a good alternative as well, happy to have a follow-up with the team to pick which solution we should go ahead with

README.md
ems-sato: Sure, lets agree on adding it as an alias and we can review this in a couple of months time

README.md
ems-sato: There is a troubleshooting block in pre-requisites section for when `conda` reports `command not found` (line 105). Maybe needs to move for more visibility?

environment.yml
ems-sato: Yes good point! I will review that page

README.md
ems-sato: Thanks for catching that! Fixed!

README.md
liamtoozer: Ah yes sorry you're right. Will resolve this comment 👍

README.md
ems-sato: Done, added alias to readme

Makefile
ems-sato: After discussion, have implemented census31 team conditional approach

README.md
ems-sato: Added to environment.yml and reflected the changes in the other runner repos

liamtoozer (approved):

README.md
```diff
@@ -82,50 +147,76 @@ echo "local" > .application-version
...
+#### Conda environment
+
+Python and Node.js versions are pinned in the committed `environment.yml`, matching
+`.python-version` and `.nvmrc` as closely as conda-forge availability allows:
 
-eval "$(pyenv init -)"
+
+> Note: conda-forge does not publish every Node patch release. Where the exact `.nvmrc` version is unavailable, pin the closest available patch below it and note the substitution in `environment.yml`.
+
+If `.python-version` or `.nvmrc` change, update `environment.yml` to match.
```
petechd: Sorry one last thing to make it clearer for future users of this, something like:
"If `.python-version` or `.nvmrc` change, or if you want to use another version of Python or Node, update `environment.yml` to match."

ems-sato: Yes that's much clearer, just made that change

.gitignore
```diff
@@ -45,6 +45,7 @@ docs/_build
 env
 env*
```
petechd: Can we just remove the wildcarded `env*` to prevent the `!environment.yml`?

petechd: Sorry, I am just curious about this line as it is not included in every PR for this task.

ems-sato: Yes, that makes sense, it will make it more consistent. I will make that change

petechd (approved):

